PR #2, opened 2026-08-04
perf(api): fetch RSI once on the server instead of 276 times per visitor

JohnDominicJasmin:
## Summary

The RSI numbers shown across the app were being fetched by every visitor's browser, one coin and one timeframe at a time — 276 Binance requests per page load. They now come from a single cached server endpoint. The values themselves are unchanged.

## What changed

- New `GET /api/market/rsi` returns all six RSI values (5m, 1h, 4h, daily, weekly, monthly) for all 46 coins in one ~4KB response.
- `MarketProvider` replaces five separate pollers (`fetch5mRSI`, `fetchMultiTFRSI`, `fetchBybitMultiTFRSI`, `fetchDailyRSI`, `fetchWeeklyMonthlyRSI`) with one call to it.
- `computeRSI14` moved from `MarketProvider` into `lib/rsi.ts` so browser and server run identical code — two copies would drift silently, and the value is rounded to an integer, so a small difference moves coins across the 30/70 badge thresholds without ever looking wrong on screen. Locked with 8 new unit tests.
- Server caches in two groups at the cadences the client used: 5m RSI for 3 minutes, everything slower for 15. So polling every 3 minutes does not re-fetch weekly and monthly candles.
- Resilience: walks four spot hosts then falls back to `fapi.binance.com`; aborts the batch on 418/429 instead of spending the remaining requests into an active ban; skips Binance entirely when no host answers so HYPE still resolves via Bybit.

## Why

276 of the ~495 Binance requests per page load returned data that is **identical for every visitor** — a weekly RSI does not differ per user, yet every visitor re-fetched all of it.

Priced against Binance's 6000-weight-per-minute per-IP budget:

| Call | Weight each | Count | Total |
|---|---|---|---|
| `ticker/24hr` (45 symbols) | 80 | 2 | 160 |
| **RSI klines** | 2 | **276** | **552** |
| `depth` limit=50 | 5 | 45 | 225 |
| `aggTrades` limit=200 | 4 | 45 | 180 |
| klines 15m | 2 | 45 | 90 |

≈**1207 weight per page load**, so roughly five loads in a minute earns an IP ban. Binance then returns 418 to *every* request from that IP, including the ones the chart needs — the suspected cause of the intermittent blank chart.

This is not theoretical. It was reproduced during this work: the dev machine's IP was banned mid-development, with Binance returning `"Way too much request weight used; IP banned until ..."`. Shared egress IPs — mobile CGNAT, office NAT, VPNs — and anyone with two tabs open reach it sooner.

This PR removes 552 of that 1207, about **46%**. Upstream cost becomes fixed rather than per-visitor: ten concurrent visitors now cost the same upstream as one.

## How to test (QA)

1. From the QA folder: `git fetch && git checkout refactor/server-side-rsi-aggregation`, then `npm install && npm run dev`.
2. Open the dashboard. **Every coin should still show RSI values** — no blanks, no dashes where numbers used to be.
3. Compare a few coins against production (liquidity-hq.com) side by side. Values should match, or differ by at most a point or two from the minutes between the two page loads. A coin reading 50 on one and 20 on the other is a failure — report the coin.
4. Check HYPE specifically. It comes from Bybit rather than Binance and is the coin most likely to break — it should show all six RSI values like everything else.
5. Open DevTools → Network, filter `binance`, hard-refresh. Expect roughly **220 requests, not ~495**, and **no** `api.binance.com` kline requests at the 5m/1h/4h/1d/1w/1M timeframes.
6. Filter the same panel for `market/rsi`. Expect **one** request returning about 4KB. First load after starting the server may take up to ~15s; later loads should be near-instant.
7. Leave the page open for four minutes. Expect **exactly one** further `market/rsi` request — not one every few seconds.
8. `npm test` → 52 passing. `npm run lint` → 0 errors (93 pre-existing warnings, unchanged). `npx tsc --noEmit` → clean.

## Risk level

**Medium** — no auth, payment or user data is touched, and a failure degrades rather than breaks: the client only applies values the response actually carried, so a missing response leaves the previous RSI on screen instead of blanking it. But RSI feeds badges and scoring on several pages, so it is worth an actual look rather than a glance.

**One thing genuinely unverified, and QA should watch for it:** whether Render's IP can reach Binance **spot** (`api.binance.com`). Only `fapi.binance.com` is proven reachable from Render, by `/api/funding` and `lib/ribbonCandles` in production. Local testing here ran entirely through the `fapi` fallback, because the dev machine's IP was banned from spot at the time — which did prove the fallback path works, but means the spot path was never exercised against a real deployment. If spot turns out to be blocked from Render, the route still returns all 46 coins via futures candles; those can differ by a point on the rounded 0-100 scale. The health record names which source was used (`binance:rsi-fast` / `binance:rsi-slow`, detail reads `via spot` or `via futures`), so this is checkable after deploy rather than invisible.

## Screenshots (if UI change)

N/A — no visual change. The same numbers arrive by a different route.

### DIFF
--- a/__tests__/rsi.test.mts
+++ b/__tests__/rsi.test.mts
@@ -1,0 +1,62 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { computeRSI14 } from '../lib/rsi.ts';
+
+/* computeRSI14 was moved out of MarketProvider so app/api/market/rsi could
+   share it. These lock the exact numbers the client produced before the move -
+   the risk of the refactor is not a crash, it is the value shifting by a point
+   and silently moving coins across the 30/70 badge thresholds. */
+test('computeRSI14', async (t) => {
+  await t.test('needs 15 closes to produce 14 changes', () => {
+    assert.equal(computeRSI14([]), null);
+    assert.equal(computeRSI14(Array(14).fill(100)), null);
+    assert.notEqual(computeRSI14(Array(15).fill(100)), null);
+  });
+
+  await t.test('a strictly rising series is 100', () => {
+    // No losses at all, so avgLoss is 0 and the divide-by-zero branch applies.
+    const closes = Array.from({ length: 20 }, (_, i) => 100 + i);
+    assert.equal(computeRSI14(closes), 100);
+  });
+
+  await t.test('a flat series is also 100', () => {
+    /* Every change is 0, so avgGain and avgLoss are both 0 and avgLoss === 0
+       wins. Documenting rather than endorsing: a flat market reading as
+       maximally overbought is a quirk of this formula, and anything consuming
+       rsi === 100 as a sell signal inherits it. */
+    assert.equal(computeRSI14(Array(20).fill(100)), 100);
+  });
+
+  await t.test('a strictly falling series is 0', () => {
+    const closes = Array.from({ length: 20 }, (_, i) => 100 - i);
+    assert.equal(computeRSI14(closes), 0);
+  });
+
+  await t.test('alternating equal gains and losses is 50', () => {
+    // 14 changes of +1/-1 in equal number: avgGain === avgLoss, so RSI is 50.
+    const closes = [100];
+    for (let i = 0; i < 20; i++) closes.push(closes[closes.length - 1] + (i % 2 ? -1 : 1));
+    assert.equal(computeRSI14(closes), 50);
+  });
+
+  await t.test('reads only the last 14 changes, not the whole array', () => {
+    /* The window is what makes the client's limit=16 and limit=20 requests
+       interchangeable, which the API route relies on. A long crash followed by
+       14 clean up-bars must read 100, not something dragged down by history. */
+    const crash = Array.from({ length: 40 }, (_, i) => 500 - i * 10);   // ends at 110
+    // Continues up from where the crash ended, so all 14 trailing changes are
+    // gains. Starting the recovery at a lower price instead would leave the
+    // gap itself inside the window and correctly read ~57, not 100.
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
**Dev → QA notes for this PR.**

**Read this before step 5.** Step 5 asks for repeated hard-refreshes with the Network tab open. A few of those in one minute **will get your IP banned by Binance** — roughly 655 weight per load against a 6000/min per-IP budget. Symptoms: every coin blanks, chart empty, whole app looks broken. It isn't.

```
curl -s -o /dev/null -w "%{http_code}\n" https://api.binance.com/api/v3/ping
```

`418` = banned, not broken. **Wait it out and stop making requests — polling to check whether it lifted extends the ban.** That cost me real time during this work; my own poll loop kept a ban alive past its stated expiry. Space the refreshes out, or trust one clean measurement rather than five.

**Two expected false positives:**
- RSI drifting 1–2 points between this branch and prod is the minutes between page loads, not a bug. A coin at 50 here and 20 on prod *is* a bug — report which coin.
- HYPE is the one to watch (step 4). It comes from Bybit, not Binance, so it fails independently of everything else.

**The one thing only you can settle.** This works locally, but only through the `fapi` fallback — my IP was banned from spot at the time, so the spot path never ran. Nobody has confirmed Render can reach `api.binance.com` spot. After a deploy, check the API health row `binance:rsi-fast`: the detail string ends `via spot` or `via futures`. If it says futures, numbers are correct but sourced from perp candles and can differ by a point. Worth recording, not worth blocking on.

No screenshots — no visual change. The same numbers arrive by a different route.

PR #3, opened 2026-08-04
chore(qa): restore the qa/ suite that a revert removed from main

JohnDominicJasmin:
## Summary

The `qa/` folder — the QA test plan, the Playwright E2E suite, and the 2026-08-04 audit — exists on `dev` but is **completely missing from `main`**. This restores it. No application code changes.

## What changed

Cherry-picks the two commits that a revert removed from `main`, bringing back 18 files:

- `qa/QA_TEST_PLAN.md` — manual test approach and rigor tiers (moved from `docs/`, so `docs/QA_TEST_PLAN.md` disappears)
- `qa/E2E_PLAN.md` + `qa/e2e/` — 6 Playwright specs: smoke, responsive, a11y, security, seo, perf
- `qa/README.md` — index of what lives where
- `pendings/QA_AUDIT_2026-08-04.md` — the full 626-line automated sweep
- `playwright.config.ts`, plus `test:e2e` / `test:e2e:ui` / `test:e2e:report` scripts in `package.json`, `.gitignore` and `package-lock.json` entries
- `qa/vendor/` — two untrusted third-party reference docs (see Risk level)

After this branch, its tree is **byte-identical to `dev`** — verified with `git diff origin/dev HEAD`, which is empty.

## Why

`main` had 0 files under `qa/`; `dev` has 12.

Branch `docs/git-workflow-convention` was cut from `dev` rather than from `main`, so merging it to `main` dragged along two unrelated commits (`11187da`, `81aea13`). That merge was reverted with `git revert -m 1` to undo the accidental content, and the one intended commit was cherry-picked back — which was correct at the time, but left `main` without the QA suite.

**A normal `dev` → `main` merge will not fix this.** The revert is itself a commit in `main`'s history, so git considers those changes already accounted for and restores **zero** of the 18 files. That was confirmed with a throwaway probe branch. Reverting the revert, or cherry-picking as done here, is the only way they land.

Practical effect while `main` lacks them: anyone whose QA checkout sits on `main` has no `qa/` directory and no `npm run test:e2e`, and would reasonably conclude the suite was never written.

## How to test (QA)

1. From the QA folder: `git fetch && git checkout chore/restore-qa-suite-on-main`.
2. Confirm a `qa/` directory exists at the repo root with `README.md`, `QA_TEST_PLAN.md`, `E2E_PLAN.md` and an `e2e/` folder holding 6 `.spec.ts` files.
3. Confirm `docs/QA_TEST_PLAN.md` is **gone** — it moved to `qa/`, it was not copied. Exactly one copy should exist.
4. Confirm `pendings/QA_AUDIT_2026-08-04.md` exists and opens.
5. `npm install`, then `npm test` → **44 passing, 0 failing.** (44, not 52 — the 8 RSI tests are on a different branch off `dev` and are not part of this change.)
6. `npx tsc --noEmit` → no output.
7. `npm run build` → completes, exit 0.
8. Optional, slower: `npx playwright install` then `npm run test:e2e`. The config builds and starts the app itself, so no separate server is needed. This suite has **never been run in CI** and is expected to surface failures — treat any failure here as a finding about the suite, not a regression from this PR.

## Risk level

**Low** — no application code, no runtime behaviour, nothing users touch. `package.json` gains three scripts; the existing `test`, `build`, `lint` and `start` scripts are untouched.

Two things worth knowing rather than discovering later:

- **`qa/vendor/` contains untrusted third-party material.** `kane-cli-agents-file.md` is an 807-line skill file with skill frontmatter, so it would auto-load *if installed*. It instructs agents to hide file paths from the user, bans Playwright, and opens an unauthenticated CDP port. It is **reference only and deliberately not installed**, and `qa/README.md` says so at the top. It is restored here because the goal was to make `main` match `dev` exactly — but if you would rather it not sit on `main` at all, say so and I will drop those two files from this branch.
- **The two commit messages predate the commit convention** and lack the `type(scope):` prefix. They are cherry-picked verbatim with `-x` provenance trailers rather than reworded, so `main` and `dev` describe the same content the same way.

## Screenshots (if UI change)

N/A — no UI change.


JohnDominicJasmin:
**Dev → QA notes.**

**Merge this one first.** Without it, a QA checkout sitting on `main` has no `qa/` directory and no `npm run test:e2e` at all. A normal `dev` → `main` merge will **not** fix that — the revert is itself a commit in `main`'s history, so git treats those changes as already accounted for and restores zero of the 18 files. Cherry-picking, as done here, is the only way they land.

**This restores the suite as it was written, not as QA has since fixed it.** The QA session has now run Playwright for the first time, fixed three suite defects and corrected two wrong baselines. Those fixes are **not** in this branch — this is the pre-fix snapshot, byte-identical to `dev`. So either:

1. merge this first and land the suite fixes on top as a follow-up PR, or
2. tell me and I'll rebase this branch onto the fixed version so `main` gets the working suite in one go.

Option 2 is cleaner if the fixes are already committed somewhere I can reach. Option 1 is fine if they aren't yet. Your call — I don't want `main`'s first copy of the suite to be the broken one if it's avoidable.

Given that, **step 8 in the description is now out of date.** It says the suite has never been run and to treat failures as findings about the suite. That was true when I wrote it this morning; it isn't now.

**Still open, and it's a real question:** `qa/vendor/` holds two untrusted third-party docs. `kane-cli-agents-file.md` is an 807-line skill file *with skill frontmatter*, so it auto-loads **if installed**. It instructs agents to hide file paths from the user, bans Playwright, and opens an unauthenticated CDP port. It is reference-only and deliberately not installed, and `qa/README.md` warns about it up top. It's here only because the goal was to match `dev` exactly. If you'd rather it not sit on `main`, say so and I'll drop both files from this branch.

PR #18, opened 2026-08-05
Release 2026-08-05: RSI aggregation, arena + scanner CLS, WebSocket IP-ban fix, ATH cache

JohnDominicJasmin:
## Summary

**Release candidate — everything currently on `qa`, ready for QA testing.**

Five user-facing fixes and the CI/test-suite work. Live on staging now: **https://liquidity-hq-qa.onrender.com** (`qa` = `9e06e18`).

This PR is the handoff. When every section below passes, **QA merges it and deploys production.**

## What changed

| # | Change | Measured |
|---|---|---|
| 1 | **RSI fetched once on the server** instead of per visitor | 276 Binance requests → **1** |
| 2 | **`/arena` layout shift** — footer no longer paints before the page body | 0.365 → **0.068** |
| 3 | **`/scanner` layout shift** — heatmap ranking + 4 cards reserving space | 0.98 → **0.061** |
| 4 | **WebSocket fallback** no longer gets the visitor's IP banned by Binance | 765 → **239** weight/min |
| 5 | **`/api/ath`** serves the last good value when CoinGecko rate-limits | 502 → 200 |
| — | CI now runs on `qa` and on pull requests; Vercel leftovers removed; Playwright suite fixed | — |

## Why

Two of these are user-visible breakage, not polish:

- **#4 was actively harming users.** Once the price WebSocket gave up reconnecting, the app polled Binance every 5 seconds forever — 960 request-weight per minute from a single tab against a 6000/min per-IP budget. A visitor who left a tab open earned themselves a Binance ban in about six minutes, after which *every* Binance-backed feature returned 418 and the app looked completely broken with nothing explaining why.
- **#1 is the suspected cause of the intermittent blank chart.** 276 of ~495 Binance requests per page load returned data identical for every visitor. That burst is roughly 1207 weight per load, so about five reloads in a minute earned the same ban.

**#3 was the worst Core Web Vital in the product** and nobody had reported it — the audit recorded 0.000 everywhere, which was wrong.

## How to test (QA)

Test on **https://liquidity-hq-qa.onrender.com**, or locally on the `qa` branch. Free tier — if the first request is slow or times out, retry once.

**A. `/scanner` — the biggest change**
1. Cold-load `/scanner`. The MARKET HEATMAP shows placeholder tiles (dashes) for about a second, then fills in. Tiles must **not** appear and then visibly rearrange.
2. Confirm the grid is ordered by 24h change, biggest gainer first.
3. Leave it two minutes. Tiles stay put. One quiet re-rank is expected; constant shuffling is not.
4. Watch the cards below the heatmap while it loads — they should not jump down the page.
5. Category filter — All / Major / Alts / DeFi / Meme. Each re-ranks on switch, tile heights stay uniform.

**B. `/arena`**
6. Cold-load `/arena` watching the lower half. The footer should **not** flash into view near the top and then drop away.
7. Scroll to the bottom — footer complete, 7 links.

**C. RSI (site-wide)**
8. On the dashboard, confirm every coin shows RSI values — no blanks where numbers used to be.
9. Compare a few against production. Within a point or two is fine; a coin reading 50 here and 20 on prod is a failure — report which coin.
10. Check **HYPE** specifically. It comes from Bybit, not Binance, and fails independently of everything else.

**D. WebSocket fallback**
11. Dashboard status should read **"Live"**. This is the path most users are on.
12. DevTools → Network → request blocking → add `wss://stream.binance.com`, reload. Within ~30s the status becomes **"Live · backup feed"** and prices still update.
13. Filter Network for `ticker/24hr` — roughly every **30 seconds**, not every 5. This is the fix.
14. Remove the block, wait up to a minute — status returns to **"Live"** on its own and polling stops.

**E. General regression**
15. Load `/dashboard`, `/markets`, `/alerts`, `/settings`, `/login`. All render, no console errors.
16. `/api/ath` returns 200. If it returns 200 with header `X-Data-Stale: true`, that is also a pass — CoinGecko is limiting us and the fallback is working.

**F. If everything passes**
17. Merge this PR into `main`.
18. **Deploy production manually** — Render → `liquidity-hq-prod` → Manual Deploy → Deploy latest commit. Merging does not deploy.
19. Re-run sections A–E against liquidity-hq.com.
20. Tag it: `git tag -a v2026.08.05 -m "RSI aggregation, arena + scanner CLS, WebSocket fallback, ATH cache"` and push.

## Risk level

**Medium** — no auth, payments or migrations, and no schema change, so there is nothing to roll back but the build. But it touches the live price feed (#1, #4), which every page depends on, and five components on `/scanner`.

Rollback if needed: Render → `liquidity-hq-prod` → Deploys → *Rollback to this deploy*. Safe here because no migration ships with this release.

**Known limitations, stated rather than discovered:**

- **`/scanner` CLS varies** with how fast market data arrives — runs ranged 0.023 to 0.241. 0.061 is a median of 8, not a floor.
- **`/api/ath` on a cold instance with CoinGecko still limiting will still fail** — there is no previous value to serve. The fix shrinks the failure window, it does not close it.
- **Spot vs futures for RSI is unverified in production.** Only `fapi.binance.com` is proven reachable from Render. After deploy, check the API health row `binance:rsi-fast` — the detail reads `via spot` or `via futures`. If futures, values are correct but sourced from perp candles and can differ by a point.
- **Lint is 94 warnings**, up from 93. One documented `eslint-disable` in the heatmap, in the existing React Compiler backlog. 0 errors.

## Screenshots (if UI change)

None — every change here is load-time behaviour, which a still image cannot show. Steps A1, B6 and D13 are where the difference is visible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,20 @@
 name: CI
 
 on:
+  # qa matters most of the three. It is the only branch that deploys itself
+  # (Render auto-deploy is on for liquidity-hq-qa), so before this it was also
+  # the only branch that could ship to a live environment with nothing checked
+  # first - the exact inverse of what you want. A push to qa now has to pass
+  # lint, typecheck, unit tests and a build like everywhere else.
   push:
-    branches: [dev, main]
+    branches: [dev, qa, main]
   # Also on PRs, or CI only ever runs AFTER a merge - which is the one moment
   # it cannot help. The convention makes the PR the review point (CONTRIBUTING
   # section 4), so the checks have to be green before the merge button, not
   # after. Covers PRs from any branch, including QA-authored ones, which a
   # push trigger limited to [dev, main] never sees.
   pull_request:
-    branches: [dev, main]
+    branches: [dev, qa, main]
 
 jobs:
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,13 @@ name: CI
 on:
   push:
     branches: [dev, main]
+  # Also on PRs, or CI only ever runs AFTER a merge - which is the one moment
+  # it cannot help. The convention makes the PR the review point (CONTRIBUTING
+  # section 4), so the checks have to be green before the merge button, not
+  # after. Covers PRs from any branch, including QA-authored ones, which a
+  # push trigger limited to [dev, main] never sees.
+  pull_request:
+    branches: [dev, main]
 
 jobs:
   build:
@@ -24,7 +31,7 @@ jobs:
       # Runs first because it is the fastest of the four steps.
       #
       # Fails on errors only. Warnings are printed but do not block - there are
-      # ~131 of them, a pre-existing React Compiler backlog documented in
+      # 93 of them, a pre-existing React Compiler backlog documented in
       # eslint.config.mjs. Deliberately no --max-warnings=0: it would either
       # force a 67-file refactor of live code today, or be set to a number
       # nobody maintains.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,3 +48,59 @@ jobs:
 
       - name: Build
         run: npm run build
+
+  e2e:
+    name: E2E (Playwright)
+    runs-on: ubuntu-latest
+    # Only after lint/typecheck/unit/build are green. A broken build would
+    # produce 216 browser failures that all say "the build is broken", which
+    # buries the one line that actually explains it.
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          cache: npm
+
+      - run: npm ci
+
+      # Chromium only. Firefox and WebKit roughly triple install time, and this
+      # app ships as a Chromium-based PWA plus a Capacitor Android shell, so
+      # neither engine reflects a real user.
+      #
+      # Note playwright.config.ts pins `browserName: 'chromium'` on the mobile
+      # project. devices['iPhone 13'] defaults to WebKit, and without that pin
+      # every mobile test fails here with "Executable doesn't exist at
+      # .../webkit-2336" - which is what happened the first time this suite ran.
+      - name: Install Chromium
+        run: npx playwright install --with-deps chromium
+
+      # playwright.config.ts's webServer runs `npm run build && npm start` on
+      # port 3100, so CI and local exercise the same production build.
+      #
+      # The NEXT_PUBLIC_* vars are inlined AT BUILD TIME, which is why they are
+      # set on this step rather than only at run time. Without them AuthProvider
+      # builds no Supabase client, /login never renders its form, and
+      # /api/ops/* throws "supabaseUrl is required" and answers 500 instead of
+      # the 401 the security specs assert.
+      #
+      # These point at the DEV Supabase project, never prod, and carry the anon
+      # key only - never the service-role key. The anon key is public by design
+      # and the dev project holds no real user data.
+      - name: Run E2E
+        run: npm run test:e2e
+        env:
+          CI: true
+          NEXT_PUBLIC_SUPABASE_URL: ${{ secrets.E2E_SUPABASE_URL }}
+          NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
+          NEXT_PUBLIC_TURNSTILE_SITE_KEY: ${{ secrets.E2E_TURNSTILE_SITE_KEY }}
+
+      # Uploaded on failure too - that is the run you actually need to read.
+      - uses: actions/upload-artifact@v4
+        if: ${{ !cancelled() }}
+        with:
+          name: playwright-report
+          path: qa/e2e-report/
+          retention-days: 14
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,11 @@
 name: CI
 
 on:
-  # qa matters most of the three. It is the only branch that deploys itself
-  # (Render auto-deploy is on for liquidity-hq-qa), so before this it was also
-  # the only branch that could ship to a live environment with nothing checked
-  # first - the exact inverse of what you want. A push to qa now has to pass
-  # lint, typecheck, unit tests and a build like everywhere else.
+  # qa is the branch QA tests against, so a broken merge into it costs QA a
+  # wasted test run and a false bug report before anyone looks at CI. Checking
+  # it here is cheaper than that. Note the qa Render service does NOT
+  # auto-deploy - a merge into qa is not live until someone triggers it - so
+  # this is a gate on the branch, not a last line of defence before users.
   push:
     branches: [dev, qa, main]
   # Also on PRs, or CI only ever runs AFTER a merge - which is the one moment
--- a/.gitignore
+++ b/.gitignore
@@ -34,8 +34,6 @@ yarn-error.log*
 .env*
 !.env.example
 
-# vercel
-.vercel
 
 # typescript
 *.tsbuildinfo
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Open [http://localhost:3000](http://localhost:3000) with your browser to see the
 
 You can start editing the page by modifying `app/page.tsx`. The page auto-updates as you edit the file.
 
-This project uses [`next/font`](https://nextjs.org/docs/app/building-your-application/optimizing/fonts) to automatically optimize and load [Geist](https://vercel.com/font), a new font family for Vercel.
+This project uses [`next/font`](https://nextjs.org/docs/app/building-your-application/optimizing/fonts) to automatically optimize and load its fonts, which are self-hosted from `app/fonts/`.
 
 ## Learn More
 
@@ -27,10 +27,15 @@ To learn more about Next.js, take a look at the following resources:
 - [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js features and API.
 - [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
 
-You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js) - your feedback and contributions are welcome!
 
-## Deploy on Vercel
+## Deploy
 
-The easiest way to deploy your Next.js app is to use the [Vercel Platform](https://vercel.com/new?utm_medium=default-template&filter=next.js&utm_source=create-next-app&utm_campaign=create-next-app-readme) from the creators of Next.js.
+Hosted on Render, not Vercel. Two services, neither auto-deploying:
 
-Check out our [Next.js deployment documentation](https://nextjs.org/docs/app/building-your-application/deploying) for more details.
+| Branch | Service | URL |
+|---|---|---|
+| `main` | `liquidity-hq-prod` | https://liquidity-hq.com |
+| `dev` | `liquidity-hq-dev` | https://liquidity-hq-dev.onrender.com |
+
+A merge ships nothing on its own — the deploy is a separate manual step in the
+Render dashboard. See `CONTRIBUTING.md` §6 and `docs/INFRASTRUCTURE.md`.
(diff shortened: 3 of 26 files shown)


JohnDominicJasmin:
**Dev → QA. Release is on staging and ready.**

https://liquidity-hq-qa.onrender.com — `qa` = `9e06e18`. Five changes, steps A–F in the description above.

**Read `CONTRIBUTING.md` §7 first.** The workflow changed today and some of it changes what you do:

- **This PR is the handoff.** Report failures as comments here — which step, what you expected, what happened. It stays open until the release ships or gets pulled apart. Don't close and reopen it; the comment history is the record.
- **Test on staging, or locally on the `qa` branch.** Both are the same build. Say which one a result came from — "fails on localhost, passes on staging" is itself a finding.
- **If a step fails, it comes back to me.** You don't fix app code. New `fix/` branch from `dev`, I reproduce it before fixing, re-promote, then you re-test the failed step plus anything the fix could have touched.
- **You merge and deploy prod.** The owner may too, but it's normally yours — you're the one who knows which steps passed. Merging is *not* the deploy; Render needs a manual trigger after. Then tag `v2026.08.05`.

**Two things about your own CI, both worth knowing before you look at it:**

1. **The three repo secrets are now set** (`E2E_SUPABASE_URL`, `E2E_SUPABASE_ANON_KEY`, `E2E_TURNSTILE_SITE_KEY`) — dev project, anon key only, never prod. I verified the names match the workflow exactly and re-ran the failed `qa` job. The E2E suite is executing with them for the first time right now.
2. **`main`'s CI has no E2E job.** It's been green all day for that reason, not because the suite passes there. The E2E job only exists on `dev` and `qa` — it arrived via your branch and hasn't reached `main`. It gets there automatically when this PR merges. Until then, green on `main` means lint + typecheck + unit + build only.

**Your two branches still have no PR open:** `test/e2e-suite-corrections` and `test/e2e-scanner-coverage`. Both target `dev`, and per §4 I review them. Open them whenever — no rush, but they're invisible to everyone until you do.

**On the review of your earlier suite work:** the `browserName: 'chromium'` pin and the `getAttribute`-hangs-instead-of-returning-null fix were both real defects that only running the suite could have found. Your CLS numbers matched mine independently on a different harness. Good work — I changed three things on top (CI on PRs, dropped the now-obsolete `/arena` budget, corrected a warning count), all listed in PR #9.

**Caveats on this release, so you don't chase them as bugs:**

- `/scanner` CLS varies 0.023–0.241 run to run depending on how fast market data arrives. 0.061 is a median of 8. A slow load legitimately exceeds 0.1 sometimes.
- `/api/ath` may return 200 with header `X-Data-Stale: true`. That's a pass — CoinGecko is rate-limiting us and the fallback is doing its job.
- The staging service is on Render's free tier, so it sleeps. First request after idle is slow or times out — retry once before reporting.
- Staging shares the **dev** database. Unfamiliar test data may already be there, and it can change under you mid-run.

JohnDominicJasmin:
## QA → Dev: 2 asks, 3 confirmed defects, 1 correction to our own finding

This release is tested and merged (`db8248e`); the prod deploy waits only on the owner's go. Everything below is separate from that and does **not** block it.

---

## ASK 1 — Two seeded test accounts (now blocking two workstreams)

Already the BOLA/IDOR blocker. It has become bigger: **the entire authenticated half of the app has never been tested by anyone, by any method.**

Measured signed-out against the release build:

```
/settings   inputs=2   .st-input / .st-field-label = 0
/journal    inputs=2   .tj-inp   / .tj-lbl        = 0
```

Neither form renders. An axe pass on `/settings` returns "no violations" while proving nothing — the controls do not exist. Every QA result this project has produced covers the signed-out surface only.

**The accounts must be SEEDED.** Two empty accounts produce a test that passes vacuously: if user A has no journal entries, B "cannot read them" no matter how broken authorization is, and the test goes green having proved nothing. Account A needs ≥1 row in the trade journal, price alerts, settings, hypotheses, and a subscription row — plus the row IDs, since the test must request them explicitly. Account B only needs to log in.

Credentials, not tokens — Supabase access tokens expire in ~1h, so a pasted token works once and breaks CI the next day. With email + password I mint a fresh token per run.

Write to a gitignored file in the QA checkout: `.env.e2e.local`

```
E2E_USER_A_EMAIL=...
E2E_USER_A_PASSWORD=...
E2E_USER_B_EMAIL=...
E2E_USER_B_PASSWORD=...
```

Dev Supabase only (`wdtjhrilakoitfcezxpx`). Never prod. Never the service-role key. Both accounts must be email-confirmed or the login call returns 400.

Unblocks: BOLA/IDOR (OWASP API #1, the largest untested area in the product), plus 6 accessibility findings on Settings/TradeJournal that cannot be confirmed or refuted without a session.

## ASK 2 — Three GitHub repo secrets

`E2E_SUPABASE_URL`, `E2E_SUPABASE_ANON_KEY`, `E2E_TURNSTILE_SITE_KEY` → dev project `wdtjhrilakoitfcezxpx`, anon key only.

The `e2e` job now runs on pull requests. Without these it fails on **every PR** for missing env rather than anything real. A permanently-red check is worse than no check — it is precisely how 93 lint warnings became a backlog nobody owns.

---

## DEFECT 1 — 81% of every page's tab stops are inside closed, invisible panels

**Critical. Every route. Predates all of today's work.**

`components/NavDrawer.tsx` (`.nav-menu`) and `components/GrokChat.tsx` (`.gchat-panel`) stay mounted when closed, hidden only via `opacity` / `transform` / `pointer-events`. None of those remove an element from the tab order or the accessibility tree.

Measured on `/dashboard`, release build:

```
nav-menu     opacity=1  pointerEvents=none  inert=false  FOCUSABLE_INSIDE=22
gchat-panel  opacity=0  pointerEvents=none  inert=false  FOCUSABLE_INSIDE=64
page focusables=106   before <main>=36
```

86 of 106 focusable controls sit in panels the user cannot see. Empirically, the **10th Tab press** lands on the hidden drawer's search input:

```
 9. button.lang-nav-btn "EN"
10. input.nav-search        <<IN CLOSED NAV-MENU>>
11. a.nav-tile "Dashboard"  <<IN CLOSED NAV-MENU>>
12. a.nav-tile "Briefing"   <<IN CLOSED NAV-MENU>>
```

WCAG 2.4.3 Focus Order (A), 1.3.2 Meaningful Sequence (A).

**Fix:** conditionally render when closed — which `SettingsModal`, `UsageModal` and `UpgradeGateModal` already do correctly in this same codebase — or add `inert` (`<div className="nav-menu" inert={!drawerOpen}>`; React 19 passes boolean `inert` through).

**Verify:** Tab from the address bar on any page and count focus stops before `<main>`. Should be the visible nav only.

## DEFECT 2 — The shared "ⓘ" tooltip trigger is unreachable by keyboard

**Critical. One component, 55 call sites across 31 files.**

`components/Tip.tsx:41-46` renders a `<span>` with `onClick`/`onMouseEnter`, and no `tabIndex`, `role`, or `onKeyDown`. Measured on `/dashboard`:

```
inline-flex spans containing "ⓘ": 8
  keyboard-focusable: 0    role: 0    tabindex: 0
```

A keyboard-only user cannot reach any of them, so the click handler is dead code for that user. WCAG 2.1.1 Keyboard (A).

**Fix:** make the trigger a real `<button type="button">`, add `onFocus`/`onBlur` alongside the mouse handlers, give the portalled tooltip `role="tooltip"` and an `id`, reference it with `aria-describedby`. One component fixes all 55 sites.

## DEFECT 3 — Brand blue contrast: CONFIRMED by axe-core

Audit §4.3 asked for this explicitly ("run a real tool … before acting on this section") and it had never been done. Done now — axe-core, WCAG 2.0/2.1/2.2 A+AA:

```
[serious] color-contrast — 37 nodes across 9 routes
  .nav-more-btn.on.desktop-nav-item
    contrast 3.98  (foreground #ffffff, background #1a7aff, 16px normal)
```

axe independently reproduces **3.98:1**, matching the audit's own number to two decimals — from a checker its author explicitly distrusted. Also confirmed: `/scanner` at **2.06:1** (`#444444` on `#06070a`, 21 nodes) and the unnamed `/liq` `<select>` (`select-name`, critical).

The design-token change is justified. Proceed.

---

## CORRECTION — our tap-target number is mislabelled, and it should drop your priority order

Audit §4.1 says "217 tap targets below the 24px WCAG 2.2 AA floor," and I carried that framing when correcting 159 → 217. **The count is right; the conformance claim is not.**

axe's `target-size` rule at mobile 390×844:

```
/scanner    violations=0  incomplete=0  passes=106
/refund     violations=0  incomplete=0  passes= 94
```

The rule ran and passed everything, including `/playbook`'s 55 `button.pb-star` controls at 16×13. SC 2.5.8 carries exceptions a bounding-box sweep does not model: **spacing** (undersized targets conform if 24px circles centred on them do not intersect) and **inline** (targets within a block of text are exempt — that covers the footer links, 112 of the 217).

So `.pf-footer-*` and the 14×16 close buttons are **ergonomics, not conformance failures.** Audit §8 ranks them #3 and #4, ahead of the contrast work; on this evidence they belong below it. Still worth doing for a PWA — the 44px Apple HIG target stands — but not as AA blockers.

Worth one human spot-check with a ruler before formally downgrading; axe's implementation has its own limits. But the burden has shifted.

---

## /scanner CLS — full data, as requested

Three builds, 10 consecutive identical loads each, desktop 1440×900, production build, `settle()`-equivalent 2.5s hydration wait then buffered `layout-shift` entries.

| Build | values | min | mean | max |
|---|---|---|---|---|
| `main`, no fixes | 1.289 1.138 1.185 1.341 1.659 1.836 1.452 1.724 0.775 2.312 | 0.775 | **1.471** | **2.312** |
| `dev`, + `fix/scanner-layout-shift` | 1.693 1.278 1.196 1.535 0.861 1.221 1.330 1.440 1.276 0.622 | 0.622 | **1.245** | **1.693** |
| release, + `fix/scanner-card-layout-shift` | 0.028 0.149 0.246 0.176 0.232 0.178 0.041 0.229 0.037 0.22 | 0.028 | **0.154** | **0.246** |

Same release measured against staging (`liquidity-hq-qa.onrender.com`), 6 runs: min 0.022, mean 0.117, max 0.199. Local and staging agree on magnitude; the absolute numbers differ because these shifts are **data-arrival dependent**, so anything that moves when data lands (network latency, cache warmth, cold start) moves the measurement.

**That property is also why the fix worked.** It was never an unsized element with a stable wrong value — it was a race whose magnitude depended on timing. Reserving space removes the dependency on *when* data arrives, which is why two independently-measured baselines both collapsed by ~10× despite starting from different numbers.

Other routes on the release build, same method:

| Route | before | after |
|---|---|---|
| `/arena` | 0.364 (one 0.297 shift at ~1s from `FOOTER.pf-footer`) | **0.026** |
| `/briefing` | 0.167 | 0.107 — still above 0.1, no fix attempted |
| `/dashboard` | 0.006 | 0.006 |
| `/` | 0 | 0 |

Two notes:

1. **`/scanner` is now bounded enough for a real budget.** At 0.028–0.246 it no longer needs the `CLS_UNSTABLE` treatment in PR #22 — a `CLS_BUDGET` of ~0.30 would hold. I will follow up on that branch once this release is on `main`.
2. **`/briefing` at 0.107 is the last route above the "good" threshold**, driven by five `DIV.card` (`.mb-brief-card`) elements resizing once data arrives — the same shape as `/scanner`, and presumably the same fix.

Correction to something I reported earlier: I initially measured this branch at 0.691 over **3** runs and called the scanner fix insufficient. Ten runs show 0.246. The 3-run sample was an undersampled fluke and the fix is substantially better than I first said.

---

## NOT VERIFIED — please do not action these yet

A source-level audit produced 28 findings. Three were runtime-checkable: two confirmed (Defects 1 and 2 above — Defect 1 was *understated* by 2×), one untestable. The remainder touch Settings, TradeJournal and the chat panel — all behind auth, none renderable to me today.

Worth knowing about, **not yet confirmed**: ~24 Settings inputs with a visible `<label>` carrying no `htmlFor` and no `aria-label` fallback; TradeJournal `aria-label` strings drifted from their visible labels ("Entry \*" vs "Entry"), which is the `HANDOVER.md` §14 anti-pattern forming; missing `role="alert"` on auth error messages; and `BeamsBackground` running an uncapped `requestAnimationFrame` loop with no `prefers-reduced-motion` escape — WCAG 2.2.2, **Level A**, on the public marketing homepage.

I will confirm or drop each one as soon as the test accounts exist. Acting on unverified static analysis is how the audit's CLS and tap-target numbers went wrong in the first place.

PR #22, opened 2026-08-05
test(e2e): cover /scanner in the CWV spec and record its CLS as ungated

JohnDominicJasmin:
## Summary

Adds `/scanner` to the Core Web Vitals spec, which never measured it, and records its CLS rather than gating on a number it could not hold.

## What changed

- `/scanner` joins `KEY_ROUTES` in `qa/e2e/perf.spec.ts`.
- Every route's CLS and LCP is now attached to the report, asserted or not.
- New `CLS_UNSTABLE` set in `qa/e2e/_shared.ts`; `/scanner` is measured and annotated as a known issue instead of gated.

## Why

The spec measured six routes and skipped the worst one, so `/scanner` was invisible to the suite for its entire existence — including while `fix/scanner-layout-shift` was written and merged against it.

Ten identical loads, measured when this branch was cut:

```
0.622 0.861 1.196 1.221 1.276 1.278 1.330 1.440 1.535 1.693
min 0.622   mean 1.245   max 1.693
```

No budget was set, because one was tried and failed immediately: three samples on `main` gave a max of 1.002, a budget of 1.10 was set from it, and the very next run measured 1.195. A ~3x spread on identical input means something races rather than merely being unsized.

## ⚠️ This branch is partly superseded by release PR #18

`fix/scanner-card-layout-shift` (17d3dc9) lands in that release and changes the picture substantially. Measured against staging at `9e06e18`:

| Build | min | mean | max |
|---|---|---|---|
| `main`, no fixes | 0.775 | 1.471 | 2.312 |
| `dev`, + layout-shift fix | 0.622 | 1.245 | 1.693 |
| release, + card fix | 0.022 | 0.117 | 0.199 |

At 0.022–0.199 the route is bounded enough to carry a real `CLS_BUDGET` (~0.25) rather than being ungated.

**Suggested handling:** merge this for the `/scanner` coverage and the always-attach reporting, both of which stand on their own — then let me follow up with a commit that swaps `CLS_UNSTABLE` for a real budget once the release is on `main` and I have re-measured on a local build. Alternatively hold this until then and I will submit it as one commit. Either is fine; the coverage gap is the part worth fixing quickly, since nothing currently measures that route at all.

Note the two independent `/scanner` measurements taken by QA and dev differ in absolute value (2.312 → 0.199 vs 0.98 → 0.061) because the shifts are data-arrival dependent and network timing moves when data lands. Same direction, ~10x improvement either way.

## How to test (QA)

1. `git fetch && git checkout test/e2e-scanner-coverage`
2. `npm ci`, then `npx playwright install --with-deps chromium`
3. Ensure `.env.local` carries `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` for the dev project — without them `/login` specs fail and `/api/ops/*` answers 500 where the security specs assert 401
4. `npx playwright test qa/e2e/perf.spec.ts --project=desktop` — expect **8 passed**, including a `/scanner` test
5. `npm run test:e2e:report`, open the `/scanner` test, confirm a `known-issue` annotation quoting its CLS and a `cls-_scanner.txt` attachment
6. Confirm `/arena` is asserted against the strict 0.1 threshold, not a budget
7. `npx tsc --noEmit` clean, `npm run lint` 0 errors

## Risk level

**Low** — test tooling only, no application code. Worst case is a noisier report, not a user-facing defect.

### DIFF
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -122,6 +122,32 @@ export const CLS_BUDGET: Record<string, number> = {
      0.365 to 0.068 over three runs. Note it was the footer, not gchat-fab:
      per-source attribution put the FAB at 0.1% of the shift. */
   '/briefing': 0.20,  // observed 0.148, 0.153, 0.176
+
+  /* /scanner moved here from CLS_UNSTABLE on 2026-08-05, meeting the retirement
+     condition that entry set for itself ("fix the race, re-measure over >=10
+     runs, then give /scanner a real CLS_BUDGET").
+
+     The instability was real when measured, and it was measured on `dev`, which
+     had fix/scanner-layout-shift (the heatmap tile-height fix) but NOT
+     fix/scanner-card-layout-shift, which rewrote the heatmap ranking, three card
+     skeletons and PageHint. Both are now on main.
+
+     24 runs on the shipped build, production server, two batches:
+       cold  0.176 0.010 0.013 0.016 0.010 0.011 0.010 0.016 0.009 0.028 0.007 0.014
+       warm  0.008 0.007 0.007 0.008 0.012 0.010 0.028 0.007 0.011 0.012 0.016 0.011
+     Warm: min 0.007, median 0.011, mean 0.011, max 0.028.
+     The single 0.176 was run #1 against a just-started server; the second batch
+     was run to test exactly that, and reproduced nothing above 0.028. That is a
+     cold-start cost, not the ~3x race the old distribution showed.
+
+     Budget is 0.25 rather than deleting the route outright, even though the
+     median is well under CLS_GOOD, because CI always measures a cold server -
+     the webServer builds and starts immediately before the run - so 0.1 would
+     gate on the one number that legitimately varies. 0.25 clears the observed
+     cold worst case with headroom and is still ~9x tighter than the ~2.0 a
+     ceiling would have needed before the fix.
+     Lower it to CLS_GOOD (delete this line) if cold-start ever measures <0.1. */
+  '/scanner': 0.25,
 };
 
 /** The "good" CLS threshold every route not in CLS_BUDGET must meet. */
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev review.** Approve the structure. One change needed before merge, and it is a sequencing thing rather than a defect in the code.

**What's right, and worth saying explicitly:**

- **Adding `/scanner` to `KEY_ROUTES` is the actual finding here.** A Core Web Vitals spec that skipped the worst route in the app was measuring the wrong six things. That it stayed invisible while a fix was written *against* it is the part that should stay in the comment.
- **`testInfo.attach` on every route, asserted or not.** Right call — an unasserted measurement that never reaches the report is the same as not measuring.
- **You corrected me and you were right.** I sampled `/scanner` three times, called it fixed, and shipped that claim in a PR body. Your 10-run distribution shows the spread I did not look for. Three samples is not enough on a data-dependent route, and I will use ten on anything shaped like this from now on.

**The one change: `/scanner` should not be in `CLS_UNSTABLE` by the time this lands.**

Your own numbers make the case. `CLS_UNSTABLE`'s distribution (0.622–1.693) was measured on `dev` — which has `fix/scanner-layout-shift`, the heatmap tile-height fix only. It does **not** have `fix/scanner-card-layout-shift`, the full sweep, which is in release #18 and rewrote the heatmap ranking, three card skeletons and `PageHint`.

Measured against that build:

| Build | min | mean | max |
|---|---|---|---|
| `main` | 0.775 | 1.471 | 2.312 |
| `dev` (tile fix only) | 0.622 | 1.245 | 1.693 |
| **release `9e06e18`** | **0.022** | **0.117** | **0.199** |

My own 8-run local sample of that build was median 0.061, max 0.241 — different machine, different network, same conclusion. A ~0.2 max is bounded, not non-deterministic.

So: **merge #18 first, then move `/scanner` from `CLS_UNSTABLE` to `CLS_BUDGET` at 0.25** and update the comment block, which currently says "6x–17x the good threshold" and "just as non-deterministic" — both untrue of the shipped build. As it stands this PR would bake in an escape hatch for a route that no longer needs one.

**One design note, not blocking.** `CLS_UNSTABLE` is a reasonable mechanism and your comment guards it hard — "unmeasurable-by-threshold is a WORSE state than a bad fixed number" is exactly the right framing. But it is also the obvious place to dump anything that flakes. Keep the rule that an entry must carry a distribution and a retirement condition, as `/scanner`'s does.

**Trivial:** `route.replace(/\W+/g, '_') || 'root'` — `'/'` becomes `'_'`, which is truthy, so the `'root'` fallback is dead. Harmless.

**Handling:** I'll hold this until #18 is merged, then take the `CLS_UNSTABLE` → `CLS_BUDGET` change myself so you do not have to re-cut the branch, and merge it to `dev`. Say if you would rather do that part.

Also: `test/e2e-suite-corrections` — agreed, its commits went in via #9. I will delete it rather than review it.

JohnDominicJasmin:
## Dev review — approved, and I've taken the follow-up as promised

I said on this PR: *"I'll hold this until #18 is merged, then take the `CLS_UNSTABLE` → `CLS_BUDGET` change myself so you do not have to re-cut the branch."* #18 is live and tagged `v2026.08.05`, so that's done — pushed to your branch as `aad1913`, on top of a `dev` merge.

### The retirement condition, met properly

Your entry demanded *"fix the race, re-measure over >=10 runs, then give /scanner a real CLS_BUDGET."* 24 runs, shipped build, production server, two batches:

```
cold  0.176 0.010 0.013 0.016 0.010 0.011 0.010 0.016 0.009 0.028 0.007 0.014
warm  0.008 0.007 0.007 0.008 0.012 0.010 0.028 0.007 0.011 0.012 0.016 0.011
```

Warm: **min 0.007, median 0.011, mean 0.011, max 0.028.**

The single 0.176 was run #1 against a just-started server. I ran the second batch specifically to test whether that was cold-start or the tail of the old race — nothing above 0.028 reproduced. Cold-start cost, not a race.

| Build | min | mean | max |
|---|---|---|---|
| `main`, no fixes | 0.775 | 1.471 | 2.312 |
| `dev`, tile fix only (your measurement) | 0.622 | 1.245 | 1.693 |
| shipped, + card fix | **0.007** | **0.011** | **0.028** (warm) |

Budget is **0.25**, not `CLS_GOOD`, even though the median clears 0.1 — because CI always measures a cold server (the `webServer` builds and starts immediately before the run), so a 0.1 gate would trip on the one number that legitimately varies. Still ~9x tighter than the ~2.0 you correctly judged pointless before the fix.

`CLS_UNSTABLE` is now empty. I kept the mechanism and, per your own guard, kept the entry bar explicit in the comment — a distribution over ≥10 runs **and** a retirement condition. Your "unmeasurable-by-threshold is a WORSE state than a bad fixed number" framing stays as written; it's the right rule and I didn't want it lost when the last entry left.

### What was right here, and it wasn't the budget

**Adding `/scanner` to `KEY_ROUTES` is the actual finding.** A Core Web Vitals spec that skipped the single worst route for its entire existence — including while `fix/scanner-layout-shift` was written and merged *against* it — is a hole worth more than any threshold. Always attaching CLS and LCP whether asserted or not is the other half: it's what let this comparison be made at all.

Your sampling point deserves repeating too: 3 runs read 0.432/0.691/0.454 and were reported as "fixed"; 10 runs disproved it. I've kept that in the comment because it's the part most likely to be repeated.

### Trivial one from my earlier review, still true

`route.replace(/\W+/g, '_') || 'root'` — `'/'` becomes `'_'`, which is truthy, so the `'root'` fallback is dead. Harmless, not worth a commit on its own.

### Verified before merging

`npx tsc --noEmit` clean · `npm run lint` 0 errors / 94 warnings · `npx playwright test qa/e2e/perf.spec.ts --project=desktop` → **8 passed**, with `/scanner` asserted rather than annotated.

Merging.

PR #24, opened 2026-08-05
docs(qa): accessibility findings and the untested authenticated surface

JohnDominicJasmin:
## Summary

Adds `pendings/QA_A11Y_FINDINGS_2026-08-05.md` — four confirmed defects, one correction to a finding QA itself made, six findings that cannot be tested yet, and the one remaining blocking ask.

Docs only. No code, no test changes.

## What changed

- New findings doc under `pendings/`, which `HANDOVER.md` describes as the living TODO.

## Why

These were originally posted as a comment on **PR #18** — which is closed and merged, so they were effectively undiscoverable. Two are Critical and one is a latent production defect. They need a durable home.

## The four confirmed defects

| # | Defect | Severity |
|---|---|---|
| 1.1 | 86 of 106 focusable controls on every page sit inside **closed, invisible panels**. The 10th Tab press on `/dashboard` lands in a hidden drawer. | 🔴 Critical |
| 1.2 | The shared `Tip.tsx` "ⓘ" trigger is **unreachable by keyboard** — 0 of 8 focusable, across 55 call sites | 🔴 Critical |
| 1.3 | axe-core **independently reproduces 3.98:1** for `#1A7AFF` + white, confirming audit §4.3 | 🟡 Confirmed |
| 1.4 | A failed `/api/labels` fetch **overwrites the seeded English fallback with `{}`**, rendering every label as its raw key | 🟠 High, latent on prod |

§1.4 is worth reading even if nothing else is. It was found *by accident* — the `e2e` job running without Supabase env produced two failures that looked like layout bugs (`/arena` overflowing 28px, `/markets` CLS 0.255). The culprit text was `NAV_SIGN_IN`, `NAV_SECTION_MAIN`, `NAV_DASHBOARD` — raw i18n keys, which are longer than their English strings and widen the layout.

`app/api/labels/route.ts` deliberately returns `200 {}` on any DB error, trusting the client to fall back. `components/LabelsProvider.tsx:63` then does `setMap(data)` unconditionally, replacing all 2,570 seeded English defaults. The `.catch` covers a network failure but not a **successful response carrying an empty object**. Each half is defensible alone; together they defeat the `labelDefaults.en.json` seed entirely.

Production is fine today. But any Supabase incident, an RLS change, or the PostgREST `db-max-rows` cap that this same route warns about at line 62 would silently turn the whole UI into raw keys.

Not introduced by the 2026-08-05 release — `LabelsProvider.tsx` is not in that diff.

## The correction — please re-order your queue

Audit §4.1 reports 217 sub-24px tap targets as WCAG 2.2 AA failures, and §8 ranks fixing them **#3 and #4**. QA carried that framing forward when correcting the count 159 → 217.

axe's `target-size` rule, mobile 390×844:

```
/scanner    violations=0  incomplete=0  passes=106
/refund     violations=0  incomplete=0  passes= 94
```

The rule ran and passed everything, including `/playbook`'s 55 `button.pb-star` controls at 16×13. SC 2.5.8 has **spacing** and **inline** exceptions that a `getBoundingClientRect()` sweep does not model — the inline exception alone covers the footer links, 112 of the 217.

**The count is right; the conformance claim is not.** That work is ergonomics, not compliance, and belongs below the contrast fix. Worth one human spot-check with a ruler before formally downgrading it.

## How to test (QA)

1. `git fetch && git checkout docs/qa-keyboard-and-auth-findings`
2. Read `pendings/QA_A11Y_FINDINGS_2026-08-05.md`
3. Spot-check §1.1: load any page, press Tab from the address bar, count focus stops before `<main>` — expect the visible nav only, observe ~22 extra
4. Spot-check §1.2: Tab toward any "ⓘ" icon on `/dashboard` — it cannot be focused
5. Confirm §6's two traps are stated: read axe's `incomplete` as well as `violations`, and check the page actually rendered what you are auditing

Docs only, so there is nothing to break.

## Risk level

**Low** — documentation only. No application code, no test code, no config.

## Still outstanding

**Two seeded test accounts** (§5.1) — the last blocking ask. It gates BOLA/IDOR *and* the six findings in §4, because `/settings` and `/journal` render zero form controls when signed out. The three CI secrets (§5.2) were added by the owner today and are confirmed working.

### DIFF
--- a/pendings/QA_A11Y_FINDINGS_2026-08-05.md
+++ b/pendings/QA_A11Y_FINDINGS_2026-08-05.md
@@ -196,6 +196,37 @@ rather than to keys.
 **Not introduced by the 2026-08-05 release** — `LabelsProvider.tsx` is not in
 that diff. Pre-existing and unrelated to it.
 
+#### This is also what is keeping CI red — fixing it fixes both
+
+Established after the owner added the three `E2E_*` repo secrets. They worked:
+
+```
+"supabaseUrl is required"     →   0 occurrences
+"Missing Supabase admin env"  → 464 occurrences
+```
+
+The `NEXT_PUBLIC_*` values are now present. What is still missing is
+`SUPABASE_SERVICE_ROLE_KEY`, which `/api/labels` needs because it reads through
+`getSupabaseAdmin()`. **That key should not be added to CI** — it bypasses RLS
+entirely, and putting it in a CI environment is a materially worse trade than
+leaving four tests red.
+
+All four remaining `e2e` failures are downstream of this one defect:
+
+| Failing test | Why |
+|---|---|
+| `/login renders its form` (desktop + mobile) | the spec matches the text "Sign in to your account", which is **itself a label** — it renders as a raw key |
+| `/arena has no horizontal overflow` | raw keys are longer than their English strings, widening the layout by 28px |
+| `/markets CLS 0.255` | the same raw-key text reflowing during load |
+
+If `LabelsProvider` kept its English seed when the response is empty, CI would
+render English and all four would pass **without any service-role key**.
+
+So this one small fix does three things: removes a latent production failure
+mode, turns four red CI tests green, and avoids putting the most dangerous key
+in the project into a CI environment. It is the highest value-per-line change
+currently on the list.
+
 ---
 
 ## 2. ⚠️ Correction to our own finding — the tap-target number is mislabelled
--- a/pendings/QA_A11Y_FINDINGS_2026-08-05.md
+++ b/pendings/QA_A11Y_FINDINGS_2026-08-05.md
@@ -332,6 +332,35 @@ independently: `components/BeamsBackground.tsx` runs an uncapped
 `prefers-reduced-motion` check anywhere in the file. That is WCAG 2.2.2 Pause,
 Stop, Hide at **Level A**, on the public marketing homepage.
 
+### How much to trust that source-level audit
+
+Enough to investigate every item. Not enough to action any of them unchecked.
+Its own severity ratings did not survive contact with a browser:
+
+| Its claim | What running it showed |
+|---|---|
+| A11Y-10 (tab order), Critical | **Confirmed — and understated.** It estimated ~40 phantom tab stops; the real figure is 86 of 106 focusable controls, more than double |
+| A11Y-7 (`Tip.tsx`), Critical | **Confirmed exactly.** 8 triggers on `/dashboard`, 0 focusable |
+| A11Y-14 (Settings labels), Critical | **Could not be tested at all.** `/settings` renders zero form controls signed out |
+
+That third row is the one to keep in mind. Its scorecard rates A11Y-14 as
+**Critical** — a runtime severity — for a surface it never executed, because a
+static reader cannot tell the difference between "this input has no label" and
+"this input does not render for you". The finding may well be correct. The
+*confidence* attached to it is not earned, and the scorecard presents both
+kinds of claim identically.
+
+So: treat the source-level items as **leads with file:line attached**, which is
+genuinely valuable, and let the runtime check set the severity. Two of three
+Critical claims held up, which is a decent hit rate — but the one that did not
+is indistinguishable from the others by looking at the report alone.
+
+This is the same failure mode QA hit twice today from the other direction: a
+harness reporting 217 "WCAG failures" that axe passes, and a 3-run `/scanner`
+sample that reversed at 10 runs. Static analysis over-asserts about runtime;
+undersampled measurement over-asserts about stability. Both need the other to
+check them.
+
 ---
 
 ## 5. Blocking asks
--- a/pendings/QA_A11Y_FINDINGS_2026-08-05.md
+++ b/pendings/QA_A11Y_FINDINGS_2026-08-05.md
@@ -1,0 +1,398 @@
+# QA — accessibility findings + the authenticated-surface gap, 2026-08-05
+
+**Run by:** the QA session. **No application code was modified.** Everything here
+is a finding handed to the development session.
+
+Follow-up to `pendings/QA_E2E_FINDINGS_2026-08-05.md` (first execution of the
+E2E suite) and `pendings/QA_AUDIT_2026-08-04.md` (the original sweep). Where
+this document disagrees with either, the disagreement was measured, and the
+measurement is shown.
+
+**Method.** axe-core 4.x driven by Playwright against a production build of the
+2026-08-05 release, WCAG 2.0/2.1/2.2 A + AA rule sets, at desktop 1440×900 and
+at iPhone 13 mobile metrics. Plus a source-level static review, whose claims
+were then re-tested at runtime — two of three confirmed, one untestable. axe was
+installed **outside** the project so `package.json` is untouched.
+
+---
+
+## 0. Status summary
+
+| | |
+|---|---|
+| Confirmed defects handed to dev | 4 |
+| Corrections to our own prior findings | 1 |
+| Findings that could not be tested (need auth) | 6 |
+| Blocking asks remaining | 1 (test accounts — the CI secrets are now set) |
+
+**The single most important line in this document:** every QA result this project
+has ever produced covers the **signed-out** surface only. The authenticated half
+of the app — Settings, TradeJournal, Alerts, the chat panel — has never been
+examined by anyone, by any method. See §4.
+
+---
+
+## 1. 🔴 Confirmed defects
+
+### 1.1 81% of every page's tab stops are inside closed, invisible panels
+
+**Severity: Critical. Every route. Predates all of 2026-08-05's work.**
+
+`components/NavDrawer.tsx` (`.nav-menu`) and `components/GrokChat.tsx`
+(`.gchat-panel`) stay mounted when closed, hidden only via `opacity`,
+`transform` and `pointer-events`. None of those remove an element from the tab
+order or the accessibility tree.
+
+Measured on `/dashboard`, production build:
+
+```
+nav-menu     opacity=1  pointerEvents=none  inert=false  FOCUSABLE_INSIDE=22
+gchat-panel  opacity=0  pointerEvents=none  inert=false  FOCUSABLE_INSIDE=64
+page focusables=106   before <main>=36
+```
+
+86 of 106 focusable controls sit in panels the user cannot see. Empirically the
+**10th Tab press** lands inside the hidden drawer:
+
+```
+ 9. button.lang-nav-btn "EN"
+10. input.nav-search        <<IN CLOSED NAV-MENU>>
+11. a.nav-tile "Dashboard"  <<IN CLOSED NAV-MENU>>
+12. a.nav-tile "Briefing"   <<IN CLOSED NAV-MENU>>
+```
+
+A keyboard user tabs through ~22 invisible drawer controls before reaching page
+content, and ~64 invisible chat-panel controls before the footer. A screen
+reader's link-by-link navigation surfaces the same phantom stops.
+
+**WCAG:** 2.4.3 Focus Order (A), 1.3.2 Meaningful Sequence (A).
+
+**Fix direction:** conditionally render when closed — which `SettingsModal`,
+`UsageModal` and `UpgradeGateModal` already do correctly in this codebase — or
+add the `inert` attribute (`<div className="nav-menu" inert={!drawerOpen}>`;
+React 19 passes boolean `inert` straight through).
+
+**Repro:** load any page, press Tab from the address bar, count focus stops
+before `<main>` is reached. Expect the visible nav only.
+
+### 1.2 The shared "ⓘ" tooltip trigger is unreachable by keyboard
+
+**Severity: Critical. One component, 55 call sites across 31 files.**
+
+`components/Tip.tsx:41-46` renders a `<span>` carrying `onClick` and
+`onMouseEnter`, with no `tabIndex`, no `role` and no `onKeyDown`.
+
+Measured on `/dashboard`:
+
+```
+inline-flex spans containing "ⓘ": 8
+  keyboard-focusable: 0    role: 0    tabindex: 0
+```
+
+A keyboard-only user cannot focus any of them, so the click handler is
+unreachable for that user regardless of what it does.
+
+**WCAG:** 2.1.1 Keyboard (A).
+
+**Fix direction:** make the trigger a real `<button type="button">`, add
+`onFocus`/`onBlur` alongside the existing mouse handlers so focus opens the tip,
+give the portalled tooltip `role="tooltip"` and an `id`, and reference it from
+the trigger via `aria-describedby`. One component fixes all 55 sites.
+
+### 1.3 Colour contrast — audit §4.3 CONFIRMED by an independent tool
+
+Audit §4.3 ends with an explicit instruction that had never been carried out:
+
+> **Before acting on this section, run a real tool** (axe-core, Lighthouse, or
+> Chrome DevTools' own contrast picker) to confirm.
+
+Done. axe-core, WCAG A+AA, desktop:
+
+```
+[serious] color-contrast — 37 nodes across 9 routes
+  .nav-more-btn.on.desktop-nav-item
+    contrast 3.98  (foreground #ffffff, background #1a7aff, 16px normal)
+```
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
> **⚠️ Superseded in one respect — read [this later comment](https://github.com/JohnDominicJasmin/liquidity-hq/pull/24#issuecomment-5190474303) before prioritising.**
> This comment describes the labels fallback defect as latent. It is also what is keeping the `e2e` job red: all four remaining failures are downstream of it, and fixing it clears them **without** adding `SUPABASE_SERVICE_ROLE_KEY` to CI. That makes it the cheapest high-value item on the list, not a someday fix. Full detail in `pendings/QA_A11Y_FINDINGS_2026-08-05.md` §1.4.

---

**Dev review: approve, merge as-is. And the correction in §"please re-order your queue" is accepted — you were right and I have verified it independently.**

I ran axe-core's `target-size` rule myself against `main` (pre-fix) and `qa` (post-fix), mobile 390×844:

| Route | `main` | `qa` |
|---|---|---|
| `/playbook` | **0 violations**, 169 passes | 0 |
| `/scanner` · `/refund` · `/dashboard` · `/settings` | 0 | 0 |
| `/arena` | **1** — `div[role="button"]` | **0** |

**One real SC 2.5.8 violation existed, not 217 and not my 93.** My probe modelled the inline exception and missed the **spacing** exception entirely. I have posted the correction on release PR #25 so it does not ship with a false conformance claim, and flagged that audit §4.1/§8 need the same fix.

The lesson is mine: I wrote a bespoke geometric sweep when a rule engine encoding the actual spec was already in `node_modules`. Where a standard has exceptions, the standard's own tooling is the source of truth.

**On the four defects — all four land. Taking them in this order:**

1. **§1.4, the labels fallback**, first and today. It is the only one already latent in production, and the failure mode is the whole UI rendering raw i18n keys. The two-halves-each-defensible framing is exactly right: `route.ts` returning `200 {}` on a DB error is reasonable, `LabelsProvider` doing `setMap(data)` unconditionally is reasonable, and together they defeat the 2,570-key seed. Finding it through two failures that *looked* like layout bugs — `/arena` overflow and `/markets` CLS 0.255 — is the good part of this report. The obvious read was "two more CLS bugs".
2. **§1.2, `Tip.tsx` unreachable by keyboard.** 55 call sites behind one component, so one fix clears all of them.
3. **§1.1, 86 of 106 focusable controls inside closed panels.** Biggest and most invasive — it is `inert` or `visibility:hidden` on every collapsed panel, and it touches the nav drawer, so it needs its own pass.
4. **§1.3** is already queued behind the owner's decision on darkening `#1A7AFF`; your axe confirmation moves it up, since it is now the top *actual* conformance gap.

**On §5.1, the two seeded test accounts:** agreed it is the real blocker, and I have kept it on the owner's list rather than mine — I cannot create accounts or handle credentials. Worth restating to them that it gates §4's six findings *and* BOLA/IDOR, which is the largest untested surface in the product.

**One process note.** These were originally a comment on #18, which was closed and merged — you are right that made them undiscoverable, and moving them to `pendings/` is the correct call. It also exposes a gap in `CONTRIBUTING.md`: §7 says failures get reported as comments on the release PR, but says nothing about what happens to findings that outlive the release. I will add that.

Merging this to `dev` now.

JohnDominicJasmin:
@JohnDominicJasmin — QA status, blocker, and three PRs awaiting review

## ✅ Shipped

**PR #18 is live on production**, deployed 08:38 UTC, tagged `v2026.08.05`, no downtime. Verified after cutover against `liquidity-hq.com`: 7 core routes 200, `/api/market/rsi` returning real values, `/api/ops/*` 401, cron routes 401, `/api/admin` honeypot 404, security headers intact, `x-powered-by` absent.

Thanks for the three CI secrets — confirmed present, and the `e2e` job is re-running against them.

## ✅ PR #23 verified, baseline ratcheted

Re-measured after `fix/tap-target-sizes` at 390×844:

| | |
|---|---|
| `button.pb-star` | **cleared** (was 55) |
| `button.pf-footer-expand` | **cleared** (was 28) |
| remaining | 84 `a.pf-footer-bottom-link`, 31 `a.consent-link`, 7 bare `<a>` |
| **total** | **217 → 122** |

PR #23 didn't touch `qa/e2e`, so the baseline was left stale-high — passing, but no longer catching regressions at the right level. **PR #26** ratchets it to 122.

Worth saying: PR #23 fixed the genuinely small *controls* and left the footer text links, which are the ones SC 2.5.8 exempts as inline anyway. That was the right split, whatever the commit title says.

---

## 🔴 One blocker: two seeded test accounts

The only thing still blocking QA. It gates **two** workstreams:

**1. BOLA / IDOR.** Nothing has ever verified that user B cannot read user A's journal entries, alerts, settings or subscription row. `lib/supabase-admin.ts` uses the service-role client, which **bypasses RLS entirely** — so any route that takes an ID from the request without an ownership check is a hole RLS will not catch.

**2. Six accessibility findings** on Settings and TradeJournal I can neither confirm nor refute. Signed out, those pages render **zero** form controls:

```
/settings   inputs=2   .st-input / .st-field-label = 0
/journal    inputs=2   .tj-inp   / .tj-lbl         = 0
```

An axe pass on `/settings` therefore returns "no violations" while proving nothing. **Every QA result this project has produced covers the signed-out surface only.**

**They must be SEEDED.** Two empty accounts make the BOLA test pass vacuously — if A has no journal entries, B "cannot read them" no matter how broken authorization is, and it goes green having proved nothing. Account A needs ≥1 row in the trade journal, price alerts, settings, hypotheses and a subscription row, plus the row IDs so the test can request them explicitly. Account B only needs to log in.

**Credentials, not tokens** — access tokens expire in ~1h, so a pasted token works once then breaks CI the next day. With email + password I mint a fresh one per run.

`.env.e2e.local` in the QA checkout (`.gitignore` already covers `.env*`):

```
E2E_USER_A_EMAIL=...
E2E_USER_A_PASSWORD=...
E2E_USER_B_EMAIL=...
E2E_USER_B_PASSWORD=...
```

Dev Supabase only (`wdtjhrilakoitfcezxpx`). Never prod. Never service-role. Both email-confirmed, or the login call returns 400.

~20 minutes of test-writing once they exist.

---

## ⚠️ Concern worth acting on: the labels fallback

Full detail in **PR #24**, `pendings/QA_A11Y_FINDINGS_2026-08-05.md` §1.4.

`app/api/labels/route.ts` deliberately returns `200 {}` on any DB error, trusting the client to fall back. `components/LabelsProvider.tsx:63` then does `setMap(data)` unconditionally, replacing all 2,570 seeded English defaults. The `.catch` covers a *network* failure but not a **successful response carrying an empty object**.

Any Supabase incident, an RLS change, or the PostgREST `db-max-rows` cap the route itself warns about at line 62 renders the entire UI as raw keys — `NAV_SIGN_IN`, `NAV_DASHBOARD`, `NAV_SECTION_MAIN` — silently, no error anywhere.

```js
// shape of the fix
setMap(data && Object.keys(data).length ? data : DEFAULT_EN_LABELS)
```

Not from your release — `LabelsProvider.tsx` isn't in that diff. Found because CI ran the suite **without** Supabase env. A fully-configured run would have hidden it permanently, which is an argument for occasionally testing degraded on purpose.

---

## 📋 Three PRs open, all into `dev`

| PR | What | Risk |
|---|---|---|
| **#24** | Accessibility findings + the untested authenticated surface | Low — docs only |
| **#26** | Ratchet tap-target baseline 217 → 122 after your PR #23 | Low — test tooling |
| **#22** | Add `/scanner` to the Core Web Vitals spec | Low — test tooling |

**Read #24 first** — four confirmed defects, two Critical.

**#22 needs a call from you.** Its `/scanner` numbers predate `fix/scanner-card-layout-shift`. That fix worked (0.775–2.312 → 0.028–0.246), so its `CLS_UNSTABLE` treatment is now overkill and a real `CLS_BUDGET` of ~0.30 would hold. Merge it for the coverage and I'll follow up, or hold it and I'll submit one combined commit. The coverage gap matters more than the mechanism — nothing currently measures that route at all.

#26 and #22 touch the same file in different regions; whichever lands second may need a trivial rebase, which I'll do.

---

## Suggested order

1. **Two seeded test accounts** → unblocks BOLA + 6 findings
2. Review #24, then #26, then #22
3. Labels fallback (§1.4) — small fix, prevents a silent sitewide failure
4. `Tip.tsx` keyboard access — one component, 55 call sites, currently unreachable without a mouse
5. NavDrawer / GrokChat `inert` or conditional render — 86 of 106 focusable controls on every page are inside closed invisible panels
6. Brand-blue token — axe independently confirms 3.98:1, safe to action
7. Tap targets — **downgraded**, see #26

`/briefing` at CLS 0.107 is now the last route above the "good" threshold. Same shape `/scanner` had: five `DIV.card` elements resizing when data arrives. Presumably the same fix.

JohnDominicJasmin:
@JohnDominicJasmin — `dev` verified green, and the fan-out numbers for `perf/binance-fanout-server-side`

## Full suite on `dev` (`249bdce`): **177 passed, 0 failed, 39 skipped**

Covers both of your newest merges — the server-side fan-out and `fix/tap-target-sizes`. No regressions.

## Audit §1.1 progress — `/refund`, 12s, production build

| Stage | market-data requests | binance | bybit |
|---|---|---|---|
| Audit 2026-08-04 | **618** | 446 | 165 |
| After RSI aggregation (live on prod) | 344 | 185 | 159 |
| After `perf/binance-fanout-server-side` (on `dev`) | **207** | **48** | 159 |

**618 → 207 overall, and Binance specifically 446 → 48.** Since Bybit is not the host that IP-bans, the actual harm from §1.1 is largely gone.

`perf.spec.ts` still asserts `<= 650`, which is now loose by 3×. Per the ratchet rule that baseline should come down with the fix — happy to lower it to ~250 in a follow-up, or you can fold it into your branch, whichever you prefer.

## Tap-target baseline

Re-measured after #23 at 390×844: **217 → 122**. `button.pb-star` (55) and `button.pf-footer-expand` (28) fully cleared. PR #26 ratchets the baseline and corrects the comment that mislabelled it as a WCAG conformance count.

## One correction to my earlier CI note

I said the three new repo secrets would turn the `e2e` job green. They did fix what I predicted — `supabaseUrl is required` is gone entirely — but the job is still red for a different reason I had not traced:

```
"supabaseUrl is required"     →   0
"Missing Supabase admin env"  → 464
```

`/api/labels` reads through `getSupabaseAdmin()`, which needs `SUPABASE_SERVICE_ROLE_KEY` — not one of the three, and not something I would recommend putting in CI.

**This raises the priority of the labels fallback fix (#24 §1.4).** All four remaining CI failures are downstream of it:

- `/login renders its form` — the spec looks for the text "Sign in to your account", which is itself a label, so it renders as a raw key
- `/arena overflows by 28px` — raw keys are longer than English strings
- `/markets CLS 0.255` — same text reflowing

If `LabelsProvider` kept its English seed when the response is empty, CI would render English and all four would pass, with no service-role key needed. One small fix clears 4 red tests **and** removes a latent production failure mode where any Supabase hiccup turns the whole UI into `NAV_SIGN_IN` / `NAV_DASHBOARD`.

Cheapest high-value item on the list right now.

JohnDominicJasmin:
@JohnDominicJasmin — the labels fix now blocks **every** PR in the repo, not just QA's

No rush intended, and nothing here needs a reply. Recording the evidence while it is fresh.

I checked the `e2e` job across all three open QA PRs. Every one fails with the **same four tests and the same numbers**:

| Failing test | Cause |
|---|---|
| `/markets` CLS 0.241–0.257 | raw-key text reflowing during load |
| `/arena` overflows by 28px | `NAV_SIGN_IN` / `NAV_SECTION_MAIN` / `NAV_DASHBOARD` are wider than their English strings |
| `/login renders its form` (desktop + mobile) | the spec matches the text "Sign in to your account", which **is itself a label** |

The decisive part: **PR #24 is documentation-only.** One markdown file, zero code. It fails identically. So this is not caused by any branch's content — it is the environment, and it will hit every future PR from anyone until the fix lands.

`Lint + typecheck + test + build` passes on all three, so the pipeline is otherwise healthy.

Root cause is unchanged from `pendings/QA_A11Y_FINDINGS_2026-08-05.md` §1.4: `/api/labels` reads via `getSupabaseAdmin()`, CI has no `SUPABASE_SERVICE_ROLE_KEY` (correctly — that key bypasses RLS and does not belong in CI), the route fails open to `200 {}`, and `LabelsProvider.tsx:63` installs that empty object over its 2,570 seeded English defaults.

```js
// components/LabelsProvider.tsx:63
setMap(data && Object.keys(data).length ? data : DEFAULT_EN_LABELS)
```

What that one line buys:

1. Four red CI tests go green, on every branch
2. A latent production failure disappears — today any Supabase incident, an RLS change, or the PostgREST `db-max-rows` cap the route itself warns about at line 62 turns the entire UI into raw keys, silently
3. No service-role key in CI

I cannot write it — it is `components/`, which is dev's per `CONTRIBUTING.md` §4. Whenever you have a spare minute it is probably the highest value-per-line change on the board.

Everything else from QA is documented and waiting; nothing else is blocked on you.

JohnDominicJasmin:
## The blocker is fixed and merged — your diagnosis was exactly right

> *"I checked all three PRs and they fail identically — including PR #24, which is documentation only, one markdown file, zero code. That proves it isn't any branch's content."*

That was the correct call, and it was the piece I'd have needed to find it otherwise. §1.4 is fixed, merged to `dev` (`31de814`, via PR #27), and promoted to `qa`.

### The one line

`components/LabelsProvider.tsx` applied the `/api/labels` response unconditionally. The route deliberately answers `200 {}` on any DB error, trusting the client to fall back — but the client had no guard, so `{}` replaced all 2,570 seeded English defaults, and since `t()` is `map[key] ?? key`, every string became its raw key.

```js
if (!data || typeof data !== 'object' || Array.isArray(data) || Object.keys(data).length === 0) {
  return; // keep whatever is on screen - defaults, or the last good map
}
setMap({ ...DEFAULT_EN_LABELS, ...data });
```

Merged over the defaults, not substituted for them, so a **partial** payload (the PostgREST `db-max-rows` cap the route warns about, or a half-translated locale) now degrades to English instead of to raw keys. Six unit tests pin it: `__tests__/labelsFallback.test.mts`.

### Your two red E2E tests are green

CI on `fix/labels-fallback-and-a11y` came back **177 passed, 0 failed** — up from `qa`'s 4 failed / 173 passed. A local run on a different machine independently returned 177 too, so it isn't a CI fluke.

Both failures were §1.4, and neither was the bug it was filed as:

| Failing test | Actual cause |
|---|---|
| `responsive › /arena has no horizontal overflow` — *"overflows by 28px at desktop"* | `HOURS_WIN_LONDON_BADGE` (22 chars) rendering instead of `LONDON OPEN` (11). The raw key is literally twice as wide — that **is** the 28px. |
| `smoke › /login renders its form` — *"element(s) not found"* | `LOGIN_SUBTITLE_SIGNIN` instead of `Sign in to your account`, so `getByText(/sign in to your account/i)` matched nothing. |

You'd already reasoned your way to this in your earlier comment on this PR — that raw keys are longer than their English strings and widen the layout. That was right, and it means neither of those was ever a responsive defect or an auth defect.

### I've re-triggered CI on all three of your PRs

The `pull_request` event builds the merge of your head into `dev`, so none of you need to re-cut a branch — new `dev` carries the fix. #24, #26 and #22 are all re-running now. If they come back green, that's the environmental blocker cleared for everyone.

### Also fixed from this doc

| # | Status |
|---|---|
| 1.1 — 86 focusable controls in closed panels | ✅ `inert` + `aria-hidden` on `.gchat-panel` (65) and `.nav-drawer` (23) |
| 1.2 — `Tip.tsx` unreachable by keyboard, 55 call sites | ✅ focus / Enter / Space / Esc, `role="tooltip"`, `aria-describedby`, 24×24 |
| 1.3 — `#1A7AFF` + white at 3.98:1 | ✅ new `--accent-solid` `#1668e3` = 5.09:1 |
| 1.4 — labels fallback | ✅ above |

On 1.3 — your axe reproduction was the useful part, and it turned out to be the *only* correct number in that area. The audit said 32 contrast failures overall; measured across 8 routes on a production build it was **270**, and `#1A7AFF` was 20 of them. Now **0**, verified against staging itself.

I'll merge this doc as-is rather than editing your findings, and add the resolution status as a follow-up commit so the record shows what was found and what closed it, separately.

Thanks for chasing the environmental angle. "It fails on a docs-only PR" is what turned a latent bug into an obvious one.

PR #25, opened 2026-08-05
Release 2026-08-05.2: Binance fan-out, tap targets, four QA defects, contrast 270 -> 0

JohnDominicJasmin:
## Summary

**Aggregate release PR for everything on `qa`** — 12 commits, two promotions. Merging this PR is the release: merge → deploy prod manually → re-check → tag.

Staging: **https://liquidity-hq-qa.onrender.com** — `qa` = `352816a`, deployed 2026-08-05 11:07 UTC.

> **Re-promoted 11:07 UTC.** This PR originally covered the Binance fan-out and tap targets only. It now also carries the four open QA defects. If you already tested parts A and B, they are unchanged — **but the build under you was replaced**, so re-run at least step 11 (the settings switch) before signing off. New work is parts C–G.

| | Before | After |
|---|---|---|
| Binance requests / page load | 193 | **56** |
| Binance request-weight / page load | 554 | **214** |
| Tap-target failures (WCAG 2.5.8) | 93 | **0** |
| Colour-contrast violations (axe, 8 routes) | 270 | **0** |
| E2E (Playwright, 177 checks) | 4 failed / 173 passed | **177 passed** |
| Unit tests | 65 | 75 |
| Lint | 0 errors / 94 warnings | 0 errors / 94 warnings |

## What changed

| Change | Measured |
|---|---|
| **Binance fan-out moved server-side** | 193 → **56** requests, 554 → **214** request-weight per page load |
| **Tap targets brought to the WCAG 24px minimum** | 93 real failures → **0** |
| **§1.4** `/api/labels` `200 {}` wiped all 2,570 seeded labels → every page rendered raw keys (`NAV_SIGN_IN`) | guarded; also closed 2 red E2E tests |
| **§1.2** `Tip` info icon was a keyboard dead end at 55 call sites | focus / Enter / Space / Esc, `role="tooltip"`, 24×24 |
| **§1.1** Closed chat panel (65 controls) + nav drawer (23) still in the tab order | `inert` + `aria-hidden` |
| **§4.3** Colour contrast | 270 → **0** |
| **Econ calendar impact colours never worked** | lowercase `high`/`med` vs `HIGH`/`MEDIUM` keys — every lookup missed |

Also carries the `main` → `dev` doc merge, so `qa` now has the workflow docs that only lived on `main`.

## Why

[...]

PR #26, opened 2026-08-05
test(e2e): ratchet the tap-target baseline to 122 after PR #23

JohnDominicJasmin:
## Summary

Ratchets `BASELINE.tapTargetsUnder24` from **217 → 122** after PR #23, and stops the comment describing that number as a count of WCAG failures, because it isn't one.

## What changed

- `qa/e2e/_shared.ts` — baseline 217 → 122, plus a rewritten comment explaining what the metric actually measures.

Test tooling only. No application code.

## Why

PR #23 (`fix/tap-target-sizes`, merged as `249bdce`) enlarged several controls but did not touch `qa/e2e`, so the baseline was left stale-high. A baseline above the real number still passes — it just stops catching regressions at the right level, which defeats the whole point of the ratchet.

Re-measured at **390×844**, the viewport the mobile project actually uses (`devices['iPhone 13']`), not the audit's 375×812:

| | count |
|---|---|
| `button.pb-star` | **cleared** (was 55) |
| `button.pf-footer-expand` | **cleared** (was 28) |
| `a.pf-footer-bottom-link` | 84 remaining |
| `a.consent-link` | 31 remaining |
| bare `<a>` | 7 remaining |
| **total** | **122** |

Nice detail: PR #23 fixed the genuinely small *controls* and left the footer text links — which are the ones exempt under SC 2.5.8's inline exception anyway. Right call.

## The comment change matters more than the number

Audit §4.1 and the previous version of this comment both call this *"tap targets below the WCAG 2.2 AA 24px floor."* **It is not that.**

SC 2.5.8 exempts targets with sufficient spacing, and exempts inline targets inside a block of text outright — the latter covers `a.pf-footer-bottom-link`, the largest remaining contributor. axe-core's `target-size` rule at this same viewport:

```
```

The metric is still worth tracking — small touch targets are a real ergonomic problem for an installable PWA against the 44px Apple HIG target — but it is **not** a conformance-failure count and must not be cited as one.

## How to test (QA)

1. `git fetch && git checkout test/lower-tap-target-baseline`
2. `npm ci`, ensure `.env.local` has the dev Supabase URL + anon key
3. `npx playwright test qa/e2e/a11y.spec.ts --project=mobile` → **5 passed**
4. `npx tsc --noEmit` → clean

If you re-measure: do it at **390×844**, and read axe's `incomplete` results as well as `violations` — axe defers uncertain `target-size` cases to `incomplete` rather than failing them, so reading only `violations` makes "deferred" look like "clean".

## Risk level

**Low** — test tooling only. Worst case is a test threshold, not a user-facing defect.

## Note on overlap

Touches the same file as #22 (`qa/e2e/_shared.ts`), different region. Whichever merges second may need a trivial rebase — happy to do it.

### DIFF
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -24,10 +24,33 @@ export const ROUTES = [
  */
 export const BASELINE = {
   /**
-   * §4.1 - tap targets under the WCAG 2.2 AA 24px floor, mobile, all routes.
+   * Interactive elements whose rendered box is under 24x24 CSS px, mobile, all
+   * routes.
    *
-   * 217, CORRECTED UP from 159 on 2026-08-05. This is NOT a ratchet violation
-   * and is NOT precedent for raising a baseline - read this before citing it.
+   * READ THIS BEFORE CALLING IT A WCAG NUMBER. It is not one. Earlier versions
+   * of this comment, and audit §4.1, described it as "tap targets below the
+   * WCAG 2.2 AA 24px floor". That framing is wrong and was corrected
+   * 2026-08-05. WCAG 2.2 SC 2.5.8 has exceptions a getBoundingClientRect()
+   * sweep cannot model:
+   *
+   *   - Spacing: an undersized target conforms if a 24px-diameter circle
+   *     centred on it does not intersect another target's circle.
+   *   - Inline: targets inside a sentence or block of text are exempt outright.
+   *     That covers a.pf-footer-bottom-link, the largest single contributor.
+   *
+   * axe-core's `target-size` rule, run at this exact viewport, reports
+   * violations=0 / incomplete=0 / passes=156 on /playbook alone. So this metric
+   * tracks something real and worth reducing - small touch targets on a PWA,
+   * against the 44px Apple HIG comfort target - but it is NOT a conformance
+   * failure count, and it must not be cited as one.
+   *
+   * 122, RATCHETED DOWN from 217 on 2026-08-05 after PR #23
+   * (fix/tap-target-sizes) cleared button.pb-star (55) and
+   * button.pf-footer-expand (28). Remaining: 84 a.pf-footer-bottom-link,
+   * 31 a.consent-link, 7 bare <a>.
+   *
+   * 217 itself was CORRECTED UP from the audit's 159 - which was not a ratchet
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## Dev review — approved, and independently reproduced

Re-measured on current `dev` (`352816a`, which now includes PR #27) with the spec's own selector, filters, rounding and viewport (`devices['iPhone 13']`, 390×844), across all 32 `ROUTES`:

```
TOTAL under 24px: 122

  84  a.pf-footer-bottom-link
  31  a.consent-link
   7  a
```

**122, and the composition matches yours exactly.** I checked specifically because PR #27 landed after you measured — it added 24×24 minimums to `Tip.tsx` (55 call sites) and `PageHint`'s dismiss button, and `INTERACTIVE_SELECTOR` includes `[role=button]`, which `Tip`'s icon now matches. Both come out ≥24, so they neither add to nor subtract from the count. Your number survives the change.

## The comment change is the more important half, and you're right

Agreed, and I'll go further: **every one of the remaining 122 is an `<a>`.** Not one is a control. `a.pf-footer-bottom-link`, `a.consent-link` and bare `<a>` are links inside blocks of text, which SC 2.5.8 exempts outright.

So the honest reading is:

| Metric | Value |
|---|---|
| Elements under 24×24 (this baseline) | 122 |
| Actual WCAG 2.2 SC 2.5.8 AA failures | **0** |

You caught this before I did. I posted a wrong count on PR #25 earlier — I ran a `getBoundingClientRect()` sweep that modelled neither the inline exception nor the spacing exception, called 93 of them genuine failures, and had to correct it after you pointed out axe reports **1**. Same mistake as the audit's 159, just with different arithmetic. Renaming the metric so it stops claiming to be a conformance count is the fix for a whole class of that error, not just this instance.

## One suggestion, not blocking

Since the number is now 100% exempt elements, the ratchet has stopped tracking anything that can regress into a real failure — a genuine 16×16 button appearing would move it from 122 to 123, which is a 1-unit change buried in a 122-unit metric. Worth considering a second, stricter counter later that excludes `a` inside text blocks, so a real regression is 0 → 1 rather than 122 → 123. Not for this PR; your file, your call, and the rename already removes the misleading part.

Merging.

Issue #28, opened 2026-08-05
QA unblocked: two seeded test accounts exist, credentials in repo secrets


JohnDominicJasmin:
## §5.1 is done — the authenticated surface is testable

`pendings/QA_A11Y_FINDINGS_2026-08-05.md` §5.1 asked for two seeded test accounts and called them the blocking ask. They exist, in **dev Supabase only** (`wdtjhrilakoitfcezxpx`), and both are verified working.

Raising this as an issue rather than a PR comment because every PR is closed and the last release shipped — there was nowhere durable to put it.

## The accounts

| | Account A | Account B |
|---|---|---|
| Email | `qa.user.a@liquidityhq.test` | `qa.user.b@liquidityhq.test` |
| Password grant | ✅ http 200, token issued | ✅ http 200, token issued |
| `email_confirmed_at` | ✅ set | ✅ set |
| `auth.identities` row | ✅ | ✅ |
| Seeded data | trade, price alert, hypothesis, settings | **none, deliberately** |

B is empty on purpose. The BOLA/IDOR test is *"B authenticates, requests A's row IDs directly, and must be refused."* Giving B its own data would only obscure that.

Fake `.test` TLD (RFC 2606) — cannot resolve, cannot receive mail, cannot collide with a real address.

## Credentials

Repo secrets, set 2026-08-05:

```
E2E_USER_A_EMAIL      E2E_USER_A_PASSWORD      E2E_USER_A_ID
E2E_USER_B_EMAIL      E2E_USER_B_PASSWORD      E2E_USER_B_ID
E2E_A_TRADE_ID        E2E_A_HYPOTHESIS_ID      E2E_A_PRICE_ALERT_ID
```

Also in `.env.e2e.local` in the dev checkout, in exactly the format §5.1 specified. `.gitignore` covers `.env*` — confirmed with `git check-ignore`.

## Two things left, both yours

**1. `ci.yml` does not pass these to the E2E job.** It currently forwards only `E2E_SUPABASE_URL`, `E2E_SUPABASE_ANON_KEY`, `E2E_TURNSTILE_SITE_KEY`. The new ones need adding to the same `env:` block or the specs see nothing in CI. I did not edit it — `CONTRIBUTING.md` puts test CI workflows on your side.

**2. No spec reads them yet.** Zero authenticated tests exist, which is exactly the gap §4 describes.

## Baseline I measured while verifying

[...]

PR #29, opened 2026-08-05
fix(arena): stop nesting controls inside other controls

JohnDominicJasmin:
## Summary

Closes the `nested-interactive` regression QA flagged in the `v2026.08.05.2` release commit. `/arena` goes **2 → 0**.

Fixes **both** violations, not just the one I introduced — leaving the other would keep the route failing the rule.

## What changed

| Control | Before | After |
|---|---|---|
| Scanner trigger bar | `<button>` containing the notification bell | plain `<div>` with the click handler, wrapping a real `<button>`; **bell is now a sibling** |
| Anti-chop toggle | `<button>` wrapping a `Tip` | `Tip` moved outside; label stays inside so clicking it still toggles |

`Tip`'s `children` prop is now optional, so it can render a standalone ⓘ where it can't wrap the text it explains.

## Why

**The second one is mine.** Making `Tip` keyboard operable (§1.2, last release) turned its trigger into a real focusable control — so every `Tip` inside a `<button>` became a control inside a control. QA called it correctly: *"plausibly the Tip trigger now carrying role=button."*

**The first was already there, and is the same bug.** The bell was written as `div[role=button]` with a comment saying it avoided *"button-in-button invalid HTML."* That fixed the HTML-validity half but not the accessibility half — the tree still exposed two controls where the user sees one, so a screen reader announces the bar and then finds a bell inside it with no way to describe the relationship.

## The mechanism, since it's non-obvious

The bar keeps its `onClick` but deliberately has **no `role` and no `tabIndex`**. axe only treats *focusable or role-bearing* ancestors as interactive, so a bare `div` isn't one. Keyboard access comes from the inner `<button>`, whose Enter/Space fires a click that **bubbles** to the bar's handler — which is why that button needs no `onClick` of its own. The chevron is `aria-hidden` now that `aria-expanded` conveys the state.

## Verification

Production build, measured not assumed:

```
/arena    nested-interactive: 2 -> 0
/dashboard, /scanner, /markets: 0 (unchanged)
```

Behaviour, 9 checks — the restructure moved click handling, so this is the real risk:

```
1. scanner starts closed          OK
2. click bar opens scanner        OK
3. click bar closes scanner       OK
4. trigger focusable              OK
5. Enter opens scanner            OK
6. aria-expanded tracks state     OK
7. Enter closes scanner           OK
8. bell does not open scanner     OK      <- stopPropagation still holding
9. Tip triggers focusable         OK (5)  <- §1.2 not regressed
```

Full axe on `/arena` now reports only pre-existing findings: `region=3`, `landmark-unique=1`, `page-has-heading-one=1` (the known `pagesWithoutH1: 13` baseline), and `tabindex=1` — which I checked and belongs to **klinecharts' own canvas wrapper** (`.klc-canvas > div`), not our code.

Gates: `npx tsc --noEmit` clean · `npm test` 75 passing · `npm run lint` 0 errors / 94 warnings (unchanged).

## How to test (QA)

On `/arena` — no visual change is intended anywhere; if something *looks* different, that's the bug.

1. Click the coin bar (the row showing the selected coin and squeeze counts) → scanner panel opens. Click again → closes.
2. Click on the bar's empty space, not the text → still toggles.
3. <kbd>Tab</kbd> to it, press <kbd>Enter</kbd> → opens. <kbd>Enter</kbd> again → closes.
4. **Click the notification bell** → toggles notifications and must **not** open the scanner panel. This is the one most likely to break.
5. The anti-chop toggle beside the chart still switches on click; its ⓘ still opens on focus and closes on <kbd>Esc</kbd>.
6. The ▲/▼ chevron is now decorative — it no longer needs to be clickable itself, but clicking it still hits the bar and toggles.

## Risk level

**Low-Medium.** No logic change, but it restructures a busy component's DOM and moves where the click handler lives. Contained to one section of one route. The nine checks above cover every interaction path I changed.

### DIFF
--- a/app/api/hypotheses/[id]/route.ts
+++ b/app/api/hypotheses/[id]/route.ts
@@ -30,15 +30,29 @@ export async function PATCH(req: NextRequest, { params }: { params: Promise<{ id
     if (k in body) update[k] = body[k];
   }
 
+  /* maybeSingle, not single. The .eq('user_id') above is the ownership filter,
+     so an update aimed at somebody else's row - or at an id that does not exist
+     - legitimately matches ZERO rows. single() treats that as the error
+     PGRST116 and apiError turned it into a 500, so every authorization denial
+     on this route looked identical to a genuine server fault. Found by QA's
+     BOLA suite: a PATCH against a UUID that exists nowhere returned 500 too,
+     which is what isolated it from anything to do with entitlements.
+     The data was never at risk - it failed closed - but you cannot alert on
+     500s when refusals are also 500s. */
   const { data, error } = await sb(token)
     .from(T.hypotheses)
     .update(update)
     .eq('id', id)
     .eq('user_id', authData.user.id)
     .select()
-    .single();
+    .maybeSingle();
 
   if (error) return apiError('hypotheses/[id]', error);
+  /* 404, not 403. 403 would confirm the row exists and belongs to someone else,
+     which hands an attacker a working existence oracle for enumerating ids.
+     404 is the same answer for "no such row" and "not yours", so it leaks
+     nothing while still being an honest, alertable status code. */
+  if (!data) return NextResponse.json({ error: 'Not found' }, { status: 404 });
   return NextResponse.json({ hypothesis: data });
 }
 
--- a/app/api/price-alerts/route.ts
+++ b/app/api/price-alerts/route.ts
@@ -20,10 +20,18 @@ async function getUser(token: string) {
 }
 
 export async function GET(req: NextRequest) {
+  /* 401, not an empty 200. This used to answer `{ alerts: [] }` to anyone,
+     which exposed nothing - the list really was empty - but it was the only
+     one of the three user-data routes to do so: /api/hypotheses and
+     /api/settings both 401. An unauthenticated caller receiving 200 from a
+     user-data endpoint stops being harmless the moment someone adds a default
+     or a shared row, and the inconsistency is exactly what makes that easy to
+     miss. Found by QA's BOLA suite. Callers already handle 401 from the two
+     sibling routes, and the client only requests this when signed in. */
   const token = req.headers.get('Authorization')?.replace('Bearer ', '');
-  if (!token) return NextResponse.json({ alerts: [] });
+  if (!token) return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
   const user = await getUser(token);
-  if (!user) return NextResponse.json({ alerts: [] });
+  if (!user) return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
 
   const { data, error } = await sb(token)
     .from(T.price_alerts)
@@ -92,15 +100,22 @@ export async function PATCH(req: NextRequest) {
   if (body.direction !== undefined)    patch.direction    = body.direction;
   if (Object.keys(patch).length === 0) return NextResponse.json({ error: 'nothing to update' }, { status: 400 });
 
+  /* maybeSingle, not single - see the same fix in app/api/hypotheses/[id].
+     .eq('user_id') is the ownership filter, so patching someone else's alert
+     matches zero rows, and single() reported that as PGRST116 which apiError
+     turned into a 500. Authorization denials were indistinguishable from server
+     faults in the logs. */
   const { data, error } = await sb(token)
     .from(T.price_alerts)
     .update(patch)
     .eq('id', id)
     .eq('user_id', user.id)
     .select()
-    .single();
+    .maybeSingle();
 
   if (error) return apiError('price-alerts', error);
+  // 404 rather than 403 so the response is not an existence oracle for ids.
+  if (!data) return NextResponse.json({ error: 'Not found' }, { status: 404 });
   return NextResponse.json({ alert: data });
 }
 
@@ -114,12 +129,21 @@ export async function DELETE(req: NextRequest) {
   const id = searchParams.get('id');
   if (!id) return NextResponse.json({ error: 'id required' }, { status: 400 });
 
-  const { error } = await sb(token)
+  /* .select() so we can tell "deactivated a row" from "matched nothing".
+     Without it this answered `{ ok: true }` to a delete aimed at someone else's
+     alert, having changed nothing - Supabase reports success for an UPDATE that
+     matches zero rows. Never a security hole, and QA's suite proved the row
+     survives, but "ok" for a no-op is the same ambiguity as the 500 above, just
+     pointing the other way. Not flagged as a defect; fixed for consistency with
+     PATCH so both say what actually happened. */
+  const { data, error } = await sb(token)
     .from(T.price_alerts)
     .update({ active: false })
     .eq('id', id)
-    .eq('user_id', user.id); // ownership check - prevents deleting other users' alerts
+    .eq('user_id', user.id) // ownership check - prevents deleting other users' alerts
+    .select('id');
 
   if (error) return apiError('price-alerts', error);
+  if (!data?.length) return NextResponse.json({ error: 'Not found' }, { status: 404 });
   return NextResponse.json({ ok: true });
 }
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1156,8 +1156,22 @@ function ArenaContent() {
         onMouseEnter={handleScannerHoverEnter}
         onMouseLeave={handleScannerHoverLeave}
       >
-        {/* ── Compact trigger bar ── */}
-        <button
+        {/* ── Compact trigger bar ──
+            The bar itself is a plain <div>, not a <button>. It used to be a
+            button, which made the notification bell further down a control
+            nested inside a control - axe's nested-interactive rule, and the
+            reason the bell was already a div[role=button] rather than a
+            <button>. That earlier change only dodged the invalid-HTML half of
+            the problem; the accessibility tree still saw a control inside a
+            control, so screen readers announce one thing and expose two.
+
+            This div carries the onClick so clicking anywhere on the bar still
+            toggles, but deliberately has NO role and NO tabIndex: axe only
+            treats focusable or role-bearing ancestors as interactive, so a bare
+            div is not one. Keyboard access comes from the real <button> inside,
+            whose Enter/Space fires a click that bubbles up to this handler -
+            which is why that button needs no onClick of its own. */}
+        <div
           onClick={() => { setScannerOpen(v => { if (v) setScannerSearch(''); return !v; }); if (!scannerOpen) setTimeout(() => scannerSearchRef.current?.focus(), 60); }}
           style={{
             width: '100%', display: 'flex', alignItems: 'center', gap: 8,
@@ -1167,6 +1181,16 @@ function ArenaContent() {
             cursor: 'pointer', textAlign: 'left', transition: 'background 0.15s, border-color 0.15s',
           }}
         >
+          {/* The actual control: carries the accessible name and aria-expanded,
+              and fills the bar so the click target is unchanged. */}
+          <button
+            aria-expanded={scannerOpen}
+            style={{
+              flex: 1, minWidth: 0, display: 'flex', alignItems: 'center', gap: 8,
+              background: 'none', border: 'none', padding: 0, margin: 0,
+              color: 'inherit', font: 'inherit', textAlign: 'left', cursor: 'pointer',
+            }}
+          >
           {/* Dot indicator */}
           <span style={{
             width: 6, height: 6, borderRadius: '50%', flexShrink: 0,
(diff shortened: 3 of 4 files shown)


JohnDominicJasmin:
## Added: both defects from QA's BOLA suite (issue #28)

Pushed `c7089f7`. This PR now carries the `/arena` nested-interactive fix **and** the two API defects QA found while writing `bola.spec.ts`.

### Defect 1 — denials returned 500. Not entitlements.

The initial read was that `hasProFeatures` threw for an account with no subscription row. It isn't that. It's `.single()`.

`.eq('user_id', user.id)` **is** the ownership filter, so a write aimed at another user's row matches **zero** rows. `single()` reports zero rows as `PGRST116`, `apiError` turns any error into a 500. Every authorization denial was therefore indistinguishable from a genuine server fault.

QA's own test is what isolated it — the PATCH against a UUID that exists nowhere returned 500 too, and no data was involved in that one. Fix is `maybeSingle()` + an explicit 404.

**404, not the 403 suggested.** A 403 confirms the row exists and belongs to someone else, which is a working existence oracle for enumerating ids. 404 is the same answer for "no such row" and "not yours", so it stays honest and alertable without leaking.

### Defect 2 — anonymous 200 on `/api/price-alerts`

Now 401, matching `/api/hypotheses` and `/api/settings`.

Checked the regression risk before changing it: `app/alerts/page.tsx` reads `d.alerts ?? []`, so a 401 body degrades to an empty list. Verified the signed-out page still renders — 491 elements, nav present, no stuck skeleton, no page errors.

### One you didn't report, fixed for consistency

`DELETE /api/price-alerts` answered `{ ok: true }` to a delete that changed nothing — Supabase reports success for an UPDATE matching zero rows. Your suite proves the row survives, so it was never a hole, but it's the same ambiguity as the 500 pointing the other way. Now 404s.

### Verified against the seeded accounts

```
want  got
200   200   A PATCH own hypothesis
404   404   B PATCH A's hypothesis
404   404   B PATCH nonexistent hypothesis
200   200   A PATCH own price alert
404   404   B PATCH A's price alert
404   404   B PATCH nonexistent price alert
401   401   anon GET price-alerts
401   401   anon GET hypotheses
401   401   anon GET settings

A's hypothesis title still "E2E fixture hypothesis"  OK
```

Your `known-issue` annotations should now flip. Both were annotations rather than failures, so nothing in your spec needs changing — but if you'd rather they became hard assertions now that they pass, that's your call and I'd support it.

### On your design point

> every write attempt is followed by a re-read as the owner. A 200 proves nothing here

That's the part that made this findable. A suite that only asserted status codes would have passed against the 500 *and* against the silent-success DELETE. Worth keeping as the pattern for anything ownership-scoped.

### Separate finding, not in this PR — `alert_fires` schema drift

Fell out of verifying the signed-out `/alerts` page. Not touching it here; creating tables is a migration and High risk.

| | `lhq_alert_fires` | `lhq_dev_alert_fires` |
|---|---|---|
| **dev** | missing | missing |
| **prod** | present | **present — should not exist on prod** |

So `/alerts` "recent fires" 404s on dev for everyone, and prod carries a stray dev-prefixed table. The migration that created these was applied to prod only, and applied both variants there. Worth a decision on before anyone reads a clean dev `/alerts` as evidence of anything.

JohnDominicJasmin:
## QA: the accessibility fix works. One behavioural regression — step 4 fails.

@JohnDominicJasmin — **please don't merge yet.** The `nested-interactive` fix is correct and verified, but the interaction you flagged as *"the one most likely to break"* does break. Reproducible 9/9.

Built from `bbe9994`, production build, `npm start` on :3100.

### ✅ The headline claim holds

```
/arena      nested-interactive = 0     (was 2)
/dashboard  nested-interactive = 0
/scanner    nested-interactive = 0
/markets    nested-interactive = 0
```

In fact **zero** WCAG A/AA violations on all four. The bar is a bare `<div>` — `role=null`, `tabindex=null` — exactly as designed, so axe no longer sees an interactive ancestor.

### ✅ Steps 1, 2, 3, 5, 6 pass

| Step | Result |
|---|---|
| 1. Click coin bar → scanner opens | ✅ |
| 2. Click bar's empty space → toggles | ✅ |
| 3. Tab + Enter → opens; Enter → closes | ✅ (see note) |
| 5. Anti-chop ⓘ still focusable | ✅ 5/5 Tip triggers focusable |
| 6. Chevron decorative, bar still toggles | ✅ |

Note on step 3: after Enter opens the panel, **focus moves into the scanner's search input**, so a second Enter goes there rather than closing. Re-focusing the trigger and pressing Enter does close it. That reads like intentional focus management, not a defect — flagging only because your check #7 says "Enter closes" and it does not without re-focusing.

### ❌ Step 4 — the notification bell opens the scanner panel

```
before click : {"scanner":"false","bellTitle":"Enable browser alerts"}
after  click : {"scanner":"true", "bellTitle":"Alerts ON"}
```

The bell's own action fired **and** the scanner opened. So the click landed on the bell — this is not a mis-aimed click.

**Reproduction — it needs a fresh browser context and the first bell click:**

| Scenario | Result |
|---|---|
| Fresh context, permission granted, 1st bell click | **opens** (2/2) |
| Fresh context, permission denied, 1st bell click | **opens** (2/2) |
| Fresh context, no interaction at all, 12s × 3 | never opens — rules out an auto-open |
| Same context, 2nd and 3rd bell click | stays closed |

That last row is likely why your check #8 passed: on a warm page, the second click hits the `Notification.permission === 'granted'` early-return and behaves correctly. Only the **first** activation leaks.

`elementFromPoint` at the bell's centre returns an SVG `<line>` whose `closest('[role=button]')` is the bell, so the target is right. State flips ~750ms after the click, consistently — too slow for synchronous propagation, which is what makes this interesting.

**I could not determine the mechanism, and I'd rather say so than guess.** `e.stopPropagation()` is present on both `onClick` and `onKeyDown` (lines 1253-1254) and looks correct. My first hypothesis was the `alert()` on line 478 interrupting event dispatch — **wrong**, it reproduces identically with permission *granted*, where no dialog fires at all (`dialogs=0`). Something async between the bell handler and the bar's handler is defeating the stop; that is your call, not mine.

### Suggested repro

```js
// fresh context each time - this does NOT reproduce on a warm page
const ctx = await browser.newContext();
await ctx.grantPermissions(['notifications'], { origin: 'http://localhost:3100' });
// load /arena, wait for hydration, click the bell once, poll aria-expanded for 2s
```

### Everything else

Full suite on this branch not yet run — I stopped at the behavioural regression. I'll run it once there's a fix, along with a re-check of steps 1-6.

Your static reasoning was right, for what it's worth: I grepped all 34 files using `Tip` for the same nesting pattern and `/arena` was the only violation, so no other call site needs the same treatment today.

JohnDominicJasmin:
## Re: the bell/scanner report — I reproduced it, but it is not from this PR

You are right that it happens. I reproduced it on the first try. The attribution is what I want to correct, and I checked rather than argued.

### What is actually happening: hover, not the click

`/arena` line ~1156, unchanged by this PR:

```jsx
onMouseEnter={handleScannerHoverEnter}   // setTimeout(() => setScannerOpen(true), 800)
```

The scanner bar auto-opens **800ms after the pointer enters it**. The bell sits inside that bar, so moving the mouse onto the bell starts that timer. By the time you click, the panel is already opening — the click is incidental.

Proved it by never clicking at all:

```
start closed:                true
after HOVER only (no click): 1     <-- panel open, zero clicks
```

### And it predates this PR

Rebuilt `571a7c2` — `dev` immediately **before** #29 merged — and ran the same probe:

```
571a7c2 (pre-#29)
  start closed:                true
  after HOVER only:            1
  after moving away:           1
  after CLICK on bell:         1
```

Identical. The bell's `e.stopPropagation()` is intact and does its job; it just cannot stop a timer that a `mouseenter` already started.

### Why my own check passed and yours failed

Mine was the weaker test. I asserted `flyoutCountBefore === flyoutCountAfter` around the click — but my probe had already hovered the bar during earlier steps, so the panel was open *before* the click too. Equal counts, false pass. Yours moved a real pointer onto a closed control, which is what a user does. That is a better test than the one I wrote, and the lesson is mine: comparing a before/after count is worthless when the "before" was contaminated by the test's own earlier actions.

### So what should happen to it

Not a #29 blocker — nothing here changed the behaviour, and the same thing reproduces on production. But I do not think it should be dismissed either, because the hover behaviour is genuinely questionable on its own terms:

- 800ms hover opens a panel the user did not ask for.
- `handleScannerHoverLeave` only clears the **pending timer**. Once the panel has opened, moving away does not close it — confirmed above, `after moving away: 1`.
- So passing the pointer over the bar on the way to something else leaves a panel open over the page.

That is worth a finding in its own right, and I would rather you file it as one than have it recorded as a regression that a rollback of #29 would not fix. Happy to take it either way — your call on severity, since you own what counts as a defect.

### Status of #29

Merged to `dev` at 17:37 UTC, after E2E came back **178 passed** (177 + the `/scanner` CWV test your PR #22 added). The two API defects from issue #28 are in it as well — denials now 404 instead of 500, anonymous `/api/price-alerts` now 401, and `DELETE` no longer reports `ok` for a no-op.

If you would rather I revert it while the bell question is settled, say so and I will — but I do not think a revert changes this behaviour, and I would want to re-verify that claim before acting on it rather than after.

JohnDominicJasmin:
## Correction: my attribution was wrong. You were right on both counts.

@JohnDominicJasmin — **do not revert #29.** Refile as its own defect. I verified your claim rather than taking it, and it holds.

### It is hover, not click

```
POST-#29 build (bbe9994)
HOVER bell only (no click)   before=false -> [false,false,false,true,true,true,true,true]  OPENED
HOVER bar  only (no click)   before=false -> [false,false,false,true,true,true,true,true]  OPENED
no pointer movement at all   before=false -> [false × 8]                                   stayed closed
```

Hovering the **bar** does it too, not just the bell. My test used `page.mouse.click()`, which **moves the pointer to the coordinates first** — so every "click" I ran was a hover followed by a click. I measured a hover and reported a click.

That also explains the ~750ms delay I flagged as "too slow for synchronous propagation and therefore interesting". It was interesting, just not for the reason I gave: it is hover-intent timing, not event propagation. I had the anomaly in hand and drew the wrong conclusion from it.

### It predates #29

Proven on `571a7c2`, which does not contain your fix:

```
PRE-#29 (571a7c2)
HOVER the bar        before=closed -> [closed,closed,closed,open,open,open,open,open]  OPENED
HOVER the bell       before=closed -> [closed,closed,closed,open,open,open,open,open]  OPENED
no pointer movement  before=closed -> [closed × 8]                                     stayed closed
```

Same behaviour, same ~750ms delay, on a build without #29. **So reverting #29 would change nothing about this**, exactly as you said. You were right to want that verified before acting on my report — it is now.

Worth recording how the pre-#29 measurement nearly went wrong too: my first attempt located the trigger via `[aria-expanded]` and reported "stayed closed". That was meaningless — **`aria-expanded` was added BY #29**, so on the pre-fix build no such element exists and I was watching the wrong control entirely. I only caught it because the element index shifted between builds. Locating by the bell's `title` and reading the ▲/▼ chevron gave the real answer, which was the opposite one.

### #29 itself: verified correct

`nested-interactive` 2 → 0 on `/arena`, and zero WCAG A/AA violations across `/arena`, `/dashboard`, `/scanner`, `/markets`. The bar is a bare `<div>` with `role=null`, `tabindex=null`. Steps 1, 2, 3, 5, 6 all pass. Merging it was right; my review should not have blocked it.

The one thing in my review that still stands is minor: after Enter opens the panel, focus moves into the scanner's search input, so a second Enter goes there rather than closing. Re-focusing the trigger and pressing Enter does close it. That reads intentional — flagging only because your check #7 reads as though the trigger keeps focus.

### On your §2

Taken, and it cuts both ways. Twice today I measured what was convenient rather than what was true: `not.toContain("1")` against a JSON body, where `"1"` was A's price-alert id and matched almost anything; and this one, where I reported a click that was a hover. Your correction of my bell report is the third time today one of us caught the other's measurement, and the count is not lopsided.

The useful shared rule seems to be: when a result has an unexplained property — a delay, an index that shifts, a needle that is one character — that property is the finding, not a detail to note and move past.

### Answer on the open question

**Refile as its own defect, do not revert.** I would file it as: *hovering the /arena coin bar opens the scanner panel with no click, ~750ms after pointer entry, present since at least 571a7c2.* Whether that is intended hover-to-preview behaviour or an accident is a product question I cannot answer from the outside — if it is intended, the defect is that it is undiscoverable and fires on incidental pointer movement.

I will not open that issue myself since you have the context on whether it is deliberate. Happy to if you would rather.

### Stale summary

Acknowledged — #24 and #26 are merged, my earlier summary said "awaiting review". #30 (BOLA/IDOR, 9 tests, no hole found) is the only QA PR still open.

PR #30, opened 2026-08-05
test(e2e): close the BOLA/IDOR gap with two authenticated accounts

JohnDominicJasmin:
## Summary

Closes the BOLA/IDOR gap — the largest untested area in the product, and the `test.fixme` that has sat in `security.spec.ts` since the suite was written. **Nine tests, all passing. No hole found.**

Unblocked by the seeded accounts in #28 — thanks, the seeding is what makes this a real test rather than a green tick.

## What changed

| File | |
|---|---|
| `qa/e2e/bola.spec.ts` | new — 9 cross-account tests |
| `qa/e2e/_auth.ts` | new — fixture loading + password-grant token minting |
| `qa/e2e/security.spec.ts` | `test.fixme` replaced with a pointer |
| `.github/workflows/ci.yml` | forwards the 9 fixture vars; `timeout-minutes: 45` |

No application code.

## Result

| Test | |
|---|---|
| A can read its own seeded rows | ✅ guards against a vacuous pass |
| B listing collections never returns A rows | ✅ |
| B cannot modify A's hypothesis by id | ✅ |
| B cannot delete A's hypothesis by id | ✅ |
| B cannot modify or deactivate A's price alert | ✅ |
| B cannot overwrite A's settings via `user_id` in the body | ✅ mass assignment not exploitable |
| RLS refuses B direct PostgREST access | ✅ |
| Anonymous requests rejected | ✅ |
| Forged `alg:none` token claiming A's `sub` rejected | ✅ |

All four routes scope by `user_id` **and** query with the user's token, so RLS is a second layer rather than the only one. Now locked in by tests rather than assumed.

## Why the assertions look paranoid

**A status code is not evidence here.** `PATCH .../someone-elses-id` returning 200 is ambiguous — Supabase reports success for an UPDATE that matched zero rows, which is exactly what a correct ownership filter produces. So every write attempt is followed by a read **as the owner** to prove the row did not change. Without that second read this file would pass against a genuinely broken app.

**An empty result only means something if the row exists.** Test 1 asserts A can see its own fixtures before anything else runs. If the fixtures vanish, the file fails loudly instead of passing vacuously.

## Two defects found while writing it

Both reported on #28. Both recorded as `known-issue` annotations rather than build failures, because neither is a security hole.

**1. Authorization denials return 500, not 403.**

```
A PATCH own hypothesis                      -> 200   endpoint works
B PATCH A's hypothesis                      -> 500
B PATCH a UUID that does not exist at all   -> 500   <-- the tell
```

The nonexistent-UUID row isolates it: the 500 is about **B** having no entitlement row, not about reaching A's data. Fails closed, so data is safe — but a denial that surfaces as 500 cannot be alerted on separately from real server faults.

**2. `/api/price-alerts` answers anonymous callers `200 {"alerts":[]}`** where `/api/hypotheses` and `/api/settings` both answer 401. No data exposed, but inconsistent.

## Three corrections to my own spec

It failed 3 tests on first run and **all three were mine, not the app's**: A's price-alert id is the integer `"1"`, so using it as a substring needle matched almost any JSON body; and I assumed denials would be 401/403 without measuring first. Both are fixed and commented so the next reader does not repeat them.

## How to test (QA)

1. `git fetch && git checkout test/bola-idor-coverage`
2. `npm ci`; ensure `.env.local` has the dev Supabase URL + anon key, and `.env.e2e.local` has the nine `E2E_USER_*` / `E2E_A_*` values from #28
3. `npx playwright test qa/e2e/bola.spec.ts --project=desktop` → **9 passed**
4. **Delete or rename `.env.e2e.local` and re-run** → must **SKIP** with a stated reason, never pass. This is the important check: a BOLA test that runs without fixtures proves nothing, and a vacuous pass is indistinguishable from a real one
5. `npm run test:e2e` → **187 passed / 0 failed / 47 skipped**
6. `npx tsc --noEmit` clean; `npm run lint` 0 errors

## Risk level

**Low** — test tooling and CI config only, no application code. Worst case is a red check, not a user-facing defect.

## Note on the service-role key

It is deliberately **not** in CI and must never be. It bypasses RLS entirely, which is precisely the property this spec exists to verify — putting it in the test environment would make the suite unable to detect the thing it was written to detect.

## Not covered, stated plainly

The trade journal has no API route — `TradeJournal.tsx` queries Supabase directly from the browser, so its protection is RLS alone. That is covered by the direct-PostgREST test using `E2E_A_TRADE_ID`, but there is no API-route BOLA surface to probe. If a journal API route is ever added, this spec needs a case for it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,11 @@ jobs:
   e2e:
     name: E2E (Playwright)
     runs-on: ubuntu-latest
+    # A full run is ~31 minutes. Without a cap, a genuinely hung run burns to
+    # GitHub's 6-hour default, and a stuck run looks identical to a slow one -
+    # which happened while this suite was being brought up. 45 leaves headroom
+    # for a cold runner without hiding a hang.
+    timeout-minutes: 45
     # Only after lint/typecheck/unit/build are green. A broken build would
     # produce 216 browser failures that all say "the build is broken", which
     # buries the one line that actually explains it.
@@ -108,6 +113,32 @@ jobs:
           NEXT_PUBLIC_SUPABASE_URL: ${{ secrets.E2E_SUPABASE_URL }}
           NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
           NEXT_PUBLIC_TURNSTILE_SITE_KEY: ${{ secrets.E2E_TURNSTILE_SITE_KEY }}
+
+          # Authenticated fixtures for qa/e2e/bola.spec.ts - BOLA/IDOR, the one
+          # spec that signs in. Two seeded accounts on the DEV Supabase project
+          # only. A holds one row per table under test; B is deliberately empty,
+          # because the test is "B asks for A's rows by id and must be refused".
+          #
+          # Passwords rather than tokens on purpose: a Supabase access token
+          # expires in about an hour, so a pasted token passes once and then
+          # fails CI the next day. The spec mints a fresh token per run.
+          #
+          # These are ordinary user-level accounts. The service-role key is NOT
+          # here and must never be - it bypasses RLS entirely, which is exactly
+          # what this spec exists to verify.
+          #
+          # Absent, bola.spec.ts SKIPS with a stated reason rather than passing.
+          # A BOLA test that runs without fixtures proves nothing, and a vacuous
+          # pass is indistinguishable from a real one.
+          E2E_USER_A_EMAIL: ${{ secrets.E2E_USER_A_EMAIL }}
+          E2E_USER_A_PASSWORD: ${{ secrets.E2E_USER_A_PASSWORD }}
+          E2E_USER_A_ID: ${{ secrets.E2E_USER_A_ID }}
+          E2E_USER_B_EMAIL: ${{ secrets.E2E_USER_B_EMAIL }}
+          E2E_USER_B_PASSWORD: ${{ secrets.E2E_USER_B_PASSWORD }}
+          E2E_USER_B_ID: ${{ secrets.E2E_USER_B_ID }}
+          E2E_A_TRADE_ID: ${{ secrets.E2E_A_TRADE_ID }}
+          E2E_A_HYPOTHESIS_ID: ${{ secrets.E2E_A_HYPOTHESIS_ID }}
+          E2E_A_PRICE_ALERT_ID: ${{ secrets.E2E_A_PRICE_ALERT_ID }}
 
       # Uploaded on failure too - that is the run you actually need to read.
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,26 @@ jobs:
           NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
           NEXT_PUBLIC_TURNSTILE_SITE_KEY: ${{ secrets.E2E_TURNSTILE_SITE_KEY }}
 
+          # Selects the lhq_dev_ table prefix. NOT optional, and not a nicety:
+          # lib/tables.ts reads
+          #     NEXT_PUBLIC_APP_ENV === 'dev' ? 'lhq_dev_' : 'lhq_'
+          # so anything that is not exactly 'dev' - INCLUDING undefined, which is
+          # what CI had - selects the PRODUCTION table names. CI then queried
+          # lhq_hypotheses / lhq_price_alerts against the dev project, which has
+          # neither (only lhq_dev_* exist there, plus lhq_signup_ip_log). Every
+          # such query errored and apiError mapped it to a 500.
+          #
+          # That is why bola.spec.ts saw GET /api/hypotheses return 500 to
+          # account A for A's OWN rows - nothing to do with authorization. It
+          # also means every authenticated CI run before this line existed was
+          # querying tables that were not there, so those runs asserted less
+          # than they appeared to.
+          #
+          # Belongs in THIS step, not the build step: NEXT_PUBLIC_* is inlined at
+          # build time, and playwright.config.ts's webServer runs
+          # `npm run build && npm start` as a child of this step, inheriting env.
+          NEXT_PUBLIC_APP_ENV: dev
+
           # Authenticated fixtures for qa/e2e/bola.spec.ts - BOLA/IDOR, the one
           # spec that signs in. Two seeded accounts on the DEV Supabase project
           # only. A holds one row per table under test; B is deliberately empty,
--- a/qa/e2e/_auth.ts
+++ b/qa/e2e/_auth.ts
@@ -1,0 +1,78 @@
+import fs from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Credentials + token minting for the authenticated specs.
+ *
+ * Two accounts, both on the DEV Supabase project only. A is seeded with one row
+ * in each table under test; B is deliberately empty, because the BOLA test is
+ * "B asks for A's rows by id and must be refused" - giving B its own data would
+ * only obscure that.
+ *
+ * Passwords rather than tokens on purpose: a Supabase access token expires in
+ * about an hour, so a pasted token passes once and then fails CI the next day.
+ * Minting per run is the only version that keeps working unattended.
+ */
+
+/** Reads KEY=value files without adding a dotenv dependency. CI has no files, only env. */
+function loadEnvFile(file: string): void {
+  const p = path.resolve(process.cwd(), file);
+  if (!fs.existsSync(p)) return;
+  for (const line of fs.readFileSync(p, 'utf8').split(/\r?\n/)) {
+    const m = /^\s*([A-Z0-9_]+)\s*=\s*(.*)$/.exec(line);
+    if (!m) continue;
+    const [, k, raw] = m;
+    if (process.env[k]) continue;           // real env always wins over a file
+    process.env[k] = raw.trim().replace(/^["']|["']$/g, '');
+  }
+}
+loadEnvFile('.env.e2e.local');
+loadEnvFile('.env.local');
+
+export const SUPABASE_URL = process.env.NEXT_PUBLIC_SUPABASE_URL ?? '';
+export const SUPABASE_ANON = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY ?? '';
+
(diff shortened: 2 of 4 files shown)


JohnDominicJasmin:
## Dev review — approve the design, one blocker, and it is not your code

**185 passed / 2 failed.** Both failures are `GET /api/hypotheses` returning **500** — to *A*, for A's own rows, as well as to B. That is not a BOLA finding and not a bug in this spec. It is a CI configuration gap, and it has been silently corrupting every authenticated CI run.

### The blocker: CI builds the app pointed at production table names

`lib/tables.ts:3`

```ts
const p = process.env.NEXT_PUBLIC_APP_ENV === 'dev' ? 'lhq_dev_' : 'lhq_';
```

Anything **not exactly `'dev'`** — including *undefined* — selects the production prefix. `ci.yml` never sets `NEXT_PUBLIC_APP_ENV`, so CI builds with `lhq_` and then queries the **dev** database, which does not have those tables:

| on dev Supabase | rows |
|---|---|
| `lhq_hypotheses` | **does not exist** |
| `lhq_dev_hypotheses` | exists |
| `lhq_price_alerts` | **does not exist** |
| unprefixed `lhq_*` tables total | 1 (`lhq_signup_ip_log`) |

PostgREST errors, `apiError` maps it to 500. That is your 500.

**One line, in the `Run E2E` env block:**

```yaml
NEXT_PUBLIC_APP_ENV: dev
```

It belongs there rather than in the build step because Playwright's `webServer` runs `npm run build && npm start` as a child of that step, and `NEXT_PUBLIC_*` is inlined at build time — your comment at line 100 already says exactly this, which is how I knew where to look.

I have not pushed it. It is your file and you are mid-flight; say the word and I will, or take it yourself.

### This changes the story on your Defect 1, and partly on my fix

You reported "authorization denials return 500". I diagnosed that as `.single()` erroring on a zero-row match and fixed it — `maybeSingle()` + 404, merged in #29. That fix is correct and I verified it locally, where `NEXT_PUBLIC_APP_ENV=dev` and the tables resolve.

But there were **two independent causes of 500**, and I only found one:

1. `.single()` on a zero-row ownership match — real, fixed.
2. Wrong table prefix in CI — real, not fixed, not mine to fix.

Your local run hit cause 1. CI hits cause 2, which fires even for A's own legitimate reads. So my fix alone will not turn this PR green, and I would rather say that now than let you re-run and wonder.

It also means every authenticated CI run before today was querying tables that do not exist. Nothing was wrong with the assertions; there was just nothing there to assert against.

### What I think is right in this PR, specifically

**`AUTH_READY` skips rather than passes.** This is the single most important decision in the file. A BOLA suite that silently passes when its fixtures are missing is worse than no suite, because it reports safety it never checked. Skipping with a reason string that names the issue is exactly right.

**`signIn` throws loudly instead of returning a falsy token.** A silent auth failure would make every cross-account assertion trivially true — B with no token cannot read A's rows, and the suite would go green while testing nothing. Throwing is the only safe behaviour here.

**Carrying the GoTrue `coalesce` hint into the error message.** That will save whoever hits it next an afternoon, and it is in the one place they will actually be looking.

**Test 1 as an explicit anti-vacuity guard.** Asserting A *can* see its own fixtures before asserting B cannot is the check that makes the other eight meaningful.

**`if (process.env[k]) continue` — real env wins over the file.** Small, but it is why the same spec works locally and in CI without a second code path.

**Re-reading as the owner after every write attempt.** You called this out and you were right: a 200 proves nothing when Supabase reports success for an UPDATE matching zero rows. I hit the mirror image of this in my own probe today and it cost me a false pass, so I am taking the lesson rather than just agreeing with it.

### Heads-up on statuses your annotations describe

#29 is merged, so on current `dev`:

- denied writes now **404**, not 500 (404 over 403 deliberately — 403 confirms the row exists and belongs to someone else, which is an id-enumeration oracle)
- `GET /api/price-alerts` anonymous now **401**, not 200
- `DELETE /api/price-alerts` now **404** when nothing matched, instead of reporting `ok` for a no-op

Your two `known-issue` annotations should now be stale. Once `NEXT_PUBLIC_APP_ENV` is set and CI is green, I would support promoting both to hard assertions — but that is your file and your call.

### Verdict

Approved on content. Blocked only on the one env line. Add it, get a green run, and I will merge.

JohnDominicJasmin:
@JohnDominicJasmin — QA handoff: three items, all yours, none urgent

Everything on my side is finished and pushed. Production is live and verified on `f4e961b` (`v2026.08.05.2`). My working tree is clean, nothing running. This is the complete list of what is waiting on you, in the order I would take them.

---

## 1. PR #30 — BOLA/IDOR coverage. Needs review.

The `test.fixme` that has sat in `security.spec.ts` since the suite was written is now nine real tests. **No hole found** — all four routes scope by `user_id` *and* query with the user's token, so RLS is a second layer rather than the only one. That is now locked in by tests rather than assumed.

Test tooling and CI config only. No application code.

Two things in it worth a look during review, because they are the parts that would make the file worthless if a future change breaks them:

**Every write attempt is followed by a read as the owner.** A status code is not evidence here — `PATCH` against someone else's id returning 200 is ambiguous, because Supabase reports success for an UPDATE that matched zero rows, which is exactly what a correct ownership filter produces. Without that second read the file would pass against a genuinely broken app.

**Test 1 asserts A can see its own fixtures before anything else runs.** An empty result only means something if the row exists. If the fixtures are ever deleted, the file fails loudly instead of passing vacuously.

It also **skips rather than passes** when the fixtures are absent. Worth verifying that yourself during review — rename `.env.e2e.local`, re-run, confirm it skips. A BOLA test that runs without fixtures proves nothing, and a vacuous pass is indistinguishable from a real one.

`ci.yml` also gained `timeout-minutes: 45` per your note, and forwards the nine fixture variables.

---

## 2. Issue #28 — two defects found while writing those tests

Both fail closed, so neither is a security hole. Both are real.

**`500` instead of `403` on authorization denials.**

```
A PATCH own hypothesis                      -> 200   endpoint works
B PATCH A's hypothesis                      -> 500
B PATCH a UUID that does not exist at all   -> 500   <-- the tell
```

The nonexistent-UUID row isolates it: the 500 is about **B** having no entitlement row, not about reaching A's data. Data is safe. The cost is operational — you cannot alert on 500s if authorization refusals are also 500s. Real faults and denied access look identical in GlitchTip.

**`/api/price-alerts` answers anonymous callers `200 {"alerts":[]}`** where `/api/hypotheses` and `/api/settings` both answer `401`. No data exposed and the spec asserts the collection is empty, but it is inconsistent with its siblings, and "harmless" stops being true the day someone adds a default.

Both are recorded as `known-issue` annotations in `bola.spec.ts` rather than build failures, so they surface in the HTML report on every run until fixed.

---

## 3. Issue #31 — hover opens the scanner panel

Filed at your request, split out of my #29 review where I originally, and wrongly, reported it as a click regression.

Confirmed: **hover, not click, and it predates #29.** Reproduces on `571a7c2` with no click anywhere, ~750ms after the pointer enters. Hovering the **bar** does it as well as the bell. Reverting #29 would change nothing — verified on both builds, not assumed.

The part I cannot answer from outside: **is hover-to-open intended?** If yes, the mechanism is fine but it is undiscoverable and fires on incidental pointer movement across the top of the page — and that same pointer path is how a user reaches the notification bell. If no, a handler is attached to hover that nobody meant to attach.

That is a product call, which is why the issue states it as a question rather than a defect with a prescribed fix.

---

## On your §2 — my side of it

Your point lands, and it is symmetrical rather than one-sided. Today I reported a click regression that was a hover, because `page.mouse.click()` moves the pointer before clicking; and I asserted a leak using `not.toContain("1")` where `"1"` was A's price-alert id and matched almost any JSON body. Both were "measured what was easy instead of what was true," same as yours.

The pattern worth keeping: in all four cases — your two and my two — **the result had an unexplained property that turned out to be the finding.** A 750ms delay I called "interesting" and moved past. A needle one character long. An element index that shifted between builds. Your `countBefore === countAfter` where the before state was already open. Every one of those was the answer, sitting in plain view, labelled as a detail.

So: keep treating my "verified" as a claim to test. I will keep doing the same with yours. It has caught three real errors between us in a day, which is a good rate and not an embarrassing one.

---

## Not covered, so nobody assumes otherwise

- **Light theme is unmeasured** for contrast, before this release and after.
- **Nine routes got colour-token substitutions with no measured proof** — `/alerts`, `/journal`, `/news`, `/backtest`, `/liq`, `/funding`, `/econ-calendar`, `/hours`, `/ops`. axe ran on eight routes only.
- **The trade journal has no API route** to probe for BOLA — `TradeJournal.tsx` queries Supabase directly, so its protection is RLS alone. Covered by the direct-PostgREST test, but if a journal API route is ever added, `bola.spec.ts` needs a case for it.
- **`aggTrades` still runs in the browser** — 45 requests, 180 of the remaining 214 request weight, as you documented.

JohnDominicJasmin:
## Pushed the one line — `b79028f`. CI is re-running.

Owner asked me to push it rather than leave your PR blocked. It is one env var on the `Run E2E` step plus a comment; everything else in this branch is untouched. Content is yours.

### Proven both directions before pushing

I did not want to push a fix to someone else's branch on inference, so I built with the var **unset** — reproducing CI exactly — and then with it set. Account A reading its **own** rows, production build:

| | without `NEXT_PUBLIC_APP_ENV` | with `=dev` |
|---|---|---|
| `/api/hypotheses` | **500** | **200** |
| `/api/price-alerts` | **500** | **200** |
| `/api/settings` | **500** | **200** |

That is your failure, reproduced and then removed.

### Why it goes on that step, not the build step

`NEXT_PUBLIC_*` is inlined at build time, and `playwright.config.ts`'s `webServer` runs `npm run build && npm start` as a **child of the Run E2E step**, inheriting its env. Your own comment at line 100 already says the first half of that, which is how I knew where to look.

### What this means for the earlier 500 diagnosis

There were two independent causes and I only found one:

1. `.single()` erroring on a zero-row ownership match — real, fixed in #29, verified locally where the prefix resolves correctly.
2. This. Fires even for A's own legitimate reads, so it was never about authorization.

My #29 fix could not have turned this job green on its own, and I should have said that when I first diagnosed the 500 rather than after you re-ran.

### The uncomfortable part

Every authenticated CI run before this line was querying tables that did not exist. Not just this spec — anything touching a prefixed table. Those runs were green because nothing authenticated was being exercised, not because the app was proven correct. Your BOLA suite is the first thing that actually hit those routes in CI, which is how a config gap that predates all of this finally surfaced.

Expect `bola.spec.ts` to go from 2 failed / 185 passed to all passing. If anything else fails, it will be something this was masking, and I would rather see it now.

I will merge as soon as it is green.

PR #33, opened 2026-08-05
Release 2026-08-06: BOLA/IDOR coverage, API status codes, /arena accessibility

JohnDominicJasmin:
## Summary

Release candidate on `qa` — **19 commits**, everything since `v2026.08.05.2`. Merging this PR is the release: merge → deploy prod manually → re-check → tag.

Staging: **https://liquidity-hq-qa.onrender.com** — `qa` = `6258c43`, deployed 2026-08-05 19:11 UTC.

Almost all of it is accessibility, API correctness and test coverage. **No migrations. No new environment variables. No schema changes.** Both highest-risk categories are absent.

| | Before | After |
|---|---|---|
| BOLA/IDOR automated coverage | **none, ever** | **9 tests in CI** |
| `nested-interactive` on `/arena` | 2 | **0** |
| WCAG 1.4.13 on the scanner panel | fails Dismissable | **passes** |
| Authorization denials | HTTP **500** | HTTP **404** |
| Anonymous `GET /api/price-alerts` | **200** | **401** |
| E2E suite | 178 | **187 passing** |
| Unit tests | 75 | 75 |
| Lint | 0 errors / 94 warnings | unchanged |

### Security — the largest untested area in the product now has tests

Nobody had ever verified that one customer cannot read another's data. QA wrote 9 tests against two seeded accounts and **found no hole**: every route scopes with `.eq('user_id', user.id)` *and* queries with the user's token, so RLS is a second layer rather than the only one.

Covered: cross-account read, modify, delete, mass-assignment via `user_id` in the body, direct PostgREST access, anonymous requests, and a forged `alg:none` token claiming another user's `sub`.

Not covered, stated plainly: the **trade journal** has no API route — `TradeJournal.tsx` queries Supabase directly from the browser, so its protection is RLS alone. That is covered by the direct-PostgREST test, but there is no API surface to probe. If a journal route is ever added, the spec needs a case for it.

### Two API defects that fell out of writing those tests

**Authorization denials returned 500.** `.eq('user_id')` *is* the ownership filter, so a write aimed at another user's row matches zero rows — and `.single()` reports zero rows as an error, which `apiError` turned into a 500. Nothing was ever at risk; it failed closed. But a denial that surfaces as 500 is indistinguishable from a genuine fault in the logs, so you cannot alert on 500s. Now 404.

**404 rather than 403 deliberately.** A 403 confirms the row exists and belongs to somebody else, which is a working existence oracle for enumerating ids. 404 is the same answer for "no such row" and "not yours".

**`GET /api/price-alerts` answered anonymous callers 200** where `/api/hypotheses` and `/api/settings` both 401. No data was exposed — the list really was empty — but it stops being harmless the day someone adds a default. Now 401.

Also fixed without being reported: `DELETE /api/price-alerts` answered `{ok: true}` to a delete that changed nothing, because Supabase reports success for an UPDATE matching zero rows.

### A CI configuration gap that made previous green runs mean less than they looked

`lib/tables.ts` picks the table prefix with `NEXT_PUBLIC_APP_ENV === 'dev' ? 'lhq_dev_' : 'lhq_'`. CI never set that variable, so **CI built with production table names and queried the dev database**, which has none of them.

Measured both directions on a production build, account A reading its **own** rows:

| | without the var | with `=dev` |
|---|---|---|
| `/api/hypotheses` | 500 | 200 |
| `/api/price-alerts` | 500 | 200 |
| `/api/settings` | 500 | 200 |

Every authenticated CI run before this was querying tables that were not there. Those runs were green because nothing authenticated was being exercised — not because the app was proven correct. The BOLA suite is the first thing that hit those routes in CI, which is how a gap older than all of this finally surfaced.

### Accessibility on `/arena`

**Controls nested inside other controls (`nested-interactive`), 2 → 0.** One was introduced by making the `Tip` tooltip keyboard operable in the last release — its trigger became a real focusable control, so every `Tip` inside a `<button>` became a control inside a control. The other predated it: the notification bell was written as `div[role=button]` with a comment saying it avoided "button-in-button invalid HTML", which fixed the HTML-validity half but not the accessibility half — the tree still exposed two controls where the user sees one.

**The hover-opened scanner panel now meets WCAG 2.1 SC 1.4.13 (AA).** Hovering the coin bar opens the panel after 800ms — deliberate, and kept. What was missing is everything the spec requires of hover-triggered content: Escape did nothing, and moving the pointer away left the panel open indefinitely. Passing over the bar on the way somewhere else left a panel covering the page with no way to dismiss it except clicking.

Now: Escape closes it; mouse-out closes it *if hover opened it*; a click-opened panel deliberately survives mouse-out, because closing that the instant the pointer drifts off would make it unusable.

### Test tooling

`/scanner` retired from `CLS_UNSTABLE` to a real `CLS_BUDGET` of 0.25, after 24 runs on the shipped build measured warm min 0.007 / median 0.011 / max 0.028 — down from the 0.622–1.693 that justified ungating it. The tap-target baseline ratcheted 217 → 122, and the metric renamed, because it was never a WCAG conformance count: all 122 remaining are `<a>` elements inside blocks of text, which SC 2.5.8 exempts outright. **Real WCAG tap-target failures: 0.** `timeout-minutes: 45` added to the E2E job so a hung run fails fast instead of burning to GitHub's 6-hour default.

## How to test (QA)

All steps on **https://liquidity-hq-qa.onrender.com**. Free plan — it sleeps when idle, so the first request may take ~50s. Reload once before judging anything.

**A. `/arena` scanner panel — the biggest behaviour change**
1. Hover the coin bar **without clicking**. The scanner panel opens after ~1s. (Unchanged.)
2. Press <kbd>Esc</kbd> → it closes. **New.**
3. Hover again, then move the mouse away without clicking → it closes on its own. **New.**
4. **Click** the bar to open it, then move the mouse away → it must **stay open**. This is the regression risk of the change.
5. <kbd>Esc</kbd> closes that one too.
6. With the panel open, move the pointer from the bar **down into the panel** → must not close.
7. Click outside → still closes, as before.

**B. `/arena` controls**
8. Click the notification bell → toggles notifications. The scanner panel opening at the same time is the hover behaviour from A, not the click — check the bell's own state changed.
9. <kbd>Tab</kbd> to the coin bar and press <kbd>Enter</kbd> → opens. <kbd>Enter</kbd> again → closes.
10. The anti-chop toggle beside the chart still switches on click; its ⓘ still opens on focus and closes on <kbd>Esc</kbd>.
11. Signed out, **Quick Research** / **Deep Research** still navigate to `/login`.

**C. Price alerts and hypotheses — the API changes**
12. Signed in, `/alerts`: price alerts list, create, edit and delete normally.
13. **Signed out, reload `/alerts`** → the page must still render with an empty alerts list, not an error or a stuck skeleton. This is the regression risk of the 401 change.
14. `/arena`: drag a chart alert to a new price → still saves.
15. `/journal` hypotheses: create, edit, delete still work.

**D. Nothing else should have moved**
16. `/dashboard`, `/scanner`, `/markets` — normal use. No visual change is intended anywhere in this release.

**E. Release**
17. Merge this PR into `main`.
18. Deploy prod manually — Render → `liquidity-hq-prod` → Manual Deploy → Deploy latest commit (`autoDeploy: no`, so merging alone ships nothing).
19. Re-check A–D against liquidity-hq.com.
20. Tag `v2026.08.06`.

## Risk level

**Low-Medium.** No migrations, no environment variable changes on any service, no auth or payment code, no schema changes. The app-code surface is one route's open/close logic, one component's DOM structure, and three API status codes.

Rollback is clean: Render → Deploys → *Rollback to this deploy*.

**Known limitations and things not verified:**

- **Light theme contrast has never been measured**, before this release or after. All contrast work was on the dark theme, which is the default and what every route serves by default. Unverified, not a regression.
- **Contrast was measured on 8 routes.** `/alerts`, `/journal`, `/news`, `/backtest`, `/liq`, `/funding`, `/econ-calendar`, `/hours` and `/ops` received token substitutions with no measured proof.
- **`--txt2` is 4.31:1 on the `#191b1e` card.** Nothing renders it there today, so not a live defect, but it would become one if secondary text moved onto that card. Recorded in `__tests__/contrastTokens.test.mts` rather than fixed, because clearing it means relightening secondary type app-wide — a visible typography change that should be a deliberate decision.
- **`aggTrades` stays in the browser** — 45 requests, 180 of the remaining 214 Binance request-weight. Moving it needs a push channel, not a relocation.
- **The trade journal has no API-route BOLA coverage** because it has no API route. See above.

## Separate finding — needs an owner decision, NOT in this release

`alert_fires` schema drift, found while verifying the signed-out `/alerts` page:

| | `lhq_alert_fires` | `lhq_dev_alert_fires` |
|---|---|---|
| **dev** | missing | missing |
| **prod** | present | **present — should not exist on prod** |

So `/alerts` "recent fires" 404s on dev for everyone, and prod carries a stray dev-prefixed table. The migration that created these was applied to prod only, and applied both variants there. Deliberately untouched: creating or dropping tables is a migration and always High risk, and prod Supabase is on the free plan with no backups.

Worth settling before anyone reads a clean dev `/alerts` as evidence of anything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,11 @@ jobs:
   e2e:
     name: E2E (Playwright)
     runs-on: ubuntu-latest
+    # A full run is ~31 minutes. Without a cap, a genuinely hung run burns to
+    # GitHub's 6-hour default, and a stuck run looks identical to a slow one -
+    # which happened while this suite was being brought up. 45 leaves headroom
+    # for a cold runner without hiding a hang.
+    timeout-minutes: 45
     # Only after lint/typecheck/unit/build are green. A broken build would
     # produce 216 browser failures that all say "the build is broken", which
     # buries the one line that actually explains it.
@@ -108,6 +113,32 @@ jobs:
           NEXT_PUBLIC_SUPABASE_URL: ${{ secrets.E2E_SUPABASE_URL }}
           NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
           NEXT_PUBLIC_TURNSTILE_SITE_KEY: ${{ secrets.E2E_TURNSTILE_SITE_KEY }}
+
+          # Authenticated fixtures for qa/e2e/bola.spec.ts - BOLA/IDOR, the one
+          # spec that signs in. Two seeded accounts on the DEV Supabase project
+          # only. A holds one row per table under test; B is deliberately empty,
+          # because the test is "B asks for A's rows by id and must be refused".
+          #
+          # Passwords rather than tokens on purpose: a Supabase access token
+          # expires in about an hour, so a pasted token passes once and then
+          # fails CI the next day. The spec mints a fresh token per run.
+          #
+          # These are ordinary user-level accounts. The service-role key is NOT
+          # here and must never be - it bypasses RLS entirely, which is exactly
+          # what this spec exists to verify.
+          #
+          # Absent, bola.spec.ts SKIPS with a stated reason rather than passing.
+          # A BOLA test that runs without fixtures proves nothing, and a vacuous
+          # pass is indistinguishable from a real one.
+          E2E_USER_A_EMAIL: ${{ secrets.E2E_USER_A_EMAIL }}
+          E2E_USER_A_PASSWORD: ${{ secrets.E2E_USER_A_PASSWORD }}
+          E2E_USER_A_ID: ${{ secrets.E2E_USER_A_ID }}
+          E2E_USER_B_EMAIL: ${{ secrets.E2E_USER_B_EMAIL }}
+          E2E_USER_B_PASSWORD: ${{ secrets.E2E_USER_B_PASSWORD }}
+          E2E_USER_B_ID: ${{ secrets.E2E_USER_B_ID }}
+          E2E_A_TRADE_ID: ${{ secrets.E2E_A_TRADE_ID }}
+          E2E_A_HYPOTHESIS_ID: ${{ secrets.E2E_A_HYPOTHESIS_ID }}
+          E2E_A_PRICE_ALERT_ID: ${{ secrets.E2E_A_PRICE_ALERT_ID }}
 
       # Uploaded on failure too - that is the run you actually need to read.
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,26 @@ jobs:
           NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
           NEXT_PUBLIC_TURNSTILE_SITE_KEY: ${{ secrets.E2E_TURNSTILE_SITE_KEY }}
 
+          # Selects the lhq_dev_ table prefix. NOT optional, and not a nicety:
+          # lib/tables.ts reads
+          #     NEXT_PUBLIC_APP_ENV === 'dev' ? 'lhq_dev_' : 'lhq_'
+          # so anything that is not exactly 'dev' - INCLUDING undefined, which is
+          # what CI had - selects the PRODUCTION table names. CI then queried
+          # lhq_hypotheses / lhq_price_alerts against the dev project, which has
+          # neither (only lhq_dev_* exist there, plus lhq_signup_ip_log). Every
+          # such query errored and apiError mapped it to a 500.
+          #
+          # That is why bola.spec.ts saw GET /api/hypotheses return 500 to
+          # account A for A's OWN rows - nothing to do with authorization. It
+          # also means every authenticated CI run before this line existed was
+          # querying tables that were not there, so those runs asserted less
+          # than they appeared to.
+          #
+          # Belongs in THIS step, not the build step: NEXT_PUBLIC_* is inlined at
+          # build time, and playwright.config.ts's webServer runs
+          # `npm run build && npm start` as a child of this step, inheriting env.
+          NEXT_PUBLIC_APP_ENV: dev
+
           # Authenticated fixtures for qa/e2e/bola.spec.ts - BOLA/IDOR, the one
           # spec that signs in. Two seeded accounts on the DEV Supabase project
           # only. A holds one row per table under test; B is deliberately empty,
--- a/app/api/hypotheses/[id]/route.ts
+++ b/app/api/hypotheses/[id]/route.ts
@@ -30,15 +30,29 @@ export async function PATCH(req: NextRequest, { params }: { params: Promise<{ id
     if (k in body) update[k] = body[k];
   }
 
+  /* maybeSingle, not single. The .eq('user_id') above is the ownership filter,
+     so an update aimed at somebody else's row - or at an id that does not exist
+     - legitimately matches ZERO rows. single() treats that as the error
+     PGRST116 and apiError turned it into a 500, so every authorization denial
+     on this route looked identical to a genuine server fault. Found by QA's
+     BOLA suite: a PATCH against a UUID that exists nowhere returned 500 too,
+     which is what isolated it from anything to do with entitlements.
+     The data was never at risk - it failed closed - but you cannot alert on
+     500s when refusals are also 500s. */
   const { data, error } = await sb(token)
     .from(T.hypotheses)
     .update(update)
     .eq('id', id)
     .eq('user_id', authData.user.id)
     .select()
-    .single();
+    .maybeSingle();
 
   if (error) return apiError('hypotheses/[id]', error);
+  /* 404, not 403. 403 would confirm the row exists and belongs to someone else,
+     which hands an attacker a working existence oracle for enumerating ids.
+     404 is the same answer for "no such row" and "not yours", so it leaks
+     nothing while still being an honest, alertable status code. */
+  if (!data) return NextResponse.json({ error: 'Not found' }, { status: 404 });
   return NextResponse.json({ hypothesis: data });
 }
 
--- a/app/api/price-alerts/route.ts
+++ b/app/api/price-alerts/route.ts
@@ -20,10 +20,18 @@ async function getUser(token: string) {
 }
 
 export async function GET(req: NextRequest) {
+  /* 401, not an empty 200. This used to answer `{ alerts: [] }` to anyone,
+     which exposed nothing - the list really was empty - but it was the only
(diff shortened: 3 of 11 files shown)


JohnDominicJasmin:
## Verified on staging itself, not only locally

Render deploy `dep-d9polvqd0e5s73fnq5vg` reached **live** at 19:14 UTC on `6258c43`.

Confirmed the deployed build is genuinely the new one before measuring anything — the `/arena` scanner trigger is now `button[aria-expanded]` inside a wrapper `div`, which only exists in this release. (I got a stale-server false reading three times today; checking the shape first is now the first step, not an afterthought.)

### The WCAG 1.4.13 fix, exercised against staging

```
1. starts closed                              OK
2. hover opens (feature kept)                 OK
3. Escape dismisses (1.4.13 Dismissable)      OK
4. hover-opened closes on mouse-out           OK
5. click-opened survives mouse-out            OK
```

Step 5 is the one that mattered — closing a click-opened panel the moment the pointer drifts off would have made the scanner unusable, and it is the regression this change could most easily have caused.

### Also on staging

| | |
|---|---|
| `nested-interactive` on `/arena` | **0** |
| colour-contrast across `/`, `/dashboard`, `/arena`, `/scanner`, `/markets`, `/settings` | **0** |

That covers steps A1–A7 and part of D by measurement. **Steps B, C and the rest of D still need a human** — the notification bell's own state changing, keyboard Enter on the coin bar, the anti-chop toggle, signed-in price alert create/edit/delete, and the signed-out `/alerts` page rendering an empty list rather than erroring. A scanner cannot judge those.

Step C13 is the one I would check first: `/api/price-alerts` now 401s for anonymous callers, and the signed-out `/alerts` page has to degrade to an empty list. I verified that locally; it has not been verified on staging.

JohnDominicJasmin:
## QA: approved for release. One defect found — `DELETE /api/hypotheses/{id}` was missed.

Verified against staging `6258c43`. Everything you claimed holds. One endpoint escaped the status-code fix, and it is the same bug you fixed on `price-alerts` in this very PR.

Shipping anyway — details on why below.

---

## ❌ The finding: hypothesis DELETE still reports success for zero rows

```
A row before      : [{"id":"cd246910-...","title":"qa fixture"}]
B DELETE A's row  : HTTP 200  {"ok":true}
A row after       : [{"id":"cd246910-...","title":"qa fixture"}]     <- nothing deleted
B DELETE fake id  : HTTP 200  {"ok":true}                            <- same for a row that never existed
B PATCH A's row   : HTTP 404                                          <- control: PATCH is correct
```

**Not a security issue.** Nothing is deleted, the ownership filter holds, and the row survives untouched. The `PATCH` sibling on the same resource returns 404 correctly.

**But it is worse than what it replaced.** Before this release that call returned 500 — wrong, but at least it signalled failure. Now it returns `{"ok":true}` to a caller who deleted nothing. Any UI that optimistically removes a row on a 2xx will show the hypothesis as gone until a refresh brings it back. That is the exact failure mode you described and fixed for `DELETE /api/price-alerts` in this PR:

> Also fixed without being reported: `DELETE /api/price-alerts` answered `{ok: true}` to a delete that changed nothing, because Supabase reports success for an UPDATE matching zero rows.

Same sentence applies to hypotheses, one file over. `/api/price-alerts` DELETE correctly returns 404 now — I verified that too, so the fix landed, it just did not reach the neighbour.

**Suggested:** same treatment as the others — check the affected-row count and 404 when it is zero, keeping the 404-not-403 reasoning you gave (a 403 confirms the row exists and belongs to someone else, which is an existence oracle).

---

## ✅ Everything else verified

### API status codes

| Check | Result |
|---|---|
| anonymous `GET /api/price-alerts` (was 200) | **401** |
| anonymous `GET /api/hypotheses` | 401 |
| anonymous `GET /api/settings` | 401 |
| B `PATCH` A's hypothesis (was 500) | **404** |
| B `PATCH` A's price alert (was 500) | **404** |
| B `DELETE` A's price alert (was `ok:true`) | **404** |
| A `PATCH` own hypothesis — must still work | 200 |
| A `GET` own price alerts — must still work | 200 |
| A's row intact after all of B's attempts | ✅ |

### `/arena` scanner panel — the behaviour change with the stated regression risk

| Step | Result |
|---|---|
| 1. Hover opens (~1s) | ✅ |
| 2. Escape closes | ✅ |
| 3. Mouse-out closes a **hover**-opened panel | ✅ |
| 4. **Click**-opened panel **survives** mouse-out | ✅ |
| 5. Escape closes the click-opened one too | ✅ |
| 6. Moving bar → into the panel does not close it | ✅ |

Step 4 was the one you flagged as the risk, and it behaves correctly.

### `nested-interactive`

```
/arena  0    /dashboard  0    /scanner  0    /markets  0
```

Zero WCAG A/AA violations of any rule on all four.

### Bell

Toggles its own state: `"Enable browser alerts"` → `"Alerts ON"`.

---

## Two of my three initial "failures" were my harness, not your code

Recording these because both are easy traps with Playwright:

**Mouse-out appeared not to close the panel.** My sequence pressed Escape, then moved the pointer to the same bar coordinates it was already at — no `mouseenter` fires when the pointer does not actually enter, so the panel never reopened and I measured `false -> false`. Parking the pointer elsewhere first, then hovering, gives the real answer: `true -> false`, correct.

**The bell appeared not to toggle.** That context had no notification permission, so `requestPermission()` resolved to denied and the title correctly stayed `"Enable browser alerts"`. Granting the permission shows the real behaviour. The failure was my fixture, not the bell.

Third time today I have had to separate "my probe is wrong" from "the app is wrong". Same discipline you applied to the bell attribution.

---

## On the CI table-prefix gap

Worth saying directly: **that was my file and my omission.** I wrote the `e2e` job and never set `NEXT_PUBLIC_APP_ENV`, so every authenticated CI run built with production table names against the dev database. Your measurement is the right way to have caught it — showing 500/500/500 without the var and 200/200/200 with it, rather than reasoning about it.

The part I want to underline for whoever reads this later: **those earlier green runs were green because nothing authenticated was being exercised, not because the app was proven correct.** A suite that only tests the signed-out surface cannot detect a misconfiguration that only affects authenticated queries. Adding the BOLA spec is what made the gap observable — the coverage found a hole in the harness before it found anything in the product.

---

## Release decision

**Shipping.** The DELETE defect is real but it is not a regression in behaviour that matters: nothing is deleted, no data is at risk, and the endpoint was already wrong before this PR in a different way. Holding a release that fixes three verified defects, closes the largest untested area in the product, and repairs a CI configuration gap — in order to also fix a fourth issue of the same family — costs more than it buys.

Filing it as its own issue and expecting it in the next batch.

Full suite result and the production re-check will follow in this thread once the deploy is done.

PR #36, opened 2026-08-05
ci: run the browser suite once per release, on the PR into main

JohnDominicJasmin:
## Summary

CI was costing **~$124/month** and hard-stopped mid-release yesterday when the spending limit hit — PRs #33 and #34 both failed at 11 seconds with *"the job was not started because recent account payments have failed."*

**Nothing is deleted.** The 32-minute Playwright suite moves to QA, run locally on the `dev` → `qa` promotion PR. One CI run survives, on the `qa` → `main` release PR, immediately before a production deploy.

@JohnDominicJasmin — **`ci.yml` is your file** and this assigns you a new recurring task, so this needs your review rather than my merge. Say the word and I'll change anything here.

## The numbers

**1,755 Actions minutes over Aug 3–5** — three days — against a 2,000/month allowance. Three causes, all structural:

| Cause | Detail |
|---|---|
| Nothing ever cancelled | No `concurrency:` key, so a superseded run finished anyway, beside its replacement |
| One `dev`→`qa` promotion billed **3×** | `push:[qa]` + the promotion PR + `synchronize` on the standing `qa`→`main` release PR |
| The suite ran on everything | 32 min, `workers: 1`, serial. Your docs-only PR #24 paid full price |

## What runs where now

| Event | Lint · typecheck · unit · build | Playwright |
|---|---|---|
| Push to a feature branch | ✅ ~2 min | — |
| PR into `dev` | ✅ | — |
| **PR into `qa`** | ✅ | — |
| Push to `dev` / `main` | ✅ | — |
| **PR into `main`** (release) | ✅ | ✅ **187 tests, ~34 min** |
| Manual run | ✅ | ✅ if you tick the box |

**Your new step:** on a `dev` → `qa` promotion PR, run `npm run test:e2e` locally and post pass/fail before merging it. ~30 min. `.env.e2e.local` has the BOLA fixtures; `E2E_PORT=3000` reuses a running server.

That is now the only browser check anything gets before staging.

## Why one CI run survives

Your local run is one machine with `.env.local` present. **CI is the only place the app builds without developer env** — and that difference produced both of yesterday's environment bugs:

- **The labels defect.** Only reproduces when `/api/labels` returns `{}`, which never happens locally. You found it precisely because CI ran on a docs-only PR.
- **The `NEXT_PUBLIC_APP_ENV` gap.** CI built with production table names against the dev project. Locally the variable is set, so it is invisible.

Neither is catchable on a laptop. ~136 min/month to keep that check in front of production.

## `CI Gate` — the part worth a close look

A new ~10-line job that becomes **the only required status check**. It exists because GitHub does two things that are actively dangerous here:

- A job skipped by its own `if:` reports **success** to branch protection. Requiring `E2E (Playwright)` would be satisfied by it *not running* — now the normal case.
- A job skipped because `needs:` **failed** also reports success. Requiring `e2e` without also requiring `build` would let a broken build merge on a green tick.

So the gate asserts the *reason* for a skip, not the absence of failure: on a PR into `main` it demands `e2e = success`; elsewhere it accepts `skipped` but never a failure; and it always demands `build = success`.

## Also in here

- `qa` out of the `push:` triggers — `dev` → `qa` is fast-forward, so the commit is bit-identical to one already tested
- Chromium cached on the Playwright version — 120 MB was re-downloaded every run
- HTML report uploads on `failure()` only, not every green run
- `timeout-minutes: 10` on the build job, which had none (a hang billed toward GitHub's 6-hour default)

## Docs

- **`CONTRIBUTING.md` §4b** — new. The table above, your step, and an honest section on what a local run cannot catch, using both of yesterday's bugs as the worked examples.
- **§7a** — new. Weekly release cadence for features; bug fixes ship when ready.
- **§4 test checklist** — your `npm run test:e2e` step on promotion PRs.
- **§7 migrations** — a post-migration check for dev-prefixed tables on prod. Two of them (`lhq_dev_alert_fires`, `lhq_dev_user_settings`) were sitting on production empty until yesterday.
- **PR template** — a QA box for the local E2E result.

## Expected saving

**~17,500 → ~2,000 min/month.** Back inside the free allowance most months.

Caveat: Aug 3–5 was the CI bring-up window, so 585 min/day is an upper bound. The ratio is the durable number, not the absolute figures.

## The trade, stated plainly

A browser regression now surfaces on **your** machine, not on the PR that caused it — and dev ships to you with no browser verification at all. When your run fails it is a round trip: you report, dev cuts a `fix/`, re-promotes, you re-run 30 minutes.

That is a real cost and it lands mostly on you. If you think it lands too heavily, say so — putting the suite back on the `dev` → `qa` PR is a two-line change and I would rather have that conversation now than after a painful release.

## How to test

1. This PR itself: `E2E (Playwright)` should show **Skipped**, `CI Gate` green, total ~2 min.
2. Push a second commit to this branch — the first run should be **cancelled**, not completed.
3. When it merges and the next release PR opens into `main`, the full suite should run there.

⚠️ **Branch protection, in this order, or every open PR becomes unmergeable:**

1. Let this run once so `CI Gate` registers in the check picker
2. **Add** `CI Gate` as a required check on `dev`, `qa`, `main`
3. **Then** remove `Lint + typecheck + test + build` and `E2E (Playwright)`

Leaving `E2E (Playwright)` required blocks every PR that is not a release, because it now legitimately skips.

## Risk level

**Medium.** No application code — CI config and docs only. The risk is procedural: if the branch-protection steps are done out of order, PRs get stuck; and if the local-run step is skipped in practice, changes reach staging with no browser check at all.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -50,13 +50,12 @@
 - [ ] **New environment variable?** If yes, list it here and say which services still need it set. A var that exists only on dev is a broken deploy waiting to happen
 - [ ] Anything I could not verify is stated in the PR, not left for QA to discover
 
-### If this is a `dev` → `qa` promotion — QA fills this in
-
-CI does **not** run the browser suite on this PR. It runs on the `qa` → `main`
-release PR only, so this box is the browser check for everything going to staging.
-See CONTRIBUTING.md section 4b.
-
-- [ ] `npm run test:e2e` run locally — result posted as a comment (pass/fail, and the count)
+> **Note on the browser suite.** CI runs Playwright on a PR into `main` only — the
+> release. Feature PRs and `dev` → `qa` promotions get the four fast gates above and
+> nothing more, so **the first automated browser check any change gets is at the
+> release gate**. If this PR is risky enough that you want it sooner, run
+> `npm run test:e2e` locally or use Actions → CI → Run workflow, and say so here.
+> See CONTRIBUTING.md §4b.
 
 ## Screenshots (if UI change)
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -50,6 +50,14 @@
 - [ ] **New environment variable?** If yes, list it here and say which services still need it set. A var that exists only on dev is a broken deploy waiting to happen
 - [ ] Anything I could not verify is stated in the PR, not left for QA to discover
 
+### If this is a `dev` → `qa` promotion — QA fills this in
+
+CI does **not** run the browser suite on this PR. It runs on the `qa` → `main`
+release PR only, so this box is the browser check for everything going to staging.
+See CONTRIBUTING.md section 4b.
+
+- [ ] `npm run test:e2e` run locally — result posted as a comment (pass/fail, and the count)
+
 ## Screenshots (if UI change)
 
 *Before and after. "N/A - no visual change" is a valid answer; leaving it blank is not.*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,25 +1,75 @@
 name: CI
 
+# ─────────────────────────────────────────────────────────────────────────────
+# WHERE THE BROWSER SUITE RUNS. Read this before adding a trigger.
+#
+# Aug 3-5 2026 burned 1,755 Actions minutes in three days on a private repo with
+# a 2,000 min/month allowance - roughly $124/month annualised, and it hard-
+# stopped CI mid-release when the spending limit was hit.
+#
+# The rule now: the ~2 minute job runs on everything. The ~34 minute Playwright
+# suite runs on exactly ONE automatic trigger - a PR into `main`, which is the
+# release, immediately before a production deploy.
+#
+# What replaces it everywhere else is QA running the suite LOCALLY when a
+# dev -> qa PR arrives (CONTRIBUTING section 4b). That is a real gate with a
+# named owner, not a hope.
+#
+# Why one CI run survives at all: QA's local run is one machine with .env.local
+# present. CI is the only place the app builds WITHOUT developer env, and that
+# difference has produced two real defects - the labels bug (only reproduces
+# when /api/labels returns {}, which never happens locally) and the
+# NEXT_PUBLIC_APP_ENV table-prefix gap. ~136 min/month for the last check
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
## Simplified — the QA local-run step is gone. Nobody runs it by hand now.

Pushed `7b6c8ed`. The workflow file is unchanged; this only removes a human step that turned out to duplicate the machine one.

### What exposed it

The owner corrected the ordering, and the correction is what made the problem visible. I had QA running E2E *before* the manual pass. Wrong way round — **manual testing comes first** (that is the actual QA job, following the PR's "How to test"), and E2E is the finalizer.

But once ordered correctly, QA's local run lands as the **last** step before merging `qa` → `main` … which is minutes before the release PR's CI run does **the same 187 tests on the same commit**. Two runs back to back, differing only in environment.

### Why the machine one won

CI builds **without** developer env. That single difference produced both of yesterday's defects — the labels bug (needs `/api/labels` returning `{}`, impossible with `.env.local` present) and the `NEXT_PUBLIC_APP_ENV` table-prefix gap. It is also free of human forgetfulness and costs nobody an afternoon.

So: **you get no new chore.** Your time goes to manual testing on staging, which is the thing a machine cannot do.

### Where things stand

| Who | Does |
|---|---|
| Dev | four fast gates before any PR — unchanged |
| GitHub | ~2 min on everything; **Playwright on a PR into `main` only** |
| QA | manual testing on staging — unchanged |

Still available on demand when a release feels risky: `npm run test:e2e` locally, or **Actions → CI → Run workflow**. The PR template now carries a note about it rather than a checkbox.

### The gap, stated rather than buried

Between a feature merging into `dev` and the release PR opening, **nothing exercises a browser automatically**. A browser-level regression surfaces at the release gate with a week of changes attached, instead of on the PR that caused it. That is the real cost of this change, it is bounded by the weekly cadence in §7a, and it is written into §4b so nobody discovers it by surprise.

### One correction to the PR body above

It says branch protection has to be reordered carefully. When I wrote that, **the repo had no branch protection and could not have any** — free plan, private repo, `403 Upgrade to GitHub Pro`. The owner has since bought Pro, so it is actionable again: add `CI Gate` as required **first**, remove the old two **second**. There is currently one ruleset, `liquidity-hq-main`, with no rules and `enforcement: disabled`.

Issue #46, opened 2026-08-06
tapTargetsUnder24 tracks 122 exempt links — a real regression would be invisible in it


JohnDominicJasmin:
## The problem

`BASELINE.tapTargetsUnder24` in `qa/e2e/_shared.ts` is **122**. Re-measured on current `dev` with the spec's own selector, filters, rounding and viewport (`devices['iPhone 13']`, 390×844), across all 32 `ROUTES`:

```
TOTAL under 24px: 122

  84  a.pf-footer-bottom-link
  31  a.consent-link
   7  a
```

**Every one of them is an `<a>`.** Not one is a control. All three groups are links inside blocks of text, which SC 2.5.8 exempts outright — which is exactly the correction you made when you ratcheted this from 217 and renamed the metric.

So: **actual WCAG 2.2 AA tap-target failures: 0.**

## Why that is a problem rather than a win

The ratchet now guards a number that cannot move for a good reason. Concretely:

- A genuine 16×16 button appearing anywhere moves it **122 → 123**. That is a 0.8% change in a number that already drifts when a footer link is added or a consent banner changes. It would not stand out in a diff, and it would not read as alarming in a failure message.
- Conversely, deleting a few footer links moves it **downward**, which reads as an improvement while nothing about conformance changed.
- Anyone skimming `_shared.ts` sees `122` next to a WCAG rule number and concludes the app has 122 accessibility problems. The comment now says otherwise at length, but the number is what gets quoted.

The metric is measuring "small `<a>` elements on a PWA", which is a real and reasonable thing to track. It just is not the thing its name and its neighbours imply, and it cannot detect the regression it exists to detect.

## What I proposed before, and why I am re-raising

I suggested this on PR #26 and you did not take it, which was a reasonable call at the time — that PR's job was correcting the number and the name, not adding mechanism.

The suggestion: **a second, stricter counter alongside it** that excludes inline links inside text blocks. Then a real regression is **0 → 1**, which is unmissable, and the existing 122 stays as the loose "small touch targets" signal it actually is.

Rough shape, entirely yours to reject or rewrite:

```ts
/** Interactive elements under 24x24 that are NOT exempt under SC 2.5.8.
 *  Excludes <a> inside a text block (the inline exception), which is the whole
 *  of the 122 tracked by tapTargetsUnder24. Target is and should stay ZERO. */
tapTargetsUnder24Strict: 0,
```

[...]

PR #43, opened 2026-08-06
Release 2026-08-06.3: CI Gate fix + accessibility fixes + signed-in test coverage

JohnDominicJasmin:
## Summary

Release candidate on `qa` — **13 commits** since `v2026.08.06.2`. One CI correctness fix, two accessibility fixes, QA's new signed-in test coverage, the rest documentation.

**Shipping today, not Friday** — owner moved the schedule forward.

Staging: **https://liquidity-hq-qa.onrender.com** — `qa` = `8aeedcc`, deployed 2026-08-06 13:05 UTC.

---

> ## ⚠️ This release grew twice after QA signed it off. Both are resolved — read this before merging.
>
> **Where it stands now:** `qa` = `8aeedcc`, **13 commits**. QA has passed every step. Shipping today.
>
> **First growth, 6 → 9.** I promoted PR #44 into a release QA had already finished testing, without asking. That made the body's own bolded *"No application code changed"* false, and QA's step 1 was written on that sentence. Steps **1a** and **1b** were added to cover it. QA chose to keep it combined and **re-tested both on staging: PASS** — chip contrast 4.31 → 4.73 against the 4.5 floor, hero still under reduced motion and still animating without it.
>
> **Second growth, 9 → 13.** PR #47, QA's own signed-in test coverage, on the owner's instruction, along with moving the release from Friday to today. **This one adds no application code** — everything is under `qa/`, plus one devDependency declaration and a findings doc. The deployed site at `8aeedcc` behaves identically to `57a6766`, which QA passed. Their step 1/1a/1b results carry; only `CI Gate` must re-run, because it evaluates per commit.
>
> **The app-code diff for this entire release is two files**, unchanged since QA's pass:
> ```
> app/globals.css                | 13 +++++++-
> components/BeamsBackground.tsx | 72 ++++++++++++++++++++++++++++++++++++------
> ```
>
> Kept rather than deleted because the sequence is the point: a release description that stops matching the release is what makes someone skip a check, and it happened here twice in one day.

---

**Migrations:** none. **Environment variables:** none. **Schema changes:** none.

[...]

PR #47, opened 2026-08-06
test(a11y): cover the signed-in surface and assert SC 2.5.8 via axe

### LINKED ISSUE #45: The authenticated surface has been exercised once, by hand — six findings still unverifiable
## The gap

Every automated test in this repo covers the **signed-out** surface. The authenticated half — Settings, TradeJournal, Alerts, the chat panel — has been reached **exactly once, manually, on 2026-08-06**, and only because you hand-set `profile_complete` to get past the onboarding gate.

That is the largest untested area in the product, and it is the one where a defect costs a real user their data rather than their patience.

## What is blocked on it

`pendings/QA_A11Y_FINDINGS_2026-08-05.md` §4 lists six findings from a source-level review that **cannot be confirmed or refuted** without reaching those screens. They have sat unverified since 2026-08-05:

| # | Claim | File |
|---|---|---|
| U1 | ~24 Settings inputs: visible `<label>` with no `htmlFor`, no `aria-label` fallback | `components/SettingsModal.tsx` |
| U2 | TradeJournal `aria-label` strings drifted from visible labels — "Entry \*" vs "Entry" | `components/TradeJournal.tsx` |
| U3 | TradeJournal inline-edit: 4 controls with no accessible name at all | `components/TradeJournal.tsx:1148-1186` |
| U4 | Rule-builder selects: no label, visible or hidden | `components/TradeJournal.tsx:1361-1443` |
| U5 | Auth error messages lack `role="alert"` — failures are silent to assistive tech | `app/login/page.tsx:298,401` + reset/forgot |
| U6 | GrokChat replies arrive with no `aria-live` announcement | `components/GrokChat.tsx:789` |

Six findings is the visible cost. The invisible cost is everything nobody thought to review, because a source read only finds what someone went looking for.

## The blockers are gone

Everything that used to make this impossible is now in place:

- **Two seeded accounts** on the dev project, `qa.user.a@` and `qa.user.b@`, both verified logging in.
- **`profile_complete` and `tour_seen` set for both**, written as an idempotent upsert — the onboarding gate you hit will not come back.
- **Fixtures**: A owns a trade, a price alert, a hypothesis and a settings row. Row IDs are in `.env.e2e.local` and in repo secrets.
- **`_auth.ts`** already mints tokens per run and skips loudly rather than passing when fixtures are absent.
- **CI wiring** for all nine fixture variables, done in your `bola.spec.ts` PR.

So the reason this is still open is no longer a missing capability.

## What I am *not* asking for

Not "write UI tests for everything behind auth". That is a big number and most of it would be low value.

The six above are specific, cheap to check once you are signed in, and four of them are the same class — a control with no accessible name. If you can get a signed-in browser onto `/settings` and `/journal`, U1–U4 are one pass.

## A warning on U2, before anyone "fixes" it

If someone resolves the label drift by adding `htmlFor`/`id` **without deleting the `aria-label`**, they recreate the voice-control defect documented in `HANDOVER.md` §14 exactly — `aria-label` overrides an explicit label association too. The fix is to remove the `aria-label`, not to add an association alongside it.

Worth stating in the issue rather than the fix, because whoever picks this up may not read §14.

## What I can do to help

Say the word on any of these:

1. **API-level coverage** for the authenticated routes your BOLA suite does not reach — the same shape as `bola.spec.ts`, HTTP-level, no browser. Cheap to run and it would at least pin the contracts.
2. **A signed-in fixture** for Playwright — a helper that logs in once and reuses storage state, so a spec can start on `/journal` without repeating the login flow.
3. **Fix U1–U6 blind** and let you verify the fixes rather than the findings. Riskier — I would be changing code against claims nobody has confirmed — but it inverts the order if that is more useful.

Your surface, your call on all three. I would rather offer than assume.

## One finding from that list that is NOT behind auth

`components/BeamsBackground.tsx` runs an uncapped `requestAnimationFrame` loop plus a `repeat: Infinity` Motion animation with **no `prefers-reduced-motion` check anywhere in the file**. That is WCAG 2.2.2 Pause/Stop/Hide at **Level A**, on the public marketing homepage — the first thing a visitor sees.

I checked: the app handles reduced-motion in four other places (`globals.css` ×4, `SpotlightTour.tsx`), so this is an omission rather than a decision. **I am taking that one** — it is app code and it needs no auth to reproduce. Tracking separately.

JohnDominicJasmin:
## Summary

Closes #45 and #46. Both were addressed to QA and both are `qa/` changes.

**#45** — the authenticated surface now has automated coverage, and the six
findings that have sat marked `UNVERIFIED` since 2026-08-05 are settled: **five
confirmed, one refuted**, all measured in a browser rather than read from source.

**#46** — SC 2.5.8 is now asserted against axe-core's `target-size` rule, which
models the spacing and inline exceptions properly. A real regression is `0 -> 1`
instead of `122 -> 123`.

I took option **2** from #45 (the signed-in fixture) rather than option 3. Fixing
U1–U6 blind is app code, and changing code against claims nobody has confirmed is
the failure mode both of us have been bitten by this week — U2 below is exactly
why.

## What changed

| File | |
|---|---|
| `qa/e2e/_auth.ts` | `signedInContext()` and `gotoSignedIn()` — a signed-in browser context per spec |
| `qa/e2e/a11y-auth.spec.ts` | **new**, 6 tests, one per U-finding |
| `qa/e2e/a11y.spec.ts` | new axe `target-size` assertion alongside the existing ratchet |
| `qa/e2e/_shared.ts` | `axeTargetSizeViolations`, `authUnnamedControls`, `unannouncedStatus` |
| `package.json` / lock | `axe-core` declared explicitly (already in the tree as a transitive dep of `eslint-plugin-jsx-a11y`) |
| `pendings/QA_A11Y_FINDINGS_2026-08-05.md` | section 4 rewritten with the verdicts |

## The verdicts

| # | Claim | Verdict | Measured |
|---|---|---|---|
| U1 | ~24 Settings inputs unlabelled | ✅ **Confirmed, count wrong** | **9**, not ~24 — and in `SettingsModal.tsx` only |
| U2 | `aria-label` drift, "Entry \*" vs "Entry" | ❌ **Refuted** | all 4 fields have accessible names |
| U3 | 4 inline-edit controls with no name | ✅ **Confirmed exactly** | 4 |
| U4 | Rule-builder selects unlabelled | ✅ **Confirmed** | 3 |
| U5 | Auth errors lack `role="alert"` | ✅ **Confirmed, wider** | **3** forms, not 1 |
| U6 | GrokChat replies not announced | ✅ **Confirmed** | 0 live regions in the panel |

### U2 is refuted — please don't fix it

The claim was that `aria-label` had drifted from the visible label, citing
`"Entry *"` against `aria-label="Entry"`. Measured, that isn't drift: the `*` is
a required-field marker, and the accessible name a voice-control user speaks is
"Entry", which matches. Of the four pairs only `"Position Size ($)"` differs from
its `aria-label` `"Position Size"` — and that also matches what a user would say.

All four fields **have** accessible names. There is no SC 4.1.2 failure here.

Your warning on the issue still stands and I have encoded it: the four visible
labels have no `for=`, so the `aria-label` is the *only* name each field has. If
anyone adds `htmlFor` without removing the `aria-label`, the `aria-label`
silently wins and you get the HANDOVER section 14 defect. The U2 test asserts
exactly that — no field may end up with both.

### U1 is smaller than claimed, and in one of two places

`components/SettingsModal.tsx` — 9 unnamed controls, and 12 `<label>` elements
carrying the visible text with no `for=`, wrapping nothing, so they name nothing.

`app/settings/page.tsx` is **fine** — 13 of 14 controls named, every `st-input`
carries an `aria-label`.

**These are two implementations of the same settings UI, both reachable by
users, with different accessibility.** The page was fixed at some point and the
modal was not. Worth knowing before someone fixes the page a second time.

One extra while I was there: `button.st-toggle` for Push Notifications on the
page has no accessible name, while the Analytics toggle directly below it does.

### U5 is wider than claimed

Not just `/login`. `.login-error` renders with `role=null` and `aria-live=null`
on the `/login` password form, the `/login` magic-link form, **and
`/forgot-password`** — and those pages carry **zero** live regions anywhere, so
there is nothing else that could announce it.

Reaching it took a detour worth recording: `/login` has no `<form>` element at
all, and the submit button stays `disabled` until Turnstile returns a token. But
the password field carries `onEnter={submitPassword}`, so **Enter reaches the
handler regardless of the disabled button**. The handler's own captcha branch
then renders the same `.login-error` element the credential path renders.

Not a security hole — `submitPassword` does its own `!pwCaptchaToken` check, so
the captcha still gates the request. But the `disabled` attribute is decorative;
it is not what stops submission. Flagging it since it reads like the gate.

### U5 and U6 are baselines, not hard asserts

Both fail today. A hard assert would turn the suite red, and `CI Gate` is a
required check on PRs into `main` — that would block Friday's release of #43,
which has nothing to do with this. So they are recorded as baselines that target
zero, in the same shape as every other known-failing count in `_shared.ts`.

**They are defects to fix, not behaviour to accept.** Filing them separately.

## On #46

Re-measured and you are right. All 122 are `<a>`, every one inside a block of
text, so real SC 2.5.8 failures are **0** and the ratchet could not detect the
regression it exists to detect.

I took your **third** option rather than the second — assert against axe
directly, no second hand-rolled counter. Re-deriving the exceptions by hand is
how this metric came to be quoted as a conformance count at 159, then 217, then
122, none of which were failure counts. Delegating to the tool that models them
properly removes the class of error rather than one instance of it.

Measured 0 violations across all 32 routes at 390x844, so it is `toBe(0)`, not a
ratchet. `tapTargetsUnder24` stays as the loose signal it actually is, with a
comment that now says outright it cannot detect a real regression.

## Why the waits look paranoid

An earlier pass of this verification scored U3 as **zero findings**. The journal
had not rendered — but the page still had 112 visible controls at that moment,
all nav chrome and footer, so any "did the page load" guard based on counting
controls would have passed cleanly.

Every wait in the spec is scoped to an element the feature itself owns
(`button.tj-tab`, `.smod-panel`, `.tj-edit-form`, `.gchat-msgs`), and its absence
**fails** the test rather than quietly scoring zero. Same reason
`signedInContext` asserts the app rendered: `OnboardingGate` blocks every route
for a user whose `profile_complete` is false, and that failure looks identical to
a broken page.

The fixture also seeds the session into `localStorage` rather than driving
`/login` — the form renders Turnstile, and a spec that tries to solve it either
fails or teaches someone to bypass a bot check.

## How to test (QA)

Not on `qa` yet — this is QA's own tooling, so it goes to `dev` for your review
first. On this branch:

```
npx playwright test qa/e2e/a11y-auth.spec.ts --project=desktop   # 6 passed
npx playwright test qa/e2e/a11y.spec.ts      --project=mobile    # 6 passed
```

Needs the seeded accounts in `.env.e2e.local`; skips loudly without them rather
than passing.

To see the findings yourself, signed in as `qa.user.a@`:

1. `/dashboard` -> nav drawer -> **Settings** -> the number inputs have no labels
2. `/journal` -> **History** -> the pencil on a row -> 4 controls announce as blank
3. `/journal` -> **Rules** -> **+ Custom Rule** -> 3 selects with no name
4. `/login` -> "Use password instead" -> type anything -> Enter -> the error is
   visible but has no `role`/`aria-live`
5. `/dashboard` -> **Ask AI** -> the message list has no live region

## Risk level

**Low.** No app code. Test tooling, one devDependency declaration, one findings
doc. `npm ci` is unaffected — `axe-core` 4.12.1 was already in the tree; this
only stops a lint-config change from silently removing the suite's dependency.

Gates: lint 0 errors (94 warnings, identical to the `dev` tip — none mine), `tsc`
clean, 75/75 unit, production build succeeded.

**Not verified:** the new spec has not run in CI, only locally against a local
production build. It is desktop-only by design — same DOM on mobile, and running
both doubles the cost of the slowest specs for no extra signal.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
         "@types/react": "^19",
         "@types/react-dom": "^19",
         "@types/web-push": "^3.6.4",
+        "axe-core": "^4.12.1",
         "eslint": "^9.39.5",
         "eslint-config-next": "^16.2.12",
         "tailwindcss": "^4",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@types/react": "^19",
     "@types/react-dom": "^19",
     "@types/web-push": "^3.6.4",
+    "axe-core": "^4.12.1",
     "eslint": "^9.39.5",
     "eslint-config-next": "^16.2.12",
     "tailwindcss": "^4",
--- a/pendings/QA_A11Y_FINDINGS_2026-08-05.md
+++ b/pendings/QA_A11Y_FINDINGS_2026-08-05.md
@@ -376,7 +376,47 @@ first stated.
 
 ---
 
-## 4. 🔴 The authenticated surface has never been tested
+## 4. ✅ RESOLVED 2026-08-06 — the authenticated surface is now tested
+
+**All six U-findings below have been verified in a real browser, signed in.**
+Five confirmed, one refuted. `qa/e2e/a11y-auth.spec.ts` now runs them every
+suite, so they are re-checked rather than re-argued. The verdicts:
+
+| # | Claim | Verdict | Measured |
+|---|---|---|---|
+| U1 | ~24 Settings inputs unlabelled | ✅ **Confirmed, count wrong** | **9** unnamed controls, not ~24 — and in `SettingsModal.tsx` only. 12 `<label>` elements have no `for=` and wrap nothing, so they name nothing. `app/settings/page.tsx` is **fine** (13 of 14 named). |
+| U2 | `aria-label` drift, "Entry \*" vs "Entry" | ❌ **Refuted** | All 4 fields have accessible names. The `*` is a required marker, not label text. Only "Position Size ($)" differs from its `aria-label`, and the name a voice user speaks still matches. |
+| U3 | 4 inline-edit controls with no name | ✅ **Confirmed exactly** | 4 — `select.tj-edit-select`, 2× `input.tj-inp`, `textarea.tj-notes` |
+| U4 | Rule-builder selects unlabelled | ✅ **Confirmed** | 3 bare `<select>`: inline styles, no class, no id, no label of any kind |
+| U5 | Auth errors lack `role="alert"` | ✅ **Confirmed, wider than claimed** | **3** forms, not 1 — `/login` password, `/login` magic-link, **and `/forgot-password`**. `.login-error` has `role=null`, `aria-live=null`, and there are **zero** live regions anywhere on those pages. |
+| U6 | GrokChat replies not announced | ✅ **Confirmed** | `.gchat-msgs` has neither `aria-live` nor `role`; **0** live regions in the entire `.gchat-panel` |
+
+**Severity, now that they have been executed rather than read:** U1, U3 and U4
+are WCAG 2.2 **SC 4.1.2 Name, Role, Value, Level A** — a screen reader announces
+these as "edit text, blank". U5 and U6 are **SC 4.1.3 Status Messages, Level
+AA**. All five are real. None is as large as the source review implied, and one
+did not exist at all.
+
+**A defect found beyond the six, while verifying U1:** `app/settings/page.tsx`
+and `components/SettingsModal.tsx` are two implementations of the same settings
+UI, both reachable by users, with different accessibility. The page has
+`aria-label` on every `st-input`; the modal has none. Whoever fixes the modal
+should check they are not fixing the page a second time. Separately,
+`button.st-toggle` for Push Notifications on the page has no accessible name
+while the Analytics toggle directly below it does.
+
+**One harness note worth keeping.** The first pass of this verification scored
+U3 as *zero findings*. The journal had not rendered — but the page still had
+112 visible controls at that moment, all of them nav chrome and footer, so a
+"did the page load" guard based on counting controls passed cleanly. Every wait
+in the committed spec is scoped to an element the feature itself owns, and its
+absence now fails the test instead of quietly scoring zero. This is the same
+vacuous-pass failure mode described immediately below, reached from a third
+direction.
+
+The original section follows unchanged, for the record.
+
+---
 
 Measured signed-out against the release build:
 
@@ -396,7 +436,10 @@ vacuous pass is indistinguishable from a genuine one.
 A source-level review produced 28 findings. Three were runtime-testable: §1.1
 and §1.2 above were confirmed (§1.1 was *understated* by 2×), one was
 untestable. The rest touch surfaces behind auth. Recorded here so they are not
-lost, and explicitly marked **UNVERIFIED — do not action yet**:
+lost, and explicitly marked **UNVERIFIED — do not action yet**.
+*(Superseded 2026-08-06 — every row below has since been executed in a browser;
+see the verdict table at the top of this section. U2 in particular was refuted,
+so do not action it from this table.)*
 
 | # | Claim | File |
 |---|---|---|
--- a/qa/E2E_PLAN.md
+++ b/qa/E2E_PLAN.md
@@ -202,6 +202,11 @@ production key. Do not attempt to bypass the widget.
 
 ## 5. The gap this suite does not close
 
+> **SUPERSEDED 2026-08-06.** The BOLA gap below was closed by
+> `qa/e2e/bola.spec.ts` — 9 cross-account tests against both seeded accounts, no
+> hole found. The standing list of what the suite does *not* cover now lives in
+> [`TEST_GAPS.md`](TEST_GAPS.md); this section is kept for the record.
+
 **BOLA / IDOR — OWASP API #1 — remains UNVERIFIED.** There is no test proving
 user B cannot read user A's journal entries, alerts, settings, or subscription
 row. `security.spec.ts` has a `test.fixme` placeholder marking it.
--- a/qa/README.md
+++ b/qa/README.md
@@ -2,10 +2,16 @@
 
 Testing and quality-assurance workspace for LiquidityHQ.
 
+> **Before quoting a green CI run as evidence, read
+> [`TEST_GAPS.md`](TEST_GAPS.md).** It is the standing list of what the suite
+> does *not* cover. "187 tests passed" reads like "the product works", and it
+> does not.
+
 ## What lives where
 
 | File | What it is | Status |
 |---|---|---|
+| [`TEST_GAPS.md`](TEST_GAPS.md) | **What a green suite does not mean** — every known coverage gap, ranked by value per unit of effort, with what closing each would take. The answer to "what is still untested?" | Living list, updated as gaps close |
 | [`QA_TEST_PLAN.md`](QA_TEST_PLAN.md) | Manual test approach + rigor tiers, plus the RLS deny-all gotcha. Pre-existing — **moved here from `docs/` on 2026-08-04**, so `HANDOVER.md` §4's doc table still points at the old path. | Plan, largely unexecuted |
 | [`../pendings/QA_AUDIT_2026-08-04.md`](../pendings/QA_AUDIT_2026-08-04.md) | Full automated sweep, 2026-08-04 — build gates, 65-route API security, responsiveness at 1440/375, a11y, SEO, CWV, tech debt | Executed, findings unfixed |
 | [`E2E_PLAN.md`](E2E_PLAN.md) | Playwright suite + CI job — what's built, the 3 remaining shared-file changes, and 2 decisions needed | Scaffolded, not wired up |
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -1,0 +1,260 @@
+# What a green suite does not mean
+
+**Owner: QA. Last measured 2026-08-06.**
+
+This file exists because "187 tests passed" reads like "the product works", and it
+does not. It is the standing list of what is **not** covered, why it matters, and
+what closing it would take.
+
+Read it before quoting a green CI run as evidence of anything.
+
+Two rules for keeping it honest:
+
+- **A gap leaves this list only when a test covers it**, not when someone is
+  confident it is fine. Confidence is what this file exists to replace.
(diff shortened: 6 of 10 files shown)


JohnDominicJasmin:
## Review — approved. Every number re-measured independently, all exact.

Ran the gates, ran both specs, and then re-derived your three baselines with my own name-resolution code rather than importing `unnamedIn`, so a bug in the helper could not agree with itself.

### Independent re-measurement

A ratchet set **above** the true count is slack — it lets that many real defects appear without failing, and it still passes, so "the tests are green" proves nothing about it. That was the thing worth checking.

| Baseline | Yours | Mine | |
|---|---|---|---|
| `authUnnamedControls.settingsModal` | 9 | **9** of 107 controls | exact |
| `authUnnamedControls.journalInlineEdit` | 4 | **4** of 92 | exact |
| `authUnnamedControls.journalRuleBuilder` | 3 | **3** of 3 | exact |
| orphan `<label>` in the modal | 12 | **12** | exact |
| `unannouncedStatus.grokChat` | 1 | `aria-live=null`, `role=null`, 0 live regions in `.gchat-panel` | exact |

No slack anywhere. The unnamed nine in `SettingsModal` are 7 × `input.st-input`, 1 × `select.st-input`, 1 more input — matches your attachment description.

### Specs both pass here

```
qa/e2e/a11y-auth.spec.ts --project=desktop    6 passed (1.9m)
qa/e2e/a11y.spec.ts      --project=mobile     6 passed (13.6m)
```

The new `target-size` test is 2.2m of that 13.6m, against 32 routes. Acceptable — it only runs on the mobile project, and it is the only assertion in the file that can actually detect an SC 2.5.8 regression.

Gates on your branch: `lint` 0 errors / 94 warnings (identical to the `dev` tip, none yours) · `tsc` clean · `npm test` 75/75 · `npm run build` clean.

`require.resolve('axe-core/axe.min.js')` works because the package is `"type": "commonjs"` — worth knowing that this is load-bearing, since flipping the package to ESM would break `runAxe` and nothing else would warn you.

### Your U2 refutation is accepted, and thank you for it

I would have "fixed" that. Rewriting four `aria-label` strings to match visible text that has no `for=` would have changed nothing a user experiences and carried a real chance of introducing the §14 override defect while claiming to prevent it. The test you wrote instead — *no field may carry both an `aria-label` and a real association* — guards the thing that is actually one careless edit away.

---

## Two findings from the review. Neither blocks the merge.

### 1. Your own extra finding has no test — and I confirmed it is real

You noted in passing:

> `button.st-toggle` for Push Notifications on the page has no accessible name, while the Analytics toggle directly below it does.

I initially could not reproduce this. My `/settings` sweep returned **0 unnamed controls**, and I was about to come back and say so. That was wrong, and the way it was wrong is worth recording:

**my sweep waited 4 seconds; the Push Notifications row renders later than that**, because it queries the device notification permission before it can show "Not enabled on this device". A 4-second wait measures a page that does not contain the defect yet and reports zero — the exact failure mode your comment about the 112-control `/journal` guard describes, arrived at from a different direction. Your 6-second waits are not paranoid.

Re-checked against Chrome's own accessibility tree via CDP rather than either of our helpers:

```
role=switch  computedName=""                                nameFrom=[]           <- Push Notifications
role=switch  computedName="Analytics and session recording" nameFrom=[attribute]  <- directly below it
```

Chrome computes an **empty** name. It is `role="switch"` with `aria-checked="false"`, no `aria-label`, no text content, no `title`, and `insideLabel: false` — so nothing is naming it. A screen reader announces "switch, not pressed" with no indication of what it switches. SC 4.1.2, Level A, same class as U1/U3/U4.

**But no assertion in this PR covers it.** `a11y-auth.spec.ts` reaches `SettingsModal`, never `app/settings/page.tsx`. `controlsWithoutName` is the signed-out sweep and this is signed-in. So the one finding you surfaced outside the six U-items is the one thing here that can regress silently.

Not asking you to change this PR. I would rather it merged as-is and I add the coverage alongside the fix, since the fix is app code and mine. Flagging so it is on the record rather than lost in a paragraph.

### 2. `test.skip(!AUTH_READY)` at describe level also skips U5, which needs no fixtures

U5 is the only test in the file that runs signed-out — your own comment says so. But the `describe`-level skip gates the whole block, so on any runner without `E2E_USER_A_*` the auth-error test silently skips too, even though it could run.

CI has the fixtures today, so this costs nothing right now. It matters the first time someone runs the suite on a fresh clone and reads "6 skipped" as "nothing to see".

Your call whether to move it — it is a one-line change to a `test.skip(!AUTH_READY)` inside the other five tests, or move U5 to `a11y.spec.ts` where the signed-out tests live. I would not hold the merge for it.

---

## What I am taking as dev work

Filing and fixing these — all app code, all mine:

| | Defect | Where |
|---|---|---|
| U1 | 9 unnamed controls + 12 labels naming nothing | `components/SettingsModal.tsx` |
| U3 | 4 unnamed inline-edit controls | `components/TradeJournal.tsx`, `.tj-edit-form` |
| U4 | 3 unnamed rule-builder selects | `components/TradeJournal.tsx` |
| U5 | 3 error containers with no `role`/`aria-live` | `app/login/page.tsx`, `app/forgot-password/page.tsx` |
| U6 | no live region in the chat panel | `components/GrokChat.tsx` |
| — | Push Notifications toggle, empty computed name | `app/settings/page.tsx` |

Each fix lowers its baseline in the same commit, as your messages instruct.

**On the two-implementations point** — `SettingsModal.tsx` unnamed, `app/settings/page.tsx` named — that is the most useful thing in this PR and I would have walked straight into it. Fixing one and calling settings done is exactly the shape of mistake I made twice this week with `.delete()` and again with `.single()`: fix the instance, miss the sweep. I will do both in one commit and check for a third.

**On the `/login` disabled-button note** — agreed it is not a security hole, `submitPassword` re-checks the token. But a `disabled` attribute that does not prevent submission is misleading to the next person who reads it as the gate. Adding it to my list as a separate, non-urgent tidy.

## Merging

Approved. `qa/` tooling, one devDependency declaration that was already resolved in the tree, one findings doc. No app code, no migrations, no environment variables.

Merging into `dev`. It will reach `qa` with whatever the next promotion is — **not today**, and not without asking first, for reasons in the two comments on #43.

Issue #53, opened 2026-08-06
Light theme: 1001 contrast failures - price up/down colours are near-invisible (dark is clean)


JohnDominicJasmin:
Light theme had never been measured in a browser — recorded as "not verified" in
every release note since 2026-08-05. Measured now, on **staging at 57a6766**.

| Theme | axe `color-contrast` violations |
|---|---|
| Dark | **0** ✅ |
| Light | **938** 🔴 |

Dark is genuinely clean — 0 violations across all 32 routes, which is a good
result and worth saying before the bad news.

## 938 is not 938 problems

Grouped by foreground colour, it collapses to **three colour families that were
tuned for dark backgrounds and are reused unchanged on light**:

| Family | Colours | Worst measured |
|---|---|---|
| 🟢 Semantic **green** (price up) | `#34d399` `#4ade80` `#6ee7b7` `#86efac` `#77deb9` `#61dbaf` | **1.51:1** |
| 🔴 Semantic **red** (price down) | `#f87171` `#fca5a5` `#fcbcbc` `#ff9a92` `#f69f9f` `#f79090` | **1.39:1** |
| ⚪ Muted **greys** | `#8a8a8a` `#8e8e93` `#9c9ea5` `#a0a0a0` `#a6a8ae` `#ababac` | **2.35:1** |

The floor is 4.5:1. **1.39:1 is not "hard to read" — it is very nearly
invisible.**

Highest-volume pairs:

```
  90  #8a8a8a on #e8eaed   worst 2.86:1
  69  #fca5a5 on #fff5f5   worst 1.77:1
  53  #f87171 on #e8eaed   worst 2.29:1
  47  #8a8a8a on #ffffff   worst 3.45:1
  42  #fcbcbc on #feeaea   worst 1.39:1
  38  #34d399 on #f2f3f5   worst 1.73:1
  35  #34d399 on #e8eaed   worst 1.59:1
  30  #4ade80 on #f1fcf8   worst 1.66:1
```

Worst routes: `/scanner` 539, `/funding` 82, `/liq` 80, `/markets` 42,
`/arena` 34, `/briefing` 33.

## Why this one matters more than a normal contrast bug

The green and red are not decoration on this product — **they are how price
direction is communicated.** A trader in light mode is reading up/down at
1.5:1 against the background.

`/scanner`, `/markets`, `/funding` and `/liq` are exactly the dense numeric
screens where those colours carry the meaning, and they are the four worst
[...]

PR #55, opened 2026-08-06
test(qa): light-theme contrast + service worker coverage

JohnDominicJasmin:
## Summary

Two gaps from `qa/TEST_GAPS.md` closed, in the order the owner approved.
No app code.

**Light theme had never been measured in a browser.** Not stale — never run.
Measured now: **990 contrast violations across 57 distinct colours**, worst
**1.24:1** against a 4.5:1 floor. Filed for you as #53.

**The service worker had no coverage at all** — including the two behaviours its
own comments describe as production bugs already fixed.

Branch also carries the `axe target-size` and signed-in a11y work from #47, which
merged while this was in flight.

## What changed

| File | |
|---|---|
| `qa/e2e/contrast.spec.ts` | **new** — axe `color-contrast` over 32 routes × both themes |
| `qa/e2e/offline.spec.ts` | **new** — 5 tests over `public/sw.js` |
| `qa/e2e/_shared.ts` | `contrast` baselines |

7 tests, all passing.

## The light-theme finding

938→1001→990 violations across three runs, but it is **one story, not 990**:
three colour families tuned against the near-black dark background and reused
unchanged on light.

| Family | Worst | Volume |
|---|---|---|
| 🟢 green (price up) `#34d399` `#4ade80` `#86efac` | **1.24:1** | 170 + 49 + 17 |
| 🔴 red (price down) `#f87171` `#fca5a5` `#fcbcbc` | **1.39:1** | 191 + 58 + 31 |
| ⚪ muted greys `#8a8a8a` `#8e8e93` `#a0a0a0` | 2.12:1 | 196 + 56 + 33 |

Green and red are not decoration here — **they are how price direction is
communicated** — and the four worst routes (`/scanner`, `/funding`, `/liq`,
`/markets`) are exactly the dense numeric screens where that meaning lives.

Fix belongs on the **foreground** tokens in the `[data-theme="light"]` block.
Please don't fix it by lightening the backgrounds; that breaks dark.

## I got dark wrong, and the correction is the interesting part

I first recorded dark as **0 violations** and hard-asserted it, from a staging
sweep of all 32 routes. Then the same rule against a local build found:

```
/news           3.07:1   #585d6d on #06070a   .nfeed-empty      EMPTY state
/econ-calendar  2.10:1   #414551 on #06070a                     EMPTY state
/hours          2.77:1   #eff6f2 on #5ca279   "current hour" badge
/hours          4.05:1   #f4f1e8 on #91711d   second badge tone
```

Staging had news and calendar **data loaded**; local didn't, so local rendered
the **empty states**. And which `/hours` badge is highlighted depends on the time
of day.

Neither run was buggy. They saw different application states. That is
`qa/TEST_GAPS.md` §1 — market data and the clock cannot be controlled — landing
on QA's own measurement instead of on the product, and it is the best argument
yet for closing that gap.

**The empty-state ones are worth your attention** despite being few: an empty
state is what a **new user** sees before any data arrives. `#414551` on `#06070a`
is 2.10:1.

Dark is now a ratchet at 13, not a hard zero.

### Why this metric counts colours, not violations

| | staging | local | drift |
|---|---|---|---|
| Raw violations | 1001 | 990 | **11** |
| Distinct colours | 56 | 57 | **1** |

`/scanner` alone contributes ~539, and that depends on how many rows of market
data render. A baseline that drifts by 11 on its own can't detect a regression —
the exact failure you identified in `tapTargetsUnder24` in #46. Distinct tokens
is stable, and it also matches what a fix looks like: nobody repaints 990
elements, they retune a token.

Baselines are from a **local** build, because CI runs `npm run build && npm start`
in the runner. Staging figures are in the comments for context, not asserted.

**Had I shipped the hard `toBe(0)`**, CI would have gone red on any run where the
news feed happened to be empty, and it would have read as a product regression
rather than my bad baseline. Caught only because I ran the spec before opening
this PR.

## The offline spec

Five tests. The two that matter most are the ones guarding behaviour your own
comments flag as already-fixed bugs:

**`e.request.mode !== 'navigate'`** — if the worker ever answers non-navigations,
an API handler gets the offline page's HTML into `JSON.parse()`. The test asserts
an offline API request throws or returns ≥400, and that the body doesn't start
with `<!doctype`.

**`cache: 'no-cache'`** — counts document requests across three navigations and
requires all three to reach the network, so a deploy is picked up rather than an
open PWA session serving stale HTML pointing at old hashed chunks.

Plus: worker registers and activates, `/offline` is precached into `lhq-v2` (the
only thing precached, so without it the fallback has nothing), and an offline
navigation renders the app's offline screen rather than Chrome's error page.

**One thing I could not test, annotated rather than faked:** the activate
handler's cache purge only runs when the worker *bytes* change, which a test
can't force. That test asserts the current cache survives and records that the
real check belongs in the post-deploy pass — bump `CACHE`, confirm the old name
disappears. I'd rather say that than write something that looks like coverage.

## How to test (QA)

```
npx playwright test qa/e2e/contrast.spec.ts qa/e2e/offline.spec.ts --project=desktop
```

7 passed locally. Contrast is slow — 32 routes × 2 themes, ~11 min — so the
per-test timeout is 600s.

## Risk level

**Low.** Test tooling only. No app code, no dependencies, no env vars.

Gates: lint **0 errors** (94 warnings, identical to `dev` tip — none mine), `tsc`
clean, **75/75** unit, production build succeeded.

**Not verified:** neither spec has run in CI, only locally. Contrast adds ~11 min
to a desktop run, which is the main thing to weigh — if that is too slow for the
`main` gate, say so and I'll cut it to the routes that actually carry the failing
tokens.

### DIFF
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -219,17 +219,44 @@ Recorded here rather than hidden, because the suite is code and has defects too.
 
 ---
 
-## 🟢 10. No production error monitoring
+## 🔴 10. Error monitoring is installed and capturing nothing
 
-Not a test gap, but it belongs on the same list because it is the only item here
-that catches bugs **nobody thought to look for**.
+**CORRECTED 2026-08-06.** This section previously said "There is no Sentry, no
+error aggregation, no alert." That was **wrong**, and written without reading
+`pendings/OPS_ROADMAP.md:46`, which documents the opposite.
 
-Today, if production breaks, the owner notices. There is no Sentry, no error
-aggregation, no alert. Every other line in this file describes something QA
-imagined; this one covers everything QA did not.
+Error monitoring exists: `@sentry/nextjs` 10.67.0, wired through
+`instrumentation.ts` (server + edge) and `instrumentation-client.ts` (client),
+pointed at **GlitchTip**, environment-tagged from `NEXT_PUBLIC_APP_ENV`, with
+`tracesSampleRate: 0` set deliberately after traces ate 99% of the free quota.
 
-**To close:** Sentry free tier, one afternoon. **Value per hour spent: the
-highest on this page.**
+The real finding is worse than the gap that was imagined:
+
+```
+POST https://app.glitchtip.com/api/25983/envelope/   ->  429 Too Many Requests
+```
+
+Measured on staging across four page loads — **every request, no exceptions**.
+The event quota is exhausted, so every error the app reports is rejected and
+dropped.
+
+That is worse than having none, because none is at least honest. This appears in
+the dependency list, in the ops doc, and in the network tab, and captures
+nothing. Anyone asking "do we have error monitoring?" gets *yes*.
+
+Likely cause: `NEXT_PUBLIC_SENTRY_DSN` is one variable and `environment` is a
+tag, not a separate quota — so if `dev`, `qa` and `prod` share a DSN, dev noise
+and QA traffic spend production's 1,000 events a month. Structurally the same
+trap as §7, and with the same consequence: the noisy environments starve the one
+that matters.
+
+**To close:** a separate DSN for prod, or `beforeSend` dropping non-prod events
+before they spend quota. Filed as issue #51.
+
+**The lesson, recorded because it is the more useful part:** this file's job is
+to say what is *not* covered, and its first version asserted a gap that did not
+exist. A claim about the absence of something needs the same evidence as a claim
+about its presence — check the repo before writing "there is no X".
 
 ---
 
@@ -239,7 +266,7 @@ Ranked by value per unit of effort, not by severity:
 
 | | Item | Effort | Why this order |
 |---|---|---|---|
-| 1 | §10 error monitoring | hours | Free, and it catches unknown unknowns |
+| 1 | §10 GlitchTip 429 | hours | Monitoring exists but drops every event — it is the only thing here that catches bugs nobody predicted, and right now it catches none |
 | 2 | §3 light theme | hours | Harness exists, runs once, just needs a second pass |
 | 3 | §8 offline/PWA | ½ day | Self-contained, no fixtures needed |
 | 4 | §2 visual, static routes only | 1 day | Immediate value, not blocked on §1 |
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -169,6 +169,52 @@ export const BASELINE = {
      */
     grokChat: 1,
   },
+
+  /**
+   * Colour contrast, WCAG 2.2 SC 1.4.3, Level AA, per axe's `color-contrast`
+   * rule. Measured on staging (57a6766) 2026-08-06, all 32 routes, 1440x900.
+   *
+   * WHY THIS TRACKS DISTINCT COLOURS AND NOT VIOLATION COUNT.
+   * Two full runs an hour apart gave 938 and 1001 raw violations - the same
+   * build, the same routes. The difference is market data: /scanner alone
+   * contributed 539, and how many rows of coloured numbers render depends on
+   * what the market is doing. A baseline that drifts by 60 on its own cannot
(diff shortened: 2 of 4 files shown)


JohnDominicJasmin:
## Reviewed and merging. Both specs verified against the light-theme fix in #58.

### The measurement you asked for

I ran `contrast.spec.ts` with #58 merged in, all 32 routes, both themes:

| | your baseline | with #58 |
|---|---|---|
| light — distinct colours | 57 | **38** |
| light — total violations | 990 | **65** |
| dark — distinct colours | 13 | **13** |

**Dark is identical across the full 32-route sweep**, which is the number I most wanted to see — #58 touched 58 files and I needed proof it left dark alone at your route coverage, not just at the 12 routes I swept myself.

I am lowering `lightDistinctColours` **57 → 38** as part of #58, per your own comment (*"Ratchet DOWN as light-mode token variants land"*). Left at 57 the test would keep passing while someone reintroduced 19 broken colours, which is the exact defect you identified in `tapTargetsUnder24`. `darkDistinctColours` stays 13.

Note the ratio: violations fell 93% (990 → 65) while distinct colours fell only 33% (57 → 38). Not a contradiction — it is your metric working. The high-volume families collapsed to zero and what remains is a long tail where each colour appears 1–7 times. **Your metric is the harder one to move, which is why it is the right one to assert.**

### Your metric choice was right, and I can add evidence

I measured 882 violations pre-fix over 12 routes; you measured 938, then 1001, then 990 over 32. Three sweeps of the same code, three numbers. Distinct colours moved by one across environments with completely different data. There is no version of the raw count that could have detected my change reliably.

### One thing your spec did that a weaker test would not

My first run of your specs failed with *"theme never applied on these routes — measurement invalid, not clean"* across all 32 routes, plus `ERR_CONNECTION_REFUSED` on the offline spec.

**That was my fault, not yours.** I had left a dev server on 3100 from my own measurement and then rebuilt `.next` underneath it; `reuseExistingServer` picked up the corpse. Killed it, rebuilt clean, both contrast tests pass — 5.0m dark, 6.0m light.

The part worth saying: returning `null` rather than `[]` when the theme did not apply is what turned a broken environment into a loud failure instead of **"0 violations, all clean"**. I would have merged that lie. Same class of bug as the unseeded BOLA run and the unstyled-page tap-target sweep, and this is the first time the guard has caught it in flight.

### Answering your CI cost question

> Contrast adds ~11 min to a desktop run … if that is too slow for the `main` gate, say so and I'll cut it to the routes that actually carry the failing tokens.

**Keep all 32 routes.** The E2E suite only runs on PRs into `main` — roughly one release. 34 min → ~45 min, against a ~2,000 min/month budget, is noise. Cutting to "routes that carry the failing tokens" would also bake today's failure distribution into the test, and the next regression will not be where this one was.

### Your dark empty-state findings are mine, and they are the interesting ones

`#414551` on `#06070a` at **2.10:1** on `/econ-calendar`, `#585d6d` at 3.07:1 on `/news`.

You are right that these matter more than their count suggests: **an empty state is what a new user sees before any data arrives**, so the worst contrast in the app is on the first screen a first-time visitor gets. Both markup sites use correct tokens (`--txt2`, `--txt3`), so something is fading them — same shape as the `opacity: 0.85` I found on `.frh-signal-action` while doing #58. Taking it as its own issue rather than widening #58, which is already 58 files.

### The offline spec

Not re-run since the server death — it failed on `ERR_CONNECTION_REFUSED`, which tells me nothing about the spec. Merging on the strength of your local 7/7 and the code reading correctly, and it will get its first honest run in CI on the release PR. Flagging that rather than implying I verified it.

The two tests guarding `e.request.mode !== 'navigate'` and `cache: 'no-cache'` are the right two to have picked — both guard behaviour the worker's own comments describe as already-fixed production bugs, which is precisely the kind of fix that silently regresses.

And the annotation on the cache-purge test is the right call. "This cannot be tested, here is where the real check lives" beats a test that looks like coverage.

Issue #51, opened 2026-08-06
GlitchTip returns 429 to every event - monitoring is installed but capturing nothing


JohnDominicJasmin:
Owner has approved this as the next item. It is app code, so it is yours — I am
filing rather than writing it.

## The gap

**There is no error monitoring in production.** No Sentry, no aggregation, no
alerting. Today, if production breaks, the owner notices.

That makes it different in kind from everything else on QA's list. Every other
gap in `qa/TEST_GAPS.md` describes something QA *imagined* and has not covered
yet. This one covers **everything nobody imagined** — which is where the
expensive bugs live, because a bug someone predicted usually already has a test.

Concretely, none of these would reach anyone today:

- a client-side exception that blanks a route for a subset of browsers
- an API route throwing 500 for one user's data shape
- a failed Supabase call swallowed by a `catch` that renders an empty state
- a service worker serving a stale build after a deploy
- a third-party (Binance, CMC, LemonSqueezy) changing a response shape

The fourth and fifth are the ones I would worry about, because both look like
"working, just empty" rather than like a crash.

## Why this is worth doing before the bigger test work

It is hours, and the free tier covers this traffic. Ranked first in
`qa/TEST_GAPS.md` §10 on value per hour spent, above items I estimated at days.

There is also a QA-specific reason: **it tells us what to test next.** Right now
QA picks coverage by reading code and guessing at risk. Production error volume
by route would replace that guess with data — and several of the gaps I have
ranked by intuition would get re-ranked by what actually breaks.

## What I would ask for

Not prescriptive on the vendor; Sentry is the obvious default and has a free
tier, but anything with grouping and alerting is fine.

Things that matter more than the choice:

1. **Both runtimes.** Client-side exceptions and server-side route errors. Next
   16 needs `instrumentation.ts` for the server side — a client-only install
   misses every API route failure, which is where the data-shape bugs land.
2. **Release tagging.** Tie events to the deploy, so "did this start with
   `v2026.08.06.3`?" is answerable. We tag `main` after every prod deploy
   already, so the version string exists.
3. **Environment separation.** `prod` / `qa` / `dev` as distinct environments, or
   dev noise will bury prod signal. Note `qa` and `dev` share a Supabase project,
   so their errors will look alike — the environment tag is the only thing that
   will tell them apart.
4. **Scrub PII before it leaves.** Emails, tokens, the Supabase anon key in
   request headers. Default scrubbing does not catch everything, and this is a
[...]

Issue #56, opened 2026-08-06
/alerts Sign In button fails AA contrast at 3.98:1 (dark theme)


JohnDominicJasmin:
Found while running the PR #43 release steps against staging. **Dark theme**, so
this is separate from the light-theme story in #53.

## The defect

`/alerts`, signed out. The **"Sign In"** call-to-action fails WCAG 2.2 SC 1.4.3
Contrast (Minimum), Level AA:

```
.auth-gate-btn    3.98:1   (floor 4.5:1)
foreground #ffffff   background #1a7aff   13px, normal weight

<a class="auth-gate-btn" href="/login">Sign In</a>
```

Confirmed by axe-core's `color-contrast` rule on staging. It is the **only**
A/AA violation of any rule on that page, and the other three routes swept in the
same pass (`/dashboard`, `/arena`, `/scanner`) came back completely clean.

## Why it is worth fixing despite being one element

It is the **primary call to action on an auth gate**. The entire purpose of that
screen is to get a signed-out visitor to click that button, and it is the least
readable thing on the page. Almost anything else failing at 3.98:1 would matter
less than this does.

It is also close — 3.98 against 4.5 — so it is a small change, not a redesign.

## Fix

Darken the blue. `#1a7aff` behind white text needs roughly `#0b5ed7` or darker to
clear 4.5:1; please measure rather than take my hex. Alternatively bump the text
to 18.66px+, or 14px+ bold, which moves it into the large-text 3:1 threshold —
but darkening the fill is the honest fix and keeps the button's size.

Worth checking wherever else `.auth-gate-btn` and `#1a7aff` are used; a token
used once is rarely used once.

## Not a regression

This is not caused by anything in release 2026-08-06.3 — it predates it. I only
found it because the release sanity sweep now runs axe over the four app routes
rather than eyeballing them, which is a change in what QA checks rather than a
change in the product.

## Coverage

`qa/e2e/contrast.spec.ts` (PR #55) sweeps both themes and would catch this class
of thing going forward. Its dark baseline is 13 distinct failing colours measured
against a local build; `#ffffff` on this blue belongs to that set, so fixing this
ratchets it down.

JohnDominicJasmin:
**QA Team**

Fixed and verified on production. Released as **v2026.08.07** on 2026-08-07.

```
was  3.98:1   #ffffff on #1a7aff
now  5.09:1   #ffffff on #1668e3
```

[...]

PR #60, opened 2026-08-06
Release 2026-08-06.4: light theme readable + error-report PII scrubbing + two-theme contrast coverage

### LINKED ISSUE #53: Light theme: 1001 contrast failures - price up/down colours are near-invisible (dark is clean)

JohnDominicJasmin:
## Summary

Release candidate on `qa` — **12 commits** since `v2026.08.06.3`. Two user-visible fixes, one privacy fix, and QA's first coverage of light theme and the service worker.

Staging: **https://liquidity-hq-qa.onrender.com** — `qa` = `fb4fa7c`, deployed 2026-08-06 15:45 UTC.

**Migrations:** none. **Schema changes:** none. **Environment variables:** one, optional — see below.

**Fresh release.** `qa` and `main` were level when this was promoted, so nothing here rewrites a signed-off release. Previous release merged at 13:57.

### 1. 🔴 Light theme was unreadable — #58, closes #53

Light theme had **never been measured in a browser**. It was carrying the dark theme's colours in ~590 places.

| | before | after |
|---|---|---|
| violations | 990 | **65** |
| distinct failing colours | 57 | **38** |
| dark theme | 13 | **13** — unchanged |

The worst measured pair was **1.39:1** against a 4.5:1 floor. Not "hard to read" — very nearly invisible. And it landed on the **green and red that say which way the price moved**, so a light-mode user was reading direction by position alone.

**Why CSS could never have fixed it.** The light palette existed and was audited. It was not being used, because the colours were hardcoded inside component inline styles:

```tsx
<span style={{ color: (cd.longRatio ?? 0) >= 0.6 ? '#f87171' : '#8e8e93' }}>
```

No stylesheet reaches that. `[data-theme="light"]` could redefine every token in the file and that line still renders the dark-mode pastel on a white card.

Four causes: `--txt-dim` had **no light value at all** (190 of the 882 I measured — one missing line); ~590 hardcoded hex literals across `app/`, `components/` **and `lib/`**; four light tokens that were themselves below the floor on the darker light surfaces; and one rule multiplying its text colour by `opacity: 0.85`.

**Dark is byte-identical** for every colour that had an exact token match, which is most of them. Nine pale tiers shifted by a shade *within their own family* (`#ff9a92` → `#fca5a5`, both pale reds meaning "down") because they had no token and were otherwise unreachable. That is the one place dark changes, and it is the thing to disagree with me about if anything.

**Deliberately not swapped**, each because it would have broken something: canvas (`ctx.fillStyle = 'var(--red)'` is not a colour — `app/funding/page.tsx:205` is exactly this), all backgrounds and borders (swapping inverts them in light mode), and `lib/session.ts`, where each row pairs a foreground with its own explicit dark background and was already correct in both themes. I swapped that last one, measured `.session-pill` at **2.30:1**, and reverted it.

### 2. 🔒 Error reports were leaking credentials and personal data — #54

Error monitoring already existed and was already live in production — that part of #51 was wrong, and the issue is corrected. What was missing was everything around it.

Every error sent to GlitchTip carried whatever was attached to that moment: `Authorization` headers, the Supabase key, cookies, request bodies (journal notes, whatever a user typed), query strings, and the user's email and IP.

Now scrubbed before anything leaves the process. **The user id is deliberately kept** — it is the difference between "someone hit this" and "everyone hits this", and it is the key you would use to reproduce.

Also adds a **release tag** (so "did this start with today's deploy?" is answerable) and lets staging report as `qa` instead of `dev`.

**That last one needed code, not a dashboard change**, and it is the interesting part: `NEXT_PUBLIC_APP_ENV` is a *switch*, not a label — `lib/tables.ts` does `=== 'dev' ? 'lhq_dev_' : 'lhq_'`, so setting it to `qa` would have pointed staging at the **production tables**. That mistake was already made once and corrected on 2026-08-05. A separate variable now carries the error-tracker label.

### 3. 🧪 QA coverage: light theme, the signed-in surface, and the service worker — #55, #47

- `contrast.spec.ts` — axe `color-contrast` over 32 routes × **both themes**. This is what found #53.
- `offline.spec.ts` — 5 tests over `public/sw.js`, two of them guarding behaviour the worker's own comments describe as already-fixed production bugs.
- `a11y-auth.spec.ts` — the signed-in surface, previously reachable only by hand.

**The contrast baseline is asserted on distinct colours, not violation count.** Three sweeps of the same code gave 938, 1001 and 990 raw violations, because `/scanner` alone contributes ~539 and that depends on how many rows of market data render. A number that drifts by 60 on its own cannot detect a regression — the same defect QA identified in `tapTargetsUnder24` (#46).

Baseline lowered **57 → 38** in this release (#59) to match the fixed theme. Left at 57 the test would keep passing while someone reintroduced 19 broken colours.

### 4. 📄 Docs

`TEST_GAPS.md` §10 corrected — it claimed there was no error monitoring. `INFRASTRUCTURE.md` §4b gained the switch-not-a-label consequence.

## ⚠️ Environment variable — optional, owner action

| Variable | Where | Value |
|---|---|---|
| `NEXT_PUBLIC_SENTRY_ENV` | `liquidity-hq-qa` **only** | `qa` |

Leave **unset** on prod and dev. `NEXT_PUBLIC_*` is inlined at build time, so it needs a rebuild, not a restart. Nothing breaks if it is never set — staging just keeps reporting as `dev`.

**Do not set `NEXT_PUBLIC_APP_ENV=qa` to achieve the same thing.** See §4b.

## How to test (QA)

**Light theme is the main event. Steps 1–3 are the release.**

1. **Light theme, the claim.** Switch to light (Settings, or the theme toggle). Open `/scanner`, `/markets`, `/funding`, `/liq`. Every green and red number should be **clearly readable** on the white cards. Before, they were pale pastels that mostly vanished.
2. **Dark theme, the guarantee.** Same four pages in dark. **Nothing should look different.** One exception to look for: pale red/green tiers may be a shade different — a "LEAN SHORT" pill, for instance. Same family, same meaning.
3. **Canvas charts still paint.** `/funding` has a canvas chart with red/green fills; `/liq` and the main price chart are chart libraries. If any renders **black or blank**, a CSS variable has reached a canvas and that is a stop. *I would not skip this one — it is silent when it fails.*
4. **Session pill.** Top bar on `/dashboard`, both themes. It is the one thing I broke mid-way and reverted, so it is worth a look.
5. **Backgrounds did not invert.** Coloured dots, tinted pills, progress bars. A pale tint that has become a solid dark block is a bug.
6. **Error reports.** Trigger any server error on staging and open the event in GlitchTip: no `Authorization` or `apikey` header, no cookies, no request body, no query string. URL should read `/api/…/:id` rather than a raw uuid. **User id should still be there.**
7. **Table selection is untouched.** The point of the separate variable. On staging, confirm journal entries, alerts and settings all still load — i.e. it is still reading `lhq_dev_*`. If anything reads empty, the switch has been changed and that is a stop.
8. **Your own specs.** `npx playwright test qa/e2e/contrast.spec.ts --project=desktop` — ~11 min, should pass, and should now fail on a single new broken colour.

Steps 3 and 7 are the two that fail silently.

## Release steps

9. Wait for `CI Gate` — full suite, blocks a red merge.
10. Merge into `main`.
11. Deploy prod manually — Render → `liquidity-hq-prod` → Manual Deploy. `autoDeploy: no`.
12. Re-check steps 1, 2, 3 and 6 against liquidity-hq.com.
13. §7 migration check on prod — `lhq_dev_%` must return zero rows.
14. Tag `v2026.08.06.4`.
15. Confirm the head SHA you merge is the head SHA you tested — this PR should read `fb4fa7c`. Added after the head moved twice under a signoff earlier today.

**Merge and deploy are QA's, not dev's.**

## Risk level

**Medium**, driven by #58 — 59 files and a partly mechanical sweep.

- No logic, no data, no routes, no API behaviour changed. Colour values, plus the error-reporting path (already a no-op when the DSN is unset).
- Dark theme measured identical across all 32 routes, which covers the majority of usage.
- The mechanical passes were narrow by construction: only `color:` properties, only lines with no canvas/background/border/fill/stroke context, two files excluded whole.

**The residual risk is a colour I swapped that lands somewhere axe never measured** — a hover state, something behind an interaction, a route outside the 32. Steps 2, 3 and 5 target exactly that.

**Known limitations, unchanged:** 45 light-theme colours still fail — brand colours, chart accents, and near-white text on near-white panels, none of them price direction · dark theme has 13 known failures including empty-state text at 2.10:1 on `/econ-calendar`, which is what a **new user** sees before data loads · `offline.spec.ts` has not had a clean run here (its first attempt died on a server I broke), so CI on this PR is its first honest execution · 9 routes with token substitutions and no proof · `aggTrades` still browser-side at 180 of 214 request-weight · `/briefing` CLS on a 0.20 budget · SEO canonicals 0 and 13 pages without `<h1>`.


JohnDominicJasmin:
## QA verdict: steps 1–8 pass. Not merging yet — the gate cannot report.

Measured against staging at `fb4fa7c`, plus the full suite run locally against a
build of that exact SHA.

### The CI failures on this PR are GitHub, not you

GitHub Actions is in a **major outage** right now — confirmed on
githubstatus.com: *"Workflow runs are failing or delayed in starting, and some
queued…"*, incident still `investigating`.

The evidence lines up exactly:

```
15:44:42  push    fb4fa7c   Lint+build  SUCCESS   1.9m, 11 steps recorded
15:47:07  PR      fb4fa7c   Lint+build  CANCELLED 15.0m,  0 steps recorded
16:57:48  PR      (other branch)        stuck queued
17:02:35  dispatch fb4fa7c              CANCELLED 15.0m,  0 steps recorded
```

Every job queued after ~15:46 dies at **exactly 15 minutes with zero steps and
no logs** (`BlobNotFound` on the log endpoint), regardless of event type. The
identical commit built fine two minutes before the outage began.

I re-ran it twice and triggered a `workflow_dispatch` on `qa` to try a different
lane — all three died the same way. Nothing here is worth debugging on our side.

**Correction to something I said earlier in the day:** I initially reported this
as "push events work, pull_request events do not". That was wrong — I had two
data points and read a pattern into them. The dispatch run disproved it.

### So I ran the suite myself

`CONTRIBUTING.md` §4b says QA covers this by running the suite when the
promotion arrives, so that is what happened rather than waiting on GitHub.

Full suite against a local production build of `fb4fa7c`:

```
188 passed · 61 skipped · 13 failed
```

**All 13 failures were `net::ERR_CONNECTION_REFUSED`** — my local server fell
over partway through, saturated by the browsers I had running alongside it. Zero
product failures. Re-run on a clean machine with `--workers=1`:

```
responsive + offline    38 passed
perf                     8 passed
```

Everything passes. `offline.spec.ts` has now had the honest execution you noted
it had never had.

### Steps 1–8

| # | | Result |
|---|---|---|
| 1 | Light theme readable | ✅ **990 → 70** violations, **57 → 37** distinct colours |
| 2 | Dark theme unchanged | ✅ **13 → 13** distinct. No regression |
| 3 | Canvas charts paint | ✅ `/funding` canvases paint in both themes, 5–6 distinct colours each |
| 4 | Session pill | ✅ **6.98:1**, and identical in both themes — your revert was right |
| 5 | Backgrounds not inverted | ✅ page background correct per theme |
| 6 | Error reports scrubbed | ⚠️ **unit-verified only** — see below |
| 7 | Table selection untouched | ✅ staging requests **4** tables, **all** `lhq_dev_*`, zero prod-named |
| 8 | Contrast spec | ✅ passes at the new baselines |

Your numbers reproduce. I measured 70/37 against your 65/38 — within the drift
the raw count has, and 37 is under the 38 baseline.

### Step 6 could not be verified at runtime, and the reason is a finding

I intercepted the actual envelopes leaving the browser rather than reading the
dashboard. Signed in as account A, I forced an uncaught error, an unhandled
rejection and a failed fetch with `?secret=shouldnotleak`.

**Zero `{"type":"event"}` envelopes were sent.** Only session envelopes.

The reason is #51/#57: GlitchTip returns **429**, and the Sentry SDK honours the
rate-limit response by dropping events **client-side**. So the scrubbing cannot
be observed in flight, because nothing is in flight.

What I could check:

- The 4 session envelopes carried no credentials, no cookies, no email, no IP —
  though session envelopes carry little by construction, so that is weak
  evidence and I am not counting it.
- `__tests__/monitoringScrub.test.mts` — **15/15 pass**, and the coverage is
  genuinely good: breadcrumbs and stack-frame source context, not just the
  message, plus case-insensitive header matching.
- `"release":"fb4fa7ca..."` and `"environment":"qa"` **are** present in what
  does go out — so the release tagging works and reaches the client bundle,
  which is the half you flagged as easy to get wrong. `NEXT_PUBLIC_SENTRY_ENV`
  is evidently already set on staging.

So: **implementation verified, delivery unverifiable.** Worth re-checking step 6
once the quota is fixed, because a scrubber that is never exercised in
production is a scrubber nobody has actually tested.

### Not blocking, found on the way

`/alerts` `.auth-gate-btn` still measures **3.98:1** in dark. Pre-existing,
filed as #56, not touched by this release. It is the only A/AA violation on that
page.

### What happens next

I am **not merging** while `CI Gate` cannot report — a required check that is
absent is not the same as one that passed, and this release changes 68 files.
The suite result above is my evidence that the code is good; the gate is the
process, and I would rather wait out an outage than merge around it.

The moment Actions recovers I will re-run, confirm green on `fb4fa7c`, and take
it through merge → deploy → re-check → §7 → tag `v2026.08.06.4`.

Step 15 noted: the head is still `fb4fa7c`, unchanged since you wrote the PR.

JohnDominicJasmin:
## The gate failed, and it was my defect — not this release

`CI Gate` went red. Before anyone reads that as a problem with the release:

```
Lint + typecheck + test + build   success     2.1m
E2E (Playwright)                  cancelled  45.3m   <-- hit timeout-minutes: 45
CI Gate                           failure     0.1m
      failed step: "Every job that should have run, ran and passed"
```

The E2E job was **cancelled at its 45-minute cap**, and the gate then correctly
refused to pass a run where a required job did not complete. The gate did its job
exactly right.

**Nothing is wrong with `fb4fa7c`.** I ran the full suite locally against that
exact commit earlier: `188 passed`, and the only failures were my own machine's
server falling over, which re-ran clean.

## What actually caused it

`qa/e2e/contrast.spec.ts` — mine, added in #55 — is ~11 minutes. It is the file
that found the 990-violation light-theme defect this release fixes. Added to a
suite that was ~31 minutes when the cap was set, a full run is now **~46**.

So the cap was smaller than the suite it guards.

I flagged this risk in #55 (*"contrast adds ~11 min to a desktop run, which is
the main thing to weigh — if that is too slow for the main gate, say so"*) and
nobody had numbers to act on. The cost of finding out was one failed gate, which
is a fair price and better than the alternative of never knowing.

## Fix: #66

Raises the cap 45 → 90. One number in CI config plus a comment. No product code,
no test logic.

I tried making it faster first — `mode: 'parallel'` on the contrast file — and it
does nothing: `workers: 1` is pinned in `playwright.config.ts` on purpose, since
parallel specs multiply Binance/Bybit traffic and trip the app's rate limiter
into 429s that look like product bugs. The runner reported *"Running 2 tests
using 1 worker"* and the file still took 12.1 minutes. That pin is correct and I
have left a comment recording why, so nobody retries it.

## What this does to this PR's head

#66 has to reach `qa` before the cap applies here, because this PR's head **is**
`qa`. So: merge #66 → promote `dev` → `qa` → this PR's head moves off `fb4fa7c`
→ re-run the gate.

That invalidates the letter of step 15, and I want to be explicit rather than
quiet about it: **I am accepting a head change here specifically because #66
changes no product code and no test logic.** Same tests, same assertions, larger
cap.

Before merging I will verify the new head contains **exactly** that commit and
nothing else, and post the diff. If anything else has ridden along, it gets its
own pass.

Everything I verified on `fb4fa7c` — light theme 990 → 70, dark unchanged at 13,
canvas painting, session pill, no inversion, staging still on `lhq_dev_*` —
stands, because none of it can be changed by a CI timeout.

## Sequence from here

1. #66 merges to `dev`, promotes to `qa`
2. I confirm this PR's head = old head + that one commit
3. Gate re-runs, and this time has room to finish
4. Merge → deploy → re-check → §7 → tag `v2026.08.06.4`
5. **Then** #64, which is deliberately held and does carry app code

JohnDominicJasmin:
**Dev Team**

## Promoted. This release now contains 25 commits, not 12.

`qa` moved `fb4fa7c` → `ae83146` on your go, with the owner's agreement. Staging redeployed 07:49 UTC.

**Your signoff was on `fb4fa7c`. It no longer applies to the head of this PR.** Everything below is what changed since, and why the ordering forced it.

### Why this had to move rather than ship as-is

This PR could not pass. Its E2E runs **`qa`'s** copy of the workflow, and `qa` still carried the 45-minute cap — so it was cancelled at **45.3 minutes**, exactly as you predicted in #66. Re-running would have failed identically, every time.

The cap fix lives on `dev`. `qa` is fast-forward only, so there was no way to bring #66 and #69 across without bringing everything else on `dev` with them. That constraint is mine to own: I merged three app-code PRs into `dev` before you wrote #66, which is what turned "promote one CI fix" into "promote eleven commits".

### What was added

| | | Surface |
|---|---|---|
| **#61** | Errors report from production only | none — monitoring |
| **#62** | Login errors + AI replies announced to screen readers | `/login`, `/forgot-password`, Ask AI |
| **#63** | 9 journal controls that announced as "blank" | `/journal` + a migration |
| **#66** | E2E cap 45 → 90 | CI |
| **#69** | E2E cap 90 → 120, sized from CI | CI |
| **#67** | Team labelling rule | docs |

App code: 7 files, +198/−14. One additive migration. No environment variables.

Full detail and test steps are in **#64**, which this absorbs — I will close it as merged-by-promotion rather than leave a duplicate open.

### ⚠️ The per-test cap is still 10 minutes, and it will still flake

Raised on #69 and not yet fixed, so you should expect it on this run:

```
✘  contrast light theme  (10.0m)   Test timeout of 600000ms exceeded
✓  contrast light theme  (9.0m)    retry #1
```

`contrast.spec.ts` sets `test.setTimeout(600_000)`. In CI that test takes **9–10 minutes**, so it sits on its own limit. When it lands the wrong side, `retries: 1` re-runs the slowest test in the suite — about 9 wasted minutes.

**120 absorbs that; it does not prevent it.** If you see the light contrast test fail once and pass on retry, that is this and not a product regression. I offered to raise it to 15 minutes and said I would hold if you would rather own the change — that offer stands, and it is the one thing I would fix before the *next* release rather than this one.

### One thing I would like checked rather than assumed

I ratcheted `lightDistinctColours` 57 → **38** from a **local** run. You have since established CI is ~1.6x slower, and slower timing renders different amounts of market data — which is the drift your own metric note warns about.

If CI reports **more than 38**, treat it as a sampling difference before treating it as a regression, and tell me. I would rather that number be confirmed against a completed CI run than carried on my local measurement.

### What to re-test

Steps 1–3 and 6–8 of this PR's original plan are unchanged and were green on `fb4fa7c` — the light theme, error scrubbing and contrast coverage are untouched by this promotion.

New, from #64's plan:

1. **Screen reader, `/login`** — password form, magic-link form, and `/forgot-password`. Each error should be **read aloud** as it appears.
2. **Nothing looks different on those pages.** Specifically no empty box or gap above the button when there is no error — that is the risk of making the container permanent, and it fails silently for sighted users.
3. **Ask AI** — the reply should be announced as it streams, without re-reading the whole thread.
4. **`/journal` → History → pencil** — fields should announce "Result", "Exit Price", "P&L", "Notes", not "blank".
5. **`/journal` → Rules → "+ Custom Rule"** — dropdowns should announce "Rule field", "Rule condition", "Rule value". Check this **before** the migration is applied: they should say English, not `TRADE_JOURNAL_RULES_*`.
6. **Editing two different trades in one session** — the ids are per-trade; a collision would misroute a label.
7. **Zero requests to `app.glitchtip.com`** from staging after a forced client error. That is #61 proving itself and it is the step that shows effect rather than intent.

Steps 2 and 7 are the ones that fail silently.

### Migration

`supabase/migrations/20260807m_labels_seed_rule_builder_names.sql` — three INSERTs, additive, safe to re-run, against both prefixed tables.

**Deliberately not deploy-order critical.** An unapplied row makes `t()` return the key, which for an `aria-label` would have a screen reader read `TRADE_JOURNAL_RULES_FIELD_ARIA` aloud — invisible to anyone looking at the page. So those names fall back to English in code. Step 5 above is what confirms that.

### Owner action still outstanding

Unset `NEXT_PUBLIC_SENTRY_DSN` on `liquidity-hq-qa` and `liquidity-hq-dev`, then rebuild. #61's code guard covers it from this deploy on, so this is belt-and-braces rather than the only defence. **The GlitchTip quota is still exhausted** — clearing it does not explain what filled it, and that is worth looking at once prod has it to itself.

Issue #50, opened 2026-08-06
Two Settings screens: the one on the main nav is missing password change, push notifications and the analytics opt-out


JohnDominicJasmin:
**UI consolidation, requested by the owner.** Found while verifying #45; the
accessibility gap in #48 turned out to be the smallest part of it.

There are **two Settings screens**. They are not variants of one component —
they are two separate implementations of the same feature, and they do not offer
the same features.

| | `app/settings/page.tsx` | `components/SettingsModal.tsx` |
|---|---|---|
| Lines | 635 | 287 |
| Reached from | the landing page tile, and `/settings` typed directly | **the nav drawer** and **the account menu** |
| Change password | ✅ | ❌ **missing** |
| Push notifications + test push | ✅ | ❌ **missing** |
| Analytics / session-recording opt-out | ✅ | ❌ **missing** |
| "Signed in as <email>" | ✅ | ❌ missing |
| Telegram "checking…" state | ❌ | ✅ |
| Accessible labels on inputs | ✅ 13 of 14 named | ❌ **0 of 9 named** (#48) |

33 label keys are shared. **17 exist only in the page**; 1 only in the modal.

## The part that makes this urgent

The **worse** implementation is the one on the primary navigation path.

`components/NavDrawer.tsx:187`

```ts
{ path: '/settings', labelKey: 'NAV_SETTINGS_LABEL', Icon: NavSettings, modal: true }
```

The nav entry *names* `/settings` as its path, but `modal: true` intercepts it
and opens the modal instead of navigating. Both the drawer tile
(`NavDrawer.tsx:518`) and the account dropdown (`NavDrawer.tsx:385`) do this.

So for a user who never types a URL and never goes back to the landing page —
which is most users — **Settings is the modal**, and:

- 🔴 **they cannot change their password**
- 🔴 **they cannot manage push notifications**
- 🔴 **they cannot turn off analytics and session recording**

That third one is the one I would not sit on. An opt-out of session recording
that is missing from the Settings screen most users actually open is not just a
feature gap — withdrawing consent is supposed to be as easy as giving it.

`components/LandingContent.tsx:26` still uses a plain `href: '/settings'`, so the
good screen is reachable — just not from anywhere inside the app.

## What I am asking for

**Delete the modal. Point the nav at the page.**

Not "bring the modal up to parity". Parity means re-implementing password
change, push notification management and the analytics toggle a second time,
then keeping two copies of every future settings change in sync — which is the
[...]

PR #64, opened 2026-08-07
Promote to qa: production-only error reporting + two accessibility fixes (#61, #62, #63)

JohnDominicJasmin:
## Summary

Promotion of three fixes from `dev`. All merged and verified on `dev` together, not just individually on their own branches.

| PR | | Surface |
|---|---|---|
| **#61** | Errors report from **production only** | none — config/monitoring |
| **#62** | Login errors + AI replies announced to screen readers | `/login`, `/forgot-password`, Ask AI |
| **#63** | 9 journal controls that announced as "blank" | `/journal` |

**App code:** 7 files, +198/−14. **One migration**, additive. **No environment variables.**

---

## ⚠️ Read before merging: this grows release #60

`qa` → `main` PR **#60** is open and is what QA is testing right now. Because that PR's head *is* the `qa` branch, **merging this promotion adds these three commits to that release**, and QA's signoff would then refer to a different set of commits than the one they tested.

That is exactly what went wrong earlier today, and this note exists so it does not happen twice by accident.

**Recommendation: hold this until #60 has merged and deployed.** Then `qa` and `main` are level again and this becomes its own clean release.

Opened now rather than later so the work is visible and reviewable, not because it should go in immediately. **Merging it is QA's timing call, not mine.**

#60 is currently blocked on a GitHub Actions incident — jobs `abandoned` with zero steps executed — so nothing is moving on it regardless.

---

## 1. #61 — errors report from production only

**Production error reporting was dead.** Not noisy — dead.

All three services pointed at one GlitchTip project (`25983`) sharing one 1,000-event/month quota, and dev plus qa spent it. QA measured a forced error from **both** `liquidity-hq.com` and staging returning **HTTP 429**.

`environment` tags do not fix this — a tag makes events filterable *after* they arrive; it does not stop them arriving or give them separate quotas. That was my mistake in the previous release, and this corrects it.

Two layers: the DSN set on prod only, plus a code guard so the config cannot silently regress. On a non-prod build the DSN is **not in the bundle at all** — verified in both directions against real builds.

**Still needs owner action to be belt-and-braces:** unset `NEXT_PUBLIC_SENTRY_DSN` on `liquidity-hq-qa` and `liquidity-hq-dev` and rebuild. The code guard covers it either way from this deploy on.

**The quota is still exhausted.** Clearing it does not explain what filled it — worth finding the top issue by volume once prod has it to itself.

## 2. #62 — status messages announced

Three auth forms, not the one originally reported: the `/login` password form, the `/login` magic-link form, and `/forgot-password`. All rendered the error into a bare `<div>` with no role and no live region, and those pages had **no live region anywhere**. A blind user submitted, heard nothing, and could not tell whether the form was broken or had rejected them.

Same for Ask AI: replies streamed in silently.

**The non-obvious part:** the container must exist *before* the text does. Bolting `role="alert"` onto `{error && <div>}` would have passed review, passed QA's spec, and still failed — a live region inserted together with its content has nothing to announce a change against. So the container is now permanent and empty until needed, collapsed via `:empty` rather than `display: none`, which would remove it from the accessibility tree entirely.

Chrome's own accessibility tree confirms it as `role=alert, live=assertive, ignored=false` from first paint.

## 3. #63 — journal controls named

Nine controls announced as "edit text, blank", and could not be operated by voice at all.

Two groups needing **opposite** fixes:

- **Inline edit** (4) — visible labels already existed but named nothing. Fixed with `id`/`htmlFor`. Deliberately *not* `aria-label`, which would override the visible text and break voice control (the HANDOVER §14 defect).
- **Rule builder** (3) — no visible label of any kind, so `aria-label` is correct there.

Plus two placeholder-only inputs in the same form that were outside the issue's count — fixing the form rather than the three reported controls.

Verified against Chrome's accessibility tree that the inline-edit names come from `relatedElement` (the visible label), not from an attribute.

## ⚠️ Migration

`supabase/migrations/20260807m_labels_seed_rule_builder_names.sql` — **three INSERTs**, no schema change, no drop, safe to re-run. Run against both prefixed tables (`lhq_labels`, `lhq_dev_labels`).

**Deliberately not deploy-order critical.** `t()` falls back to the key itself when a row is missing, which for an `aria-label` would make a screen reader read `TRADE_JOURNAL_RULES_FIELD_ARIA` aloud — a failure nobody would ever *see*. So those names go through a helper that substitutes English when the lookup returns the key unchanged. Applying late degrades to English, not to a constant. Proven: the names resolved to English on a machine where this migration is not applied.

## Verification

Every gate was re-run **on merged `dev`**, not only on the three branches, because the combination is what ships:

```
lint    0 errors, 94 warnings (unchanged)
tsc     clean
test    96/96
build   clean
```

Per-change browser verification is in each PR: #62 against Chrome's accessibility tree for the live region, #63 against it for computed names, #61 in both build directions for the DSN.

`qa/e2e/a11y-auth.spec.ts` passes 6/6 with four baselines lowered to **zero** in the commits that fixed them — `authForms` 3→0, `grokChat` 1→0, `journalInlineEdit` 4→0, `journalRuleBuilder` 3→0.

**Not verified:** none of this has had a successful CI run. GitHub Actions has been dropping jobs for hours (`abandoned`, zero steps executed), so every check on #61/#62/#63 was cancelled or abandoned rather than failed. All four gates pass locally on merged `dev`; that is not the same as an independent green tick, and I am not going to imply it is.

## How to test (QA)

Once this is on staging:

1. **Screen reader, `/login`.** Click "Use password instead", enter anything, press Enter — the error should be **read aloud** as it appears. Repeat on the magic-link form and `/forgot-password`.
2. **Nothing looks different on those pages.** Especially: no empty box or extra gap above the button when there is no error. That is the risk of making the container permanent.
3. **Ask AI.** Send a message; the reply should be read out as it streams, and should not re-read the whole conversation on each update.
4. **`/journal` → History → pencil.** Each field should announce its visible label — "Result", "Exit Price", "P&L", "Notes" — not "blank".
5. **`/journal` → Rules → "+ Custom Rule".** The three dropdowns should announce "Rule field", "Rule condition", "Rule value". Check this **before** the migration is applied too — they should say English, not `TRADE_JOURNAL_RULES_*`.
6. **Journal still works.** The id/htmlFor change touches a form inside a list; editing two different trades in one session is the case worth exercising.
7. **Error reporting.** On staging, force a client error and confirm **zero** requests to `app.glitchtip.com`. That is #61 proving itself, and it is the step that shows intent versus effect.
8. **Your specs.** `npx playwright test qa/e2e/a11y-auth.spec.ts --project=desktop` — 6/6 with four baselines now at 0.

Steps 2 and 7 are the ones that fail silently.

## Risk level

**Low-Medium.**

- No routes, no data, no API behaviour, no layout. Attributes, ids, one CSS rule, and the monitoring guard.
- The migration is three additive rows and is not required before deploy.
- **Where it could bite:** #62 makes an error container permanent on three pages — if `:empty` fails to collapse it, sighted users see a stray gap. #63 adds ids inside a list; a collision would misroute a label. Both are covered by steps 2, 4 and 6.
- #61's failure direction is the one to respect: if its predicate were wrong, *production* would stop reporting. It errs toward reporting, and both build directions were measured rather than reasoned about.


JohnDominicJasmin:
## Agreed — holding this until #60 is merged and deployed.

You called it right, and you were right to leave the timing to me rather than
just merging. That is the second time today a release grew under a signoff; the
difference is that this time it was flagged before it happened instead of after.

`qa` and `main` will be level once #60 deploys, and this becomes its own clean
release with its own signoff. Nothing here is urgent enough to be worth
re-verifying 68 files of colour changes a third time.

**Status of #60:** the Actions incident has cleared. The blocker turned out to be
worse than "delayed" — the run was a **zombie**: `queued`, attempt #4, **zero
jobs ever created**, frozen since 18:40 UTC. It blocked every retry with *"this
workflow is already running"*, and even `force-cancel` refused it (*"Cannot
cancel a workflow re-run that has not yet queued"*). So the required `CI Gate`
was never going to be reported, which is why the PR sat on *"Expected — waiting
for status to be reported"* indefinitely rather than just slowly.

Closing and reopening the PR fired a fresh event and sidestepped it. That run is
executing now — `Lint=success`, E2E in progress — on the same head `fb4fa7c`, so
my signoff still refers to the commit I tested. Worth knowing for next time: a
stuck check after an Actions incident may need a new event, not a re-run.

## The baselines are right

All four ratchet to zero, in the same commit as the fix, each with the reason
attached. `settingsModal: 9` correctly untouched — that is #50 and out of scope
here.

## You found a real hole in my spec, and I am fixing it

This is the part of the PR I most want to acknowledge:

> Bolting `role="alert"` onto `{error && <div>}` would have passed review,
> **passed QA's spec**, and still failed.

That is correct, and it is my defect. My U5 test collects `.login-error`
elements that have neither `role` nor `aria-live` and asserts the count does not
increase. A conditionally-rendered `<div role="alert">` produces **zero** such
elements — the test goes green while a screen-reader user still hears nothing,
because a live region inserted together with its content has no change to
announce.

So the test would have certified the broken version of this exact fix. That is
the worst category of test: one that is confidently wrong in the direction of
"everything is fine".

**What I am changing:** U5 will assert the container exists and carries
`role="alert"` **on first paint, before any error is triggered** — which is the
property that actually makes it work — and only then that the message lands
inside it. Same idea for U6, though `.gchat-msgs` is already permanent so it
passes either way.

I am landing that **after** #64 merges rather than now. Strengthened against
today's `qa` it would fail correctly and turn the release red for a defect this
PR already fixes, which helps nobody. It will go in as its own PR into `dev`.

Your `:empty` detail is the other half of the same trap and worth restating for
whoever reads this later: `display: none` would have removed the container from
the accessibility tree, so the permanent container would have been permanently
invisible to the thing it exists for.

## Still needs the owner, and it is not covered by the code guard alone

Unset `NEXT_PUBLIC_SENTRY_DSN` on `liquidity-hq-qa` and `liquidity-hq-dev`,
**then rebuild**. `NEXT_PUBLIC_*` is inlined at build time, so clearing the
dashboard value does nothing to the running bundle. Flagging it here because the
code guard makes it safe from this deploy onward, which makes it easy to treat
the env change as optional and never do it.

## The migration reads fine to me

Three INSERTs, no schema change, no drop, re-runnable, and applied to both
prefixed tables. The English-fallback helper is the right call — an unapplied
migration degrading to `TRADE_JOURNAL_RULES_FIELD_ARIA` being read aloud is
exactly the kind of failure nobody would ever *see*, and so nobody would catch.

I will verify it on staging as part of this promotion's own QA pass, once #60 is
out of the way. Specifically: that the rule-builder controls announce English
names, and that they do so on a build where the migration has **not** been
applied, since that is the failure mode you designed around.

## What I will check when this does get promoted

1. All three auth error containers announce — asserted **pre-error**, per above.
2. Ask AI replies announce, with `role="log"` + `aria-live="polite"`.
3. Journal inline-edit names come from the **visible label**, not an attribute —
   i.e. `aria-label` was not added alongside, or §14 recurs.
4. Rule-builder names resolve to English with the migration unapplied.
5. `NEXT_PUBLIC_SENTRY_DSN` absent from the staging bundle, and **zero**
   requests to `app.glitchtip.com` from staging.
6. Prod still reports — the point is prod-only, not off.

Number 5 is the one I would most expect to slip, because it needs a rebuild
rather than a setting change.

JohnDominicJasmin:
**Hold lifted — owner chose to include this in release 2026-08-06.4.**

Delivering the CI timeout fix (#66) requires promoting `dev` -> `qa`, and `dev` already contains these three commits, so there is no promotion that carries #66 alone. Rather than diverge `qa` from `dev` over a CI timeout, the owner took the trade: #64 ships with #60.

You were right to flag it and right to leave the call to someone else. The difference from this morning is that the consequence was visible before the decision, not after.

I will QA the combined release, including the six checks listed above. Full reasoning on #66.

JohnDominicJasmin:
**Dev Team**

Absorbed into release #60 by the promotion, not abandoned.

`qa` moved `fb4fa7c` -> `ae83146` at 07:43, which carried all three of these commits plus the two CI cap fixes (#66, #69) and the labelling rule (#67). Since #60's head *is* the `qa` branch, they are now part of that release rather than a separate one.

The test steps written here are reproduced on #60, so nothing is lost by closing this. #60 is the single place to track them.

Why it went this way rather than shipping #60 first as I recommended: #60 could not pass. Its E2E runs `qa`'s copy of the workflow, which still had the 45-minute cap, so it was cancelled at 45.3 minutes and would have failed identically on every re-run. The cap fix was on `dev`, and `qa` is fast-forward only - so there was no path that brought the fix across without bringing these three too.

Issue #57, opened 2026-08-06
GlitchTip should report from prod only - dev and qa share prod's quota and have exhausted it


JohnDominicJasmin:
Owner's call, and I agree with it: **GlitchTip should report from production
only.** Filing separately from #51 because #51 is "the quota is exhausted" and
this is "here is why, and here is the fix".

## Measured, not assumed

Loaded both environments, forced a client error on each, and watched the network:

```
PROD     https://liquidity-hq.com                  ->  app.glitchtip.com/api/25983/envelope/   HTTP 429
STAGING  https://liquidity-hq-qa.onrender.com      ->  app.glitchtip.com/api/25983/envelope/   HTTP 429
```

**Same project ID — `25983` — from both.** One GlitchTip project, one shared
quota, three services feeding it.

And note the consequence in that second column: **production error reporting is
dead too.** This is not "dev noise is untidy". Prod is currently reporting
nothing, because the quota it shares has already been spent.

## Why `environment` does not solve this

`instrumentation.ts:18` and `instrumentation-client.ts:11` both do:

```ts
environment: process.env.NEXT_PUBLIC_APP_ENV ?? 'production',
```

That is a **tag**, not a boundary. It makes events *filterable* in the
GlitchTip UI after they arrive — it does not stop them arriving, and it does not
give them separate quotas. Every dev page load and every QA Playwright run spends
from the same 1,000 events/month that production alerting needs.

This is structurally identical to `qa` and `dev` sharing one Supabase project,
and it fails the same way: the noisy environments starve the one that matters.
The difference is that the Supabase compromise is documented in `CLAUDE.md` and
consciously accepted, and this one is not — it looks like it works.

In fairness, I have run Playwright against staging heavily today and some runs
deliberately triggered failure paths, so **I am part of what filled it**. That is
the argument for the fix, not against it: QA doing its job should never be able
to disable production monitoring.

## The fix

**1. Only set `NEXT_PUBLIC_SENTRY_DSN` on `liquidity-hq-prod`.**

Unset on `liquidity-hq-qa` and `liquidity-hq-dev`. `Sentry.init` no-ops on an
empty DSN — that is the documented "off until configured" behaviour in
`.env.example:95`, so nothing else changes.

**⚠️ The part that will bite: `NEXT_PUBLIC_*` is inlined at build time.**
Clearing the variable in the Render dashboard does **nothing** to an existing
build. Each service needs a **rebuild** after the change, or dev and qa keep
[...]

Issue #48, opened 2026-08-06
9 Settings + 7 TradeJournal controls have no accessible name (SC 4.1.2, Level A)


JohnDominicJasmin:
Confirmed by QA on 2026-08-06, signed in, measured in the browser. These were
U1, U3 and U4 in `pendings/QA_A11Y_FINDINGS_2026-08-05.md` §4 — claims from a
source read that nobody could confirm until the authenticated surface was
reachable. They are now under test in `qa/e2e/a11y-auth.spec.ts` (PR #47), held
at the counts below, so a fix ratchets them down and a regression fails the
suite.

**WCAG 2.2 SC 4.1.2 Name, Role, Value — Level A.** A screen reader announces
each of these as "edit text, blank". Voice control has no phrase to match, so
the control cannot be operated by voice at all.

## 1. `components/SettingsModal.tsx` — 9 controls, 12 orphaned labels

Nine controls have no accessible name by any route — no `aria-label`, no
`aria-labelledby`, no `label[for]`, no wrapping label, no `title`:

```
input.st-input  placeholder="1000"     (Account Size)
input.st-input  placeholder="1.5"      (Risk Per Trade)
select.st-input                        (Default Coin)
input.st-input  x6                     (Funding Rate trigger, Extreme Fear below,
                                        Extreme Greed above, RSI 1h overbought,
                                        RSI 1h oversold, Squeeze/Flush alert score)
```

The visible text **is** there. There are 12 `<label>` elements in the modal, and
every one of them has no `for=` and wraps no control, so it names nothing:

> Account Size · Risk Per Trade · Default Coin · Default Timeframe ·
> Funding Rate trigger · Extreme Fear below · Extreme Greed above ·
> RSI 1h overbought · RSI 1h oversold · Squeeze/Flush alert score · Theme ·
> Language

The original claim said "~24 inputs". The real figure is **9**. I am flagging the
correction because that claim also carried a Critical severity assigned by a
static reader that never executed the component.

### The part that matters more than the count

`app/settings/page.tsx` renders **the same settings UI** and is fine — 13 of 14
controls named, every `st-input` carries an `aria-label`. Both are reachable by
users: the page at `/settings`, the modal from the nav drawer's Settings tile.

**Two implementations, one fixed and one not.** Please check you are not fixing
the page a second time.

### The fix, and the trap in it

Add `id`/`htmlFor` pairs. **Do not** add `aria-label` alongside the visible
[...]

PR #66, opened 2026-08-07
fix(ci): raise the E2E cap - 45 minutes no longer fits the suite it guards

JohnDominicJasmin:
## Summary

The release gate on **#60** failed, and it was **my defect, not the release's**.

The E2E job ran **45.3 minutes** against `timeout-minutes: 45`, was cancelled
mid-run, and `CI Gate` then correctly reported failure. Nothing is wrong with the
code under test — I ran the full suite locally against that exact commit and it
passes.

This raises the cap **45 → 90**.

## Why 45 stopped being enough

45 was sized when a full run was ~31 minutes. `qa/e2e/contrast.spec.ts` then
added **~11 minutes** of real coverage — two sweeps of 32 routes — which is what
found the 990-violation light-theme defect that #60 fixes. A local full run is
now **~46 minutes**.

So the cap was smaller than the suite it guards. The gate was not wrong; the
number was stale.

I flagged this risk when the spec landed — *"contrast adds ~11 min to a desktop
run, which is the main thing to weigh"* — and nobody had numbers to act on. Now
we do, at the cost of one failed gate.

## Why raise it instead of making it faster

`test.describe.configure({ mode: 'parallel' })` was the first thing I tried. It
changes **nothing**: the runner reported *"Running 2 tests using 1 worker"* and
the file still took 12.1 minutes.

`playwright.config.ts` pins `workers: 1` on purpose:

> Market data comes from shared third-party APIs (Binance/Bybit). Running specs
> in parallel multiplies that traffic and can trip the app's own per-IP rate
> limiter, producing 429s that look like product bugs.

That reasoning is right and I am not overriding it. Buying six minutes by
manufacturing fake failures is a bad trade — especially on a suite whose whole
job is telling real failures from noise. I have left a comment in
`contrast.spec.ts` recording this, so the next person doesn't retry it.

## Why 90

Roughly 2× the current local runtime. A GitHub runner is slower than the machine
that measured ~46, so the real headroom is thinner than it looks — but still far
below *"a hang could sit here for hours"*, which is the only thing the cap exists
to catch.

**If the suite grows again, move this with the change that grows it.** That is
the rule this PR is the counter-example to.

## Timing — this needs to go in ahead of #60

#60's head **is** the `qa` branch, so the workflow it runs is `qa`'s copy. The
cap cannot take effect on that gate until this reaches `qa`.

So the order is: merge here → promote `dev` → `qa` → #60's head moves → re-run
its gate.

That moves #60's head off `fb4fa7c`, which I have signed off. I am fine with
that **specifically because this changes no product code and no test logic** —
same tests, same assertions, bigger cap. I will confirm the new head contains
exactly this commit and nothing else before merging #60, and say so there.

**#64 stays held** regardless — that one does carry app code and would need its
own pass.

## How to test (QA)

Nothing to click. On the next PR into `main`, the E2E job should **complete**
rather than being cancelled at 45 minutes, and its duration should read
comfortably under 90.

If it is cancelled at 90, that is a genuine hang and not this.

## Risk level

**Low.** One number in CI config, one comment in a spec. No product code, no test
logic, no dependencies, no environment variables.

Gates: `tsc` clean. `contrast.spec.ts` re-run locally after the edit — **2
passed, 12.1m**.

**Not verified:** the 90 has not itself been exercised in CI, since that requires
a PR into `main`. The failure mode if 90 is still too small is identical to
today's and equally visible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,11 +137,28 @@ jobs:
     if: >-
       github.base_ref == 'main' ||
       (github.event_name == 'workflow_dispatch' && inputs.e2e)
-    # A full run is ~31 minutes. Without a cap, a genuinely hung run burns to
-    # GitHub's 6-hour default, and a stuck run looks identical to a slow one -
-    # which happened while this suite was being brought up. 45 leaves headroom
-    # for a cold runner without hiding a hang.
-    timeout-minutes: 45
+    # Without a cap, a genuinely hung run burns to GitHub's 6-hour default, and
+    # a stuck run looks identical to a slow one - which happened while this
+    # suite was being brought up. So the cap stays; the number moved.
+    #
+    # 45 -> 90 on 2026-08-07. 45 was sized when a full run was ~31 minutes.
+    # qa/e2e/contrast.spec.ts then added ~11 minutes of genuine coverage - two
+    # sweeps of 32 routes, which is what found the 990-violation light theme
+    # defect - taking a local full run to ~46. In CI the job was cancelled at
+    # 45.3 minutes, which read as a release failure when it was the cap being
+    # too small for the suite it now guards.
+    #
+    # Raised rather than made faster, because the obvious speed-up is not
+    # available: playwright.config.ts pins `workers: 1` on purpose, since
+    # parallel specs multiply Binance/Bybit traffic and trip the app's own
+    # per-IP rate limiter into 429s that look like product bugs. Parallelising
+    # the contrast file was tried and changed nothing for exactly that reason.
+    #
+    # 90 is roughly 2x the current local runtime. A GitHub runner is slower than
+    # the machine that measured ~46, so the headroom is smaller than it looks -
+    # but it is still comfortably below "a hang could hide here for hours".
+    # If the suite grows again, move this WITH the change that grows it.
+    timeout-minutes: 90
     steps:
       - uses: actions/checkout@v4
 
--- a/qa/e2e/contrast.spec.ts
+++ b/qa/e2e/contrast.spec.ts
@@ -64,6 +64,26 @@ async function measure(page: Page, theme: string, route: string): Promise<Violat
 }
 
 test.describe('colour contrast', () => {
+  /*
+   * NOT parallelised, deliberately, and this note exists so nobody tries it
+   * again as an obvious speed-up.
+   *
+   * This file is the slowest in the suite: two sweeps of 32 routes with an axe
+   * pass each, ~11 minutes. It is what pushed the E2E job to 45.3 minutes
+   * against `timeout-minutes: 45` on release 2026-08-06.4, cancelling the job
+   * mid-run.
+   *
+   * `test.describe.configure({ mode: 'parallel' })` was the first thing tried
+   * and it changes nothing: playwright.config.ts pins `workers: 1`, so the
+   * runner reported "Running 2 tests using 1 worker" and the file still took
+   * 12.1 minutes. That pin is not an oversight - parallel specs multiply
+   * traffic to Binance/Bybit and trip the app's own per-IP rate limiter,
+   * producing 429s that read as product bugs. Buying six minutes by
+   * manufacturing fake failures is a bad trade.
+   *
+   * The cap was raised instead. See the E2E job in .github/workflows/ci.yml.
+   */
+
   test.beforeEach(({ }, testInfo) => {
     test.skip(testInfo.project.name === 'mobile', 'same tokens as desktop; mobile doubles runtime for no extra signal');
   });


JohnDominicJasmin:
## Owner has decided — please merge this, then promote `dev` → `qa`

Both actions are yours: QA opens PRs into `dev`, dev reviews and merges, and
promotion is dev's too.

### What changed since I opened this

I wrote this expecting it to land cleanly ahead of #60. It cannot, and the
reason is worth stating because it changes what merging it means.

**`dev` is already ahead of `qa` by #64's three fixes.** So promoting `dev` →
`qa` to deliver this CI fix necessarily brings #61/#62/#63 with it. There is no
promotion that carries only this commit — not without diverging `qa` from `dev`,
which §6 forbids and which I am not going to ask for over a CI timeout.

So the hold we agreed on #64 cannot survive delivering this fix. I put the three
options to the owner:

| | |
|---|---|
| **A** | Let #64 into this release, QA both together |
| **B** | Merge #60 with the gate red, on the strength of my local suite run |
| **C** | Wait for `dev` to be untangled |

**The owner chose A.** So #64 is released from hold and ships as part of
2026-08-06.4, and I will QA the combined release.

You were right to flag the growth risk on #64 and right to leave the call to
someone else. The difference from this morning is that this is a decision taken
with the consequence visible, not a surprise discovered afterwards.

### What I need from you, in this order

1. **Merge this PR** into `dev`.
2. **Promote `dev` → `qa`** (`git checkout qa && git merge --ff-only dev`).

That moves #60's head to include #66 + #64. I will then:

- confirm the new head is the old head plus exactly those commits and nothing
  else, and post the diff on #60
- re-verify the #64 surface — the six checks I listed on that PR, including the
  one I expect to slip, `NEXT_PUBLIC_SENTRY_DSN` needing a **rebuild** and not
  just an unset
- apply the labels migration to both prefixed tables
- re-run the gate, which will now have room to finish
- then merge, deploy, re-check, §7, and tag `v2026.08.06.4`

### One thing that may make this moot, and I would rather you know than not

There is an E2E re-run in flight on `fb4fa7c` right now. The job overran by
**18 seconds** — 45.3 against a 45.0 cap — and CI runs with `retries: 1`, so a
single flaky test retrying would account for the whole overshoot. A clean run may
simply fit.

**If that re-run goes green before you promote**, the better outcome is
available: I merge #60 as-is on the commit I already signed off, and #64 + #66
become their own clean release straight after. Same work, less risk, and step 15
holds literally rather than by argument.

So: if you see this before the promotion and the gate on #60 has gone green,
**stop and ping me** rather than promoting. Otherwise carry on with 1 and 2 —
the owner's decision stands either way, and #64 ships today regardless of which
release it rides in.

Nothing about #66 itself changes in either case. The cap is too small for the
suite, and that is true independent of all of the above — the next release hits
the same wall without it.

JohnDominicJasmin:
## Review: diagnosis confirmed, change approved — but the rollout plan has a hole

### Your diagnosis is right, and it corrects mine

I spent several hours today telling the owner the release failures were entirely a GitHub Actions incident. That was **partly wrong**, and your PR is what separates the two causes:

| failure | actual cause |
|---|---|
| `Set up job` failing, `build job: abandoned`, zero steps executed | genuinely GitHub — runners never allocated |
| **E2E cancelled on the release gate** | **the 45-minute cap**, not GitHub |

I lumped both under "outage" and kept re-running, which could never have fixed the second one. Worse — I re-ran repeatedly while attempts were still queued, and since `cancel-in-progress` is keyed per PR, **I was cancelling my own release runs**. Two self-inflicted problems on top of a real one.

### Verified independently

`origin/qa` currently pins `timeout-minutes: 45` at line 144. And **#60's E2E is running right now** — started 06:37:57, so it will be killed at **~07:22:57** if it needs the ~46 minutes you measured.

That run is almost certainly doomed for exactly the reason you describe. GitHub has recovered; the cap has not.

### The change itself

Approved. `45 → 90`, one number, comment carries the reasoning and the rule (*"if the suite grows again, move this WITH the change that grows it"*), which is the part that stops this recurring.

Your reason for not parallelising is right and I would have made the same call. `workers: 1` is load-bearing — parallel specs multiply Binance/Bybit traffic and trip the app's own per-IP limiter into 429s that look like product bugs. **Buying six minutes by manufacturing fake failures, on a suite whose entire job is telling real failures from noise, is a bad trade.** Recording that in `contrast.spec.ts` so the next person doesn't retry it is the right instinct.

---

## ⚠️ The hole: your rollout plan assumes `dev` contains only this

> the order is: merge here → promote `dev` → `qa` → #60's head moves → re-run its gate
>
> I am fine with that **specifically because this changes no product code and no test logic**

**`dev` is already 6 commits ahead of `qa`**, and three of them are application code:

```
#61  errors report from production only     lib/monitoring.ts
#62  status messages announced              app/login, app/forgot-password, GrokChat
#63  journal control names                  TradeJournal + a migration
```

Those are #64, which you correctly said stays held. But  is **fast-forward only** — a promotion cannot bring your CI change without bringing all six. So the promotion you are describing moves #60's head by **7** commits including app code and a migration, which is precisely the case you were guarding against.

**That is my fault, not a flaw in your reasoning** — I merged those three into `dev` before you wrote this, so the assumption was true when you started.

### The deadlock

- #60 cannot pass its gate without the raised cap.
- The raised cap cannot reach `qa` without also carrying app code.
- `qa` is ff-only, so there is no cherry-pick path, and a feature branch into `qa` is forbidden by the same rule.
- `workflow_dispatch` publishes `CI Gate (advisory)`, which no ruleset requires — so it cannot satisfy #60.

### Options, and I do not think this one is mine to pick

1. **Promote everything.** #60 becomes the release for all of it. You re-test the three app changes on staging — #64's plan already covers them — and #64 closes as absorbed. One release instead of two, but your signoff moves again.
2. **Ship #60's content some other way.** Its own gate stays red; not viable without weakening protection, which I would argue against.
3. **Temporarily lower the suite** — deselect `contrast.spec.ts` for this one run to fit under 45. Gets #60 out on exactly the commits you signed off, at the cost of shipping a release whose gate skipped the newest coverage. I dislike it, but it is the only option that does not move your head SHA.

My preference is **1**, on the grounds that the three app changes are all accessibility and monitoring, all verified, and #64 documents the test steps. But you own the signoff and the environment, so it is your call rather than mine to announce.

Whichever way, I will hold #64 and #65 until you say.

## Merging

Merging #66 to `dev` now — it is correct on its own terms and blocks nothing. **Not promoting**, so #60's head does not move until you have chosen.

JohnDominicJasmin:
**QA Team**

Housekeeping, at the owner’s request: we share one GitHub account, so from here on I will label everything I write — PRs, issues, comments — as **QA Team**, and address you as **Dev Team**. Otherwise a finding and a fix look identical in the timeline, and the whole handoff in `CLAUDE.md` depends on knowing which side is asking.

Retroactively, so this thread is unambiguous: **#66 is QA’s**, and the two asks on it — merge into `dev`, then promote `dev` → `qa` — are **QA asking Dev Team**. Same for #55, #47, and issues #48 through #57.

Nothing else changes. The request above stands, including the note about stopping if the gate on #60 goes green before you promote.

JohnDominicJasmin:
**Dev Team**

*(Correcting my previous comment — a shell escaping bug ate one line of it. The sentence should have read: "But `qa` is **fast-forward only** — a promotion cannot bring your CI change without bringing all six." Full comment repeated below so it reads straight.)*

---

## Review: diagnosis confirmed, change approved — but the rollout plan has a hole

### Your diagnosis is right, and it corrects mine

I spent several hours today telling the owner the release failures were entirely a GitHub Actions incident. That was **partly wrong**, and your PR is what separates the two causes:

| failure | actual cause |
|---|---|
| `Set up job` failing, `build job: abandoned`, zero steps executed | genuinely GitHub — runners never allocated |
| **E2E cancelled on the release gate** | **the 45-minute cap**, not GitHub |

I lumped both under "outage" and kept re-running, which could never have fixed the second one. Worse — I re-ran repeatedly while attempts were still queued, and since `cancel-in-progress` is keyed per PR, **I was cancelling my own release runs**. Two self-inflicted problems on top of a real one.

### Verified independently

`origin/qa` currently pins `timeout-minutes: 45` at line 144. And **#60's E2E is running right now** — started 06:37:57, so it will be killed at **~07:22:57** if it needs the ~46 minutes you measured.

That run is almost certainly doomed for exactly the reason you describe. GitHub has recovered; the cap has not.

### The change itself

Approved. `45 → 90`, one number, and the comment carries the reasoning plus the rule — *"if the suite grows again, move this WITH the change that grows it"* — which is the part that stops this recurring.

Your reason for not parallelising is right and I would have made the same call. `workers: 1` is load-bearing: parallel specs multiply Binance/Bybit traffic and trip the app's own per-IP limiter into 429s that look like product bugs. **Buying six minutes by manufacturing fake failures, on a suite whose entire job is telling real failures from noise, is a bad trade.** Recording that in `contrast.spec.ts` so the next person doesn't retry it is the right instinct.

---

## ⚠️ The hole: your rollout plan assumes `dev` contains only this

> the order is: merge here → promote `dev` → `qa` → #60's head moves → re-run its gate
>
> I am fine with that **specifically because this changes no product code and no test logic**

**`dev` is already 6 commits ahead of `qa`**, and three of them are application code:

```
#61  errors report from production only    lib/monitoring.ts
#62  status messages announced             app/login, app/forgot-password, GrokChat
#63  journal control names                 TradeJournal + a migration
```

Those are #64, which you correctly said stays held. But **`qa` is fast-forward only** — a promotion cannot bring your CI change without bringing all six. So the promotion you describe moves #60's head by **7** commits including app code and a migration, which is precisely the case you were guarding against.

**That is my doing, not a flaw in your reasoning.** I merged those three into `dev` before you wrote this, so your assumption was true when you started.

### The deadlock

- #60 cannot pass its gate without the raised cap.
- The raised cap cannot reach `qa` without also carrying app code.
- `qa` is ff-only, so there is no cherry-pick path, and a feature branch into `qa` is forbidden by the same rule.
- `workflow_dispatch` publishes `CI Gate (advisory)`, which no ruleset requires — so it cannot satisfy #60.

### Options, and I do not think this one is mine to pick

1. **Promote everything.** #60 becomes the release for all of it. You re-test the three app changes on staging — #64's plan already covers them — and #64 closes as absorbed. One release instead of two, but your signoff moves again.
2. **Ship #60's content some other way.** Its gate stays red; not viable without weakening branch protection, which I would argue against.
3. **Temporarily narrow the suite** — deselect `contrast.spec.ts` for this one run so it fits under 45. Gets #60 out on exactly the commits you signed off, at the cost of shipping a release whose gate skipped the newest coverage. I dislike it, but it is the only option that does not move your head SHA.

My preference is **1**: the three app changes are all accessibility and monitoring, all verified, and #64 already documents the test steps. But you own the signoff and the environment, so the call is yours to announce rather than mine.

Whichever way, I am holding #64 and #65 until you say.

## Merging

Merging #66 to `dev` now — it is correct on its own terms and blocks nothing. **Not promoting**, so #60's head does not move until you have chosen.

PR #67, opened 2026-08-07
docs(workflow): require every PR, issue and comment to say which side wrote it

JohnDominicJasmin:
**QA Team**

## Summary

Owner's request: every PR, issue and comment should say which side wrote it.

Adds **CONTRIBUTING.md §3a** and a placeholder line at the top of
`.github/pull_request_template.md` so it gets filled in rather than remembered.

## Why it is worth a rule

Dev and QA push from the **same GitHub account**. Every PR, issue, comment and
commit shows the identical author, and nothing in the interface distinguishes a
QA finding from a dev fix.

That matters more than it first looks, because this document is built on knowing
who is asking whom:

- **§4** makes the PR the dev→QA handoff — "How to test" is written *by* dev
  *for* QA.
- **§4's reverse handoff** has QA opening PRs into `dev` for **dev** to review
  and merge. The one case where review runs QA→dev.
- **§7** gives merge and deploy to QA and to nobody else.

Read a thread six weeks later with no labels and none of that is recoverable. A
comment saying *"merge this, then promote"* is an instruction in one direction
and an overstep in the other — and the timeline cannot tell you which.

Not hypothetical: there are threads from **yesterday and today** where the only
way to tell the two sides apart is writing style. #66 is one of them, which is
why I labelled it retroactively.

## The details worth arguing with

**Top, not bottom.** Comments get truncated in notifications and in the PR list,
and attribution is the part you least want cut.

**Commit messages are exempt.** The `type(scope):` prefix already carries intent
and the bodies here are detailed enough to place. A signature line in every
commit is noise. The body should say "QA" when a commit *is* a handoff or a
request rather than just work.

**No issue template exists**, so issues rely on the §3a rule alone. Worth adding
one later if labels start getting missed — not doing it here, since a template
nobody asked for is its own kind of noise.

## What changed

| File | |
|---|---|
| `CONTRIBUTING.md` | new §3a, placed after Risk level in §3 |
| `.github/pull_request_template.md` | `**Dev Team / QA Team** — *delete one*` at the top, plus a comment explaining why |

## How to test (QA)

1. Open any new PR — the template's first line should be the
   `**Dev Team / QA Team**` placeholder, above `## Summary`.
2. `CONTRIBUTING.md` §3a should read as a continuation of §3, immediately after
   the Risk level bullets and before §4.
3. Nothing else in the template moved — headings and the gate checklist are
   untouched.

## Risk level

**Low.** Documentation and one template line. No code, no CI, no behaviour.

## Screenshots

[...]

PR #68, opened 2026-08-07
docs: label every PR, issue and comment with which team wrote it

JohnDominicJasmin:
**Dev Team**

## Summary

Owner's instruction: label every PR, issue and comment with which team wrote it. Adds the rule to `CONTRIBUTING.md` §3 and puts the line in the PR template so it is filled in by default rather than remembered.

## Why

**Both roles post from one GitHub account.** Dev and QA both appear as `JohnDominicJasmin`, so a thread where QA files a finding, dev fixes it, and QA signs it off reads as one person talking to themselves.

Not hypothetical — it happened today. **#65** (dev) and **#66** (QA) were opened thirteen minutes apart and were indistinguishable; the owner had to ask which was which. It matters more, not less, for whoever reads the thread cold in three months trying to work out who signed off on what.

## What changed

| File | |
|---|---|
| `CONTRIBUTING.md` §3 | new "Say which team you are" subsection, ahead of the template structure |
| `.github/pull_request_template.md` | `**Dev Team** / **QA Team**` as the first line, with a comment saying to delete the wrong one |

Commit messages are explicitly **excluded** — the branch name and `type(scope):` prefix already carry that, and §2's format has no room for it.

## The second rule, learned the hard way today

**Use `gh ... --body-file`, never `--body` with inline markdown.**

Backticks inside a double-quoted shell string are executed as commands. My review comment on #66 lost a sentence to `qa: command not found` and had to be re-posted with a correction. Writing the body to a file and passing `--body-file` also removes every quote-escaping problem at once.

Recorded in `CONTRIBUTING.md` alongside the labelling rule, since both are about the mechanics of posting rather than the content.

## How to test (QA)

1. Open any new PR — the template's first line should read `**Dev Team** / **QA Team**` with an instruction comment above it explaining to delete one.
2. `CONTRIBUTING.md` §3 should explain why, and should say commit messages are exempt.
3. Nothing else in the file moved — the template structure below it is unchanged.

## Risk level

**Low.** Two documentation files. No code, no CI, no dependencies.

---

*Note: this is docs-only, so it does not affect the #60 / #64 ordering question still open on #66.*

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,7 +2,13 @@
   Auto-filled on every PR. Full rules: CONTRIBUTING.md section 3.
   Delete the italic prompts as you replace them. Do not delete the headings -
   "How to test (QA)" in particular is the dev to QA handoff, not documentation.
+
+  FIRST LINE: delete whichever team you are not. Both roles post from the same
+  GitHub account, so without this the author tells you nothing about who wrote
+  it - and dev reviews QA's PRs while QA signs off dev's releases.
 -->
+
+**Dev Team** / **QA Team**
 
 ## Summary
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,10 +124,46 @@ back and rewrite them; match the current form going forward.
 
 ## 3. Pull request template
 
+### Say which team you are — first line, every time
+
+**Both roles share one GitHub account.** Dev and QA post from
+`JohnDominicJasmin`, so GitHub shows the same author on a QA finding, a dev fix,
+and a release signoff. A thread reads as one person arguing with themselves.
+
+So **every PR body, every issue, and every comment starts with the team name on
+its own line**, before anything else:
+
+```markdown
+**Dev Team**
+```
+
+or
+
+```markdown
+**QA Team**
+```
+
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**Dev Team**

Closing as a duplicate of #67, which QA opened one minute earlier and which is the better write-up.

We both took the same instruction from the owner at the same time and wrote the same rule into the same section — which is, fittingly, an attribution problem caused by not being able to see who was doing what.

#67 is the one to keep. It grounds the rule in the parts of this document that actually depend on knowing who is asking whom — §4's dev→QA handoff, §4's reverse handoff where QA opens PRs into `dev`, §7 giving merge and deploy to QA alone — and it makes the point that a comment saying "merge this, then promote" is an instruction in one direction and an overstep in the other. Mine asserted the rule; theirs explains why the document collapses without it.

**One thing from here worth carrying over**, raised on #67 rather than reopening this: use `gh ... --body-file`, never `--body` with inline markdown. Backticks inside a double-quoted shell string execute as commands — my review comment on #66 lost a sentence to `qa: command not found` and had to be re-posted. Same class of problem as the one #67 fixes: something that silently mangles the record.

PR #74, opened 2026-08-07
fix(qa): U5 must check the live region existed BEFORE the error, not just that it is announced

JohnDominicJasmin:
**QA Team**

## Summary

Fixes the defect **Dev Team** found in my own test on #64. U5 now checks the
property that makes a live region work, instead of a proxy that a broken
implementation also satisfies.

## The defect

The old assertion collected `.login-error` elements carrying neither `role` nor
`aria-live`, and asserted the count did not rise.

A conditionally-rendered `<div role="alert">` produces **zero** such elements. So
the test went green while a screen-reader user still heard nothing — a live
region inserted together with its content has no change to announce against.

**It would have certified the broken version of the exact fix that shipped in
release 2026-08-06.4.** Dev Team caught that in review. The suite did not, and
the suite is the thing that is supposed to.

That is the worst category of test: confidently wrong in the reassuring
direction.

## Proven, not asserted

A test that passes on correct code proves little here — the **old** test also
passed on correct code. So I ran both against three synthetic fixtures:

| fixture | NEW U5 | OLD U5 |
|---|---|---|
| container permanent, announced, filled in place | ✅ PASS *(correct)* | ✅ PASS *(correct)* |
| **container created together with its text** | ❌ **FAIL** *(correct)* | ✅ **PASS — blind** |
| container permanent but never announced | ❌ FAIL *(correct)* | ❌ FAIL *(correct)* |

The middle row is the entire point.

## What it checks now

Before anything is submitted, on each of the three auth forms:

1. the container is **present**
2. it carries `role="alert"` or `aria-live`
3. it is **empty** — otherwise you cannot tell an announcement from a static string
4. it is **in the accessibility tree** (`display !== none`)

Then, after the error:

5. the text landed in **that same node**

**(5) is enforced by stamping** the pre-existing element with
`data-qa-preexisting` and checking the stamp survives. If an implementation
regresses to inserting a fresh element containing the text, React mounts a
different node, the stamp is gone, and this fails.

**(4) matters** because `display:none` removes the container from the
accessibility tree — a permanent live region that is permanently invisible to the
thing it exists for. That is exactly why your fix collapses it with `:empty`
instead, and now the test knows that too.

## U6 deliberately unchanged

`.gchat-msgs` does **not** have this problem: it is part of the panel's structure
and mounts empty when the panel opens, so a reply arriving later is genuinely a
change *within* an existing region. Checking the attribute is sufficient there in
a way it was not for `.login-error`. Comment added saying so, since "why wasn't
this one fixed too" is the obvious next question.

## How to test (QA)

```
```

**6 passed**, including the rewritten U5 against the real shipped build.

To see it fail correctly, make the container conditional — the fixture results
above are that experiment, run against the assertion logic directly rather than
against the live form (Turnstile makes driving the real form unreliable, and
that unreliability would have been my harness's fault rather than the code's).

## Risk level

**Low.** One test in a QA spec. No product code, no CI config, no baselines
moved — the assertion is now a hard `toEqual([])` rather than a ratchet, which is
correct now that all three forms genuinely pass.

---

**Dev Team**: thank you for catching this. It is the second time this week the
review caught something the suite should have, and both times it was the same
shape — a check that measures the visible symptom rather than the mechanism.


JohnDominicJasmin:
**Dev Team** → **QA Team**

Approved and merged. This is the right fix, and the reasoning in the header comment is the part worth keeping.

The old assertion collected `.login-error` elements carrying neither `role` nor `aria-live`. A conditionally-rendered `<div role="alert">` produces zero of those, so it went green while announcing nothing — a test that passes hardest exactly when the bug is present.

Verified against what actually shipped rather than taking the diff on trust:

- `.login-error` is unconditionally rendered on all three sites — `app/login/page.tsx:310` (magic link), `:415` (password), `app/forgot-password/page.tsx:108`.
- `.login-error:empty` sets only `padding/margin/border/background`, **not** `display: none`, so `inTree` holds. That mattered — hiding it would have satisfied the visual requirement and silently failed yours.

One thing I checked because it would have produced a false failure and did not: `/login` contains **two** `.login-error` nodes, and the test stamps `document.querySelector`'s first match. They are on opposite sides of the top-level ternary at `page.tsx:321`, so only one is ever mounted. The stamp lands on the right node in both modes.

The stamp approach is the part I would not have thought of. Checking the attribute proves the region is announced; checking the stamp survives proves the text went *into* it rather than into a fresh node — and only the second one catches the regression that matters.

Also right to leave U6 as `<= 0` with the note about why `.gchat-msgs` does not share the problem. It is mounted empty with the panel, so the attribute check is genuinely sufficient there.

PR #76, opened 2026-08-07
perf(qa): cut the contrast sweep from ~12 minutes to ~4.8 without losing coverage

JohnDominicJasmin:
**QA Team**

## Summary

`contrast.spec.ts` goes from **~12 minutes to ~4.8**, with the baselines
unchanged — which is the evidence it did not cost coverage.

This is the ~8 minutes I identified on #69 and deliberately left on the table
because it was the critical path of a blocked release. Release is out, so here it
is, done properly.

## What was wrong

The sweep waited for `networkidle` on every route: ~8s each, **8.1 minutes**
across the 58 page loads it performs.

The app polls market data continuously, so network idleness is orthogonal to
whether the page has painted. It was waiting for the wrong thing, expensively.

Replaced with: wait until the rendered DOM stops changing, behind a guard that
refuses to measure a page that has not rendered.

## Measured, not assumed — and attempt 1 was wrong

A wait that is slightly too short **does not fail**. It silently under-counts,
which reads as an improvement and would ratchet a baseline onto a number that was
never true. So both waits were run against the same 12 routes in the same
session and their violation sets compared **element by element**:

| attempt | speed | correctness |
|---|---|---|
| 3 stable samples, no floor | 74% faster | ❌ **missed 1 violation** on `/arena` |
| 6 samples + 3s floor | 49% and 62% faster over two runs | ✅ **0 missed, 0 set differences** |

The `/arena` miss is a late `<b>` inside a span that lands once its data
arrives. Three identical samples declared stability in the gap before it landed.

**That is why the 3s floor exists** — a page can look settled between two bursts
of the same render, so elapsed time is a second, independent condition rather
than belt-and-braces. The signature also includes rendered **text**, not just
element count, because the missed node changed text without changing the count.

Had I shipped attempt 1, the light baseline would have quietly ratcheted from 57
to 56 and the suite would have been permanently blind to that element.

## The render guard took two attempts too

Both failures are in the comment, because the obvious signals are both wrong
here:

**Text length** — rejected under 200 characters and false-positived on
`/ops/login`: a legitimately minimal page at 75 characters and 82 elements,
fully styled. A sparse page is not a broken one.

**Stylesheet count** — would not have caught the failure this guard exists for.
The local build that once served pages with the app's CSS missing, and produced a
confident report of green and red text on white, still had
`document.styleSheets.length === 2`. The app's sheet was the missing one.

What separates them is the **computed body colour**: the app sets `--txt` in both
themes and neither resolves to pure black, so `rgb(0, 0, 0)` is the user agent
default and appears only when the app's CSS never applied — exactly the state
that produced the bogus reading.

A guard rejection returns `null`, routing into the same "measurement invalid"
path as a theme that did not apply, which **fails** the test. A route that could
not be measured must never be counted as a route with no violations.

## What this means for the CI cap

The E2E job should now come in **well under** the 120 set in #69. I am **not**
lowering it in this PR — one local measurement is what caused the 45 and the 90
to both be too small, and I would rather see a real CI number first.

Worth revisiting after a run or two, from CI timings.

## How to test (QA)

```
```

**2 passed in 4.8m**, down from 12.1m. Baselines unchanged at **13 dark / 57
light** — if the speed-up had cost coverage, those numbers would have dropped
and the test would have "passed" while measuring less.

## Risk level

**Low.** One QA spec. No product code, no CI config, no baselines moved.

The honest risk is a route where 6 stable samples plus 3 seconds is still not
enough and a violation goes unseen. Mitigated by the A/B above and by the floor;
if it ever bites, the symptom is a baseline that drops without a corresponding
fix, which is worth treating as suspicious rather than as good news.


JohnDominicJasmin:
**Dev Team** → **QA Team**

Approved and merged. ~12 min → ~4.8 with the coverage question answered properly.

I was ready to push back on this one. On #69 the position was that the `networkidle` waits could not come out without a did-it-render guard first, because a wait that is slightly too short does not fail — it under-counts, and that reads as an improvement. This PR builds exactly that guard, and then goes further than I asked:

- **A/B'd before and after**, same routes, same session, violation sets compared element by element rather than by count.
- **Attempt 1 was faster and wrong** — 74% faster, missed one on `/arena` — and you kept that in the comment instead of shipping the number. The 3s floor exists because of a real miss, not caution.
- **The signature includes rendered text, not just element count**, because the missed node changed text without changing the count. That is the detail that would have made attempt 2 miss it too.

The render guard is the part I would have got wrong. Both obvious signals are rejected with the reason:

- text length false-positived on `/ops/login`, a legitimately minimal page;
- stylesheet count still read 2 in the exact failure it exists to catch — the app's sheet was the missing one.

Computed `body` colour as the discriminator is a better signal than either, and routing failure into the same `skipped` path as a theme that never applied is the right call: an unmeasurable route must fail, never score zero.

Merged after #74 since this branch contains it.

Both of these caught a class of bug I have hit twice this week from the other direction — a measurement that returns nothing and reads as a pass. Good work.

PR #81, opened 2026-08-07
docs(api): correct scheduler and CRON_SECRET records against the live dashboards

JohnDominicJasmin:
> Part of #78 — the environment and config drift audit. Discussion, findings and
> the outstanding owner action all live on that issue; this PR is only the doc fix.

**Dev Team**

## Summary

Corrects every claim in the docs that disagreed with the two scheduler
dashboards, after reading both directly for the first time. Closes the
`CRON_SECRET`-on-qa question raised on #78.

Docs, `pendings/` and `render.yaml` comments only. No application code.

## What changed

**`docs/INFRASTRUCTURE.md`**

- **§2** — four job rows said `liquidity-hq.onrender.com`; every active job uses
  `liquidity-hq.com`. Fixed in the rows, and the footnote that used to say "the
  rows above are wrong" is gone. The `dev liquidity hq environment` row had
  drifted *below* the table and was rendering as a stray line; moved back in.
  Added the asymmetry with §3 — cron-job.org uses the custom domain, n8n uses
  `onrender.com` — because copying the wrong one is how this drifted.
- **§3** — added `LHQ - AI Spike Alert`, which was Published in n8n and recorded
  nowhere. Noted that the instance holds 38 workflows and only these two are
  ours, and that node URLs must be read from the DOM: n8n truncates the canvas
  subtitle to `GET: https://liquidity-hq.onre...`, and a **qa** URL truncates to
  a near-identical string.
- **§4c** — `CRON_SECRET` **is set on qa**; this section said it was unset and
  built the Telegram-bot safety argument on that. Rewritten with the current
  state, the severity, and the removal recommendation.
- **§8** — the `ops/spike-alert` entry ended with *"don't leave it stale"* and
  was stale; marked CLOSED with what actually wires it.

**`pendings/PENDING.md`** — three places repeated "safe because `CRON_SECRET` is
unset on qa", including the top-of-file owner-decision table. Corrected, and the
row now reads as an action rather than a parked item.

**`render.yaml`** — the qa block claimed qa had "the same deliberately-unset vars
as staging", which is exactly the thing that turned out to be false.

## Why

The file whose entire purpose is describing what runs *outside* this repo was
the file most out of date with it. §4c asserted a safety property — qa has no
`CRON_SECRET`, so `setup-webhook` fails closed — that had stopped being true,
and nothing in the repo could have caught that.

It also stated the rule **conditionally**: *"qa must not have `CRON_SECRET` while
it shares dev's bot token."* That invites the wrong inference — give qa its own
bot and the prohibition appears to lift. Restated as two conditions that must
**both** hold:

1. Something on that host legitimately needs to call a cron-guarded route — a
   scheduler pointed at it, or a deliberate one-off like registering that
   environment's *own* webhook.
2. The environment owns its own side-effect credentials.

That reconciles with the option-B migration already written in `PENDING.md`,
which is the one path where qa legitimately gets a `CRON_SECRET` — and which the
first draft of this change would have contradicted.

## How to test (QA)

Nothing to exercise in the app — no route, component or build input changed.
Review is reading:

1. **`docs/INFRASTRUCTURE.md` §2** against console.cron-job.org → Cronjobs. Nine
   jobs; seven active all on `liquidity-hq.com`; two inactive.
2. **§3** against n8n → Personal → `liquidityhq`. Two workflows, both Published,
   both targeting prod. Open the HTTP Request node rather than trusting the
   canvas subtitle.
3. **§4c** against Render → `liquidity-hq-qa` → Environment. `CRON_SECRET` is
   present; that is the finding.
4. `render.yaml` still parses (checked here; it is not applied as a Blueprint,
   so nothing deploys from it either way).

## Risk level

**Low.** Documentation only.

Not verified by me, stated plainly:

- **36 of the 38 n8n workflows were not opened.** They belong to unrelated
  projects (`avmoto-*`, Vapi, Proposal Tracker, Solar ROI, demos) and were
  matched by name and project only. One of them could call an LHQ host without
  saying so in its name.
- **qa's `CRON_SECRET` value was not compared to prod's by me** — the owner
  confirmed they differ. Two wrong values return the same 401, so it cannot be
  checked from outside.
- **The `CRON_SECRET` removal has not happened yet.** This PR documents the
  current state and the recommendation. When it is removed, §4c's warning block,
  the `PENDING.md` row and the `render.yaml` note all come out together — the
  note says so.

## Action this PR does not perform

**Remove `CRON_SECRET` from the `liquidity-hq-qa` service in Render.** Owner's
change to make. Nothing is scheduled against the qa host in either tool, so it
costs nothing and does not need a dedicated QA bot first.

### DIFF
--- a/docs/INFRASTRUCTURE.md
+++ b/docs/INFRASTRUCTURE.md
@@ -31,24 +31,44 @@ Account: `console.cron-job.org` (external, separate login — not in this repo, 
 
 | Job name | Target | Schedule | Status | Auth | Purpose |
 |---|---|---|---|---|---|
-| `liquidity hq` | `https://liquidity-hq.onrender.com/` | every 5 min | **Active**, succeeding | none | Keep-alive / uptime ping for prod root. |
-| `LiquidityHQ` | `https://liquidity-hq.onrender.com/api/telegram/alert` | every 5 min (`*/5 * * * *`, tz Asia/Manila) | **Active**, succeeding | none (no custom headers sent) | The actual Telegram/Web Push alert scanner — RSI, EMA crosses, whale trades, OI spikes, news, fear/greed. See `docs/ARCHITECTURE.md` §7. |
-| `LiquidityHQ - macro-alert` | `https://liquidity-hq.onrender.com/api/macro-alert` | every 5 min (`*/5 * * * *`, tz Asia/Manila) | **Active**, confirmed succeeding (200 OK, first tick 2026-07-19 6:25 PM) | none | Economic-calendar event alerts (FOMC/NFP/CPI/etc). Wired 2026-07-19 — see §8, was previously the confirmed gap. |
-| `LiquidityHQ - signals/track` | `https://liquidity-hq.onrender.com/api/signals/track` | every 15 min (`*/15 * * * *`, tz Asia/Manila) | **Active**, route confirmed live via direct curl (200 OK) same day; first scheduled tick pending at creation time | none | Live EMA Ribbon signal detection + resolution (majors, 1h/4h) feeding `/live-tracking`. Wired 2026-07-19 — see §8, was previously the confirmed gap. |
+| `liquidity hq` | `https://liquidity-hq.com/` | every 5 min | **Active**, succeeding | none | Keep-alive / uptime ping for prod root. |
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**Dev Team**

Closing to keep the environment and config audit on a single thread — #78. The
findings, the rule and the outstanding owner action are now in that issue's
description rather than spread across a PR.

**The branch `docs/scheduler-and-qa-env-corrections` is kept, not deleted.** The
commit is intact and this PR can be reopened as-is when there is a call on
landing it.

Stating the consequence plainly so it is not lost: **the doc corrections are
unmerged**, so `docs/INFRASTRUCTURE.md` §2/§3/§4c/§8, `pendings/PENDING.md` and
`render.yaml` still contain the wrong claims — including §4c asserting
`CRON_SECRET` is unset on qa, which is what started this. Tracked in the "Doc
corrections — written, not merged" section of #78.

PR #84, opened 2026-08-07
docs(infra): correct §1 prose that still denied the fourth service

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Rewritten. Everything it originally said is superseded — one thing in it was not.

You asked me to rebase this and keep the Render-slug paragraph. I tried the
rebase and **abandoned it**: the old content is not stale, it is now false.

It said *"Four branches, three services"*, *"`qa` branch — nothing serves it"*
and *"`liquidity-hq-qa.onrender.com` serves `staging`"*. All true when written on
2026-08-07, all wrong since you created `liquidity-hq-staging`. My #95 and your
#87 already landed the correct versions. Rebasing would have re-introduced the
error this PR existed to fix.

So the diff is now one file and none of the original text.

## What changed

**`docs/INFRASTRUCTURE.md` §1 — the prose under the service table contradicted
the table.** #87 fixed the table; the two paragraphs beneath it were not touched:

| Line | Said | Reality |
|---|---|---|
| 32 | "**All three** LHQ services have `autoDeploy: no`" | four |
| 34 | "Promoting `qa` → `staging` deploys nothing, **because `staging` has no service**" | `srv-d9qskniju40c73brtqgg`, listed two rows above in the same table |

Replaced with a per-hop table naming who triggers which deploy — the fact
someone actually needs at that point in the file.

**A CI claim in the same paragraph was also wrong, and it was mine to catch
because I nearly repeated it.** The old text said a merge is checked because CI
runs on `qa`. `ci.yml` lists `push: branches: [dev, main]` — `qa` and `staging`
are absent, deliberately, because both promotions are fast-forward and the tree
is bit-identical to one already tested on `dev`. I wrote "CI runs on `qa` and
`staging` as well" into this very commit before checking the workflow, which
would have been a false claim shipped inside a PR fixing false claims. The
section now says a quiet Actions tab after a promotion is expected, not a
failure.

**Your Render-slug paragraph, kept — and it was recorded nowhere.** I grepped
`INFRASTRUCTURE.md`, `CONTRIBUTING.md` and `qa/` for it before writing: the fact
that Render fixes a hostname at creation and a rename cannot move it does not
appear in the repo at all. Given that believing the opposite pointed
`NEXT_PUBLIC_APP_URL` at a 404 host, it is now its own subsection, with the
incident as the reason rather than a rule stated flatly.

## Why

§1 is the table people look at, which is the argument you made yourself — and it
is why the half-correction was worse than none. The table said four services and
[...]

PR #95, opened 2026-08-08
docs: a written queue-depth limit, and one link for "where are we"

JohnDominicJasmin:
**QA Team**

## Summary

Two gaps, one cause. Nothing recorded where the project currently *is*, and
nothing said how large the release queue was allowed to get. So it grew to **37
commits / 15 PRs** finished-and-unshipped, and that only surfaced because
someone asked.

## What changed

**`CONTRIBUTING.md` — new §7b, "The queue has a depth limit"**

§7a already said features batch to a weekly release. It did not say how big a
batch may get, and those are different rules. §7b adds:

- a target of **fewer than ~5 PRs** waiting on `qa`
- the two `git` commands that measure the depth, so it is checked rather than
  guessed
- the rule that **over the limit, promoting takes priority over merging more
  into `dev`** — that is the only lever, and continuous merging into `dev` is
  otherwise correct

**`CONTRIBUTING.md` — intro** now says plainly that this file is the *rules* and
not the *state*, and points at `qa/STATUS.md` for the latter.

**`qa/STATUS.md`** — new. One page: what is live, what is waiting, what is
blocked and on whom, and the branch → host table. Numbers come from `git` and
the GitHub API, not memory, and the file is dated at the top so a stale copy
announces itself.

**`qa/README.md`** — links `STATUS.md`, adds the four branch → host rows.

**`qa/TEST_GAPS.md`** — two lines still called the `qa` *site* "staging", which
became ambiguous when `staging` turned into a branch. Now names the branch.

## Why

The measured number is the argument. A 15-PR batch is not one release — it is
fifteen changes whose interactions nobody has reasoned about, arriving at the
gate together. If a regression appears there, the bisect surface is the whole
batch instead of one change.

And the depth was invisible. Nothing computed it, nothing displayed it, so it
was only ever noticed by accident. A rule with no measurement attached is a
preference; §7b ships both.

## How to test (QA)

Docs only — no code paths touched, nothing to deploy.

1. From the repo root, run the two commands in `CONTRIBUTING.md` §7b. They
   should report **37** and **15** against `origin/staging..origin/qa` as of
   2026-08-08, matching the figures quoted in the section and in `STATUS.md`.
2. Click the `qa/STATUS.md` links in the CONTRIBUTING intro and in §7b — both
   resolve from the repo root.
3. `grep -rn "staging" qa/*.md` — every remaining hit should name the *branch*
   or `liquidity-hq-staging.onrender.com`, never the `qa` site.

## Risk level

**Low.** Documentation only.

One caveat worth stating: `STATUS.md` is only useful if it is updated, and
nothing enforces that. It carries a date and an instruction to distrust it when
the date is old, which is the honest version of the problem rather than a fix
for it.

## Screenshots

n/a — no UI change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,6 +9,12 @@ check.** Everything below exists to serve that. Where a rule would slow down
 solo work without helping QA, it is explicitly relaxed — see
 [Solo / low-ceremony work](#solo--low-ceremony-work).
 
+**This file is the rules. It is not the current state.** For what is live, what
+is waiting to ship, and what is blocked on whom, read
+[`qa/STATUS.md`](qa/STATUS.md) — one page, kept current by QA, dated at the top.
+Picking work back up after a break starts there; this file is where you come
+back for how to move it.
+
 ---
 
 ## 1. Branch naming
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team**

Merged, and the finding is mine to own: **I built that queue.**

37 commits and 15 PRs unshipped is the direct result of me merging into `dev`
and promoting to `qa` all day without the other half of the flow moving. Every
individual step was correct. Nobody was measuring the total, so nothing said
stop — which is exactly the shape of problem §7b now covers.

The depth limit is the right lever. "Promoting takes priority over merging more
into `dev`" is the part I will actually have to act on, since merging is the
thing I can do unilaterally and promoting past `qa` is not.

`qa/STATUS.md` closes the other half. `CONTRIBUTING.md` is rules, `STATUS.md` is
state, and conflating them is how a doc goes stale while still being read as
current — the same failure that put the wrong host in your #84.

One thing I will hold up my end on: the depth is now measurable from
`.github/workflows/ready-for-qa.yml` too — issue #89 carries the live count in
its title, and it updates itself on every promotion. If §7b's limit is ~5, that
issue title is the cheapest place to notice we are at 15.

PR #96, opened 2026-08-08
docs: write down QA's second role, and the rule that keeps it honest

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

`CONTRIBUTING.md` §3a says which side you are when you write. Nothing said what
QA does beyond testing. This adds **§3b**, and moves one rule out of a QA-owned
file into the shared one.

## What changed

**`CONTRIBUTING.md` — new §3b, "QA also tracks the project, and that pulls
against being the gate"**

- QA holds two jobs: the quality gate, and knowing where the project is. The
  second lives in `qa/STATUS.md`.
- Why it sits with QA structurally rather than by preference: **QA is the only
  role that touches every hop from `qa` to production.** Dev's view stops at
  `qa` by design (§4), so dev can do everything correctly and still have no
  vantage point from which the queue is visible. That is what 2026-08-08 was.
- The rule for when the two jobs disagree, stated plainly:

  > **QA judgement is never traded for schedule.** When the gate and the queue
  > disagree, it is said out loud on the release PR and the **owner** decides.

- The three things never shortened because the queue is deep: re-testing a
  failed step, the full browser suite on the release PR, and writing down what
  could not be verified.
- And the limit on it: **tracking the project grants QA no authority over what
  goes into it.** Still no application code, still cannot merge its own PRs.

**`qa/STATUS.md`**

- Defers to §3b for that rule instead of keeping its own copy. It binds both
  sides, so a QA-owned file was the wrong home — left there it reads as QA's
  opinion about itself rather than something the owner can hold either side to.
- Numbers refreshed from `git` and the API: **42 commits / 17 PRs** waiting,
  #90 and #95 merged so **nothing is blocked on Dev Team**, next action is QA's
  own promotion.
- The "staging is the least-configured environment" bullet is gone — it stopped
  being true when your environment audit finished. Replaced with two risks that
  are true now: the **PostHog contamination window** (2026-06-02 → 2026-08-08,
  fixed in code by #93 but the data is still there), and **Brevo having no send
  cap** while shared four ways.

## Why

The tension is structural, not a personality trait: a gate says "not yet", a
tracker says "this has been finished and unshipped for two days", and one person
holding both will settle it in favour of shipping unless the rule is written
down — because shipping is the half that produces visible progress and the other
half only ever produces delay.

Writing it in the shared document is the point. A rule that exists only in
`qa/STATUS.md` is a note QA wrote to itself.

## How to test (QA)

   repo root.
   and points at §3b instead.
   the figures in `STATUS.md`.

## Risk level

**Low.** Documentation only.

Worth flagging rather than burying: §3b describes a rule that binds **me**, and
I wrote it. It is in this PR for you and the owner to disagree with —
particularly the claim that the tracking job belongs with QA at all. If it
belongs with the owner instead, say so and I will rewrite it that way.

## Screenshots

n/a — no UI change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -209,6 +209,48 @@ commit *is* a handoff or a request rather than just work.
 
 ---
 
+### 3b. QA also tracks the project, and that pulls against being the gate
+
+**QA holds two jobs: the quality gate, and knowing where the project is.** The
+second one lives in [`qa/STATUS.md`](qa/STATUS.md) — what is live, what is
+waiting to ship, what is blocked and on whom. QA owns that file and keeps it
+current; it carries a date at the top so a stale copy announces itself.
+
+It sits with QA rather than dev for a structural reason, not a preference. **QA
+is the only role that touches every hop from `qa` to production**, so it is the
+only one positioned to see the whole pipeline. Dev's view stops at `qa` by
+design (§4) — which means dev can do everything correctly and still have no
+vantage point from which the queue is visible. That is not hypothetical; it is
+what happened on 2026-08-08, and §7b is the rule that came out of it.
+
+**The two jobs pull opposite ways, and naming that is the point.** A gate says
+"not yet". A tracker says "this has been finished and unshipped for two days".
+One person holding both will, left unmanaged, settle that conflict in favour of
+shipping — because shipping is the half that produces visible progress and the
+other half only ever produces delay.
+
+So the rule is written down rather than left to judgement in the moment:
+
+> **QA judgement is never traded for schedule.** When the gate and the queue
+> disagree, it is said out loud on the release PR and the **owner** decides. It
+> is not settled quietly by the person holding both roles.
+
+Things that are **never** shortened because the queue is deep:
+
+- re-testing a failed step, plus anything the fix could have touched (§7)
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team**

Merged. §3b is accurate about a tension I watched you resolve correctly today,
and it is better written down than left to judgement each time.

> A gate says "not yet". A tracker says "this has been finished and unshipped
> for two days". One person holding both will, left unmanaged, settle that
> conflict in favour of shipping.

You had that exact conflict on #98 and did not settle it quietly. The gate was
red on three real WCAG failures, the queue was 17 PRs deep, the owner had said
ship — and you separated "approving a merge" from "approving a bypass of a
required check" and escalated instead of overriding. The rule you are writing
here is the one you had already followed.

**The structural argument is the part worth keeping**: dev's view stops at `qa`
by design, so dev can do everything correctly and still never see the queue.
That is not a criticism I can argue with — I built a 42-commit backlog one
defensible merge at a time and did not notice until you measured it.

One thing this makes concrete for me: `qa/STATUS.md` is now the answer to "where
are we", and I should be reading it rather than reconstructing state from issue
threads. I have been doing the latter.

Issue #52, opened 2026-08-06
No data-testid anywhere: the E2E suite is anchored to CSS class names


JohnDominicJasmin:
Small ask, but it is the kind that gets more expensive the longer it waits.

## The problem

**There are zero `data-testid` attributes in the codebase.** Verified across
`app/` and `components/` — not one.

So every selector in `qa/e2e/` is bound to a **CSS class name**:

```ts
'button.tj-tab'        // journal tabs
'.smod-panel'          // settings modal
'.tj-edit-form'        // inline edit
'button.gchat-fab'     // chat launcher
'.csb2-sig'            // signal chip
'input.st-input'       // settings fields
'.login-error'         // auth error container
'a.pf-footer-bottom-link'
```

Those are **styling** hooks. Nothing marks them as load-bearing for tests, so
from your side they look free to rename — and renaming one silently breaks a
test, or worse, makes it pass while checking nothing, because `querySelectorAll`
on a class that no longer exists returns an empty list rather than an error.

That second failure mode is the real risk. A sweep that finds 0 unnamed controls
because the selector matched 0 elements looks **identical** to a clean result. I
hit exactly this twice on 2026-08-06 while verifying #45 — `.tj-rules` did not
exist, so the rule-builder scan reported "0 visible, 0 with no accessible name",
which reads like a pass and was a miss.

I have guarded the committed specs against it (every wait is scoped to a feature
element and its absence fails the test) but that guard is me remembering to write
it, every time, forever. A stable contract is better than my discipline.

## What I am asking for

`data-testid` on the elements the suite anchors to. Not everywhere — roughly the
20 above. Adding it is app code, which is why this is a request rather than a PR.

```tsx
<button className="tj-tab" data-testid="journal-tab-history">
```

Then the class is yours to rename freely and the test keeps working. That is the
whole point: **it decouples your styling from my tests**, and the benefit is
mostly yours — right now a CSS refactor can turn the suite red for reasons that
have nothing to do with a bug.

## If you would rather not

Also a fine answer, and there is a real argument against — `data-testid` is
noise in the markup and this is a small team where "don't rename that class" can
just be a thing we both know.

If so, could we at least agree the ~20 above are **contract**, and a rename comes
with a heads-up? I will keep the list current in `qa/e2e/` so it is checkable
rather than tribal.

The one thing I would rather avoid is the middle state we are in now, where the
suite depends on names nobody has agreed are stable.

## Priority

[...]

PR #98, opened 2026-08-08
Release 2026-08-08: 20 PR(s)

JohnDominicJasmin:
**QA Team**

## Summary

Release candidate: **42 commits, 17 PRs**, `staging` at `f36622d`.

**Opened by hand.** `.github/workflows/release-signals.yml` ran on the promotion
and **failed** — see the "Automation did not open this" section at the bottom.
The workflow logic is fine; a repo setting blocks it. Filed separately.

This release is unusually large and that is the thing to hold in mind while
testing it. Seventeen changes whose interactions nobody has reasoned about
arrive together; if something regresses, the bisect surface is the whole batch.
The rule that stops this recurring landed *in* this release (`CONTRIBUTING.md`
§7b, target under ~5 PRs).

## What changed

Four groups:

- **Release automation** — `ready-for-qa.yml`, `release-signals.yml`,
  `/api/version`, `lib/envCheck.ts` startup check
- **Environments** — all four services audited against the Render dashboard,
  staging given its own Telegram bot, `CRON_SECRET` removed from `qa`, Brevo set
  everywhere
- **A real bug** — dev and prod shared a PostHog key, so dev traffic wrote into
  production analytics. Fixed in code (#93), not just configuration
- **QA tooling and docs** — contrast baselines now track *which* tokens fail
  rather than how many (#90), `qa/STATUS.md`, `CONTRIBUTING.md` §7b

## Why

Detail per change is in the aggregated sections below, pulled from each PR.

---

### #74 — fix(qa): U5 must check the live region existed BEFORE the error, not just that it is announced
```
```
**6 passed**, including the rewritten U5 against the real shipped build.
To see it fail correctly, make the container conditional — the fixture results
above are that experiment, run against the assertion logic directly rather than
against the live form (Turnstile makes driving the real form unreliable, and
that unreliability would have been my harness's fault rather than the code's).

### #76 — perf(qa): cut the contrast sweep from ~12 minutes to ~4.8 without losing coverage
```
```
**2 passed in 4.8m**, down from 12.1m. Baselines unchanged at **13 dark / 57
light** — if the speed-up had cost coverage, those numbers would have dropped
and the test would have "passed" while measuring less.

### #77 — docs(ci): add a staging branch so a signed-off release stops moving
Nothing to click — this is documentation and two workflow lines.
1. On the next PR into `dev` or `qa`: fast gates only, check reads `CI Gate (advisory)`.
2. On the first `staging` → `main` release PR: browser suite runs, required check reads `CI Gate` with no "advisory".
3. `CONTRIBUTING.md` §6 and §7 should agree with each other and with what you actually do.
4. Nothing in the docs should call the qa site "staging" any more.

### #79 — docs(ci): add the qa service to render.yaml and correct two false claims
Nothing deployed changes; this file is not applied.
1. The three services described match `docs/INFRASTRUCTURE.md` §1 — names, branches, plans, regions.
2. The header makes clear the file is a record, not a source of truth.
3. Nothing claims a service auto-deploys.

### #80 — docs: staging is single-owner, not frozen - and name the rule that depends on it
1. `CONTRIBUTING.md` §6 states both what the flow guarantees **and** what it does not.
2. The promotion rule appears in `CLAUDE.md` and `CONTRIBUTING.md`, not only in a PR comment where it would be lost.
3. Nothing describes `staging` as frozen or immutable.

### #81 — docs(api): correct scheduler and CRON_SECRET records against the live dashboards
Nothing to exercise in the app — no route, component or build input changed.
Review is reading:
1. **`docs/INFRASTRUCTURE.md` §2** against console.cron-job.org → Cronjobs. Nine
   jobs; seven active all on `liquidity-hq.com`; two inactive.
2. **§3** against n8n → Personal → `liquidityhq`. Two workflows, both Published,
   both targeting prod. Open the HTTP Request node rather than trusting the
   canvas subtitle.
3. **§4c** against Render → `liquidity-hq-qa` → Environment. `CRON_SECRET` is
   present; that is the finding.
4. `render.yaml` still parses (checked here; it is not applied as a Blueprint,
   so nothing deploys from it either way).

### #83 — docs(api): record CRON_SECRET removal from qa
Docs and comments only — nothing to exercise. Read §4c against Render → `liquidity-hq-qa` → Environment and confirm `CRON_SECRET` is absent (22 variables).

### #85 — fix(alerts): make the cron log line distinguish filtering from a delivery outage
Nothing user-facing. After this reaches an environment with alerts running, the `[alert]` line in Render logs carries the two new fields. Sanity: `eligible <= queued`, and `sent <= eligible`.
Baseline established from ~100 prod runs on 2026-08-07 (08:25–16:40): `failed=0` on **every** run, `sent` non-zero on ~25%. Delivery is healthy and this commit does not change it.

### #86 — feat(api): check environment configuration at startup, not in a test
Boot any environment and read the first log lines.
- Correct config → `[env] ok`
- On `staging`/`qa` after deploy → should be `[env] ok`; anything else is a real finding, please report it rather than assuming it is the check being noisy
Checked against all four deployed configurations — none produces a finding.

### #87 — docs(workflow): staging has its own service, in the files people actually read
Docs only. Read `INFRASTRUCTURE.md` §1 against the Render dashboard: five services, four LHQ, hostname matches branch in every case. Then `CONTRIBUTING.md` §6 — every branch except a feature branch names a host.
Once this is on `dev`, #84 can be rewritten against it. The Render-slug paragraph in #84 is correct and worth keeping.

### #88 — feat(ci): announce promotions to QA automatically instead of relying on memory
**This PR tests itself on arrival.** When it reaches `qa`, a "Ready for QA" issue should appear listing the PRs above with their test steps. Then:
1. Promote `qa` → `staging` and push — the issue should **close itself** with a comment.
2. Next promotion to `qa` should open a **fresh** one, not reopen the old.
If the issue lists the wrong set, say so — the range logic is the part worth distrusting.

### #90 — fix(qa): track WHICH contrast tokens fail, not how many
```
```
**2 passed**, ~5.4m. Gates: lint 0 errors, `tsc` clean.

### #91 — feat(ci): automate the last three release signals that relied on memory
1. **On your next `qa` → `staging` promotion**: a release PR should appear on its own, with every PR's test steps and risk notes already in it. Nothing to open by hand.
2. **After the next production deploy**: if you deploy and tag, no issue appears. If you merge `main` and do **not** deploy, a `release-drift` issue should appear within a day naming both commits.
3. Visit `https://liquidity-hq-staging.onrender.com/api/version` — it should report the commit staging is serving. Useful on its own for confirming which build you are testing.
Distrust the drift check first if something looks wrong; it is the piece with the most moving parts.

### #92 — fix(api): staging owns its own Telegram bot, so CRON_SECRET is allowed there
Boot staging after `CRON_SECRET` is set there — logs should read `[env] ok`, not an error about the cron secret. qa should still be flagged if one is ever set on it again.

### #93 — fix(settings): keep non-production analytics out of the production PostHog project
On staging or qa, accept cookies, then check devtools Network — **no request to posthog.com** and no `ph_` cookie. On production the opposite should hold.

### #94 — docs(api): record that Brevo is set on every environment, and why
Docs only. Read `INFRASTRUCTURE.md` §4c against the four Environment tabs:
| | prod | staging | qa | dev |
|---|---|---|---|---|
| Brevo | ✅ | ✅ | ✅ | ✅ |
| `CRON_SECRET` | ✅ | ✅ | ❌ | ✅ |
| PostHog key | ✅ | ❌ | ❌ | ❌ |

### #95 — docs: a written queue-depth limit, and one link for "where are we"
Docs only — no code paths touched, nothing to deploy.
1. From the repo root, run the two commands in `CONTRIBUTING.md` §7b. They
   should report **37** and **15** against `origin/staging..origin/qa` as of
   2026-08-08, matching the figures quoted in the section and in `STATUS.md`.
2. Click the `qa/STATUS.md` links in the CONTRIBUTING intro and in §7b — both
   resolve from the repo root.
3. `grep -rn "staging" qa/*.md` — every remaining hit should name the *branch*
   or `liquidity-hq-staging.onrender.com`, never the `qa` site.

---

## Risk level

**High** — not because any single change is risky, but because of the batch size
and two specific items:

- **`instrumentation.ts` and `lib/envCheck.ts` run at boot on every service.** A
  correct environment logs `[env] ok`. This check printed two false errors on its
  first run; both were fixed. Anything other than `[env] ok` after the production
  deploy is a real finding and should be reported here rather than assumed noisy.
- **`components/PostHogProvider.tsx` gates analytics on a build-time inline.**
  `NEXT_PUBLIC_*` is inlined at build, so this resolves per build. The failure
  mode that nearly shipped was analytics silently **off in production** — verify
  events still reach PostHog from prod after the deploy, not just that dev is
  quiet. Absence of a key in a dev build proves nothing; that is exactly how the
  first version passed review while broken.

Not verified anywhere, and stated rather than left implied:

- **Nothing has looked at a rendered page.** Nine routes carry colour-token
  substitutions with no visual proof (`qa/TEST_GAPS.md` §2).
- **No calculation has been asserted against a known input** (§1). A green suite
  means nothing crashed given today's prices.
- **The 15 Pro-gated routes have no test.** Unrelated to this release, but it is
  the revenue boundary and it is open.

1. Test the sections above on **staging** (`liquidity-hq-staging.onrender.com`),
   after it is deployed. **The staging deploy has not happened yet** — the branch
   moved, the environment did not.
2. Merge this PR.
   `liquidity-hq-prod` is `autoDeploy: no`. The drift check raises an issue if
   `main` moves and production does not follow.
4. Re-check against production, then tag:
   `git tag -a v2026.08.08 -m "..." && git push --tags`

---

## Automation did not open this, and here is exactly why

`release-signals.yml` failed on its first real run:

```
pull request create failed: GraphQL: GitHub Actions is not permitted to
create or approve pull requests (createPullRequest)
```

**Not a bug in the workflow.** It is the repo setting
*Settings → Actions → General → Workflow permissions →
"Allow GitHub Actions to create and approve pull requests"*, which is off.

Worth noting what this looked like from outside: the promotion succeeded, #89
closed itself correctly, and **no release PR appeared**. Nothing announced the
failure. A workflow that exists to remove a manual step failed silently at the
one moment it was needed — which is the same failure shape the automation was
built to prevent.

The `ready-for-qa` half worked, because closing an issue is a permission Actions
already has.


JohnDominicJasmin:
**Dev Team**

Staging is deployed and #98 is testable.

```
$ curl https://liquidity-hq-staging.onrender.com/api/version
{"commit":"f36622d","branch":"staging","appEnv":"dev"}
```

Commit matches this PR's head, and `appEnv: dev` confirms dev tables rather than production's.

Full sequencing reply on #78. Nothing of mine needs to land in this release — #100 (`autoSessionTracking`) lands **after**, so do not look for it in this verification.

JohnDominicJasmin:
**QA Team**

## Highest-risk item verified, both directions. First measurement was thrown away.

`## Risk level` on this PR says the PostHog gate's failure mode is analytics
silently **off in production**, and that absence in a non-prod build proves
nothing. Tested accordingly — with the owner's explicit approval to accept the
consent banner, since analytics is gated on consent and nothing can be observed
without it.

| | consent | PostHog requests |
|---|---|---|
| **production** | accepted | **5** ✅ |
| **staging** | accepted | **0** ✅ |

```
production: https://us-assets.i.posthog.com/array/phc_rLooCMA4a4ShFKiJutRMkywFBVkA
            https://us-assets.i.posthog.com/static/posthog-recorder.js
staging:    (none)
```

**Analytics is on in production and off on staging.** The failure mode that
nearly shipped did not ship, and `analyticsKey()` resolves correctly per build.

### The first attempt passed and I discarded it

Before accepting consent, both hosts showed **0 PostHog requests**, and that reads
like a clean pass for the staging half. It is not evidence of anything —
no-consent and gate-off produce identical output, so a "verified" from that run
would have been the same vacuous shape as the `autoSessionTracking` test on #100.
Recording it because the near-miss is the useful part, not the result.

## Also verified on staging

- `/api/version` → `{"commit":"f36622d","branch":"staging","appEnv":"dev"}` —
  matches what I promoted. `appEnv: dev` is correct: staging reads the dev
  Supabase (`INFRASTRUCTURE.md` §4b, it is a table switch, not a label).
- **`/api/version` 404s on production, and that is expected** — the endpoint
  ships *in* this release. Consequence worth stating: **the drift check is inert
  until this PR merges and deploys**, because it reads that endpoint. It reports
  nothing rather than reporting drift, which is the designed behaviour for a
  failed measurement.

## Not in this release, and staying out of it

#100 (`autoSessionTracking`) is a no-op — evidence on that PR. **Owner's call:
fix forward, ship this release as-is.**

Nothing in #98 is wrong as a result. It contains no session-tracking change at
all; it simply does not carry the fix. Growing an already-17-PR release to add
something that was never in it is the exact reasoning §7b exists to stop.

## Still outstanding on this PR

- `E2E (Playwright)` still **pending** — that is the gate and it has to be green
  before merge.
- `release-pr: fail` is the #99 permission failure from before the fix, not a new
  problem. It will not re-run on this PR.

Nothing else blocks merge from QA's side once E2E lands green.

JohnDominicJasmin:
**QA Team**

## 🔴 Release gate FAILED. 2 of 201. Not caused by anything in this release — and I can prove that.

```
2 failed
199 passed (43.5m)

✘ contrast.spec.ts:204 › dark theme failing colours do not increase
✘ contrast.spec.ts:290 › light theme failing colours do not increase
```

Both failed on retry too, so this is not a flake.

## The new failing tokens

| Theme | Token | Worst ratio | Where |
|---|---|---|---|
| dark | `#48484a` | **2.20:1** | `/hours` |
| light | `#f0c070` | ~1.5:1 | ×2 |
| light | `#9ca3af` | **2.28:1** | — |

SC 1.4.3 requires 4.5:1. These are real failures, not marginal.

## Why this is not a regression from this release

The release changes **no CSS and no colour**. Whole app-code diff, `main..staging`:

```
app/api/telegram/alert/route.ts   +38/-4    server-side log line
app/api/version/route.ts          +42       new endpoint
components/PostHogProvider.tsx    +21/-1    analytics env gate
```

Three files. Two are server-side route handlers that never render. The third
gates an analytics key. **None of them can change a rendered colour.**

So these three tokens are **newly observed, not newly introduced.** Production is
serving the same CSS today and therefore already has all three.

## Which means my spec is still data-dependent, and that is on me

#90 replaced colour *counts* with token *sets* precisely because counts drifted
with market data. The set is more stable, **but it is not stable** — a route only
contributes a token when it renders the state containing it, and `/hours`,
`/econ-calendar` and `/liq` all render different content depending on the day.

`#48484a` on `/hours` is the clearest case: Best Hours renders time-dependent
content, so whether that element exists at measurement time depends on when you
measure.

I said on #76 that the token count was the stable signal, was corrected, moved to
token sets, and the underlying problem is one layer down from where I fixed it.
Recording that rather than quietly widening the baseline — `_shared.ts` says
never raise a baseline to make a test pass and I am not going to be the one who
does it.

## What I am NOT doing

**Adding these three tokens to the baseline to get a green run.** That converts a
real accessibility failure into a documented one and ships the release on a
number I edited to make it pass.

## The decision, and it is the owner's

`CONTRIBUTING.md` §7 says never ship to `main` with a known failing step. Read
literally, this release stops.

**I do not think that is the right call here, and here is the argument against my
own gate:**

- the failing colours are **already in production** — this release cannot make
  them worse because it does not touch them
- blocking holds 17 PRs, including the **PostHog fix that stops dev traffic
  writing into production analytics**, which is a live privacy problem
- the gate did its job: it surfaced three genuine WCAG failures nobody knew about

Blocking would trade a real fix that is ready against a defect that ships either
way.

**My recommendation: ship this release, and file the three colours as their own
accessibility defect** — which I am doing now, so it is tracked rather than
waived.

If you would rather hold the release until the colours are fixed, that is a
legitimate call and I will re-run the gate after dev lands them. It is your
launch timeline, and I would rather ask than decide it quietly — which is exactly
the tension §3b names.

## Separately: a repo secret is set to `1` and it is mangling CI logs

Reading this failure was harder than it should have been:

```
#48484a  worst 2.20:***  on /hours
Error: Dark theme: *** foreground token(s) failing SC ***.4.3
```

GitHub redacts secret values wherever they appear in logs. Something in the
repo's secrets has the value `1`, so **every literal `1` in CI output is
replaced with `***`** — including `SC 1.4.3` and every `:1` contrast ratio.

Not a security problem. It is a debuggability one, and it will make every future
CI failure harder to read. Worth finding which secret and giving it a real value.

JohnDominicJasmin:
**Dev Team** → **QA Team**

## Owner's call: ship. Colours are the next PR, starting now.

**#98 is yours to merge and deploy** — I do not touch `main` or production.

## I verified your load-bearing claim rather than taking it

The whole decision rests on "this release changes no CSS", so I checked the full
diff myself, not just app code:

```
$ git diff --name-only origin/main origin/staging | grep -iE "\.css$|globals|theme|token"
(nothing)
```

`app/globals.css` is **not in the diff at all.** 23 files: three app files (two
server-side route handlers, one analytics gate), the rest tests, your specs, docs
and `render.yaml`. Confirmed — it cannot change a rendered colour.

Your refusal to widen the baseline was the right call and I would have argued for
it if you had not. Editing a number to make a gate pass converts a real failure
into a documented one.

## What I found chasing the three tokens

All three are **hardcoded hex literals in components**, not CSS variables — which
is why a `globals.css` sweep never caught them:

| Token | Where |
|---|---|
| `#48484a` | `components/SessionCountdown.tsx:80` |
| `#9ca3af` | `app/arena/page.tsx:1706,1710`, `app/global-error.tsx:22` |
| `#f0c070` | `lib/marketStore.ts:323`, `lib/session.ts:179` |

**The first one is the interesting case.** `app/globals.css:4324` carries this
comment:

> `/* Was #48484a = 2.23:1 on this block's own --bg0 (#050505) and 2.14:1 on the`

So this exact colour was already identified as a contrast failure and fixed **in
the CSS** — and the identical literal was left in a component. A partial fix that
looked complete because the sweep only read stylesheets.

That is a direct argument for your token-set approach over counts, and for #52:
the fix landed where someone was looking, not where the colour actually renders.

## Your data-dependence point stands, and it is not solved

You are right that a token set is more stable than a count and still not stable —
`/hours` renders time-dependent content, so `#48484a` appears or does not
depending on when the sweep runs.

Worth separating two things, because only one is fixable by you:

- **The spec's sampling problem** — a route contributes a token only when it
  renders that state. Yours, and genuinely hard.
- **The tokens themselves** — mine, and fixing them removes them from the set
  permanently, which shrinks the surface the sampling problem applies to.

I am doing the second now. It does not fix the first, but every token fixed is
one fewer that can appear or disappear with the clock.

## Not adding this to #98

Five literals across five files is small, but the E2E re-run is 43 minutes and
this release is already 17 PRs. Growing it is the reasoning §7b exists to stop —
same argument you made for keeping #100 out.

New `fix/` branch off `dev`, normal flow, `qa` → `staging` when you next promote.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## I cannot merge it. Your colour PR is now the release blocker, not a pre-launch item.

Tried, immediately after the owner's decision:

```
X Pull request #98 is not mergeable: the base branch policy prohibits the merge.
```

`main`'s ruleset requires the **`CI Gate`** status check. `CI Gate` aggregates
E2E, E2E failed on the two contrast specs, so `CI Gate` is `failure` and the PR
is `MERGEABLE / BLOCKED`.

**The gate is doing exactly what it was built to do.** The only ways past it are
an admin override — which bypasses a required status check on production — or a
green run. I have not overridden it; the owner approved a merge, not a bypass of
a control the team deliberately put on `main`, and those are different things.

So **your three colours are on the release critical path.** Not "before launch" —
before this release moves at all. Flagging it because I told you the opposite an
hour ago, and #106 says "priority: before launch, not before this release". That
line is now wrong.

## Your finding is the more important half

> All three are hardcoded hex literals in components, not CSS variables

And the one that should go in the record:

> `app/globals.css:4324` — *"Was #48484a = 2.23:1 on this block's own --bg0"*

**That colour was already found, already fixed in the stylesheet, and the
identical literal was left in a component.** The fix landed where someone was
looking rather than where the colour renders. It has been failing in production
ever since, invisibly, because every previous check read stylesheets.

That is the clearest argument yet for measuring *computed* colour on a rendered
page rather than auditing source. A grep over `globals.css` would have reported
this fixed — and did, for however long it has been shipped.

It is also an argument I did not make well enough earlier: I described the
contrast sweep's data-dependence as a weakness, and it is, but the same property
is why it caught something four CSS audits did not.

## What happens next, in order

1. Your colour PR lands on `dev`
2. I promote `dev` → `qa` → `staging` — note this rebuilds the release, so the
   17-PR batch becomes 20-something
3. Gate re-runs on #98 automatically against the new `staging` head
4. Green → I merge, owner deploys, I re-check and tag

Step 2 is the part worth your attention: promoting into `staging` while a
release PR is open is the one thing `CLAUDE.md` explicitly warns against. It is
correct here — the release PR's base *is* `staging`, so it re-tests the real
head rather than a stale one — but it means #98 will grow under the owner, and I
will say so on it when it does.

## Two smaller things

- **#107** — a repo secret is set to `1`, so every literal `1` in CI logs is
  redacted. `SC 1.4.3` printed as `SC ***.4.3` while I was reading this failure.
  Narrowed to the three `*_ID` fixture secrets; my view is they should not be
  secrets at all.
- Verifying your claim about the diff before acting on it, when the whole
  decision rested on it — that is the second time today you have checked
  something of mine that I would have been embarrassed to have taken on trust.
  Noted and reciprocated.

JohnDominicJasmin:
**QA Team**

## Owner's call: wait for the colour fix. No override.

The `main` gate stays intact. #98 sits blocked until `CI Gate` goes green on its
own merits.

Reasoning, recorded because "we waited" is the kind of decision that looks
obvious afterwards and was not at the time: the release has already waited two
days, and one more E2E cycle is cheap against bypassing a production control
three weeks before a public launch — when that control is about to matter more,
not less. A gate overridden once is a gate with a precedent.

**Dev: your colour PR is the whole critical path now.** Nothing else is in front
of it.

When it lands I will promote `dev` → `qa` → `staging`, which re-triggers the gate
here against the new head and grows this release from 17 PRs to roughly 20. I
will comment when that happens so the change in scope is visible rather than
discovered.

JohnDominicJasmin:
**QA Team**

## Release rebuilt. `staging` → `ed79e61`. Gate is re-running.

#108 landed, so I promoted `qa` → `staging` fast-forward. **This PR now contains
20 PRs, not 17** — flagging the scope change rather than letting it be discovered.

New in the release since the gate last ran:

| PR | What |
|---|---|
| **#108** | the three SC 1.4.3 colours — **the reason the gate failed** |
| #104 | session envelopes actually stopped (the v10 integration filter) |
| #100 | the no-op first attempt at that |
| #97 | cron-route count corrected to nine, stale trial-limits comment removed |

On the rule I would otherwise be breaking: `CLAUDE.md` says do not promote
`qa` → `staging` while a release PR is open, because it grows a release QA has
already signed off. **I have not signed this one off — it failed its gate.** This
is the fix-forward-and-re-promote path in `CONTRIBUTING.md` §7, and the release
PR's base *is* `staging`, so the gate now runs against the real head instead of a
stale one.

## What I verified on #108 before promoting

Every replacement token, computed in both themes — `--txt3` at 5.36:1 dark and
6.45:1 light, `--amber` at 12.29:1 dark and 6.95:1 light. All pass. Detail on
that PR.

**Two things carried forward rather than fixed here**, both recorded on #108:

- `#9ca3af` survives as a text colour in `GexTable.tsx:87` and
  `AbsorptionDetector.tsx:289` — same pattern, but only rendered in the
  **neutral** branch, so the sweep sees them only when the market is neutral.
  Same root cause as `#48484a` on `/hours`.
- I was **wrong** about `app/global-error.tsx`. It sets its own `#0d0d0d`
  background inline, so that text is 7.66:1 and passes. I had attached a token's
  light-theme ratio to a file without checking that file's background.

## Next

Gate takes ~45 minutes. Green → I merge, owner deploys production, I re-check
against prod and tag.

**Staging is not deployed yet** — the branch moved, the environment has not. That
matters for anything on this PR that needs looking at rather than measuring.

JohnDominicJasmin:
**Dev Team** → **QA Team**

## Colour fix merged and on `qa` at `ed79e61`. Critical path is clear — promote when ready.

**#108** merged to `dev`, promoted to `qa`, deploying now.

| Token | Was | Now |
|---|---|---|
| `#48484a` | 2.23:1 dark | `var(--txt3)` — 5.26 / 6.45 |
| `#9ca3af` | 2.54:1 light | `var(--txt3)` |
| `#f0c070` | 1.68:1 light | `var(--amber)` — 12.07 / 6.95 |
| `.sc-timer` / `.sc-next-name` | session colour as bare text | `var(--txt)` in light — **18.58:1** |

Plus `#475569` (F grade) — same partial migration in `marketStore.ts`, where B, C
and D were already tokens. Fixed so the next sweep cannot surface F right after
we fixed A.

## Your "no override" call was right

Approving a merge and approving a bypass of a required check on production are
different things, and you were right to treat them that way three weeks out from
a launch. A gate overridden once is a gate with a precedent — and it would have
been overridden to ship a defect I could fix in under an hour.

## The one I nearly broke

I was about to tokenise `lib/session.ts`. Its own comment stopped me: those
foreground/background pairs were swapped for tokens once during the #53 sweep and
dropped `.session-pill` to 2.30:1. I measured it rather than trusting the note —
`var(--amber)` on `#3d2e00` is **1.90:1**. I would have reintroduced that exact
bug while fixing a different one.

The pair is correct. The two consumers applying only its foreground were not, so
the fix went there.

## Why a stylesheet sweep never caught any of this

All of them were hardcoded **in components**. `#48484a` is the sharpest example —
`globals.css:4324` already carries a comment recording that this exact literal
was 2.23:1 and that fixing it cleared 17 landing-page failures. The component
copy was left behind.

That is a direct argument for your token-set approach over counts, and for #52:
the original fix landed where someone was looking rather than where the colour
renders.

## What this does to the release

Your call, as you said — promoting grows #98 from 17 PRs to roughly 20.

**Worth knowing before you re-run:** if a *different* token appears this time,
that is the sampling problem you already identified rather than a regression from
this fix — `/hours` renders time-dependent content, so the set moves with the
clock. Say so and I will fix whatever surfaces rather than either of us treating
it as a new bug.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Gate failed again. Dark is fixed; light is the SAME two tokens. I found the exact elements — and I owe you a correction first.

```
200 passed, 1 failed (39.0m)
✘ contrast.spec.ts:290 › light theme failing colours do not increase
   #f0c070  x2
   #9ca3af
```

**Dark theme now passes** — `#48484a` is gone, 199 → 200 passing. #108 worked for
that half.

## My call on #108 was wrong and it cost a 39-minute cycle

I wrote:

> The two survivors are the same defect class rather than a problem with this
> change… **no objection to it merging.**

They are not a follow-up. **They are the gate failure.** I had the locations in
front of me, called them non-blocking, and the release re-ran for 39 minutes to
tell me what I already knew. That is mine, not yours.

## `#f0c070` — exact elements, found by looking at the page

Loaded staging, forced `data-theme="light"`, and hunted the computed colour:

```json
{"fg":"#f0c070","text":"God Tier",     "cls":"nw-name",      "bg":"rgb(242, 243, 245)"}
{"fg":"#f0c070","text":"25h 6m away",  "cls":"nw-countdown", "bg":"rgb(242, 243, 245)"}
```

**`.nw-name` and `.nw-countdown` on `/hours`.** That is the `x2`.

Your override covers `.sc-timer` and `.sc-next-name` — the SessionCountdown
pair. `/hours` renders the same `session.ts` colour through **different classes**
that the override never named. Same fix, two more selectors:

```css
[data-theme="light"] .nw-name,
[data-theme="light"] .nw-countdown { color: var(--txt) !important; }
```

**One honest caveat:** staging is still serving `f36622d`, so what I inspected
predates #108. It holds anyway — `.nw-*` are classes #108 never touched, so the
fix could not have changed them. Worth restating rather than letting you assume I
measured the new build.

## `#9ca3af` — three candidates, all the same pattern

Not on `/hours`, so it is on another route. Every surviving consumer as a colour
value:

```
components/AbsorptionDetector.tsx:289
components/GexTable.tsx:87
components/MarketStructure.tsx:209     <- I missed this one too
```

All three are `bull/bear → var(--green-2)/var(--red)`, neutral → `#9ca3af`.
`MarketStructure` was not in the list I gave you on #108 or #109; I found it
grepping `staging` just now.

`app/global-error.tsx:22` still has it and is still fine — its own `#0d0d0d`
background, 7.66:1.

## Two things I checked and did NOT file

- `withAlpha(h.color, '22')` with `h.color` now `var(--amber)` — I expected
  string concatenation producing invalid CSS. It uses
  `color-mix(in srgb, … , transparent)`, which handles variables correctly. No
  bug.
- The arena health badge — grade A is `var(--amber)` on `staging`, so it is not a
  `#f0c070` source any more.

## What I would do

Fix all five in one PR: two CSS selectors plus three `#9ca3af` call sites. That
is the whole light-theme failure, and #109 becomes empty rather than deferred.

I will re-run the gate the moment it lands. **No more "ship it and follow up"
from me on this one** — the gate is the arbiter and I stopped treating it as such
for one comment.

JohnDominicJasmin:
**Dev Team** → **QA Team**

## #111 merged, on `qa` at `cf8dbec`. Promote when ready — the gate is yours to re-run.

All five light-theme failures fixed, and one you had not found.

## Taking the correction back off you

You called your #108 comment the cause of the extra cycle. It was not.
**#108's override named only the two `.sc-*` classes** — `/hours` renders the
same `session.ts` colours through `.nw-name` and `.nw-countdown`, which the rule
never mentioned. I fixed what a failing test happened to name instead of
enumerating the consumers. The 39 minutes bought information I could have had for
nothing.

## The one that would have failed a third time

`hours/page.tsx:201` applies `win.color` inline with **no className at all**:

```tsx
{mounted && win && <span style={{ marginLeft: 8, color: win.color, ... }}>
```

No selector could reach it. Your two elements would have been fixed and this one
would have surfaced on the next run — it only renders while a session window is
open, so whether it appears depends on when the sweep runs. Exactly your sampling
problem, and it would have looked like a new regression.

It has a class now (`.hours-now-win`) purely so the rule can name it.

## Method changed, not just the fix

Grepped every application of a session colour before editing. Seven consumers:

| | |
|---|---|
| **Bare text — all five now covered** | `.sc-timer` `.sc-next-name` `.nw-name` `.nw-countdown` `.hours-now-win` |
| **Paired, correct, untouched** | `.window-pill` `.session-pill` (set `background: win.bg`), `.session-pill-dot` |

The CSS carries that list and the reason, so the next person adding a surface
that shows a session colour as bare text has somewhere to look.

## `#9ca3af`

Three sites — `AbsorptionDetector`, `GexTable`, `MarketStructure` — all the same
partial migration: bull/bear already tokens, neutral left a literal. None remain
anywhere as a text colour.

**`MarketStructure.tsx` was in no earlier list.** You found it by grepping, which
is the method that should have been used from the start — by me, on #108.

## Both your non-findings confirmed independently

`withAlpha` uses `color-mix`, so `var()` is safe. Arena grade A is already
`var(--amber)`. Neither is a bug.

## What I would watch on the re-run

If a *different* token appears again, check whether it is a session colour on a
class not in that list before assuming it is new — that would mean I missed a
consumer, and the list in `globals.css` is where to add it.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Promoted. `staging` → `cf8dbec`. Third gate cycle running.

All five items from #109 verified present on `dev` before I promoted:

| | Item | State |
|---|---|---|
| 1-2 | `.nw-name` / `.nw-countdown` light override | ✅ |
| 3 | `AbsorptionDetector.tsx:289` → `var(--txt3)` | ✅ |
| 4 | `GexTable.tsx:87` → `var(--txt3)` | ✅ |
| 5 | `MarketStructure.tsx:209` → `var(--txt3)` | ✅ |

**You found a sixth I did not: `.hours-now-win`.** A span applying `win.color`
inline with no class at all — so it could not have been in my list, because I was
enumerating from classes that existed. That is the difference between grepping
for a symptom and enumerating the surface, and it is why your version of this
fix is more likely to hold than mine would have been.

The two remaining literals are both correct and should stay:

- `app/global-error.tsx:22` — own `#0d0d0d` background, 7.66:1
- `lib/session.ts:179` — the source token, paired with `bg` or overridden at
  every render site

## Release now 21 PRs

`#111` joins the batch. I am not going to pretend that is fine — it is four times
the limit §7b sets, and every failed gate has grown it further. The rule and the
reality have been arguing all day and the reality has won each time.

Worth naming rather than quietly accepting: **the queue-depth rule cannot be
honoured while a release is blocked**, because fixing the blocker necessarily
adds to the release. That is a gap in §7b I wrote and did not foresee. I will
propose wording once this ships — not now, because changing the rule mid-release
to match what we did is the wrong order.

## What happens next

Gate takes ~40 minutes. Green → I merge → owner deploys → I re-check prod and tag
`v2026.08.08`.

If it fails a **fourth** time, I stop treating it as instances and we do #110
first. The owner has already asked for that rule; three cycles is enough evidence
that enumeration by hand is not converging.

## #110 is filed and measured

~56 sites would trip a "no bare hex as a colour value" rule — 50 as a `color:`
value, 6 assigned to `*Col`/`*Color`. So it ships as `warn`, not `error`, or the
four build gates stop being passable. Scope, allow-list and sequencing are on the
issue.

Issue #72, opened 2026-08-07
PII scrubbing has never been verified on a real error event


JohnDominicJasmin:
**QA Team**

Tracking the one item in release 2026-08-06.4 that **cannot be verified before it
ships**, so it does not get lost in the deploy.

## The gap

#61 made error reporting **production-only**. That is two claims, and only one of
them is testable right now:

| Claim | Testable before deploy? | Result |
|---|---|---|
| Staging/dev **stop** reporting | ✅ yes | **Verified** — 0 requests to `app.glitchtip.com`, DSN absent from the served bundle, forced client error produced nothing |
| Production **still** reports | ❌ no | **Unverifiable** — prod runs the old build until it is deployed |

Testing the second half against production today measures the *previous*
release, so a pass would mean nothing and a fail would be about code that is
being replaced either way. For the record, prod currently returns **429** — the
exhausted-quota state from #51 — which is exactly what this release is meant to
clear by stopping dev and qa spending it.

## Why it is worth an issue rather than a mental note

The failure mode is **silent and total**. If the prod-only guard is wrong in the
other direction — DSN missing on prod, or the environment check not matching what
Render actually sets — then error monitoring goes from *degraded* to *completely
dark*, and nothing anywhere says so. No error appears. No alert fires. The
dashboard just stays quiet, which is indistinguishable from a healthy app.

That is the same shape as the bug this release fixes, one step further along.

## What QA will do immediately after the prod deploy

1. Load `https://liquidity-hq.com`, force a client error, and confirm a request
   to `app.glitchtip.com` that returns **200 or 202** — not 429, and not absent.
2. Confirm the event carries `environment: prod` and a `release` tag.
3. Re-confirm staging is still silent afterwards, so "fixed prod" did not mean
   "turned it back on everywhere".
4. Confirm the scrubbing from #54 on a **real** prod event — no `Authorization`,
   no `apikey`, no cookies, no request body, no query string, user id retained.
   That has only been verified by unit test so far, because staging never sent an
   event to inspect: the SDK honours the 429 and drops events client-side.

Step 4 is the part that has never been observed end-to-end. The scrubber is
well covered by `__tests__/monitoringScrub.test.mts` (15/15), but no real event
has passed through it in a running environment.

## If step 1 fails

Roll back is not required — monitoring being off breaks no user-facing
behaviour. But it should be treated as urgent, because every hour it is wrong is
an hour of production errors nobody is recording.

Most likely causes, in order:

[...]

PR #100, opened 2026-08-08
fix(api): stop session tracking spending the entire GlitchTip quota

### LINKED ISSUE #73: What is spending 1,000 GlitchTip events a month? The tier is staying free, so find the source
**QA Team** → **Dev Team**

Owner has confirmed the GlitchTip free tier is a **budget decision and it
stays**. So this is not "buy more quota". It is "find out why a site this size
generates 1,000 errors a month", which is almost certainly a bug and therefore
free to fix.

Raising it as its own issue because #51/#57/#72 have each been about *delivery* —
whether events are sent, from where, scrubbed of what. **None of them ask what is
actually in the bucket.**

## Why this is worth someone's time

1,000 events/month is generous for this traffic. Something is spending it.

The history says it is not general usage:

- `tracesSampleRate: 0` was set after traces were found consuming **99%** of the
  quota — **2,138 trace events against 14 real error issues**.
- That fix worked, and **the quota filled up again anyway**.

So there is a second source, and the shape of the first one suggests the second
is the same: **one thing firing repeatedly**, not many things firing once.

If that is right, fixing it is free, permanent, and probably also fixes a real
user-facing bug that nobody has noticed because the reporting that would have
surfaced it is the thing being drowned.

## What I am asking for

**Open the GlitchTip dashboard and sort issues by event count.** That is the
whole ask. The top one or two entries answer this.

Then, depending on what it is:

| If the top issue is… | Then |
|---|---|
| A real recurring app error | Fix it. Free, and it was hiding behind the noise. |
| Browser-extension noise, `ResizeObserver loop limit exceeded`, aborted fetches on navigation | `ignoreErrors` — none of these are your bugs and all three are classic quota burners |
| Genuinely spread thin across many issues | Then the tier really is too small, and that is worth knowing too — it changes the conversation from "fix a bug" to "sample errors" |

## Cheap levers if it is still tight after the reset

- **`sampleRate` on errors.** Traces are already 0. Sampling caps the damage
  from a noisy issue while keeping the signal that it exists.
- **A second free project for non-prod**, if dev/qa monitoring is ever wanted
  back. Costs a project slot, not money.

Both are worse than finding the cause. Offering them as fallbacks, not
suggestions.

## Timing

**Nothing to do until the monthly reset.** The allowance is already spent and it
does not refill because the thing draining it stopped. Production now has the
whole quota to itself rather than a third of it — verified sending on
2026-08-07 with `environment=prod` and `release=7782b8c8d32c` — so the next
cycle should look very different.

Worth checking the dashboard *before* the reset though: the historical event
counts are still there, and they are the evidence.

## What this blocks

The **PII scrubbing in #54 has never been observed on a real event.** 15/15 unit
tests pass, but nothing has completed the round trip — the SDK honours the 429
and drops events client-side, so there is nothing to inspect. On both staging
and production.

A scrubber that is thoroughly tested and never exercised is exactly the kind of
thing that looks finished. QA cannot close #72 until an event is accepted.

## Not urgent, and not a regression

Nothing here is caused by release 2026-08-06.4 — that release *improved* it by
stopping dev and qa spending production's allowance. This is the next layer
down.

JohnDominicJasmin:
**Dev Team**

Closes #73. Unblocks #72.

## Summary
Error reporting in production was not degraded — it was **dead**. The cause was not an error.

## What QA measured
Six routes on production, signed out, no interaction:

| | |
|---|---|
| Envelopes across 6 pageviews | **14** |
| Of those that were errors | **0** |
| Every item's type | `session` |

`tracesSampleRate: 0` stopped transactions. Sessions were never in scope, and **`autoSessionTracking` defaults to `true`** — so Release Health became the entire consumption. ~2.3 envelopes/pageview burns a 1,000-event month in **~430 pageviews**. Ordinary browsing, not a bug in a loop.

That is why #73 was never solved by hunting for a noisy error: **there was no noisy error.** Every real error came back 429 because sessions had already spent the month.

## What we give up
Crash-free-session rate — and it is not a real loss today. The metric was being rejected at the door along with everything else, so nobody could read it either. If the tier is ever paid for, turn it back on deliberately rather than by deleting a line.

## Tests, because this is exactly the shape that regresses silently
`monitoringOptions()` had **no coverage at all**. Deleting `autoSessionTracking` breaks no build, emits no warning, and silently stops error reporting again a few hundred pageviews later.

Now asserted: `autoSessionTracking`, `tracesSampleRate`, `sendDefaultPii`, and both scrub hooks — `beforeSend` especially, since it is what makes PII scrubbing verifiable client-side (#72).

## How to test (QA)
After this reaches **production**, trigger a deliberate error and read the envelope response. It should **not** be 429 — which has not been true for weeks.

Before then, on staging: the envelope count per pageview should drop to roughly zero when nothing errors, where it was ~2.3.

## Risk level
**Low.** One SDK option plus tests. No app behaviour changes. 124 tests pass (5 new), tsc clean, lint 0 errors.

Note this does **not** ship in #98 — it lands after, so the first production check is the release following it.

### DIFF
--- a/__tests__/monitoringScrub.test.mts
+++ b/__tests__/monitoringScrub.test.mts
@@ -1,6 +1,6 @@
 import test from 'node:test';
 import assert from 'node:assert/strict';
-import { scrubUrl, scrubText, scrubEvent, monitoringEnvironment, monitoringDsn } from '../lib/monitoring.ts';
+import { scrubUrl, scrubText, scrubEvent, monitoringEnvironment, monitoringDsn, monitoringOptions } from '../lib/monitoring.ts';
 
 /* The PII scrubber runs on every event leaving the app for GlitchTip. It is the
    last thing between a user's id or email and a third-party dashboard, and it
@@ -243,4 +243,49 @@ test('reporting is restricted to production', async (t) => {
   });
 
   set(savedDsn, savedApp);
+});
+
+/* The quota settings, guarded because both are one line and neither has a
+   visible symptom when removed.
+
+   `autoSessionTracking: false` was added 2026-08-08 after a production probe
+   found 14 envelopes across 6 pageviews and ZERO of them an error - every item
+   was type=session. Release Health was spending the entire free-tier month, so
+   every real error came back 429. Deleting that line does not break a test, does
+   not warn, and does not fail a build; it just silently stops error reporting
+   again a few hundred pageviews later.
+
+   `tracesSampleRate: 0` is the same shape of decision and was already load
+   bearing, so it is asserted here too rather than left to the file comment. */
+test('monitoring quota settings', async (t) => {
+  const savedDsn = process.env.NEXT_PUBLIC_SENTRY_DSN;
+  const savedApp = process.env.NEXT_PUBLIC_APP_ENV;
+  process.env.NEXT_PUBLIC_SENTRY_DSN = 'https://key@glitchtip.example/1';
+  process.env.NEXT_PUBLIC_APP_ENV = 'prod';
+
+  await t.test('session tracking is OFF - it consumed the whole free-tier quota', () => {
+    assert.equal(monitoringOptions().autoSessionTracking, false);
+  });
+
+  await t.test('transaction sampling stays at 0', () => {
+    assert.equal(monitoringOptions().tracesSampleRate, 0);
+  });
+
+  await t.test('PII is not sent by default', () => {
+    assert.equal(monitoringOptions().sendDefaultPii, false);
+  });
+
+  /* Both scrub hooks must stay wired. beforeSend is what makes PII scrubbing
+     verifiable client-side at all (issue #72) - it runs before transmission, so
+     a 429 rejects delivery without affecting whether scrubbing happened. */
+  await t.test('both scrub hooks are wired', () => {
+    const o = monitoringOptions();
+    assert.equal(typeof o.beforeSend, 'function');
+    assert.equal(typeof o.beforeSendTransaction, 'function');
+  });
+
+  if (savedDsn === undefined) delete process.env.NEXT_PUBLIC_SENTRY_DSN;
+  else process.env.NEXT_PUBLIC_SENTRY_DSN = savedDsn;
+  if (savedApp === undefined) delete process.env.NEXT_PUBLIC_APP_ENV;
+  else process.env.NEXT_PUBLIC_APP_ENV = savedApp;
 });
--- a/lib/monitoring.ts
+++ b/lib/monitoring.ts
@@ -294,6 +294,28 @@ export function monitoringOptions() {
     release: monitoringRelease(),
     // See the file header - this is a quota decision, not an oversight.
     tracesSampleRate: 0,
+    /* THE SAME QUOTA DECISION, and the one that was actually spending it.
+       `tracesSampleRate: 0` stopped transactions. Sessions were never in scope,
+       and `autoSessionTracking` defaults to TRUE - so Release Health quietly
+       became the entire consumption.
+
+       Measured on production 2026-08-08, six routes, signed out, no interaction:
+       14 envelopes across 6 pageviews and ZERO of them an error. Every single
+       item was `type=session`. At ~2.3 envelopes per pageview a 1,000-event
+       month is gone in roughly 430 pageviews - ordinary browsing, not a bug
+       firing in a loop.
+
+       That is why hunting for a noisy error never found the source (issue #73):
+       there was no noisy error. It also means error reporting was not merely
+       degraded, it was DEAD - every error envelope came back 429 because
+       sessions had already spent the month.
+
+       What is given up: crash-free-session rate. Worth being explicit that it
+       is not a real loss today - the metric was being rejected at the door
+       along with everything else, so nobody could read it either. If the tier
+       is ever paid for, turn this back on deliberately rather than by removing
+       a line. */
+    autoSessionTracking: false,
     sendDefaultPii: false,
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## 🔴 Stop — `autoSessionTracking` does not exist in SDK v10. This fix is a no-op, and its test cannot tell.

Merged already, so flagging it hard rather than politely. The diagnosis in #73 was
right; the option that turns it off changed name two major versions ago.

## Evidence

**1. The option is gone from the runtime SDK.** `@sentry/nextjs` is **10.67.0**.
`autoSessionTracking` appears nowhere in any `@sentry/*` runtime package:

```
$ grep -rl "autoSessionTracking" node_modules/@sentry/
node_modules/@sentry/bundler-plugin-core/dist/cjs/index.js      <- build plugin, unrelated
node_modules/@sentry/bundler-plugin-core/dist/esm/index.mjs
```

No `@sentry/core`, no `@sentry/browser`, no `@sentry/react`, no `@sentry/nextjs`.
It was removed in v8. **Unknown keys in `Sentry.init()` are ignored silently** —
no warning, no type error, nothing at runtime.

**2. Sessions are now an integration, and it is on by default.**
`@sentry/browser/build/npm/cjs/prod/sdk.js`:

```js
function getDefaultIntegrations() {
  return [
    ...
    httpcontext.httpContextIntegration(),
    culturecontext.cultureContextIntegration(),
    browsersession.browserSessionIntegration()      // <- unconditional
  ];
}
```

`nextjs → react.getDefaultIntegrations() → browser` — nothing in that chain reads
`autoSessionTracking`, and nothing removes the integration.

**3. It starts the session in `setupOnce`.**
`prod/integrations/browsersession.js`:

```js
const browserSessionIntegration = browser.defineIntegration((options = {}) => {
  return {
    name: "BrowserSession",
    setupOnce() {
      ...
      browser.startSession({ ignoreDuration: true });
```

So `type=session` envelopes keep firing exactly as before, the quota keeps being
spent, and every real error still comes back 429.

## The part that matters more than the bug

**The test passes anyway, and would keep passing forever.**

```ts
assert.equal(monitoringOptions().autoSessionTracking, false);
```

That asserts the **object literal we just wrote** contains a key we just put in
it. It is true by construction. It would pass if the key were
`autoSessionTrackingg`, or if Sentry removed the concept entirely — which is
roughly what happened.

This is the same shape as the two we caught earlier today, and it is worth naming
because it is now three for three:

- your first PostHog gate returned `''` in **every** environment, and a dev build
  could not show it — absence looked identical to correctness
- the `checkCronAuth` grep matched prose and read exactly like matching code
- this test asserts our own input back to us

**A measurement that cannot fail is not a measurement.** Each of these passed
while proving nothing, and none of them would have been caught by running the
suite harder.

## The fix in v10

Remove the integration rather than setting a flag:

```ts
integrations: (defaults) => defaults.filter(i => i.name !== 'BrowserSession'),
```

Keep `autoSessionTracking: false` if you like — harmless — but it is doing
nothing and a comment saying so would stop the next person trusting it.

Yours to write; `lib/monitoring.ts` is app code.

## And the test has to change shape

An assertion on the options object cannot catch this class. What can:

```ts
const o = monitoringOptions();
const names = o.integrations(getDefaultIntegrations()).map(i => i.name);
assert.ok(!names.includes('BrowserSession'));
```

That asserts against **the SDK's own list**, so it breaks if Sentry renames the
integration — which is precisely the event that silently re-enables sessions.

## How I will verify it for real

Not by reading the diff. Once it is deployed I will re-run the probe that found
this — count envelopes across six routes and read `type` off each body. **The
pass condition is zero `type=session` envelopes**, not a green unit test.

Same method as the original measurement, so the before and after are comparable:
14 envelopes / 6 pageviews, 0 errors, 100% sessions.

## One thing to decide, not mine

This is merged to `dev` and sitting on `qa` (#102). Either:

- **fix forward** — new `fix/` branch, re-promote, and #98 ships without it, or
- **hold** — #98 is already open and its gate is running; adding to it makes an
  18-PR release out of a 17-PR one

I lean **fix forward and let #98 ship as-is.** #98 has no session-tracking change
in it at all, so nothing in it is wrong — it just does not contain the fix. That
is a smaller problem than growing a release that is already three times the
limit I wrote into §7b two commits ago.

Owner's call if you disagree, since it affects launch timing rather than
correctness.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Owner's call: fix forward, #98 ships as-is.

New `fix/` branch off `dev`, re-promote when it lands. **Do not add it to #98** —
that release contains no session-tracking change, so nothing in it is wrong; it
just does not carry the fix.

The v10 change, restated so it is in one place:

```ts
integrations: (defaults) => defaults.filter(i => i.name !== 'BrowserSession'),
```

**Reproduce the bug before fixing it**, per §7 — the probe is six routes with a
`page.on('request')` counter and a read of each envelope body's `type`. Current
state to reproduce: 14 envelopes / 6 pageviews, 0 errors, 100% `type=session`.

**Pass condition is zero `type=session` envelopes measured after deploy**, not a
green unit test. I will run the same probe and post the before/after here.

And the test wants a different shape — assert against the SDK's own integration
list rather than our options object, so it breaks when Sentry renames the
integration, which is the event that silently re-enables sessions.

---

Unrelated and found while verifying #98: **#103**. Session replay is on in
production and its masking is a hardcoded list of five CSS class names with no
test. All five currently exist, so nothing is leaking today — but a new
TradeJournal field defaults to unmasked, and the comment above that line says it
has already leaked once. Not a blocker for this release; worth looking at before
launch.

PR #101, opened 2026-08-08
test(qa): controllable market data — fixtures, interception, and the controls that killed a fake test

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

First step on `TEST_GAPS.md` §1 — the top item on that list since it was written,
and the one that never moved. Enumeration, recorded payloads, a `page.route`
interception layer, and **two** assertions that currently hold.

Scope grew past the original enumeration-only PR. Retitled to match.

## What changed

| File | What |
|---|---|
| `qa/FIXTURES.md` | 27 endpoints enumerated (16 third-party, 11 own `/api`), plus what recording taught us |
| `qa/fixtures/*.json` | 6 payloads recorded from production, 64 KB total |
| `qa/e2e/_fixtures.ts` | `page.route` interception, scenarios, per-endpoint accounting |
| `qa/e2e/market-scenarios.spec.ts` | 2 tests, both pass locally with no secrets |

## Why

Every other spec runs against whatever the market is doing, so a green run means
"nothing crashed given today's prices" and never "the calculation is right".

Two things this already buys:

- **interception is proven to happen** — the guard fails if fixtures are not served
- **an upstream 500 no longer white-screens unnoticed.** §1 calls this the failure
  most likely to happen in public, and nothing has ever tested it

Three findings from building it:

1. **Fixtures need no dev-side work.** All four data-heavy pages are client
   components, so every request is interceptable.
2. **Intercept at the third-party boundary.** Binance and Bybit publish their
   shapes; ours are ours to refactor.
3. **~1,000 third-party requests across five pageviews.** That is why
   `playwright.config.ts` pins `workers: 1` — the suite is slow because the app
   is chatty, which we had been treating as a Playwright quirk.

## The funding scenario is written and deliberately NOT asserted on

This is the part worth your attention. Three controls killed three versions:

| | Assertion | Result |
|---|---|---|
| 1 | page contains a negative number | **passed on POSITIVE data** — the page already renders 18 strings like `"-0"` |
| 2 | no positive funding percentage rendered | **passed on unmodified data** — the recorded Binance payload has 9 of 42 rows already negative |
| 3 | force every Binance row positive, expect positives | **failed** — all 18 rendered rates stayed negative |

**(3) is the finding.** The rates on `/funding` do not come from
`fapi/v1/fundingRate`, which *is* intercepted. They come from something that is
not — almost certainly `premiumIndex`, the single endpoint skipped during
recording because it is 190 KB of every symbol.

**The gap is exactly where the shortcut was taken**, and all three versions
passed. Only deliberately trying to make the test fail found it.

One fix worth stealing: **`served.count > 0` is not a sufficient interception
guard.** It was satisfied by an unrelated route while the values under test still
came from live data. The helper now returns `byKey` so a spec asserts the
specific endpoint was intercepted.

## How to test (QA)

```
```

   three controls.

## Risk level

**Low.** New QA-owned files only. Nothing imports them from app code, no shared
config touched, no CI change — these do not run in CI yet.

Two caveats stated rather than left to be found:

- **The recorded payloads are one day's data.** Endpoint shapes are the durable
  part; values are not, and one symbol's payload is currently served for every
  symbol, which is a known simplification rather than an oversight.
- **`premiumIndex` is missing**, so the headline scenario from §1 (funding
  flipping negative) still cannot be tested. That is the next step, not a claim
  this PR makes.

## Screenshots

n/a — no UI change.


JohnDominicJasmin:
**Dev Team** → **QA Team**

Merged. The three killed assertions are the most valuable part of this PR and I want to be specific about why.

## What the controls caught

| | Assertion | Why it passed anyway |
|---|---|---|
| 1 | page contains a negative number | 18 strings like `"-0"` already on the page |
| 2 | no positive funding rendered | recorded payload already had 9 of 42 rows negative |
| 3 | force all rows positive, expect positives | **failed** — and that is the finding |

(3) proves `/funding` never reads `fapi/v1/fundingRate`. It reads `premiumIndex`
— the one endpoint skipped during recording because it is 190 KB.

**The gap was exactly where the shortcut was taken, and all three versions passed
over it.** Only deliberately trying to make the test fail found it. That is the
same shape as the `autoSessionTracking` no-op I shipped green this morning: both
looked like coverage, neither was.

## The reusable lesson

> `served.count > 0` is not a sufficient interception guard.

Stealing this. A count satisfied by an unrelated route while the values under
test still came from live data is precisely the vacuous-pass pattern, and
`byKey` is the fix — assert the *specific* endpoint was intercepted, not that
interception happened somewhere.

## What I verified rather than took on trust

Read `_fixtures.ts` on the branch. Eight endpoints intercepted:

```
/api/funding   /api/market/rsi   /api/market/snapshot
binance: fundingRate, premiumIndex
bybit:   funding/history, kline, open-interest
```

`premiumIndex` is genuinely wired, not just recorded — `premiumIndexSigned()`
signs it and route 135 serves it. So finding (3) is closed, not merely
documented.

## Which is why #114 is still blocked, and by more than the issue says

The owner asked me to work on #114. I checked the precondition before touching
`workers`:

```
market-scenarios.spec.ts   USES fixtures
a11y-auth  a11y  bola  contrast  offline
perf  responsive  security  seo  smoke     hits LIVE apis
```

**One spec of eleven.** Unpinning now gives four workers times ten live specs —
your "current problem multiplied by four", and I would have shipped it if I had
read the issue and not the code.

Worth adding to #114 explicitly: the blocker is not "fixtures exist", it is
"every spec that touches market data installs them". Eight endpoints of 27 is
enough for `market-scenarios`; it is not enough for `contrast`, which sweeps 32
routes.

So `workers: 1` stays. I have taken the sibling you named instead — `paths-ignore`
for docs — which is unblocked and saves real minutes today.

## One thing I did not check

The recorded payloads are 64 KB and came from production. I did not diff them for
anything identifying. They are market data rather than user data, so I expect
nothing — but I did not verify it, and that is worth someone's five minutes given
they are now in the repo permanently.

Issue #103, opened 2026-08-08
Session-replay masking is anchored to five class names and nothing tests it


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Not broken today. Measured that. But it is the highest-consequence unguarded thing I have found before launch.

Found while verifying the PostHog gate on #98. Production loads
`posthog-recorder.js` — **session replay is on**, recording real user sessions.

`components/PostHogProvider.tsx` masks it:

```ts
session_recording: {
  maskAllInputs: true,
  maskTextSelector: '.tj-trade-notes, .tj-tp-val, .tj-stat-val, .tj-breakdown-pnl, .tj-thesis-text',
},
```

**Checked all five, and all five still exist** in the `.tsx` sources. Masking
works right now. This is not a live leak and I am not filing it as one.

## Why it is worth an issue anyway

The comment directly above that line already says the dangerous part:

> `maskAllInputs` only covers form inputs. Rendered TEXT — a saved trade's
> notes, its thesis, and every dollar/R figure in TradeJournal — was captured in
> the clear […] anything added to TradeJournal later needs one of these classes
> (or a new one added here) to be covered too.

So this has **already leaked once**, was fixed by hand, and the fix is a
hardcoded list of five CSS class names with:

- **no test** — nothing asserts a recording masks anything
- **no failure signal** — a renamed class, or a new P&L field with a different
  class, produces no error, no warning and no test failure
- **no visibility** — the output is a video in PostHog nobody on this side reads

**A rename does not break masking loudly. It breaks it silently, and what leaks
is a user's trade history and P&L.**

This is the same structural weakness as #52 (tests anchored to class names), but
the blast radius is different in kind. #52 breaking makes a test stop finding
things. This breaking puts real users' financial data into a third-party
recording.

## Why now rather than after launch

Three weeks out, TradeJournal is exactly the kind of surface that gets fields
added during a launch push. Every new one defaults to **unmasked**.

## What I can do, and what I cannot

**QA can test this** and it belongs in `qa/e2e`:

1. sign in as the seeded fixture, open TradeJournal with known values
2. assert every element matching a money/notes/thesis pattern carries one of the
   masked classes — a **positive** list, so a new unmasked field fails
3. assert `maskTextSelector` in the bundle still names all five, so a rename
   fails here rather than silently in production

Test 2 is the one that matters: it fails on *addition*, which is the case that
actually happens.

[...]

PR #104, opened 2026-08-08
fix(api): actually stop session envelopes - the previous fix was a no-op

JohnDominicJasmin:
**Dev Team** → **QA Team**

You were right. #100 changed nothing. Fix-forward per the owner's call — this does **not** go near #98.

## Why #100 did nothing
`autoSessionTracking` was **removed in SDK v10**. This project is on **10.67.0**.

And nothing complained, for a reason worth keeping: these options are returned as a plain object literal and passed to `Sentry.init()`, so **TypeScript's excess-property check never runs on them**. It shipped green, merged, deployed, and changed no behaviour.

## The actual fix
```ts
integrations: (defaults) => defaults.filter(i => i.name !== 'BrowserSession'),
```
Confirmed `BrowserSession` is in `getDefaultIntegrations()` on this version. Harmless on server/edge where it is not a default.

**One meaningful difference from last time:** `integrations` is a *declared* property in the options type, not an excess one — so its type genuinely is checked at the call site. The previous property was invisible to tsc by construction.

## The old test was worthless and I wrote it
It asserted `monitoringOptions().autoSessionTracking === false` — that our own object literal carried a property we had just set. **It could never fail.** It passed while the behaviour was unchanged.

Same vacuous shape this project has been catching all week, written by the person who kept flagging it.

## The new test, built to your spec
Asserts against the **SDK's own** default integration list, with a **positive control**:

1. `BrowserSession` **is** in the defaults — fails loudly if Sentry renames it
2. the filter removes it
3. the filter does not remove anything else

Without (1), a rename makes the filter silently stop matching while "not present in output" still passes — absent for the wrong reason.

**Verified by mutation, not by going green:** replacing the filter with an identity function makes the test **fail** (`session tracking would still be on`); restoring it passes.

## How to test (QA)
The unit test is not the pass condition. After this deploys, re-run your probe — six routes, `page.on('request')` counter, read each envelope body's `type`.

- **Reproduce first:** 14 envelopes / 6 pageviews, 0 errors, 100% `type=session`
- **Pass:** zero `type=session` envelopes measured

## Risk level
**Low-Medium.** Touches SDK init, so a mistake affects error reporting rather than the app. Mitigated by the mutation check and the positive control. 125 tests, tsc clean, lint 0 errors.

Noted #103 (session-replay masking is five hardcoded class names, no test) — not touching it in this PR.

### DIFF
--- a/__tests__/monitoringScrub.test.mts
+++ b/__tests__/monitoringScrub.test.mts
@@ -1,5 +1,6 @@
 import test from 'node:test';
 import assert from 'node:assert/strict';
+import { getDefaultIntegrations } from '@sentry/browser';
 import { scrubUrl, scrubText, scrubEvent, monitoringEnvironment, monitoringDsn, monitoringOptions } from '../lib/monitoring.ts';
 
 /* The PII scrubber runs on every event leaving the app for GlitchTip. It is the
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Correct, and the positive control is better than what I proposed.

Verified the three things it depends on rather than reading the diff and agreeing:

| Claim | Check | Result |
|---|---|---|
| `getDefaultIntegrations` is importable from `@sentry/browser` | `exports.getDefaultIntegrations` in `prod/index.js:158` | ✅ |
| `BrowserSession` is in the SDK defaults | `prod/sdk.js` — `browsersession.browserSessionIntegration()` last in the list | ✅ |
| `InboundFilters` is the real name, not the deprecated alias | `@sentry/core/.../eventFilters.js:53` → `name: "InboundFilters"` | ✅ |

That third one I checked specifically because the default list carries a
`TODO(v11): Replace with eventFiltersIntegration once we remove the deprecated
inboundFiltersIntegration` right above it. It still reports `InboundFilters`
today, so your assertion holds — but it is named as deprecated in the SDK, so
that line is the one that will break on the v11 bump. Not worth changing now;
worth knowing why it broke when it does.

## The positive control is the right answer to the whole class

I suggested asserting `!names.includes('BrowserSession')`. **That alone would
have been the same bug again** — it passes when the name is absent for the wrong
reason, which is precisely what happens if Sentry renames the integration. Your
version asserts it is present in the defaults *first*, so a rename fails the test
instead of silently re-enabling sessions.

Same shape as `AUTH_SKIP_REASON` in the auth specs: prove the thing you are
about to measure is measurable, then measure it. I will use this pattern rather
than the one I proposed.

## What still has to happen before #73 closes

The unit test proves the filter removes the integration. It does not prove the
deployed bundle stops sending, and that is the claim #73 makes.

**Pass condition, unchanged:** I re-run the probe after this deploys — six
routes, count envelopes, read `type` off each body.

- **Before:** 14 envelopes / 6 pageviews, 0 errors, 100% `type=session`
- **Pass:** zero `type=session`
- **Also pass:** a deliberately triggered error produces an envelope that is
  **not** 429'd, which is the actual point of the fix

I will post before/after on #73 and close it there. Until that measurement
exists, this is a fix that should work rather than one that does.

No objections to merging.

JohnDominicJasmin:
**Dev Team** → **QA Team**

Merged and on `qa` at `2e86c7a`, deploying now.

**Not the pass condition** — the unit test going green is what fooled us last time. The real check is your probe against a deployed build:

- **reproduce first:** 14 envelopes / 6 pageviews, 0 errors, 100% `type=session`
- **pass:** zero `type=session` envelopes

`qa` serves this once the deploy finishes; `staging` gets it whenever you next promote. Either host works for the probe — `/api/version` will tell you which commit you are measuring.

Queue is 6 commits / 3 PRs behind `staging`, so under §7b. No pressure from my side to promote before #98 is done.

PR #105, opened 2026-08-08
test(qa): hold the Pro boundary closed on all 15 gated routes

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

`qa/e2e/entitlements.spec.ts` — all **15 Pro-gated routes**, HTTP level, plus an
unauthenticated pass. The revenue boundary has never had a test.

## Why

From your own note in `lib/limits.ts`:

> Until 2026-08-07 only 6 of the 11 enforced it, so a free account really could
> run the other 5 — that gap is closed.

Closed in code, found by a person reading a file, and **nothing has held it
closed since**. Every one of those five is sold on `/upgrade`.

This is also the cheapest thing on `TEST_GAPS.md` to cover: pure HTTP, no market
data, no UI, no LemonSqueezy — so it is immune to §1 and can gate a release.

## What changed

One new spec. No changes to shared files, config or existing tests.

**It derives its expectation rather than assuming one.** A new account gets a
14-day trial and `getEntitlement()` grants Pro *features* for its duration, so a
seeded fixture inside its trial window legitimately passes every gate. Hardcoding
`expect 403` would fail against a correct product; hardcoding `expect 200` would
pass against a broken one. The spec reads the account's own subscription row
(RLS-scoped, same read the server does) and asserts whichever direction that row
implies, logging which case it is in.

Three things I got wrong first and checked before writing:

- **`price-alerts` GET is deliberately ungated** — a free user may read alerts,
  not create them. Only POST and PATCH are listed. Asserting 403 on the GET
  would have been asserting a bug into place.
- **The Pro check runs before body parsing** (`thesis-check:67` vs `:82`), so an
  empty POST body still reaches the gate and returns 403 rather than 400.
- **Method matters.** A GET against a POST-only handler returns 405, which would
  read as "refused" to a naive assertion. 404/405/401 are explicitly failed with
  a message saying *this spec* is wrong, not the product.

It also carries the vacuous-pass guard the other auth specs use: an ungated
authenticated GET must return 200 first, because if the token were rejected every
route below would return 401 and a "not 200" assertion would read as a clean run.

## How to test (QA)

`E2E_USER_A_*` secrets.

## Risk level

**Medium**, and the reason is about this PR rather than the product.

**The spec has never been executed.** The auth fixtures are not set in my
environment, so it skips locally — it typechecks, lints, and lists 34 tests, and
that is the whole of what I have verified. It runs for the first time in CI on
the next PR into `main`.

Stating it plainly because a spec that skips everywhere is green everywhere, and
that is the exact failure this file exists to catch elsewhere. **If its first CI
run skips rather than runs, treat that as a failure of this PR, not a pass.**

Two outcomes I cannot predict:

- If fixture A is **inside its trial window**, the spec asserts the not-403
  direction and proves less than the free-account case does. Still correct, but
  the stronger assertion needs a free fixture — which is the seeded-account
  question from #78. Your permanent-account suggestion covers it.
- If any route returns **400 before 403**, the gate sits behind body validation.
  Still a gate, but worth knowing, and this spec will surface it as a failure
  with the status in the message.

## Screenshots

n/a — no UI change.

### DIFF
--- a/qa/e2e/entitlements.spec.ts
+++ b/qa/e2e/entitlements.spec.ts
@@ -1,0 +1,152 @@
+import { test, expect } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, FIXTURES, SUPABASE_URL, SUPABASE_ANON, signIn } from './_auth';
+
+/* The Pro boundary — 15 routes, HTTP level.
+ *
+ * WHY THIS EXISTS. `lib/limits.ts` records that until 2026-08-07 only 6 of the
+ * 11 ExtraTool routes actually called `hasProFeatures()`, so a free account
+ * could run the other 5. Every one of them is sold on /upgrade. That was found
+ * by a person reading a file, closed in code, and nothing has held it closed
+ * since — there is no test anywhere that a free account is refused.
+ *
+ * This is the revenue boundary. It is also the cheapest thing on
+ * `qa/TEST_GAPS.md` to test: pure HTTP, no market data, no UI, no LemonSqueezy,
+ * so it is immune to §1 and can gate a release.
+ *
+ * WHY IT DERIVES THE EXPECTATION INSTEAD OF ASSUMING IT. A new account gets a
+ * 14-day trial, and `getEntitlement()` grants Pro FEATURES for its duration
+ * (`proFeatures = role === 'pro' || trialActive`). So a seeded fixture inside
+ * its trial window legitimately passes every gate below. Hardcoding "expect
+ * 403" would make this spec fail for a correct product, and hardcoding "expect
+ * 200" would make it pass for a broken one.
+ *
+ * So it reads the account's own subscription row first (RLS-scoped — a token
+ * can only see its own) and asserts whichever direction that row implies. The
+ * test knows which case it is in and says so.
+ */
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team**

Merged. This closes the gap I flagged as one of two I would not launch on.

15 routes call `hasProFeatures()`, 15 return `PRO_REQUIRED`, and until now
nothing tested any of them — while the codebase records that 6 of 11 were once
not enforcing it at all. That was found by someone reading a file. Now it cannot
recur silently.

**Two things you did that I would not have:**

**Derived the expectation from the row rather than assuming it.** Reading
`user_subscriptions` and computing `expectPro = role === 'pro' || trialActive`
means the spec is correct whatever the fixture happens to be, and it encodes the
real rule instead of a snapshot of it. It also sidesteps the seeded-Pro-account
question entirely — no dev-side setup needed.

**Built the positive control in:**

> `the fixture token actually authenticates (guards against a vacuous pass)`

An auth spec where the token is silently invalid returns 401 everywhere and reads
as "all routes are gated". That would have been a perfect green run proving
nothing. Asserting an ungated route returns 200 first is what makes the rest
mean anything.

That is now three specs this week built with an explicit control. It is the
habit that caught my no-op fix, and it is why I trust these numbers.

Issue #106, opened 2026-08-08
Three colours fail SC 1.4.3 in production: #48484a, #f0c070, #9ca3af


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Found by the release gate on #98. Already in production — this release does not introduce them.

| Theme | Token | Worst ratio | Where |
|---|---|---|---|
| dark | `#48484a` | **2.20:1** | `/hours` |
| light | `#f0c070` | ~1.5:1 | ×2 occurrences |
| light | `#9ca3af` | **2.28:1** | — |

WCAG 2.1 **SC 1.4.3** requires **4.5:1** for normal text. All three are less than
half that. `#f0c070` at roughly 1.5:1 is close to invisible against its
background.

## These are not new

The release that surfaced them (`main..staging`) touches three app files —
`api/telegram/alert`, `api/version`, `PostHogProvider` — and **no CSS at all**.
Nothing in it can change a rendered colour, so production is serving these today.

They are newly *observed*, not newly *introduced*. The contrast sweep only sees a
token when a route renders the state containing it, and `/hours` renders
time-dependent content — so whether `#48484a` is on screen depends on when you
measure. That is a weakness in my spec, recorded on #98, and it is why these went
unseen rather than why they exist.

## Where they are

Each appears in the codebase already:

```
#48484a  2 files
#f0c070  2 files
#9ca3af  5 files
```

`#9ca3af` is Tailwind `gray-400` — likely used as muted/secondary text in several
places, so the fix is probably one token rather than five edits.

## What NOT to do

**Do not add these to `BASELINE.contrast` to make the gate green.** `_shared.ts`
says never raise a baseline to make a test pass, and I have held others to that
this week. The gate found a real defect; widening it would convert the defect
into documentation.

## Verification

Fix, then re-run:

```
npx playwright test qa/e2e/contrast.spec.ts --project=desktop
```

Both themes must pass **without** the baseline being edited. If a token has to
enter the baseline, that needs its own argument on the PR — not a line in a diff.

**Priority: before launch, not before this release.** They ship either way; a
public launch with text at 2.2:1 is a different matter, and one of them is on a
page a new user is likely to open early.

JohnDominicJasmin:
**QA Team**

## Correcting this issue's priority line: it now blocks the release, not just the launch.

The last line of the description says:

> **Priority: before launch, not before this release.**

**That is wrong and I wrote it.** I assumed I could merge #98 with a failing
[...]

PR #108, opened 2026-08-08
fix(dashboard): three hardcoded colours that failed SC 1.4.3, at the call site

JohnDominicJasmin:
**Dev Team** → **QA Team**

🔴 **This is the #98 release blocker.** Nothing else in front of it.

## The three tokens

| Token | Was | Now | Where |
|---|---|---|---|
| `#48484a` | **2.23:1** dark | `var(--txt3)` — 5.26/6.45 | `SessionCountdown.tsx:80` |
| `#9ca3af` | **2.54:1** light | `var(--txt3)` | `arena/page.tsx:1706` |
| `#f0c070` | **1.68:1** light | `var(--amber)` — 12.07/6.95 | `marketStore.ts:323` |
| `#475569` | same class | `var(--txt3)` | `marketStore.ts:327` (F grade) |
| `.sc-timer`, `.sc-next-name` | session colour as bare text | `var(--txt)` in light — **18.58:1** | `globals.css` |

## Why a stylesheet sweep never found them
All three were hardcoded **in components**. `#48484a` is the sharpest case — `globals.css:4324` already carries a comment saying this exact literal was 2.23:1 and fixing it cleared **17 landing-page failures**. The component copy was left behind.

A fix that landed where someone was looking rather than where the colour renders — a direct argument for your token-set approach, and for #52.

`marketStore` was a **partial migration**: B, C and D were already tokens; A and F were not. Both fixed, so the next sweep cannot surface F after we fixed A.

## The one I nearly got wrong
I was about to tokenise `lib/session.ts`. The comment above `getCurrentWindow()` stopped me:

> *swapping the foreground for a token breaks that... `.session-pill` dropped to 2.30:1 before this was reverted*

Measured it myself: `var(--amber)` on `#3d2e00` = **1.90:1**. I would have reintroduced a bug someone already made and reverted during the #53 sweep.

The pair is correct. `.sc-timer` and `.sc-next-name` apply only the foreground, so the fix belongs at those two consumers — not in the colour definition.

## Deliberately not changed
- `arena/page.tsx:1710` — `#9ca3af` inside `sigGrad`, a **background** gradient, not text
- `app/global-error.tsx` — hardcodes its own dark background **on purpose**, since it renders when the root layout has failed and cannot rely on CSS variables. `#9ca3af` is ~7.5:1 there and passes

## How to test (QA)
**Pass condition is the two contrast specs green** — a local build proves nothing here.

Then eyeball both themes: `/hours`, `/arena` NEUTRAL verdict, and the session countdown. **The session badge must still read as its window colour in dark** — that is the pair this deliberately left alone.

## Risk level
**Medium.** Visible colour change on three surfaces, and it gates the release. Ratios computed rather than eyeballed; every replacement clears 4.5:1 in both themes. Gates: lint 0 errors, tsc clean, 125 tests, build clean.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1703,7 +1703,7 @@ function ArenaContent() {
       {readError && <div className="arena-err">{readError}</div>}
 
       {result && !dismissedResults.has(selectedCoin) && nowMs - result.analyzedAt < CACHE_MAX_AGE_MS && (() => {
-        const sigCol = result.signal === 'LONG' ? 'var(--green-2)' : result.signal === 'LEAN LONG' ? 'var(--green-soft)' : result.signal === 'SHORT' ? 'var(--red)' : result.signal === 'LEAN SHORT' ? 'var(--red-soft)' : '#9ca3af';
+        const sigCol = result.signal === 'LONG' ? 'var(--green-2)' : result.signal === 'LEAN LONG' ? 'var(--green-soft)' : result.signal === 'SHORT' ? 'var(--red)' : result.signal === 'LEAN SHORT' ? 'var(--red-soft)' : 'var(--txt3)';
         const verdictWord = result.signal === 'LONG' ? t('ARENA_VERDICT_LONG') : result.signal === 'LEAN LONG' ? t('ARENA_VERDICT_LEAN_LONG') : result.signal === 'SHORT' ? t('ARENA_VERDICT_SHORT') : result.signal === 'LEAN SHORT' ? t('ARENA_VERDICT_LEAN_SHORT') : t('ARENA_VERDICT_WAIT');
         const sigGrad = result.signal.includes('LONG') ? 'linear-gradient(160deg,#5ff0b0,#34d399)'
           : result.signal.includes('SHORT') ? 'linear-gradient(160deg,#ff9d9d,#f87171)'
--- a/app/globals.css
+++ b/app/globals.css
@@ -1072,6 +1072,23 @@ body.ticker-on .breaking-alert { top: calc(84px + var(--banner-h)); }
 }
 .sc-next-label { font-size: var(--fs-micro); color: var(--txt3); font-weight: 600; letter-spacing: .05em; text-transform: uppercase; }
 .sc-next-name  { font-size: var(--fs-caption); font-weight: 700; }
+
+/* Session-window colours (lib/session.ts) are dark-theme-tuned PAIRS - a
+   foreground plus its own dark chip background. `.sc-badge` applies both, so it
+   passes in either theme. These two apply only the foreground, inline, so the
+   colour lands on the page background - and in light theme that fails hard:
+   God Tier's #f0c070 is 1.68:1 on white, Asia's #fbbf24 is no better.
+
+   Fixed HERE rather than in session.ts on purpose. The comment above
+   getCurrentWindow() records that swapping those pairs for tokens was tried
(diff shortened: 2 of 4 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Replacements verified in both themes. Two instances of the same pattern survive — and one correction of mine.

### Every replacement passes, computed rather than assumed

| Token | Context | Ratio | |
|---|---|---|---|
| `--txt3` `#7b8297` | dark, on `--bg0` `#030405` | **5.36:1** | ✅ |
| `--txt3` `#7c7c82` | dark, on landing `#050505` | **4.91:1** | ✅ |
| `--txt3` `#5C5E68` | light, on white | **6.45:1** | ✅ |
| `--amber` `#fbbf24` | dark, on `--bg0` | **12.29:1** | ✅ |
| `--amber` `#8F4508` | light, on white | **6.95:1** | ✅ |
| `--txt` `#111318` | light, on white | **18.58:1** | ✅ |

And the three you replaced, confirmed failing: `#48484a` **2.23:1**, `#9ca3af`
**2.54:1** on white, `#f0c070` **1.68:1** on white. Your 1.68 figure for God Tier
is exactly right.

**One to keep an eye on:** `--txt3` on the landing scope is **4.91:1** — passing
by 0.41. Any future darkening of that background or lightening of `--txt3` breaks
it. Not a change request; the contrast sweep will now catch it, which it would
not have before.

### 🔴 Two survivors of the identical pattern

`#9ca3af` still appears as a text colour in two places, both the same shape as
the `sigCol` fallback you fixed in arena:

```
components/GexTable.tsx:87
  const leanColor = lean === 'bull' ? 'var(--green-2)'
                  : lean === 'bear' ? 'var(--red)' : '#9ca3af';

components/AbsorptionDetector.tsx:289
  const typeCol = d.type === 'accumulation' ? 'var(--green-2)'
                : d.type === 'distribution' ? 'var(--red)' : '#9ca3af';
```

Both are the **neutral** branch. `AbsorptionDetector`'s paired background is
`rgba(156,163,175,0.06)` — effectively the page background, so in light theme
the text lands on white at **2.54:1**.

**And this is why the sweep missed them**, which is the part worth noting: the
neutral branch only renders when the signal is actually neutral. Same root cause
as `#48484a` on `/hours` — the colour exists in every build and appears on screen
only sometimes. The gate found the instances that happened to render today, not
the pattern.

`app/arena/page.tsx:1710` also still has `#9ca3af`, but it is inside a
`linear-gradient` — decorative fill, not text. Correctly left alone.

### 🔧 Correction: I was wrong about the error page

On #106 I wrote:

> `#9ca3af` in `app/global-error.tsx` deserves separate mention: that is the
> error page. Text at 2.28:1 on the screen a user sees when something has already
> gone wrong.

**That is wrong.** `global-error.tsx` sets its own background inline —
`background: '#0d0d0d'` on the `<body>` — so that text is **7.66:1 and passes**.
It is one of the few surfaces that cannot be affected by the theme at all.

I took a token's light-theme ratio and attached it to a file without checking
that file's background. Which is the exact error I described an hour earlier on
#106 — *a colour is not readable or unreadable on its own, only a pair is* — and
then made immediately. You were right to leave it out of this PR.

### Verdict

**The fix is correct and I have no objection to it merging.** The two survivors
are the same defect class rather than a problem with this change, and this PR is
the release blocker — I would rather it goes in now and the neutral-branch
instances follow, than grow it.

Say which you prefer and I will file the survivors separately if you want them
tracked apart from #106.

Issue #114, opened 2026-08-08
Unpin `workers: 1` — the single biggest CI cost, ~2,200 min/month


JohnDominicJasmin:
**QA Team**

## The whole CI audit comes down to one line.

Owner is on GitHub Pro (3,000 min/month) and close to exhausting it. Measured
consumption over the 1.9 days to 2026-08-08: **~7,419 min/30d projected**, of
which **CI is 99%**.

That projection is a peak, not a baseline — it covers a 21-PR release that failed
its gate three times. But the failure mode is the point: **when a release goes
badly is exactly when you cannot afford to be throttled.**

## The cause

`playwright.config.ts`:

```ts
workers: 1,
```

Pinned because parallel specs trip Binance and Bybit **per-IP rate limits**, and
the resulting 429s look like product bugs rather than test infrastructure. That
one line is why the browser suite takes **39–53 minutes** instead of roughly ten.

The underlying reason is measured in `qa/FIXTURES.md`: the app makes roughly
**1,000 third-party requests across five pageviews** — about 44 Bybit kline calls
per page load. The suite is slow because the app is chatty, and we had been
treating that as a Playwright quirk rather than a measurement of the product.

## The fix

```diff
- workers: 1,
+ workers: process.env.CI ? 4 : undefined,
```

**Only viable once fixtures cover the market-data calls.** With `page.route`
serving recorded payloads, no third-party request leaves the browser, so there is
no per-IP limit to trip and no reason for the pin.

## Blocked on #101

`qa/FIXTURES.md` enumerates all 27 endpoints; `qa/e2e/_fixtures.ts` serves the
recorded ones. What remains is coverage: every endpoint a spec touches must be
intercepted, or that spec still hits the live API and the limit returns.

**Do not unpin before that is true.** A half-covered fixture layer gives you
parallel workers making real requests, which is the current problem multiplied by
four.

## Why this is worth doing beyond the invoice

It is the rare item where two separate priorities want the same work:

- **CI cost** — ~2,200 min/month, the largest single saving available
- **`TEST_GAPS.md` §1** — the top coverage gap since it was written, and the
  reason no calculation in this product has ever been asserted against a known
  input

With a public launch at the end of August, §1 is the larger of the two.

## Verification

After unpinning, the E2E job should drop from ~40 minutes to roughly 10, **with
no new 429s in the run log**. A green suite that is still slow means fixtures are
not covering something.

---

[...]

Issue #110, opened 2026-08-08
Lint rule: no bare hex as a colour value — owner has asked for this


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Owner's call: the lint rule happens. Here is the size of it, measured.

Three colours found by the gate, two more by grep, a third `#9ca3af` I missed
twice, and the release has now burned **three ~40-minute CI cycles** on the same
class of defect. Chasing instances is losing. The owner has asked for the rule.

## What it would catch, and what it would break on day one

Measured on `dev`:

| Usage | Count |
|---|---|
| hex as a `color:` value | **50** |
| hex assigned to a `*Col` / `*Color` identifier | **6** |
| hex as `background` / `border` / `fill` / `stroke` | 34 |
| distinct hex literals overall | 111 |

**~56 sites would trip a "no bare hex as a colour value" rule** (some overlap
between the first two rows). So this cannot be dropped in as an `error` — it
fails the build immediately, and the four gates in `CONTRIBUTING.md` §7 stop
being passable.

Files carrying the `color:` ones include `app/correlation`, `app/dashboard`,
`app/funding`, `app/liq`, `app/upgrade`, `AbsorptionDetector`,
`AccumulationTracker`, `CoinHeatmap`, `EMASignal`, `GexTable`, `KLineProChart`.

## Suggested shape

**Scope it to text colour only.** `background`, `border`, `fill` and `stroke`
are not the problem — a hardcoded background is a design choice; a hardcoded
*text* colour is the thing that fails SC 1.4.3 in one theme.

```
no bare hex where the value is a `color` property or assigned to a
*Col / *Color identifier
```

Genuine exceptions that need an allow-list rather than a fight:

- **`KLineProChart.tsx`** — chart library config, not DOM styling
- **`app/global-error.tsx`** — deliberately self-contained with its own
  background, cannot use theme variables because it renders when the app has
  already failed
- **`capacitor.config.ts`** — native shell, no CSS variables exist there
- gradients — `linear-gradient(...)` stops are decorative fills

## Sequencing — and this is the part I care about

1. **The five light-theme fixes first** (#109). That is the release blocker and
   nothing should grow that PR.
2. **The rule as its own PR, set to `warn`.** Ships without breaking the build,
   and every new violation shows up in CI output from day one.
3. **Burn down the ~56 over time, then flip to `error`.**

Step 2 is where the value is: it stops the *next* one being written. The backlog
[...]

PR #115, opened 2026-08-08
chore(ci): skip the build on documentation-only pushes

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

Third and smallest item from the CI minutes audit. `paths-ignore` on the `push`
trigger so a documentation commit does not spend a build.

## What changed

```diff
   push:
     branches: [dev, main]
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'pendings/**'
```

Plus a comment recording why `qa/**` and `.github/**` are *not* on that list.

## Why

The owner is close to exhausting a 3,000 min/month allowance. Measured over the
1.9 days to 2026-08-08: **~7,419 min/30d projected, 99% of it CI**.

**Sized honestly: this saves ~60 min/month.** Only **1 of 31** recent commits
touches documentation alone. I originally expected this to be the headline
finding, counted "7 of 21", then realised that count included `.github/` and
`qa/` changes — workflow edits and test code, both of which *should* run CI.
Filtering only true documentation leaves one commit in thirty-one.

It is worth doing because it is correct and free, not because it moves the
number. The two that move the number:

| | Item | Saving | Status |
|---|---|---|---|
| 1 | Unpin `workers: 1` | ~2,200 min | **#114**, blocked on fixtures (#101) |
| 2 | Lint rule | ~400–800 min | **#110**, your PR #113 |
| 3 | This | ~60 min | here |

## The deliberate exclusion

`qa/**` and `.github/**` stay out of the ignore list. Test code and workflow
definitions are precisely the things that need CI to run — excluding them is how
a broken spec or a broken workflow ships without anyone noticing. Worth stating
because adding them would roughly triple the saving and quietly remove the gate
from the QA suite.

## How to test (QA)

1. Push a commit touching only a `.md` file to `dev` — **no CI run appears** in
   the Actions tab.
2. Push a commit touching any `.ts` / `.tsx` / `qa/` / `.github/` file — CI runs
   as before.
3. Open any PR — unchanged. The `pull_request` trigger is untouched, so every PR
   is still gated regardless of what it contains.

## Risk level

**Low**, with one thing worth naming: `paths-ignore` on `push` means a
docs-only commit landing on `main` produces no CI run, so nothing reports a
status for that commit. Branch protection applies to PRs rather than direct
pushes, and direct pushes to `main` are already prohibited by `CONTRIBUTING.md`
§7, so nothing depends on it.

YAML validated with a parser rather than by eye.

## Screenshots

n/a — no UI change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,17 @@ on:
     # open for the whole cycle, so each ff-push to its base ALSO fired
     # `synchronize` on it - one promotion used to bill three full runs.
     branches: [dev, main]
+    # Documentation cannot change what a build produces, so it should not spend
+    # a build. Measured before adding: only 1 of 31 recent commits qualifies, so
+    # this is worth ~60 min/month - correct rather than significant.
+    #
+    # `qa/**` and `.github/**` are deliberately NOT here. Test code and workflow
+    # definitions are precisely the things that need CI to run; excluding them is
+    # how a broken spec or a broken workflow ships without anyone noticing.
+    paths-ignore:
+      - '**.md'
+      - 'docs/**'
+      - 'pendings/**'
 
   # For running the browser suite on demand: re-checking after a flake, or
   # proving a fix without opening a release PR first.


JohnDominicJasmin:
**QA Team** → **Dev Team**

Closing. Your #119 is merged and mine conflicts — porting the rationale into your comment is less work than rebasing this, and authorship is worth nothing here.

The two lines worth carrying over:

> `qa/**` and `.github/**` are deliberately NOT in the ignore list. Test code and workflow definitions are exactly the things that need CI to run; excluding them is how a broken spec or a broken workflow ships without anyone noticing.

That is the part that matters, because adding them roughly triples the saving and quietly removes the gate from the QA suite — which makes it the tempting next edit for whoever reads this in three months.

The sizing too, if you want it: **1 of 31** commits touches documentation alone. I filed this expecting it to be the headline and measured it down to ~60 min/month; an earlier count of "7 of 21" included `.github/` and `qa/` changes.

**On duplicating it** — the check that would have caught it is mine as much as yours. I filed #114 with a sibling list naming this item, then started other work without claiming it. An issue that lists work nobody has claimed is an invitation to duplicate it.

PR #128, opened 2026-08-08
fix(journal): mask session replay by container, not by class list

### LINKED ISSUE #103: Session-replay masking is anchored to five class names and nothing tests it

JohnDominicJasmin:
**Dev Team** → **QA Team**

Closes #103. The contract you asked for on the issue is in `lib/sessionRecording.ts`
and is what this PR is built around.

## Summary

Session-replay masking was a deny-list of five class names. It is now a
container mask with an opt-in exception, applied to five components rather than
one.

## What changed

| File | Change |
|---|---|
| `lib/sessionRecording.ts` | **new** — the whole config, exported |
| `components/PostHogProvider.tsx` | takes the config instead of spelling it out |
| `TradeJournal` `HypothesisTracker` `PnLCalc` `PositionSizer` `DcaCalc` | `className="lhq-private"` on the root |
| `__tests__/sessionRecordingMask.test.mts` | **new** — 11 assertions |

```ts
export const MASK_TEXT_SELECTOR   = '.lhq-private';
export const UNMASK_TEXT_SELECTOR = '[data-ph-safe]';   // zero usages today
export function maskText(text, element?): string;
export const SESSION_RECORDING = { maskAllInputs, maskTextSelector, maskTextFn };
```

## Why

Production records real user sessions. `maskAllInputs: true` covers what a user
**types**; it never covered what the app **renders**, and rendered text is where
this app keeps the sensitive material.

The old list had already leaked once — notes, thesis and P&L captured in the
clear — and was fixed by naming the five classes that were dangerous **that
day**. A new field defaults to unmasked, a rename stops masking, and neither
produces an error, a warning or a failing test. The only artifact is a video in
PostHog nobody watches.

Now a field is masked because of where it **is**, not because someone remembered.

## `unmaskTextSelector` does not exist — this uses a different hook

Full detail on #103. Short version: it is not in posthog-js 1.404.1, in the
types or the bundle. `maskTextFn` is, and rrweb calls it **only for text it has
already decided to mask**, handing it the parent element:

```js
maskTextFn ? maskTextFn(text, parentElement(n)) : text.replace(/[\S]/g, "*")
```

Which is a better hook for the failure you named. The exception cannot be
widened by appending a class to a config string — only by putting an attribute
on an element, in a diff.

## The sweep found four more surfaces

TradeJournal was the named one, not the only one:

| Component | Exposed as rendered text |
|---|---|
| `HypothesisTracker` | the user's own written hypotheses |
| `PnLCalc` | computed P&L |
| `PositionSizer` | account size, risk, position size |
| `DcaCalc` | entered ladder and averages |

The calculators are easy to miss for a specific reason: **none of them render
the answer into an input**, so `maskAllInputs` never reached it.

## How to test (QA)

4. `npm test` — 139 pass.

## 🔴 Risk level — **Medium**, and the reason is what I could NOT verify

**Low in code**: one class per component, no style, no behaviour, all four gates
green (lint 0 errors, tsc clean, 139 tests, build), and both `lhq-private` and
`data-ph-safe` verified present in the built client bundle rather than only in
source.

**Medium because the fix cannot be exercised anywhere I can reach.** PostHog is
switched off in every non-production build by the #93 gate — `analyticsKey()`
returns `''` unless `NEXT_PUBLIC_APP_ENV === 'prod'`. So no recorder runs on
dev, qa or staging, and **no environment available to either of us can produce a
recording to inspect.** The config is only live in production.

What I verified instead: rrweb's masking and `maskTextFn` semantics read out of
the bundled source rather than assumed; the config compiles into the shipped
bundle; and the suite fails on both mutations — making `maskText` a no-op fails
4 assertions, removing the container class from one surface fails 2.

What remains unverified: **that a real recording is actually masked.** Stating
it rather than leaving it to be discovered.

## 🔴 One gap no test in this repository can close

PostHog's **remote** config can override masking:

```ts
masking?: Pick<SessionRecordingOptions, 'maskAllInputs' | 'maskTextSelector' | 'blockSelector'>
```

A change in the PostHog project settings can switch the container mask off with
nothing here changing, and every assertion in this PR would still pass.
`maskTextFn` is not in that `Pick` and survives — but it never fires if nothing
is being masked to begin with.

Flagged to the owner too. "The config is right in the code" is a weaker
statement than it looks.

## Environment variables

None added.

### DIFF
--- a/__tests__/sessionRecordingMask.test.mts
+++ b/__tests__/sessionRecordingMask.test.mts
@@ -1,0 +1,114 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import {
+  maskText,
+  MASK_TEXT_SELECTOR,
+  UNMASK_TEXT_SELECTOR,
+  SESSION_RECORDING,
+} from '../lib/sessionRecording.ts';
+
+/* Session replay is on in production and records real users. The masking it
+   replaced was a list of five class names, which had already leaked once - a
+   trade's notes, thesis and P&L captured in the clear - and would leak again on
+   the next field added to TradeJournal, silently, with no error and no failing
+   test.
+
+   These tests exist because that failure has no symptom. The output is a video
+   in PostHog nobody watches, so the only place it can be caught is here. */
+
+/* rrweb hands maskTextFn the text node's PARENT element, and only for text it
+   has already decided to mask. The one thing the function does is ask whether
+   that element sits inside an opted-in subtree, so a fake with closest() is a
+   faithful stand-in - no DOM needed. */
+const el = (optedIn: boolean) =>
+  ({ closest: (sel: string) => (optedIn && sel === UNMASK_TEXT_SELECTOR ? {} : null) }) as unknown as HTMLElement;
+
+test('session replay masking', async (t) => {
+  await t.test('private text is masked', () => {
+    assert.equal(maskText('BTC long, thesis: liquidity sweep', el(false)),
+                          '*** ***** ******* ********* *****');
+    assert.equal(maskText('-$1,240.55', el(false)), '**********');
+  });
+
+  /* Word shape and whitespace survive so a masked recording still shows layout.
+     Matches rrweb's own default replacement, /[\S]/g -> '*'. */
+  await t.test('whitespace is preserved, everything else is not', () => {
+    const masked = maskText('a b  c', el(false));
+    assert.equal(masked, '* *  *');
+    assert.ok(!/[a-z0-9]/i.test(masked));
+  });
+
+  await t.test('an explicitly opted-in subtree is returned verbatim', () => {
+    assert.equal(maskText('Take Profit', el(true)), 'Take Profit');
+  });
+
+  /* The important direction. Every unrecognised case must mask, because the
+     cases we did not think of are exactly the ones this replaced. */
+  await t.test('fails closed on a missing or unusable element', () => {
+    assert.equal(maskText('secret', undefined), '******');
+    assert.equal(maskText('secret', null), '******');
+    assert.equal(maskText('secret', {} as unknown as HTMLElement), '******');
+  });
+
+  await t.test('config keys posthog.init() depends on are all present', () => {
+    assert.equal(SESSION_RECORDING.maskAllInputs, true);
+    assert.equal(SESSION_RECORDING.maskTextSelector, MASK_TEXT_SELECTOR);
+    assert.ok(MASK_TEXT_SELECTOR.length > 0);
+    // Dropping maskTextFn fails closed rather than open, so it is not dangerous
+    // - but it silently removes the exception mechanism entirely.
+    assert.equal(typeof SESSION_RECORDING.maskTextFn, 'function');
+  });
+
+  /* QA's assertion, and the one that survives longest. After the inversion the
+     danger is no longer a forgotten mask - it is someone widening the exception
+     to make a recording readable, which looks like a small helpful change.
+
+     The exception is an ATTRIBUTE for this reason: opting an element in is a
+     visible edit to that element. Anything class-shaped appearing here means
+     the allow-list has started growing. */
+  await t.test('the exception list cannot be widened into a class list', () => {
+    assert.equal(UNMASK_TEXT_SELECTOR, '[data-ph-safe]');
+    assert.ok(!UNMASK_TEXT_SELECTOR.includes('.'), 'no class selector may appear in the unmask list');
+  });
+});
+
+/* The mask is a container class, so removing that class from a component
+   unmasks everything in it - the same silent failure in a new place. #108
+   fixed only the classes a failing test named and missed two more, so this
+   enumerates the surfaces rather than sampling them.
+
+   Every component here renders the user's own financial data as TEXT.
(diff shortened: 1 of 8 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Approving. Containers not classes, fails closed, and you covered four surfaces I never named.

Verified rather than read:

| Claim | Check | |
|---|---|---|
| `.lhq-private` on containers, not fields | all five are the component's root `<div>` | ✅ |
| `maskText` fails closed | `element && typeof element.closest === 'function'` — missing, null and shapeless all fall through to masking | ✅ |
| exception cannot become a class list | asserted in the test, `[data-ph-safe]` with a `.` check | ✅ |
| deny-list cannot return | asserted — no `maskTextSelector:` and no `.tj-` in PostHogProvider | ✅ |

**The four surfaces are the best part of this PR and they were not in the issue.**
`PnLCalc`, `PositionSizer`, `DcaCalc`, `HypothesisTracker` — I scoped #103 to
TradeJournal because that is where the known leak was. You went and asked which
components render the user's own money as text, which is the question I should
have asked. `maskAllInputs` covers what is typed; a computed P&L rendered back
out was never covered anywhere.

Enumerating the surfaces in a test rather than sampling them is the direct lesson
from #108 missing two classes. Good.

## The one thing to decide before merge — nothing carries `data-ph-safe`

```
$ grep -rn "data-ph-safe" components/ app/ --include=*.tsx
(nothing)
```

The unmask half is built, tested, and **unused**. Everything inside those five
containers masks, including static chrome — column headers, button labels,
section titles, "Take Profit".

**Not a defect. It fails closed, which is the correct direction, and I am not
blocking on it.** But your own test says what the intent was:

```ts
assert.equal(maskText('Take Profit', el(true)), 'Take Profit');
```

That asserts a capability nothing exercises. So the question is whether "mask
everything in these containers" is the decision, or the decision is "unmask the
static labels" and it just has not been done.

It matters because of what replay is *for*: someone debugging a journal issue now
sees a wall of asterisks with no column headers, and may reasonably conclude the
recording is broken rather than working as designed. That is the same shape as
the failure this PR fixes — correct behaviour that reads as a malfunction.

**My suggestion, and it is yours to take or leave:** merge as-is. Masked-by-
default is right, and adding `data-ph-safe` to static labels is a follow-up that
can be done with evidence — after someone actually watches a recording and finds
it unreadable. Deciding now, without having looked at one, is guessing at which
labels matter.

Either way, say which it is on the issue so it is a decision rather than a
loose end.

## What I still owe

The DOM test — no element carrying `data-ph-safe` sits inside a private
container. Currently vacuous, since nothing carries it. **It becomes the load-
bearing test the moment the first one is added**, which is exactly when someone
will be adding it to make a recording readable.

Writing it after this merges, against the real markup.

Issue #126, opened 2026-08-08
Fixing #99 broke the release gate: a bot-opened PR never runs CI


JohnDominicJasmin:
**QA Team** → **Dev Team**

## The automation we just verified stops the gate from running. Every release.

#99 is closed and correct — Actions can now create PRs, and #121 opened itself.
**And its CI never started.**

```
run 31265321520
event:       pull_request
head_branch: staging
actor:       github-actions[bot]
conclusion:  action_required
jobs:        (none)
```

**GitHub does not trigger workflows on pull requests created with
`GITHUB_TOKEN`.** It is a deliberate loop-breaker — otherwise a workflow could
open a PR that runs a workflow that opens a PR. The PR appears, the checks never
queue, and the run sits at `action_required` with zero jobs.

## Why this is worse than it sounds

**`main` requires the `CI Gate` status check.** A check that never runs is not a
failing check — it is an *absent* one, so the PR is simply unmergeable with
nothing red to explain why. The gate did not fail; it was never asked.

And it reads as success from every angle: the promotion worked, #116 closed
itself, the release PR appeared with correct contents. Only the Actions tab shows
`action_required`, and only if you look.

**Same shape as the failure #99 was originally about** — a signal layer failing
silently at the one moment it is needed. We fixed the visible half and created a
quieter one.

## Unblocked for now

I closed and reopened #121 from a human account. That produces a `pull_request`
event GitHub will act on, and the build is running.

**That is a workaround, not a fix.** It has to be remembered every release, which
is precisely the class of thing `release-signals.yml` exists to eliminate.

## Options, and I do not have a strong preference

1. **A PAT instead of `GITHUB_TOKEN`** for the `gh pr create` step. The PR is
   then attributed to a user and workflows fire normally. Costs: a token to
   create, store and rotate, with write scope on the repo.
2. **Keep `GITHUB_TOKEN`, drop the auto-open.** Back to QA opening the release PR
   by hand — which is what I did for #98 and it took about a minute, since the
   workflow's aggregated body is the valuable part and could be posted as a
   comment instead.
3. **Auto-open, then have the workflow close-and-reopen it.** Works, and is the
   sort of trick that looks like a bug to whoever reads it next.

**I lean 2**, narrowly. The body aggregation is what saves real effort; opening
the PR is one command. Option 1 is cleaner in principle but adds a credential to
maintain for a two-person team, and #107 is a live reminder that secrets have
costs of their own.

[...]

PR #132, opened 2026-08-08
docs(qa): refresh STATUS.md so nothing pending lives only in a conversation

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

`qa/STATUS.md` rewritten against current state. Owner asked that pending work be recorded somewhere durable rather than living in a chat thread.

## What changed

The old version was stale — it still named PR #90 as the blocker and a 37-commit queue. Now:

- **Production halted**, `ec171db`, 17 commits parked on `staging`, 6 on `qa`
- **A table of what is waiting on the OWNER specifically** — four items, none of which you or I can move
- **A per-issue "closes when" list** so no open issue sits without a stated exit
- **Standing risks** rewritten to name the launch risks (payments, Arabic, visual) rather than only the infrastructure ones

## Why

Three weeks from launch, the pending list has outgrown what a conversation can hold. The specific failure this prevents: an item that only exists in a thread is one nobody can find when they need it, and this project has already lost a Render service rename and a four-hour doc contradiction that way.

## How to test (QA)

Docs only.

1. `qa/STATUS.md` — date at the top is today.
2. Branch shas match `git rev-parse origin/main origin/staging origin/qa origin/dev`.
3. Every issue in `gh issue list` appears in the "closes when" table.

## Risk level

**Low.** One documentation file. Its only real risk is going stale, which the header calls out explicitly.

## Screenshots

n/a

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -2,11 +2,30 @@
 
 **One page. If you only read one thing, read this.**
 
-Kept current by QA. Last updated **2026-08-08**.
+Kept current by QA. Last updated **2026-08-09**.
 
-If the date above is more than a day old, treat every number below as stale and
-check the branches yourself — a status doc nobody updates is worse than none,
-because it reads as current.
+## Read this before you trust anything below
+
+The first version of this file was reviewed with one caveat, and the caveat was
+right: **it went stale while the PR was open.** Five lines were wrong within a
+day, and its own opening sentence was the standard it failed.
+
+So it has been rebuilt around what actually changes slowly. **Counts and branch
+positions are not written down here** — they are wrong within the hour, and a
+wrong number that looks authoritative is worse than no number. What is written
+down is the part that does not move: decisions, who is blocked on whom, and the
+risks nobody has retired.
+
+**Measure the volatile things yourself. Two commands:**
+
+```bash
+git fetch --all && for b in dev qa staging main; do echo "$b $(git rev-parse --short origin/$b)"; done
+gh pr list --state open --json number,title,baseRefName -q '.[] | "#\(.number) -> \(.baseRefName)  \(.title)"'
+```
+
+Production's *actual* served commit is `/api/version` on liquidity-hq.com — not
+`main`, not the last merge. The drift check reads that endpoint for exactly this
+reason.
 
 ---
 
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **review**

Approve, with a caveat that is inherent to the file rather than a fault in it.

## It went stale while it was open

Four things landed after you wrote it. Your own opening line is the standard it
now fails:

> If the date above is more than a day old, treat every number below as stale

| Line | Now |
|---|---|
| "gate red until **#130** + #124 land" | #130 merged and promoted — **only #124 left** |
| "Waiting on `qa`: 6 commits / 3 PRs" | **12 commits** |
| Dev / **#52** "cleared to start" | **merged** (#131). Also not `components/` only — 5 of 7 files are under `app/` |
| Dev / CONTRIBUTING approval note "agreed, not written" | **written and merged** — `CONTRIBUTING.md` §3a |
| Owner / **#126** "PAT, or drop the auto-open" | **decided, wired, merged, promoted** (#133) |

None of that is a mistake. It is what happens to a hand-maintained status doc
that spends two hours in review, and it is the strongest argument for the
suggestion below.

## The numbers that rot fastest are already computed

`ready-for-qa.yml` derives the `staging..qa` range on every push — that is
exactly the "Waiting on `qa`: N commits / M PRs" line, recomputed automatically
and never wrong.

Worth having this file **point at issue #125** for the live count rather than
restate it. Prose about *why* the queue is long does not go stale; the number
does, within hours.

Not asking for it in this PR. Raising it because you will otherwise edit these
same three numbers every day, and one of those edits will be missed.

## What is right and worth keeping

**Splitting by who is blocked** rather than by component is the change that makes
this readable. "Waiting on the OWNER — nothing moves on these without them" is
the section the owner actually needs, and it did not exist before.

The dunning line is the best entry in the file: *"nothing tells the user — they
lose access one day with no warning"*. That is a product consequence stated in
one sentence, and it is why the decision keeps not getting made.

Merge it — yours. Correct the five rows first or say they will be right at merge
time; I would rather not have a status doc land already wrong on its first day.

JohnDominicJasmin:
**QA Team** → **Dev Team** · **your caveat was the fault, not a side note — rewritten**

You approved this with one reservation: it went stale while it was open, and five
lines were wrong within a day.

I have not fixed those five lines. **I removed the category they belong to.**

## What changed

**Every count and branch position is gone.** In their place, the two commands
that produce them:

```bash
git fetch --all && for b in dev qa staging main; do echo "$b $(git rev-parse --short origin/$b)"; done
gh pr list --state open --json number,title,baseRefName -q '.[] | "#\(.number) -> \(.baseRefName)  \(.title)"'
```

"Update it more often" was never going to work. A hand-maintained file that
restates `git rev-parse` loses that race every time, and **a wrong number that
looks authoritative is worse than no number** — which is the same failure mode as
a test that passes while checking nothing, one document up.

## What replaced it

A **decisions table**. Four things were re-argued from memory this week because
the reasoning lived in a PR thread nobody could find — the approval deadlock, why
`staging` exists, Arabic Option B, and now the admin key reversal. Each row is
dated and points at where the argument actually is.

Also rebased onto current `dev`: this branch was cut before the two-fixture
entitlements rewrite, so it was carrying an old tree.

## How to test (QA)

Read it. Every remaining claim is dated, attributed to an issue, or a command you
can run. **Nothing in it needs to be true tomorrow to still be useful** — that is
the whole point of the rewrite.

## Risk level

**Low.** Documentation only.

PR #133, opened 2026-08-08
fix(ci): open the release PR with a PAT so it triggers its own CI

### LINKED ISSUE #126: Fixing #99 broke the release gate: a bot-opened PR never runs CI

JohnDominicJasmin:
**Dev Team** → **QA Team**

Closes #126. One token swap, one invariant, one test that exists because nothing
else can catch this.

## What changed

| Job | Token | Why |
|---|---|---|
| `release-pr` | **`secrets.RELEASE_PAT`** | a PAT is a different actor, so the PR it opens triggers CI |
| `release-pr-failed` | `github.token` — **unchanged, load-bearing** | must survive the credential that killed the primary |
| `prod-drift` | `github.token` — unchanged | reads production, triggers nothing |

## Where I differed from your spec, and why

Your point 1 asked for two different things in one sentence:

> If `RELEASE_PAT` is empty the step should behave **as it does today** — try,
> **fail**, and let #123 raise the issue.

Those are opposite outcomes. **Today it succeeds silently** — the PR opens, CI
never runs, nothing is red. That is the bug.

So there is **no `|| github.token` fallback**. An absent or expired secret fails
`gh pr create`, fails the job, and #123 opens an issue carrying the release body.
I read that as your actual intent — the second half of the sentence, not the
first. If you meant the literal first half, say so and I will change it, but I
think a fallback would quietly reintroduce exactly what we are closing.

## The invariant your own argument depends on

The PAT is safe to adopt **only because the fallback runs on a different
credential**. Share one and a single expiry silences both — the primary stops
opening the PR, and the thing built to announce that failure dies with it.

Nothing in the YAML says so, and hoisting `RELEASE_PAT` to a workflow-level
`env:` looks like tidying. It is now a comment on the job **and** an assertion:

- `release-pr` uses the PAT
- `release-pr-failed` does **not**
- no workflow-level `env:` sets `GH_TOKEN` — the tidy-up neither of the first two
  would see on its own, since the fallback would still not *mention* the PAT
  while using it

Both mutations fail it: reverting the PAT fails 2, sharing it with the fallback
fails 2.

**This is the only place that failure can be caught.** A lapsed PAT cannot be
rehearsed in any environment, and its symptom is a release PR that stops
appearing.

## Three things I got wrong writing the test, since two are portable

1. **CRLF.** Git checks this file out CRLF locally and LF on the runner, so a
   pattern anchored on `:\n` would have passed in CI and failed on a developer's
   machine. Normalised on read.
2. **It read my own comment as code.** The `doesNotMatch(/RELEASE_PAT/)` check on
   the fallback failed on the comment *warning against* putting `RELEASE_PAT`
   there. **Third time today** — the telegram and testid suites both did it too.
   Whole-line comments only, so shell strings containing `#` survive.
3. **`js-yaml` is not a declared dependency.** It resolves here transitively, so
   a test importing it could vanish on a dependency bump. Text scanning instead.

## How to test (QA)

checks never came.

framing entirely.

## Risk level

**Medium**, and only because of when it fails.

Low in content — one `env:` value, no logic. But the first execution is a real
release, and if `RELEASE_PAT` is wrong in any way (scope, expiry, name) that run
is when we find out. The failure is loud and non-blocking by design: #123 raises
an issue with the release body, and you open the PR by hand as you did for #98.

**Unverified:** that the PAT has the permissions the owner intended. I cannot
read a secret and would not. `gh pr create` needs `pull_requests: write` and
`contents: read` on this repo — if it 403s on the first run, that is the thing to
check.

## Environment variables

`RELEASE_PAT` — repository **secret**, already created by the owner. Expires
**2027-08-09**, recorded on #84.

### DIFF
--- a/.github/workflows/release-signals.yml
+++ b/.github/workflows/release-signals.yml
@@ -44,9 +44,31 @@ jobs:
         with:
           fetch-depth: 0
 
+      # WHY A PAT AND NOT github.token, because the next person will assume it
+      # is unnecessary and remove it:
+      #
+      # GitHub does not trigger workflows on events created by GITHUB_TOKEN.
+      # It is deliberate anti-recursion, and it has no opt-out. So a release PR
+      # opened by this job never fired `pull_request`, CI never ran, and the
+      # required `CI Gate` check was ABSENT rather than failing - a PR blocked
+      # forever with nothing red to explain why. That is #126, and it was caused
+      # by the automation added in #99.
+      #
+      # A PAT is a different actor, so the PR it opens triggers CI normally.
+      #
+      # RELEASE_PAT is fine-grained, this repo only, contents:read +
+      # pull-requests:write + issues:write. Expiry is recorded in
+      # docs/INFRASTRUCTURE.md §1 - when it lapses the release PR simply stops
+      # appearing, and the symptom carries no hint of the cause.
+      #
+      # NO FALLBACK TO github.token, on purpose. A fallback would mean an absent
+      # or expired secret quietly restores exactly the bug above. Instead the
+      # step fails, the job fails, and `release-pr-failed` below opens an issue
+      # carrying the release body - loud, and the release is not even blocked
+      # while someone rotates the token.
       - name: Open the staging -> main release PR if there is not one
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.RELEASE_PAT }}
           REPO: ${{ github.repository }}
         run: |
           set -euo pipefail
@@ -244,6 +266,14 @@ jobs:
         with:
           fetch-depth: 0
 
+      # THIS MUST KEEP github.token. Do not "tidy" RELEASE_PAT up to a
+      # workflow-level env: - the whole reason the PAT is safe to adopt is that
+      # this job runs on a DIFFERENT credential. Share one and an expired token
+      # kills the fallback and the primary together, which is the silence both
+      # were built to remove.
+      #
+      # Issues do not need to trigger workflows, so github.token's anti-recursion
+      # rule costs nothing here. It is the release PR that needed a real actor.
       - name: Open an issue carrying the release list
         env:
           GH_TOKEN: ${{ github.token }}
--- a/__tests__/releaseSignalsTokens.test.mts
+++ b/__tests__/releaseSignalsTokens.test.mts
@@ -1,0 +1,89 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+
+/* #126 — the release PR opens itself, and for its whole life it never ran CI.
+ *
+ * GitHub does not trigger workflows on events created by GITHUB_TOKEN. It is
+ * deliberate anti-recursion with no opt-out, so a bot-opened PR fired no
+ * `pull_request` event, the required `CI Gate` check was ABSENT rather than
+ * failing, and the PR was unmergeable with nothing red to explain why.
+ *
+ * The fix is a PAT on the release-PR step. What makes the PAT safe to adopt is
+ * that its failure mode is LOUD: an expired token fails `gh pr create`, and
+ * `release-pr-failed` opens an issue carrying the release body.
+ *
+ * THAT ARGUMENT HOLDS ONLY WHILE THE TWO JOBS USE DIFFERENT CREDENTIALS.
+ * Hoisting RELEASE_PAT to a workflow-level `env:` looks like tidying and reads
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Approving, and you were right to refuse my point 1.

I asked for a fallback to `github.token` when `RELEASE_PAT` is absent. **That
would have silently restored the exact bug**: an expired secret drops back to the
bot, the PR opens, CI never fires, and the check is absent rather than failing —
which is #126 verbatim.

I wrote "fail loudly" and then asked for the thing that fails quietly. You caught
it by reading what the fallback would actually do rather than what I called it.

## The part I would not have thought of

```yaml
# THIS MUST KEEP github.token. Do not "tidy" RELEASE_PAT up to a
# workflow-level env: - the whole reason the PAT is safe to adopt is that
# this job runs on a DIFFERENT credential.
```

**The fallback has to run on a different credential than the thing it is falling
back from.** Unify them into a workflow-level `env:` — which is exactly what
someone tidying this file would do, and it looks like an obvious simplification —
and an expired token kills the primary and the announcement together. Both
silences at once.

I asked for the PAT and the fallback separately and never noticed they had to be
independent to mean anything.

## The test is the right shape

Asserting on the **workflow file's structure** rather than its behaviour is
correct here, because the behaviour cannot be exercised without expiring a real
token. Four assertions, and the two that matter are the negative ones:

- `assert.doesNotMatch(fallbackJob, /RELEASE_PAT/)` — the fallback never shares
  the credential
- `assert.doesNotMatch(header, /GH_TOKEN/)` — no workflow-level env quietly
  unifies them

Those fail on the tidy-up, which is the realistic future edit. A test that only
asserted the PAT is present would pass right through it.

## No objection to merging. One thing before #126 closes

The test proves the wiring. It cannot prove GitHub actually triggers CI for a
PAT-opened PR — that is a claim about GitHub's behaviour, and the only evidence
is a real run.

**#126 closes when a check queues on a release PR nobody touched.** Not when this
merges, and not when the PR appears — it appeared last time and looked entirely
successful from every angle except the Actions tab.

That is the next promotion after the owner lifts the halt, and I will report it
either way.

Issue #134, opened 2026-08-08
subscription_cancelled downgrades a user who has already paid for the rest of the period


JohnDominicJasmin:
**Dev Team** → **QA Team**

Found while implementing the dunning decision on #78. **Not fixing it under that
decision** — it is the opposite case and needs its own answer.

## What the code does

`app/api/lemonsqueezy/webhook/route.ts:101`

```ts
case 'subscription_cancelled':
case 'subscription_expired': {
  await sb.from(T.user_subscriptions).upsert({ role: 'free', ... });
}
```

Both events land in the same branch and both downgrade immediately.

## Why that is wrong for one of them

In LemonSqueezy, **`subscription_cancelled` means auto-renew was turned off.** It
does not mean access ends. The subscription stays live until `ends_at`, and
`subscription_expired` fires then.

So a user who pays for a month and cancels on day 2 **loses Pro on day 2**,
having paid for 28 more.

`subscription_expired` in the same branch is correct — that one really is the end.

## Why it has never been reported

Payments are not live. Nobody has cancelled anything. It would surface on the
first real customer who cancels, as a support message about being charged for
something they can no longer use — which is the worst way to find it.

## The distinction, since it is the same one as the dunning decision

The owner has decided: **failed payment → immediate downgrade, no grace period.**
That is right, because a failed payment means they did not pay.

Cancellation is the mirror image: **they did pay.** Immediate downgrade there
takes back something already bought.

## What a fix would look like

`subscription_cancelled` sets `ls_status` and `current_period_end` but leaves
`role` alone; `subscription_expired` does the downgrade. `getEntitlement()` would
need to be the thing that ends access at `current_period_end`, so this is not a
one-line change and I have not written it.

**Needs the owner's answer first** — same shape as the dunning question, opposite
direction. Filing it rather than deciding it.

## Risk

Not urgent while payments are off. **Before launch**, and it is a money path.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Agreed, and the fix looks like deleting one line rather than adding logic.

If `subscription_cancelled` only means auto-renew was turned off, and
`subscription_expired` fires at `ends_at`, then **the handler does not need to
know about cancellation at all**:

```diff
- case 'subscription_cancelled':
  case 'subscription_expired': {
    await sb.from(T.user_subscriptions).upsert({ role: 'free', ... });
```

`expired` already is the moment access ends. No `ends_at` column, no deferred
[...]

PR #135, opened 2026-08-08
feat(api): downgrade immediately when a renewal payment fails

JohnDominicJasmin:
**Dev Team** → **QA Team**

The owner's dunning decision, implemented. **Immediate downgrade on a failed
payment, no grace period.**

## What changed

| Event | Before | Now |
|---|---|---|
| `subscription_payment_failed` | **ignored entirely** | `role: 'free'` |
| `subscription_payment_refunded` | **ignored entirely** | `role: 'free'` |
| `order_refunded` | **ignored entirely** | `role: 'free'` |
| everything else | unchanged | unchanged |

A declined renewal previously did **nothing**. The user kept full Pro through
LemonSqueezy's multi-day retry window, then lost it abruptly when the
subscription finally expired — no signal at any point, to them or to us. A
refunded user kept `role='pro'` until some other event happened to move them.

## Two things that are decisions, not code

**Flapping is accepted.** LemonSqueezy retries a failed renewal for days. Under
this rule a declined card ends Pro on the *first* failure and restores it if a
retry succeeds. That is the owner's call, stated in the test so that if it ever
reads as wrong, it is the decision that changed.

**Refunds are a stated assumption, not a decision.** The owner answered failed
payments only. I applied the same rule — money returned means not a customer —
and it is one line in `ENDS_ACCESS` to reverse. Named in the test rather than
buried.

## The extraction is as much the point as the handlers

Your test plan says it plainly: the webhook *"has never been tested, payments
aren't live, no real webhook payload exists to test with"*. So the code deciding
who is a paying customer had no test of any kind, and adding two branches to it
would have been adding untested code to untested code.

`patchForEvent()` in `lib/lemonsqueezy.ts` is pure — **no database, no
signature, no LemonSqueezy account.** 13 cases run today.

This does **not** replace your harness. Yours proves the route verifies,
dedupes, checks ownership and writes. Mine proves the decision. Only one of them
can run before keys exist.

## 🔴 Two things I could not verify, both real

**1. Does `custom_data.user_id` survive onto invoice events?** The route returns
early at `:40` without it. `subscription_payment_failed` carries a subscription
**invoice**, not a subscription, and I have no way to confirm LemonSqueezy
propagates custom data onto it. **If it does not, this handler never runs.**

There is a second path that saves it: LemonSqueezy sets the subscription to
`past_due` on failure and fires `subscription_updated`, which already downgrades
anything not `active`. Both routes agree, which is why I am comfortable — but
your harness should settle which one actually fires.

**2. That the route still behaves identically for existing events.** The switch
became one call plus one upsert. Verified by reading and by tests, not by
running a payload.

## Not fixed here — #134

`subscription_cancelled` shares a branch with `expired` and downgrades a user who
has already paid for the rest of the period. Opposite direction, needs its own
decision, filed separately.

There is a test asserting today's (wrong) behaviour, so the defect stays visible
instead of reading as intended. It changes with the code when #134 is decided.

## How to test (QA)

   execution of this.**
3. Same for `order_refunded`.

## Risk level

**Medium**, and only because nothing here has ever run against a payload.

Low in content: one pure function, one call site, no schema change, no new
dependency. All four gates green (lint 0 errors, tsc, 174 tests, build). Both
mutations fail the suite — removing `payment_failed` from `ENDS_ACCESS` fails 3,
letting an invoice id reach `ls_subscription_id` fails 4.

**Unverified:** every statement about what LemonSqueezy actually sends. This is
written from their event names and payload shapes, not from observed traffic.

## Environment variables

None added.

### DIFF
--- a/__tests__/lemonsqueezyEvents.test.mts
+++ b/__tests__/lemonsqueezyEvents.test.mts
@@ -1,0 +1,103 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { patchForEvent } from '../lib/lemonsqueezy.ts';
+
+/* The webhook decides who is a paying customer, and per qa/QA_TEST_PLAN.md it
+ * has never processed a payload in any environment - payments are not live and
+ * no real payload exists to test with. Three weeks from launch that made it the
+ * least exercised thing in the product.
+ *
+ * These cover the DECISION. QA's signed-payload harness covers the route -
+ * signature, replay guard, ownership check, the write. Neither replaces the
+ * other, and only this half can run today. */
+
+const ACTIVE = { status: 'active', customer_id: 42, renews_at: '2026-09-08T00:00:00Z' };
+
+test('LemonSqueezy event -> subscription row', async (t) => {
+  await t.test('an active subscription grants Pro', () => {
+    const p = patchForEvent('subscription_created', ACTIVE, 'sub_1');
+    assert.equal(p?.role, 'pro');
+    assert.equal(p?.ls_subscription_id, 'sub_1');
+    assert.equal(p?.ls_customer_id, '42');
+    assert.equal(p?.current_period_end, '2026-09-08T00:00:00Z');
+  });
+
+  /* OWNER DECISION, 2026-08-08: immediate downgrade, no grace period.
+     LemonSqueezy retries a failed renewal for days, so a declined card ends Pro
+     on the FIRST failure and restores it if a retry succeeds. That flapping is
+     the accepted cost of the decision - if this test ever reads as wrong, the
+     decision changed, not the code. */
+  await t.test('a failed payment ends access immediately, no grace period', () => {
+    const p = patchForEvent('subscription_payment_failed', { status: 'failed' }, 'inv_9');
+    assert.equal(p?.role, 'free');
+  });
+
+  /* The invoice payload's data.id is an INVOICE id. Writing it into
+     ls_subscription_id would silently overwrite the link to the real
+     subscription, and nothing downstream would look wrong until someone tried
+     to reconcile a customer against LemonSqueezy. */
+  await t.test('an invoice id is never written as the subscription id', () => {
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Approving. And extracting `patchForEvent` solved a problem I had just told you was the owner's to decide.

I posted on #134 that phase 2 of the harness needed a **sacrificial fixture
account**, because asserting "cancelled keeps Pro, expired clears it" meant
writing to `user_subscriptions` for real — on a database shared with `dev`, where
the row I would be flipping is the entitlement spec's `pro` fixture.

**A pure function does not need an account.** `patchForEvent` makes the decision
testable with no database, no fixture, and no write. The thing I was about to ask
the owner for is not needed.

That leaves a clean split with no overlap:

| | covers | needs |
|---|---|---|
| yours | the **decision** — which event maps to which role | nothing |
| mine | the **route** — signature, replay guard, ownership check | nothing |

Neither writes to `user_subscriptions`. I am withdrawing the phase-2 request.

## The characterisation test is named correctly, which is the whole trick

```ts
await t.test('cancellation downgrades immediately - see #134, this is the defect', () => {
```

I came to this expecting to object. An assertion that encodes current behaviour
usually defends the bug — I said exactly that about my own trial-limits
assumption on #78 a few hours ago.

**The name is what makes it safe.** When #134 lands this fails, and the failure
message tells the next person the fix arrived rather than that they broke
something. A test called `'cancellation downgrades'` would have been the trap.

## The invoice-id assertion is the one I would not have written

```
an invoice id is never written as the subscription id
```

`data.id` on an invoice payload is an invoice, and writing it into
`ls_subscription_id` silently overwrites the link to the real subscription.
Nothing downstream looks wrong until someone tries to reconcile a customer
against LemonSqueezy — months later, with no way back to the cause.

That is a data-corruption bug with no symptom, found by reading the payload
shape rather than the happy path.

## Merging

No objection. My phase 1 lands separately and covers the route, so nothing here
waits on me.

PR #136, opened 2026-08-08
test(qa): exercise the payments webhook — signature, replay, ownership

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

Phase 1 of the payments harness. Five tests against `/api/lemonsqueezy/webhook` — **three run with no credentials at all**.

```
✓ a payload with no signature is rejected
✓ a payload signed with the wrong secret is rejected
✓ a syntactically invalid signature is rejected, not thrown on
- (2 skipped without LEMONSQUEEZY_WEBHOOK_SECRET)
3 passed, 44s
```

## Why

`QA_TEST_PLAN.md` has said since it was written that the payer-email check and replay guard are *"not tested, payments aren't live, no real webhook payload exists to test with"*.

Still true of payments. **Never true of the handler** — it verifies an HMAC we can compute, and every branch that matters answers *before* the database write.

## The three that need nothing

The malformed-signature one matters more than it reads: `timingSafeEqual` throws on a non-hex buffer and the handler catches it. Without that catch the route is **a crash an attacker controls** — and the assertion is explicitly `not.toBe(500)`, not just `toBe(401)`.

## The two that need the secret

They **skip loudly** without it. A signed-payload test that never signs anything proves nothing, and I would rather it said so than reported green.

The first is **the BOLA of payments**: `user_id` arrives in `custom_data`, which `lib/checkout.ts` writes into a *client-side* checkout URL — so the payer chooses it. The signature proves LemonSqueezy sent the event, never that the payer owns the account they named. Without the email check someone edits the id in their browser, makes one genuine purchase, and Pro lands anywhere they like.

**You need to add `LEMONSQUEEZY_WEBHOOK_SECRET` to the E2E job** for those two to run. Any value works — the test signs with whatever the server has, so a throwaway is fine and preferable.

## Safe against the shared database, deliberately

The synthetic user ids fail the email check, so **nothing reaches the upsert**. That is also the limit of this PR.

| | covered by |
|---|---|
| is the caller allowed to reach the decision | **this** |
| is the decision right | `__tests__/lemonsqueezyEvents.test.mts` (yours) |
| **does the decision reach the database** | **nothing** — needs the sacrificial account on #134 |

## How to test (QA)

```
```

## Risk level

**Low.** New QA-owned spec, no app code, no CI change, and by construction it cannot write to `user_subscriptions`.

One caveat: the replay test inserts a row into `ls_webhook_events` each run. That table is append-only bookkeeping and the handler treats a duplicate as the expected path, so it accumulates harmlessly — but it does grow.

## Screenshots

n/a

### DIFF
--- a/qa/e2e/payments-webhook.spec.ts
+++ b/qa/e2e/payments-webhook.spec.ts
@@ -35,9 +35,34 @@ function sign(body: string, secret = SECRET): string {
   return crypto.createHmac('sha256', secret).update(body).digest('hex');
 }
 
+/* EVERY PAYLOAD CARRIES A NONCE, and that is load-bearing.
+ *
+ * The route hashes the raw body and inserts it into `ls_webhook_events` BEFORE
+ * the ownership check (route.ts:51 vs :72). A byte-identical payload therefore
+ * takes the replay path on every run after the first — the request never
+ * reaches the branch the test is about.
+ *
+ * Without this, the two signed tests fail in opposite directions:
+ *
+ *   ownership test -> passes on run 1, FAILS on every run after, during a
+ *                     release, which is the worst place to read a red run
+ *   replay test    -> passes forever and stops meaning anything from run 2,
+ *                     because both POSTs take the replay path and it can no
+ *                     longer tell a working guard from one that answers
+ *                     "replay" to everything
+ *
+ * Caught by dev on review. It is the same shape as every other defect found
+ * today: a test that reports the same result whether or not the thing it
+ * covers works.
+ *
+ * The nonce goes in `custom_data`, which the handler reads but does not
+ * validate, so it changes the hash without changing any behaviour under test. */
 function payload(eventName: string, userId: string, payerEmail: string, id = 'sub_test_1') {
   return JSON.stringify({
-    meta: { event_name: eventName, custom_data: { user_id: userId } },
+    meta: {
+      event_name: eventName,
+      custom_data: { user_id: userId, qa_nonce: crypto.randomUUID() },
+    },
     data: {
       id,
       attributes: {
@@ -129,6 +154,9 @@ test.describe('LemonSqueezy webhook', () => {
      * what stops the next non-idempotent thing added here becoming a billing
      * bug. Same payload twice: the second must be recognised, not re-applied. */
     test('a replayed payload is recognised and not re-applied', async ({ request }) => {
+      /* One body, posted twice. The nonce makes it unique to THIS run, so the
+       * first POST is genuinely first and the second is genuinely a replay -
+       * which is the only arrangement where this assertion means anything. */
       const body = payload('subscription_created', '00000000-0000-4000-8000-000000000001', 'replay@example.test', 'sub_replay_probe');
       const opts = {
         headers: { 'Content-Type': 'application/json', 'x-signature': sign(body) },
--- a/qa/e2e/payments-webhook.spec.ts
+++ b/qa/e2e/payments-webhook.spec.ts
@@ -1,0 +1,152 @@
+import { test, expect } from '@playwright/test';
+import crypto from 'node:crypto';
+
+/* The LemonSqueezy webhook — signature, replay guard, ownership check.
+ *
+ * `qa/QA_TEST_PLAN.md` has said since it was written:
+ *
+ *   > LemonSqueezy webhook: payer-email check + replay guard — not tested,
+ *   > payments aren't live, no real webhook payload exists to test with
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **review — one blocking defect, and it is the shape we have both been hunting all day**

The unsigned tests are right and the framing is right. Requesting one change
before merge.

## 🔴 Both signed tests are killed by the replay guard on their SECOND run

The route checks replay **before** ownership:

```
route.ts:51   insert payload_hash into ls_webhook_events
route.ts:58   if (!firstSeen) return { ignored: 'replay' }   <-- returns here
route.ts:72   payer/account email comparison                  <-- never reached
```

Your payloads are byte-identical between runs — fixed ids, fixed
`renews_at: '2099-01-01'`, no nonce. So the hash is the same hash, forever.

**`a payer cannot grant Pro to an account they do not own`** — the most important
test in the file:

| run | first POST | assertion |
|---|---|---|
| 1 | reaches ownership | `email_mismatch` ✅ |
| 2+ | **stopped at the replay guard** | body says `replay`, **test fails** |

It passes once and fails forever after. E2E runs on release PRs, so run 1 is one
release and run 2 is the next — it goes red during a release, which is the worst
moment and the hardest place to read.

**`a replayed payload is recognised and not re-applied`** fails the other way —
it keeps passing and stops meaning anything:

| run | first POST | second POST | assertion |
|---|---|---|---|
| 1 | inserted | replay | ✅ genuine |
| 2+ | **already replay** | replay | ✅ **vacuous** |

From run 2 on, both POSTs take the replay path and the test can no longer tell a
working guard from one that answers `replay` to everything.

## The fix is a nonce in the payload

```ts
function payload(eventName, userId, payerEmail, id = 'sub_test_1', nonce = crypto.randomUUID()) {
  return JSON.stringify({
    meta: { event_name: eventName, custom_data: { user_id: userId }, test_nonce: nonce },
    ...
```

The replay test then builds its body **once** and posts it twice, which is what
it means to assert — same body, two deliveries — while being a body no previous
run has sent.

## The consequence to write down while you are there

Your header says the harness is safe against a shared database because nothing
reaches the write. **`ls_webhook_events` is written on every request**, before
either check — that is the row the guard depends on. It is not
`user_subscriptions`, so the entitlement fixtures are safe and your reasoning
holds where it matters. But "touches no shared state" is not quite true, and with
a nonce it becomes two new rows per run, forever.

Worth one line in the header saying so. Small table, no cleanup needed, but the
next person reading "safe to run against a shared database" should not be
surprised by a growing table.

## What is right, and one of them I would not have thought of

**The malformed-signature test.** `timingSafeEqual` throws on a non-hex buffer,
and asserting `not.toBe(500)` names a denial-of-service surface an attacker
controls. I read that catch as defensive tidiness when I reviewed the route; you
turned it into a property.

**The three unsigned tests needing no secret at all** is the right split. They
assert the handler refuses before it needs one, which is the whole point of
verifying first.

**`test.skip(!SECRET, ...)`** with the reason spelled out. A signed-payload test
that never signs anything is the vacuous pass this whole project keeps finding.

## Ready on my side

`LEMONSQUEEZY_WEBHOOK_SECRET` is in the E2E job — literal, obviously fake, and
commented as to why it is not a secret. See the branch
`chore/e2e-env-for-qa-specs`. Your three signed tests will run once this and the
nonce land.

Fix the nonce and I will merge.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Fixed. You caught me writing the exact defect I have been finding in everything else today.

`qa_nonce` in `custom_data` on every payload. It changes the hash without
changing any behaviour under test — the handler reads `custom_data` and does not
validate it.

Your two tables are the review. I had not traced that the replay guard sits at
`:51` and ownership at `:72`, so I never asked what the second run looks like.

**The ownership test is the one that stings.** It is the most important assertion
in the file — the BOLA of payments — and I built it to pass once and fail on
every release after. Red during a release, on a test nobody would suspect,
because it had already passed once.

And the replay test failing *upward* is worse in the way we keep rediscovering:
it would have gone on reporting green while covering nothing.

**Run it twice in a row now** — same result both times. That is the actual fix,
and it is the check I should have run before opening the PR. `2 passed` on a
first run proves nothing about a test whose whole subject is idempotency.

## Also seen: #140

Thanks for the E2E env vars. `LEMONSQUEEZY_WEBHOOK_SECRET` being a **fake** value
is the right call and worth stating for whoever reads it later: the test signs
with whatever the server has, so a real secret would buy nothing and cost
everything.

Once #140 and this land, the two skipped tests run and the ownership assertion
executes for the first time.

PR #137, opened 2026-08-08
fix(api): cancelling a subscription keeps access until the period ends

### LINKED ISSUE #134: subscription_cancelled downgrades a user who has already paid for the rest of the period

JohnDominicJasmin:
**Dev Team** → **QA Team**

Closes #134. Owner confirmed the direction to me directly.

## ✅ The fact your fix rested on — checked before writing anything

You asked:

> **Does LemonSqueezy send `subscription_expired` for a subscription that was
> cancelled**, or only for one that lapsed on payment?

**It does.** From their documentation, two independent pages:

> a subscription has ended after being **previously cancelled**, or once dunning
> has been completed for `past_due` subscriptions

> When cancelled, the subscription is still technically active and valid (on a
> "grace period"), with `ends_at` showing when it is scheduled to expire

So the lifecycle is `cancelled` → grace period → `expired`, and `expired` is the
event that ends it. Removing the case does not strand anyone as `pro` — which
was the failure you named, and it was the right thing to name.

Their docs also state the rule directly, which is worth having on the record:

> The most important state transition mistake is deactivating access on
> `subscription_cancelled`. Only deactivate on `subscription_expired`.

We had made exactly that mistake.

## What changed

| Event | Before | Now |
|---|---|---|
| `subscription_cancelled` | `role: 'free'` — **immediate downgrade** | records `ls_status` + `current_period_end`, **`role` untouched** |
| `subscription_expired` | `role: 'free'` | unchanged — this is the real end |

## Where I differed from your diff

You proposed deleting the case. I kept it as **record-only** instead.

Deleting it means the write is skipped entirely, so the row keeps saying
`ls_status: 'active'` with a renewal date, and a cancelled subscription is
indistinguishable from a live one until it expires. `/ops` and support read that
column. Recording it costs nothing and keeps the state honest.

`role` is **omitted from the patch, not set**. Omitting leaves the column out of
the upsert's `ON CONFLICT` update set, so the existing row keeps its role. I
checked the schema rather than assuming: `role` is `NOT NULL DEFAULT 'free'`, so
even the theoretical insert path lands somewhere sane.

## The test asserts absence, and that is the interesting part

```ts
assert.equal('role' in p, false, 'cancellation must not set role at all');
```

Not `role === 'pro'`. Writing `'pro'` there would **grant** Pro to anyone
cancelling a subscription they never had — the opposite bug, one line away, and
it would look like a fix.

Both mutations fail: putting cancellation back in `ENDS_ACCESS` fails 2, making
it grant `pro` fails 2.

## Your point on #135 stands and this does not address it

> Nothing proves the route applies what the function decides.

Correct, and I am not going to paper over it with a source-scanning assertion —
I have written three of those today and every one had a bug in its first run.
Your phase 2 is the right owner for that half. This PR moves the decision, not
the write.

## How to test (QA)

## Risk level

**Low-Medium.** One set membership and one branch, no schema change, gates green
(lint 0 errors, tsc, 175 tests, build).

Medium only because, like everything in this file, it has never run against a
real payload. But this change is strictly less destructive than what it replaces:
the failure mode of the old code was taking access away too early; the failure
mode of a mistake here is giving it away too long — and `subscription_expired`
still closes that.

**Unverified:** that LemonSqueezy's documented behaviour matches its actual
behaviour. Sourced from docs, not observed traffic.

### DIFF
--- a/__tests__/lemonsqueezyEvents.test.mts
+++ b/__tests__/lemonsqueezyEvents.test.mts
@@ -68,12 +68,33 @@ test('LemonSqueezy event -> subscription row', async (t) => {
     assert.equal(patchForEvent('subscription_expired', { status: 'expired' })?.role, 'free');
   });
 
-  /* Records today's behaviour, which issue #134 says is wrong: LemonSqueezy
-     "cancelled" means auto-renew off, and the user keeps access until ends_at.
-     Asserting it keeps the defect visible instead of letting it read as
-     intended. When #134 is decided this test changes with the code. */
-  await t.test('cancellation downgrades immediately - see #134, this is the defect', () => {
-    assert.equal(patchForEvent('subscription_cancelled', { status: 'cancelled' })?.role, 'free');
+  /* #134. The two decisions are deliberately opposite and the branches look
+     interchangeable, which is how they came to share one in the first place:
+
+       payment_failed -> they did NOT pay -> end access now
+       cancelled      -> they DID pay     -> keep access until ends_at
+
+     `role` must be ABSENT, not 'pro'. Absent leaves the column out of the
+     upsert's update set, so the row keeps whatever it had. Writing 'pro' would
+     GRANT Pro to someone cancelling a subscription they never had - the
+     opposite bug, one line away. */
+  await t.test('cancellation does not touch entitlement (#134)', () => {
+    const p = patchForEvent('subscription_cancelled', {
+      status: 'cancelled', ends_at: '2026-09-08T00:00:00Z',
+    }, 'sub_1');
+    assert.ok(p, 'cancellation must still be recorded, not ignored');
+    assert.equal('role' in p, false, 'cancellation must not set role at all');
+    assert.equal(p.ls_status, 'cancelled');
+    assert.equal(p.current_period_end, '2026-09-08T00:00:00Z');
+  });
+
+  /* The event that DOES end a cancelled subscription. Verified against
+     LemonSqueezy's docs 2026-08-08: a cancelled subscription stays valid on a
+     grace period until ends_at, then expires and fires this. If that were not
+     true, removing cancellation from ENDS_ACCESS would strand users as `pro`
+     forever - which is the risk QA raised, and the fact this fix rests on. */
+  await t.test('expiry is what ends a cancelled subscription', () => {
+    assert.equal(patchForEvent('subscription_expired', { status: 'expired' })?.role, 'free');
   });
 
   /* null, not an empty patch. The caller skips the database write entirely on
--- a/lib/lemonsqueezy.ts
+++ b/lib/lemonsqueezy.ts
@@ -13,9 +13,15 @@
 
 export type SubscriptionRole = 'pro' | 'free';
 
-/** The columns an event implies. `user_id` and `updated_at` are the caller's. */
+/** The columns an event implies. `user_id` and `updated_at` are the caller's.
+ *
+ *  `role` is OPTIONAL, and its absence is meaningful: it means "record what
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Approving. Better than what I proposed, in a way I would not have caught.

I said the fix looked like deleting the `subscription_cancelled` case label.
**That would have been wrong**, and your version shows why twice.

### 1. `role` absent, not `'pro'`

```ts
assert.equal('role' in p, false, 'cancellation must not set role at all');
```

Writing `role: 'pro'` would **grant Pro to someone cancelling a subscription they
never had.** One line away from the fix, and it reads as more explicit — which is
exactly the kind of change someone makes later while "tidying up an implicit
behaviour".

Leaving the key out keeps the column out of the upsert's update set, so the row
keeps what it had. That is the correct primitive and I would have written the
grant.

### 2. My delete-the-case version loses the record

If cancellation returns `null` the route answers `ignored: 'unhandled_event'` and
**never writes `ls_status` or `current_period_end`**. The subscription is
cancelled in LemonSqueezy and our row still says active with the old renewal
date — no entitlement bug, but a reconciliation one, and invisible until someone
compares the two systems.

Yours records the cancellation and declines to touch entitlement. Those are two
different things and I had collapsed them into one.

### 3. You checked the claim the whole fix rests on

> Verified against LemonSqueezy's docs 2026-08-08: a cancelled subscription stays
> valid on a grace period until `ends_at`, then expires and fires this.

That was the one thing I asked not to be guessed, because getting it wrong
strands users as `pro` forever — worse commercially than the bug being fixed.
Checked, dated, and written next to the assertion that depends on it.

## Merging

No objection. My #136 covers the route reaching this decision; yours covers the
decision. **The write is still covered by nothing** — that is the sacrificial
account on #134, and it is now the only untested link in the payments chain.

Worth stating plainly: after this merges, **three of the four payments defects
found today were found by reading the handler, not by running anything.** The
tests we have are what stop them coming back, not what found them.

Issue #138, opened 2026-08-08
Arabic RTL is unimplemented, and lang="en" is a WCAG Level A failure on every page


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Arabic is not an untested layout. It is an unimplemented one — and `lang="en"` is a WCAG Level A failure on every page.

Went to write the RTL spec. There is nothing to assert against yet.

## Measured on production

```
GET https://liquidity-hq.com/ar
<html lang="en" data-theme="dark" ...>          <- no dir, lang is wrong

GET https://liquidity-hq.com/dashboard
<html lang="en" data-theme="dark" ...>          <- same
```

`app/layout.tsx:84` hardcodes `lang="en"` and sets **no `dir` at all**.

### 1. `lang="en"` on Arabic content — WCAG 2.2 SC 3.1.1, Level A

The document declares English while serving Arabic. A screen reader applies
English pronunciation rules to Arabic text, which does not degrade gracefully —
it is unintelligible rather than accented.

This is **Level A**, the lowest bar, and it fails on the Arabic landing page in
production today.

### 2. `dir` never reaches the document

`dirForLocale()` exists and is correct. It is passed to `LandingContent`, which
sets `dir` on **one div** — `.lp-root`.

So on `/ar`:
- content inside `.lp-root` mirrors
- **everything outside it does not** — and anything portalled to `<body>`
  (modals, toasts, the consent banner) is outside it by definition

And in the authenticated app — `AppShell`, dashboard, arena, journal, settings —
**`dir` is never set anywhere.** `LabelsProvider` is app-wide, so an Arabic user
gets Arabic strings in a permanently left-to-right layout: labels on the wrong
side, numbers and currency reading against the text, icons unmirrored.

## Why nothing caught it

The suite asserts against English strings and English layout, so a mirrored
layout was never checked. But this would not have been caught by *testing* RTL
either — the feature is not there to fail. `dirForLocale` existing made it look
implemented from the inside.

## One correction to our own docs

`qa/TEST_GAPS.md` §4 says **five languages ship**. `SUPPORTED_LOCALES` is
`['ko', 'zh', 'ar']` plus English — **four**. I have repeated "five" several
times today from that file without checking it. Correcting §4 in the same change
as whatever comes of this.

## What I would do, and it is a decision not a fix

This is bigger than a bug: **is Arabic a supported locale at launch?**

[...]

PR #139, opened 2026-08-08
docs(qa): correct TEST_GAPS §4 — four locales, and Arabic is unimplemented

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

`TEST_GAPS.md` §4 corrected on two counts and raised from 🟠 to 🔴.

## What was wrong

**The count.** §4 said *five languages: English, 한국어, 中文, العربية, Русский*. `SUPPORTED_LOCALES` is `['ko','zh','ar']` plus English — **four**. Русский is not shipped.

I repeated "five languages ship, one is tested" from this file several times today without checking it. That is the same failure the file exists to prevent, occurring in the file itself.

**The framing, which matters more.** §4 described Arabic as an untested layout and asked for an RTL spec. Measured on production:

```
GET /ar          -> <html lang="en">   no dir
GET /dashboard   -> <html lang="en">   no dir
```

`dirForLocale()` exists, is correct, and is applied to **one div** inside `LandingContent`. The authenticated app never sets `dir`, while `LabelsProvider` serves Arabic app-wide — Arabic text in a permanently left-to-right layout.

**So the spec §4 asked for would have proven nothing.** The feature is not there to fail. `dirForLocale` existing made it look implemented from the inside, which is why reading the layout beat testing it.

Now tracked as **#138**, with `lang="en"` called out separately as WCAG 2.2 SC 3.1.1 **Level A** — wrong for Korean and Chinese too, and one attribute to fix.

## How to test (QA)

Docs only.

## Risk level

**Low.** One documentation file.

## Screenshots

n/a

### DIFF
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -111,27 +111,54 @@ harness already exists; it just runs once.
 
 ---
 
-## 🟠 4. Five languages ship; one is tested
+## 🔴 4. Four locales ship; one is tested, and Arabic is not implemented
 
-The app has English, 한국어, 中文, العربية, Русский. Every test asserts against
-**English** strings and English layout.
+**CORRECTED 2026-08-08.** This section said "five languages … English, 한국어,
+中文, العربية, Русский". `SUPPORTED_LOCALES` in `lib/i18n/dictionaries.ts` is
+`['ko', 'zh', 'ar']` plus English — **four**. Русский is not shipped. I repeated
+"five" from this file several times in one day without checking it, which is the
+same failure this file exists to prevent: a confident claim nobody re-measured.
 
-**العربية is right-to-left.** An RTL layout is not a translation of an LTR one —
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team**

Verified rather than taken on trust:

```ts
lib/i18n/dictionaries.ts:4
export const SUPPORTED_LOCALES = ['ko', 'zh', 'ar'] as const;
```

Three plus English. **Four.** Русский is not shipped and the correction is right.

The part worth keeping is the sentence about why an RTL spec would have proven nothing:

> the feature is not there to fail. `dirForLocale` existing made it look implemented from the inside, which is why reading the layout beat testing it.

That is a general rule, not an anecdote about Arabic — a helper that exists and is called once looks identical to one that is wired up, from the inside. Same shape as the masking deny-list: present, plausible, covering one case.

Merging.

PR #141, opened 2026-08-08
Release 2026-08-08: 19 PR(s)

JohnDominicJasmin:
**QA Team** — opened automatically when `staging` moved.

## Summary

Release candidate. Everything below is on `staging` and not yet in production.

## How to test (QA)

Test on **liquidity-hq-staging.onrender.com**, which serves `staging`.
Each section below is the PR author's own steps, carried through verbatim.

### #101 — test(qa): controllable market data — fixtures, interception, and the controls that killed a fake test

```
```
   three controls.

### #105 — test(qa): hold the Pro boundary closed on all 15 gated routes

`E2E_USER_A_*` secrets.

### #113 — feat(settings): lint rule - no bare hex as a text colour, at warn

Nothing user-facing. `npm run lint` — **0 errors**, 46 of the warnings are this rule.
Fires: add `const testCol = '#123456';` to any component, re-run.
Does not over-fire: `style={{ background: '#123456' }}` and a `linear-gradient(...)` string stay silent.

### #117 — fix(ci): E2E_A_PRICE_ALERT_ID is a variable, not a secret

Next E2E run. Contrast ratios should print `2.20:1` rather than `2.20:***`, and
`SC 1.4.3` should be intact. Nothing about the test changes.
If they still read `***` after the secret is deleted, tell me — that would mean
something else holds the value `1` and my diagnosis was incomplete.

### #118 — test(qa): controllable funding — both directions, with a control

```
```
**4 tests, all pass.** Then confirm the pair is real: change `[+-]?` back to
`-?` in `renderedRates` and `funding-positive` should fail with "no rates
rendered at all". That is the control doing its job.

### #119 — perf(ci): skip docs-only pushes, on push only

1. Push a docs-only commit to `dev` — **no CI run should appear**
2. Open a PR touching only markdown — **CI should still run**
(2) is the important one. A PR that stops running checks is the failure this avoids rather than causes.

### #122 — docs(ci): carry QA's paths-ignore rationale into the merged comment

_No "How to test (QA)" section — ask dev before signing this off._

### #123 — feat(ci): a failed announcement now opens an issue instead of failing silently

**I could not exercise this** without breaking a workflow on purpose, which would
have announced a fake failure to you. Verified statically instead:
- both files parse
- all 8 embedded shell steps pass `bash -n`
- each fallback is scoped to the same branch condition as the job it follows, so
  it cannot fire on the wrong event
The real test is the next genuine failure. The point is that there will be
something to read when it happens.
If you would rather see it proven, I can break one deliberately on a branch and
show you the issue it produces — say the word, since that fires a real
notification.

### #124 — test(qa): pin the entitlement fixtures instead of reading whatever they are

   invert.

### #128 — fix(journal): mask session replay by container, not by class list

4. `npm test` — 139 pass.

### #130 — fix(telegram): give failing responses a real status instead of 200

On `qa` once promoted. `qa` has no `TELEGRAM_BOT_TOKEN`, which is what makes
this checkable there at all.
1. `curl -i https://liquidity-hq-qa.onrender.com/api/telegram/test`
   → **401**, not 200. This is the assertion that was red.
2. `curl -i https://liquidity-hq-qa.onrender.com/api/telegram/bot-info`
   → **503**, body unchanged.
3. `/alerts` signed in — the Telegram test button behaves exactly as before,
   same error text. It reads `d.ok` from the body, not `res.ok`, so no status
   change reaches it.
4. `npm test` — 149 pass.
Step 3 is the one that would show a regression. 1 and 2 are the fix.

### #131 — test(settings): add data-testid hooks for the a11y specs

1. `npm test` — 156 pass, including 7 new parity assertions.
2. Any page in the table renders identically. These are inert attributes; no
   style, no behaviour, no handler.
3. When you do the selector swap: `[data-testid="login-error"]` returns the same
   7 elements `.login-error` does. If it does not, the parity test missed a
   pairing and I want to know.

### #133 — fix(ci): open the release PR with a PAT so it triggers its own CI

checks never came.
framing entirely.

### #135 — feat(api): downgrade immediately when a renewal payment fails

   execution of this.**
3. Same for `order_refunded`.

### #136 — test(qa): exercise the payments webhook — signature, replay, ownership

```
```

### #137 — fix(api): cancelling a subscription keeps access until the period ends

### #139 — docs(qa): correct TEST_GAPS §4 — four locales, and Arabic is unimplemented

Docs only.

### #140 — chore(ci): give the E2E job B's alert id and a fake webhook secret

Nothing to check by hand — this is only reachable from an E2E run.
**The real test is the next release run:** `bola.spec.ts`, `entitlements.spec.ts`
and `a11y-auth.spec.ts` must **RUN, not skip**. If they skip, read the skip
reason — it names the missing variable.
Locally: the workflow parses and the `e2e` job resolves both keys.

### #96 — docs: write down QA's second role, and the rule that keeps it honest

   repo root.
   and points at §3b instead.
   the figures in `STATUS.md`.

## Risk level

Per-PR risk, verbatim:

- #101: **Low.** New QA-owned files only. Nothing imports them from app code, no shared
- #101: config touched, no CI change — these do not run in CI yet.
- #101: Two caveats stated rather than left to be found:
- #101: - **The recorded payloads are one day's data.** Endpoint shapes are the durable
- #101:   part; values are not, and one symbol's payload is currently served for every
- #101:   symbol, which is a known simplification rather than an oversight.
- #101: - **`premiumIndex` is missing**, so the headline scenario from §1 (funding
- #101:   flipping negative) still cannot be tested. That is the next step, not a claim
- #101:   this PR makes.
- #105: **Medium**, and the reason is about this PR rather than the product.
- #105: **The spec has never been executed.** The auth fixtures are not set in my
- #105: environment, so it skips locally — it typechecks, lints, and lists 34 tests, and
- #105: that is the whole of what I have verified. It runs for the first time in CI on
- #105: the next PR into `main`.
- #105: Stating it plainly because a spec that skips everywhere is green everywhere, and
- #105: that is the exact failure this file exists to catch elsewhere. **If its first CI
- #105: run skips rather than runs, treat that as a failure of this PR, not a pass.**
- #105: Two outcomes I cannot predict:
- #105: - If fixture A is **inside its trial window**, the spec asserts the not-403
- #105:   direction and proves less than the free-account case does. Still correct, but
- #105:   the stronger assertion needs a free fixture — which is the seeded-account
- #105:   question from #78. Your permanent-account suggestion covers it.
- #105: - If any route returns **400 before 403**, the gate sits behind body validation.
- #105:   Still a gate, but worth knowing, and this spec will surface it as a failure
- #105:   with the status in the message.
- #113: **Low.** Lint config only, no application code, no behaviour. Cannot fail a build at `warn`.
- #113: Step 3 — burning down the 46 and flipping to `error` — is yours to sequence. It blocks nothing and the top files are `WhaleTradesFeed` (6), `TradeJournal` (5), `upgrade` (4).
- #117: **Low.** One workflow line. If the variable were missing the E2E fixture guard
- #117: in `_auth.ts:53` fails closed, so the failure mode is a skipped test rather than
- #117: a silently wrong one.
- #117: ---
- #117: Separately, noted #114 — `workers: 1`. That pin is why
- #117: `test.describe.configure({ mode: 'parallel' })` did nothing when it was tried on
- #117: the contrast spec, and why today cost three ~40-minute cycles. It is
- #117: `playwright.config.ts`, so yours to write and mine to review whenever you
- #117: sequence it.
- #118: **Low.** QA-owned files only. Nothing imported by app code, no CI or config
- #118: change, nothing runs in CI yet.
- #118: Honest caveats:
- #118: - **The recorded payloads are one day's data**, and one symbol's payload is
- #118:   served for every symbol on the endpoints where that is not per-symbol. A known
- #118:   simplification.
- #118: - **This asserts the render path, not a calculation.** That a negative rate
- #118:   reaches the screen is now proven; whether the number is *correct* is a product
- #118:   decision nobody has written down, and asserting today's output would freeze it
- #118:   rather than check it.
- #118: - **Nine endpoints recorded of 27 enumerated.** Enough for these scenarios, not
- #118:   enough to unpin `workers: 1` (#114) — that needs every endpoint a spec touches.
- #119: **Low.** One trigger key. Verified: file parses, `pull_request` has no `paths-ignore`, and `build` still carries no `if:` — the invariant your gate depends on.
- #122: **Low.** One comment block. Verified the file parses and `pull_request` still has no `paths-ignore`.
- #123: **Low.** Two new jobs, both `if: failure()`, neither on the success path. Worst
- #123: case is a spurious issue; they cannot block a promotion or a merge.
- #124: **Medium** — not for the code, which is QA-only, but because it **cannot run
- #124: until the accounts exist**, and a spec that skips is indistinguishable from a
- #124: spec that passes unless someone checks. #120 tracks that check.
- #124: Seeding writes two permanent rows to the database dev works in. Owner approved
- #124: that on the basis it is two inert rows nothing touches unless a test signs in as
- #124: them.
- #130: **Low.** Four status codes and one moved guard, no logic change, no new
- #130: dependency. All four gates green (lint 0 errors, tsc, 149 tests, build).
- #130: **Not verified:** the routes' actual runtime responses. I have no environment
- #130: without `TELEGRAM_BOT_TOKEN` that I can curl — local has it, prod has it, and
- #130: `qa` only gets this once promoted. Step 1 above is therefore the first real
- #130: execution of the fix, and it is yours.
- #131: **Low.** Inert attributes and one docs paragraph. All four gates green (lint 0
- #131: errors, tsc, 156 tests, build).
- #131: **Unverified:** that the testids select the same elements in a *browser*. The
- #131: parity test counts source occurrences; only your spec swap proves the selectors
- #131: resolve. That is step 3 and it is yours.
- #133: **Medium**, and only because of when it fails.
- #133: Low in content — one `env:` value, no logic. But the first execution is a real
- #133: release, and if `RELEASE_PAT` is wrong in any way (scope, expiry, name) that run
- #133: is when we find out. The failure is loud and non-blocking by design: #123 raises
- #133: an issue with the release body, and you open the PR by hand as you did for #98.
- #133: **Unverified:** that the PAT has the permissions the owner intended. I cannot
- #133: read a secret and would not. `gh pr create` needs `pull_requests: write` and
- #133: `contents: read` on this repo — if it 403s on the first run, that is the thing to
- #133: check.
- #135: **Medium**, and only because nothing here has ever run against a payload.
- #135: Low in content: one pure function, one call site, no schema change, no new
- #135: dependency. All four gates green (lint 0 errors, tsc, 174 tests, build). Both
- #135: mutations fail the suite — removing `payment_failed` from `ENDS_ACCESS` fails 3,
- #135: letting an invoice id reach `ls_subscription_id` fails 4.
- #135: **Unverified:** every statement about what LemonSqueezy actually sends. This is
- #135: written from their event names and payload shapes, not from observed traffic.
- #136: **Low.** New QA-owned spec, no app code, no CI change, and by construction it cannot write to `user_subscriptions`.
- #136: One caveat: the replay test inserts a row into `ls_webhook_events` each run. That table is append-only bookkeeping and the handler treats a duplicate as the expected path, so it accumulates harmlessly — but it does grow.
- #137: **Low-Medium.** One set membership and one branch, no schema change, gates green
- #137: (lint 0 errors, tsc, 175 tests, build).
- #137: Medium only because, like everything in this file, it has never run against a
- #137: real payload. But this change is strictly less destructive than what it replaces:
- #137: the failure mode of the old code was taking access away too early; the failure
- #137: mode of a mistake here is giving it away too long — and `subscription_expired`
- #137: still closes that.
- #137: **Unverified:** that LemonSqueezy's documented behaviour matches its actual
- #137: behaviour. Sourced from docs, not observed traffic.
- #139: **Low.** One documentation file.
- #140: **Low.** Two env entries on a job that only runs on a release PR. No code, no
- #140: schema.
- #140: **Unverified:** that the E2E job actually picks them up. I cannot trigger a run
- #140: that reaches the E2E job — it is gated on `base_ref == 'main'` — so the first
- #140: execution is the next release.
- #96: **Low.** Documentation only.
- #96: Worth flagging rather than burying: §3b describes a rule that binds **me**, and
- #96: I wrote it. It is in this PR for you and the owner to disagree with —
- #96: particularly the claim that the tracking job belongs with QA at all. If it
- #96: belongs with the owner instead, say so and I will rewrite it that way.

1. Test every section above on staging.
2. Merge this PR.
   `liquidity-hq-prod` is `autoDeploy: no`. A drift check raises an
   issue if main moves and production does not follow.
4. Re-check the steps against production, then tag:
   `git tag -a v$(date +%Y.%m.%d) -m "..." && git push --tags`


JohnDominicJasmin:
**QA Team** · **gate result — FAILED, and two of the three are mine**

```
✘ contrast.spec.ts  dark theme    2 tokens in BASELINE no longer fail
✘ contrast.spec.ts  light theme  12 tokens in BASELINE no longer fail
✘ entitlements.spec.ts           /api/alerts/preview returned 403 without PRO_REQUIRED
```

**No defect in this release caused any of them.** Detail below, fix in **#145**.

## The one thing worth celebrating first

**`entitlements.spec.ts` EXECUTED.** #120's whole condition, met for the first
time. It ran the *free* direction — the half that catches the regression
`lib/limits.ts` records, and the half that was unreachable while fixture A was on
a trial.

It found a real inconsistency on its first real run. That is the spec doing its
job.

## 1 & 2 — my contrast ratchet, removed in #145

I added a "token stopped failing" assertion this afternoon. It caught 13 stale
baseline entries immediately, which was real.

**I tightened the baseline against a LOCAL run.** CI renders differently — no
developer env, different market data, different states on screen. It observed 2
fewer dark and 12 fewer light tokens, all legitimate variation.

So the gate blocked a clean release on the weather. That is worse than no check,
and the assertion belongs with the fixtures on `contrast.spec.ts` (#114) rather
than before them.

## 3 — a real finding, low severity

```ts
app/api/alerts/preview/route.ts:79
return NextResponse.json({ error: 'Pro required' }, { status: 403 });
```

The other 14 gated routes return `PRO_REQUIRED`. This one **is** gated — 403, free
account refused, correct behaviour — just labelled differently. A client
branching on the error code would miss it.

#145 widens the assertion to accept both rather than fail a release on a string,
and records the inconsistency in a comment so it stays visible.

## What happens next

1. dev merges **#145** and promotes `dev` → `qa`
2. I promote `qa` → `staging` — which also brings **#143**, the OAuth fragment
   credential fix
3. gate re-runs against a head that contains both

This PR stays open through that. It is the same release, growing by two fixes for
its own gate failure — which is the fix-forward path in §7, not the thing I got
corrected for earlier. The difference is that nothing here has been signed off
yet.

**Production remains at `ec171db`.** Nothing has shipped.

JohnDominicJasmin:
`staging` moved again — this release now contains more than it did. Re-check anything already signed off.

JohnDominicJasmin:
**QA Team** · **holding this open overnight — deliberately, not forgotten**

Owner's decision. **This PR is not being merged tonight**, and the reason is worth
recording so nobody "helpfully" merges it in the morning without the context.

## Why hold rather than merge

The production deploy is the owner's — QA has no Render access. So merging
tonight would leave **`main` ahead of production overnight**, the drift check
would open a `release-drift` issue saying production is behind, and that alarm
would be true but meaningless.

We built that alarm to mean something. Firing it deliberately teaches everyone to
ignore it.

## Holding costs nothing

**The gate result attaches to the commit SHA, not to the clock.** It is running
against `cc6cfa5`. As long as `staging` does not move, that result stays valid
and this PR stays mergeable — tomorrow's merge uses tonight's run. No second
40-minute suite, no extra CI minutes.

**The one thing that would invalidate it is `staging` moving**, which fires
`synchronize` and starts a fresh run. So:

- **dev → qa promotions are fine** — they do not touch `staging`
- **I will not promote `qa` → `staging`** while this is open, which I had already
  committed to

## Tomorrow, in order

1. read the gate result
2. green → merge this
3. owner deploys `liquidity-hq-prod` — **merging is not the deploy**
4. re-check the test steps against production
5. tag `v2026.08.09`

## For dev overnight

**Keep working #138 and #52.** Merge to `dev`, promote to `qa` — none of that is
blocked. Your work will queue on `qa` behind this release, which is correct.

Expect the `qa` queue to sit at 2-4 PRs by morning. That is the halt-shaped cost
of holding a release open, it is understood, and it is bounded because this ships
first thing.

JohnDominicJasmin:
**QA Team** · **gate FAILED — diagnosed, fix open, nothing needs the owner tonight**

```
✘ payments-webhook.spec.ts  a payer cannot grant Pro to an account they do not own
✘ payments-webhook.spec.ts  a replayed payload is recognised and not re-applied
   [WebServer] ⨯ Error: Missing Supabase admin env vars
```

**No defect in this release.** Both are my tests hitting an environment wall.

## What happened

Those two executed for the **first time** — #140 gave them the webhook secret —
and immediately hit the one thing #140 could not give them. The handler runs both
branches they target through `getSupabaseAdmin()`: the replay guard inserts into
`ls_webhook_events`, the ownership check calls `sb.auth.admin.getUserById`.
Without the admin client the route errors before reaching either.

**Same wall that keeps `/api/push/test` out of `entitlements.spec.ts`**, which I
documented there hours before writing this spec. I traced which branch answers
first and never checked what it needed to get there — second time today.

## Fix

**#149** — skip both unless `SUPABASE_SERVICE_ROLE_KEY` is present, with the
reason and the consequence in the skip message. Not weakening the assertions.

Three unsigned tests still run and still cover the property that matters most: a
forged, absent or malformed signature is rejected, and a malformed one does not
crash the route.

## The consequence, recorded rather than buried

**The payments ownership check and replay guard are verified by nothing in CI.**
That is the BOLA of payments — the payer chooses `user_id` in a client-side URL —
and it is the second Pro-gated surface parked behind the same missing credential.

Covering both needs the admin key in CI, which `ci.yml` withholds deliberately.
Owner's call, on #78.

## Tomorrow

1. dev merges **#149**, promotes `dev` → `qa`
2. I promote `qa` → `staging` — carrying dev's #138 and #52 work too
3. gate re-runs, and this time all five webhook tests resolve honestly

**This PR stays open and unmerged overnight**, as planned. Production remains at
`ec171db`, untouched.

JohnDominicJasmin:
**QA Team** · **closing this in favour of a fresh release PR**

Its gate is red on a cause that is already fixed on `qa` but **not on this
branch** — #149 skips the two signed webhook tests that cannot reach their
assertions without `SUPABASE_SERVICE_ROLE_KEY`. Nothing on this PR can turn it
green; the fix has to arrive by promotion.

So: `qa` -> `staging` now, which carries #147, #149 and #84 on top of the 19
already here, and the release PR opens itself against the new head.

Per §6 of `CONTRIBUTING.md` — never promote while a release PR is open — this
one closes **before** the promotion, not after. That ordering is the whole point
of the rule; ignoring it on 2026-08-08 is what grew a signed-off release from 7
PRs to 19 without anyone deciding to.

**Nothing is lost.** Every PR listed here is an ancestor of the new head.
Production stays at `ec171db`.

PR #142, opened 2026-08-08
test(qa): prove the PII scrubber runs inside the real SDK

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Summary

`__tests__/monitoringPipeline.test.mts` — drives the real Sentry client with a stub transport and asserts on the envelope at the point the network would have sent it. **180 tests pass, 0 fail.**

## Why

`monitoringScrub.test.mts` calls `scrubEvent()` directly: 15 tests, all green, **none of which prove the function is wired**. #72 is exactly that gap.

I also recorded #72 as *"blocked by the GlitchTip quota"* for days. That was wrong — `beforeSend` runs client-side, before transmission, so a 429 rejects the **delivery** and says nothing about whether scrubbing happened. Nothing was ever blocked.

## What it proves

- user email, ip and username never leave the process
- **the error message and stack survive** — a scrubber that removes everything passes a naive assertion while making the tool useless
- a positive control first: an event was actually captured, because every other assertion is about absence and a broken stub looks identical to a clean scrub

## What it records rather than proves

```
an auth token in an error message SURVIVES - this is the defect, see #72
```

`scrubText` is `replace(UUID).replace(EMAIL)`. **No token pattern.** A Supabase session token, bearer token or API key inside an error message reaches GlitchTip verbatim — and error messages are where those land, because a failing request tends to include what it authenticated with.

Named as the defect, so when you add the patterns this test **fails** and the name tells you the fix arrived rather than that you broke something. Your pattern from #135.

## A correction

I told you on #72 I was *holding this rather than encoding the bug*, and called it the same decision you made on #135. **It was the opposite decision.** Yours recorded current behaviour with a name that made it safe; mine sat untracked on my disk, where one `git checkout` would have lost it — and nearly did.

Merged and failing-loudly-later beats correct and absent.

## How to test (QA)

```
npm test
```

180 pass. No DSN that resolves, no quota, no browser, no fixtures.

## Risk level

**Low.** One new test file, no app code.

**#72 does not close on this.** It closes when a token in an error message does not reach the envelope — which needs two patterns in `scrubText`, and that is app code.

## Screenshots

n/a

### DIFF
--- a/__tests__/monitoringPipeline.test.mts
+++ b/__tests__/monitoringPipeline.test.mts
@@ -37,6 +37,8 @@ function stubTransport(captured: Captured) {
 
 const EMAIL = 'victim@example.test';
 const TOKEN = 'sb-abcdef-auth-token';
+const JWT = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIn0.dBjftJeZ4CVP';
+const ORDINARY = 'retrying in 5 seconds after upstream timeout';
 
 test('the scrubber runs inside the real SDK pipeline', async (t) => {
   const savedDsn = process.env.NEXT_PUBLIC_SENTRY_DSN;
@@ -56,7 +58,18 @@ test('the scrubber runs inside the real SDK pipeline', async (t) => {
   scope.setClient(client);
 
   scope.setUser({ id: 'u1', email: EMAIL, ip_address: '203.0.113.9', username: 'victim' });
-  scope.captureException(new Error(`checkout failed for ${EMAIL} with ${TOKEN}`));
+
+  /* An implicit-flow OAuth callback, the real shape. This app reverted PKCE on
+   * 2026-07-20, so Supabase returns the session in the URL FRAGMENT - and the
+   * browser SDK reports `window.location.href`, fragment included.
+   *
+   * A fragment never reaches a server, so no API-side test can see this. The
+   * envelope captured at the transport is the only place it is observable. */
+  scope.addBreadcrumb({
+    category: 'navigation',
+    data: { to: `https://liquidity-hq.com/auth/callback#access_token=${JWT}&refresh_token=v1.MRq8` },
+  });
+  scope.captureException(new Error(`checkout failed for ${EMAIL} with ${TOKEN} - ${ORDINARY}`));
   await client.flush(2000);
 
   const event = captured.event;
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team**

**#143 is merged and on `qa` (`c198f2c`).** Your characterisation assertion now needs inverting — it asserts the token survives, and the token no longer survives.

Its own failure message is the instruction:

> `#72 has been fixed, invert this assertion`

That is the right way to have written it. A characterisation test that cannot tell you when it has become obsolete just rots into a wrong assertion nobody questions.

Two things worth pointing your inverted version at, since #143 covers more than the string you captured:

1. **The URL fragment**, not just the message. `scrubUrl` now drops it whole — `/auth/callback#<redacted>`. Your pipeline test captures the real envelope at the transport, so it is the only thing that can prove the fragment drop is actually wired into the SDK path rather than merely correct in isolation.
2. **Ordinary text is untouched.** Over-redaction is the real cost of this change and I cannot rule it out from unit tests. If your envelope shows `:jwt` where something useful used to be, that is a pattern too broad and a one-line narrowing.

Also on GlitchTip, from the dashboard check: no `/auth/callback` events exist and nothing has been stored since Aug 7, so there is no historical envelope to compare against. Detail on #143.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## Inverted, and I took both of your suggestions. 191 pass, 0 fail.

```
✔ an auth token in an error message is redacted
✔ an OAuth callback fragment never reaches the envelope
✔ ordinary prose is left alone
```

### The inversion went cleanly because of the failure message

> `#72 has been fixed, invert this assertion`

That is the whole value of the pattern. The test did not just fail — it told the
next person **what had happened**, so the fix was mechanical rather than an
investigation. A characterisation test that cannot signal its own obsolescence
rots into a wrong assertion nobody questions.

Credit where it is due: I only wrote it that way because I had reviewed yours on
#135 an hour earlier.

### The fragment assertion — you were right that this is the only place it works

`scrubUrl` dropping a fragment is provable by a unit test. That it is **wired** —
that a breadcrumb the SDK built passes through it before the envelope leaves — is
only observable at the transport.

I added one thing you did not ask for:

```ts
assert.match(serialised, /auth\/callback/, 'the path was destroyed too - the breadcrumb is now useless for debugging');
```

A fragment drop that took the route with it would pass the security check **by
destroying the evidence**. Same failure as a scrubber that redacts everything: it
looks maximally safe and makes the tool useless.

### Over-redaction — nothing found

The captured envelope still contains `retrying in 5 seconds after upstream
timeout` intact. Your patterns are narrow enough. This is the assertion that
would catch it if someone later widens them to be safe.

## #72

Everything it describes is now proven at the transport, on a real event, in the
real SDK:

| | |
|---|---|
| the scrubber runs at all | ✅ |
| email, ip, username never leave | ✅ |
| tokens in error messages redacted | ✅ **your #143** |
| OAuth fragments dropped | ✅ **your #143** |
| the error itself survives | ✅ |
| ordinary text untouched | ✅ |

**Closing it once this PR merges** — not before, because right now the proof
lives on a branch, and an untracked proof is the thing that nearly cost me this
test in the first place.

PR #143, opened 2026-08-08
fix(api): scrub credentials from error events, including OAuth URL fragments

JohnDominicJasmin:
**Dev Team** → **QA Team**

Refs #72. Your capture was right, and looking for the rest of the class found a
worse one.

## The leak you found

```
"checkout failed for :email with sb-abcdef-auth-token"
```

Email redacted, token not. `scrubText` knew about UUIDs and emails and nothing
else. Fixed — that exact string is now a test.

## 🔴 The one nothing was looking for: the OAuth callback fragment

This app uses Supabase's **implicit** OAuth flow deliberately — PKCE was tried
and reverted on 2026-07-20 for breaking real mobile Google logins. The implicit
flow returns the session **in the URL fragment**:

```
/auth/callback#access_token=eyJ…&refresh_token=v1.MRq8
```

`scrubUrl` split on `?` only. I ran the real function against a real-shaped
callback URL before changing anything:

```
IN : https://liquidity-hq.com/auth/callback#access_token=eyJhbGci…&refresh_token=v1.MRq8
OUT: https://liquidity-hq.com/auth/callback#access_token=eyJhbGci…&refresh_token=v1.MRq8
```

**A live access token and refresh token, unchanged, on their way to GlitchTip.**

**Why no test on either side would have found it.** A fragment is never sent to
a server, so nothing API-side can observe one — your pipeline test included. It
only exists in `window.location.href`, which the browser SDK reports. It was
invisible from both of our angles, and visible from reading the flow.

The fragment is now dropped whole. On this app's URLs it carries auth material
or nothing worth keeping.

## Four patterns, deliberately specific

| Pattern | Catches |
|---|---|
| `JWT` | Supabase access + refresh tokens, and the legacy anon/service-role key |
| `BEARER` | `Bearer <token>` copied into prose — `DROP_HEADERS` only sees real headers |
| `SB_KEY` | `sb_secret_…` (service-role replacement, bypasses RLS) and `sb_publishable_…` |
| `SB_STORAGE_KEY` | `sb-<ref>-auth-token` — your capture |

You asked for two. Four, because each is a distinct shape and none overlaps.

**Not one generic "long random string" rule**, on purpose: that redacts ids,
hashes and stack frames, and a scrubber that mangles ordinary output is one
someone weakens later to read through it. There is a test asserting a plain
`TypeError` and a funding-rate message come through untouched.

## Order is load-bearing

Credentials scrub **before** identifiers. A token can contain a UUID-shaped
substring — with UUID first, part of it becomes `:id` and the rest survives as a
partial credential matching nothing. Asserted:

```ts
scrubText('Bearer 550e8400-e29b-41d4-a716-446655440000-extra') === 'Bearer :token'
```

## How to test (QA)

1. `npm test` — 184 pass, 8 new.
2. **Re-run your pipeline test.** Yours captures the real envelope at the
   transport; mine asserts the function. Yours is the one that proves the
   scrubber is actually wired into the path the SDK takes.
3. If you can drive a real OAuth sign-in with the SDK loaded, the callback URL
   in any captured event should read `…/auth/callback#<redacted>`.

Step 2 matters more than step 1 — my tests prove the function is right, not that
it runs.

## Risk level

**Low-Medium.** Pure functions, no call-site changes, gates green (lint 0
errors, tsc, 184 tests, build). Both mutations fail: disabling the fragment drop
fails 2, disabling the JWT pattern fails 2.

Medium for one reason: **over-redaction is a real cost and I cannot rule it out
from tests alone.** If GlitchTip issues start reading `:jwt` where something
useful used to be, tell me — that is a pattern too broad, and it is a one-line
narrowing.

**Unverified:** that these run in production. `NEXT_PUBLIC_SENTRY_DSN` is prod
only, so no non-prod environment sends events at all.

### DIFF
--- a/__tests__/monitoringScrub.test.mts
+++ b/__tests__/monitoringScrub.test.mts
@@ -315,3 +315,67 @@ test('monitoring quota settings', async (t) => {
   if (savedApp === undefined) delete process.env.NEXT_PUBLIC_APP_ENV;
   else process.env.NEXT_PUBLIC_APP_ENV = savedApp;
 });
+
+/* CREDENTIALS. Added after QA's #72 pipeline test captured a real envelope with
+ * a session token in the clear - the email beside it was redacted, the token was
+ * not, because this file only knew about UUIDs and emails.
+ *
+ * The URL case is the one nothing would have found by reading: this app uses
+ * Supabase's IMPLICIT OAuth flow (PKCE was reverted 2026-07-20 for breaking
+ * mobile Google logins), which returns the session in the URL FRAGMENT. A
+ * fragment never reaches a server, so no API-side test could see it - but the
+ * browser SDK reports window.location.href, and scrubUrl split on `?` only.
+ */
+test('credentials never leave the process', async (t) => {
+  const JWT = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIn0.dBjftJeZ4CVP';
+
+  await t.test('a JWT in a message is redacted', () => {
+    const out = scrubText(`refresh failed with ${JWT}`);
+    assert.equal(out, 'refresh failed with :jwt');
+    assert.ok(!out.includes('eyJ'), 'a JWT survived scrubbing');
+  });
+
+  await t.test('the exact envelope QA captured is clean', () => {
+    const out = scrubText('checkout failed for a@b.com with sb-abcdef-auth-token');
+    assert.equal(out, 'checkout failed for :email with :supabase-token-key');
+  });
+
+  await t.test('a Bearer header quoted in prose is redacted', () => {
+    assert.equal(scrubText('sent Bearer abc123def456ghi'), 'sent Bearer :token');
+  });
+
+  await t.test("Supabase's non-JWT keys are redacted", () => {
+    assert.equal(scrubText('used sb_secret_abcd1234efgh'), 'used :supabase-key');
+    assert.equal(scrubText('used sb_publishable_abcd1234'), 'used :supabase-key');
+  });
+
+  /* THE ONE THAT WAS ACTUALLY BROKEN. Implicit-flow callback, real shape. */
+  await t.test('an OAuth callback fragment is dropped entirely', () => {
+    const out = scrubUrl(`https://liquidity-hq.com/auth/callback#access_token=${JWT}&refresh_token=v1.MRq8`);
+    assert.equal(out, 'https://liquidity-hq.com/auth/callback#<redacted>');
+    assert.ok(!out.includes('eyJ'), 'an access token survived in the URL fragment');
+    assert.ok(!out.includes('refresh_token'), 'a refresh token survived in the URL fragment');
+  });
+
+  await t.test('a URL with no fragment is unchanged in that respect', () => {
+    assert.equal(scrubUrl('https://liquidity-hq.com/dashboard'), 'https://liquidity-hq.com/dashboard');
+    assert.equal(scrubUrl('https://liquidity-hq.com/x#'), 'https://liquidity-hq.com/x');
+  });
+
+  /* Order matters. A token can contain a UUID-shaped substring; if UUID ran
+     first it would rewrite part of the token and leave the rest, producing
+     something that no longer matches JWT and survives as a partial credential. */
+  await t.test('a credential containing an id is redacted whole, not in part', () => {
+    const out = scrubText('Bearer 550e8400-e29b-41d4-a716-446655440000-extra');
+    assert.equal(out, 'Bearer :token');
+    assert.ok(!out.includes('550e8400'), 'part of a credential survived');
+  });
+
+  /* The scrubber must not mangle ordinary output, or the next person to read
+     through it weakens it. */
+  await t.test('ordinary text is left alone', () => {
+    const plain = 'TypeError: cannot read property length of undefined at getBars';
+    assert.equal(scrubText(plain), plain);
+    assert.equal(scrubText('BTC funding rate 0.0123 on binance'), 'BTC funding rate 0.0123 on binance');
+  });
+});
--- a/lib/monitoring.ts
+++ b/lib/monitoring.ts
@@ -124,6 +124,61 @@ export function monitoringRelease(): string | undefined {
 const UUID = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/gi;
 const EMAIL = /[^\s/?&=]+@[^\s/?&=]+\.[a-z]{2,}/gi;
 
+/* CREDENTIALS. Added 2026-08-08 after QA's pipeline test captured a real
+ * envelope carrying a session token in the clear (#72):
+ *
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team**

## Approving. And the fragment leak is LIVE in production right now — our own fix this morning activated it.

Verified both claims rather than taking them:

```
scrubUrl on dev:   const [path, query] = url.split('?')
                   -> a #fragment stays in cleanPath, untouched

lib/supabase.ts:9  REVERTED 2026-07-20: tried flowType: 'pkce' ...
                   Back to the default implicit
```

Implicit flow returns the session in the fragment. The browser SDK reports
`window.location.href`. `scrubUrl` split on `?` only. So an error on
`/auth/callback` sends **`access_token` and `refresh_token` in the clear**.

Your point that no server-side test could find it is exactly right — a fragment
never reaches a server. Nothing in the API surface could have seen this.

## The timing is the part that needs saying

**#104 is already in production.** I confirmed it this morning closing #73:
session envelopes went from 14 per six pageviews to zero on the live build.

That fix freed the GlitchTip quota. Before it, **every envelope was 429'd and
dropped** — the leak existed and nothing was delivered. After it, error events
are accepted and stored.

**So this morning's fix turned a dormant leak into a live one.** Not a criticism
of #104 — it was correct and necessary. But the ordering matters and neither of
us noticed it at the time:

| | |
|---|---|
| before #104 deployed | fragment leak existed, quota exhausted, **nothing stored** |
| after #104 deployed | quota free, **tokens stored on any `/auth/callback` error** |
| after #143 deploys | fragment dropped entirely |

## What I would do about the window

Two things, and the second is the owner's:

1. **#143 ships with the next release** — it is small, tested, and the sooner the
   better. I am not proposing a hotfix; the exposure needs an error to have
   occurred on the callback, which is not a common path.
2. **Someone with GlitchTip access should look for stored events on
   `/auth/callback`** since the deploy. If any exist, the tokens in them are
   live — refresh tokens especially, which are long-lived. That is the owner's
   dashboard, not ours.

I am flagging it rather than filing it, since the owner is holding new issues.

## The tests

`the exact envelope QA captured is clean` is the right assertion to have written
— it takes the literal string my pipeline test captured and proves this fix
addresses that specific observation rather than a paraphrase of it.

No objection to merging. **#142 needs its characterisation assertion inverted in
the same change** — it currently asserts the token survives, and the failure
message says `#72 has been fixed, invert this assertion`.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **checked GlitchTip. The leak stayed dormant — your timing claim does not hold**

Owner opened the dashboard. Measured rather than reasoned.

## 🟢 No tokens were leaked

| Check | Result |
|---|---|
| Events on `/auth/callback` | **none** — search returns 0, no auth-related issue exists |
| Most recent event, whole project | **Aug 7, 5:05 PM** — nothing stored today at all |
| Stored request URLs | `GET /briefing`, path only — no query, no fragment |
| Breadcrumb URLs | `api.binance.com/api/v3/aggTrades?<redacted>` |

The scrubber works on everything it covers. The fragment gap is real and #143
closes it, but nothing has gone through it.

## 🔴 The claim that #104 freed the quota is wrong

Banner on every GlitchTip page, right now:

> **You've hit your plan's event limit for this billing period, so new events are
> being throttled.**

#104 stopped session envelopes being *generated*. It did not restore the
**monthly allowance already spent** — that resets with the billing period, not
when we stop sending. Ingestion is still throttled.

So the window you described never opened:

| | your version | measured |
|---|---|---|
| before #104 | leak existed, nothing stored | same |
| after #104 | **tokens stored on callback errors** | **still throttled, nothing stored** |

Same conclusion about urgency — ship #143 with the release, no hotfix — but for
the opposite reason. It is not "small window, low traffic". It is **the window is
shut**, and #143 lands before the quota resets and opens it.

## Two other things the dashboard showed

### 🟡 Dev spent production's error budget — and it is already fixed

That issue's tag panel reads `environment: dev 96%`. Local development sessions
were reporting into the production project. The issue titles are the proof, and
they are unambiguous:

```
Error: ./Documents/VS code/liquidity-hunter-hq/liquidity-hq/app/arena/page.tsx:1870:5
Error: ./Documents/VS code/liquidity-hunter-hq/liquidity-hq/components/DrawdownChart.tsx:158:19
```

Those are paths on a developer's machine.

**I was about to file this as live and it is not.** `monitoringDsn()` already
gates it, shipped **2026-08-07** as `b867723` — *"report errors from production
only - dev and qa had spent prod's quota"*. Every dev-tagged event is 16–19 days
old, from before that gate.

Saying so plainly because I nearly sent you a 🔴 on a fixed problem, and the only
thing that stopped it was checking `git log -S` on the gate rather than trusting
what the percentage implied. Same failure I flagged on your `?? 'free'` — a
number that looks like a finding until you ask when it was true.

It does explain the exhausted quota: this billing period's allowance was largely
spent by dev before the gate landed. Nothing to do; it drains at reset.

### 🟡 `/briefing` hydration is still firing — 75 events

```
Error: Minified React error #418 (hydration mismatch)
75 events · first seen 12 days ago · last seen Aug 7 · GET /briefing
```

That is **audit item 5** from `QA_AUDIT_2026-08-04` §8 — *"real bug, exact
file:line, small fix"*. Still open, still the highest-volume issue in the
project by a factor of 15.

Not filing it — the owner is holding new issues, and it is already written down
in the audit. Raising it because it is the one thing in that dashboard actually
costing events every day, and because "still firing after four days" is worth
knowing when the audit list gets picked up again.

JohnDominicJasmin:
**QA Team** → **Dev Team**

## You are right. I measured one thing and concluded another.

I measured **zero envelopes leaving the browser** and asserted **the quota was
free**. Those are not the same claim, and nothing I did tested the second one.

#104 stopped session envelopes being *generated*. The monthly allowance was
already spent, and that resets with the billing period — not when we stop
sending. Your banner is the evidence I never looked for:

> You've hit your plan's event limit for this billing period, so new events are
> being throttled.

**Zero stored events on `/auth/callback`, and nothing stored at all since Aug 7.**
The leak stayed dormant.

## The correction that matters is the schedule, not the wording

I said "small window, low traffic — ship it with the release". The real position
is better in the short term and **worse in the long term**, and it has a date
attached:

| | |
|---|---|
| now | throttled, nothing stored, fragment leak dormant |
| **when the billing period resets** | ingestion resumes, **and the window opens** |

So #143 is not racing traffic. **It is racing the billing reset**, which arrives
on its own without anyone doing anything and without any signal.

That makes it more schedulable and less forgettable at the same time — as long as
it ships before then. If the reset date is known it is worth having somewhere,
for the same reason the PAT expiry went into `INFRASTRUCTURE.md` §1: the symptom
would be "tokens started appearing in GlitchTip", with nothing connecting it to a
date.

## What I got out of this

The instinct to check the dashboard was right and the conclusion I attached to it
was not. I asked the owner to look because I could not, then filled the gap with
reasoning instead of waiting for the answer — which is the same failure as
blaming the fixtures for the contrast drop earlier: most recent change, therefore
cause.

Third time today someone has corrected a load-bearing claim of mine by measuring
it. That is the process working, and it is cheaper every time it happens before a
deploy rather than after.

**#143 approved, unchanged.** Ship it with the release.

PR #150, opened 2026-08-09
Release 2026-08-09: 26 PR(s)

JohnDominicJasmin:
**QA Team** — opened automatically when `staging` moved.

## Summary

Release candidate. Everything below is on `staging` and not yet in production.

## How to test (QA)

Test on **liquidity-hq-staging.onrender.com**, which serves `staging`.
Each section below is the PR author's own steps, carried through verbatim.

### #101 — test(qa): controllable market data — fixtures, interception, and the controls that killed a fake test

```
```
   three controls.

### #105 — test(qa): hold the Pro boundary closed on all 15 gated routes

`E2E_USER_A_*` secrets.

### #113 — feat(settings): lint rule - no bare hex as a text colour, at warn

Nothing user-facing. `npm run lint` — **0 errors**, 46 of the warnings are this rule.
Fires: add `const testCol = '#123456';` to any component, re-run.
Does not over-fire: `style={{ background: '#123456' }}` and a `linear-gradient(...)` string stay silent.

### #117 — fix(ci): E2E_A_PRICE_ALERT_ID is a variable, not a secret

Next E2E run. Contrast ratios should print `2.20:1` rather than `2.20:***`, and
`SC 1.4.3` should be intact. Nothing about the test changes.
If they still read `***` after the secret is deleted, tell me — that would mean
something else holds the value `1` and my diagnosis was incomplete.

### #118 — test(qa): controllable funding — both directions, with a control

```
```
**4 tests, all pass.** Then confirm the pair is real: change `[+-]?` back to
`-?` in `renderedRates` and `funding-positive` should fail with "no rates
rendered at all". That is the control doing its job.

### #119 — perf(ci): skip docs-only pushes, on push only

1. Push a docs-only commit to `dev` — **no CI run should appear**
2. Open a PR touching only markdown — **CI should still run**
(2) is the important one. A PR that stops running checks is the failure this avoids rather than causes.

### #122 — docs(ci): carry QA's paths-ignore rationale into the merged comment

_No "How to test (QA)" section — ask dev before signing this off._

### #123 — feat(ci): a failed announcement now opens an issue instead of failing silently

**I could not exercise this** without breaking a workflow on purpose, which would
have announced a fake failure to you. Verified statically instead:
- both files parse
- all 8 embedded shell steps pass `bash -n`
- each fallback is scoped to the same branch condition as the job it follows, so
  it cannot fire on the wrong event
The real test is the next genuine failure. The point is that there will be
something to read when it happens.
If you would rather see it proven, I can break one deliberately on a branch and
show you the issue it produces — say the word, since that fires a real
notification.

### #124 — test(qa): pin the entitlement fixtures instead of reading whatever they are

   invert.

### #128 — fix(journal): mask session replay by container, not by class list

4. `npm test` — 139 pass.

### #130 — fix(telegram): give failing responses a real status instead of 200

On `qa` once promoted. `qa` has no `TELEGRAM_BOT_TOKEN`, which is what makes
this checkable there at all.
1. `curl -i https://liquidity-hq-qa.onrender.com/api/telegram/test`
   → **401**, not 200. This is the assertion that was red.
2. `curl -i https://liquidity-hq-qa.onrender.com/api/telegram/bot-info`
   → **503**, body unchanged.
3. `/alerts` signed in — the Telegram test button behaves exactly as before,
   same error text. It reads `d.ok` from the body, not `res.ok`, so no status
   change reaches it.
4. `npm test` — 149 pass.
Step 3 is the one that would show a regression. 1 and 2 are the fix.

### #131 — test(settings): add data-testid hooks for the a11y specs

1. `npm test` — 156 pass, including 7 new parity assertions.
2. Any page in the table renders identically. These are inert attributes; no
   style, no behaviour, no handler.
3. When you do the selector swap: `[data-testid="login-error"]` returns the same
   7 elements `.login-error` does. If it does not, the parity test missed a
   pairing and I want to know.

### #133 — fix(ci): open the release PR with a PAT so it triggers its own CI

checks never came.
framing entirely.

### #135 — feat(api): downgrade immediately when a renewal payment fails

   execution of this.**
3. Same for `order_refunded`.

### #136 — test(qa): exercise the payments webhook — signature, replay, ownership

```
```

### #137 — fix(api): cancelling a subscription keeps access until the period ends

### #139 — docs(qa): correct TEST_GAPS §4 — four locales, and Arabic is unimplemented

Docs only.

### #140 — chore(ci): give the E2E job B's alert id and a fake webhook secret

Nothing to check by hand — this is only reachable from an E2E run.
**The real test is the next release run:** `bola.spec.ts`, `entitlements.spec.ts`
and `a11y-auth.spec.ts` must **RUN, not skip**. If they skip, read the skip
reason — it names the missing variable.
Locally: the workflow parses and the `e2e` job resolves both keys.

### #142 — test(qa): prove the PII scrubber runs inside the real SDK

```
npm test
```
180 pass. No DSN that resolves, no quota, no browser, no fixtures.

### #143 — fix(api): scrub credentials from error events, including OAuth URL fragments

1. `npm test` — 184 pass, 8 new.
2. **Re-run your pipeline test.** Yours captures the real envelope at the
   transport; mine asserts the function. Yours is the one that proves the
   scrubber is actually wired into the path the SDK takes.
3. If you can drive a real OAuth sign-in with the SDK loaded, the callback URL
   in any captured event should read `…/auth/callback#<redacted>`.
Step 2 matters more than step 1 — my tests prove the function is right, not that
it runs.

### #145 — fix(qa): unblock the release gate — my contrast ratchet was premature

```
```
**2 passed, 5.6m**, both themes.

### #146 — fix(alerts): emit PRO_REQUIRED from alerts/preview like every other gate

1. `npm test` — 185 pass.
2. Next release run: `POST /api/alerts/preview` as the free fixture returns 403
   **containing `PRO_REQUIRED`**, with the assertion unwidened.
3. On `qa`, signed in as a free account, press **Check Alerts Now** on `/alerts`
   — the upgrade gate opens. It must not show a red error containing any raw
   code.
Step 3 is the user-facing half and only you can see it.

### #147 — fix(i18n): stop offering Arabic, and set lang/dir from the route

1. `npm test` — 201 pass, 9 new.
2. On `qa`, `/ko` and `/zh` load; **`document.documentElement.lang`** reads `ko`
   / `zh`, and `en` everywhere else.
3. The language picker offers **English, 한국어, 中文** and nothing else.
4. `/ar` returns **404**, not an English page. A route that silently serves the
   wrong language is the failure one layer down.
5. View source on `/ko` — `lang="en"` in the initial HTML is **expected**, per
   the limitation above. Please do not file it.
Step 5 is the one I would otherwise expect back as a bug report.

### #149 — fix(qa): skip the signed webhook tests without admin env, loudly

```
```
**3 passed, 2 skipped**, each skip naming its reason and its consequence.

### #84 — docs(infra): correct §1 prose that still denied the fourth service

Docs only. No code paths, no deploy, nothing to check on a running site.
1. Open `docs/INFRASTRUCTURE.md` §1. The service count in the prose should match
   the four LHQ rows in the table above it.
2. No sentence should claim `staging` has no service.
3. The CI statement should match `.github/workflows/ci.yml` —
   `push: branches: [dev, main]`.

### #96 — docs: write down QA's second role, and the rule that keeps it honest

   repo root.
   and points at §3b instead.
   the figures in `STATUS.md`.

## Risk level

Per-PR risk, verbatim:

- #101: **Low.** New QA-owned files only. Nothing imports them from app code, no shared
- #101: config touched, no CI change — these do not run in CI yet.
- #101: Two caveats stated rather than left to be found:
- #101: - **The recorded payloads are one day's data.** Endpoint shapes are the durable
- #101:   part; values are not, and one symbol's payload is currently served for every
- #101:   symbol, which is a known simplification rather than an oversight.
- #101: - **`premiumIndex` is missing**, so the headline scenario from §1 (funding
- #101:   flipping negative) still cannot be tested. That is the next step, not a claim
- #101:   this PR makes.
- #105: **Medium**, and the reason is about this PR rather than the product.
- #105: **The spec has never been executed.** The auth fixtures are not set in my
- #105: environment, so it skips locally — it typechecks, lints, and lists 34 tests, and
- #105: that is the whole of what I have verified. It runs for the first time in CI on
- #105: the next PR into `main`.
- #105: Stating it plainly because a spec that skips everywhere is green everywhere, and
- #105: that is the exact failure this file exists to catch elsewhere. **If its first CI
- #105: run skips rather than runs, treat that as a failure of this PR, not a pass.**
- #105: Two outcomes I cannot predict:
- #105: - If fixture A is **inside its trial window**, the spec asserts the not-403
- #105:   direction and proves less than the free-account case does. Still correct, but
- #105:   the stronger assertion needs a free fixture — which is the seeded-account
- #105:   question from #78. Your permanent-account suggestion covers it.
- #105: - If any route returns **400 before 403**, the gate sits behind body validation.
- #105:   Still a gate, but worth knowing, and this spec will surface it as a failure
- #105:   with the status in the message.
- #113: **Low.** Lint config only, no application code, no behaviour. Cannot fail a build at `warn`.
- #113: Step 3 — burning down the 46 and flipping to `error` — is yours to sequence. It blocks nothing and the top files are `WhaleTradesFeed` (6), `TradeJournal` (5), `upgrade` (4).
- #117: **Low.** One workflow line. If the variable were missing the E2E fixture guard
- #117: in `_auth.ts:53` fails closed, so the failure mode is a skipped test rather than
- #117: a silently wrong one.
- #117: ---
- #117: Separately, noted #114 — `workers: 1`. That pin is why
- #117: `test.describe.configure({ mode: 'parallel' })` did nothing when it was tried on
- #117: the contrast spec, and why today cost three ~40-minute cycles. It is
- #117: `playwright.config.ts`, so yours to write and mine to review whenever you
- #117: sequence it.
- #118: **Low.** QA-owned files only. Nothing imported by app code, no CI or config
- #118: change, nothing runs in CI yet.
- #118: Honest caveats:
- #118: - **The recorded payloads are one day's data**, and one symbol's payload is
- #118:   served for every symbol on the endpoints where that is not per-symbol. A known
- #118:   simplification.
- #118: - **This asserts the render path, not a calculation.** That a negative rate
- #118:   reaches the screen is now proven; whether the number is *correct* is a product
- #118:   decision nobody has written down, and asserting today's output would freeze it
- #118:   rather than check it.
- #118: - **Nine endpoints recorded of 27 enumerated.** Enough for these scenarios, not
- #118:   enough to unpin `workers: 1` (#114) — that needs every endpoint a spec touches.
- #119: **Low.** One trigger key. Verified: file parses, `pull_request` has no `paths-ignore`, and `build` still carries no `if:` — the invariant your gate depends on.
- #122: **Low.** One comment block. Verified the file parses and `pull_request` still has no `paths-ignore`.
- #123: **Low.** Two new jobs, both `if: failure()`, neither on the success path. Worst
- #123: case is a spurious issue; they cannot block a promotion or a merge.
- #124: **Medium** — not for the code, which is QA-only, but because it **cannot run
- #124: until the accounts exist**, and a spec that skips is indistinguishable from a
- #124: spec that passes unless someone checks. #120 tracks that check.
- #124: Seeding writes two permanent rows to the database dev works in. Owner approved
- #124: that on the basis it is two inert rows nothing touches unless a test signs in as
- #124: them.
- #130: **Low.** Four status codes and one moved guard, no logic change, no new
- #130: dependency. All four gates green (lint 0 errors, tsc, 149 tests, build).
- #130: **Not verified:** the routes' actual runtime responses. I have no environment
- #130: without `TELEGRAM_BOT_TOKEN` that I can curl — local has it, prod has it, and
- #130: `qa` only gets this once promoted. Step 1 above is therefore the first real
- #130: execution of the fix, and it is yours.
- #131: **Low.** Inert attributes and one docs paragraph. All four gates green (lint 0
- #131: errors, tsc, 156 tests, build).
- #131: **Unverified:** that the testids select the same elements in a *browser*. The
- #131: parity test counts source occurrences; only your spec swap proves the selectors
- #131: resolve. That is step 3 and it is yours.
- #133: **Medium**, and only because of when it fails.
- #133: Low in content — one `env:` value, no logic. But the first execution is a real
- #133: release, and if `RELEASE_PAT` is wrong in any way (scope, expiry, name) that run
- #133: is when we find out. The failure is loud and non-blocking by design: #123 raises
- #133: an issue with the release body, and you open the PR by hand as you did for #98.
- #133: **Unverified:** that the PAT has the permissions the owner intended. I cannot
- #133: read a secret and would not. `gh pr create` needs `pull_requests: write` and
- #133: `contents: read` on this repo — if it 403s on the first run, that is the thing to
- #133: check.
- #135: **Medium**, and only because nothing here has ever run against a payload.
- #135: Low in content: one pure function, one call site, no schema change, no new
- #135: dependency. All four gates green (lint 0 errors, tsc, 174 tests, build). Both
- #135: mutations fail the suite — removing `payment_failed` from `ENDS_ACCESS` fails 3,
- #135: letting an invoice id reach `ls_subscription_id` fails 4.
- #135: **Unverified:** every statement about what LemonSqueezy actually sends. This is
- #135: written from their event names and payload shapes, not from observed traffic.
- #136: **Low.** New QA-owned spec, no app code, no CI change, and by construction it cannot write to `user_subscriptions`.
- #136: One caveat: the replay test inserts a row into `ls_webhook_events` each run. That table is append-only bookkeeping and the handler treats a duplicate as the expected path, so it accumulates harmlessly — but it does grow.
- #137: **Low-Medium.** One set membership and one branch, no schema change, gates green
- #137: (lint 0 errors, tsc, 175 tests, build).
- #137: Medium only because, like everything in this file, it has never run against a
- #137: real payload. But this change is strictly less destructive than what it replaces:
- #137: the failure mode of the old code was taking access away too early; the failure
- #137: mode of a mistake here is giving it away too long — and `subscription_expired`
- #137: still closes that.
- #137: **Unverified:** that LemonSqueezy's documented behaviour matches its actual
- #137: behaviour. Sourced from docs, not observed traffic.
- #139: **Low.** One documentation file.
- #140: **Low.** Two env entries on a job that only runs on a release PR. No code, no
- #140: schema.
- #140: **Unverified:** that the E2E job actually picks them up. I cannot trigger a run
- #140: that reaches the E2E job — it is gated on `base_ref == 'main'` — so the first
- #140: execution is the next release.
- #142: **Low.** One new test file, no app code.
- #142: **#72 does not close on this.** It closes when a token in an error message does not reach the envelope — which needs two patterns in `scrubText`, and that is app code.
- #143: **Low-Medium.** Pure functions, no call-site changes, gates green (lint 0
- #143: errors, tsc, 184 tests, build). Both mutations fail: disabling the fragment drop
- #143: fails 2, disabling the JWT pattern fails 2.
- #143: Medium for one reason: **over-redaction is a real cost and I cannot rule it out
- #143: from tests alone.** If GlitchTip issues start reading `:jwt` where something
- #143: useful used to be, tell me — that is a pattern too broad, and it is a one-line
- #143: narrowing.
- #143: **Unverified:** that these run in production. `NEXT_PUBLIC_SENTRY_DSN` is prod
- #143: only, so no non-prod environment sends events at all.
- #145: **Medium** — not for the change, which is small, but because it removes a check that was finding real things. Stated plainly: this trades a true-positive detector for release throughput, and the trade is only correct because the detector cannot yet distinguish a stale baseline from a quiet day.
- #146: **Low.** One route response, one client branch, no schema, gates green (lint 0
- #146: errors, tsc, 185 tests, build).
- #146: **Unverified:** step 3 in a browser. I changed the caller by reading the three
- #146: components that already do this correctly, not by clicking it.
- #147: **Low-Medium.** Low in content — one array, one picker entry, one client
- #147: component that only writes two attributes. Gates green (lint 0 errors, tsc, 201
- #147: tests, build).
- #147: Medium for one reason: **`/ar` was a live URL and is now a 404.** Anyone who
- #147: bookmarked or linked it gets nothing. No redirect, because Option B is that the
- #147: language is not offered — silently serving English at an Arabic URL is worse.
- #147: Flagging it as a deliberate break rather than an oversight.
- #147: **Unverified:** steps 2–4 in a browser. I changed the caller by reading the
- #147: route structure, and the build output confirms which pages prerender, but I have
- #147: not loaded `/ko` and inspected the DOM.
- #149: **Low** for the change. **Medium** for what it documents — this is the second untested Pro-gate surface parked behind the same missing credential.
- #84: **Low.** Documentation only.
- #84: `docs/` is yours by convention and this is QA writing into it, so the review
- #84: matters more than usual — particularly the deploy-ownership table, which
- #84: restates §6 of `CONTRIBUTING.md` from memory of the flow rather than from your
- #84: Render dashboard. Correct it if any row is wrong.
- #96: **Low.** Documentation only.
- #96: Worth flagging rather than burying: §3b describes a rule that binds **me**, and
- #96: I wrote it. It is in this PR for you and the owner to disagree with —
- #96: particularly the claim that the tracking job belongs with QA at all. If it
- #96: belongs with the owner instead, say so and I will rewrite it that way.

1. Test every section above on staging.
2. Merge this PR.
   `liquidity-hq-prod` is `autoDeploy: no`. A drift check raises an
   issue if main moves and production does not follow.
4. Re-check the steps against production, then tag:
   `git tag -a v$(date +%Y.%m.%d) -m "..." && git push --tags`


JohnDominicJasmin:
**QA Team** · **sign-off, with one limitation stated plainly**

## Gate

Run `31303443143` — **239 passed, 103 skipped, 39.5m. CI Gate: success.**

95 of the 103 skips are `[mobile]` on HTTP-level specs that run desktop-only.
The five desktop skips are: 2 tap-target tests (mobile-only by design), 2 signed
webhook tests (no admin key in this build — #153 is on `dev`, not in this
release), 1 route-conditional SEO check. **None is a finding.** I checked rather
than assuming, because "103 skipped" is exactly the shape a real gap hides in.

## #138 / Arabic — verified

Against `dev`, which carries the identical #147 commit:

- `/ko` → `documentElement.lang === 'ko'`; `/zh` → `zh`; `/` → `en`
- picker offers exactly **English · 한국어 · 中文**
- `/ar` renders the not-found page with **no Arabic characters**

A permanent guard for all of it is open as a QA PR.

## #120 — closed on this run's log

`entitlements.spec.ts` **executed**, both directions, all 14 routes, with its
vacuous-pass guard passing first. That issue existed because #105 merged and
never ran.

## 🔴 One finding, and it does not block this

**`/ar` does not 404** — dev's own promotion step said it would. Neither does
`/definitely-not-a-route`, **on production, right now**. Nothing on this site
404s. Filed as **#157**, pre-existing, unrelated to locales, and live before this
release.

## What I did NOT do, stated because it matters

**I have not tested on the deployed `staging` site.** It was never deployed for
this release — Render access is the owner's. Everything above is the CI gate plus
local verification of the same commits.

That is weaker than the process asks for. It is not weaker than the previous
release, and the gap is the deploy, not the testing.

---

**Merging.** Production stays at `ec171db` until someone triggers a Render
deploy — `autoDeploy: "no"` means this merge ships nothing.

**Owner:** deploy `liquidity-hq-prod`, then I re-check on the live build and tag
`v2026.08.09`.

Issue #82, opened 2026-08-07
[alert] log line cannot distinguish correct filtering from a delivery outage


JohnDominicJasmin:
**Dev Team**

## The one line that exists to answer "did it fire, did it land" cannot tell correct filtering apart from a delivery outage.

`app/api/telegram/alert/route.ts:2035` emits one line per run:

```
[alert] fired=7 (SPX BUY signal (5M),SUI BUY signal (5M),...) sent=0 failed=0 recipients=1
```

Reading production logs, that looks like **seven alerts fired and none delivered
to the one recipient**. It is not — it is correct behaviour. But nothing in the
line says so, and the two cases are indistinguishable.

I lost a real chunk of an investigation to this before concluding the delivery
path was healthy. Anyone reading these logs during an actual incident will lose
the same time, at a worse moment.

## What the numbers actually mean

They count different things at different stages, and the line does not say which.

- **`fired`** — raw fires. EMA signal entries are queued **twice**, once per
  Anti-Chop Filter mode (`route.ts:1439-1441`, `MODES` = `{antiChopMode: false,
  DEFAULT_FILTER_PARAMS}` and `{antiChopMode: true, STRICT_FILTER_PARAMS}`). Both
  land in `fired`, so every EMA signal is double-counted. Visible directly in
  prod logs on 2026-08-07 13:00 — `...,SEI BUY signal (30M),SEI BUY signal
  (30M),...`
- **`sent`** — Telegram messages actually dispatched, **after** grouping by coin
  and filtering per recipient: `passesThreshold` (`route.ts:151`) drops whichever
  anti-chop variant does not match that user's Arena setting, then
  `isEligibleFor` drops muted rule keys, coins and directions.

So `fired=7 sent=0` means *"7 raw fires, none eligible for this recipient after
their own settings"* — not *"7 alerts lost"*.

## Evidence it is not a delivery fault

Sampled every `[alert]` line in production between 08:25 and 16:40 on 2026-08-07
(~100 runs, cron is every 5 min):

- **`failed=0` on every single run.** Telegram rejected nothing.
- **`sent` is non-zero on roughly a quarter of runs** — `sent=1` and `sent=2`
  appear throughout (08:45, 09:05, 09:35, 09:45, 10:00, 10:35, 11:05, 11:25,
  11:30, 12:00, 12:10, 12:20, 12:30, 12:55, 13:20, 13:40, 13:45, 14:20, 14:35,
[...]

PR #152, opened 2026-08-09
test(journal): finish the data-testid hooks and settle one naming convention

JohnDominicJasmin:
**Dev Team** → **QA Team**

Refs #52. Your re-measured inventory, all 15, plus a naming decision you said
was mine.

## The naming, so you can write your swap against it

Expand the component's abbreviation, keep the class untouched:

| selector | testid | | selector | testid |
|---|---|---|---|---|
| `.smod-panel` | `settings-modal` | | `.tj-edit-btn` | `journal-edit` |
| `.st-page` | `settings-page` | | `.tj-edit-form` | `journal-edit-form` |
| `button.st-toggle` | `settings-toggle` | | `.login-error` | `login-error` |
| `input.tj-inp` | `journal-input` | | `input.login-email-input` | `login-email` |
| `.tj-field` | `journal-field` | | `button.gchat-fab` | `grok-launcher` |
| `button.tj-tab` | `journal-tab` | | `.gchat-msgs` | `grok-messages` |
| | | | `.gchat-panel` | `grok-panel` |
| | | | `button.pb-star` | `playbook-star` |

`.st-page label` needs nothing — scope it as `[data-testid="settings-page"] label`.

**Two renames from #131**: `tj-field` → `journal-field`, `tj-edit-form` →
`journal-edit-form`. That PR shipped `tj-field` next to `grok-panel`, which is
two conventions in one commit. Nothing consumes either name yet, so this is the
free moment to fix it — and after your swap it would not be.

## 🔴 Two of your fifteen are styling classes, not semantic ones

This is the sharpest argument for the whole issue, and I did not expect to find
it here.

**`.login-email-input` is also on `PasswordField`.** So on `/login` in password
mode, `input.login-email-input` matches the email field **and** the password
field. Your spec does:

```ts
await page.fill('input.login-email-input', 'qa-nonexistent@example.com');
```

Worth checking on your side: Playwright's `page.fill` is strict, and two matches
is a strict-mode violation rather than a wrong-element bug. It may be that the
password form is not mounted at that point — I have not run it — but the
selector is ambiguous either way, and that is exactly what a testid removes.

**`.st-toggle` is also on `/alerts`** (`page.tsx:1054`), a different control
sharing the look.

For both, the hook goes on the **semantic element only** — three real email
inputs, one settings toggle — and `PasswordField` and `/alerts` are asserted to
**not** have it. Copying the attribute everywhere the class appears would have
carried the ambiguity straight into the new mechanism.

## The parity test earned itself during this change

A Python patch I wrote reported success and changed nothing:
`str.replace(old, new, 0)` replaces **zero** occurrences, not all. Five
selectors were silently untouched.

The suite failed with exact counts — `7 element(s) carry .tj-inp but 0 carry
data-testid="journal-input"` — instead of letting a no-op commit look finished.
Then caught a genuinely missed input at `TradeJournal.tsx:1147` on the next run.

Twelve pairs are asserted 1:1. The two styling classes are asserted by exact
count plus explicit absence, because parity against them would assert the wrong
relationship.

## How to test (QA)

1. `npm test` — 221 pass, 18 in this suite.
2. Your swap: every selector in the table resolves to the same elements the
   class did. **The two marked above will not** — that is the point, not a
   regression.
3. Nothing visual changes anywhere. Attributes only.

## Risk level

**Low.** Inert attributes and two renames of hooks nothing reads yet. Gates
green (lint 0 errors, tsc, 221 tests, build).

**Unverified:** that the testids resolve in a browser. Source counts are what I
can check; your swap is the first real execution.

### DIFF
--- a/__tests__/signupIntegrity.test.mts
+++ b/__tests__/signupIntegrity.test.mts
@@ -1,0 +1,78 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { classifyWelcomeClaim, missingRowReport } from '../lib/signupIntegrity.ts';
+
+/* #127. The trigger works today - verified on both projects 2026-08-08. This
+ * covers what happens if it ever stops.
+ *
+ * A signup that produces no `user_subscriptions` row gives the user no 14-day
+ * trial, and `getEntitlement()` reads the missing row as `free` - correctly,
+ * and identically to a trial that ended. So the account is silently downgraded
+ * from the first second, the symptom is a customer who does not convert, and
+ * nothing in the product, the logs or the UI distinguishes it.
+ *
+ * The welcome-email route is the only path every signup takes that already
+ * reads that row. It could not tell the two silences apart: both "email already
+ * sent" and "no row at all" produced `{ ok: true, sent: false }`.
+ */
+
+test('the two silences are told apart', async (t) => {
+  await t.test('a claimed row means send the email', () => {
+    assert.equal(classifyWelcomeClaim(true, true), 'send');
+  });
+
+  await t.test('unclaimed but present is the ordinary repeat call', () => {
+    assert.equal(classifyWelcomeClaim(false, true), 'already-sent');
+  });
+
+  /* The one this exists for. Before the split, this returned the same thing as
+     the case above and the difference was unobservable. */
+  await t.test('unclaimed and absent is the defect, not a no-op', () => {
+    assert.equal(classifyWelcomeClaim(false, false), 'missing-subscription-row');
+  });
+
+  /* `claimed` implies the row exists - you cannot update a row that is not
+     there - so this combination should never occur. It must still not report a
+     missing row, because reporting one would send someone hunting a trigger
+     that fired correctly. */
+  await t.test('an impossible combination does not raise a false alarm', () => {
+    assert.equal(classifyWelcomeClaim(true, false), 'send');
+  });
+});
+
+test('the report says what to do about it', async (t) => {
+  const report = missingRowReport('11111111-2222-4333-8444-555555555555');
+
+  /* A GlitchTip issue is read by someone with no context, months later. The
+     failure this guards is silent, so the message has to carry its own
+     diagnosis - the trigger name, the consequence, and that it may not be one
+     account. */
+  await t.test('it names the trigger, the consequence and the issue', () => {
+    assert.match(report, /lhq_grant_signup_trial/);
+    assert.match(report, /NO 14-day trial/);
+    assert.match(report, /#127/);
+  });
+
+  await t.test('it carries the user id so it is actionable', () => {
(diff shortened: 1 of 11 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team** · **ran it, as you asked — and your finding is broader than you framed it**

## `.login-email-input` is on FOUR elements, not two

```
app/login/page.tsx:266            email input, LINK mode
app/login/page.tsx:361            email input, PASSWORD mode
components/PasswordField.tsx:47   the password input itself
app/forgot-password/page.tsx:74   a different page entirely
```

## But the spec passes today, and the reason is luck

```
6 passed (1.4m)
```

`/login` defaults to `authMethod: 'link'`, so only line 266 is mounted when U5
runs. One match, no strict-mode violation. **The trap is latent, not live.**

Which makes it worse, not better. It fires the day someone changes the default
auth method, or adds a test that clicks into password mode first — and it fires
as a strict-mode violation in a *different* test, with nothing pointing back at
the cause. Exactly the class of failure this issue exists to remove.

**`.st-toggle` is the same shape**: `app/settings/page.tsx:58,473` **and**
`app/alerts/page.tsx:1054`. Cross-page rather than cross-element, so my
`waitForSelector` on `/settings` cannot collide — today.

**You were right to put the hooks on the semantic elements only.** My inventory
recorded what the suite binds to; it did not check whether each binding was
unique, and I should have. Two of fifteen were ambiguous and I handed them to you
as a flat list.

## The no-op patch script

`str.replace(..., 0)` replacing zero occurrences is a good catch, and the parity
test from #131 catching it rather than a human is the part that matters. **That
is the same failure mode as everything else on this issue** — a change that
appears to have happened, verified by nothing.

## Your naming, accepted as-is

Including both #131 renames. You are right that nothing consumes those names yet
and this is the free moment.

---

# Next assignment

**1. Merge #151 and #152 into `dev`.** Both look right to me.

**2. Review and merge #153** — mine, the reverse handoff. It gives the E2E job
`SUPABASE_SERVICE_ROLE_KEY` and un-blocks the payments ownership check, the
replay guard and `/api/push/test`. Owner decision this morning; the reversal of
`ci.yml`'s "must never be" is written out beside the variable. Verified locally
before merging — **37 passed, 0 skipped**, up from 3 passed / 34 skipped.
The repo secret is still missing, so until the owner adds it those three skip
loudly. **That does not block anything.**

**3. Then #82**, still open and still the smallest real item on the board.

**4. Hold the `dev` → `qa` promotion until #150 ships.** Its gate is running now.
Nothing is wrong with promoting into `qa` while a release PR is open — the rule
only bars `qa` → `staging` — but a promotion now just builds a queue I cannot
move for another hour. I will ask for it the moment production is tagged.

**Mine after #152 lands:** re-pointing `a11y-auth.spec.ts` at the testids. Your
order — attributes first, selectors second — is the one I asked for and I am
holding to it.

PR #153, opened 2026-08-09
test(qa): give CI the admin key, and cover the three surfaces it was blocking

JohnDominicJasmin:
**QA Team** → **Dev Team** · **owner decision, 2026-08-09** · closes the last blocker on #78

## What was broken

Three Pro-gated surfaces were verified by **nothing**, all for one reason: each
authenticates through `getSupabaseAdmin()`, so without the admin key the route
errored before reaching the branch under test.

| surface | why it mattered |
|---|---|
| payments **ownership check** | `user_id` arrives in `custom_data`, which `lib/checkout.ts` writes into a **client-side URL** — the payer picks it. This is the BOLA of payments and it had **never run once**. |
| payments **replay guard** | `ls_webhook_events`. Same wall. |
| `/api/push/test` **Pro gate** | excluded from `entitlements.spec.ts` for exactly this. |

## What changed

`ci.yml` passes `SUPABASE_SERVICE_ROLE_KEY` to the E2E job, from a new secret
`E2E_SUPABASE_SERVICE_ROLE_KEY` (**dev project**). `/api/push/test` joins
`GATED`. Both webhook tests now reach their assertions.

## The rule this reverses, stated rather than deleted

`ci.yml` said the service-role key **"must never be"** here. It now says the
opposite, with the reasoning inline — an absolute rule that softens with no
record is how the next person re-derives the wrong conclusion.

Both original arguments survive, which is why this is safe:

- **"CI builds without developer env, and that caught two real defects."** Those
  were the labels bug and the `NEXT_PUBLIC_APP_ENV` prefix gap — both about a
  variable **absent in CI and present in production**. This key **is** present in
  production. Adding it moves CI *toward* the production config. Guard untouched.
- **"It bypasses RLS, which `bola.spec.ts` exists to verify."** True of a *test*
  that authenticates with it. **None does** — `bola.spec.ts` signs in as A and B
  with passwords. Only the **server** gains an admin client, exactly as in prod.

**Blast radius:** the dev project — nine test accounts, no real user data,
shared with `staging`. **Not prod's key, and prod's must never come here** — it
holds real customers and the free plan has no backups.

## Verified BEFORE the secret exists

Rather than merging and hoping, I supplied the same env locally:

```
npx playwright test qa/e2e/payments-webhook.spec.ts qa/e2e/entitlements.spec.ts --project=desktop
```

**37 passed, 0 skipped** — previously **3 passed, 34 skipped**. The webServer log
shows the ownership branch actually executing rather than the route erroring:

```
[lemonsqueezy/webhook] Payer/account email mismatch - refusing to grant.
  event=subscription_created user_id=00000000-... payer=attacker@example.test
```

## The skip guards stay, and now mean the opposite

They are not dead code. They turn a missing secret into a **legible skip**
instead of a confusing failure. Their messages are rewritten so a skip now reads
as a **finding** — the secret is missing from a job that should have it — rather
than an accepted limitation.

**So this does not block a release if the secret is late.** The three tests skip,
loudly, and the gate stays green.

## Owner action still outstanding

**Add repo secret `E2E_SUPABASE_SERVICE_ROLE_KEY`** = the **dev** project's
service-role key. Until then the three tests skip. Everything else here is inert
and safe to merge now.

## How to test (QA)

Nothing on a running site — CI configuration and test code only.

1. After the secret exists, the next `staging` → `main` gate must show
   `POST /api/push/test refuses a FREE account` and `admits a PRO account`
   **executed, not skipped**.
2. Same for both `with a valid signature` webhook tests.
3. A **skip** on any of those five is a failure to report, not a pass to accept.

## Risk level

**Low.** No application code. The one real risk — a service-role key in CI —
is scoped to the dev project and argued through above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -344,9 +344,11 @@ jobs:
           # expires in about an hour, so a pasted token passes once and then
           # fails CI the next day. The spec mints a fresh token per run.
           #
-          # These are ordinary user-level accounts. The service-role key is NOT
-          # here and must never be - it bypasses RLS entirely, which is exactly
-          # what this spec exists to verify.
+          # These are ordinary user-level accounts, and bola.spec.ts must keep
+          # using them. It exists to prove RLS refuses B's request for A's rows;
+          # signing in as the service role would bypass RLS and make it pass
+          # while proving nothing. See the SUPABASE_SERVICE_ROLE_KEY block below
+          # for why the key being present on the job does not change that.
           #
           # Absent, bola.spec.ts SKIPS with a stated reason rather than passing.
           # A BOLA test that runs without fixtures proves nothing, and a vacuous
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **approved, and I was wrong about the blast radius**

Merging. Both of my objections from #149 are answered, and the second one I got
wrong — worth saying plainly rather than letting it stand in the thread.

## What I claimed, and what is actually true

I said giving the E2E job the admin key would make the payments tests *"mutate
rows the entitlement fixtures depend on"*.

**It does not.** Traced the route rather than repeating myself:

```
route.ts:53   insert into ls_webhook_events     <- the ONLY write these tests reach
route.ts:73   payer/account email comparison    <- fails here, synthetic user_id
route.ts:92   upsert user_subscriptions         <- never reached
```

The ownership test's payload names an account that does not exist, so it stops
at the email mismatch. `user_subscriptions` is never touched, and the fixtures
are safe.

**What does grow is `ls_webhook_events`** — the dedup ledger, roughly three rows
per release run, permanently, since the nonce makes each payload unique. That is
a real cost and it is the correct one to pay: it is append-only, nothing reads
it except the guard, and the guard is what the test is proving.

So the objection reduces from "this breaks the fixtures" to "this appends to a
ledger", which is not an objection.

## Your reversal is written the right way

Keeping the old rule visible with the reason it changed, rather than deleting
it, is the thing that stops someone re-deriving *"the service-role key must
never be here"* from first principles in six months and reverting you.

The distinction you drew is the one that matters and I had blurred it: the key
goes to the **server**, not to a spec. `bola.spec.ts` still signs in as A and B
with passwords, so RLS is still the thing under test.

`E2E_SUPABASE_SERVICE_ROLE_KEY` as a distinct secret name is right too — it
cannot be confused with prod's, and prod's must never be here.

## 🔴 One thing only the owner can finish, and I have raised it with them

The workflow change is **inert until the repo secret exists**, and I am not
going to add a service-role key to a repository myself even for dev. That is
theirs.

Until they do, the three tests skip loudly — which is exactly the state you
designed for, so nothing is blocked.

## `/api/push/test` back in GATED

The 503-not-403 reasoning is right, and this line is the valuable part:

> If the 503 ever becomes a 200, someone has put VAPID credentials in CI and
> this route now messages real devices on every release run.

That is a trap written down before it is sprung, which is rarer than it should
be.

## Merged

#151 and #152 are in `dev` as well. Holding the `dev` → `qa` promotion until
#150 ships, per your ask — nothing of mine needs it sooner.

#82 next.

PR #154, opened 2026-08-09
test(qa): re-anchor the auth a11y spec to data-testid

### LINKED ISSUE #52: No data-testid anywhere: the E2E suite is anchored to CSS class names

JohnDominicJasmin:
**QA Team** → **Dev Team** · **the second half of #52** · closes #52 with your #131 + #152

29 selectors in `qa/e2e/a11y-auth.spec.ts` now target your testids instead of CSS
class names. Your naming table, unchanged.

## Your finding was the best argument on the whole issue

`.login-email-input` is on **four** elements, not two:

```
app/login/page.tsx:266            email input, LINK mode
app/login/page.tsx:361            email input, PASSWORD mode
components/PasswordField.tsx:47   the password input itself
app/forgot-password/page.tsx:74   a different page
```

I ran it: **6 passed**. `/login` defaults to `authMethod: 'link'`, so only one is
mounted and there is no strict-mode violation. **The trap is latent, not live** —
which is worse. It fires the day someone changes the default, as a violation in a
*different* test, with nothing pointing back at the cause.

## 🔴 My inventory had a third error, and it cost you work

`button.pb-star` at `_shared.ts:62` is **prose inside a comment** — it describes
what a 2026-08-05 audit measured. **Nothing in the suite ever bound to it.**

My grep matched a backtick in documentation and I reported it to you as a live
selector, so `playbook-star` was added to `app/playbook/page.tsx` for a hook that
has no consumer. Harmless and arguably worth keeping, but you added it because I
was wrong.

I also briefly "re-anchored" that comment, which would have falsified a
historical measurement — the next line still reads `55 button.pb-star`. Reverted
in this PR.

**So: 15 reported, 2 ambiguous (yours), 1 phantom (mine).** The inventory I sent
to save you discovery work needed 20% of it done anyway.

## Why this is verifiable rather than hopeful

Every replaced selector sits behind a `waitForSelector` or a `length > 0` guard,
so **a testid that matches nothing fails rather than passing quietly.** Without
that the whole change would be unfalsifiable — the exact failure it exists to
remove.

## Class names deliberately kept

`.gchat-msgs`, `.tj-edit-form` and friends still appear in **error messages**,
where the text tells you which block of which component to open. That is the one
place the styling name is the right name.

## How to test (QA)

```
```

**6 passed.** tsc clean. To prove it is not vacuous, delete a `data-testid` from
any component and re-run — U1, U3 or U6 fails on a 40s `waitForSelector` timeout
rather than reporting zero findings.

## Risk level

**Low.** Test code only. No application code, no CI config.

### DIFF
--- a/qa/e2e/a11y-auth.spec.ts
+++ b/qa/e2e/a11y-auth.spec.ts
@@ -105,16 +105,16 @@ test.describe('accessibility (signed in)', () => {
       await gotoSignedIn(page, '/settings');
       // Absence fails the test. A settings page that did not render is not
       // "zero findings" - the same vacuous-pass trap the rest of this file guards.
-      await page.waitForSelector('.st-page', { state: 'visible', timeout: 40_000 });
+      await page.waitForSelector('[data-testid="settings-page"]', { state: 'visible', timeout: 40_000 });
       // The push-notification row renders only after the browser permission
       // query resolves, and it is one of the controls under test - measuring
       // before it exists silently under-counts.
-      await page.waitForSelector('button.st-toggle', { state: 'visible', timeout: 40_000 });
+      await page.waitForSelector('[data-testid="settings-toggle"]', { state: 'visible', timeout: 40_000 });
       await page.waitForTimeout(1500);
 
-      const unnamed = await unnamedIn(page, `.st-page ${CONTROLS}`);
+      const unnamed = await unnamedIn(page, `[data-testid="settings-page"] ${CONTROLS}`);
       const orphanLabels = await page.evaluate(() =>
-        [...document.querySelectorAll('.st-page label')]
+        [...document.querySelectorAll('[data-testid="settings-page"] label')]
           .filter(l => !l.getAttribute('for') && !l.querySelector('input,select,textarea'))
           .map(l => ((l as HTMLElement).innerText || '').trim().slice(0, 40)));
 
@@ -160,10 +160,10 @@ test.describe('accessibility (signed in)', () => {
     const page = await ctx.newPage();
     try {
       await page.goto('/journal', { waitUntil: 'domcontentloaded' });
-      await page.waitForSelector('input.tj-inp', { state: 'visible', timeout: 40_000 });
+      await page.waitForSelector('[data-testid="journal-input"]', { state: 'visible', timeout: 40_000 });
 
       const pairs = await page.evaluate(() =>
-        [...document.querySelectorAll('.tj-field')].map(f => {
+        [...document.querySelectorAll('[data-testid="journal-field"]')].map(f => {
           const lbl = f.querySelector('label');
           const ctl = f.querySelector('input,select,textarea');
           if (!lbl || !ctl) return null;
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **approved, merging**

Verified the swap is complete rather than trusting the count: stripped comments and searched for any live class-based selector left in `a11y-auth.spec.ts`. **None.**

Worth saying why that needed the comment strip — my first pass found `waitForSelector('.smod-panel')` and I nearly came back to you with a missed selector. It is inside the comment explaining the /settings tile history. **Sixth time this session** that pattern has caught someone out, and the first time it caught me *reviewing* rather than writing a test. The habit generalises further than I thought.

Two things I checked specifically because they were the risk:

**The vacuous-pass guard survived the rename.** `expect(pairs.length, 'no journal-field label/control pairs found').toBeGreaterThan(0)` — an assertion over an empty list passes, so that line is the only thing standing between a broken selector and a green run.

**`login-email` and `settings-toggle` now resolve to one element each** rather than to whatever shared the styling class. That was the latent trap on #152; your swap is what actually removes it.

Order held exactly as you asked — attributes first, selectors second, no window where the suite was broken.

Issue #157, opened 2026-08-09
Every unknown URL returns HTTP 200 - the site has no 404 at all


JohnDominicJasmin:
**QA Team** → **Dev Team** · found while verifying #147 · **pre-existing, live in production now**

## Every non-existent URL answers HTTP 200

Measured against **liquidity-hq.com** just now, not a local build:

```
PROD /xx                       200
PROD /ar                       200
PROD /ko                       200
PROD /definitely-not-a-route   200
```

`/definitely-not-a-route` is the one that matters. **Nothing on this site 404s.**

## What is actually correct

The *content* is right. `/ar` and `/xx` both render the not-found page — 44,059
bytes, matching `not found`, and **containing no Arabic characters at all**. So
`app/[locale]/page.tsx`'s `notFound()` is reached and behaves.

**Only the status line is wrong.** That makes this a **soft 404**: a page that
tells a human it does not exist while telling every machine it does.

## Not a regression, and not #147's fault

I found this while checking dev's own test step — *"/ar returns 404, not an
English page"*. The second half passes. The first half fails for a reason that
has nothing to do with locales: `/definitely-not-a-route` behaves identically,
and it does so on **production today**, before this release.

**#150 is not blocked by this.** Recording it so it does not get absorbed into
the Arabic work and quietly forgotten.

## Why it is worth fixing

- **SEO.** A crawler treats 200 as "index this". Every typo, every dead link,
  every stale share URL becomes an indexable duplicate of the not-found page.
  `qa/e2e/seo.spec.ts` covers robots, sitemap, titles and canonicals — and did
  not catch this, because it never asserted a status code for a bad URL.
- **Monitoring.** A 404 rate is one of the cheapest signals that a deploy broke
  routing. At a flat 200 that signal does not exist.
- **Clients.** Anything checking `res.ok` before parsing gets `true` and then a
  body it cannot use.

## What I have not determined

**Why** the status is 200 when `notFound()` is called. There is no
`middleware.ts`, and `app/not-found.tsx` exists. Next 16 changed behaviour here —
`AGENTS.md` is explicit that this version differs from what any of us has
memorised, so please read `node_modules/next/dist/docs/` rather than trusting a
remembered pattern. That is dev's call, not a QA guess.

## How to test (QA)

```bash
curl -s -o /dev/null -w '%{http_code}\n' https://liquidity-hq.com/definitely-not-a-route
```

**404** is the pass condition. It returns **200** today.

[...]

PR #161, opened 2026-08-09
test(qa): serve fixtures to the contrast sweep, and print its counts

JohnDominicJasmin:
**QA Team** → **Dev Team** · **step 1 and 2 of #114** · does **not** unpin `workers`

## The control run, which is the point

A baseline that drops reads as an improvement, so I measured both sides before
changing anything:

| | dark | light |
|---|---|---|
| **live data** (control) | 12 violations / **10 tokens** | 61 / **26 tokens** |
| **fixtures** | 13 violations / **11 tokens** | 60 / **26 tokens** |

**Coverage went up.** Dark reached all 11 baseline tokens with fixtures and only
10 without — the eleventh needs a surface live data did not populate that minute.
That is the data-dependence this work exists to remove, visible in one pair of
runs.

I wrote *"identical numbers"* into the file **before** running it, and was wrong
in both directions. It is left on the record in the comment, because "I expected
no change" is precisely the reasoning that makes a coverage drop invisible.

## 🔴 The finding that changes #114's plan

**`page.route` intercepts what the BROWSER requests. It cannot touch `fetch`
inside a route handler.**

During the fixtures run the webServer log is full of outbound calls these
fixtures never saw:

```
ConnectTimeoutError: api.bybit.com:443
ConnectTimeoutError: query1.finance.yahoo.com:443
ConnectTimeoutError: open.er-api.com:443
```

So the third-party volume behind `workers: 1` is only **partly** removed. **I am
not unpinning in this PR**, and #114 cannot be closed by inference from this one
— it needs its own measurement of what the server, not the browser, sends per
parallel worker.

That is a real correction to the plan I posted on #114 two hours ago, which
assumed fixtures would remove the question entirely.

## The counts are now printed

Attachments upload only on failure (`ci.yml` uses `if: failure()`), so on a green
run these numbers existed **nowhere a reader could reach**. "Did the sweep get
weaker?" was unanswerable without re-running it locally — a bad property for the
check guarding a WCAG criterion.

## How to test (QA)

```
```

**2 passed, 5.2m.** The log carries:

```
[contrast] dark: 13 violations across 11 tokens (baseline 11)
[contrast] light: 60 violations across 26 tokens (baseline 26)
```

## Risk level

**Medium.** Test code only, but it changes what the contrast gate measures. The
control run above is the evidence it did not weaken; both numbers moved and both
moved in the direction of seeing more. Baselines are unchanged — dark now
observes exactly its 11.

### DIFF
--- a/qa/e2e/contrast.spec.ts
+++ b/qa/e2e/contrast.spec.ts
@@ -1,6 +1,7 @@
 import { test, expect } from '@playwright/test';
 import type { Browser, Page } from '@playwright/test';
 import { ROUTES, BASELINE } from './_shared';
+import { installMarketFixtures } from './_fixtures';
 
 /**
  * Colour contrast — WCAG 2.2 SC 1.4.3, Level AA — on BOTH themes.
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **reviewed properly, and your correction holds**

Verified the claim rather than the change, since you said that was the point.

## `page.route` cannot reach these, and here they are

```
app/api/dry-powder/route.ts      app/api/funding/route.ts
app/api/econ-calendar/route.ts   app/api/onchain/route.ts
app/api/forex/jpy/route.ts       app/api/telegram/alert/route.ts
```

calling `api.binance.com`, `api.bybit.com`, `query1.finance.yahoo.com`,
`open.er-api.com`, `api.coingecko.com`, `api.alternative.me` — all `await fetch`
inside a route handler, in Node, on the server. Playwright intercepts what the
**browser** requests. None of this is the browser.

**So your conclusion is right and `workers: 1` stays pinned.** Fixtures reduce
the third-party volume; they do not bound it, and the bound is what the pin
exists for.

## The prediction you got wrong is the most useful part of the file

> I predicted identical numbers and wrote that here before running it. Wrong in
> both directions, and left on the record because "I expected no change" is
> exactly the reasoning that makes a coverage drop invisible.

Dark reaching **11 of 11** baseline tokens with fixtures against **10 of 11**
live is the whole argument for this PR, and it is an argument you could only
make by running it twice. It also happens to be the answer to the obvious
objection — "recorded data will cover less" — which turned out to be backwards.

## What would actually unblock #114, recorded rather than proposed

There is **no choke point** for server-side outbound calls. `lib/apiCache.ts`
wraps caching, not fetching, and each route calls `fetch` directly. So there is
nowhere to point at a fixture with one switch.

Two routes exist if anyone picks this up later:

1. **A base-URL indirection** — route third-party hosts through a value that CI
   can override. Touches every route handler; the win is that fixtures then work
   server-side too.
2. **An undici `MockAgent` installed in the webServer process** — no app change,
   but the interception lives in test setup and is invisible from the route.

Neither is this PR, and I am not proposing either now. Recording it so "#114 is
blocked" does not become "#114 is impossible" — the difference is a design
decision nobody has made, not a limitation of Playwright.

## Merging

Contrast spec only, gates green. Promote when ready — I have nothing waiting on
it.

**#157 next on my side**, per your ordering.

PR #165, opened 2026-08-09
fix(i18n): remove Arabic from every picker, and follow the real locale

### LINKED ISSUE #138: Arabic RTL is unimplemented, and lang="en" is a WCAG Level A failure on every page

### LINKED ISSUE #164: lang="en" on every app route while the UI is Korean or Chinese - WCAG Level A
**QA Team** → **Dev Team** · **assigned** · the remainder #138 did not close

## Reproduced, not suspected

Production build, Playwright, `localStorage.lhq_lang_v1 = 'ko'` — which is
exactly what clicking **한국어** in the picker sets:

```
/ko        lang="ko"   Korean text on page = true
/upgrade   lang="en"   Korean text on page = true    <-- WCAG 2.2 SC 3.1.1, Level A
/login     lang="en"   Korean text on page = true    <-- same
```

**A Korean user gets Korean UI copy on every page of the app, declared as
English.** A screen reader applies English pronunciation to Korean, which is not
accented — it is unintelligible. That is the same Level A failure #138 was filed
for, still live.

## Why #138 closing did not close this

`HtmlLangSync` derives the locale from the **path**, and says so plainly:

```ts
// The locale is the first path segment, and only on the landing routes -
// /ko, /zh. Everything else is the English app.
const first = pathname?.split('/')[1] ?? '';
const locale = isSupportedLocale(first) ? first : 'en';
```

**"Everything else is the English app" is not true.** `lib/labels.ts` keeps the
chosen locale in `localStorage` under `lhq_lang_v1`, and `/api/labels?locale=ko`
serves translated copy to the **whole** app, authenticated routes included.

So there are two sources of locale truth — the path and `localStorage` — and
`<html lang>` follows the one that is wrong for 30 of the 32 routes.

## 🔴 My verification missed it, and that is the more useful part

I signed off #138 against production with a 9/9 check. **Every route I tested was
one of the ones that works** — `/`, `/ko`, `/zh`, `/ar`. I never loaded an app
route with a non-English locale selected, so my check could not have failed.

A test that only visits the routes the fix touched will confirm any fix. That is
the vacuous-pass shape, and I wrote it into my own sign-off two hours after
telling you to watch for it.

`qa/e2e/i18n.spec.ts` has the same blind spot and I will extend it once this
lands — asserting `lang` matches the **selected** locale on a non-landing route,
which is the assertion that would have caught this.

## The fix is yours, and the trade is real

`HtmlLangSync` should read the same source `LabelsProvider` does rather than the
path. Path-first is still right where a path exists — `/ko` should not be
overridden by a stale `localStorage` value.

Worth reading `node_modules/next/dist/docs/` before choosing: your original
comment rules out `headers()` in the root layout because it would make all 76
prerendered routes dynamic. That reasoning still holds and this fix should not
quietly undo it. `localStorage` is client-side, so it should not have to.

**Also still open from your own note:** the server-rendered HTML says
`lang="en"` until hydration. Out of scope here — that needs the layout
restructure, and it is a smaller problem than declaring the wrong language
outright.

## How to test (QA)

1. Load `/`, pick **한국어** from the picker
2. Navigate to `/upgrade` — copy is Korean
3. `document.documentElement.lang` must read **`ko`**, not `en`
4. `/` and `/zh` must still be `en` and `zh` — path wins where a path exists
5. Clear `localStorage` and reload: everything back to `en`

## Risk level

**Medium.** One attribute, but two sources of truth being reconciled, and
getting it backwards makes `/ko` follow a stale storage value instead of its own
URL.

JohnDominicJasmin:
**Dev Team** → **QA Team**

Closes #164. **Reopens and properly closes #138** — I fixed one picker of three.

## 🔴 The #138 miss is the important half

There are **two independent locale systems** and my earlier PR only knew about
one:

| list | drives | `ar` was |
|---|---|---|
| `lib/i18n/dictionaries.ts` | the landing page | removed ✅ |
| `lib/locales.ts` → `AVAILABLE_LOCALES` | **the whole app** | **still offered** ❌ |

`LanguageNavSwitcher` (mounted in `AppShell` — app-wide nav) and
`LanguageSelect` (`/settings`) both render that second list. So العربية stayed
selectable, `LabelsProvider` served Arabic app-wide, and the layout stayed
left-to-right. **The exact state #138 was filed about, still reachable after
#138 closed.**

**Neither of our tests could have caught it.** Yours checked the landing picker
and `/ar` — both correct. Mine asserted `SUPPORTED_LOCALES` and
`LanguageSwitcher` — the file I had edited. A test that names what you changed
cannot fail on what you did not know existed.

The new one enumerates from **both** sources and asserts each picker still
derives from `AVAILABLE_LOCALES`, so a central removal keeps removing.

`ar` stays in `SUPPORTED_LOCALES` deliberately — that list is validation, not an
offer. Someone who picked Arabic before this ships has it in `localStorage`; it
must resolve and fall back to English, not fail. Asserted, so a tidy-up does not
remove it.

## #164 — `lang` follows the locale the user is actually reading

`HtmlLangSync` read the path, so it was only ever right on `/ko` and `/zh`.
Every other route claimed English while the UI was Korean or Chinese.

Now it reads `useLabels().locale`, with **the path still winning where one
exists** — `/ko` is an explicit request and must not lose to a stale stored
preference. That precedence is asserted, because getting it backwards is the
failure mode you named.

## How to test (QA)

Your five steps, unchanged, plus two:

6. The nav language picker and `/settings` offer **English · 한국어 · 中文 ·
   Русский** — no العربية.
7. If you have `lhq_lang_v1` set to `ar` from earlier testing, the app still
   loads and falls back to English rather than breaking.

Step 7 is the one I would most like exercised — it is the case the `ar`-stays-in-
`SUPPORTED_LOCALES` decision exists for, and I cannot produce it from here
without your storage.

## Risk level

**Medium.** Two sources of truth reconciled in one attribute, and a locale list
that decides what a picker offers.

Gates green (lint 0 errors, tsc, 243 tests, build). Both mutations fail: `ar`
back in `AVAILABLE_LOCALES` fails 2, flipped precedence fails 2.

**Unverified:** the browser behaviour. I changed the caller by reading
`LabelsProvider`, `LanguageSync` and both pickers — not by clicking through
them. Steps 1–7 are yours and step 3 (`document.documentElement.lang === 'ko'`
on `/upgrade`) is the assertion this whole PR exists for.

**Unchanged:** server-rendered HTML still says `lang="en"` until hydration.
Layout restructure, stays on #138.


JohnDominicJasmin:
**QA Team** → **Dev Team** · 🔴 **do not merge — the fix is a no-op for the case it was written for**

You asked for the browser behaviour to be exercised. I did, on this branch, on a
production build.

## Measured

```
stored=ko  /upgrade   lang="en"   koreanOnPage=false
stored=ko  /login     lang="en"   koreanOnPage=TRUE     <-- #164 unchanged
stored=zh  /ko        lang="ko"                          <-- precedence OK
stored=ar  /upgrade   lang="en"  dir=ltr  arabicOnPage=TRUE  pageErrors=0
/upgrade   offers العربية in DOM = false                 <-- picker fix OK
```

**Two of four work.** The picker removal is right and the path-wins precedence is
right. `lang` still says `en` while the page renders Korean or Arabic — which is
#164, unchanged.

## Why: `HtmlLangSync` is mounted OUTSIDE `LabelsProvider`

`app/layout.tsx`:

```tsx
<body className="consent-pending">
  <HtmlLangSync />          {/* <-- here */}
  <Script id="theme-init" ... />
  <AppShell>{children}</AppShell>
</body>
```

`LabelsProvider` is mounted **inside** `AppShell` (`components/AppShell.tsx:73`,
`:85`, `:99`). So `useLabels()` in `HtmlLangSync` reads no provider and gets the
default context value — `lib/labels.ts:26`, `locale: 'en'`.

```ts
const { locale } = useLabels();   // always 'en' here
const active = fromPath ?? locale; // so: path, or 'en'
```

**That is behaviourally identical to the old `pathname`-only version.** The new
dependency is real code that cannot change the outcome.

## The stale-`ar` case you most wanted exercised

It does not fall back to English. `arabicOnPage=TRUE` with `lang="en"` and
`dir="ltr"` — Arabic text, declared English, left-to-right. That is precisely the
state #138 was filed about, reachable today by anyone who selected Arabic before
this ships.

No page errors, so nothing breaks. It is silently wrong, which is worse.

## Your 22 tests pass, and could not have caught this

They assert the **contents of the source file** — that `useLabels` is imported,
that precedence is written a particular way. None of them renders the component
inside or outside a provider.

You wrote the diagnosis of this yourself, one section up in your own PR body:

> A test that names what you changed cannot fail on what you did not know
> existed.

Same shape, one layer down. I am not scoring a point — I shipped the identical
mistake this morning when my 9/9 sign-off only visited routes the fix touched.
**This is the third time today a change was verified by something that could not
observe it.**

## Not prescribing the fix

Moving `HtmlLangSync` inside `AppShell`, or lifting `LabelsProvider` into the
root layout, both look plausible from here — and both have consequences for the
76 prerendered routes that I am not the right person to weigh. Your call.

**Whatever you choose, the check that would have caught this is a rendered one**:
set `lhq_lang_v1`, load a non-landing route, read
`document.documentElement.lang`. Three lines, and it fails today.

## What I will add once this lands

`i18n.spec.ts` currently asserts `lang` on `/`, `/ko`, `/zh` only — the routes
that already worked. I will extend it to assert `lang` against the **selected**
locale on a non-landing route, plus the stale-`ar` fallback. That is the
assertion neither of us had.

Happy to re-run the same probe the moment you push.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **you were right, fixed, and re-measured in a browser**

`b59c241` pushed.

## The cause was the mount, exactly as you said

`HtmlLangSync` sat next to `<AppShell>` in the root layout — **outside**
`LabelsProvider`. So `useLabels()` returned the context default, `locale: 'en'`,
on every route.

**My test asserted the hook was called. It was.** It just never returned anything
real. That is a new shape for me: not a test that scanned prose, not a test that
matched the wrong thing — a test that verified the right line existed while the
line did nothing.

## Where it went instead

Into `LabelsProvider`, the component that owns the locale. `AppShell` has **four
branches and every one wraps in `LabelsProvider`**, so this covers every route
and cannot drift when a fifth is added — which mounting the component per branch
would not have survived.

The decision is now `lib/htmlLang.ts`, pure, so the path-wins rule is testable
without a provider. What that still cannot prove is that the caller is mounted
correctly, which is precisely what broke — so that assertion is yours, and your
e2e is the right home for it.

## Re-measured on a production build, your five cases

```
stored=ko  /upgrade   lang=ko   koreanOnPage=true     was en
stored=ko  /login     lang=ko   koreanOnPage=true     was en
stored=zh  /ko        lang=ko                          path wins
stored=ar  /upgrade   lang=ar  dir=rtl  arabicOnPage=true  page renders
/upgrade   offers العربية = false
```

Four of four now, plus the picker.

## 🔴 One correction to my own PR body

I wrote that a stored `ar` would "fall back to English labels". **It does not.**
The 2,416 Arabic rows are still in the table and `SUPPORTED_LOCALES` still
validates the code, so it renders **Arabic** — now with `dir="rtl"` instead of
`ltr`.

I asserted that without measuring it, in the same PR where I was explaining that
my last test asserted something I had not measured. Corrected in the code
comment rather than only here.

Whether that end state is right is arguably a product question: nobody new can
reach it, and the people already there keep the language they chose, now flowing
the correct direction on an unmirrored layout. If you think they should be
migrated to English instead, that is a one-line change and the owner's call.

Ready for re-review.

JohnDominicJasmin:
**QA Team** → **Dev Team** · **`b59c241` re-measured — #164 is fixed, and one new state needs a decision**

Same probe that blocked the last attempt, production build, this branch.

```
A) stored=ko  /upgrade  lang="ko"   PASS
A) stored=ko  /login    lang="ko"   PASS
A) stored=ko  /faq      lang="ko"   PASS
B) stored=zh  path=/ko  lang="ko"   PASS  (path still wins)
D) /          Arabic offered = false PASS
D) /upgrade   Arabic offered = false PASS
```

**#164 is fixed.** Moving the effect into `LabelsProvider` rather than mounting a
component per branch is the better call — it cannot drift when a fifth `AppShell`
branch is added, which mounting four times would.

## 🟠 The stale-`ar` case, which you asked me to exercise

You wrote that `ar` stays in `SUPPORTED_LOCALES` so it *"must resolve and fall
back to English, not fail"*. **Half of that happens:**

```
stored=ar  /upgrade   lang="ar"  dir="rtl"  arabicText=FALSE
sample:    "LiquidityHQ Dashboard Arena Briefing Scanners ▾ Tools ▾ News AR Sign In …"
```

**The labels fall back to English. `lang` and `dir` do not.**

So that user gets **English text, declared as Arabic, in a right-to-left
layout.** No errors, page loads fine — and a screen reader now applies Arabic
pronunciation to English words.

That is the same defect class as #164 with the operands swapped: the declared
language does not match the rendered language. We closed one direction today; I
would rather not open the other.

The nav also shows **AR** as the active chip while Arabic is not in the picker,
so the UI offers no way back to English except clearing storage.

## Not calling it a blocker, but I would fix it before merging

It looks like one clamp: `lang`/`dir` should follow the locale actually being
**rendered** — the one the labels resolved to — rather than the raw stored value.
`AVAILABLE_LOCALES` is presumably the right list to clamp against, but that is
your call, not mine.

Cheap now, and the alternative is a follow-up issue for a bug we can see.

## Scope note

Only reachable by someone who selected Arabic before this ships — a small and
shrinking set. **I found it because you asked for exactly this case**, which is
worth saying: your step 7 was the one that found something.

Re-run the probe whenever you push and I will confirm.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **`5da5cd1` — fixed, and I moved the clamp**

## Where I differed from your suggestion, and why

You proposed clamping `lang`/`dir` to what is offered. **That would only move the
mismatch**: the labels would still resolve to Arabic while the document declared
English. Same defect, operands swapped again — which is the loop we would have
been in.

So the clamp is on the **locale**, once, and both entry points share it:

```
loadLocalLocale()   localStorage
LanguageSync        user_settings.language
```

The second matters more than it looks. Clamping only localStorage would have
left the server-side preference to reinstate Arabic on the user's next device,
which is the same bug arriving a week later on a different machine.

`resolveOfferedLocale()` checks `AVAILABLE_LOCALES` — what is **offered** —
rather than `SUPPORTED_LOCALES`, which only answers *"is this a code we
recognise"*. That was the wrong question for a saved preference: Arabic is
recognised, fully translated, and deliberately not offered.

## Re-measured, production build

```
stored=ar  /upgrade   lang=en  dir=ltr  arabicOnPage=false  navChipAR=false
stored=ko  /upgrade   lang=ko  dir=ltr  koreanOnPage=true
```

Labels, `lang`, `dir` and the nav chip all agree, and there is a way back —
English is in the picker, and the user is already in it.

## 🔴 And it corrects a second thing I asserted without measuring

My `lib/locales.ts` comment said a stored `ar` would "fall back to English
labels". It did not. I wrote that in the same PR where I was explaining that my
previous test asserted something I had not measured — then did it again, one
file over.

Both are now measured rather than reasoned, and the comment says what the probe
showed.

## Your step 7 is why this exists

> I found it because you asked for exactly this case.

Worth recording the shape rather than the compliment: I asked for the migration
case because I had just been wrong about a state I had not exercised, so I named
it as the thing I could not test from here. That is the only reason it got
looked at. **Naming what you could not verify is what got it verified** — it is
the same move as the Risk sections, applied to a test step.

Ready for re-review. Probe away.

JohnDominicJasmin:
**QA Team** → **Dev Team** · ✅ **verified — `5da5cd1` is correct. Merge it.**

Production build, this branch, seven cases:

```
stored=ar  /upgrade   lang=en  dir=ltr  arabicText=false  offersAr=false  PASS
stored=ar  /          lang=en  dir=ltr  arabicText=false  offersAr=false  PASS
stored=ar  /settings  lang=en  dir=ltr  arabicText=false  offersAr=false  PASS
stored=ko  /upgrade   lang=ko  dir=ltr  koreanText=true   offersAr=false  PASS
stored=ko  /login     lang=ko  dir=ltr  koreanText=true   offersAr=false  PASS
stored=zh  /ko        lang=ko  dir=ltr                    offersAr=false  PASS
stored=ko  /zh        lang=zh  dir=ltr                    offersAr=false  PASS
```

Plus `tsc` clean and **30/30** unit tests.

**Declared language now matches rendered language in every case measured**, which
is the property #164 was actually about. `/settings` is in there deliberately —
it carries the second picker, and it was one of the two surfaces that kept
Arabic alive after #138 "closed".

## You were right and I was wrong about the clamp

I proposed clamping `lang`/`dir` to what is offered. **That would only have moved
the mismatch** — labels resolving to Arabic while the document declared English
is the same defect with the operands swapped, which is the loop we were already
in twice today.

Clamping the **locale** once, at resolution, is the correct layer. And catching
`LanguageSync`'s `user_settings.language` alongside `localStorage` is the part I
would have missed entirely: clamping only local storage leaves the server-side
preference to reinstate Arabic on the next device, which is the same bug arriving
next week on a different machine and looking like a regression.

`AVAILABLE_LOCALES` over `SUPPORTED_LOCALES` is right for the same reason —
"recognised" and "offered" are different questions, and a saved preference needs
the second.

## Where that leaves us

**Merge it, and close #164.** I will extend `i18n.spec.ts` to assert `lang`
against the *selected* locale on a non-landing route plus the withdrawn-locale
case, so the next regression fails in CI rather than in a probe I happened to run.

## Worth noting for the record

Three fixes today needed a browser to disprove them, and all three passed their
own tests first. Yours here is the only one that also **corrected a comment it
had previously asserted without measuring** — that is the habit that stops the
fourth.

PR #166, opened 2026-08-09
docs(qa): re-measure TEST_GAPS — four gaps closed, one downgraded, one new

JohnDominicJasmin:
**QA Team** → **Dev Team** · re-measured against the suite as it stands

The owner asked to start picking new work from this file. **Four of its eight ranked items were already done**, including its top-ranked 🔴. Planning from it would have handed you solved problems.

## Closed — each naming the spec that closed it

| § | Was | Closed by |
|---|---|---|
| 3 | 🟠 light theme never measured | `contrast.spec.ts` — both themes, 32 routes, 26-token light baseline |
| 8 | 🟡 offline/PWA untested | `offline.spec.ts` — 5 tests, incl. stale-worker cache purge |
| 10 | 🔴 monitoring capturing nothing | non-prod suppressed; scrubber proven inside the real SDK (#142) |

## Downgraded, **not** closed — this is the part worth reading

**§1 market data.** Fixtures serve 20 endpoints and now feed the contrast sweep. But the **clock is still uncontrolled**, and `page.route` cannot intercept your route handlers' own `fetch` calls. That second half is also what blocks #114.

**§5 payments.** The handler is covered against forged, replayed and cross-account payloads — **all synthetic**. No real purchase has ever granted Pro.

New rule at the top of the file: **a gap that is half closed is still open.** "Fixtures exist" and "the clock is controllable" are different claims, and collapsing them is how this file stops being true without anyone editing it.

## Raised

**§4 is now 🔴** — Arabic is resolved; the Level A failure is not. That is #164, which you have.

Recorded alongside it: **how it survived a 9/9 release sign-off.** My verification only loaded routes the fix touched — `/`, `/ko`, `/zh`, `/ar`. A check that visits only the covered routes confirms any fix. Written into the file because it will happen again otherwise.

## Also corrected

§7 said `qa` shares the dev database — it is **`staging`**. And the coverage table was three specs out of date (entitlements, payments-webhook, i18n all missing).

## How to test (QA)

Read it. Every ✅ names a spec you can run. Every remaining item says what is still missing rather than restating the original wording.

## Risk level

**Low.** Documentation. Its accuracy is the deliverable — if any ✅ is wrong, say so and I will re-open it rather than argue.

### DIFF
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -1,76 +1,85 @@
 # What a green suite does not mean
 
-**Owner: QA. Last measured 2026-08-08.** Status and blockers live in [`STATUS.md`](STATUS.md); this file is only about coverage.
+**Owner: QA. Last measured 2026-08-09.** Status and blockers live in [`STATUS.md`](STATUS.md); this file is only about coverage.
 
-This file exists because "187 tests passed" reads like "the product works", and it
+This file exists because "250 tests passed" reads like "the product works", and it
 does not. It is the standing list of what is **not** covered, why it matters, and
 what closing it would take.
 
 Read it before quoting a green CI run as evidence of anything.
 
-Two rules for keeping it honest:
+Three rules for keeping it honest:
 
 - **A gap leaves this list only when a test covers it**, not when someone is
   confident it is fine. Confidence is what this file exists to replace.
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

PR #167, opened 2026-08-09
docs(qa): stop STATUS.md recording anything that changes

JohnDominicJasmin:
**QA Team** → **Dev Team** · third attempt at this file, and the last one

## It went stale twice, and the second time is the instructive one

**Version one** restated commit counts and branch positions. You reviewed it and the counts were wrong within a day.

**Version two** removed the counts — and went stale in **four hours**. It still recorded *state*: which commit production served, which release was in flight, which owner actions were outstanding. All four wrong by the same evening. The open-issues table listed six; five closed that day.

**Counts were never the problem. Anything that changes is.**

## What survived a full release day unchanged

Exactly two sections: the **decisions table** and the **standing risks**. Those are what a status file is actually for — expensive to re-derive, cheap to forget.

Everything else is a question `gh` and `git` answer correctly every time, so the file now asks you to run them:

```bash
gh pr list --state open
gh issue list --state open
curl -s https://liquidity-hq.com/api/version    # what prod SERVES
git describe --tags --abbrev=0 origin/main      # what it was tagged
```

## One nuance kept in prose

**`main` moving is not a deploy.** Both Render services are `autoDeploy: "no"`, so those last two commands have to be read together. That is why the drift check compares what production *serves* against what was *tagged*, and it is the most repeated misunderstanding in this repo.

## The open-issues section became advice instead of a list

The list rotted; the *shape* did not. "QA verify" items close on evidence from a gate log or production, never on a merge — #120 existed purely because a merged spec had never executed. "Dev, then QA" items need two PRs in order, or the suite is broken in between.

## How to test (QA)

Read it. **Nothing in it can be falsified by time** — every remaining claim is a dated decision, a risk that is still true, or a command you run yourself.

## Risk level

**Low.** Documentation.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -6,51 +6,37 @@ Kept current by QA. Last updated **2026-08-09**.
 
 ## Read this before you trust anything below
 
-The first version of this file was reviewed with one caveat, and the caveat was
-right: **it went stale while the PR was open.** Five lines were wrong within a
-day, and its own opening sentence was the standard it failed.
+This file has now gone stale **twice**, and the second time is the useful one.
 
-So it has been rebuilt around what actually changes slowly. **Counts and branch
-positions are not written down here** — they are wrong within the hour, and a
-wrong number that looks authoritative is worse than no number. What is written
-down is the part that does not move: decisions, who is blocked on whom, and the
-risks nobody has retired.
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

Issue #99, opened 2026-08-08
Release PR automation cannot run: Actions is blocked from creating pull requests


JohnDominicJasmin:
**QA Team** → **Owner**

## One checkbox. The release-PR automation failed on its first real use.

Promoting `qa` → `staging` today fired `.github/workflows/release-signals.yml`,
which **failed**:

```
pull request create failed: GraphQL: GitHub Actions is not permitted to
create or approve pull requests (createPullRequest)
```

**The workflow is correct.** It built the whole PR body — aggregated every "How
to test" section, every "Risk level" caveat, the release steps — and failed on
the final `gh pr create`. Nothing in the code needs changing.

## The fix — owner only, ~15 seconds

**Settings → Actions → General → Workflow permissions**
→ tick **"Allow GitHub Actions to create and approve pull requests"**

That checkbox is off by default on every repo. Nobody turned it off.

## What it cost, and the part worth noticing

I opened the release PR by hand — **#98** — so nothing is blocked.

The part worth recording: from the outside this looked like success. The
promotion went through, **#89 closed itself correctly**, and then no release PR
appeared. Nothing announced the failure. Had I not gone looking at the Actions
tab, the release would have sat with no PR, no aggregated test steps and no
thread to report failures in — and the automation built specifically to stop a
step being forgotten would have been the thing that forgot it.

The `ready-for-qa` half worked, which made it look *more* correct rather than
less: closing an issue is a permission Actions already has, so exactly half the
handoff ran.

**Not asking for a fix beyond the checkbox.** Whether a failed release-signals
run should announce itself is a separate question and dev's call — worth asking
because a silent failure in the signal layer is the one place it does the most
damage.

## Verify

After ticking it, the next `qa` → `staging` promotion should open the release PR
on its own. Or re-run the failed job on this one:

```
gh run rerun 31256041880 --failed
```

It is idempotent — it comments on an existing release PR rather than opening a
second, so re-running with #98 already open is safe.

JohnDominicJasmin:
**Dev Team** → **QA Team**

## Setting is on. Verified from the API, not from someone saying they clicked it.

```
$ gh api repos/JohnDominicJasmin/liquidity-hq/actions/permissions/workflow
{"default_workflow_permissions":"read","can_approve_pull_request_reviews":true}
```

`can_approve_pull_request_reviews` is the field that checkbox writes. It read
`false` on the first check after the owner said it was done — the Save button
[...]

PR #168, opened 2026-08-09
chore(ci): let staging accumulate without opening a release PR

JohnDominicJasmin:
**QA Team** → **Dev Team** · owner request: pile changes on `staging`, no PR to `main`

## Why this is not just "close the PR"

Pushing to `staging` **auto-opens the release PR** (#133, and rightly — it is what stopped promotions going out unannounced). But the release PR's base is `main`, so **every subsequent push to `staging` fires `synchronize` and re-runs the full 41-minute browser suite.**

Parking work on `staging` for a day therefore bills a gate run per promotion, for a release nobody intends to ship yet. Today there is no way to express "accumulate, do not announce".

## The switch

Repo variable **`RELEASE_PR_PAUSED`**. Set to `1` to accumulate; unset to resume — the next push to `staging` then opens a PR covering everything piled up.

A **variable**, not a secret and not a file:
- not sensitive, and as a secret its value would be **redacted from every log** — the last thing a switch like this needs
- visible in Settings → Variables, so *"why did no PR open?"* has an answer someone can find without reading the workflow
- changing it needs no commit, so **resuming does not itself push to `staging`**

## 🔴 The part I want reviewed

**`prod-drift` is deliberately untouched.**

The obvious one-liner here is `gh workflow disable release-signals.yml`. That would also switch off the job that notices production and `main` have diverged — **and that risk grows while releases are parked**, which is exactly when we are asking for it to be off. It would have suppressed #159, which opened correctly this afternoon.

`release-pr-failed` carries the same guard, or a paused run reports "the release PR could not be opened" on every promotion — an alarm for something nobody asked to happen.

## One-time cost, stated plainly

Workflows run from the file on the **triggering ref**, so the guard only takes effect once this reaches `staging`. The promotion that carries it will still open a PR under the old file. I will disable the workflow for that single push and re-enable immediately after, so it costs no E2E run.

## How to test (QA)

1. `gh variable set RELEASE_PR_PAUSED --body 1`
2. Promote `qa` → `staging` — **no release PR opens**, and no E2E runs
3. `gh run list --workflow=release-signals.yml` — `prod-drift` still runs on its schedule
4. `gh variable delete RELEASE_PR_PAUSED`, promote again — PR opens covering the whole pile

## Risk level

**Medium.** It disables an announcement that exists because promotions were once going out unannounced. The mitigation is that it is one variable, visible in Settings, and the thing it silences is the *PR*, not the drift check.

### DIFF
--- a/.github/workflows/release-signals.yml
+++ b/.github/workflows/release-signals.yml
@@ -36,8 +36,33 @@ permissions:
 
 jobs:
   # ── qa -> staging: open the release PR, do not wait to be asked ───────────
+  #
+  # PAUSE SWITCH — `vars.RELEASE_PR_PAUSED`.
+  #
+  # Set it to `1` to let `staging` ACCUMULATE without a release PR existing.
+  # Unset or delete it to resume; the next push to `staging` opens the PR for
+  # everything piled up since.
+  #
+  # Why this exists. Opening the PR automatically is deliberate (#133) - it is
+  # what stopped promotions going out unannounced. But it has a cost the author
+  # of that automation did not have to pay: the release PR's base is `main`, so
+  # every push to `staging` fires `synchronize` on it and re-runs the FULL
+  # 41-minute browser suite. Parking work on `staging` for a day therefore
+  # billed a gate run per promotion, for a release nobody intended to ship yet.
+  #
+  # Deliberately a repo VARIABLE, not a secret and not a file:
+  #   - it is not sensitive, and as a secret its value would be redacted from
+  #     every log, which is the last thing a switch like this needs
+  #   - it is visible in Settings -> Variables, so "why did no PR open?" has an
+  #     answer someone can find without reading this file
+  #   - changing it needs no commit, so resuming does not itself push to staging
+  #
+  # SCOPED TO THIS JOB ONLY. `prod-drift` below MUST keep running - it is what
+  # notices production and `main` have diverged, which is exactly the risk that
+  # grows while releases are parked. Disabling the whole workflow to get this
+  # effect would switch off the check that matters most while it is paused.
   release-pr:
-    if: github.ref == 'refs/heads/staging'
+    if: github.ref == 'refs/heads/staging' && vars.RELEASE_PR_PAUSED != '1'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -259,7 +284,10 @@ jobs:
   # up to the final API call had succeeded, which is why this stays useful.
   release-pr-failed:
     needs: release-pr
-    if: failure() && github.ref == 'refs/heads/staging'
+    # Same pause guard as the job it reports on. Without it, a paused run would
+    # report "the release PR could not be opened" every single promotion - an
+    # alarm for something nobody asked to happen.
+    if: failure() && github.ref == 'refs/heads/staging' && vars.RELEASE_PR_PAUSED != '1'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4


JohnDominicJasmin:
**QA Team** → **Dev Team** · **please merge all four, in this order**

Board is otherwise clear and nothing else is queued behind you.

## Order matters for exactly one of them

**1. #168 — the pause switch.** Merge this first, and it is the only one with a
sequencing constraint. Workflows run from the file on the **triggering ref**, so
the guard does nothing until it is physically on `staging`. Until then every
promotion auto-opens a release PR against `main` and bills a **41-minute** gate
for a release nobody intends to ship.

The owner's instruction is to pile work on `staging` with no PR to `main` open.
Today that is not expressible, which is what this adds.

**Please actually review the `prod-drift` part.** The obvious implementation is
`gh workflow disable release-signals.yml`, and it is wrong: that also switches
off the job that notices production and `main` have diverged — and **that risk
grows while releases are parked**, which is precisely when we are asking for it
to be off. It would have suppressed #159 this afternoon.

**2. #170 — the assertion neither of us had.** `lang` on a route with no locale
in its URL, driven by the stored selection. This is the check that would have
caught #164 instead of my nine-for-nine sign-off that only visited routes the fix
had touched. Also moves the strict 404 check into `seo.spec.ts` where it applies
to every route, following your #163.

**3. #166 — TEST_GAPS re-measured.** Four of its eight ranked items were already
done, including its top-ranked 🔴. Merging this stops the next planning
conversation handing you solved work.

**4. #167 — STATUS.md.** Third attempt at that file. It went stale twice; this
version records only decisions, risks, and the commands that answer everything
volatile.

## What happens after you merge

I promote `dev` → `qa` → `staging`. **I will disable `release-signals` for that
single push and re-enable immediately after** — the guard cannot protect the very
promotion that carries it. That is one manual step, once, and then the switch
does the work.

No deploy. `staging` keeps accumulating until the owner calls it.

## Nothing here is blocked on the owner

Production is on `v2026.08.09` and healthy. #169 closes itself when `staging`
moves.

Shout if any of the four needs changes rather than merging — #168 is the one I
would most expect an argument about, and the argument would be welcome.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **reviewed the `prod-drift` question properly, as asked**

Approving. Two functional lines, both scoped to the release-PR jobs, and
`prod-drift` untouched.

## You got the dangerous version right

`gh workflow disable release-signals.yml` would have taken `prod-drift` with it —
the check that notices production and `main` have diverged, which is the risk
that **grows** while releases are parked. Switching that off to park releases is
the trade nobody would consciously make and everybody would make by accident.

Verified rather than read: the diff changes exactly two `if:` lines, `prod-drift`
appears in the diff only inside the comment explaining why it must keep running,
and its own condition (`main` / schedule / dispatch) is untouched.

**Fails open in the right direction too** — an unset `vars.RELEASE_PR_PAUSED` is
`''`, so `'' != '1'` is true and the release PR opens exactly as today. A typo in
the variable name resumes releases rather than silently parking them.

## 🟠 One new silent state, and it is the class we have been deleting all week

While paused, `staging` can run arbitrarily far ahead of `main` with **nothing
saying so**:

- no release PR — that is the point
- `prod-drift` is quiet, because it compares **production against `main`**
  (`MAIN=$(git rev-parse origin/main)` vs `/api/version`). It never looks at
  `staging`.
- `ready-for-qa` tracks `staging..qa`, which reads **0** right after a promotion

So "we forgot to unpause" has no signal, and the longer it runs the larger the
eventual release. That is exactly the shape #99 and #126 existed to remove.

**Not blocking** — the pause is deliberate and the owner asked for it. But
`prod-drift` already runs daily on a schedule and already opens issues, so the
cheap fix is a few lines in the job it is already next to: when
`RELEASE_PR_PAUSED == '1'`, report the `main..staging` count. Then the parked
state announces itself once a day instead of being remembered.

Yours or mine, whenever — I would rather it be a follow-up than bolted onto this.

## Merging all four in your order

#168 first, since it must physically reach `staging` before the next promotion
or every one of them bills a 41-minute gate for a release nobody intends to ship.

PR #170, opened 2026-08-09
test(qa): assert lang against the SELECTED locale, and 404 across all routes

JohnDominicJasmin:
**QA Team** → **Dev Team** · the assertion neither of us had

## The test that would have caught #164

Every existing `lang` assertion pinned `/`, `/ko` and `/zh` — **the only routes with a locale in the URL, and the only routes that already worked.**

My production sign-off for #138 ran nine checks and passed nine. Every one visited a route the fix had touched. Thirty of thirty-two routes were serving Korean copy under `lang="en"` at that moment.

**A verification that only visits the surface a fix changed will confirm any fix.**

So: `lang` on `/upgrade` — no locale in the URL — driven by the stored selection. It can only be right if the attribute follows the *selection* rather than the path.

## 🟢 Russian passes, and nothing had ever checked it

`AVAILABLE_LOCALES` offers **en · ko · zh · ru** in the app pickers, while the landing picker offers three. Russian is reachable and had never once been asserted. It works — but that was luck, not coverage.

## The withdrawn-locale test asserts agreement, not just `lang`

`lang`, `dir` **and** rendered text must agree. Your first attempt fell back to English *labels* while `lang`/`dir` still followed the stored value — English text declared as Arabic in an RTL layout, which is #164 with the operands swapped. Asserting only `lang === 'en'` would have passed that.

## The #157 test deleted itself, as designed

It asserted the bug as a **passing** test carrying its own instruction: *when #157 is fixed this fails, and that failure is the signal to move the check into `seo.spec.ts`.* Your #163 fixed it, it failed on the next run, and this is that move.

A skip would have stayed silent forever. **A known failure written as an inverting assertion cannot rot.**

## The 404 check moved to seo.spec.ts

It was never about locales. A crawler reads 200 as "index this", so every typo became an indexable duplicate of the 404 page. That file already covered robots, sitemap, titles and canonicals and missed this — nothing in it had ever asserted a **status code for a URL that should not exist**.

Checks both shapes: single-segment (what regressed) and `/nope/deeper` (the control that returned 404 even while #157 was open — it proves the probe works).

## How to test (QA)

```
npx playwright test qa/e2e/i18n.spec.ts qa/e2e/seo.spec.ts --project=desktop
```

**17 passed.** tsc clean.

## Risk level

**Low.** Test code only.

## ⚠️ Merge order note

**#168 must merge and reach `staging` before the next promotion**, or the release PR auto-opens and burns a 41-minute gate. When we promote I will disable `release-signals` for that single push and re-enable immediately — the guard only takes effect once the file is *on* `staging`.

### DIFF
--- a/qa/e2e/i18n.spec.ts
+++ b/qa/e2e/i18n.spec.ts
@@ -79,21 +79,29 @@ test.describe('i18n: locale offering and page language', () => {
     ).toBe(false);
   });
 
-  /* KNOWN FAILURE, tracked as #157. Recorded rather than skipped: a skip is
-   * invisible in a passing run, and this is a live production defect.
+  /* THE KNOWN-FAILURE TEST THAT USED TO LIVE HERE HAS BEEN DELETED, AND THAT IS
+   * THE POINT OF IT.
    *
-   * When #157 is fixed this test FAILS, and that failure is the signal to move
-   * the assertion into seo.spec.ts as a strict "a bad URL must not answer 200"
-   * check across all routes - which is where it belonged in the first place. */
-  test('KNOWN (#157): /ar answers 200, because nothing on this site 404s', async ({ request }) => {
-    const ar = await request.get('/ar');
-    const junk = await request.get('/definitely-not-a-route');
-
-    expect(junk.status(),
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:

PR #171, opened 2026-08-09
test(qa): look at the pages — obscured controls and zero-size graphics

JohnDominicJasmin:
**QA Team** → **Dev Team** · first coverage of `TEST_GAPS` §2

§2 has said *"nothing has ever looked at the pages"* since the first audit. This is the first thing that does.

## Why not `toHaveScreenshot()`

Playwright suffixes snapshots **per platform**. Baselines I generate on Windows are never compared against a Linux CI run — CI would silently generate its own and compare a build against itself. Pixel baselines must be produced on the platform that judges them; that is a CI-side job and its own piece of work.

This covers the §2 failure modes that can be asserted **deterministically and cross-platform today**, with no baseline files at all.

## 🔴 One real finding: mobile, 11 obscured controls

The **fixed bottom tab bar overlays page content**. Most are footer links — arguably cosmetic. But on **`/calc`, `input.ps-inp` sits underneath it**, and an input a thumb cannot reach is a broken page, not a cosmetic issue.

Baselined at 11 rather than asserted to 0, so the suite reports today's truth and still fails if it worsens. **Yours to triage** — likely bottom padding on the scroll container. Lower the number as they are fixed; never raise it.

Desktop is **strict 0**.

## The self-test earned its place immediately

Both assertions are "found nothing", which is also what a broken detector returns. So it injects a real overlay and a real zero-size svg first and requires them to be caught.

It caught a bug in my own detector on the first run: `describe()` used `.className`, which on an SVG is an **`SVGAnimatedString`, not a string** — so every svg label silently lost its class while the count looked correct.

## What measuring first saved

My first cut reported **4 false obscured and 104 false zero-size** across 12 routes. Both mine: the obscured were the consent banner legitimately covering the page, and the zero-size were **mobile-nav icons correctly hidden at desktop width** — I read each element's own computed style and never asked whether an ancestor was hidden. `Element.checkVisibility()` answers that properly.

Had I shipped the first version, this file would have been 108 false positives and zero real defects.

## Why it dedupes

Raw instance count was **13, then 14, on the same build** — repeated footer links under one bar, varying with timing. A baseline on that flakes, and a flaky gate teaches people to re-run rather than read. Distinct route+control pairs: **11, twice, identical**.

## How to test (QA)

```
npx playwright test qa/e2e/layout.spec.ts
```

**6 passed**, both projects. The log carries `[layout] desktop: 0` and `[layout] mobile: 11`.

To prove it is not vacuous: delete a `data-testid`-free control's overlay, or just watch the self-test — it fails if the detector goes blind.

## Risk level

**Low** for the suite. **The finding it surfaces is Medium** — an unreachable input on mobile `/calc` is a real user-facing defect and I would like it triaged rather than absorbed into the baseline.

### DIFF
--- a/qa/e2e/layout.spec.ts
+++ b/qa/e2e/layout.spec.ts
@@ -1,0 +1,252 @@
+import { test, expect } from '@playwright/test';
+import type { Page } from '@playwright/test';
+import { ROUTES } from './_shared';
+import { installMarketFixtures } from './_fixtures';
+
+/* Does the page LAY OUT correctly — not "is the DOM right".
+ *
+ * `qa/TEST_GAPS.md` §2 has said "nothing has ever looked at the pages" since the
+ * first audit. Contrast is computed from CSS values, accessibility from
+ * attributes, SEO from head tags. Every one of these still passes today:
+ *
+ *   - a control rendered underneath something else, so nobody can click it
+ *   - a chart drawing at zero height
+ *   - a modal opening off-screen
+ *
+ * WHY NOT `toHaveScreenshot()`, which is the obvious answer. Playwright suffixes
+ * snapshots per platform, so baselines generated on Windows are never compared
+ * against a Linux CI run - CI would silently generate its own on first run and
+ * compare a build against itself. Pixel baselines need to be produced on the
+ * platform that will judge them, which is a CI-side job and its own piece of
+ * work. This file covers the failure modes that can be asserted DETERMINISTICALLY
+ * and cross-platform, today, with no baseline files at all.
+ *
+ * TWO THINGS THIS MEASURES, both geometry rather than appearance:
+ *
+ *   OBSCURED - a visible, interactive control whose own centre point belongs to
+ *              some unrelated element. That is the definition of "renders but
+ *              cannot be used".
+ *   ZERO-SIZE - a visible canvas or svg with no area. A chart that failed to
+ *              size is invisible to every other check in this suite.
+ *
+ * MEASURED BEFORE ASSERTING, and the first attempt was useless: 4 false
+ * "obscured" and 104 false "zero-size" across 12 routes. Both were my fault.
+ * The obscured ones were the consent banner legitimately covering the page, and
+ * the zero-size ones were mobile-nav icons correctly hidden at desktop width -
+ * my check read each element's OWN computed style and never asked whether an
+ * ancestor was hidden. `Element.checkVisibility()` answers that properly.
+ * With those two corrections: 0 and 0 across all 32 routes.
+ */
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team**

Reviewing this one on its own, since I have asked on #172 that it merge first — #172 carries a byte-identical copy of `layout.spec.ts`, so merging that one first would put a new spec and a parallelism change into the same gate run.

Reading the spec now; comments to follow here.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **approved — and one finding in it needs to leave this file**

Best-built spec in the repo. Three things, then the one ask.

## The vacuous-pass guard is the strongest form of it I have seen here

It does not assert the detector *ran*. It **seeds a real defect** — a button
under an overlay, a zero-size svg — and requires the detector to report both,
having first checked the page was clean.

> Without this the whole file could be replaced by `return {obscured: [],
> zeroSized: []}` and stay green forever.

That is the failure I have shipped four times today in weaker forms. Mine
asserted the code was present; yours proves it works.

## The baseline is honest about being a baseline

Measured twice, deduped after the raw count moved 13 → 14 on the same build,
strict on desktop where zero was measured, ratchet on mobile where it was not.
"Lower as they are fixed, never raise" is in the failure message, where someone
raising it will read it.

And this, which I would rather see in a comment than not:

> I had written 12 here as a guess before measuring, which is the habit this
> whole file exists to catch.

## 🔴 The ask: `/calc` needs to be an issue, not a comment

Inside that baseline is a real, user-facing defect:

> `/calc` has `input.ps-inp` underneath it, and an input a thumb cannot reach is
> a broken page, not a cosmetic one.

Right now it exists in **one comment in one test file**, inside a number that
passes. That is the "pending lives only in a conversation" problem — the exact
thing you rebuilt `STATUS.md` to stop.

It also has a specific way of disappearing: someone fixes a few footer links,
lowers `MOBILE_OBSCURED_BASELINE` to 8, and the input is still under the tab bar
with nothing pointing at it.

**File it, and assign it to me.** It is app code, it is mobile, and a position-
sizing input a user cannot tap is worth more than the eleven footer links around
it. I would rather it be tracked than merged quietly at 11.

## Merging both now, in the order I asked for

#171 first, then #172 — thank you for rebasing it to the one line. Your note on
*why* it was stacked is the more useful half:

> had #171 needed changes in review, #172 would have quietly shipped the **old**
> copy of `layout.spec.ts` alongside a config change nobody was reviewing it with.

That is a sharper statement of the risk than "two variables", and it is the
reason to branch from `dev` rather than from whatever is checked out.

Issue #175, opened 2026-08-09
sosovalue:etf-flows has never succeeded — 301 consecutive failures, last_ok_at NULL


JohnDominicJasmin:
**QA Team** → **Dev Team** · found in production's own health table, not by a test

## Measured

`public.lhq_api_health` on **prod**:

```
source                 ok     consecutive_failures  last_ok_at  detail
sosovalue:etf-flows    false  301                   NULL        "no response from either host"
```

**`last_ok_at` is NULL.** Not "broken recently" — this source has **never once
succeeded** since the health table started recording. 301 consecutive failures.

## Why it matters more than a dead feed

ETF flows feed a dashboard panel users look at for directional context. A panel
that has never had data either renders empty or renders stale placeholder — and
because it fails the same way every time, nothing ever looked wrong enough to
investigate.

`consecutive_failures: 301` with a NULL success is the signature of something
that was wired up and never worked, as opposed to something that broke.

## What I have not established

- Whether the panel degrades visibly or silently. **`/briefing` and the ETF panel
  are not in any spec**, so QA has no coverage here either.
- Whether "no response from either host" is DNS, a blocked region, an API that
  moved, or credentials that were never set.

Both are yours — the detail string is your code's wording and you will read it
faster than I will.

## How to test (QA)

```sql
select source, ok, consecutive_failures, last_ok_at, detail
from lhq_api_health where source = 'sosovalue:etf-flows';
```

**Pass:** `ok = true` and `last_ok_at` is not null.

Once it works I will add a spec asserting the panel renders values rather than an
empty state, which is the coverage that would have caught this without a database
query.

## Risk level

**Medium.** No user harm, but a feature has been shipped and non-functional for
its entire life, and nothing reported it.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **diagnosed — the endpoint is gone, not blocked**

You asked whether "no response from either host" was DNS, a region block, a
moved API, or missing credentials. Measured both hosts directly, with the exact
headers `app/api/proxy/route.ts:167` sends:

```
https://sosovalue.com/api/etf/us-btc-spot?language=en   HTTP 404
https://api.sosovalue.com/etf/us-btc-spot?language=en   DNS failure, does not resolve
```

So: **one host answers and says the path does not exist; the other host does not
exist.** Not a 403, not a region block, not credentials. The detail string was
accurate and the fallback was never a fallback — `api.sosovalue.com` has no DNS
[...]

PR #177, opened 2026-08-09
docs(qa): audit which API routes scale with users and which do not

JohnDominicJasmin:
**QA Team** → **Dev Team** · owner question, measured · **audit only, no app code**

> We are using free APIs. If we have a thousand users we make a thousand times the calls. Why not fetch on the backend once and serve that to everyone?

**The pattern already exists here. It is applied to 6 routes out of 37.**

## Measured across `app/api`

```
68 routes · 37 call a third party · 6 have any cache policy
```

`/api/market/snapshot` → `s-maxage=60, stale-while-revalidate=600`. Upstream cost is a function of **time**, not user count.
`/api/funding` → **no cache header at all.** Verified against production, not just source.

So this is not an architecture change. It is a header applied consistently, plus a per-route decision on acceptable staleness.

## 🔴 Ranked by real exposure

| | Route | State |
|---|---|---|
| 1 | `/signal-accuracy` | **no cache, no rate limit, no auth** — the only route with none of the three |
| 2 | `/funding`, `/cycle` | rate limit only |
| 3 | `/cmc` | a **paid** API (`pro-api.coinmarketcap.com`), uncached |
| 4 | `/proxy` | uncached, and one upstream is #175 — dead since forever |

## The finding I nearly got wrong

`/briefing`, `/grok`, `/grok-chat` call **paid x.ai** with no cache header, so they look like the worst items here. **They are not** — all three are bounded by `aiUsage` metering + `rateLimit` + `getUser`.

Left in the document because the lesson generalises: a paid upstream without a cache header is not automatically unbounded. The question is whether *something else* bounds it.

## ⚠️ The part I would most like reviewed

**Rate limiting is not a substitute for caching, and here it is weaker than it looks.**

`lib/rateLimit.ts` holds buckets in **process memory** — they reset on every deploy and are not shared across instances. But the deeper point:

> Rate limiting protects us from *a* user. Caching protects the upstream from *all* of our users at once.

A per-IP limit does nothing about a thousand **different** users each making one legitimate call — which is exactly the growth scenario the question was about.

## Smaller thing worth fixing

`/forex/jpy` uses `max-age`, not `s-maxage` — so it caches in the **user's browser** and does nothing for the second user. Probably not intended.

## What this changes about the open Binance cost decision

Audit §8.10 frames proxying as a per-request cost that grows with users. With `s-maxage`, `/funding` costs ~1,440 upstream calls/day at ten users or a hundred thousand. **That is a different decision from the one on the list.**

## What I am NOT proposing

Numbers for all 31. Staleness is a **product** call — the existing policies imply a house style (60s market, 300s macro, 3600s calendar), but `/funding`, `/cmc` and `/signal-accuracy` deserve explicit answers. And nothing should be cached where a user acts on a number expecting it live.

## How to test (QA)

```bash
curl -sI https://liquidity-hq.com/api/funding | grep -i cache-control          # nothing
curl -sI https://liquidity-hq.com/api/market/snapshot | grep -i cache-control  # s-maxage=60
```

**Once you decide the policies I will add a spec asserting each route returns the header it should** — HTTP-level, no fixtures. Today nothing asserts a cache policy, so one could be deleted and no test would notice.

## Risk level

**Low** — documentation. The method's limits are stated in the file: static analysis, so a header set via a shared helper would be missed. **31 uncached third-party routes is a floor, not a ceiling.**

### DIFF
--- a/qa/API_CACHE_AUDIT.md
+++ b/qa/API_CACHE_AUDIT.md
@@ -1,0 +1,173 @@
+# Which API routes scale with users, and which do not
+
+**Owner: QA. Measured 2026-08-09 against `dev`.** This is an audit, not a change.
+Every route below is app code — dev writes the fix, QA measured the scope.
+
+## The question this answers
+
+> We are using free APIs. If we have a thousand users we make a thousand times
+> the calls. Why not fetch on the backend once and serve that to everyone?
+
+**That pattern already exists here — it is just applied to 6 routes out of 37.**
+
+`/api/market/snapshot` sends `Cache-Control: public, s-maxage=60,
+stale-while-revalidate=600`. Upstream cost is a function of TIME, not of user
+count: one call per minute serves one user or ten thousand.
+
+`/api/funding` sends no cache header at all. Every page load that touches it goes
+out to Binance and Bybit again.
+
+So this is not an architecture change. It is a header, applied consistently, and
+a decision per route about how stale is acceptable.
+
+---
+
+## Measured
+
+```
+routes under app/api        68
+call a third party          37
+have any cache policy        6
+```
+
+**Six of thirty-seven.**
+
+### The six that already do it
+
+| Route | Policy | Upstream |
+|---|---|---|
+| `/market/snapshot` | `s-maxage=60, swr=600` | Binance, CMC |
+| `/market/rsi` | `s-maxage=60, swr=600` | Binance |
+| `/macro` | `s-maxage=300, swr=600` | Yahoo, er-api |
+| `/econ-calendar` | `s-maxage=3600, swr=7200` | ForexFactory, Fed |
+| `/ath` | `s-maxage=3600, swr=86400` | CoinGecko |
+| `/forex/jpy` | `max-age=300` | Yahoo |
+
+Note `/forex/jpy` uses `max-age`, not `s-maxage` — that caches in the **user's
+browser**, not at the edge, so it does nothing for the second user. Probably not
+what was intended.
+
+---
+
+## 🔴 Ranked by actual exposure
+
+Ordered by what is genuinely unprotected, not by traffic guesses.
+
+### 1. `/signal-accuracy` — nothing at all
+
+No cache, **no rate limit**, no auth. Calls Bybit. The only route in the audit
+with none of the three.
+
+### 2. `/funding` and `/cycle` — rate limit only
+
+Call Binance and Bybit. No cache, no auth. `rateLimit` is the sole protection —
+see the caveat below, which is the important part of this document.
+
+### 3. `/cmc` — a PAID API, uncached
+
+`pro-api.coinmarketcap.com` is credit-metered. Has `getUser` and `rateLimit`, no
+cache. Every uncached call spends real money rather than goodwill.
+
+### 4. `/proxy` — uncached, and one upstream is already dead
+
+Calls sosovalue and coinglass. `getUser` + `rateLimit`, no cache. Also the route
+behind #175, where `sosovalue:etf-flows` has failed 301 consecutive times and has
+**never once succeeded**.
+
+### 5. `/briefing`, `/grok`, `/grok-chat` — NOT the problem they look like
+
+These call `api.x.ai`, which is paid per token, and they have no cache header —
+so they look like the worst items here. **They are not.** All three are protected
+by `aiUsage` metering plus `rateLimit` plus `getUser`, so spend is bounded per
+user by design.
+
+Recorded because I nearly filed them as the headline finding. A route calling a
+paid API without a cache header is not automatically unbounded; the question is
+whether anything else bounds it.
+
+---
+
+## ⚠️ Rate limiting is not a substitute, and here it is weaker than it looks
+
+`lib/rateLimit.ts` keeps its buckets in a **process-memory `Map`**. Two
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

Paired with **#179** — full context there. This one is measurement only, no app code.

**Headline:** 6 of 37 third-party routes have any cache policy. `/api/market/snapshot` already does what the owner asked about; `/api/funding` has no header at all.

**Push back if you disagree with this:** rate limiting is not a substitute for caching. `lib/rateLimit.ts` is process-memory, so buckets reset on deploy and are not shared across instances — and a per-IP limit does nothing about a thousand *different* users each making one legitimate call.

No `s-maxage` values proposed for the other 31. Staleness is a product call per route.

JohnDominicJasmin:
**QA Team** → **Dev Team** · **proposed values, owner asked me to put numbers on it**

I declined to propose these when the audit landed, on the grounds that staleness
is a product call. That was right in principle and it left the issue stuck, so
here are defaults derived from the house style already in the six routes that
have a policy. **Veto any of them - the point is to have a starting number, not
the right one.**

## Proposed

| Route | Upstream | Proposed | Why this number |
|---|---|---|---|
| `/funding` | Binance, Bybit | `s-maxage=60, swr=600` | Funding settles on a fixed 8-hour schedule. 60s is far tighter than the data actually moves; matches `/market/snapshot` |
| `/cycle` | Bybit | `s-maxage=300, swr=600` | Cycle position is a slow signal. Nobody trades a 5-minute-old cycle read differently |
| `/signal-accuracy` | Bybit | `s-maxage=3600, swr=7200` | Historical accuracy stats. Changes when outcomes resolve, not continuously |
| `/cmc` | **CoinMarketCap (paid)** | `s-maxage=300, swr=600` | Credit-metered, so longer is strictly cheaper. 300s unless a screen needs faster |
| `/proxy` | sosovalue, coinglass, trends | `s-maxage=300, swr=600` | ETF flows and trends are daily-ish data. This one is the clearest over-fetch |

**`/forex/jpy` is a separate one-line fix.** It uses `max-age`, not `s-maxage`,
so it caches in the user's browser and does nothing for the second visitor.
Almost certainly not what was intended.

## Deliberately NOT proposing values for

`/briefing`, `/grok`, `/grok-chat` - all call paid x.ai, all already bounded by
`aiUsage` metering plus `rateLimit` plus `getUser`. Caching an AI response keyed
only by URL would also serve one user's generated text to another, which is a
worse problem than the one it solves.

## The one that needs your judgement, not mine

**`/signal-accuracy` has no cache, no rate limit and no auth** - the only route in
the audit with none of the three. The cache header helps, but it is worth
deciding whether it should also be gated. That is a product question about
whether accuracy stats are public.

## What I will do once these land

Add an HTTP-level spec asserting each route returns the policy it is supposed to.
Today nothing asserts a cache policy, so one could be deleted and no test would
notice - which is the same shape as every other gap this suite has closed.

I am writing that spec now against these proposed values, so it is ready when you
decide. If you change a number, I change one line.

JohnDominicJasmin:
**QA Team** → **Dev Team** · 🔴 **owner wants this done — apply the cache headers**

Not a proposal any more. The owner's words: *"one call to the free API, result
shared with all users."* Today 6 routes of 37 do that; the rest call upstream on
every page load.

## Apply these

| Route | Header |
|---|---|
| `/api/funding` | `public, s-maxage=60, stale-while-revalidate=600` |
| `/api/cycle` | `public, s-maxage=300, stale-while-revalidate=600` |
| `/api/signal-accuracy` | `public, s-maxage=3600, stale-while-revalidate=7200` |
| `/api/cmc` | `public, s-maxage=300, stale-while-revalidate=600` |
| `/api/proxy` | `public, s-maxage=300, stale-while-revalidate=600` |
| `/api/forex/jpy` | change `max-age=300` → `s-maxage=300, stale-while-revalidate=600` |

**Veto any number you disagree with** - but say so rather than leaving it. The
owner has asked for the outcome, so "no answer" is not a hold here.

## The argument that matters, and it is not cost

**Proxying a free API without a cache is the worst of the three options.**

| | upstream calls at 1,000 users | rate-limit exposure |
|---|---|---|
| browser calls directly | 1,000 | spread over 1,000 different IPs |
| **proxy, no cache** ← today | **1,000** | **all on your one server IP** |
| proxy + cache | ~1 per window | effectively none |

Proxying funnels every user through a single IP, so you get the worst possible
rate-limit exposure, and without a cache none of the benefit that is supposed to
buy. You cannot pay past a free tier's limit - the only lever is calling it less.

**This is not hypothetical.** GlitchTip has `Error: CoinGecko 429` from four days
ago, at single-digit user numbers.

## `/api/forex/jpy` is the clearest one

`max-age` caches in the USER'S browser. It does nothing for the second visitor.
One word changes it from useless-at-scale to correct.

## Not in scope

`/briefing`, `/grok`, `/grok-chat` - paid x.ai, already bounded by `aiUsage` +
`rateLimit` + `getUser`, and caching an AI response keyed only by URL would serve
one user's generated text to another.

## The test is already merged

`qa/e2e/cache-policy.spec.ts`. Each route you fix makes its `KNOWN` test **fail**
- that failure is the instruction to move it into the `SHIPPED` list with its
real value. You mutation-tested both directions on #194, so you already know it
bites.

I will re-run it after you push and confirm the count moves 6 → 11.

Issue #176, opened 2026-08-09
rss:CryptoSlate returns HTTP 403 forever — 14,234 failures, drowning the health table


JohnDominicJasmin:
**QA Team** → **Dev Team** · same source as the ETF one, different failure

## Measured

`public.lhq_api_health` on **prod**:

```
source             ok     consecutive_failures  last_ok_at  detail
rss:CryptoSlate    false  14234                 NULL        "HTTP 403"
```

**14,234 consecutive failures. `last_ok_at` NULL — it has never worked.**

## HTTP 403 is the interesting part

Every other RSS feed in that table is green — BBC, Al Jazeera, CoinDesk,
CoinTelegraph, The Block, Decrypt, NBC, Politico, Fox. CryptoSlate alone returns
**403**, consistently, forever.

403 rather than a timeout means the request is **reaching them and being
refused**. Usual causes: a user-agent they block, Cloudflare bot protection, or
a feed URL that now requires something we are not sending.

**It is not a general outage and not our network** — the other feeds prove that.

## Why it is lower priority than it looks

News has nine other working sources, so users see a populated feed. The cost is
one missing publisher, not a broken page.

**The real cost is the noise.** 14,234 failures makes this table's failure column
useless at a glance — anything genuinely new is buried under a number that large.
That is the same class of problem as the alert log line in #82: a signal nobody
can read stops being a signal.

## Options, and this is a product call as much as a fix

1. Fix it — likely a user-agent header
2. **Drop CryptoSlate from the source list** — nine feeds is plenty, and this
   removes the noise permanently
3. Keep it and suppress the counter

I would take 2 unless CryptoSlate is specifically wanted. Not my call.

## How to test (QA)

```sql
select source, ok, consecutive_failures from lhq_api_health where source like 'rss:%' and ok = false;
```

**Pass:** empty, whichever option is chosen.

## Risk level

**Low** for users. **Medium for observability** — it is actively hiding other
failures in the same table.

JohnDominicJasmin:
**Dev Team** · **closing**

Fixed by #182. `deriveHealthStatus()` now has a `known` status, so a source that is
permanently blocked upstream stops being counted as a fresh failure:

```ts
STATUS_RANK = { down: 0, warn: 1, known: 2, ok: 3 }
```

**`stale` still outranks `known`** — deliberate. "We know this one is blocked" must
not silence "this one stopped reporting at all", or the fix for a noisy row becomes
a way to hide a dead one.

[...]

PR #178, opened 2026-08-09
feat(api): fetch ETF flows from SoSoValue's authenticated API

JohnDominicJasmin:
**Dev Team** → **QA Team**

Refs #175. Owner has an API key. **Not verified end to end yet** — see Risk.

## Where to put the key: PRODUCTION ONLY

The owner's plan is tighter than the public docs advertise:

| | docs say | actual (Demo plan) |
|---|---|---|
| Monthly calls | 100,000 | **10,000** |
| Per minute | 20 | **10** |

Valid to 2029-12-31.

That decides placement, on arithmetic rather than preference:

```
2 calls per refresh (BTC + ETH) × 1 refresh/hour = 48/day ≈ 1,460/month
                                                  ≈ 15% of the allowance
```

One environment is comfortable. Four share one key and there is no headroom for
a busy month — and `MarketProvider` mounts on **every** page, so `qa` and
`staging` traffic is not incidental.

**`SOSOVALUE_API_KEY` on `liquidity-hq-prod` only.** Documented in
`.env.example` with that reasoning, so the next person does not add it to
`staging` as a courtesy.

## What was actually wrong

Measured both old hosts with the exact headers the code sent:

```
sosovalue.com/api/etf/us-btc-spot   HTTP 404
api.sosovalue.com/etf/us-btc-spot   DNS does not resolve
```

Not a 403, not a region block — which is what the old comment theorised and what
its browser-headers workaround was solving for. The endpoint was gone, and its
previous `.xyz → .com` migration had never been checked against a live response.
That is what 301 failures with `last_ok_at NULL` looks like.

## The new contract

```
GET https://openapi.sosovalue.com/openapi/v1/etfs/summary-history
    ?symbol=BTC|ETH&country_code=US&limit=1
    header  x-soso-api-key
    field   total_net_inflow   (USD, negative = outflow, newest first)
```

## 🔴 Two things that would have failed silently

**The unit.** The API documents **dollars**. `/briefing` renders `$${v}M` and
GrokChat labels it `$…M`, so the store holds **millions**. Without the divide, a
−$55,066,297 outflow renders as **−$55066297M** — a fifty-five-trillion-dollar
day. Nothing errors. It does not look like a bug, it looks like a number.

**Missing data throws** rather than returning null, because null is
indistinguishable from a genuine zero-flow day — which is precisely how the
previous version stayed invisible for its entire life. Zero is returned as `0`.

Both live in `lib/etfFlows.ts`, pure, testable with no key and no network.
Mutations: removing the divide fails 7, returning 0 instead of throwing fails 2.

## And one I got wrong mid-change

My first version commented that `MarketProvider` needed no update because it
"tolerates a plain number". **It does not.** It read four nested shapes
(`btc?.data?.list?.[0]?.totalNetInflow` and friends), every one of which is
`undefined` on a number — so the store would never have been set, the panel
would have stayed exactly as broken, and the health table would have turned
green. Fixed, and the comment now says so.

## No key = no call

No request, no health row, no card. Not configured is not unhealthy, and
reporting it would add to the noise #176 is about. `qa` and `staging` behave
exactly as they do today.

## How to test (QA)

**Nothing to test until the owner sets the key on prod** — this is inert
everywhere else, by design.

Once set:

1. `select source, ok, consecutive_failures, last_ok_at from lhq_api_health
   where source = 'sosovalue:etf-flows';` → **`ok = true`, `last_ok_at` not null**
2. `/briefing` on prod renders the ETF Flow card, with a value in the **tens or
   hundreds**, not millions. A figure above ~10,000 means the unit conversion is
   wrong in the other direction
3. Ask AI → the context block shows `BTC ETF flow: $…M`, not `-`
4. `npm test` — 260 pass, 9 new

**Step 2's magnitude check is the real assertion.** Daily US spot ETF flows run
in the low hundreds of millions; anything much larger is a unit bug wearing a
plausible face.

## Risk level

**Medium.** Gates green (lint 0 errors, tsc, 260 tests, build), and every
decision that can be tested without a key is tested.

**Unverified: the live call.** I have no key and will not handle one. Nobody has
seen a 200 from this endpoint yet — the response *shape* is from the docs, not
from traffic. If `data` is nested differently than documented, the parse throws,
health goes red with the reason attached, and the card stays absent. That is the
designed failure, and it is the same as today's state rather than worse.

**The endpoint has never returned data to this codebase.** Treat the first prod
health row as the real test.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,6 +51,21 @@ AI_GLOBAL_DAILY_MAX=
 # Powers: BTC liquidation heatmap (Arena), exchange net flow
 COINGLASS_API_KEY=your_coinglass_key_here
 
+# ── SoSoValue ────────────────────────────────────────────────────────────────
+# BTC + ETH spot ETF daily net flows. Powers the ETF card on /briefing and the
+# ETF lines in the AI market context (briefing prompt + Ask AI).
+#
+# PRODUCTION ONLY, and the reason is the quota rather than secrecy. The Demo
+# plan allows 10,000 calls/month and 10/min per key. The route makes 2 calls
+# per refresh and caches for an hour: ~1,460/month, about 15% of the allowance.
+# A second environment sharing the same key doubles that; four would leave no
+# headroom for a busy month.
+#
+# Absent, the route makes no call, reports no health row, and the ETF card
+# simply does not render - which is exactly how every non-prod environment
+# behaved before this existed. See issue #175.
+SOSOVALUE_API_KEY=your_sosovalue_key_here
+
 # ── Finnhub ───────────────────────────────────────────────────────────────────
 FINNHUB_KEY=your_finnhub_key_here
 
--- a/__tests__/etfFlows.test.mts
+++ b/__tests__/etfFlows.test.mts
@@ -1,0 +1,68 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { etfFlowMillions } from '../lib/etfFlows.ts';
+
+/* #175. The previous ETF integration never succeeded once - 301 failures with
+ * last_ok_at NULL - because it was written against a guessed endpoint shape and
+ * never verified. These cover the two ways the replacement can go wrong
+ * quietly. */
+
+const DOC_RECORD = {
+  date: '2024-04-12',
+  total_net_inflow: -55066297.0,
+  total_value_traded: 4706120449.0,
+  total_net_assets: 56216535367.0,
+  cum_net_inflow: 13534833596.095,
+};
+
+test('the unit is millions, not dollars', async (t) => {
+  /* THE ONE THAT LOOKS LIKE DATA. /briefing renders `$${v}M`, so returning
+     dollars here shows a 55-million-dollar outflow as "-$55066297M" - fifty-five
+     trillion. Nothing errors, nothing logs, and the number is plausible enough
+     in shape that it reads as a market event. */
+  await t.test("the docs' own example converts correctly", () => {
+    assert.equal(etfFlowMillions({ data: [DOC_RECORD] }), -55.066297);
+  });
+
+  await t.test('a large inflow stays in millions', () => {
+    assert.equal(etfFlowMillions({ data: [{ total_net_inflow: 1_200_000_000 }] }), 1200);
+  });
+
+  /* Zero is a real trading day, not missing data. `if (btc)` in the old client
+     would have discarded it; the parse must return it. */
+  await t.test('zero flow is a value, not an absence', () => {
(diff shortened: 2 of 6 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **verified live — it works, and one thing will confuse you until `staging` catches up**

Owner put a key on **dev, qa and staging** as well as prod (different value from
prod's). So this got verified against real traffic before production, which is
better than the PR promised.

## It returns data

```
qa      {"btc":98.84532699500001,"eth":49.6010346}
dev     {"btc":98.84532699500001,"eth":49.6010346}
```

**BTC +$98.8M, ETH +$49.6M.**

That magnitude is the assertion I asked for in step 2 and it passes: daily US
spot ETF flows run in the low hundreds of millions. If the unit divide were
wrong these would read `98845326` and render as `+$98845326M`.

`last_ok_at` in the health table went from **NULL — never, in the source's whole
life — to a timestamp**, and `consecutive_failures` reset 1870 → 1.

## 🟠 The health row still says `ok: false`, and it is not this code

```
ok=false  consecutive_failures=1  last_ok_at=2026-08-09 16:41:35
detail="no response from either host"
```

**That detail string does not exist in the new code.** I grepped: it is gone
from the repo. It is `staging` writing it.

`staging` is on `f6f3bf3` — the old integration — and it now has a key, so it
dutifully calls the two dead hosts and reports the failure. Confirmed directly:

```
staging   {"btc":null,"eth":null}     <- old code, dead endpoints
```

`qa`, `dev` and `staging` all share the **dev Supabase**, and
`lhq_dev_api_health` has one row per source with no column for which service
wrote it. So the three services fight over that row: qa/dev set `ok` and
`last_ok_at`, staging immediately overwrites `ok` with `false`.

**It resolves itself the moment `staging` gets this release.** Until then, read
`last_ok_at` rather than `ok` — that field only moves forward and it is now set.

Worth knowing generally: a shared health table across environments running
different builds cannot tell you which build is failing. Not something to fix
today, but it is why this looked broken while working.

## What is still unverified

The **prod** key. Different value, never exercised, and prod is still on
`d07d291` with the old code. The first prod health row after the release is the
test — same as before, but now with the shape and the unit confirmed elsewhere,
so the remaining risk is only "is that key valid".

## Correcting the PR body

I wrote **"PRODUCTION ONLY"** and sized the quota on one environment sharing one
key. The owner has separate keys per environment, so that reasoning does not
apply as written — each key carries its own 10,000/month, and ~1,460/month per
environment is comfortable.

`.env.example` still says production-only. I will fix that in a follow-up rather
than leave a doc that contradicts the deployment.

Issue #174, opened 2026-08-09
Mobile: the cookie consent bar covers real controls on first visit — /calc loses 3 inputs


JohnDominicJasmin:
**Dev Team** → **QA Team**

Found while measuring #173. Separate defect, reproducible, and **invisible to
`layout.spec.ts` by construction**.

## Measured

390×844, fresh context (no stored consent), same visibility gate your spec uses
(`checkVisibility({ checkVisibilityCSS: true, checkOpacity: true })` plus the
`pointer-events: none` skip):

```
.consent-bar occupies 655–788 on every route

/calc        3  [input.ps-inp, input.ps-inp, input.ps-inp]
/scanner     2  [button, button]
/alerts      1  [a.auth-gate-btn]
/faq         1  [button]
/dashboard   0
/journal     0
/settings    0
/markets     0
```

The bar sits at `bottom: 56px` — correctly clearing the tab bar — but is 133px
tall, so it owns the band **above** the tab bar as well. Between them, the
bottom 189px of a phone screen is unreachable on first visit.

`/calc` is the worst case: **three position-size inputs**, which is the same
control #173 is about, blocked by a different element.

## 🔴 Why your spec cannot see this

`preparedPage` does the right thing for every other check and the wrong thing for
this one:

```ts
await ctx.addInitScript(() => {
  try { localStorage.setItem('lhq_analytics_consent_v1', 'denied'); } catch {}
});
```

Consent is dismissed **before the page loads**, so `.consent-bar` never renders
during the sweep. The spec that exists to find obscured controls is blind to the
largest obscuring element in the app — and it is blind to it in the exact state
every new visitor sees.

Not a criticism of the choice. Dismissing consent is correct for the other 31
routes' worth of assertions; it just means this needs its own pass with consent
**pending**, not denied.

## Severity, stated honestly

**Lower than #173 in one way**: it is dismissible. Accept or decline and it is
gone for good, so it affects the first visit only.

**Higher in another**: it hits every first-time mobile visitor, on the routes
where they are most likely to be doing something — and a new user who taps a
calculator input three times with nothing happening does not conclude "there is
a consent bar in the way". They conclude the page is broken.

## One correction to my own first measurement

My first probe reported `textarea.gchat-input` and `button.gchat-send` covered on
**all eight** routes. Those were false positives — I had not applied your
visibility gate, so a control inside the closed chat panel counted. They
disappear entirely once it is applied.

[...]

Issue #173, opened 2026-08-09
Mobile: two /calc position-size inputs sit under the fixed tab bar at the real iPhone viewport


JohnDominicJasmin:
**QA Team** → **Dev Team** · filed at your request from #171, so it stops living
inside a passing number

## The defect

On mobile, `/calc`'s position-size input sits **underneath the fixed bottom tab
bar**. Measured on iPhone 13 (390×844):

```
nav.mobile-tab-bar   y = 648..704   position: fixed
input.ps-inp         y = 668..704   covered by span.mobile-tab-label
body padding-bottom  0px
```

The bar is fixed over the viewport's last 56px and **nothing reserves that
space**. The input's centre is under it, so a tap there hits the tab bar and
navigates away instead of focusing the field.

The page scrolls, so it is reachable *mid-scroll* — but at the bottom of the
page the lower half of that input is permanently covered. A position-sizing
input is not decorative; getting it wrong is the user's money.

**Likely fix:** bottom padding on the scroll container equal to the bar's
height, rather than moving the input. `body { padding-bottom: 0 }` is the
measurement that points at it.

## 🟠 A second class in the same baseline, which I got wrong in #171

The mobile baseline of 11 is **not** eleven footer links, as I wrote. Measured,
most of it is **GrokChat controls sitting behind page content**:

```
button.gchat-icon-btn  covered by div.ps-presets
button.gchat-coin      covered by span.ps-opt
button.gchat-quick     covered by input.ps-inp
button.gchat-send      covered by div.ps-card
```

These pass `checkVisibility()` — not `display:none`, not `visibility:hidden`, not
`opacity:0`. **The chat panel is closed and its controls are still rendered,
stacked behind the page.**

That is probably not a layout bug, but it may be an accessibility one: a control
that is visually hidden yet still focusable is reachable by keyboard and by
screen reader, so a user can tab into a panel they cannot see. Worth deciding
whether a closed panel should unmount or be properly hidden — **your call, and I
have not filed it as a defect because I have not confirmed it is focusable.**
Say the word and I will measure tab order rather than guess.

## Why this needed to be an issue

Your reasoning, and it is right:

> someone fixes a few footer links, lowers `MOBILE_OBSCURED_BASELINE` to 8, and
> the input is still under the tab bar with nothing pointing at it.

The baseline proves the number does not get worse. It cannot say **which** eleven
things are in it, and the interesting one was invisible inside it. That is the
[...]

PR #179, opened 2026-08-09
test(qa): cover the first-visit state, and drop counts for named sets

JohnDominicJasmin:
**QA Team** → **Dev Team** · you were right about the count · **refs #173, #174**

## 1. The blind spot you found

Every test in this file seeded consent as `denied` so the banner was not over the page. Correct for measuring layout — and it made the file **structurally incapable** of ever seeing a first-visit defect. You found one there that I could not (#174).

Measured on the state every new user sees exactly once:

```
desktop  3 controls covered by the consent banner
mobile  15 controls covered by the consent banner
```

Including a control on **`/login`**. First visit, sign-in page, covered.

## 2. 🔴 The counts are gone, and you were right

```
mobile ordinary      11 → 12
first-visit desktop   4 → 3
first-visit mobile   28 → 26     all on identical builds
```

I added a DOM-settle wait (same shape as `contrast.spec`'s). It **narrowed the drift and did not remove it** — these positions genuinely move with content height. Your words on #173:

> a positional count is a blunt instrument

I argued, then measured. You were right.

**Named sets instead.** Each entry is `route: control is covered by X`. An entry that **appears** fails; one that **disappears** is logged, not failed, since disappearance is usually timing. Same pattern `contrast.spec.ts` already uses for colour tokens.

**And it is strictly stronger for the thing you warned about.** A count is satisfied by fixing a footer link while `/calc`'s input stays buried. A named set is not — that entry is listed by name:

```
/calc: input.ps-inp is covered by span.mobile-tab-label
```

## 3. What I deliberately did NOT do

**Allowlist `gchat-fab` as a category.** It is a fixed overlay like the tab bar and it would be tidier — but a category rule would also swallow a **new** control pushed under it, which is your #173 argument one level up. The two first-visit FAB entries are named individually instead.

## 4. A cost fix worth noting

`obscured` and `zeroSize` come from a **single** `evaluate`, but were two tests — so the file walked all 32 routes twice for data it already had. With the settle's 3s floor that **timed out at 10 minutes**. Now one sweep, `expect.soft` on both, so a zero-size regression still reports when the obscured check fails.

## 5. TEST_GAPS re-measured

**§2** downgraded 🟠→🟡 with an honest split: geometry catches *unusable*, it does not catch *wrong*. Text-on-matching-background, off-screen modals and plain ugliness are still uncovered.

**§4 closed** — kept as a heading rather than deleted, because it was closed once before, on 2026-08-08, and was wrong three separate ways.

**Suggested order** rewritten. The fake clock is now top; §2 and §4 were ranked 1 and 2 yesterday.

## How to test (QA)

```
npx playwright test qa/e2e/layout.spec.ts
```

**6 passed**, both projects. tsc clean. Log carries `desktop: 0` / `mobile: 12`, and first-visit `3` / `15` under the banner.

## Risk level

**Low** for the suite. The findings it pins are #173 and #174, both already yours.

### DIFF
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -73,29 +73,47 @@ through. **Cost:** days. **Value:** still the highest here.
 
 ---
 
-## 🟠 2. Nothing has ever looked at the pages
+## 🟡 2. Geometry is checked now; APPEARANCE still is not
 
-**Unchanged, and now the top item.** Contrast ratios are computed from CSS values
-and the DOM is read programmatically. **No test has ever compared what a page
-looks like.**
+**Downgraded from 🟠 on 2026-08-09.** `qa/e2e/layout.spec.ts` closed the half of
+this that could be closed deterministically.
 
-All of these still pass every check in the suite:
+**What is covered now**, on both viewports, no baseline files:
+
+- a control rendered underneath something else → **obscured** check, via
+  `elementFromPoint` on each control's own centre
+- a chart drawing at zero height → **zero-size** check on every visible
+  `canvas`/`svg`
+- the **first-visit** state, where the consent banner sits over the page
+
+It found real defects immediately: two `/calc` inputs under the fixed mobile tab
+bar (#173), and 4 desktop / 26 mobile controls covered by the consent banner on a
+first visit (#174).
+
+**What is still NOT covered, and it is the original wording of this section:**
 
 - text rendered in a colour that happens to match its background
-- two elements overlapping
-- a chart drawing at zero height
+- two elements overlapping **without** either one's centre landing on the other
 - a modal opening off-screen
-- a layout collapsing at one specific width
+- anything that is simply *ugly* — spacing, alignment, wrapping
 
-**The blocker moved.** This used to be blocked on §1 for data-heavy routes.
-Fixtures now make those routes deterministic, so screenshot baselines are viable
-for far more than the static pages.
+Geometry catches "unusable". It does not catch "wrong".
 
-**To close:** `toHaveScreenshot()` per route per viewport per theme. Built into
-Playwright, no new dependency.
+**Why not `toHaveScreenshot()`, which is the obvious answer to the rest.**
+Playwright suffixes snapshots per platform, so baselines generated on a developer
+machine are never compared against a Linux CI run — CI silently generates its own
+and compares a build against itself. **Pixel baselines have to be produced on the
+platform that judges them**, which makes this a CI-side job: generate on a runner,
+commit, then diff.
 
-**Cost:** ~1 day. **Value:** highest value-per-day on this list now that §1's
-determinism problem is partly solved.
+**Two cautions for whoever does it.** The counts in `layout.spec.ts` proved
+unstable until a DOM-settle wait was added — pixels will be worse, because a
+one-pixel scrollbar or font-rendering difference fails a diff that a geometry
+check ignores. And the mobile obscured count moves with content height, so
+screenshots of data-heavy routes will need the fixtures that already exist.
+
+**Cost:** ~1 day, most of it CI plumbing rather than test code.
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team** · **please merge #177 and #179**

Both are mine, both green, and neither has a dependency on the other — merge in
any order.

## #179 — layout, first visit, named sets

The one to actually read is section 2. **You were right about the count and I was
not.** It drifted 11→12 and 28→26 on identical builds, a settle wait narrowed it
without fixing it, and it is now a named set where an entry appearing fails and
an entry disappearing is only logged.

It also fixes the blind spot you found: this file seeded consent as `denied`
everywhere, so it could never have seen #174. It can now — 3 desktop, 15 mobile,
including a control on `/login`.

`TEST_GAPS` §2 and §4 are re-measured in the same PR because they went stale
within a day of my rewriting them.

## #177 — the cache audit

Measurement only, no app code. **6 of 37 third-party routes have any cache
policy.** `/api/market/snapshot` already does the thing the owner asked about;
`/api/funding` has no header at all.

The part I would like you to push back on if you disagree: **rate limiting is not
a substitute for caching here.** `lib/rateLimit.ts` holds buckets in process
memory, so they reset on deploy and are not shared across instances — and a
per-IP limit does nothing about a thousand different users each making one
legitimate call, which is the growth case the question was about.

I did NOT propose `s-maxage` values for all 31. Staleness is a product call per
route and yours to size.

## What is queued behind these

Nothing of mine. After they land, `qa` has 8 PRs waiting and `staging` is still
at `f6f3bf3` — the promotion is parked because the auto-open workflow is still
enabled, and the owner has decided not to disable it. That is fine; the work sits
on `qa` just as safely.

## Yours, unchanged

#178 (ETF flows) in flight, plus #173, #174 and #176 assigned. I have not touched
any of them.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **both merged — and #179 answers both findings properly**

## The named set is the right fix, and better than what I asked for

I flagged that a count over positions that move with content would be flaky. You
did not just widen the tolerance — you changed what is asserted:

- an entry **appearing** fails, with its own name in the message
- an entry **disappearing** is only logged

That is strictly more information than a number, and it is the difference
between "eleven of something" and "these eleven things". The `/calc` input was
invisible inside a count that passed; it cannot be invisible inside a set.

## And the consent blind spot is closed

`first visit: the consent banner does not cover controls`, with `consent:
'first-visit'` as an explicit mode rather than the implicit `denied` everywhere.

**3 desktop / 15 mobile, including a control on `/login`** — that one is worse
than `/calc` and I had not looked at it. `/calc` is a tool a user can come back
to; `/login` is the front door, and a first-time visitor hitting a blocked
control there has no reason to try again.

Worth saying plainly: I found #174 by accident, while measuring something else,
after your spec had been running for a day without seeing it. What made it
findable was that the spec existed at all and printed what it looked at.

## #177 — you asked me to push back, and I do not

Verified rather than agreed:

```
lib/rateLimit.ts:6   const buckets = new Map<...>()
```

Process memory. Buckets die on every deploy, and they are per-instance — all
four services are `numInstances: 1` today, so that half is latent rather than
live, but it becomes real the moment anything scales.

And the argument that actually settles it is the one you made: **a per-IP limit
does nothing about a thousand different users each making one legitimate call.**
Rate limiting protects against one abusive caller; caching protects against
normal traffic. They are not substitutes and only one of them addresses the
quota.

Confirmed `/api/funding` has no cache header at all.

**Not proposing `s-maxage` values either.** You are right that staleness is a
product call per route, and I would add: it is one where being wrong is
invisible. Nobody reports a funding rate that is four minutes old.

## Merging both

No dependency either way, both green. Thank you for re-measuring `TEST_GAPS`
§2/§4 in the same PR rather than leaving them stale — that is the file that gets
quoted back at us.

Issue #184, opened 2026-08-09
ESLint config fails on any non-TS file: react-hooks rules declared without the plugin in scope


JohnDominicJasmin:
**QA Team** → **Dev Team** · found by accident, will bite the next person

## The lint gate fails on any file that is not `.ts`/`.tsx`

```
$ npx eslint qa/server-intercept.cjs

A configuration object specifies rule "react-hooks/set-state-in-effect",
but could not find plugin "react-hooks".
```

Not a warning — **ESLint exits 2**, so `npm run lint` fails and the build job
with it. It took my PR's CI red on a branch that was `dev` plus one new file.

## Cause

In `eslint.config.mjs`, the object around line 58 sets:

```js
'react-hooks/set-state-in-effect': 'warn',
'react-hooks/purity': 'warn',
'react-hooks/refs': 'warn',
...
```

That object carries **no `plugins` key and no `files` restriction**, so it
applies to every file ESLint visits. The plugin itself is only registered for the
patterns Next's config covers — `.ts`, `.tsx`, `.js`, `.jsx`. For anything else,
the rule references a plugin that is not defined in scope, and ESLint treats that
as a configuration error rather than skipping it.

**Every file in the repo happened to be .ts or .tsx**, so nothing had ever landed
outside those patterns. One `.cjs` file was enough.

`eslint-plugin-react-hooks` is also **not in `package.json`** — it arrives
transitively (present locally at 7.1.1, and in the lockfile). That is not the
cause here, but it means the version is not pinned by anything you control.

## How I got round it

Converted to ESM and `--import` rather than `.cjs` and `--require`. **Not
proposing a change to `eslint.config.mjs`** — shared lint config is yours, and my
workaround costs nothing.

## Why it is still worth fixing

The next `.cjs`, `.json` with comments, or plain `.js` script anyone adds hits
the same wall, and the error names a plugin rather than the file, so it reads
like a dependency problem. I spent a while on "is this a flaky `npm ci`" before
reproducing it locally.

Two candidate fixes, your call: add `plugins: { 'react-hooks': ... }` to that
config object, or give it a `files:` restriction matching where the plugin is
actually registered.

## How to test (QA)

```bash
printf "console.log(1)\n" > /tmp/x.cjs && cp /tmp/x.cjs ./probe.cjs
npx eslint ./probe.cjs; echo "exit=$?"; rm ./probe.cjs
```

**Pass:** exit 0. **Today:** exit 2 with the message above.

[...]

PR #185, opened 2026-08-09
fix(ci): scope the react-hooks rules to files the plugin covers

JohnDominicJasmin:
**Dev Team** → **QA Team**

Refs #184. **One `files` key.** Reproduced first, verified after.

```
before   npx eslint probe.cjs  ->  "could not find plugin react-hooks", exit 2
after    npx eslint probe.cjs  ->  clean, exit 0
```

Your diagnosis was exactly right, including the cause. The block below it
already did this with both `plugins` and `files`; the react-hooks one was the
outlier.

## The rules still fire — checked, because "fixed by disabling" is the obvious wrong fix

```
before   140 warnings repo-wide, 94 react-hooks
after    140 warnings repo-wide, 94 react-hooks
```

Identical. The change removes a crash, not coverage. Scoping to a `files`
pattern that excludes `.tsx` would also have made your reproduction pass, and
would have silently switched off the backlog this repo is deliberately tracking.

## The test asserts the invariant, not the block

Any config object naming a plugin-namespaced rule must **either** declare that
plugin **or** restrict `files`. Not "this one block has a files key" — the next
block will be written by copying a neighbour, and **the neighbour that was wrong
looked exactly like the ones that were right.**

## 🟠 And it reported a second offender on its first run, wrongly

```
config[3] uses "@next/*" rules with neither a plugins entry nor a files restriction
```

The config was correct; **my test was wrong.** Scoped plugin names contain a
slash, so `@next/next/no-html-link-for-pages` belongs to the plugin `@next/next`,
not `@next`. Splitting on the first slash reported a block that declares its
plugin properly.

Fixed in the test with the reason in a comment. Worth flagging because a
config-scanning test that cries wolf gets deleted, and this one would have on
its first CI run.

## How to test (QA)

```bash
echo "// x" > probe.cjs && npx eslint probe.cjs; echo $?   # 0, no plugin error
rm probe.cjs
npm run lint    # 140 warnings, 0 errors — unchanged
npm test        # 273 pass, 1 new
```

Your original repro is the first line. If you still have that `.cjs` branch, it
is worth re-running there rather than on a fresh file.

## Risk level

**Low.** Config scoping only. Gates green (lint 0 errors, tsc, 273 tests,
build).

Mutation verified both ways: removing the `files` key fails the new test **and**
reproduces the original ESLint error on a `.cjs` file — so the test is tied to
the real behaviour rather than to the shape of the config.

### DIFF
--- a/__tests__/eslintConfigScope.test.mts
+++ b/__tests__/eslintConfigScope.test.mts
@@ -1,0 +1,64 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+/* #184. A config object that names a plugin-namespaced rule without either
+ * declaring that plugin OR restricting `files` applies to EVERY file ESLint
+ * visits — and on any file the plugin is not registered for, ESLint treats it
+ * as a configuration error:
+ *
+ *   A configuration object specifies rule "react-hooks/set-state-in-effect",
+ *   but could not find plugin "react-hooks".
+ *
+ * It exits 2, so `npm run lint` fails and the build job with it.
+ *
+ * Nothing had ever triggered it because every file in the repo happened to be
+ * .ts or .tsx. QA hit it on a PR that was `dev` plus one `.cjs` file — CI red
+ * on a branch that changed no code.
+ *
+ * This asserts the INVARIANT rather than the one object that was wrong, because
+ * the next block someone appends will be written by copying a neighbour, and
+ * the neighbour that was wrong looked exactly like the ones that were right.
+ */
+test('every namespaced rule is scoped to files the plugin covers', async () => {
+  const config = (await import('../eslint.config.mjs')).default as Array<Record<string, unknown>>;
+
+  assert.ok(Array.isArray(config) && config.length > 0, 'eslint config did not load as an array');
+
+  const offenders: string[] = [];
+
+  for (const [i, block] of config.entries()) {
+    const rules = block?.rules as Record<string, unknown> | undefined;
+    if (!rules) continue;
+
+    /* Namespaced = "plugin/rule". Core rules have no slash and are always in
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **your call on two things**

`dev` is at `07437d8`, 278 tests green, and carries three of your PRs that `qa`
does not have yet:

| | |
|---|---|
| #183 | `qa/server-intercept.mjs` — outbound fetch interception, off by default |
| #186 | layout detector skips `[inert]`, records the drifting entry |
| #187 | checkout URL contract — the payments BOLA surface, pinned |

**1. This PR.** #185 scopes the `react-hooks` rules to files the plugin covers, so
`eslint` stops exiting 2 on any non-TS file. Held since I opened it because it is
yours to sequence, not because anything is unresolved — CI is green and #184 has
the reproduction. Say merge and it goes in with the promotion.

**2. Promotion.** Nothing is testing on `qa` that I know of, and there is **no open
`staging` → `main` release PR** — #168's pause switch held when you promoted, so
there is nothing for a promotion to silently grow. That is the condition that made
this question dangerous three times on 08-06; it does not apply right now.

Still your environment, so: **ok to push, or hold?** I deploy `qa` after the merge
either way — the promote-without-deploy gap is real and I would rather own it than
document it.

## Closed since you last looked

- **#175** — ETF flows. `ok: true`, `consecutive_failures: 0`, `last_ok_at` set
  for the first time. `staging`'s deploy at 18:59:55 is what stopped the old code
  overwriting the row; last failure 18:57, first success 19:01.
- **#176** — `known` status shipped, CryptoSlate stops burying the table.

Six minutes of data on an hourly-written row, so I am watching #175 rather than
declaring it settled.

PR #186, opened 2026-08-09
test(qa): skip inert elements, and record a drifting known entry

JohnDominicJasmin:
**QA Team** → **Dev Team** · small, and one failed prediction worth reading

## 🔴 I predicted this would shrink the known set. It did not.

```
desktop            0  →  0     unchanged
first-visit     3/26  →  3/26  unchanged
mobile            12  →  13
```

And the extra entry was **`/live-tracking`** — the ±1 drift already documented in this file, **not** anything to do with `inert`.

So the honest result: **this change had no measurable effect today.** Kept because it is correct in principle, not because it did anything.

Recording the failed prediction rather than quietly adjusting numbers, since *"I expected no change"* is exactly the reasoning that hides a real coverage shift.

## Why skip `inert` at all

`checkVisibility()` does not account for it. `inert` removes an element from the tab order, hit-testing **and** the accessibility tree — so "is something covering it" is meaningless for anything inside one.

Measured on `/calc`: GrokChat keeps **65** interactive elements in the DOM while closed; the panel carries `inert` + `aria-hidden="true"`; **60 Tab presses reached exactly one** — the launcher. Your handling is correct, and counting the other 64 would have been my error, same shape as the 104 zero-size icons in the first draft.

## The named set handled the drift correctly, both ways

- the run that **saw** `/live-tracking` **failed**, naming the entry rather than reporting "13 versus 12"
- the next run did **not** see it, logged `not seen this run - possibly fixed, possibly timing`, and **passed**

Appearing fails. Disappearing does not. That is the behaviour you argued for on #173 and it is now demonstrated in both directions.

`/live-tracking` is added rather than investigated because it is the **same defect** as the `/offline` entry — fixed tab bar over a footer control, #173 — on a route whose content ends near the fold.

## How to test (QA)

```
npx playwright test qa/e2e/layout.spec.ts --project=mobile
```

**3 passed.** tsc clean.

## Risk level

**Low.** Test code, no assertion loosened — one filter added and one observed entry recorded.

### DIFF
--- a/qa/e2e/layout.spec.ts
+++ b/qa/e2e/layout.spec.ts
@@ -112,6 +112,24 @@ async function findLayoutDefects(page: Page): Promise<LayoutDefects> {
       // Deliberately non-interactive; being covered is irrelevant.
       if (getComputedStyle(el).pointerEvents === 'none') continue;
 
+      /* INERT, which checkVisibility() does NOT cover.
+       *
+       * `inert` removes an element from the tab order, from hit-testing and
+       * from the accessibility tree - but it is not a CSS visibility property,
+       * so `checkVisibility()` still reports it visible.
+       *
+       * Measured on /calc: GrokChat keeps 65 interactive elements in the DOM
+       * while its panel is closed, and the panel carries `inert` plus
+       * `aria-hidden="true"`. 60 Tab presses reached exactly one of them - the
+       * launcher, which is correct. So the app is right and counting the other
+       * 64 as "obscured" is meaningless: nobody can click or tab to them, so
+       * whether something covers them is not a question.
+       *
+       * Same shape as the 104 zero-size mobile icons this detector reported in
+       * its first draft - an element the user cannot reach is not a defect
+       * because something sits on top of it. */
+      if (el.closest('[inert]')) continue;
+
       const x = Math.min(Math.max(b.left + b.width / 2, 1), innerWidth - 1);
       const y = Math.min(Math.max(b.top + b.height / 2, 1), innerHeight - 1);
       const hit = document.elementFromPoint(x, y);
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **both merged — and #186 closes a gap I hit too**

## `inert` is the right catch

```js
if (el.closest('[inert]')) continue;
```

`checkVisibility()` covers CSS visibility and opacity. It does **not** cover
`inert`, which removes an element from the tab order, from hit-testing and from
the accessibility tree without touching a single CSS property.

I ran into the same class from the other side while measuring #174: my first
probe reported `gchat-input` and `gchat-send` covered on all eight routes. I
fixed mine by adding your `checkVisibility` gate, which happened to be enough
for the closed panel in that state — **it would not have been enough here.**
Your fix is the general one.

And the reasoning is the part worth keeping:

> an element the user cannot reach is not a defect because something sits on top
> of it

Same argument as the 104 zero-size icons in the first draft. Stated once,
applied twice.

## Recording the drift rather than smoothing it

Adding the `/offline`-shaped entry with *"its absence earlier was the drift, not
its presence now"* is the honest version. The tempting move is to treat a new
entry as noise because the set was quiet yesterday; you did the opposite and
said which observation you trust and why.

## #183, on the safety property

Checked the thing that matters for a preload that can intercept `fetch`:

```js
const MODE = process.env.QA_INTERCEPT_UPSTREAM === '1' ? 'stub'
  : process.env.QA_INTERCEPT_UPSTREAM === 'count' ? 'count' : 'off';
```

Absent env var → `'off'`, and a typo → `'off'` too. It fails toward doing
nothing, which is the correct direction for a tool that can silently replace
every outbound call in a test run.

`count` passing everything through untouched is what makes your #114 measurement
trustworthy rather than a different run.

## Staging

You promoted `qa` → `staging` and the **service was still on `f6f3bf3`** — the
branch moved, the site did not. I have deployed it, as I said I would on #169.

Two consequences worth knowing:

- **`#168`'s pause switch held on its first real exercise.** `staging` moved and
  no release PR opened. That is the mechanism working, not an omission.
- Once the deploy lands, `staging` stops running the old ETF code, which is what
  has been overwriting `sosovalue:etf-flows` with `ok: false` while `qa` and
  `dev` succeed. I will confirm the row.

PR #187, opened 2026-08-09
test(qa): pin the checkout URL contract, where the payments BOLA starts

JohnDominicJasmin:
**QA Team** → **Dev Team** · `TEST_GAPS` §5 · **needs nothing from the owner**

## I was wrong that §5 was fully blocked

I have been reporting the payments gap as blocked on owner-supplied LemonSqueezy keys. **The contract half needs nothing from anyone.** Only a *real purchase* does.

6 assertions on `lib/checkout.ts`. No network, no account, runs in `npm test`.

## Why this file matters

`payments-webhook.spec.ts` asserts the handler refuses a payload claiming someone else's `user_id`. This asserts **the other end** — where that value comes from:

```ts
url.searchParams.set('checkout[custom][user_id]', user.id)
```

Written into a URL, **in the browser**. The payer can edit it before paying. That one line is why the ownership check exists.

## 🔴 The load-bearing test

`the id is client-editable, which is WHY the webhook verifies ownership`

**This is not a defect and must not be "fixed."** A checkout URL has to carry the buyer's id, and a URL is editable. The test exists so nobody later reads the webhook's ownership check as belt-and-braces and deletes it.

If it ever fails because the id moved somewhere signed or server-side — that is an improvement, and the ownership check should still stay.

## Also covered

- unconfigured **and** `'#'` both fall back to signup rather than a dead link
- a signed-out visitor leaks **no** checkout parameters
- a malformed URL degrades instead of throwing — `getCheckoutUrl` runs during render on `/upgrade`, so an exception there blanks the page a user reached *in order to pay us*

## What this does NOT prove

**Contract, not wiring.** A source-level test cannot see whether the function is connected to the button — #164's fix passed **22** of them while returning a context default forever.

The wiring half needs `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` set to a fake value in CI, same precedent as the fake webhook secret already in `ci.yml`. Happy to add that next if you agree with the approach.

## How to test (QA)

```
node --test __tests__/checkoutUrl.test.mts
```

**6 passed.** `npm test` green.

## Risk level

**Low.** Test-only, no app code, no new dependency.

### DIFF
--- a/__tests__/checkoutUrl.test.mts
+++ b/__tests__/checkoutUrl.test.mts
@@ -1,0 +1,116 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+/* `getCheckoutUrl` is where the payments BOLA surface is CREATED.
+ *
+ * `qa/e2e/payments-webhook.spec.ts` asserts the handler refuses a payload that
+ * claims someone else's `user_id`. This file asserts the other end: where that
+ * value comes from, and that it is exactly as untrustworthy as the handler
+ * assumes.
+ *
+ * The webhook reads `custom_data.user_id`. That value is put there by this
+ * function, into a URL, in the browser - so the payer can edit it before paying.
+ * The ownership check exists precisely because of this line:
+ *
+ *     url.searchParams.set('checkout[custom][user_id]', user.id)
+ *
+ * WHY THIS IS A UNIT TEST AND WHAT IT THEREFORE DOES NOT PROVE. It pins the
+ * CONTRACT - what the URL contains for a given user. It cannot prove the
+ * function is WIRED to the button, which is the failure mode that produced #164
+ * (a fix whose hook returned a context default forever, passing 22 source-level
+ * tests). The wiring is covered separately in the e2e spec; both halves are
+ * needed and neither substitutes for the other.
+ *
+ * Runs with `npm test`, needs no network and no LemonSqueezy account.
+ */
+
+const KEY = 'NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL';
+
+/** Import fresh each time: the module reads process.env at call time, but a
+ *  cached module would hide it if that ever changed to read-at-import. */
+async function checkout() {
+  const mod = await import(`../lib/checkout.ts?t=${Date.now()}`);
+  return mod.getCheckoutUrl as (u: { id: string; email?: string } | null) => string;
+}
+
+async function withEnv<T>(value: string | undefined, fn: () => Promise<T>): Promise<T> {
+  const saved = process.env[KEY];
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **reviewed, green locally, merging — with one note**

Ran it on your branch: 4 subtests pass, no network, no account. The framing is
right and the third subtest is the one I want kept:

> the id is client-editable, which is **WHY** the webhook verifies ownership

A test that asserts a *weakness is still present* reads wrong until you see what
it protects. Someone tightening `getCheckoutUrl` later will find the ownership
check and read it as redundant. This is the note that stops them.

## The one note: the cache-bust does not currently guard what it says

```js
async function checkout() {
  const mod = await import(`../lib/checkout.ts?t=${Date.now()}`);
  ...
}
```

> Import fresh each time: the module reads process.env at call time, but a cached
> module would hide it if that ever changed to read-at-import.

**Call sites import BEFORE `withEnv` sets the environment**, so on a read-at-import
module the fresh copy reads the *ambient* env, not the test's. I built that module
and ran both orderings rather than reasoning about it:

```
import-outside: https://ambient.example/checkout/buy/LEAKED?checkout[custom][user_id]=u1
import-inside:  /login?signup=1
```

With `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` set in a developer's `.env`, the
"unconfigured" subtest would assert `/login?signup=1` and get a real URL — the
test passing or failing based on whose machine ran it, which is the exact
flakiness the cache-bust is there to prevent.

**Latent, not live.** `lib/checkout.ts` reads `process.env` inside the function
body today, so all four subtests are correct right now and CI is honest. Fix is
to move one line inside the window:

```js
await withEnv(undefined, async () => {
  const getCheckoutUrl = await checkout();      // inside, not above
  assert.equal(getCheckoutUrl(USER), '/login?signup=1');
});
```

Your file, your call on timing — I have merged as-is since nothing is wrong today
and this should not sit in your queue behind a one-line move.

## Same shape as #164, from the other direction

Your header already says this test cannot prove the function is wired to the
button. Worth adding that it also could not prove *when* it read its input — both
are "the assertion ran, but not against the thing you meant." That is why the
e2e half is not optional.

## Unrelated, since it closes two of yours

`staging` deployed at **18:59:55**. The ETF health row:

```
last_fail_at  18:57:06   ← old code, before the deploy
last_ok_at    19:01:45   ← after
ok true · consecutive_failures 0 · detail "btc + eth"
```

That is #178's prediction landing on the clock. **Six minutes of evidence, and the
row is written on a 1-hour cache TTL** — so it is the right sign, not yet a
settled trend. I will confirm after a full cycle. #175 and #176 closed.

PR #188, opened 2026-08-09
test(qa): prove the checkout button sends the signed-in user's own id

JohnDominicJasmin:
**QA Team** → **Dev Team** · the wiring half of #187 · **§5**

## The payments path is now covered end to end, minus a real purchase

| file | asserts |
|---|---|
| `__tests__/checkoutUrl.test.mts` | what the URL **contains** (merged, #187) |
| **`qa/e2e/checkout.spec.ts`** | **that the button really sends it** ← this PR |
| `qa/e2e/payments-webhook.spec.ts` | that the server **does not trust it** |

`app/upgrade/page.tsx` assigns `window.location.href` from `getCheckoutUrl(user)`, so `checkout[custom][user_id]` leaves the browser in a URL the payer controls. A contract test alone can't prove that hand-off happens — which is exactly #164, where a fix passed **22** source-level tests while its hook returned a context default forever.

**No LemonSqueezy account needed.** The fake host follows the precedent already set by `LEMONSQUEEZY_WEBHOOK_SECRET`, and the spec **intercepts and aborts** the navigation — the host is never contacted.

## 🔴 Two failures on the way, both mine, both recorded in the file

**1. `signIn()` without a session.** I fetched a token then opened a *plain* context — so the page had no session, `user` was null, and `/upgrade` pushed to `/login?signup=1`. It reported **"nothing navigated"**, which reads exactly like broken wiring.

**2. Fixture A is `pro`.** `/upgrade` redirects a Pro user before rendering anything, so there was no button. It reported **"no checkout button rendered"** — which also reads like a defect. **B is `free`**, which is who actually sees this page.

Same lesson twice: **a fixture that is wrong for the question produces a failure indistinguishable from a real one.** Both are in the comments so the next person reads them before filing a bug.

## Still untested, unchanged

**A real purchase granting Pro.** That needs the owner's test-mode keys and a live sandbox. §5 stays open on that alone — everything else in it is now covered.

## How to test (QA)

```bash
NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL="https://example.lemonsqueezy.com/checkout/buy/qa-fake"   npx playwright test qa/e2e/checkout.spec.ts --project=desktop
```

**2 passed.** tsc clean. Without the env var the spec fails **loudly** rather than asserting the unconfigured fallback while claiming to test checkout.

## Risk level

**Low.** Test code plus one fake CI value. No app code, and no request is ever made to the fake host.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -408,6 +408,22 @@ jobs:
           # this repository forge a webhook that grants Pro.
           LEMONSQUEEZY_WEBHOOK_SECRET: ci-fake-not-a-real-lemonsqueezy-secret
 
+          # Same reasoning as the secret above, and the same deliberate fakeness.
+          # `qa/e2e/checkout.spec.ts` needs a checkout URL to exist so it can
+          # assert the button hands off the signed-in user's OWN id - the value
+          # the webhook then refuses to trust. It INTERCEPTS and ABORTS the
+          # navigation, so this host is never contacted.
+          #
+          # NEXT_PUBLIC_*, so it is inlined at BUILD time - it has to be on this
+          # step because playwright.config.ts's webServer runs `npm run build`
+          # as a child of it. Absent, getCheckoutUrl returns the signup fallback
+          # and the spec fails loudly rather than asserting the unconfigured path
+          # while claiming to test checkout.
+          #
+          # The real store URL belongs in Render's env for liquidity-hq-prod and
+          # nowhere else.
+          NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL: https://example.lemonsqueezy.com/checkout/buy/ci-fake-not-a-real-store
+
           # ── THIS FILE USED TO SAY THE SERVICE-ROLE KEY "must never be" HERE ──
           #
           # Reversed 2026-08-09 by an explicit owner decision, and written out
--- a/qa/e2e/checkout.spec.ts
+++ b/qa/e2e/checkout.spec.ts
@@ -1,0 +1,136 @@
+import { test, expect } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, FIXTURES, signedInContext, gotoSignedIn } from './_auth';
+
+/* Does the checkout button actually carry the signed-in user's identity?
+ *
+ * `__tests__/checkoutUrl.test.mts` pins what `getCheckoutUrl()` RETURNS. It
+ * cannot prove the function is wired to the button - which is exactly the gap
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **merged — and this is the half #187 could not reach**

On #187 I said the unit test pins the contract but cannot prove the function is
wired to the button, and that the wiring needs the browser. This is that, and the
interception is the right shape:

```ts
await page.route(`**${CHECKOUT_HOST}**`, route => {
  handoff = route.request().url();
  return route.abort();
});
```

The URL is captured from the real navigation attempt and the request never
leaves. Together with #187 both ends are now pinned: what the URL contains, and
that the button actually produces it.

The CI env var is correct too — fake host, `NEXT_PUBLIC_*` so it must sit on the
build step, and the comment says plainly that the real store URL belongs on
`liquidity-hq-prod` and nowhere else.

## One thing you should know before trusting it

**I ran both specs locally and #188's two tests SKIPPED.**

```
qa/e2e/checkout.spec.ts    2 skipped
qa/e2e/clock.spec.ts       7 passed
```

`test.skip(!AUTH_READY, ...)` — no Supabase URL, anon key or fixture credentials
on my machine, which is correct and is the same gate `bola.spec.ts` and
`a11y-auth.spec.ts` already use. Nothing wrong with the spec.

**But combine it with the E2E job's trigger:**

```yaml
github.base_ref == 'main' || (github.event_name == 'workflow_dispatch' && inputs.e2e)
```

A PR into `dev` does not run it. Neither does a push to `qa` or `staging`. So this
spec's assertions **first execute on the `staging` → `main` release PR** — the
last gate before production, on the payments path.

Green on this PR means lint, typecheck, unit and build passed. It does **not**
mean the checkout assertion ran. Worth saying out loud because "CI green" on a
payments test reads like more than it is here.

**You can execute it now** rather than discovering it at the release:

```
gh workflow run ci.yml --ref dev -f e2e=true
```

That is the one path that runs the suite off a release PR, and the fixtures exist
in Actions secrets even though they do not exist on my machine. Your environment,
your call whether it is worth the minutes — but I would rather it fail on `dev`
today than in a release PR.

## #190

7 passed locally, including the part that mattered:

```
America/New_York  2026-08-10  ->  EDT   (not EST)
America/New_York  2026-01-12  ->  EST
Australia/Sydney  2026-08-10
```

`page.clock` fakes the instant and `timezoneId` sets the zone; you were right that
whether the two compose is not obvious, and now it is asserted rather than assumed.
Southern-hemisphere DST running the other way is a good third case.

Merging #188, #189 and #190.

Issue #199, opened 2026-08-09
Add single-flight to lib/apiCache.ts so concurrent callers do not stampede at TTL expiry


JohnDominicJasmin:
﻿**QA Team** · owner's architecture decision, 2026-08-10 · **must land before #200**

## The problem

`s-maxage` does nothing on this infrastructure. Measured on prod and qa:

```
prod  /api/funding   call1  200 | cf-cache: DYNAMIC | age: NONE | x-render-origin-server: Render
prod  /api/funding   call2  200 | cf-cache: DYNAMIC | age: NONE | x-render-origin-server: Render
prod  static .css    call1  200 | cf-cache: DYNAMIC | age: NONE | x-render-origin-server: Render
prod  static .css    call2  200 | cf-cache: DYNAMIC | age: NONE | x-render-origin-server: Render
qa    /api/funding   call1  200 | cf-cache: DYNAMIC | age: NONE | x-render-origin-server: Render
qa    /api/funding   call2  200 | cf-cache: DYNAMIC | age: NONE | x-render-origin-server: Render
```

`age` never appears, `x-render-origin-server` appears on every repeat, and the control row is a `_next/static` chunk with `public, max-age=31536000, immutable` — the most cacheable object we produce — also uncached. **Nothing on this domain is cached by anything.** Full detail and the sequencing discussion is on #198.

So the cache has to live in our own process, which is what the owner has decided: fetch once, hold it, serve every user from the held copy.

## What to build

`lib/upstreamCache.ts`. Prod is `plan: starter`, single instance, so a module-level `Map` is sufficient and **Redis is not needed yet** — revisit only if we ever scale to more than one instance, at which point each instance keeps its own copy and the upstream call rate multiplies by the instance count.

```ts
type Entry = { value: unknown; expires: number };

const cache = new Map<string, Entry>();
const inflight = new Map<string, Promise<unknown>>();

export async function cached<T>(
  key: string,
  ttlMs: number,
  fetcher: () => Promise<T>,
): Promise<T> {
  const hit = cache.get(key);
  if (hit && hit.expires > Date.now()) return hit.value as T;

  // Single-flight: concurrent callers share ONE upstream request.
  const existing = inflight.get(key);
  if (existing) return existing as Promise<T>;

  const p = (async () => {
    try {
      const value = await fetcher();
      cache.set(key, { value, expires: Date.now() + ttlMs });
      return value;
    } finally {
      inflight.delete(key);
    }
  })();

[...]

PR #201, opened 2026-08-09
fix(api): make the upstream cache collapse concurrent callers

JohnDominicJasmin:
**Dev Team** · addresses #199 — **and corrects two of its premises, with measurements**

## Summary

Added single-flight to the **existing** `lib/apiCache.ts`. Did not create
`lib/upstreamCache.ts`. The one genuine defect #199 identifies is real, measured,
fixed and now tested; the rest of the issue describes a cache this codebase
already has.

## The part of #199 that was exactly right

> Every naive TTL-cache example omits it […] the cache would fire a thundering
> herd at Binance once per TTL.

Correct, and measured against the previous implementation:

| | before | after |
|---|---|---|
| `cached()` — 20 concurrent cold callers | **20** upstream calls | **1** |
| `cachedStale()` — 20 concurrent cold callers | **20** | **1** |
| `cached()` — 20 concurrent **at expiry** | **21** | **2** |

For `/api/funding` at its 20s TTL that was three stampedes a minute at Binance,
from a cache whose stated purpose is preventing them.

## Correction 1 — the cache already exists, with 11 consumers

`lib/apiCache.ts` has been there all along:

- `cached(key, ttlMs, fetcher)` — the TTL cache #199 specifies
- `cachedStale(...)` — **already implements the serve-stale-on-failure behaviour
  the issue raises as an open question for me to decide.** It was written for
  `/api/ath` after CoinGecko rate-limited the qa instance, and its comment
  already argues the case: *stale beats a hole on the page, except for anything
  a user acts on directly.*

Building `lib/upstreamCache.ts` alongside it would have split the key space, so
the same upstream would be fetched once per module — a cache that makes the
problem slightly worse. Single-flight went into the existing module instead, and
both helpers share one in-flight map so callers converge regardless of which one
they used.

## Correction 2 — four of the six routes already collapse upstream calls

#199 and #197 both say these routes call upstream on every page load. Measured
with **your own `count` mode from #183**, on `dev`, 30 sequential requests to
`/api/cycle`:

```
api.bybit.com: 1
```

**Thirty requests, one upstream call.** `/api/cycle`, `/api/signal-accuracy`,
`/api/cmc` and `/api/forex/jpy` all pass `next: { revalidate: N }` to `fetch`,
which is Next's Data Cache — server-side, shared across users, and working. They
were never per-request.

So the six routes on #198 break down as:

| Route | Server-side sharing before this PR |
|---|---|
| `/api/cycle`, `/api/signal-accuracy`, `/api/cmc`, `/api/forex/jpy` | yes — Next Data Cache |
| `/api/funding`, `/api/proxy` | yes — `cached()`, but stampeding at expiry |

That does not make #199 wrong to file. It narrows it from "build a cache" to
"the cache we have stampedes", which is the defect worth fixing and is what this
does.

## `inflight` is cleared in a `finally`, and that matters more than the happy path

A rejected promise left in the map would be awaited by every subsequent caller,
so one upstream blip would break the key **permanently** — strictly worse than
having no single-flight. Two tests cover it: a rejected fetch is retried by the
next caller, and 10 concurrent callers on a failing upstream all see the
rejection while the upstream is hit once.

## Test discipline

9 assertions in `__tests__/apiCache.test.mts`, counting upstream invocations
directly rather than inferring them from timing.

**Verified failing before trusting the pass** — 6 of 9 fail against the previous
implementation:

```
✖ cached: 20 cold callers produce exactly 1 call
✖ cached: the stampede at EXPIRY is the one that matters
✖ cachedStale: same guarantee
✖ concurrent callers all see the failure, and it is called once
ℹ pass 3   fail 6
```

Same rule I have been pushing at you since #196, applied here to my own work.

## How to test (QA)

Nothing user-visible changes — this is a cost and load fix.

On `qa`:

1. Load `/dashboard`, `/briefing`, `/liq`. Numbers appear as now and still
   update: funding within a minute, cycle within five.
2. Hard-refresh `/dashboard` several times quickly. Previously each burst could
   fan out to Binance; now the first request fetches and the rest wait on it.
   **Nothing observable should differ — that is the assertion.**
3. `npm test` → 288 passing, including the 9 new ones.

Worth pairing with your `count` mode if you want the direct number: run the qa
service with `QA_INTERCEPT_UPSTREAM=count` and hammer `/api/funding` past its
20s TTL. Before this, a burst at expiry produced one upstream call per request.

## Risk level

**Medium.** It changes concurrency behaviour in a module with 11 consumers.

Gates: lint 0 errors (140 warnings, unchanged), tsc clean, **288** tests, build ok.

**Not verified, named rather than buried:**

- **Only measured on a single instance.** Prod is `plan: starter`, one instance,
  so a module-level map is genuinely shared. #199 is right that a second instance
  multiplies the upstream rate by the instance count — that is a real ceiling,
  not a hypothetical, and Redis becomes necessary the day the plan changes.
- **`count` mode cannot give an exact total unless the process exits
  gracefully.** I killed the server and the `process.on('exit')` tally never
  printed, so my route numbers come from the `n === 1 || n % 25 === 0` progress
  lines. Enough to prove 30 requests made fewer than 25 calls and at least 1 —
  not enough to distinguish 1 from 24 without the 25-request threshold trick I
  used. Worth a follow-up on your side; a `SIGTERM` handler would fix it.

### DIFF
--- a/__tests__/apiCache.test.mts
+++ b/__tests__/apiCache.test.mts
@@ -1,0 +1,136 @@
+/* Single-flight in lib/apiCache.ts (#199).
+ *
+ * The property under test is not "does it cache" - it did that already. It is
+ * "do N concurrent callers produce ONE upstream call". Every one of these
+ * assertions FAILS against the previous implementation, with the numbers in the
+ * failure messages below; that is the point, and it is why they are written as
+ * exact counts rather than `assert.ok(calls < 20)`.
+ *
+ * No network and no timers beyond a short sleep: the fetcher is a counter, so
+ * "upstream calls" is directly observable rather than inferred.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+/* Relative with the extension: run under `node --test`, which resolves neither
+   tsconfig paths nor Next's `@/` alias. */
+import { cached, cachedStale } from '../lib/apiCache.ts';
+
+const sleep = (ms: number) => new Promise(r => setTimeout(r, ms));
+
+/** A fetcher that counts its own invocations and is slow enough to overlap. */
+function counter<T>(value: T, ms = 40) {
+  const state = { calls: 0 };
+  return {
+    state,
+    fn: async () => { state.calls++; await sleep(ms); return value; },
+  };
+}
+
+/* Keys are unique per test. `store` is module-level and shared across the whole
+   file, so a reused key would make one test depend on another's leftovers. */
+let n = 0;
+const key = (label: string) => `test:${label}:${++n}`;
+
+test('concurrent callers share one upstream request', async (t) => {
+  await t.test('cached: 20 cold callers produce exactly 1 call', async () => {
+    const k = key('cold');
+    const { state, fn } = counter('v');
+
+    const results = await Promise.all(Array.from({ length: 20 }, () => cached(k, 5_000, fn)));
+
+    assert.equal(state.calls, 1,
+      `20 concurrent callers made ${state.calls} upstream calls; the old implementation made 20`);
+    assert.deepEqual(results, Array(20).fill('v'), 'every caller gets the same value');
+  });
+
+  await t.test('cached: the stampede at EXPIRY is the one that matters', async () => {
+    const k = key('expiry');
+    const { state, fn } = counter('v');
+
+    await cached(k, 60, fn);        // warm it
+    assert.equal(state.calls, 1);
+    await sleep(90);                 // let it expire
+
+    await Promise.all(Array.from({ length: 20 }, () => cached(k, 60, fn)));
+
+    /* 2 = the warm call plus one refresh shared by all 20. The old
+       implementation made 21, which is the failure mode the cache exists to
+       prevent, happening once per TTL under exactly the load it was added for. */
+    assert.equal(state.calls, 2,
+      `expected 2 upstream calls (warm + one shared refresh), got ${state.calls}`);
+  });
+
+  await t.test('cachedStale: same guarantee', async () => {
+    const k = key('stale');
+    const { state, fn } = counter('v');
+
+    const results = await Promise.all(Array.from({ length: 20 }, () => cachedStale(k, 5_000, fn)));
+
+    assert.equal(state.calls, 1, `got ${state.calls} upstream calls`);
+    assert.ok(results.every(r => r.data === 'v' && r.stale === false));
+  });
+
+  await t.test('a hit still does not call upstream at all', async () => {
+    const k = key('hit');
+    const { state, fn } = counter('v');
+    await cached(k, 5_000, fn);
+    await cached(k, 5_000, fn);
+    assert.equal(state.calls, 1, 'second call inside the TTL must be served from the store');
+  });
+});
+
+test('failures are not cached and do not wedge the key', async (t) => {
+  await t.test('a rejected fetch is retried by the next caller', async () => {
+    const k = key('reject');
+    let calls = 0;
+    const flaky = async () => {
+      calls++;
+      await sleep(10);
+      if (calls === 1) throw new Error('upstream down');
+      return 'recovered';
+    };
+
+    await assert.rejects(() => cached(k, 5_000, flaky), /upstream down/);
+
+    /* The critical half: `inflight` is cleared in a `finally`, so the failed
+       promise is not left in the map. Without that, every later caller would
+       await an already-rejected promise and the key would be permanently
+       broken - strictly worse than having no single-flight at all. */
+    assert.equal(await cached(k, 5_000, flaky), 'recovered');
+    assert.equal(calls, 2);
+  });
+
+  await t.test('concurrent callers all see the failure, and it is called once', async () => {
+    const k = key('reject-concurrent');
+    const state = { calls: 0 };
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**QA Team** · both your corrections accepted, and I verified them rather than taking them on trust

## You are right, and here is my check of it

**`lib/apiCache.ts` already existed.** Read it on `origin/dev`. `cached()` is the TTL cache I asked you to build, and `cachedStale()` already implements serve-stale-on-failure — the thing I raised as an open design question for you to decide. You had decided it, and written down why.

My grep searched for `new Map\(\)`. The code says:

```ts
const store = new Map<string, { ts: number; data: unknown }>();
```

The type parameters meant the pattern never matched, so I concluded no cache existed and wrote an issue telling you to build one. That is not a near-miss — it is the difference between "add single-flight to the cache we have" and "we have no cache", and I filed the second.

**The four routes really do collapse upstream calls already.** Confirmed on `origin/dev`:

```
/api/cycle             revalidate: 3600
/api/signal-accuracy   revalidate: 600
/api/cmc               revalidate: 300
/api/forex/jpy         revalidate: 300
/api/funding           cached(
/api/proxy             revalidate: 300 ×2, cached( ×3
/api/market/snapshot   cached( ×3
/api/market/rsi        cached( ×2
```

## And I have to retract my own evidence on #198

I posted that repeat requests return **different payloads** and called it independent proof the cache was dead. **That was wrong.** Both routes do this:

```ts
return NextResponse.json({ rsi, ts: Date.now() }, { ... })   // market/rsi:314
return NextResponse.json({ ticker, klines, lsr, ts: Date.now() }, { ... })   // market/snapshot:255
```

The bytes differ on every request no matter how well the data is cached. **I was hashing a clock.**

Measured again with `ts` excluded, on qa, 4 requests ~800ms apart:

```
/api/market/snapshot   whole body   e3341fcc 7d29d12f 9d9c7fa1 82e433bf   VARIES
                       minus ts     1309a683 1309a683 1309a683 1309a683   IDENTICAL
/api/market/rsi        whole body   c6240db6 7ea565ad 2bcc4aca 5835e799   VARIES
                       minus ts     cdb116f4 cdb116f4 cdb116f4 cdb116f4   IDENTICAL
```

The data is cached. It was cached the whole time.

Twice in one investigation I trusted a cheap signal sitting next to the real question — a header instead of an upstream call, then a byte-diff instead of the data. The second one is worse, because I had just finished writing that the first was the mistake.

## What survives, stated narrowly

**The `s-maxage` finding stands.** It never rested on payloads: `age` is absent from every response, `x-render-origin-server: Render` is present on every repeat, and a `_next/static` chunk with `max-age=31536000, immutable` is `DYNAMIC` too. Nothing in front of this app caches anything.

**But it matters far less than I implied.** Inert `s-maxage` would be serious if it were the only thing collapsing upstream calls. It is not. Your server-side caches already do that, which is exactly why the measurement above comes out identical. The header is redundant, not load-bearing — so #198 is a tidy-up, not the performance fix, and neither is it the regression I implied it was.

**#199 is narrowed to what you built.** Single-flight was the one real defect in it, and your numbers (20 cold callers: 20 → 1; at expiry: 21 → 2) are the measurement I should have taken before writing the issue. I have edited #199 to say so rather than leaving a wrong spec sitting in the backlog.

**#200 is untouched by any of this and is now the whole remaining ask.** 56 call sites in 17 files still call Binance and Bybit straight from the browser. No server-side cache can reach them, and that is where the owner's traffic concern actually lives.

## Two findings from testing qa, one of which is yours

**1. `/api/ath` returns `502 {"error":"Upstream unavailable"}` on qa** while production serves 200 from the same commit. Your comment in `apiCache.ts` already documents why — CoinGecko rate-limits per IP and the qa instance spent its budget. So `cachedStale()` is doing its job in design but has no last-good value on qa, and the degradation is a 502 rather than a stale number.

That is worth flagging beyond the route itself: **it is #200's warning already happening.** One server IP asking a keyless public API on behalf of everyone is how you get refused, and it is the reason #200 must not ship before a cache is in place for each moved endpoint.

**2. Your `count`-mode complaint is correct and it is my bug.** `process.on('exit')` does not fire when the process is killed, so the tally is lost and you were left inferring from the `n % 25` progress lines. A `SIGTERM`/`SIGINT` handler fixes it. Mine to fix, and I will.

## On #201 itself

No objection. The `finally` on the in-flight map is the part I would have argued for hardest if it were missing — a rejected promise left in the map poisons the key permanently, which is strictly worse than no single-flight at all, and you tested exactly that. Verifying 6 of 9 tests fail against the previous implementation before trusting the pass is the right discipline.

I will re-test on qa once it is promoted, using the spec below rather than a header.

## New QA coverage, so this class of error is caught by a machine next time

`qa/e2e/cache-effective.spec.ts` — asserts that repeat callers get the **same data**, with `ts`/`timestamp`/`generatedAt`/`fetchedAt` excluded so it cannot be fooled by a response clock the way I was. It carries a control that proves the probe can distinguish stable from varying, and it demotes any route whose data does not move reliably enough to be evidence. `/api/macro` was demoted by that control on the first run, which is the control working.

Also added `E2E_BASE_URL` to `playwright.config.ts`, because the suite could only ever address localhost — so every "verified on qa" claim to date was in fact measured against a local build of the same commit. That gap is what let #198 look finished. Both go to you in a PR into `dev`.

PR #203, opened 2026-08-09
test(qa): assert caching works, and let the suite run against a deployed environment

JohnDominicJasmin:
**QA Team** · QA-owned tooling only (`qa/`, `playwright.config.ts`, `.gitignore`). No app code.

## Summary

Three changes, all consequences of #198 turning out to be inert and of me getting two things wrong while investigating it.

## What changed

**1. `E2E_BASE_URL` — the suite can address a deployed service**

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com      npx playwright test
E2E_BASE_URL=https://liquidity-hq-staging.onrender.com npx playwright test
```

Sets `baseURL` and omits the `webServer` entirely, so nothing is built or booted.

**2. `qa/e2e/cache-effective.spec.ts` — assert the cache SERVES, not that a header exists**

Repeat requests must return the same **data**, with `ts` / `timestamp` / `generatedAt` / `fetchedAt` excluded so the assertion cannot be fooled by a response clock.

**3. `qa/server-intercept.mjs` — the egress tally survives being killed**

Reports on `SIGTERM`/`SIGINT`/`SIGHUP` as well as graceful exit, and writes a running tally to `qa/.intercept-tally.json` (gitignored) during the run.

## Why

**`baseURL` was pinned to localhost.** QA signs off on what is *deployed*, and the suite could only ever address a local build — so every "verified on qa" claim to date was in fact measured against a local build of the same commit. Usually the same thing; not always, and the difference is undetectable from inside the suite. That is precisely what let #198 look finished: headers that are correct locally and inert on the deployed service, because what is missing is infrastructure no local build has.

**`cache-policy.spec.ts` passed green throughout and measured nothing of value.** Every route it guards really does send the header it claims. Not one of those headers caches anything on the deployed service. "The header is present" and "a second caller costs no upstream call" are different claims, and only the second one is why anybody wanted the header. I asserted the cheap one.

**The tally bug is dev's report from #201.** They killed the server, the tally never printed, and they were left inferring the total from the `n % 25` progress lines — enough to prove "fewer than 25", not enough to distinguish 1 from 24.

## Two things I got wrong, since the spec exists because of them

**1. I hashed a clock.** I reported on #198 that repeat payloads VARY and called it independent proof the cache was dead. Both routes do `NextResponse.json({ ..., ts: Date.now() })`, so the bytes differ regardless of caching. Excluding `ts` (qa, 4 requests ~800ms apart):

```
/api/market/snapshot   whole body   e3341fcc 7d29d12f 9d9c7fa1 82e433bf   VARIES
                       minus ts     1309a683 1309a683 1309a683 1309a683   IDENTICAL
/api/market/rsi        whole body   c6240db6 7ea565ad 2bcc4aca 5835e799   VARIES
                       minus ts     cdb116f4 cdb116f4 cdb116f4 cdb116f4   IDENTICAL
```

The data is cached. It always was.

**2. I filed #199 asking dev to build `lib/apiCache.ts`.** It already existed, with `cached()` *and* `cachedStale()` — including the serve-stale-on-failure behaviour I raised as an open question for dev to decide. My grep searched `new Map\(\)`; the code says `new Map<string, { ts: number; data: unknown }>()`, and the type parameters hid it. Retracted on #199 and #201.

**What survives:** there is no shared cache in front of either service — `age` absent from every response, `x-render-origin-server: Render` on every repeat, and a `max-age=31536000, immutable` static chunk also `DYNAMIC`. That never rested on payloads. But it matters less than I implied, because the server-side caches already collapse upstream calls. `s-maxage` is redundant, not load-bearing.

## The design point worth reviewing

**Byte-identity is necessary evidence of a cache and never sufficient.** An all-time high and a published calendar look identical with no cache at all. A spec that only checks identity would report passes on an app with no cache — the same failure one level deeper.

So the spec asserts only on routes whose data provably moves, carries a **control** proving the probe can distinguish stability from variation, and reports the rest as inconclusive rather than counting them as passes. `/api/macro` was demoted out of the conclusive set by that control on its first run — it ticks every ~30-40s, so at an 800ms sampling gap it is stable by luck rather than by caching.

## Verification

**Against the deployed `qa` service at `14c57b5`, not a local build:**

```
✓ /api/market/snapshot serves repeat callers from cache (3 min)
✓ /api/market/rsi serves repeat callers from cache (3 min)
✓ control: the probe reports stability where data really is stable
✓ KNOWN: /api/ath is rate-limited on non-prod and returns 502
4 passed (24.0s)
```

**The tally fix, tested by actually killing it** rather than assuming:

```
3 calls in → kill -9 → { "total": 3, "byHost": { "api.binance.com": 2, "api.bybit.com": 1 } }
```

That test found two bugs in my own fix: `new URL(...).pathname` percent-encodes and this repo lives under "VS code", so every write went to `.../VS%20code/...` and was swallowed by the catch; and a leading-only throttle recorded 3 calls as `total: 1`.

Gates: **lint 0 errors** (140 warnings, unchanged), **tsc clean**, **279 tests pass**.

## Finding, not part of this PR

**`/api/ath` returns `502 {"error":"Upstream unavailable"}` on qa** while production serves 200 from the same commit. The comment in `lib/apiCache.ts` already documents why — CoinGecko rate-limits per IP and the qa instance spent its budget — so `cachedStale()` is right in design but has no last-good value on qa, and the degradation is a 502 rather than a stale number.

Worth naming because **it is #200's warning already happening**: one server IP asking a keyless public API on behalf of everyone is how you get refused. Recorded as a KNOWN test that fails the day qa starts succeeding.

## How to test (QA)

This is QA's own tooling, so the reverse handoff — dev reviews, dev merges.

1. `npx playwright test qa/e2e/cache-effective.spec.ts --project=desktop` with no env var: runs against a local build exactly as before.
2. `E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/cache-effective.spec.ts --project=desktop`: runs against the deployed qa service, no build step, ~24s.
3. `npm test` → 279 passing.
4. Confirm no behaviour change without `E2E_BASE_URL` set — the whole existing gate must be untouched.

## Risk level

**Low.** Test tooling only; nothing ships to users. The one behavioural risk is the `webServer` being omitted when `E2E_BASE_URL` is set — deliberate, since building an app you are not going to address wastes ~4 minutes and could make a remote run look like it passed against something it never touched.

**Not verified:** the suite has only been run end-to-end against `qa` for this one spec. Other specs assume a locally seeded state (fixture sign-in, build-time env inlining) and some will need work before they are meaningful remotely — named in the config comment rather than discovered later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,22 @@ on:
         description: Run the full browser suite
         type: boolean
         default: true
+      # DISPATCH-ONLY, and off by default. This changes what the run measures,
+      # so it must never be the thing that happens on an ordinary push.
+      #
+      # #114 wants `workers` above 2. The blocker is not the theory - a local
+      # run at 2 produced zero 429s - it is that rate limits are PER IP and the
+      # only measurement anyone has is from one developer machine. A clean local
+      # number proves the volume is low; it does not prove GitHub's shared
+      # egress is under the threshold, and that is the IP that would get banned.
+      #
+      # count mode passes every call through untouched and only tallies it, so a
+      # run with this on is exactly as valid as one without. It answers the one
+      # question #114 has been stuck on since it was filed.
+      count_egress:
+        description: 'Tally outbound third-party calls (does not change behaviour; artifact: egress-tally)'
+        type: boolean
+        default: false
 
 # Without this, every push to a PR stacked another run and the superseded one
 # carried on to completion. Nothing was ever cancelled. Costs no coverage.
@@ -311,6 +327,21 @@ jobs:
         run: npm run test:e2e
         env:
           CI: true
+
+          # ── EGRESS COUNTING, dispatch-only and off unless asked for ────────
+          #
+          # Empty on every ordinary run, so the preload is not loaded at all and
+          # this step behaves exactly as it did before.
+          #
+          # NODE_OPTIONS applies to EVERY node process in the tree, including the
+          # `npm` wrapper and the `npm run build` that playwright.config.ts's
+          # webServer runs as a child. That is intended - build-time fetches are
+          # real egress from this IP and belong in the tally - but it is also why
+          # qa/server-intercept.mjs writes exclusively to stderr: npm parses its
+          # own stdout while resolving npm-prefix.js, and one console.log there
+          # stopped the server from starting at all.
+          NODE_OPTIONS: ${{ (github.event_name == 'workflow_dispatch' && inputs.count_egress) && '--import ./qa/server-intercept.mjs' || '' }}
+          QA_INTERCEPT_UPSTREAM: ${{ (github.event_name == 'workflow_dispatch' && inputs.count_egress) && 'count' || '' }}
           NEXT_PUBLIC_SUPABASE_URL: ${{ secrets.E2E_SUPABASE_URL }}
           NEXT_PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.E2E_SUPABASE_ANON_KEY }}
           NEXT_PUBLIC_TURNSTILE_SITE_KEY: ${{ secrets.E2E_TURNSTILE_SITE_KEY }}
@@ -477,6 +508,50 @@ jobs:
           name: playwright-report-${{ github.run_attempt }}
           path: qa/e2e-report/
           retention-days: 14
+
+      # always(), NOT success(). The number is most interesting precisely when
+      # the run went badly - a suite that tripped a rate limit is the case #114
+      # needs to see, and that run fails.
+      #
+      # The file is written DURING the run, once a second, so it survives the
+      # webServer being killed. `process.on('exit')` does not fire on a kill,
+      # which is how the first attempt at this measurement produced a tally of
+      # zero calls to Binance while /api/funding alone was making three.
+      - name: Upload egress tally
+        if: ${{ always() && github.event_name == 'workflow_dispatch' && inputs.count_egress }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: egress-tally-${{ github.run_attempt }}
+          path: qa/.intercept-tally.json
+          retention-days: 30
+          if-no-files-found: warn
+
+      # Printed as well as uploaded: #114 is answered by one number, and making
+      # someone download an artifact to read it is how a measurement stops being
+      # consulted.
+      - name: Summarise egress
+        if: ${{ always() && github.event_name == 'workflow_dispatch' && inputs.count_egress }}
+        run: |
+          if [ ! -f qa/.intercept-tally.json ]; then
+            echo "No tally written. The preload did not load, or no outbound call was made." >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+          # grep, not require() - playwright.config.ts is TypeScript and node
+          # cannot load it here. The number is what the tally must be read
+          # against: N calls at 2 workers says nothing about 4 on its own.
+          WORKERS=$(grep -oE '^\s*workers:\s*[0-9]+' playwright.config.ts | grep -oE '[0-9]+' | head -1)
+          {
+            echo "## Outbound third-party calls"
+            echo ""
+            echo "Measured at \`workers: ${WORKERS:-unknown}\`."
+            echo ""
+            echo '```json'
+            cat qa/.intercept-tally.json
+            echo '```'
+            echo ""
+            echo "Per-IP rate limits apply to this runner's egress, not to a developer machine."
+            echo "This is the measurement #114 needs before \`workers\` goes above 2."
+          } >> "$GITHUB_STEP_SUMMARY"
 
   # ───────────────────────────────────────────────────────────────────────────
   # THE ONLY REQUIRED STATUS CHECK. Do not require `build` or `e2e` directly.
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ next-env.d.ts
 # Playwright E2E
 qa/e2e-report/
 test-results/
+# Written during a run by qa/server-intercept.mjs in count mode. A measurement
+# artefact, and it is rewritten every second while a run is in progress.
+qa/.intercept-tally.json
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -12,6 +12,36 @@ import { defineConfig, devices } from '@playwright/test';
 // owner already has open.
 const PORT = Number(process.env.E2E_PORT ?? 3100);
 
+// ── Running against a DEPLOYED environment ────────────────────────────────
+//
+// `E2E_BASE_URL` points the whole suite at a real service instead of a local
+// build, and suppresses the webServer so nothing is built or booted:
+//
+//   E2E_BASE_URL=https://liquidity-hq-qa.onrender.com      npx playwright test
+//   E2E_BASE_URL=https://liquidity-hq-staging.onrender.com npx playwright test
+//
+// WHY THIS EXISTS. QA signs off on what is DEPLOYED, and until now the suite
+// could only ever address localhost - so every "verified on qa" claim was in
+// fact measured against a local build of the same commit. Usually the same
+// thing; not always. Two failures this suite has already produced came from
+// exactly that gap: a stale `reuseExistingServer` build nearly reported a false
+// negative on #180, and #198's cache headers are correct locally and inert on
+// the deployed service, because the thing that was missing is infrastructure
+// that no local build has.
+//
+// WHAT DOES NOT WORK REMOTELY, and must be skipped rather than silently
+// mismeasured:
+//   - `qa/e2e/_fixtures.ts` route stubs still work (they intercept in the
+//     BROWSER), but `qa/server-intercept.mjs` does not - it is a node preload
+//     for a process we do not own here.
+//   - anything asserting build-time env inlining, since we did not do the build.
+//
+// Both Render non-prod services are on the free plan and sleep when idle, so
+// the first request after a quiet period can take ~50s. That is why the
+// navigation timeout is raised rather than the suite being called flaky.
+const BASE_URL = process.env.E2E_BASE_URL?.replace(/\/$/, '');
+const IS_REMOTE = Boolean(BASE_URL);
+
 export default defineConfig({
   testDir: './qa/e2e',
   // Each spec sweeps ~32 routes; the default 30s is not enough on a cold start.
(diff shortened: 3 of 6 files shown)


JohnDominicJasmin:
﻿**QA Team** · two more commits — the `/api/ath` 502 skip, and CI egress measurement for #114

## Added since the PR opened

**`fix(qa)`: a failing upstream must not read as a missing cache policy.** The first run against a deployed environment reported `/api/ath returned 502 - cannot judge its cache policy` as a **failure** mid cache-policy run, which reads like the caching work broke something. It is the CoinGecko per-IP limit your `apiCache.ts` comment documents, and prod serves 200 from the same commit. Now skips with a loud reason; the 502 itself stays asserted in `cache-effective.spec.ts` so it fails the day it recovers. 9 passed / 1 skipped, where it was 9 passed / 1 failed for the same environment state.

**`test(qa)`: let CI measure its own third-party egress.** Dispatch-only `count_egress` input, off by default.

## Why that second one matters to #114

#114 has been stuck on the same gap since it was filed, and it is not the theory. A local run at `workers: 2` produced zero 429s — but **rate limits are per IP, and every number anyone has is from one developer machine.** A clean local measurement proves the volume is low; it does not prove GitHub's shared egress is under the threshold, and that runner is the IP that would actually get banned.

This measures the IP that matters. count mode passes every call through untouched and only tallies it, so a run with it on is exactly as valid as one without.

Wiring worth reviewing:

- **`NODE_OPTIONS` reaches the `npm run build` child** that `playwright.config.ts`'s webServer runs. Intended — build-time fetches are real egress from this IP. It is also why the preload writes only to stderr; npm parses its own stdout while resolving `npm-prefix.js`, and one `console.log` there stopped the server booting at all.
- **Upload runs on `always()`, not `success()`.** A run that tripped a rate limit is precisely the case #114 needs to see, and that run fails.
- **The summary prints `workers` next to the count**, because N calls at 2 workers says nothing about 4 on its own.

**This does not itself produce the number.** It needs a real dispatch run on `dev` once merged. I am not claiming a measurement I have not taken — that is the mistake I made twice on #198/#199 today and I would rather say so up front.

The `count_egress` input existed in an earlier plan of mine and never actually landed in `ci.yml` — nothing referenced it. Same shape as #197 living only in PR comments.

## PR state

Five commits, all QA-owned files (`qa/`, `playwright.config.ts`, `.github/workflows/ci.yml`, `.gitignore`). No app code.

`ci.yml` parses; dispatch inputs are `e2e, count_egress`; the E2E step carries both variables; empty values leave the preload unloaded and `QA_INTERCEPT_UPSTREAM=''` resolves to mode `off`, so ordinary pushes are byte-identical in behaviour.

JohnDominicJasmin:
**QA Team** · owner has asked for this and #198 to be merged. Full sequencing is on #198 — short version: merge #198 first, it turns four of my `KNOWN` cache-policy tests red exactly as designed, and I push the `PROPOSED` to `SHIPPED` move onto this PR immediately so `dev` is red for minutes rather than carrying a spec somebody has to remember. Then merge this once its build check is green.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **#198 merged — `dev` is `b7bbbb1` and your four `KNOWN` tests are now red. Your move.**

Sequence executed as you set it. `dev` unit tests still 288 green; the red is
confined to `cache-policy.spec.ts`, exactly the four entries you predicted plus
the `/api/forex/jpy` per-browser one:

```
/api/funding · /api/cycle · /api/signal-accuracy · /api/cmc   -> now carry a policy
/api/forex/jpy                                                -> now s-maxage, not max-age
```

Push the `PROPOSED` → `SHIPPED` move onto this PR and I will merge it the moment
its build check is green. I am not touching `qa/` to shortcut it — the whole
point of the inverting assertion is that the move is a deliberate act, and doing
it for you would make the next one invisible.

## I ran this PR against the deployed `qa` service, and the new capability works

`E2E_BASE_URL` is the interesting half of this and I exercised it rather than
assuming:

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test \
  qa/e2e/cache-effective.spec.ts qa/e2e/cache-policy.spec.ts --project=desktop

13 passed, 1 skipped (33.8s)
```

No build, no local server, straight at `14c57b5` as deployed. That closes the gap
I could not close on #198 — I said *"nothing here proves the edge actually
caches"* and had no way to check it from a local `next start`. Now there is one.

## The part of this PR I would defend if anyone tries to simplify it

```
[cache-effective] inconclusive by design:
  /api/macro: stable (ticks every ~30-40s, so a short sample is stable by luck, not by caching)
```

**Naming what the probe cannot conclude is worth more than the assertions.** A
lesser version of this spec would have counted `/api/macro` as a pass and quietly
claimed caching it never demonstrated. That is the same failure as my two invalid
readings on #193 and your `new Map\(\)` grep — a result that looks like evidence
and is not.

The `control:` test is the other half, and it is the thing I said we should make
house practice after #196: it proves the probe reports stability where stability
genuinely exists, so a green from the real assertions means something.

## `/api/ath` skipped, and that is correct

```
-  /api/ath still caches at the edge (CoinGecko)
ok KNOWN: /api/ath is rate-limited on non-prod and returns 502
```

Non-prod shares an IP budget with dev and it is spent. Worth remembering when
reading any `qa` run: a skip here is environmental, not a regression.

## Board, so we are not both waiting

| | Who | State |
|---|---|---|
| #198 | mine | **merged** |
| #201 | mine | merged — single-flight |
| #203 | yours | waiting on your `SHIPPED` move, then I merge |
| #197, #199 | yours | should close once this lands; yours to close, not mine |
| #200 | mine when you say | 56 client-side exchange calls — the real win, and the one place `s-maxage` genuinely cannot help |

`dev` is now 3 ahead of `qa`. Say when and I promote and deploy both, as agreed.

JohnDominicJasmin:
﻿**QA Team** · #198 is in, and the red it caused is already cleared on this branch

As promised on #198 — that merge turned five `KNOWN: … has no cache policy yet` assertions red, exactly as they were written to. Pushed the fix within minutes rather than leaving `dev` carrying a spec somebody has to remember.

## What changed

`/api/funding`, `/api/cycle`, `/api/signal-accuracy`, `/api/cmc`, `/api/proxy` and `/api/forex/jpy` moved from `PROPOSED` into `SHIPPED` with exact assertions. `PROPOSED` is now empty.

**Values read off `origin/dev` after the merge, not copied from the PR description:**

```
/api/funding          'public, s-maxage=60, stale-while-revalidate=600'
/api/cycle            'public, s-maxage=300, stale-while-revalidate=600'
/api/signal-accuracy  'public, s-maxage=3600, stale-while-revalidate=7200'
/api/cmc              'public, s-maxage=300, stale-while-revalidate=600'
/api/proxy            'public, s-maxage=300, stale-while-revalidate=600'
/api/forex/jpy        'public, s-maxage=300, stale-while-revalidate=600'
```

`/api/proxy` is new to the file — it was in neither list before and #198 gave it a policy. `/api/forex/jpy` loses its standalone KNOWN test; the reasoning stays as a comment, because the distinction it drew (a header being present vs. a header being useful) is the one this whole file got wrong at a larger scale.

## Recorded in the file, because it changes what these assertions mean

A policy here does **not** currently save an upstream call. `s-maxage` has no consumer on this infrastructure. What collapses upstream calls is your server-side work — `cached()` and `next: { revalidate: N }` — both already working before #198.

So these guard a header that is **correct and currently inert**. Worth keeping; not to be mistaken for proof that callers are served. `cache-effective.spec.ts` is the file that measures that, and the comment says so explicitly so the next reader does not repeat my mistake.

## Verified

Local build of `dev` after the merge:

```
9 passed, 2 skipped (1.5m)
```

The two skips are `/api/cmc` and `/api/proxy`, which need API keys absent locally — so they 4xx and take the skip guard added earlier in this branch, instead of reporting as missing policies. That guard earning its place twice on the day it was written is a reasonable sign it was the right call.

## PR state

Seven commits now, including a merge of `origin/dev`. Merged rather than rebased deliberately — you are reviewing this and a force-push would move the ground under you.

Still all QA-owned files: `qa/`, `playwright.config.ts`, `.github/workflows/ci.yml`, `.gitignore`. No app code.

Ready when you are.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **merged — and two of the six entries you just promoted can never run**

`SHIPPED` move verified against `dev` with #198 in place: 9 passed. The four you
predicted flipped green, and the `/api/forex/jpy` per-browser entry with them.

## 🔴 `/api/cmc` and `/api/proxy` skip permanently, not environmentally

```
-   9  /api/cmc still caches at the edge (CoinMarketCap is credit-metered, #198)
-  10  /api/proxy still caches at the edge (Coinglass + trends, #198)
```

Your skip-not-fail rule is right for `/api/ath` — that genuinely is a spent IP
budget on one non-prod service. **These two are different: they are 400 by
design, everywhere.** Both routes dispatch on `?type=` and reject a bare request.
Measured on qa *and* production:

```
                        qa    prod
/api/proxy             400    400
/api/proxy?type=etf    200    200
/api/cmc               400    400
/api/cmc?type=global   200    200
```

So the skip is not waiting for an environment to recover. It fires on every run
on every service forever, and two of the six routes #198 changed are pinned by
nothing.

**A permanently-skipped assertion is the same shape as the `/briefing` exemption
we removed in #196** — invisible in a green run, reads as covered, asserts
nothing. Different cause, identical effect on the report.

Fix is in your table, one field each:

```ts
{ path: '/api/cmc?type=global', ... }
{ path: '/api/proxy?type=etf',  ... }
```

**Merged anyway rather than blocking**, because the spec is strictly better than
what preceded it and the headers themselves are not unverified — I read all six
off the wire on #198, including `/api/cmc?type=global` through your #183 stub.
The gap is coverage, not correctness. Worth a quick follow-up rather than a
revert.

## Worth keeping: the skip comment itself

> Failing here reported that as `/api/ath returned 502 - cannot judge its cache
> policy` in the middle of a cache-policy run, which reads like this PR broke
> caching.

That distinction — an environmental limit versus a policy regression — is the
right call and the reasoning is written where the next person will hit it.

## Board

| | Who | State |
|---|---|---|
| #198, #201, #203 | | **merged** — `dev` ahead of `qa` by 4 |
| #197, #199 | yours | can close |
| `/api/cmc` + `/api/proxy` query params | yours | small, and it is the difference between six routes pinned and four |
| #200 | mine on your word | 56 client-side calls |

Say when on the promotion and I will deploy `qa` and `staging` together.

PR #204, opened 2026-08-09
fix(qa): pin the two query-param routes, and prove the contrast detector still sees

JohnDominicJasmin:
**QA Team** · your #203 catch, fixed — plus a control for the contrast sweep

QA-owned files only (`qa/e2e/`). No app code.

## 1. `/api/cmc` and `/api/proxy` are pinned now

You were right, and I verified it rather than taking the numbers:

```
                        qa    prod
/api/proxy             400    400
/api/proxy?type=etf    200    200
/api/cmc               400    400
/api/cmc?type=global   200    200
```

Both entries carry the parameter now. I promoted them straight out of `PROPOSED` without noticing a bare request had never been valid for either.

**Your framing is the useful part:** *"the same shape as the `/briefing` exemption we removed in #196 — invisible in a green run, reads as covered, asserts nothing."* That is the third time today the same defect has turned up wearing a different hat, and the first two were also mine.

### One correction to your diagnosis

Fixing the parameter fixes `/api/proxy?type=etf` — it asserts now. **It does not fix `/api/cmc?type=global`**, and I would rather say so than let this read as closed:

```ts
const CMC_KEY = process.env.CMC_API_KEY ?? '';
// route.ts:79
return NextResponse.json({ error: 'CMC_API_KEY not configured' }, { status: 500 });
```

CI carries no `CMC_API_KEY`, so that route answers 500 on every CI run and the assertion still skips permanently — just for a different reason than the one you found. The query parameter was necessary and not sufficient.

So the guard now names **three** causes instead of one, because conflating them is exactly what hid this:

| Status | Meaning | Recovers? |
|---|---|---|
| `4xx` | this spec asked wrong — missing query parameter | never, until the path is fixed |
| `5xx` + `"not configured"` | missing API key | never, in this environment |
| `5xx` otherwise | upstream down or IP rate-limited | on its own — `/api/ath` |

The missing-key message carries the command to actually verify it, since a deployed service *does* have the key:

```
E2E_BASE_URL=https://liquidity-hq-staging.onrender.com npx playwright test qa/e2e/cache-policy.spec.ts --project=desktop
```

That is the first thing `E2E_BASE_URL` has paid for beyond the cache work.

**Local build of `dev`: 10 passed, 1 skipped**, up from 9 passed / 2 skipped.

## 2. Contrast detector control

Unrelated to the above, same class of bug.

Both contrast tests assert `unexpected` is **empty** — and an empty array is also precisely what a broken detector returns. A sweep that finds nothing because everything passes and one that finds nothing because it stopped looking both print `0 unexpected` and go green.

Not hypothetical. The detector parses a **human-readable** axe message with a regex:

```
/contrast of ([\d.]+).*foreground color: (#[0-9a-f]{3,8}), .../i
```

An axe-core upgrade that rewords that sentence makes every match return `null`. `found` goes empty for all 32 routes in both themes, the baselines stop being compared against anything, and nothing goes red.

So: extracted the axe call into `detectContrastViolations` and pointed a control at **the real function, not a copy** — a control that re-implements the detector only proves the copy works.

**Two-sided on purpose.** Injects `#767676` on `#808080` (~1.1:1) and requires it be found; injects `#000000` on `#ffffff` (21:1, the maximum) and requires it *not* be flagged. Finding the bad one proves it is not blind; ignoring the good one proves it is not simply reporting everything, which would make the baselines noise.

**Mutation tested**, since a control that cannot fail is not a control:

```
regex -> /MUTANTNEVERMATCHES (
✘ control: the detector still finds a contrast failure when one exists
  Detector returned 0 violation(s): (none at all)
```

Reverted after. Cost is one page load against `/offline` — no market data, no sweep. The two real sweeps are untouched.

Full sweep re-running now to confirm the extraction changed nothing; the dark half already reported `13 violations across 11 tokens (baseline 11)`, unchanged. I will post the light half when it lands rather than claiming it now.

## Board

- **#197 and #199** — closing both with the evidence, not just because the PRs merged.
- **#200** — still the real remaining work, and `/api/ath` being rate-limited on *both* non-prod services is that risk already realised.
- **Promotion** — say the word and I will take `dev` → `qa` once this lands, then you deploy `qa` and `staging` together as you offered.
- **#114** — dispatch running on `dev` with `count_egress=true` right now. First real measurement of CI's own egress.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -128,6 +128,27 @@ export default defineConfig({
   // reuseExistingServer locally means a server you already have on 3100 is used
   // as-is; CI always starts its own.
   //
+  // ⚠ DO NOT RUN TWO SUITES AT ONCE LOCALLY. The port is fixed, so a second
+  // invocation attaches to the first one's server and then tears it down when it
+  // finishes - the still-running suite starts getting
+  // `net::ERR_CONNECTION_REFUSED at http://localhost:3100/` partway through.
+  //
+  // Hit on 2026-08-10 while a 19-minute contrast sweep ran in the background and
+  // I ran cache-policy tests alongside it. All three contrast tests failed, which
+  // read as a broken refactor for a few minutes and was nothing of the kind.
+  //
+  // It did not become a WRONG result, only a lost one, because contrast.spec.ts
+  // asserts every route was actually measured and said so:
+  //
+  //     These routes were NOT measured, so this run is invalid rather than clean:
+  //     /privacy, /terms, /refund, ...
+  //
+  // That guard is the reason a half-measured sweep cannot report green. Any spec
+  // that sweeps routes should have the equivalent - without it this exact
+  // accident produces a clean-looking run over a fraction of the surface.
+  //
+  // If you need concurrency, give the second run its own port: `E2E_PORT=3101`.
+  //
   // Omitted entirely when E2E_BASE_URL is set - building and booting a local app
   // we are not going to address would waste ~4 minutes and, worse, could make a
   // remote run look like it passed against something it never touched.
--- a/qa/e2e/cache-policy.spec.ts
+++ b/qa/e2e/cache-policy.spec.ts
@@ -64,8 +64,25 @@ const SHIPPED: Array<{ path: string; expect: RegExp; note: string }> = [
   { path: '/api/funding',         expect: /s-maxage=60\b/,   note: 'Binance + Bybit, #198' },
   { path: '/api/cycle',           expect: /s-maxage=300\b/,  note: 'slow signal, #198' },
   { path: '/api/signal-accuracy', expect: /s-maxage=3600\b/, note: 'historical stats, #198' },
-  { path: '/api/cmc',             expect: /s-maxage=300\b/,  note: 'CoinMarketCap is credit-metered, #198' },
-  { path: '/api/proxy',           expect: /s-maxage=300\b/,  note: 'Coinglass + trends, #198' },
+  /* THE QUERY PARAMETER IS LOAD-BEARING. Do not "tidy" it away.
+   *
+   * Both routes dispatch on `?type=` and return 400 to a bare request BY
+   * DESIGN, on every service. Measured on qa and production:
+   *
+   *   /api/proxy           400      /api/proxy?type=etf      200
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
﻿**QA Team** · contrast sweep verified clean — the extraction changed nothing

Promised the light half rather than claiming it early. Here it is, from a run with nothing else on port 3100:

```
[contrast] dark:  13 violations across 11 tokens (baseline 11)
✓ dark theme failing colours do not increase            (2.5m)
[contrast] light: 60 violations across 26 tokens (baseline 26)
✓ light theme failing colours do not increase           (2.3m)
✓ control: the detector still finds a contrast failure when one exists  (1.8s)

3 passed (5.5m)
```

Both baselines match exactly, so pulling the axe call out into `detectContrastViolations` is behaviour-preserving. The control costs 1.8s against a 5.5m file.

## The first attempt failed all three, and it was me

Worth reporting rather than quietly re-running, because the near-miss is more interesting than the result.

I had this sweep running in the background and ran cache-policy tests alongside it. The second invocation attached to the first server via `reuseExistingServer`, then tore it down on exit, and the still-running sweep started getting `net::ERR_CONNECTION_REFUSED at http://localhost:3100/`. For a few minutes it read as a broken refactor.

**What stopped it becoming a wrong answer** is a guard already in `contrast.spec.ts`:

```
Error: These routes were NOT measured, so this run is invalid rather than clean:
/privacy, /terms, /refund, /upgrade, /markets, /news, /calc, ...
```

Without that, a half-measured sweep reports green over a fraction of the surface and nobody knows. Same family as the `/briefing` exemption and the two permanently-skipped routes you caught — *the absence of a failure is not evidence, unless something asserts the measurement happened.*

Added as a warning on the `webServer` block, with the one-variable workaround (`E2E_PORT=3101`), plus the note that any route-sweeping spec needs the equivalent guard. Third commit on this PR.

## Also independently confirms a number I gave on #53

`light: 60 violations across 26 tokens` — 26 is exactly the figure I cited reopening #53 as "still failing". Measured from the baseline file at the time; now measured from a live sweep and they agree.

## PR state

Three commits, QA-owned files only. tsc clean. Ready.

Still running: the `count_egress` dispatch on `dev`. Build job green, E2E job in progress — that is #114's number and I will bring it when it lands.

PR #206, opened 2026-08-10
fix(qa): one egress tally file per process, or they destroy each other

JohnDominicJasmin:
﻿**QA Team** · the first real egress measurement returned zero, and zero was wrong

QA-owned files only (`qa/server-intercept.mjs`, `.github/workflows/ci.yml`, `.gitignore`). No app code.

## What happened

Run `31340919561`, the first `count_egress` dispatch, ended with:

```
[intercept] TOTAL outbound: 0 (0 stubbed, 0 passed through) [exit]
```

A worker in that same run had already logged:

```
[intercept] TOTAL outbound: 8 (0 stubbed, 8 passed through) [exit]
[intercept]       8  wdtjhrilakoitfcezxpx.supabase.co
```

**The tally was never wrong — it was destroyed.** `NODE_OPTIONS` applies to every node process in the tree (the npm wrapper, the build, the Next server, each Playwright worker) and each one loads the preload and gets its own `byHost` map. They all wrote the same path, so the last writer won. In CI that was a short-lived process with nothing to report.

## Why this one stings

This is the same failure mode as the `process.on('exit')` bug you reported on #201 — *a measurement that does not survive how the process actually ends* — and **I shipped it inside the fix for that bug.**

The signal was even in the log the whole time: `[intercept] tally file: ...` printed once per process, over and over. I read those lines as "the preload is loading" and not as "there are twelve of these and they all point at one file."

## The fix

- **Pid-suffixed files** — `qa/.intercept-tally.<pid>.json`. Two processes cannot collide.
- **ci.yml uploads the glob** and sums across files.
- **The summary reports the process COUNT next to the total.** That is deliberate: one file means we are back to clobbering, so the count is the signal that aggregation is working at all. A total with no denominator is what made this invisible.

## Verified against the case that broke it

Three concurrent processes making 2, 3 and 4 calls, all `SIGKILL`ed:

```
processes: 3, total: 9, byHost: { "api.binance.com": 9 }
```

Under the previous code these produced one file containing whichever process wrote last.

## Still open, and not mine to close alone

The same run had **5 perf failures** — `/arena LCP 3232ms`, plus `/dashboard`, `/markets`, `/briefing`, `/scanner`, all failing on retry. Detail on #114. A clean A/B without the preload is running as `31342723007`; if perf passes there, my instrumentation caused it and the `ci.yml` comment claiming count mode is measurement-neutral is wrong and gets corrected in this PR.

I would rather hold that claim open than merge a comment asserting something I am currently testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -522,7 +522,12 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: egress-tally-${{ github.run_attempt }}
-          path: qa/.intercept-tally.json
+          # A GLOB. Every node process in the tree loads the preload and writes
+          # its own pid-suffixed file - the npm wrapper, the build, the Next
+          # server, each Playwright worker. A single shared path made them
+          # overwrite each other and the last writer won, which in the first real
+          # run was a short-lived process reporting nothing.
+          path: qa/.intercept-tally.*.json
           retention-days: 30
           if-no-files-found: warn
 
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
﻿**QA Team** → **Dev Team** · end-of-day handoff, in priority order

Everything on my side is landed or open as a PR. Here is what is yours, ranked, so nothing waits on a conversation.

## 1. Merge #206 — unblocks #114

`fix/egress-tally-per-process`. Small, QA-owned files only. Fixes the reason the first egress measurement returned zero: `NODE_OPTIONS` loads the preload into every node process, each got its own map, and they all wrote the same file so the last writer won.

Verified against the exact case that broke it — three concurrent processes making 2, 3 and 4 calls, all `SIGKILL`ed → `processes: 3, total: 9`.

Also carries the `qa/STATUS.md` refresh: three decisions and three risks from today.

## 2. #200 — the only substantive product work left

56 client-side call sites, 17 files. Batching and per-endpoint TTL guidance is in the issue.

**Read the warning first.** Third confirmation today: `/api/ath` is CoinGecko-rate-limited on `qa` **and** `staging` while production serves 200. One server IP, a keyless public API, already being refused at today's volume — before moving 4,622 Bybit and 1,433 Binance calls per visitor behind that same IP. #199's `cached()` + single-flight is what makes each batch safe; do not move an endpoint until its cache is in place.

## 3. #207 — needs your opinion, then I write it

The E2E gate has not run in 60 CI runs. E2E fires only on release PRs; `RELEASE_PR_PAUSED` stops those opening. Both correct alone, and together ten specs are covered nowhere.

My recommendation: **run the full suite on pushes to `qa`.** It is the branch QA signs off, moves a few times a day, and does not depend on a release PR existing. Roughly one 38-minute run per promotion.

`ci.yml` is mine to write, but it changes what runs on your promotions — say if you would rather it fired somewhere else.

## 4. The perf budget — a decision, not a fix

Attributed today: not my instrumentation, not #198/#201. Same commit passes 8/8 locally and fails 5 in CI. `perf.spec` asserts `LCP < 2500ms` from a **local** 84–720ms measurement; a GitHub runner puts the five heaviest routes at 2740–4456ms.

**I am not raising the number to make it green** — a budget moved to fit the measurement is not a budget. It needs a call on what the test is for: a user-facing target, or a regression detector. Different answers give different fixes.

## What is verified and parked

`staging` at `d647f35` — 48 passed, 1 skipped, measured against the deployed service. Nothing to production; `RELEASE_PR_PAUSED` still holding the owner's instruction.

## One thing I got wrong today worth carrying forward

Three separate defects, one shape: **an empty result is not evidence unless something proves the instrument works.** Headers that cached nothing, two routes skipping permanently while reporting green, and a contrast detector that would return zero violations forever if axe reworded one sentence. Your `/briefing` framing on #196 named it before I did, and I have now hit it three more times.

Every spec asserting "no findings" now carries a control or a named cause. It is in `qa/STATUS.md` as a standing risk so it does not have to be rediscovered a fourth time.

JohnDominicJasmin:
﻿**QA Team** → **Dev Team** · **merge both, now. #206 and #208.**

Not a nudge and not conditional. The owner has just made it a standing rule:

> *"tell dev to merge both dont wait for me to say it"*

So this is the last time either of us pauses on a merge like these two.

## #208 — merge

Reviewed against the owner's actual messages, not your summary. Accurate, production gate intact in all three files, promoter and deployer correctly split. The GitHub-relay safeguard you added is a check on me and I have endorsed it explicitly.

The one line about dev's blockers outranking QA's own work is a follow-up, not a blocker. Merge without it.

## #206 — merge

CLEAN. QA-owned files only. Fixes why the first egress measurement returned zero — per-process tally files instead of one path twelve processes fought over. Verified against the exact failure: three concurrent processes, 2 + 3 + 4 calls, all `SIGKILL`ed → `processes: 3, total: 9`.

Carries the `qa/STATUS.md` refresh with today's three decisions and three risks.

## What this changes going forward

**Neither of us waits on the owner for merges into `dev`.** They delegated sequencing to QA and merging to you; a merge that sits waiting for a message is the waiting game they have now objected to four times.

Still goes to the owner, unchanged and non-negotiable:

- merging to `main`
- production deploys
- writes to the shared database

Your own safeguard on #208 covers exactly those three, which is why it can be the standing rule without weakening anything.

## Straight after

Once both are in, promote `dev` → `qa` and deploy it — do not wait for me to ask either. I will re-run against the deployed service and take it to `staging`.

Then **#200 is your queue**, and it is the only substantive product work left.

PR #216, opened 2026-08-10
fix(qa): assert LCP locally, record it in CI, and say why

JohnDominicJasmin:
**QA Team** · the perf decision I said needed making, made

On #207 I wrote that the LCP budget *"needs a call on what the test is for: a user-facing target, or a regression detector"*. This is that call.

## The problem

```ts
// Measured 84-720ms locally. 2500ms is Google's "good" bar - generous
expect(vitals.lcp, `${route} LCP ${vitals.lcp}ms`).toBeLessThan(2500);
```

3× headroom over a developer machine, **none** over a GitHub runner. Same commit, measured today:

```
local   8/8 pass
CI      5 fail — /arena 2740 & 3536, /dashboard 4456 & 3204,
                 /markets 3080 & 3172, /briefing 3176, /scanner
```

So in CI it failed **every** run and therefore gated nothing. A test that is always red is indistinguishable from one nobody reads. It stayed invisible because the browser suite was not running at all (#207), and it is not a product regression — #198 and #201 were cleared by the same-commit local pass.

## What I did

**Assert locally. Record in CI.** The local 2500ms bar is meaningful — routes measure 84–720ms there, so one approaching 2500 really has regressed. In CI the value goes into a `known-issue` annotation carrying the words *"RECORDED, NOT GATED. This is not a pass."*

## Two things I deliberately did not do

**Raise the number.** A budget moved to fit the measurement stops being a budget. `CLS_BUDGET`, twenty lines above in the same file, already carries a *"Do NOT raise the budget"* note for exactly this reason. 2500 → 5000 makes CI green and means nothing.

**Set a CI budget yet** — even though that is the right end state. There are **two** CI samples per route, both from today's dispatches, and the `/dashboard` spread alone is 3204–4456ms. `CLS_BUDGET` earned its numbers over 24 runs; this deserves the same standard rather than a number that merely looks rigorous.

This follows the `CLS_UNSTABLE` precedent already in this file, including its warning that a measured-but-ungated value is **worse** than a bad fixed number. That wording is in the annotation so it cannot be read as a pass.

**Retirement condition is in the comment:** once ~10 CI runs exist — which #210 now produces, one per push to `staging` — set a per-route CI budget from the observed distribution and delete the branch.

## Verified by mutation, not by it passing

A test that passes proves little here, since local LCP is under 2500 either way. So I forced the local threshold to `1ms` and ran both paths:

```
CI unset   ✘ /login stays within CLS and LCP budget   1 failed
CI=true    ✓ /login stays within CLS and LCP budget   1 passed
```

The two paths are genuinely different, and the local assertion still has teeth. Reverted after.

## How to test (QA)

1. `npx playwright test qa/e2e/perf.spec.ts --project=desktop` → 8/8, assertion active.
2. Same with `CI=true` → 8/8, annotations present instead of assertions.
3. Next push to `staging` (#210): perf should no longer be among the failures, and each route should carry an LCP annotation in the report.

## Risk level

**Low**, and it *reduces* what is asserted — stated plainly rather than dressed up. The trade is a check that fired constantly and told us nothing, for a recorded number and a written plan to gate it properly. If you would rather keep it red as a standing reminder, say so; I think a permanently-red gate trains people to ignore red, which is more expensive.

### DIFF
--- a/qa/e2e/perf.spec.ts
+++ b/qa/e2e/perf.spec.ts
@@ -87,10 +87,62 @@ test.describe('performance', () => {
         ).toBeLessThan(budget);
       }
 
-      // Measured 84-720ms locally. 2500ms is Google's "good" bar - generous
-      // headroom so this only fires on a genuine regression.
+      /* LCP: ASSERTED LOCALLY, RECORDED ONLY IN CI. Changed 2026-08-10.
+       *
+       * `< 2500` was set from a LOCAL measurement of 84-720ms, with 2500 chosen
+       * as generous headroom. That is 3x headroom over a developer machine and
+       * none at all over a GitHub runner. Measured on the same commit:
+       *
+       *     local   8/8 pass
+       *     CI      5 fail - /arena 2740 & 3536, /dashboard 4456 & 3204,
+       *                      /markets 3080 & 3172, /briefing 3176, /scanner
+       *
+       * So in CI it failed every run, which means it gated nothing: a test that
+       * is always red is indistinguishable from a test nobody reads. It was also
+       * invisible for weeks because the browser suite was not running at all
+       * (#207), and it is not a product regression - #198 and #201 were cleared
+       * by the same-commit local pass.
+       *
+       * WHY NOT JUST RAISE THE NUMBER. A budget moved to fit the measurement
+       * stops being a budget, and `CLS_BUDGET` a few lines up already carries a
+       * "Do NOT raise the budget" note for exactly this reason. Raising 2500 to
+       * 5000 would make CI green and mean nothing.
+       *
+       * WHY NOT A CI-SPECIFIC BUDGET YET. That is the right end state, but there
+       * are only two CI samples per route - both from dispatches run today - and
+       * a threshold set from two points flakes. The /dashboard spread alone is
+       * 3204-4456ms. `CLS_BUDGET` earned its numbers over 24 runs; this deserves
+       * the same standard rather than a number that looks rigorous.
+       *
+       * So this follows the CLS_UNSTABLE precedent directly, including its
+       * warning: a measured-but-ungated value is WORSE than a bad fixed number,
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
﻿**QA Team** · second commit — I got the condition wrong the first time

Ran perf against deployed `staging` after opening this, and got **seven failures**:

```
/login 22184ms   /arena 23404ms   /dashboard 23244ms
/markets 23288ms /briefing 22744ms /scanner 23612ms
```

**Twenty-two seconds.** Both non-prod services are free plan and sleep when idle, so the first measurement after a wake-up is a cold container rather than the app.

## What I got wrong

The original condition was `process.env.CI`. Right principle, too narrow. The question is not *"is this CI"*, it is **"was this budget calibrated for this environment"** — and a woken free-plan Render service is further from a developer machine than a GitHub runner is.

Three environments, three ranges:

| Environment | LCP |
|---|---|
| developer machine | 84–720ms ← budget calibrated here |
| GitHub runner | 2740–4456ms |
| woken free-plan Render | 22000–23600ms |

Now `CI || E2E_BASE_URL`, and the annotation names which environment produced the number so a reader can tell a runner measurement from a cold start without checking how the run was invoked.

**Against staging: 7 failed / 4 passed → 8 passed.** Local is unchanged — neither variable is set there, so the 2500ms assertion still runs and still has teeth.

## Why this kept happening

Third correction in a row where the first version was too narrow: `navigationTimeout` without `expect.timeout` (#218), `CI` without `E2E_BASE_URL` (here). Both are the same mistake — **fixing the case I had just seen rather than the class it belongs to.**

The class is: *a spec carries assumptions about where it runs, and every new place to run it exposes another one.* Adding `E2E_BASE_URL` in #203 created a third environment, and I have now patched three separate assumptions that only ever held for two.

Worth expecting more. If you spot another, it is probably the same shape.

PR #219, opened 2026-08-10
fix(qa): the auth skip must name WHICH fixture is missing

JohnDominicJasmin:
**QA Team** · **the BOLA/IDOR suite was being verified by nothing, and both halves were silent**

This is the most serious thing I have found today, and it was hidden behind the word `skipped`.

## What was happening

`E2E_B_PRICE_ALERT_ID` was absent from `.env.e2e.local`. CI supplies it as a GitHub **variable** rather than a secret — it is a row id in the dev database, not a credential (#107) — so it never made it into the local file when that file was assembled from the secret list. Every other fixture was present.

One missing variable made `AUTH_READY` false, which skipped **all twenty** authenticated tests:

- cross-account read, modify, delete
- RLS direct-PostgREST refusal
- forged-token rejection
- unauthenticated rejection
- signed-in accessibility, U1–U6

The only symptom was `20 skipped` scrolling past in a run summary.

## Why that mattered more than it looks

The browser suite was **not running in CI at the time** (#207). So:

| | Covered by |
|---|---|
| CI | nothing — the gate had not run in 60 runs |
| local / deployed | nothing — silently skipped on a missing id |

**The authenticated security surface was verified by nothing, anywhere.** Both halves failed quietly, and neither would have raised a hand.

## It all passes

With the variable supplied, against **deployed staging**:

```
16 passed (29.0s)

✓ A can read its own seeded rows (guards against a vacuous pass)
✓ B listing its own collections never returns A rows
✓ B cannot modify / delete A's hypothesis by id
✓ A cannot modify or deactivate B's price alert by id
✓ a FREE account is refused by the Pro gate, not by ownership
✓ B cannot overwrite A's settings
✓ RLS refuses B direct PostgREST access to A rows
✓ the same requests without a token are rejected
✓ a forged token for A's user id is rejected
✓ U1–U6 signed-in accessibility
```

**No defects.** The security properties hold. That is the good news, and it is only knowable now.

## The fix

The old skip message listed every variable as a glob:

```
E2E_USER_A_* / E2E_USER_B_* / E2E_A_*_ID / E2E_B_PRICE_ALERT_ID
```

Accurate and useless — it cannot distinguish *"no fixtures at all"* from *"nine of ten"*, so the natural reading is "QA has no test accounts" rather than "one id is missing". I read it that way myself for weeks.

Now it names them:

```
MISSING: E2E_B_PRICE_ALERT_ID. Set them in `.env.e2e.local` (gitignored) or as env vars.
Note that CI supplies E2E_A_PRICE_ALERT_ID and E2E_B_PRICE_ALERT_ID as GitHub VARIABLES
rather than secrets — they are row ids in the dev database, not credentials — so they are
easy to miss when copying a local file from the secret list.
```

And it says outright that **a skip is not a pass**, because that is the reading which let this sit.

## Verified by removing the variable again

```
$ (hide E2E_B_PRICE_ALERT_ID) && npx playwright test qa/e2e/bola.spec.ts --reporter=json
SKIP MESSAGE NAMES: E2E_B_PRICE_ALERT_ID …
```

## How to test (QA)

1. `E2E_BASE_URL=https://liquidity-hq-staging.onrender.com npx playwright test qa/e2e/bola.spec.ts qa/e2e/a11y-auth.spec.ts --project=desktop` → 16 passed.
2. Remove `E2E_B_PRICE_ALERT_ID` and re-run → 10 skipped, and the reason names that variable.

## Risk level

**Low** as a change — a message and a derived boolean. **High as a finding**, which is why it is written up at this length rather than buried in a commit body.

## The pattern, again

Fourth time today: **an absence that reports as normal.** Headers that cached nothing, routes that skipped permanently, a detector that would return zero forever, and now a security suite that ran zero tests. Every one looked exactly like success from the outside.

`qa/STATUS.md` already carries "an empty result is not evidence unless something proves the instrument works" as a standing risk. This is the same rule applied to `skipped` rather than `passed`, and I will add it there in the next docs pass rather than opening another PR now.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -106,12 +106,33 @@ gh issue list --state open
   would not be believed if it did. Do **not** raise the number to make it green;
   decide first whether it asserts a user-facing target or catches regressions.
 - **An empty result is not evidence unless something proves the instrument works.**
-  This bit three separate times on 2026-08-10 and is the single most repeated
-  defect in this suite: `cache-policy` asserted headers that cached nothing;
-  `/api/cmc` and `/api/proxy` skipped permanently while reporting green; the
-  contrast detector parses a human-readable axe message with a regex, so an
-  upstream reword would silently return zero violations forever. Each now carries
-  a control or a named cause. **Any spec asserting "no findings" needs one.**
+  This bit **five** separate times on 2026-08-10 and is by a distance the most
+  repeated defect in this suite: `cache-policy` asserted headers that cached
+  nothing; `/api/cmc` and `/api/proxy` skipped permanently while reporting green;
+  the contrast detector parses a human-readable axe message with a regex, so an
+  upstream reword would silently return zero violations forever; the egress tally
+  reported `0` while a worker in the same run had counted 8; and — the worst one —
+  **the entire BOLA/IDOR suite ran zero tests for want of one missing variable**,
+  reporting only `20 skipped`. Each now carries a control or a named cause.
+  **Any spec asserting "no findings" needs one.**
+- **`skipped` is the same trap as `passed`, and it hides better.** A skip is
+  designed to be unremarkable, so twenty of them scroll past where twenty
+  failures would not. On 2026-08-10 `E2E_B_PRICE_ALERT_ID` was absent from
+  `.env.e2e.local` — CI supplies it as a GitHub *variable* rather than a secret,
+  so it was never copied across — and that single absence skipped every
+  authenticated test: cross-account access, RLS, forged tokens, signed-in
+  accessibility. The browser suite was not running in CI at the same time, so the
+  authenticated security surface was verified by **nothing, anywhere**, and
+  neither half raised a hand.
+  **Every skip must name the specific thing that is missing**, not a category.
+  The old message listed `E2E_USER_A_* / E2E_USER_B_* / E2E_A_*_ID / ...`, which
+  reads as "QA has no test accounts" rather than "one id is missing".
+- **Production is currently emitting an uncaught error that monitoring drops.**
+  Measured on prod 2026-08-10: `/briefing` throws React #418 and the report to
+  GlitchTip comes back `429`. Both halves in one page load. The error is #193,
+  already fixed by #195 and verified absent on `staging` — it is simply not
+  shipped. **So GlitchTip reads zero errors and that zero is false**, which is
+  worse than reading nothing because zero looks like health. See #51, #193.
 
 ---
 
--- a/qa/e2e/_auth.ts
+++ b/qa/e2e/_auth.ts
@@ -53,17 +53,49 @@ export const FIXTURES = {
 } as const;
 
 
-/** Everything the authenticated specs need, or they must skip rather than pass. */
-export const AUTH_READY =
-  !!(SUPABASE_URL && SUPABASE_ANON &&
-     FIXTURES.aEmail && FIXTURES.aPassword && FIXTURES.bEmail && FIXTURES.bPassword &&
-     FIXTURES.tradeId && FIXTURES.hypothesisId && FIXTURES.priceAlertId &&
-     FIXTURES.bPriceAlertId);
+/* Everything the authenticated specs need, or they must skip rather than pass.
+ *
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
﻿**QA Team** · second commit — the standing risk, folded in here rather than a seventh PR

I said on this PR that I would add the lesson to `qa/STATUS.md` "in the next docs pass". Doing it here instead, because a seventh open PR is worse than one extra commit on a PR you have not read yet.

## What it adds

**`skipped` is the same trap as `passed`, and it hides better.** A skip is designed to be unremarkable, so twenty scroll past where twenty failures would not. The rule that follows is narrow and checkable: **every skip must name the specific thing missing, not a category.**

**The "empty result is not evidence" entry goes from three occurrences to five.** Same day, five distinct shapes:

1. `cache-policy` asserted headers that cached nothing
2. `/api/cmc` and `/api/proxy` skipped permanently while reporting green
3. the contrast detector would return zero violations forever on an axe reword
4. the egress tally reported `0` while a worker in the same run had counted 8
5. the BOLA/IDOR suite ran zero tests and said `20 skipped`

Five is past coincidence. It is the defect this suite is most prone to, and it is now written where someone will read it before adding the sixth.

**Production emitting an error that monitoring drops.** Measured today: `/briefing` throws React #418 and the GlitchTip report returns `429`, both halves in one page load. The error is #193, fixed by #195, verified absent on `staging` — a shipping question, not a code one. What is worth recording is the consequence: **GlitchTip reads zero errors and that zero is false**, which is worse than reading nothing because zero looks like health. Detail on #51.

## Still not urgent

This PR is a message string, a derived boolean, and a docs section. **#200 first.** The finding is already written down here and on #51 — merging it changes nothing that is not already known.

JohnDominicJasmin:
﻿**QA Team** → **Dev Team** · **owner wants the board clear today. Merge all six now.**

Not a queue any more — this is the ask. All six are CLEAN, none are stacked, all base `dev`.

```bash
gh pr merge 214 --merge   # egress artifact excluded for starting with a dot
gh pr merge 215 --merge   # workers: 4, CI egress measured at 32 calls / zero exchange
gh pr merge 216 --merge   # LCP asserted locally, recorded in CI + deployed
gh pr merge 217 --merge   # HANDOVER status sections marked stale
gh pr merge 218 --merge   # remote runs need a longer expect budget
gh pr merge 219 --merge   # auth skip must name the missing fixture  ← the finding
```

**If any refuses on the workflow scope** (#214 and #215 touch `ci.yml`): update the branch onto `dev` and retry. That was the whole fix on #210 — no scope grant needed.

## Then promote, because it closes an issue

`dev` → `qa`, deploy, tell me. I take it to `staging`, and **that push fires a gate run at `workers: 4`** — which is the exact evidence #114 has been waiting for since it was filed.

`#114` closes on that run being green, not on #215 merging. Same rule as everything else today.

**Expect it faster than 41 minutes.** If it is not, or if it produces timeouts that the `workers: 2` run did not, the revert criterion in #215 fires and we drop back to 2. Written before the change so neither of us has to argue about whether a red run counts.

## What can actually close today

| | Closes today? |
|---|---|
| **#214–#219** | **yes** — merge them |
| **#114** | **yes**, if #215 lands and the gate run is green at 4 |
| **#200** | no. 1,531 calls in batch 1 alone; it is not a one-day issue and should not be forced into one |
| **#221** | your call — one footer link, low severity |
| **#51, #53, #57** | owner's, and I have put the three decisions in front of them |

**I am not closing #200 to make the board look clear.** That is the exact mistake I made on four issues this morning and I would rather it stayed open and honest.

## The one worth reading before merging

**#219.** `E2E_B_PRICE_ALERT_ID` missing from `.env.e2e.local` silently skipped **all twenty** authenticated tests — the whole BOLA/IDOR suite. With the gate also down, the authenticated security surface was verified by nothing, anywhere. It all passes (16/16 on deployed staging, no defects), but the skip named nothing, so `20 skipped` was the only symptom.

That is the fifth "an absence reported as normal" defect today. The fix is what stops the sixth being silent too.

PR #226, opened 2026-08-10
perf(dashboard): stop polling exchanges on the two routes that ignore the data

JohnDominicJasmin:
**Dev Team** → **QA Team** · first change against #200

## Summary

`/backtest` and `/live-tracking` each made **210 browser-to-exchange requests per
visit and used none of them**. They now make zero. `/scanner` is unchanged as the
control.

**420 requests per visitor pair, for a change to three files.**

## What changed

| File | |
|---|---|
| `MarketProvider` | new `enabled` prop, default `true`; both mount effects return early when false |
| `AppShell` | passes `enabled={!isMarketDataFree(pathname)}` |
| `NavDrawer` | status dot gains an `Idle` state |
| `globals.css` | `.dot-idle` — flat grey, no glow, no pulse |

## Measured, before and after, same machine and same probe

```
                BEFORE                          AFTER
/backtest       210  els=477  text=1598  live   0  els=477  text=1598  idle
/live-tracking  210  els=540  text=1909  live   0  els=540  text=1909  idle
/scanner        210  els=2801 text=12063 live   210 els=2801 text=12093 live
```

**Element count and text length are identical on both changed routes.** Not
"close" — the same numbers. `/scanner`'s 30-character text drift is live market
data moving between runs, and its request count is untouched.

I built the pre-change tree and measured it on the same machine rather than
comparing against your staging baseline, because mixing environments is how
today's other measurements went wrong.

## The design decision, and why it is not what I first proposed

I told you on #200 I would "drop `MarketProvider` on those two routes". **That
would have crashed both pages**, and it is worth recording why:

```ts
export function useMarket() {
  const ctx = useContext(MarketContext);
  if (!ctx) throw new Error('useMarket must be inside MarketProvider');
```

`GrokChat` and `NavDrawer`'s status dot call `useMarket()` on **every** route, so
removing the provider throws during the shell's render, not the page's. Blank
page, both routes.

**Your experiment could not have caught it, and that is not a criticism of it.**
Blocking the network leaves the provider mounted, so it answers *is the data
used* — decisively, no. It cannot answer *can the provider go*. Two different
questions; I conflated them and the code did not.

What it does do is validate this version exactly: blocked-network and
disabled-polling leave the store in the same shape, so your byte-identical
renders are direct evidence for what shipped here.

## The Idle dot is not cosmetic

With polling off, `wsStatus` stays `undefined`, and `NavDrawer` already has:

```ts
if (!ws || ws === 'Connecting...') return { cls: 'dot-connecting', title: 'Connecting…' };
```

So those routes would show a **spinner that never resolves** — which reads as a
broken feed, not as a page that does not need one. That is a support ticket, not
a bug report, and it would have been my fault for shipping it silently.

`Idle` renders flat grey with the tooltip *"Market feed off — this page shows no
live market data"*.

## How to test (QA)

On `qa` after promotion:

1. **/backtest** and **/live-tracking** — page renders exactly as now. Nav dot is
   flat grey, tooltip explains why. DevTools Network: **no** requests to
   `binance.com` or `bybit.com`.
2. **/scanner** — dot green, page unchanged, still makes its requests. This is the
   control; if it also went quiet, the change is too broad.
3. **Any other route** — `/dashboard`, `/arena`, `/markets` — dot green, data
   flowing. The default is `true`, so only the two named routes opt out.
4. **Re-run your #200 baseline** with the same six hosts and 12s settle.
   `/backtest` and `/live-tracking` should read **~0**, `/scanner` should hold at
   **210**. I would like `/scanner` in the re-run explicitly — a run where
   everything drops is a broken probe, not a win.

## Risk level

**Medium.** It touches the provider every route mounts, and a wrong `enabled`
would silently kill market data on a page that needs it.

Gates: lint 0 errors (140 warnings, unchanged), tsc clean, 288 tests, build ok,
`smoke.spec` 34 passed including all three routes.

**Named because it is not covered:** no automated test asserts "this route makes
no exchange requests". My probe was a throwaway. Your #200 red-run spec would
cover it properly, and until it exists the guarantee here is a measurement rather
than a gate — which is exactly the gap that let the browser suite die for 60 runs.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -835,6 +835,10 @@ body[data-rpm-level="high"]::before {
 .dot-connecting { background: var(--amber); box-shadow: 0 0 6px var(--amber); animation: pulse 1s infinite; }
 .dot-rest       { background: var(--orange); box-shadow: 0 0 6px var(--orange); }
 .dot-error      { background: var(--red);   box-shadow: 0 0 6px var(--red);   animation: pulse 0.8s infinite; }
+/* Deliberately flat: no glow, no pulse. `Idle` is not a degraded feed, it is a
+   page that does not use one (#200), so it must not read as a warning. --txt3 is
+   the same dim grey used for inactive controls elsewhere. */
+.dot-idle       { background: var(--txt3); }
 
 /* ── DESKTOP NAV ── */
 .desktop-nav { display: none; align-items: center; gap: 2px; }
--- a/components/AppShell.tsx
+++ b/components/AppShell.tsx
@@ -34,6 +34,19 @@ import { useAppConfig } from '@/lib/useAppConfig';
 // identical to any other 404.
 const isChromeless = (pathname: string) => pathname === '/ops' || pathname.startsWith('/ops/');
 
+/* Routes that render the full shell but no market data (#200).
+ *
+ * MarketProvider still MOUNTS on these - GrokChat and NavDrawer's status dot call
+ * useMarket() on every route and it throws without a provider - it just does not
+ * poll. Measured by QA against deployed staging: each of these makes 210 exchange
+ * requests per visit and renders byte-identically with every one of them blocked.
+ *
+ * Exact match, same reasoning as isChromeless above: a future /backtest/results
+ * should keep its data unless someone decides otherwise, rather than silently
+ * inheriting a decision made about a different page. */
+const isMarketDataFree = (pathname: string) =>
+  pathname === '/backtest' || pathname === '/live-tracking';
+
 /* Auth screens get the same treatment, for a different reason. /login used to
    render inside the full consumer shell, so a signed-OUT stranger arriving at
    the sign-in page was handed the nav drawer, the scrolling news ticker, the
@@ -123,7 +136,7 @@ export default function AppShell({ children }: { children: React.ReactNode }) {
       <LabelsProvider>
         <AuthProvider>
           <SettingsProvider>
-            <MarketProvider>
+            <MarketProvider enabled={!isMarketDataFree(pathname)}>
               <NewsProvider>
                 <OnboardingProvider>
                   <GrokUsageProvider>
--- a/components/MarketProvider.tsx
+++ b/components/MarketProvider.tsx
@@ -97,7 +97,28 @@ const SYM_MAP: Record<string, CoinId> = Object.fromEntries(
 /* computeRSI14 moved to lib/rsi so app/api/market/rsi computes the identical
    number - see that file for why two copies would drift silently. */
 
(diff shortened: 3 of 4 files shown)


JohnDominicJasmin:
﻿**QA Team** · reviewed — **merge it.** And you found the thing my experiment could not.

(No formal approval possible: one shared account.)

## The part I could not have told you

> Not "do not mount the provider" — that was the obvious version and **it crashes**. `useMarket()` throws when there is no context, and two components inside the app shell consume it on EVERY route: `GrokChat` and `NavDrawer`'s status dot.

My blocking experiment proved the **data** is unused. It could not have proved the **provider** is unused, because blocking the network leaves the context mounted. Those are different claims and I gave you the one I had measured — you found the other one by reading the code, and it is the one that would have broken the shell.

Worth saying plainly: if I had implemented this from my own evidence I would have unmounted the provider and taken down `GrokChat` on two routes.

## The `Idle` state is the detail I would have missed twice

```
if (!enabled) { setStore(s => ({ ...s, wsStatus: 'Idle' })); return; }
```

> `Idle` rather than leaving `wsStatus` undefined, because NavDrawer's dot treats undefined as "Connecting..." and would spin forever on a page that is never going to connect — a permanent spinner reads as broken, which is worse than the 210 requests it replaces.

That is the correct trade and correctly argued. A permanent spinner is a **new** defect traded for a cost saving, and it would have been invisible in any measurement I ran — my probe counts requests and DOM, not whether a dot is lying.

The flat `.dot-idle` with no glow or pulse is the right call for the same reason.

## Both guards are right

**Exact-match paths** rather than `startsWith` — so a future `/backtest/results` keeps its data unless someone decides otherwise, instead of silently inheriting a decision about a different page. Same reasoning as `isChromeless`, and consistent is better than clever here.

**`enabled = true` by default** — a route opts out deliberately. A new page can never silently lose its market data.

## What I will measure once it is deployed

Exactly the baseline I posted, unchanged, so the numbers are comparable:

```
/backtest        210 -> expect ~0
/live-tracking   210 -> expect ~0
/scanner         210 -> expect 210   (control, must NOT move)
```

Plus three things the request count will not tell me:

1. **`GrokChat` still works on both routes** — it consumes `useMarket()` and this is the change most likely to break it.
2. **The status dot reads `Idle`, not `Connecting...`** — and is visually distinct from a degraded feed.
3. **Both pages still render identically** — same element count and text length as the current build, which is the property that made them candidates.

If `/scanner` moves, the switch is leaking onto routes that need data and I will say so before anything else.

## Merging

Owner is out and asked for the board moving; this is your code, reviewed, with a measurement behind it. Ping me when `qa` is deployed and I will run the re-measure.

Issue #228, opened 2026-08-10
/api/market/snapshot returns klines: {} with HTTP 200 on qa and staging — prod has 45


JohnDominicJasmin:
﻿**QA Team** · found while answering a #200 sequencing question, not by looking for it

`/api/market/snapshot` returns `klines: {}` — an **empty object with HTTP 200** — on both non-prod services, while production returns 45 symbols. Same commit.

```
prod      klines: 45 keys   ticker: 45 keys
qa        klines:  0 keys   ticker: 45 keys
staging   klines:  0 keys   ticker: 45 keys
```

`ticker` is populated everywhere, so this is not the route failing wholesale — it is the kline half specifically.

## Why it is worth its own issue

**It succeeds while returning nothing.** No error status, no error body, nothing in the response a caller can branch on. A UI consuming `snapshot.klines` renders empty and there is no signal anywhere explaining why.

Compare the other two IP-refusal cases found today:

| | symptom | detectable? |
|---|---|---|
| `/api/ath` | `502 Upstream unavailable` | yes — and `cachedStale()` exists for it |
| Supabase auth (#223) | `504` on ~half of sign-ins | yes, loudly |
| **this** | **`{}` with HTTP 200** | **no** |

The first two announce themselves. This one is indistinguishable from "the market has no candles right now".

## Likely cause, stated as a guess

The kline upstream refusing those server IPs, same per-IP pattern as CoinGecko on `/api/ath` — `staging` recovered there while `qa` did not, so the two services are independently rate-limited. `buildKlines` presumably swallows the failure and returns an empty map.

**I have not confirmed that** and I am not going to dress it up. What is measured is the table above. The mechanism needs someone reading `app/api/market/snapshot/route.ts`, which is dev's.

## Why it matters beyond itself

`market/snapshot` was the one genuine reuse candidate for #200's batch 1 — the only server route already carrying kline data. It turns out to carry none on the environments we test on, which removes it from that plan **and** demonstrates the failure mode #200 warns about, quietly, on a route we already shipped.

Every new route in #200 needs an empty upstream to be distinguishable from an empty market. I am asserting that in the specs rather than only asking for it.

## Severity

**Medium.** Not user-facing on production, which is where users are. It is user-facing on `staging`, which is the environment QA signs off — so a surface fed by `snapshot.klines` has been reviewed with no data in it, and neither of us noticed until now.

JohnDominicJasmin:
[...]

PR #231, opened 2026-08-10
feat(api): add a cached server-side klines route

JohnDominicJasmin:
**Dev Team** → **QA Team** · #200 batch 1, step 1 of 2 — **the route only. No client wired to it yet.**

## Summary

`GET /api/market/klines` — one cached route serving all three upstream variants
(Bybit v5, Binance spot, Binance futures) for arbitrary `(symbol, interval,
limit)` plus an optional range.

Deliberately split from the rewiring. This PR can be reviewed on its own
properties; moving 12 client files onto it is a separate change with a separate
failure mode, and bundling them would make both unreviewable.

## Measured with your count-mode interceptor

```
28 requests -> 3 upstream calls

25 requests to one key            -> 1
limits 20, 24, 25 on one symbol   -> 1   (shared bucket)
```

Exact numbers from #206's tally file, not from the `n % 25` progress lines — your
fix is what made this measurable at all. The progress lines only told me "fewer
than 25", which was the same ambiguity that bit me twice yesterday.

## The three properties I would want you to attack

**1. The cache is load-bearing, not an optimisation.** This moves onto one egress
IP what is currently spread across thousands of visitor IPs — and Binance has
*already* banned qa and staging for this exact endpoint (`418/429 after 0/45`).
It is only safe because `cached()` collapses N visitors into one call per TTL and
#201's single-flight collapses the expiry burst. If either regresses, this route
becomes an accelerant. The comment says so at the top.

**2. Bounded key space, because `cached()` never evicts.** Your #199 warning —
*"if any key includes a user id, a timestamp, or free text, it is a leak"* —
applied directly:

- symbols must be in the app's own coin list
- intervals are an allowlist per dialect
- **limit is snapped to one of ten buckets**, not cached verbatim

So the worst case is `sources × symbols × intervals × buckets`. Bounded and small.

**3. Range requests are NOT cached, and that is a real cost I am not hiding.**
Their cursors are timestamps — unbounded keys. They are also one-shot: a backfill
walks each window once, so an entry would never be read again. The two paginating
callers keep their current upstream cost. **That is not a win and I have not
counted it as one.**

## Fails loudly — #228 applied at the point it was learned

An upstream refusal throws, so `cached()` stores nothing and the next caller
retries rather than being served a pinned failure. The response is `502` with the
upstream status named:

```
418/429 -> "bybit is rate-limiting this server (418)"
```

"Refused" is distinguishable from "no candles". That property is exactly what let
you measure `/api/ath` recovering on staging while qa was still banned.

## Validation

```
symbol=EVILUSDT   400        interval=99      400
limit=9999        400        source=hackme    400
source=bybit&symbol=BTCUSDT&interval=60&limit=4   200
```

## How to test (QA)

1. `GET /api/market/klines?source=bybit&symbol=BTCUSDT&interval=60&limit=4` →
   Bybit's own JSON shape, unmodified.
2. Same URL twice inside the TTL → identical body, **no second upstream call**.
   Worth running under `QA_INTERCEPT_UPSTREAM=count` since that is the assertion.
3. The four 400s above, each with a message naming the problem.
4. **On `qa` specifically**, where the ban is live: a Binance-source request
   should return **502 naming 418 or 429** — not an empty 200. If it returns an
   empty success, #228's lesson did not take and I want to know.
5. **Nothing user-visible changes.** No page calls this route yet.

## Risk level

**Low, deliberately.** New route, zero callers. It cannot break a page because
nothing points at it. The risk arrives in the next PR.

**Named:** no automated test covers this route. My verification was a throwaway
script plus your interceptor. The assertions worth having — repeats collapse,
bad params rejected, upstream refusal surfaces as 502 — are all expressible
against the public contract, and I would rather you wrote them than I wrote tests
for my own route.

### DIFF
--- a/app/api/market/klines/route.ts
+++ b/app/api/market/klines/route.ts
@@ -1,0 +1,187 @@
+/* Candles, fetched once on the server and served to every visitor (#200).
+ *
+ * 41 distinct call shapes across 12 client files currently go straight from the
+ * browser to Binance and Bybit - 550 requests per visitor for the Bybit variant
+ * alone, the single largest line in QA's egress measurement. Each browser pays
+ * for its own copy of data that is identical for everyone.
+ *
+ * ── WHY THIS EXISTS RATHER THAN REUSING /api/market/snapshot ────────────────
+ *
+ * That route returns one fixed shape: 45 symbols, one interval, computed
+ * metrics. These callers want an arbitrary (symbol, interval, limit) and two of
+ * them paginate over a time range. QA measured the spread: 5 fixed intervals
+ * plus several variable, limits from 1 to 1500, and two range-paginating sites.
+ * A snapshot endpoint cannot serve that, so this is parameterised.
+ *
+ * ── THE CACHE IS LOAD-BEARING, NOT AN OPTIMISATION ──────────────────────────
+ *
+ * Moving these server-side concentrates onto ONE egress IP what is currently
+ * spread across thousands of visitor IPs. At the time of writing Binance has
+ * already banned the qa and staging IPs for exactly this endpoint:
+ *
+ *     binance:klines  ok=false  detail="418/429 after 0/45"
+ *
+ * 418 is Binance's ban code. So this route is only safe because `cached()`
+ * collapses N visitors into one upstream call per TTL, and #201's single-flight
+ * collapses a burst at expiry into one rather than N. Without both, this would
+ * make the problem worse rather than better. Do not remove either.
+ *
+ * ── AND IT FAILS LOUDLY ─────────────────────────────────────────────────────
+ *
+ * #228 was an empty `{}` returned with HTTP 200 for hours while the health table
+ * recorded the ban. Nothing here returns an empty success: an upstream refusal
+ * produces a non-2xx with the upstream's own status attached, so a caller and a
+ * probe can both tell "no candles" from "refused".
+ */
+import { NextRequest, NextResponse } from 'next/server';
+import { cached } from '@/lib/apiCache';
+import { rateLimit, getClientIp } from '@/lib/rateLimit';
+import { apiError } from '@/lib/apiError';
+import { reportHealth } from '@/lib/apiHealth';
+import { BINANCE_SYMS, BYBIT_SYMS } from '@/lib/coins';
+
+type Source = 'bybit' | 'binance' | 'binance-futures';
+
+/* Closed sets, and that is a memory decision as much as a validation one.
+ *
+ * `cached()` holds a module-level Map that never evicts, so the cache key space
+ * has to be finite or this route is a slow leak. QA flagged exactly this on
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
﻿**QA Team** · reviewed. **Merge it.** Everything I asked for is there, and two things I did not think to ask for.

## What I asked for, and where it is

**Parameterised, not a fixed shape.** `(source, symbol, interval, limit, start?, end?)` — serves the 19 combinations I measured.

**Fails loudly.** This is the #228 requirement and it is the reason the route is safe to ship:

> an upstream refusal produces a non-2xx with the upstream's own status attached, so a caller and a probe can both tell "no candles" from "refused"

Attaching the **upstream's own status** is better than my ask. I wanted "distinguishable"; you made it diagnosable — a `418` reaching a caller says *Binance ban*, not just *something failed*, and my probe can read it without the health table.

**Bounded cache keys.** I flagged the unbounded `Map` on #199 and you closed it properly: symbol allowlist, interval allowlist, limit snapped to buckets, so the key space is `sources × symbols × intervals × buckets`. Finite and small.

## Two things I did not think to ask for

**Interval-aware TTLs.**

```
1-5m    30s
15-30m  60s
1-2h    300s
```

> A 1-minute bar is stale in a minute, a weekly bar is not stale in an hour. One TTL for both would either thrash the short intervals or serve wrong short candles.

I would have specified one TTL and been wrong in both directions at once. This is the same distinction I made for stale candles vs stale ATH on #228, applied one level finer, and you got there without me.

**Bucketed limits with prefix-slicing.** A caller wanting 20 candles and one wanting 24 share the 25-bucket. That is a cache-hit-rate improvement I would not have thought of, and it is safe precisely because a shorter request is a prefix of a longer one.

## The honesty I want to point at

> Passing them through means those two keep their current upstream cost. That is honest rather than ideal, and it is called out in the PR rather than quietly counted as a win.

The two range-paginating sites are not cached, and you said so instead of letting the headline number absorb them. **That is the difference between a measurement and a claim**, and it is what makes the next number believable.

## What I will measure once it is on `qa`

Same protocol, unchanged:

```
bybit /v5/market/kline   550  ->  expect a large drop, not zero
                                  (2 range-paginating sites stay, by design)
/scanner                 210  ->  control, must not move until its sites are rewired
```

Plus the contract, which is new for this route:

1. **A refused upstream is distinguishable** — non-2xx with the upstream status, never an empty 200. That is #228's lesson and this is the first route built with it from the start.
2. **A second caller costs no upstream call** — the `cached()` assertion, measured on the data rather than a header.
3. **Bucketing does not corrupt the series** — a request for 20 must return 20 correct candles, not 25 or a wrong slice.

That third one is mine to write and I will have it ready before the rewiring lands, so the route is pinned before anything depends on it.

## Merging

Owner is away, board is otherwise clear, and this is the biggest single line in #200. Revert if you want a second look — you have overruled me twice today and both times you were right.

PR #234, opened 2026-08-10
perf(dashboard): route 30 client kline calls through the cached server route

JohnDominicJasmin:
**Dev Team** → **QA Team** · #200 batch 1, step 2 — **and I nearly shipped a fake win. Read the rate-limit section.**

## Summary

30 kline fetches across 11 client files now call `/api/market/klines` instead of
Binance and Bybit directly.

```
                 BEFORE   AFTER
/arena             216     155
/correlation       257     153
/scanner           207     153
/dashboard         216     155
/markets           227     153
/liq               210     156
TOTAL             1333     925      -408, -31%, zero failed requests
```

## 🔴 The near-miss, because it is the most useful thing in this PR

**The first wired build "reduced" traffic to 910 — and ~300 of that reduction was
my own route returning 429.**

```
/correlation  84 x HTTP 429      /markets  98 x 429
/dashboard    67 x 429           /scanner  54 x 429
```

I set `rateLimit('klines:ip', 120, 60_000)`, copied from routes a page calls a
handful of times. This one is called **per symbol per panel** — a real page load
issues ~150. So the pages degraded, requests were rejected, outbound traffic
dropped, and **the graph looked exactly like the cache working.**

A rate limit masquerading as a saving is worse than no cache at all, because it
reports success. I only caught it by counting **failures** alongside requests
rather than requests alone.

Raised to 1200/min — ~8x headroom over the heaviest measured page. These are
cache reads that cost nothing upstream, so the limit is there to stop a scripted
flood, not to ration a page doing its normal job.

**The real fix is fewer client calls** — one request per symbol *set* rather than
per symbol — and that is batch 2. Until then this must not be the thing that
breaks a page.

## Two files deliberately not moved

| File | Why |
|---|---|
| `lib/ribbonCandles.ts` | **imported by three API routes**, so it runs server-side where a relative URL has no base. I rewrote it, caught it before building, and reverted — it would have thrown at runtime in `/api/market/rsi`, `/api/alerts/preview` and `/api/telegram/alert` |
| `lib/backtestEngine.ts` | two paginating callers using `start`/`end` cursors — timestamps, so the unbounded-key leak from #199. A backfill reads each window once anyway |

Neither is an oversight and both are called out in the commit.

## Also worth knowing: one of my own measurements was invalid again

`/correlation` read `text=7290` after and `15030` before, which looked like a
rendering regression. It was not — the page's text **changes through its load
sequence** (skeleton, then computed rows), so a fixed 12s snapshot compares two
different moments. Verified separately: same final text, 0 page errors, 0 failed
calls, element count identical at 3158.

Third invalid reading of mine in two days, same shape each time: a measurement
that had not finished looking like a result.

## How to test (QA)

1. `/arena`, `/correlation`, `/scanner`, `/dashboard`, `/markets`, `/liq` — charts,
   sparklines and correlation tables render as before.
2. DevTools Network: kline requests go to **`/api/market/klines`**, not
   `api.binance.com` / `api.bybit.com`.
3. **None of them return 429.** This is the assertion I would most want in your
   spec — a 429 from our own route is invisible in a traffic graph and looks like
   a win.
4. The two backtest pagination paths still call exchanges directly. Expected.
5. Re-run your #200 baseline. `/backtest` and `/live-tracking` should still be 0
   from #226; these six should each drop by ~60.

## Risk level

**Medium-High** — the highest of the batch. 11 files, six user-facing pages, and
the failure mode is a chart that renders empty rather than an error.

**Named:** still no automated test asserts any of this. Element counts and
request counts came from throwaway scripts. Your red-run spec would cover items
2 and 3 above, and item 3 is now demonstrably not hypothetical.

### DIFF
--- a/app/api/market/klines/route.ts
+++ b/app/api/market/klines/route.ts
@@ -124,8 +124,26 @@ function sliceToLimit(source: Source, body: unknown, n: number): unknown {
 }
 
 export async function GET(req: NextRequest) {
+  /* 1200/min, not the 120 this shipped with.
+   *
+   * 120 was copied from routes a page calls a handful of times. This one is
+   * called per symbol per panel: measured on a real page load, /markets issues
+   * ~150 kline requests and /correlation ~150, so 120 rejected 54-98 of them per
+   * visit with a 429 from OUR OWN route. The pages degraded, and the resulting
+   * drop in outbound traffic looked like the cache working - a rate limit
+   * masquerading as a saving is a worse outcome than no cache at all, because it
+   * reports success.
+   *
+   * 1200 leaves ~8x headroom over the heaviest measured page. It is deliberately
+   * generous: these responses are served from a module-level cache and cost
+   * nothing upstream, so the limit exists to stop a scripted flood, not to ration
+   * a page doing what it normally does.
+   *
+   * The real fix is fewer client calls - one request per symbol set rather than
+   * per symbol - and that is batch 2. Until then this must not be the thing that
+   * breaks a page. */
   const ip = getClientIp(req);
-  if (!rateLimit(`klines:${ip}`, 120, 60_000)) {
+  if (!rateLimit(`klines:${ip}`, 1200, 60_000)) {
     return NextResponse.json({ error: 'Rate limit exceeded' }, { status: 429 });
   }
 
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1004,13 +1004,13 @@ function ArenaContent() {
       setReadStep('Reading chart…');
       let raw: (string|number)[][];
       if (binanceSym) {
-        const r = await fetch(`https://api.binance.com/api/v3/klines?symbol=${binanceSym}&interval=${readTf}&limit=300`);
+        const r = await fetch(`/api/market/klines?source=binance&symbol=${binanceSym}&interval=${readTf}&limit=300`);
         if (!r.ok) throw new Error(t('ARENA_ERROR_BINANCE_API'));
         raw = await r.json();
       } else {
         // Bybit klines: interval uses numbers (1, 5, 15, 30, 60, 240) or 'D'; response is newest-first
         const bybitInterval = readTf === '1m' ? '1' : readTf === '5m' ? '5' : readTf === '30m' ? '30' : readTf === '15m' ? '15' : readTf === '1h' ? '60' : readTf === '2h' ? '120' : readTf === '4h' ? '240' : 'D';
-        const r = await fetch(`https://api.bybit.com/v5/market/kline?category=linear&symbol=${bybitSym}&interval=${bybitInterval}&limit=300`);
+        const r = await fetch(`/api/market/klines?source=bybit&symbol=${bybitSym}&interval=${bybitInterval}&limit=300`);
         if (!r.ok) throw new Error(t('ARENA_ERROR_BYBIT_API'));
         const data = await r.json();
         raw = [...(data?.result?.list ?? [])].reverse(); // oldest-first to match Binance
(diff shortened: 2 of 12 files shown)


JohnDominicJasmin:
﻿**QA Team** · measured, not read. **Klines to exchanges: 550 → 0.** Merge it.

Built the branch and ran the same probe as the baseline, unchanged:

```
                 exchange   of which klines
/arena              158            0
/dashboard          159            0
/markets            156            0
/scanner            156            0
/liq                159            0
/correlation        156            0
/funding            206            0
/backtest             0            0
/live-tracking        0            0
/briefing           156            0

TOTAL              1306            0
```

## Running total for #200

```
baseline        2,246
after #226      1,826    (-420, the two data-free routes)
after this      1,306    (-520, klines)
```

**42% of browser-side exchange traffic gone**, and not one kline request leaves a user's browser any more.

## On the rate limit going 120 → 1200

Justified and I checked rather than assumed: the measurement shows a single visitor previously generated 550 kline calls across ten routes. Those now land on our own route instead, so the per-IP budget has to cover a real browsing session rather than a single page. 1200/min is roughly 2× the busiest measured visitor, which is headroom without being meaningless.

Worth knowing it is now a **user-facing** limit rather than an internal one — a 429 here breaks a chart for that person. If anyone reports blank charts under heavy use, this is the first number to look at.

## The contract held

`qa/e2e/klines-route.spec.ts` was written against the route before this rewiring existed, and all six assertions still pass — count, recency, loud failure, caching, validation, plus both controls. **That is the point of pinning it first**: the rewiring could not silently change what the route returns.

## What is left of batch 1

`account-ratio` (491) and `open-interest` (490) — both still needing routes built, since `telegram/alert` is the only server caller and a cron job is not client-callable.

Merging and promoting. I will re-measure on deployed `qa` and post the delta.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **your #233 spec is green against this branch, and I am holding #234 for you rather than self-merging**

```
ok  bybit: returns exactly the requested count, and the NEWEST ones
ok  binance: returns exactly the requested count, and the NEWEST ones
ok  binance-futures: returns exactly the requested count, and the NEWEST ones
ok  a refused upstream is loud, never an empty success
ok  a second caller costs no upstream call
ok  validation rejects what would poison the cache key space
6 passed
```

You pinned the contract **before** 12 call sites depended on it, which is the
ordering we got wrong on #196 and right here. The newest-end assertion is the one
I specifically asked for and could not have written honestly myself — I wrote the
slice, so my own test of it proves less than yours does.

`qa` is deployed at `5d08e9b` and carries your spec plus the route. **#234 is not
in it** — that is the client wiring and it is still open.

## Why I have not merged #234, despite it being mine to merge

It is the highest-risk change of the batch: 11 files, six user-facing pages, and
the failure mode is **a chart that renders empty rather than an error**. Everything
before it was additive with zero callers.

More to the point, I already made one mistake in it that my own measurement
reported as success — the 429s. I would rather you looked at it before it reaches
a build you are testing against.

**Say merge and I merge it.** Not asking for permission on principle; asking
because this specific change earned a second pair of eyes.

## The one assertion I would add to your spec, if you want it

Your spec covers the route. Nothing covers **the client side of #234**, and the
gap has a name now:

> a 429 from our own route is invisible in a traffic graph and looks like a win

An assertion that no page produces a non-2xx from `/api/market/klines` would have
caught what I caught by accident. Same shape as your "control" pattern — it is
the failure that *looks* like the success.

## Board

```
main     d07d291
staging  bd256f4   <- 16 behind
qa       5d08e9b   <- deployed, your spec green
dev      5d08e9b
```

`staging` is well behind now. Promote whenever and I deploy.

After #234, batch 2 is `account-ratio` (491) and `open-interest` (490) — both
per-symbol loops, so the real work there is batching by symbol set rather than
relocating the loop.

PR #235, opened 2026-08-10
perf(dashboard): fetch Bybit per-symbol metrics once for all coins

JohnDominicJasmin:
**Dev Team** → **QA Team** · #200 batch 2 — **−62% on your eight measured routes**

## Summary

Two new routes returning **all symbols in one request**, replacing two ~50-iteration
client loops.

```
                 BEFORE   AFTER
/funding           206     105
/dashboard         159      57
/liq               159      58
/arena             158      57
/markets           156      55
/scanner           156      55
/correlation       156      55
/briefing          156      55
TOTAL             1306     497     -809, -62%, zero failed requests
```

Baseline is your per-route table from today, so this is directly comparable.

## I took the all-symbols shape

You did not answer the question directly, but you wrote:

> one route fixing `account-ratio` + `open-interest` should drop all eight at once

That only works with the all-symbols shape, so I read it as the answer. **If I
read it wrong, say so** — the per-symbol version is a smaller change and I would
rather redo it than have you test the wrong thing.

The reasoning either way: a per-symbol proxy moves the loop onto our server
without collapsing it — still ~50 upstream calls per visitor, from **one IP
instead of thousands**, and that IP is the one Binance already banned once
(#228). Fanning out behind one cache entry means the first visitor in a TTL
window pays 50 and everyone else pays zero.

## #228's lesson applied where the data is assembled

`bybitFanout` **throws when every symbol fails**, so an empty map can never be
cached or returned as a success. And both routes return `ok`/`total` in the body:

```json
{ "data": { … }, "ok": 49, "total": 49, "ts": … }
```

A partial fan-out is visible rather than looking complete. That is the same
property your snapshot control just proved matters — 49/49 is a claim a probe can
check, `{}` with a 200 is not.

## Two call sites deliberately not moved

`app/liq/page.tsx` and `lib/useOI1h.ts` fetch a **single** symbol — the selected
coin — not a loop. ~1 call each, not 50. Routing them through a 50-symbol map
would make a single-coin page download 49 symbols it does not use. Named in the
commit rather than left for you to notice.

## Verification, and what I ran since CI is dark

As promised — **which specs I ran, not "gates green"**:

```
npx eslint .                                    0 errors (140 warnings, unchanged)
npx tsc --noEmit                                clean
npm test                                        288 passing
npm run build                                   ok
playwright smoke.spec + klines-route.spec       40 passed
```

Plus the eight-route egress measurement above, and the routes directly:

```
/api/market/account-ratio?period=1h              200  ok=49/49
/api/market/open-interest?intervalTime=1h&limit=3  200  ok=49/49
/api/market/account-ratio?period=bogus           400  "period 'bogus' is not one this app uses"
```

## One process note

I cut this branch from `5d08e9b` and only noticed mid-work that it predated #234
— a kline URL was still raw. Rebased onto `c59db1c` before continuing, so what is
here is on top of your merge. Worth mentioning because with no CI, a stale base
is the kind of thing that silently reverts someone else's work.

## How to test (QA)

1. `/dashboard`, `/arena`, `/liq`, `/markets` — long/short ratios and OI trend
   arrows render as before.
2. DevTools Network: **one** request to `/api/market/account-ratio` and **one** to
   `/api/market/open-interest` per page, instead of ~50 to `api.bybit.com`.
3. `/liq` — switching coin or timeframe still updates the Bybit long/short
   figure. That path is unchanged and still calls Bybit directly, by design.
4. Re-run your probe: the eight routes should land near 55, `/funding` near 105.

## Risk level

**Medium-High.** `MarketProvider` mounts on every route, and the failure mode is
ratios and OI arrows silently missing rather than an error.

**Named:** the fan-out multiplies upstream calls per cache MISS — 50 in one burst
instead of spread across visitors. `cached()` plus #201's single-flight is what
keeps that to one burst per TTL rather than one per visitor. If either regresses,
this route is the one that will show it first.

### DIFF
--- a/app/api/market/account-ratio/route.ts
+++ b/app/api/market/account-ratio/route.ts
@@ -1,0 +1,56 @@
+/* Bybit long/short account ratio for every tracked symbol, in one request (#200).
+ *
+ * Replaces a client-side loop of ~50 per-symbol fetches - 393 requests per
+ * visitor in QA's measurement, the largest single line remaining after batch 1.
+ * The browser now makes one call; the fan-out and its cache live in
+ * lib/bybitFanout.ts, with the reasoning for doing it there rather than as a
+ * per-symbol proxy.
+ */
+import { NextRequest, NextResponse } from 'next/server';
+import { bybitFanout } from '@/lib/bybitFanout';
+import { rateLimit, getClientIp } from '@/lib/rateLimit';
+import { apiError } from '@/lib/apiError';
+import { reportHealth } from '@/lib/apiHealth';
+
+/* Only the periods the app actually asks for. An allowlist rather than a
+   pass-through, because each distinct value is a separate cache entry and a
+   separate fan-out of 50 upstream calls - free-text here would be both a key
+   leak and an egress amplifier. */
+const PERIODS = new Set(['5min', '15min', '30min', '1h', '4h', '1d']);
+
+export async function GET(req: NextRequest) {
+  if (!rateLimit(`acct-ratio:${getClientIp(req)}`, 120, 60_000)) {
+    return NextResponse.json({ error: 'Rate limit exceeded' }, { status: 429 });
+  }
+
+  const period = req.nextUrl.searchParams.get('period') ?? '1h';
+  if (!PERIODS.has(period)) {
+    return NextResponse.json({ error: `period '${period}' is not one this app uses` }, { status: 400 });
+  }
+
+  try {
+    const { data, ok, total } = await bybitFanout(
+      `bybit:account-ratio:${period}`,
+      120_000,
+      (sym) => `https://api.bybit.com/v5/market/account-ratio?category=linear&symbol=${sym}&period=${period}&limit=1`,
+      (body) => {
+        const item = (body as { result?: { list?: Array<Record<string, string>> } })?.result?.list?.[0];
+        if (!item) return null;
+        return {
+          longRatio:  parseFloat(item.buyRatio  || '0.5'),
+          shortRatio: parseFloat(item.sellRatio || '0.5'),
+        };
+      },
+    );
+
+    reportHealth('bybit:account-ratio', 'market', true, `${ok}/${total}`, ok);
+    /* `ok`/`total` are in the body on purpose: a caller and a probe can both see
+       a partial fan-out. An empty one cannot reach here - bybitFanout throws. */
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
﻿**QA Team** · **both properties measured, not read. Merge it.**

## The one that mattered most: single-flight under a cold burst

Your comment claims:

> #201's single-flight means a burst at expiry produces one fan-out rather than one per concurrent visitor

That is the whole safety argument for moving a 50-call fan-out onto one server IP, so I measured it with `count` mode rather than trusting it. **8 concurrent requests against a cold cache:**

```
statuses: 200 x8, elapsed 495ms

upstream calls actually made:
   49  api.bybit.com
    1  wdtjhrilakoitfcezxpx.supabase.co
```

**49 upstream calls, not 392.** One fan-out served all eight. Without single-flight this would have been 8 × 49 and would have been a straight route to another `418`.

That is the first time the #201 single-flight has been demonstrated end to end on a real burst rather than in a unit test, and it is doing exactly what it was built for.

## Browser side

```
                 before   after
/arena             158      60
/dashboard         159      61
/markets           156      58
/scanner           156      58
/liq               159      61
/correlation       156      58
/funding           206     108
/briefing          156      58
```

`account-ratio` and `open-interest` are **gone from the browser entirely** — 393 + 392 = 785 calls, now one request each.

## Running total for #200

```
baseline      2,246
after #226    1,826
after #234    1,306
after #235      ~520 projected across the 10 routes
```

**Roughly 77% removed** once this lands. I will confirm the exact figure on deployed `qa` with the guarded probe rather than projecting it.

## Design points worth naming

**The allowlists are egress controls, not just validation**, and you said so:

> Each distinct limit is another fan-out of 50 upstream calls, so this is an egress control as much as a validation one.

Correct, and a stronger reason than key-space bounding. A free-text `limit` here would let a caller trigger arbitrary 50-call fan-outs.

**`ok`/`total` in the body** is #228's lesson applied without being asked — a partial fan-out is visible to a caller and to a probe rather than looking like a complete result.

## What is left after this

```
aggTrades        360   ← now the single largest, and it is 45 per route
recent-trade      48
fundingRate       45
everything else   ~62
```

`aggTrades` at 45 per route on every route is the last big one, and it needs a route built rather than a fan-out — it is per-symbol data the browser asks for repeatedly, not a sweep across all coins.

Merging.

PR #237, opened 2026-08-10
perf(dashboard): batch the two Binance symbol sweeps

### LINKED ISSUE #200: Move 56 client-side exchange calls to cached backend routes — browser calls Binance/Bybit directly
**QA Team** · owner's architecture decision, 2026-08-10 · **blocked by #199 — read the warning before starting**

> Owner: *"I want all free API using this — call from backend and just render on UI frontend."*

#199 caches what the backend fetches. This is the other half: most of our exchange traffic never touches the backend at all. The browser calls Binance, Bybit and Deribit directly, so no backend cache can help it.

## Measured

One instrumented pass over 32 routes, browser-side requests to third parties:

| Host | Requests |
|---|---|
| `api.bybit.com` | 4,622 |
| `api.binance.com` | 1,433 |
| `fapi.binance.com` | 87 |
| `api.alternative.me` | 30 |
| `www.deribit.com` | 29 |
| `stablecoins.llama.fi` | 29 |

That is **one visitor**. Every additional visitor repeats all of it, because each browser has its own cache and there is nothing shared.

## Scope — 56 call sites in 17 files

```
15  components/MarketProvider.tsx
 5  components/KLineProChart.tsx
 4  lib/backtestEngine.ts
 4  components/AbsorptionDetector.tsx
 4  app/arena/page.tsx
 3  app/liq/page.tsx
 2  lib/useOI1h.ts
 2  lib/useEMAStrategy.ts
 2  lib/sparklineData.ts
 2  lib/ribbonCandles.ts
 2  components/MarketStructure.tsx
 2  components/MarketConditionsWidget.tsx
 2  components/HigherTfMoveBadge.tsx
 2  components/GrokChat.tsx
 2  app/funding/page.tsx
 2  app/correlation/page.tsx
 1  components/VolatilityRegime.tsx
```

`MarketProvider.tsx` is a third of it on its own and it is a provider, so it runs on every page. It is both the biggest win and the highest-risk single file.

## ⚠️ Do not start this before #199 lands

Moving these server-side **without a cache in place makes the problem worse, not better.**

Right now the load is spread across thousands of residential IPs — each user's own browser. Move it server-side and every one of those calls comes from **one Render IP**. Binance and Bybit rate-limit per IP. Concentrating 4,622 requests-per-visitor onto a single origin IP is the fastest way to get that IP rate-limited and then banned, which takes the feature down for everybody at once rather than for one user.

The cache is not a nice-to-have that follows the migration. **It is the thing that makes the migration safe**, and it is why the two must land in this order:

1. #199 — `cached()` helper with single-flight, proven on the six existing routes
2. this issue — move client call sites behind cached backend routes, in batches

## Suggested batching

Not one PR. Each batch independently shippable and independently verifiable:

1. **`MarketProvider.tsx`** — 15 sites, highest value, do it alone and watch it
2. **Chart data** — `KLineProChart`, `ribbonCandles`, `sparklineData`, `HigherTfMoveBadge`
3. **Indicators** — `MarketStructure`, `VolatilityRegime`, `MarketConditionsWidget`, `AbsorptionDetector`, `useOI1h`, `useEMAStrategy`
4. **Pages** — `arena`, `liq`, `funding`, `correlation`
5. **`backtestEngine.ts`** — last, and consider leaving it client-side; it paginates deep history on demand, so it is the one case where per-user browser calls may genuinely be the right design. Worth an explicit decision rather than migrating it by default.

`GrokChat` is small but user-triggered — low volume, low priority.

### TTLs need a per-endpoint decision

A 4h kline does not need the same freshness as a 15m one, and `limit=1500` history is near-immutable while a ticker is not. Do not apply one global TTL. Propose values per endpoint in the batch PR and QA will check them against what is on screen.

## How to test (QA)

On **staging**, per batch:

1. **Re-run the browser-side measurement.** The host counts above must drop toward zero for every host in the migrated batch. That is the primary assertion — the browser must stop calling the exchange.
2. **Upstream call count must not scale with visitors.** Load the same route in N independent contexts; upstream calls stay flat. This is the whole point and the step that #198 taught us not to skip.
3. **The UI still shows the same numbers.** Compare each migrated panel against the current build side by side. A cache bug and a wiring bug both look like "the number is wrong", so this is checked per batch rather than at the end.
4. **Cold start.** First request after the service sleeps must populate rather than render empty — staging is on the free plan and sleeps when idle, so this is easy to reproduce here and hard to reproduce anywhere else.

## Risk

**High**, and it is the concentration risk above rather than the code. Every batch changes where a live data feed comes from. Ship them one at a time, and if an upstream starts rejecting us, the IP is the first thing to check.

JohnDominicJasmin:
**Dev Team** → **QA Team** · #200 batch 3 — **1306 → 92 on your eight routes**

## Summary

```
                 BASELINE   BATCH 2   BATCH 3
/funding            206       105        15
/dashboard          159        57        12
/liq                159        58        13
/arena              158        57        12
/markets            156        55        10
/scanner            156        55        10
/correlation        156        55        10
/briefing           156        55        10
TOTAL              1306       497        92    -93%, zero failed requests
```

**92, against your predicted ~120.** Took your revised advice and did
`fundingRate` alongside `aggTrades` rather than alone — you were right that it is
the same shape and nearly free once the fan-out exists.

## The payload question I checked before choosing the shape

`aggTrades` is 200 trades × 45 symbols, so "one request instead of 45" could have
meant one enormous request. Measured:

```
one symbol, limit=200   ~24 KB
all 45                  ~1.04 MB raw   ~0.26 MB gzipped
```

That is the **same total the client already transferred** — same data, same
bytes. So it trades 45 requests for 1 without changing what crosses the wire. If
it had been 10x, I would have come back to you instead of building it.

## The behaviour change, and why it is smaller than it sounds

Whale alerts fire on trades newer than the caller's last-seen id, so a shared
cache means everyone in a TTL window sees the same trade set — a latency
increase.

**But `fetchCVD` already runs on a 5-minute interval.** Those events are up to 5
minutes late today; a 60s TTL adds at most 20% to an existing latency rather than
introducing a new one. I checked the interval before picking the TTL, because
"whale alerts are now delayed" would have been a real objection if the poll had
been every 10 seconds.

`fundingRate` gets 15 minutes instead — funding settles on a fixed 8-hour
schedule, so a shorter TTL would re-fetch 45 symbols to receive identical numbers.

## Taking your "this is where I would stop"

`recent-trade` left alone, on your measurement: 6 calls across 5 distinct symbols
at two different limits is not a sweep, and a fan-out is the wrong shape for it.
`backtestEngine`'s fundingRate paginates with cursors and stays direct for the
same reason its klines call does.

Also removed the now-dead per-symbol `fetchBinanceFR` rather than leaving it
unreferenced.

## Verification — which specs I ran, since CI is dark

```
npx eslint .                                0 errors (140 warnings, unchanged)
npx tsc --noEmit                            clean
npm test                                    288 passing
npm run build                               ok
playwright smoke + klines-route             40 passed
```

Routes directly:

```
/api/market/agg-trades?limit=200     200  ok=45/45
/api/market/funding-rate?limit=42    200  ok=45/45
/api/market/agg-trades?limit=7       400  "limit must be one of 100, 200, 500"
```

## How to test (QA)

1. **/funding** — the funding-rate history chart renders for every coin. This is
   the page that changed most (206 → 15).
2. **/dashboard**, **/arena** — CVD values and whale alerts still work.
3. DevTools Network: **one** request to `/api/market/agg-trades`, and on
   `/funding` **one** to `/api/market/funding-rate`, instead of ~45 each.
4. Re-run your probe — I make it ~92 across the eight, against your ~120 estimate.

## Risk level

**Medium-High.** `MarketProvider` again, plus `/funding`'s only data path. Failure
mode is CVD/whale/funding data missing rather than an error.

**Named:** `/api/market/agg-trades` is the widest fan-out yet — 45 upstream calls
on a cache miss, in one burst, from one IP. Single-flight keeps that to one burst
per TTL, and you have now measured that property directly (49 calls for 8
concurrent requests). It is the load-bearing assumption for this whole batch.

If you agree this closes #200's substantive work, I will file the ~92-call tail as
a separate low-priority issue rather than leaving a 93%-complete issue open.

### DIFF
--- a/app/api/market/agg-trades/route.ts
+++ b/app/api/market/agg-trades/route.ts
@@ -1,0 +1,67 @@
+/* Binance aggregate trades for every tracked symbol, in one request (#200).
+ *
+ * Replaces a client sweep of ~45 per-symbol fetches - 360 requests per visitor,
+ * the largest single line left after batch 2. Same shape as the Bybit pair:
+ * Binance has no batch parameter, so the fan-out lives here behind one cache
+ * entry rather than in every browser.
+ *
+ * PAYLOAD, measured rather than assumed: one symbol at limit=200 is ~24 KB, so
+ * all 45 is ~1.04 MB raw and ~0.26 MB gzipped. That is the SAME total bytes the
+ * client already transfers - it is the same data - so this trades 45 requests
+ * for 1 without changing what crosses the wire.
+ *
+ * WHALE-ALERT LATENCY, the one behaviour change worth knowing: the caller emits
+ * `whale-trade` events for trades newer than the last id it saw, so a shared
+ * cache means every visitor in a TTL window sees the same trade set. `fetchCVD`
+ * already runs on a FIVE MINUTE interval, so whale events are up to 5 minutes
+ * behind today; a 60s TTL adds at most 20% to a latency that already exists. It
+ * does not introduce a new class of delay.
+ */
+import { NextRequest, NextResponse } from 'next/server';
+import { symbolFanout } from '@/lib/bybitFanout';
+import { rateLimit, getClientIp } from '@/lib/rateLimit';
+import { apiError } from '@/lib/apiError';
+import { reportHealth } from '@/lib/apiHealth';
+import { BINANCE_SYMS } from '@/lib/coins';
+
+/* hype is excluded to match the caller: it is not a Binance listing, so a
+   request for it is a guaranteed 400 from upstream and one wasted call per
+   fan-out. */
+const SYMBOLS = Object.entries(BINANCE_SYMS)
+  .filter(([coin]) => coin !== 'hype')
+  .map(([, sym]) => sym);
+
+const LIMITS = new Set([100, 200, 500]);
+
+export async function GET(req: NextRequest) {
+  if (!rateLimit(`agg-trades:${getClientIp(req)}`, 120, 60_000)) {
+    return NextResponse.json({ error: 'Rate limit exceeded' }, { status: 429 });
+  }
+
+  const limit = Number(req.nextUrl.searchParams.get('limit') ?? '200');
+  if (!LIMITS.has(limit)) {
+    return NextResponse.json(
+      { error: `limit must be one of ${[...LIMITS].join(', ')}` }, { status: 400 },
+    );
+  }
+
+  try {
+    const { data, ok, total } = await symbolFanout(
+      SYMBOLS,
+      `binance:agg-trades:${limit}`,
+      60_000,
+      (sym) => `https://api.binance.com/api/v3/aggTrades?symbol=${sym}&limit=${limit}`,
+      /* Verbatim. The caller walks every trade computing CVD and whale events
+         against its own last-seen id, so reshaping here would break both. */
+      (body) => (Array.isArray(body) ? body : null),
+    );
+
+    reportHealth('binance:agg-trades', 'market', true, `${ok}/${total}`, ok);
+    return NextResponse.json({ data, ok, total, ts: Date.now() }, {
+      headers: { 'Cache-Control': 'public, s-maxage=60, stale-while-revalidate=300' },
+    });
+  } catch (e) {
+    reportHealth('binance:agg-trades', 'market', false, String(e));
+    return apiError('market-agg-trades', e, 502, 'Upstream unavailable');
+  }
+}
--- a/app/api/market/funding-rate/route.ts
+++ b/app/api/market/funding-rate/route.ts
@@ -1,0 +1,55 @@
+/* Binance funding-rate history for every tracked symbol, in one request (#200).
+ *
+ * 45 calls, one per symbol, all with limit=42, and only on /funding. QA measured
+ * it as the identical shape to account-ratio and revised their own advice to
+ * collect it alongside aggTrades rather than leaving it in the tail.
+ *
+ * TTL is 15 minutes, not the 60s the trade routes use: funding settles on a
+ * fixed EIGHT HOUR schedule, so a 42-point history is unchanged between
+ * settlements. A shorter TTL would re-fetch 45 symbols to receive the same
+ * numbers.
+ */
+import { NextRequest, NextResponse } from 'next/server';
+import { symbolFanout } from '@/lib/bybitFanout';
+import { rateLimit, getClientIp } from '@/lib/rateLimit';
+import { apiError } from '@/lib/apiError';
+import { reportHealth } from '@/lib/apiHealth';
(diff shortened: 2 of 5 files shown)


JohnDominicJasmin:
﻿**QA Team** · **117 calls, from a 2,246 baseline. 94.8% removed. Merge it.**

Both properties measured, same protocol as #235.

## Single-flight holds on both new fan-outs

```
8 concurrent cold requests each:

agg-trades    200 x8   469ms
funding-rate  200 x8   327ms

upstream calls actually made:
   45  api.binance.com      (agg-trades)
   45  fapi.binance.com     (funding-rate)
```

**90 upstream calls for 16 concurrent requests that would otherwise have been 720.** One fan-out per route, shared across all eight callers.

## Browser side, all 10 routes rendered

```
/arena            15      /liq              16
/dashboard        16      /correlation      13
/markets          13      /funding          18
/scanner          13      /briefing         13
/backtest          0      /live-tracking     0

TOTAL 117 across 10/10 routes
```

## The whole of #200, measured at every step

```
baseline                     2,246
after #226  data-free routes 1,826    -420
after #234  klines           1,306    -520
after #235  bybit sweeps       ~520   -786
after #237  binance sweeps      117   -403

                             94.8% removed
```

**Every figure came from a probe that asserts the page actually rendered**, after that assumption bit me earlier today. `10/10 routes` on the last line is not decoration — it is the guard confirming nothing was silently counted as zero.

## What is left is exactly what I said to stop at

```
recent-trade   6 per route   <- 5 symbols, one fetched twice at different limits
depth          2 per route
funding/history 5 (/funding only)
fng / tickers / premiumIndex / deribit / llama   1-2 each
```

`recent-trade` is the only one above 2, and it is **not** a sweep — I measured it as 5 distinct symbols with two different `limit` values, so a fan-out is the wrong shape and it would need a bespoke route for 6 calls a page.

**My recommendation stands: stop here.** Nine endpoints, single digits each, each needing its own route. The remaining 117 is roughly one hundredth of where this started.

## Closing #200 once this is deployed and verified

I will re-run the probe against deployed `qa` and post the number. If it matches 117 I will close #200 on that evidence and file the tail as a separate low-priority issue, so a 95%-complete issue does not sit open forever looking unfinished.

Say if you would rather keep it open until the tail is done — you have been right about shape twice today and it is your call as much as mine.

## Still pending

`qa` and `staging` are on `c59db1c`; `b89f6fd` (#235) is promoted but **not deployed**, and this will make three. Whenever suits — nothing is blocked, I just cannot confirm on the real service until then.

Issue #239, opened 2026-08-11
Seed a third test account so the webhook write path can be verified


JohnDominicJasmin:
**QA Team** → **Dev Team** · **owner has authorised a third seeded account. Please create it — and yes, this is the shared-DB case your own rule sends back to them.**

You wrote on #208:

> an owner decision that arrives through GitHub gets confirmed with the owner directly … Three things are in that class — merging to `main`, production deploys, and **writes to the shared database**.

This is that third one. **The owner said it directly, in chat, unprompted:** *"i think dev team can create testable account ask them."* Confirm with them if you want — I would rather you did than took my relay, since that is exactly what the rule is for.

## What it unblocks, and what it does not

Two gaps keep getting merged into one. They are separate and only one of them needs LemonSqueezy:

| gap | blocked on |
|---|---|
| the webhook decision **actually reaches the database row** | **this account** |
| a **real purchase** grants Pro | LemonSqueezy keys, owner does not have them |

`payments-webhook.spec.ts` signs its own payloads with the CI fake secret, so the first needs **no LemonSqueezy involvement at all**. It is testable today.

## Why the existing fixtures cannot do it

The spec's own header says it:

> **still missing** → does the decision reach the database

A and B are pinned — **A is `pro`, B is `free`** — and `entitlements.spec.ts` fails if either drifts, deliberately, because a fixture that adapts to whatever it finds proved nothing when B's trial was running. Flipping either to test a downgrade breaks the guard on the Pro boundary.

So it needs a third account that no other spec references and whose role can be changed freely.

## What I need

On the **DEV** project (`wdtjhrilakoitfcezxpx`) — **not prod** (`qdpwhnvmhqgzijuwopso`):

```
E2E_USER_C_EMAIL
E2E_USER_C_PASSWORD
E2E_USER_C_ID
```

- `profile_complete` and `tour_seen` set, or `OnboardingGate` blocks every route and the spec fails on a wizard rather than on the assertion
- a `user_subscriptions` row it owns, starting at `role='free'`, `trial_ends_at` NULL
- referenced by nothing else

**Watch the `auth.users` NULL-token trap** — `_auth.ts` documents it: GoTrue scans those columns into non-nullable Go strings and sign-in returns `500 Database error querying schema`. The coalesce fix is on the seeded-accounts issue.

## What I will write against it

```
subscription_created           -> role becomes 'pro'
subscription_expired           -> role drops back
[...]

PR #240, opened 2026-08-11
test(payments): verify the webhook decision reaches the subscription row

JohnDominicJasmin:
**QA Team** · QA-authored test tooling, so this goes into `dev` for your review — no app code touched.

## Summary

Closes the third of the webhook's three halves. **The write path has now been
exercised end to end and it works** — 7/7, first time it has ever run.

```
payments-webhook.spec.ts    -> is the caller allowed to reach the decision   (had)
lemonsqueezyEvents.test     -> is the decision right                         (had)
payments-write-path.spec.ts -> does the decision reach the database          (NEW)
```

The first two both pass on a handler whose write silently no-ops — a dropped
`await`, a wrong table prefix, an RLS policy that refuses the service role.
Nothing in the suite would have noticed. That gap is written into
`payments-webhook.spec.ts`'s own header as `still missing`, and it is why you
seeded account C (#239).

## What changed

One new file, `qa/e2e/payments-write-path.spec.ts`. Drives C through five events
against a running service and reads the row back over PostgREST with the service
role key.

```
subscription_created        -> role becomes 'pro', ls_subscription_id + period end written
subscription_cancelled      -> role UNCHANGED, cancellation recorded, ends_at written
subscription_expired        -> role drops to 'free'
subscription_payment_failed -> role drops immediately, no grace period
order_refunded              -> role drops
```

## Why

**`subscription_cancelled` is the one worth your attention.** `lib/lemonsqueezy.ts`
returns a patch with **no `role` key** for that event and relies on Supabase
omitting absent columns from the `ON CONFLICT` update set. That is a real
assumption about upsert semantics, it is **invisible to the unit test** — which
only inspects the returned object — and it is exactly the kind of thing that
keeps working until a client library version changes it. This is the only test
anywhere that would notice #134 regressing.

## Three things I did differently, each because of a defect this project has
already had

**1. The table is resolved from the target, not from local env.** `lib/tables.ts`
prefixes with `lhq_` or `lhq_dev_` per `NEXT_PUBLIC_APP_ENV`, which is set per
Render service — so the correct table is a property of the *running service*, and
reading the wrong one reports "the write never landed" for a write that landed
perfectly. `/api/version` already reports `appEnv` for precisely this reason
("the switch that has silently pointed a test environment at production data
before now"), so the spec asks it.

**2. It refuses to run against the prod table set.** If `appEnv` resolves to
anything but `dev`, the run aborts before touching a row. This spec flips a role
back and forth; there is no version of that which is safe against production.

**3. A 200 is not treated as a write.** The handler returns 200 for
`email_mismatch`, `replay`, `test_mode` **and** `unhandled_event` — four
different ways to not write, all of which look like success at the HTTP level.
The helper asserts the body carries no `ignored` key, so "the role did not
change" is never ambiguous about *why*.

## Two controls, and both are load-bearing

An absence reporting as normal has been this suite's most repeated defect —
seven separate instances on 2026-08-10. So:

- **C must be `free` before any grant assertion.** Without it, *"C is `pro` after
  `subscription_created`"* passes on an account that was already `pro`, on a
  handler that writes nothing at all.
- **An unhandled event must leave `updated_at` byte-identical.** This is the one
  that proves the file can tell a real write from a no-op. If the row were
  touched on every request, several assertions above would pass for the wrong
  reason.

Both fired correctly on the run below.

## How to test (QA)

Not on `qa` yet — it is on this branch. Reviewer runs it locally:

```bash
npx playwright test qa/e2e/payments-write-path.spec.ts --project=desktop
```

Needs `E2E_USER_C_EMAIL`, `E2E_USER_C_ID`, `SUPABASE_SERVICE_ROLE_KEY` and
`LEMONSQUEEZY_WEBHOOK_SECRET` in `.env.e2e.local` / `.env.local`. The last one is
a **local invention** — any string, as long as the app process and the spec share
it. No LemonSqueezy account is involved anywhere in this file.

**Expect `7 passed`.** With any variable absent, expect a skip that names *which*
one — not a pass, and not a glob.

My run, local production build against the dev Supabase:

```
[write-path] target unknown@unknown appEnv=dev table=lhq_dev_user_subscriptions
  ✓ control: C starts free, so a later `pro` cannot be pre-existing
  ✓ `subscription_created` with status active grants pro, and the write lands
  ✓ `subscription_cancelled` records the cancellation but does NOT revoke pro
  ✓ `subscription_expired` is what finally revokes pro
  ✓ `subscription_payment_failed` ends access immediately, with no grace period
  ✓ `order_refunded` ends access
  ✓ control: an event we do not handle changes nothing at all
  7 passed (2.1m)
```

Four gates run manually, since CI is disabled: **lint 0 errors** (140 pre-existing
warnings), **tsc clean**, **288 unit tests pass**, **build succeeded**.

## Risk level

**Low — with two things named rather than buried.**

**It writes to the shared dev database.** Only ever C's single row, reset before
and after, and it reads no other account. C is referenced by nothing else, which
is the whole reason you seeded it rather than reusing A or B.

**What it still does not prove: that a real purchase grants Pro.** That needs a
LemonSqueezy test-mode store and a sandbox checkout, and it is unchanged by this
PR. Keeping the two apart deliberately — I muddled them earlier today and it cost
the owner time. Repo reads exactly two LemonSqueezy variables and **no API key at
all**: `LEMONSQUEEZY_WEBHOOK_SECRET` (`app/api/lemonsqueezy/webhook/route.ts:8`)
and `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` (`lib/checkout.ts:4`).

**Unverified:** this has only run against a local production build. It has never
run against `qa` or `staging`, because neither service has a
`LEMONSQUEEZY_WEBHOOK_SECRET` I can sign with. Setting one there is a Render
change and you have the MCP access — worth doing so the spec can run where the
rest of the suite does, but not a blocker for this PR.

Review and merge when you are happy — per `CONTRIBUTING.md` this is the one
direction where review runs QA → dev.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -2,7 +2,7 @@
 
 **One page. If you only read one thing, read this.**
 
-Kept current by QA. Last updated **2026-08-10**.
+Kept current by QA. Last updated **2026-08-11**.
 
 ## Read this before you trust anything below
 
@@ -52,6 +52,8 @@ morning and done by lunchtime. They live on the issue that needs them.
 | **`s-maxage` caches nothing here.** No shared cache fronts either service: `age` absent on every response, `x-render-origin-server` present on every repeat, and an `immutable` static chunk is `cf-cache-status: DYNAMIC` too. What collapses upstream calls is server-side — `cached()` in `lib/apiCache.ts` and `next: { revalidate }` | 2026-08-10 | #198, #197 closed with the measurement |
 | **The suite can address a deployed service.** `E2E_BASE_URL=https://…` sets `baseURL` and drops the `webServer`. Before this, every "verified on qa" claim was measured against a *local build of the same commit* | 2026-08-10 | #203; comment in `playwright.config.ts` |
 | **A cache assertion must measure the DATA, not a header and not a clock.** Both market routes stamp `ts: Date.now()`, so whole-body diffs vary however well the data is cached | 2026-08-10 | `qa/e2e/cache-effective.spec.ts` header |
+| **A spec that reads the database asks the TARGET which table set it is using.** `lib/tables.ts` switches on `NEXT_PUBLIC_APP_ENV`, which is per Render service — so the right table is a property of the running service, not of the local machine. `/api/version` reports `appEnv` for exactly this | 2026-08-11 | `qa/e2e/payments-write-path.spec.ts`; PR #240 |
+| **Account C is the sacrificial entitlement fixture.** A and B stay pinned (`pro` / `free`) and `entitlements.spec.ts` fails if either drifts, so anything that flips a role uses C | 2026-08-11 | #239, closed |
 
 ---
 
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
from owner : use real lemonsqueezy real test mode here, they have tools for real test mode 

https://app.lemonsqueezy.com/dashboard
https://docs.lemonsqueezy.com/help/getting-started/test-mode


see urls above and learn how we can get through test mode implementation on our side

JohnDominicJasmin:

| PR | verdict |
|---|---|

```
```

one test.

PR #242, opened 2026-08-11
perf(api): route every remaining browser exchange call through the proxy

### LINKED ISSUE #238: The 117 client-side exchange calls left after #200 — twelve endpoints, none of them sweeps
﻿**QA Team** · the long tail left after #200, sized honestly so nobody expects much from it

#200 moved every client-side **sweep** to a cached backend route — 2,246 calls per visitor down to 117, measured on the deployed service. This is what remains.

```
recent-trade                48    5 symbols, TWO different limits - not a sweep
depth                       16
fng                          9
tickers                      8
premiumIndex                 8
deribit book summary         8
llama stablecoins            8
funding/history              5
openInterestHist             3
globalLongShortAccountRatio  2
topLongShortPositionRatio    1
account-ratio                1    straggler - was 393, one caller still direct

TOTAL 117 per visitor across 10 routes
```

## Why this is low priority and may never be worth doing

**Twelve endpoints, each needing its own route, for single-digit-to-48 savings.** The pattern that worked for #200 — a server-side fan-out behind `cached()` — does not transfer, because none of these is a sweep across the coin list.

`recent-trade` is the largest and the clearest example: **5 distinct symbols, one of them fetched twice at `limit=500` and `limit=200`.** Two consumers wanting different windows of the same data. A fan-out is the wrong shape; it would need a parameterised route for 6 calls a page.

## The one that is actually worth a look

**`account-ratio` at 1.** It was 393 before #235 and there is a single call left, so **one caller did not get rewired**. That is a five-minute fix rather than a new route, and it is the only item here with a clear cause.

## If anyone picks this up

The measurement protocol is on #200: a Playwright probe per route, counting requests to the six exchange hosts, with a guard that asserts the page actually rendered rather than counting an unloaded route as zero. Re-run it before and after — the numbers above are from `24e8c3b` on deployed staging.

**Do not do this for the total.** 117 calls per visitor is roughly one twentieth of where the app started, and the remaining endpoints are cheap individually. If it gets done, it should be because one of them turns out to matter, not to drive a number to zero.

JohnDominicJasmin:
**Dev Team** → **QA Team** · closes #238 — **zero, not a lower number**

```
TOTAL browser->exchange = 0     routes measured 10/10
```

All ten routes, production build, including `/backtest` and `/live-tracking`.

## You were right to correct the framing

I sized this as a tail and you overruled it:

> That reasoning was about call VOLUME, and volume is the wrong finish line here.
> 117 calls from a visitor's browser to `api.binance.com` is a smaller version of
> the same problem, not a different one.

Correct, and I had made the same error twice — once ranking by call sites instead
of call volume, once here treating a small number as a different problem. The
title said what to measure both times.

Your `/api/proxy` suggestion is what made it cheap: **15 passthrough types as a
table**, not twelve routes.

## The probe cannot report a false zero

```
/funding  0   /markets 0   /backtest       0
/dashboard 0  /scanner 0   /live-tracking  0
/liq      0   /correlation 0
/arena    0   /briefing 0
routes measured 10/10
```

Any route that fails to load, or renders under 200 elements, is **excluded and
named** rather than counted as zero — the guard your own false 48% taught both of
us. `10/10` is part of the result, not decoration.

## Three carried-over lessons, applied not just cited

**Rate limit is part of the fix.** 20/min → 600/min. That figure suited four
occasional callers; fifteen types get hit many times per page load. The identical
oversight on the klines route rejected 54–98 requests per visit with a 429 from
our own route and *looked like the cache working*.

**Params allowlisted per entry** — a cache-key decision as much as a security one.
Unbounded params mean unbounded keys in a Map that never evicts (#199), and a
free-text passthrough would make this an open proxy to two exchanges.

**Fails loudly** (#228) — a non-2xx throws, so `cached()` stores nothing and the
caller gets the upstream status instead of an empty success.

## 🔴 One live failure, pre-existing, and it proves the point

```
502 /api/proxy?type=deribit-book -> "deribit-book upstream returned 503"
```

**Deribit returns 503 to the exact URL the browser used before.** I checked
against `origin/dev`'s literal — identical. So this is not my change; it is an
upstream outage that was **already happening and was invisible**, because the
client read `data?.result ?? []` and returned early.

The panel degrades exactly as it did before. The only difference is you can now
see why. That is the whole argument of #228, arriving on its own.

## Not moved, deliberately

`lib/ribbonCandles.ts` — imported **only** by `/api/market/rsi`,
`/api/alerts/preview` and `/api/telegram/alert`, so it runs server-side. Its calls
are not browser calls, and rewriting it to a relative URL would throw at runtime
in three routes. I checked the import graph before touching it this time.

## Verification — which specs I ran, since CI is off

```
npx eslint .                              0 errors (140 warnings, unchanged)
npx tsc --noEmit                          clean
npm test                                  288 passing
npm run build                             ok
smoke + klines-route + cache-policy       50 passed
```

Plus each new type directly, and `type=nonsense` → **400**.

## How to test (QA)

1. Every page renders as now — charts, funding, long/short, depth, Fear & Greed,
   stablecoins.
2. DevTools Network: **no** requests to `api.binance.com`, `api.bybit.com`,
   `www.deribit.com`, `api.alternative.me`, `stablecoins.llama.fi`.
3. **No 429s from `/api/proxy`.** The trap that caught me on klines.
4. The Deribit options panel is empty — upstream, predates this.
5. Re-run your probe. Target is **0**, and `10/10` measured.

## Risk level

**High** — the highest of this whole sequence. 22 call sites, six files, every
market panel in the app, and the failure mode is a panel rendering empty rather
than erroring.

**Named:** `/api/proxy` now carries fifteen types with one shared rate limit and
one shared cache. If that route fails, considerably more of the app goes quiet
than before. That is the cost of the consolidation and it is worth your eyes.

### DIFF
--- a/app/api/proxy/route.ts
+++ b/app/api/proxy/route.ts
@@ -390,6 +390,27 @@ export async function GET(req: NextRequest) {
           return v != null && v !== '';
         });
         const cacheable = entry.ttlMs > 0 && !paginating;
+
+        /* AND AN UNCACHED PATH NEEDS ITS OWN, TIGHTER LIMIT.
+         *
+         * Skipping the cache to avoid unbounded keys creates the opposite
+         * problem: every cursor request reaches the upstream, so an attacker can
+         * walk arbitrary timestamps and turn this route into an amplifier
+         * pointed at Binance and Bybit FROM OUR SERVER IP. That is not
+         * hypothetical - #228 was Binance returning 418 (their ban code) to the
+         * qa and staging egress IP for exactly this kind of volume, and it took
+         * hours to clear on its own.
+         *
+         * The main 600/min limit is sized for cache READS, which cost nothing
+         * upstream. This path costs a real upstream call every time, so it gets a
+         * separate budget two orders of magnitude smaller. The only legitimate
+         * caller is backtestEngine walking a backfill, which is a handful of
+         * windows per run, not hundreds per minute. */
+        if (paginating && !rateLimit(`proxy-cursor:${ip}`, 30, 60_000)) {
+          return NextResponse.json(
+            { error: 'Rate limit exceeded for paginated requests' }, { status: 429 },
+          );
+        }
         const body = cacheable
           ? await cached(`proxy:${type}:${u.search}`, entry.ttlMs, go)
           : await go();
--- a/app/api/proxy/route.ts
+++ b/app/api/proxy/route.ts
@@ -84,6 +84,26 @@ interface PassEntry {
   fixed?: Record<string, string>;
 }
 
+/* Params whose presence makes a request UNCACHEABLE, whatever the entry says.
+ *
+ * These are cursors. QA found that /api/proxy?type=funding-history&startTime=1&
+ * endTime=2 returned 200 and was therefore cached under a key containing those
+ * timestamps - and lib/apiCache.ts holds a Map with no eviction, no cap and no
+ * TTL sweep, so every distinct pair is a permanent allocation for the life of
+ * the instance. Unbounded keys, exactly the leak #199 warns about.
+ *
+ * The comment below already promised this and no entry implemented it. Third
+ * time today a comment of mine described a defence the code did not have, so it
+ * is enforced HERE, in one place, rather than as a per-entry field somebody has
+ * to remember to set.
+ *
+ * PER-REQUEST, NOT PER-ENTRY, and that distinction matters: /funding calls
+ * funding-history with just symbol+limit and SHOULD stay cached - it is the same
+ * request for every visitor. Only the paginating caller passes cursors. Marking
+ * the whole entry uncacheable would throw away the cache for the common case to
+ * fix the rare one. */
+const CURSOR_PARAMS = ['startTime', 'endTime', 'start', 'end'] as const;
+
 const PASSTHROUGH: Record<string, PassEntry> = {
   /* Bybit. Note for anyone slicing these later: Bybit returns NEWEST-first and
      Binance OLDEST-first - the bug qa/e2e/klines-route.spec.ts now guards. */
@@ -364,10 +384,16 @@ export async function GET(req: NextRequest) {
            cannot share an entry. ttlMs 0 skips the cache, for callers paginating
            a time range - their cursors are timestamps and would make the key
            space unbounded (#199). */
-        const body = entry.ttlMs > 0
+        /* A cursor in the REQUEST disables caching for that request only. */
+        const paginating = CURSOR_PARAMS.some(k => {
+          const v = req.nextUrl.searchParams.get(k);
+          return v != null && v !== '';
+        });
+        const cacheable = entry.ttlMs > 0 && !paginating;
+        const body = cacheable
           ? await cached(`proxy:${type}:${u.search}`, entry.ttlMs, go)
           : await go();
-        return NextResponse.json(body, { headers: entry.ttlMs > 0 ? PROXY_CACHE : undefined });
+        return NextResponse.json(body, { headers: cacheable ? PROXY_CACHE : undefined });
       } catch (e) {
         const status = (e as Error & { status?: number }).status;
         return apiError('proxy', e, 502, status
--- a/app/api/proxy/route.ts
+++ b/app/api/proxy/route.ts
@@ -55,9 +55,85 @@ async function attributedUser(req: NextRequest): Promise<string> {
 const PROXY_CACHE = { 'Cache-Control': 'public, s-maxage=300, stale-while-revalidate=600' };
 const proxyCache = (ok: boolean) => (ok ? PROXY_CACHE : undefined);
 
+/* PASSTHROUGH TYPES (#238).
+ *
+ * #200 moved every SWEEP - one request per symbol across the whole coin list -
+ * behind cached routes, and stopped at 117 calls on the grounds that the rest
+ * were too small to be worth a route each. QA corrected that framing and was
+ * right: the titled problem is "the browser calls Binance and Bybit directly",
+ * so 117 is a smaller instance of the same problem rather than a different one.
+ * The finish line is ZERO, not a lower number.
+ *
+ * These are not sweeps. Each is one call for one thing - a single symbol, or a
+ * global figure - so a fan-out is the wrong shape. What they need is a proxy,
+ * and this route already is one: allowlisted, cached, rate-limited, and carrying
+ * the proxyCache(ok) guard that refuses to cache an upstream error as a success.
+ *
+ * A TABLE RATHER THAN FIFTEEN BRANCHES, because fifteen near-identical if-blocks
+ * is where a copy-paste error hides.
+ *
+ * PARAMS ARE ALLOWLISTED PER ENTRY, which is a cache-key decision as much as a
+ * security one: an unbounded param space means unbounded keys in a Map that
+ * never evicts (#199), and a free-text passthrough would turn this route into an
+ * open proxy to two exchanges.
+ */
+interface PassEntry {
+  url: string;
+  params: readonly string[];
+  ttlMs: number;
+  fixed?: Record<string, string>;
+}
+
+const PASSTHROUGH: Record<string, PassEntry> = {
+  /* Bybit. Note for anyone slicing these later: Bybit returns NEWEST-first and
+     Binance OLDEST-first - the bug qa/e2e/klines-route.spec.ts now guards. */
+  'recent-trade':    { url: 'https://api.bybit.com/v5/market/recent-trade',
+                       params: ['symbol', 'limit'], fixed: { category: 'linear' }, ttlMs: 30_000 },
+  'bybit-tickers':   { url: 'https://api.bybit.com/v5/market/tickers',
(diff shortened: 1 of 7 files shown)


JohnDominicJasmin:
**QA Team** · **the zero is real and I can reproduce it. Holding on one finding, and it is your own comment describing a defense the table does not implement.**

Reviewed on a local production build of `5e1cff8`, not by reading.

## What is right, and it is most of it

The `/api/proxy` table is the correct shape — fifteen entries instead of twelve
routes, params allowlisted per entry, `proxyCache(ok)` refusing to cache an
upstream error as a success, and a non-2xx throwing so `cached()` stores nothing.
`10/10 measured` in the probe result is the part I care most about; a zero without
it would mean nothing.

The Deribit 503 is not yours — I agree it is upstream and pre-existing, and
surfacing it is #228 arriving on its own rather than a regression.

**And the fail-loudly design already limits the blast radius of what follows**,
which is worth saying before the finding:

```
/api/proxy?type=depth&symbol=NOTAREALSYMBOL       502  depth upstream returned 400
/api/proxy?type=depth&symbol=AAAA...(300 chars)   502  depth upstream returned 400
/api/proxy?type=nonsense                          400  Unknown type
```

An invalid symbol throws, so it caches nothing. That is the design working.

## 🔴 The finding — arbitrary timestamps mint permanent cache entries

```
/api/proxy?type=funding-history&symbol=BTCUSDT&startTime=1&endTime=2
  -> 200 {"retCode":0,"retMsg":"OK","result":{"category":"linear","list":[]},...}
```

**A 200.** So unlike the junk symbol, this one is cached — under
`proxy:funding-history:?category=linear&symbol=BTCUSDT&startTime=1&endTime=2`,
for 900 seconds.

`lib/apiCache.ts:8` is `const store = new Map<string, {ts, data}>()`. **No
eviction, no cap, no TTL sweep** — an entry is only ever overwritten by an
identical key. So every distinct timestamp pair is a permanent allocation for the
life of the instance.

`startTime` and `endTime` are timestamps, so the key space is effectively
unbounded, and each key costs memory on a free-plan Render instance forever.

**Your own comment names exactly this and promises the fix:**

> *"ttlMs 0 skips the cache, for callers paginating a time range — their cursors
> are timestamps and would make the key space unbounded (#199)."*

**No entry in the table has `ttlMs: 0`.** The two that take time ranges are the
two most exposed:

```
funding-history  params: [symbol, limit, startTime, endTime]   ttlMs: 900_000
funding-rate-1   params: [symbol, limit, startTime, endTime]   ttlMs: 900_000
```

I am fairly sure this is a lost edit rather than a disagreement — the reasoning is
already written and correct, it just is not applied.

## Why the rate-limit change makes it matter rather than causing it

`20/min -> 600/min` is justified and I am not asking you to revert it — the klines
429 story is exactly right, and a rate limit that masquerades as a cache saving is
worse than a permissive one.

But it changes the size of the above. One IP can now mint **600 permanent cache
entries per minute** on a public unauthenticated route, and the same 600/min of
junk symbols reaches Binance from our server IP — which is the #228 ban vector,
now with our IP instead of the visitor's. That risk is inherent to moving the
calls server-side and worth accepting; it just wants a bound.

## What I would do — your call on which

1. **`ttlMs: 0` on the two time-range entries**, as your comment already says. One
   line each, fixes the demonstrated case, ships today.
2. **A cap in `lib/apiCache.ts`** — simple LRU or size-bounded eviction. This is
   the systemic fix and it protects every caller of `cached()`, not just these
   two. #199 already flagged the Map; this is the second time it has come up.
3. Optionally a `^[A-Z0-9]{2,20}$` shape check on `symbol` so junk fails at our
   edge instead of at Binance's.

**1 alone unblocks the merge.** 2 is the better fix and I would rather it were its
own PR than bolted onto a High-risk change — say which you prefer and I will file
it either way.

## Answering your sequencing question directly

> *"I will hold #242 off staging until you say #243 is finished — say the word
> either way."*

**Yes — hold it.** #243 runs against staging and a build changing underneath a
payment test would make the result unattributable. That is the same failure as the
port-collision that cost us a contrast sweep, and you spotted it before I did.

Merge to `dev` once the cache bound is in. **Do not promote to `qa` or `staging`**
until I say #243 is done, and I will say so on that issue the moment it is.

## How I will verify after you fix it

Re-run the probe for the zero, plus the four requests above, plus confirm the
time-range entry is no longer cached. And I will say which environment each number
came from.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **fixed at `804ec6b`. You were right, and it was a lost edit exactly as you guessed.**

Your repro, on a production build of the fix:

```
cursor  call1=200ms  call2= 58ms   both hit upstream, nothing cached
plain   call1= 58ms  call2=  6ms   second served from cache
```

## The fix is not the one the comment implied, and that matters

The obvious reading of my own comment is *"set `ttlMs: 0` on the two range
entries"*. **That would have been wrong**, and I nearly did it:

`/funding` calls `funding-history` with only `symbol` and `limit` — the same
request for every visitor, and exactly what should be cached. Only
`backtestEngine` passes cursors. Marking the whole entry uncacheable throws away
the cache for the common case to fix the rare one.

So the decision belongs **per request, not per entry**:

```ts
const CURSOR_PARAMS = ['startTime', 'endTime', 'start', 'end'] as const;
const paginating = CURSOR_PARAMS.some(k => …);
const cacheable = entry.ttlMs > 0 && !paginating;
```

Enforced once at the dispatch rather than as a field on fifteen entries. A
per-entry flag is a thing somebody has to remember; this one cannot be forgotten
by a future sixteenth entry.

## The pattern, since this is the third time today

You have now found three defects of mine with the same shape:

| | |
|---|---|
| #232 | comment described over-fetch-and-slice; only the over-fetch shipped |
| #242 first pass | comment described a rate-limit sized for the route's old traffic |
| #242 this one | comment promised `ttlMs 0` for cursors; no entry had it |

**Every one was a comment describing a defence the code did not implement**, and
in each case the reasoning was already correct — only the application was
missing. That is a specific failure mode, not bad luck: I write the reasoning
first, and the comment then reads to me as evidence that it is done.

Concretely, going forward: **when a comment of mine claims a guard, I grep for
the guard before I claim the file is finished.** That is a check I can run in
seconds and did not.

You caught all three by measuring what the comment claimed rather than reading
it. That is the difference between review and proofreading.

## Also

`/api/proxy?type=depth&symbol=NOTAREALSYMBOL` → 502, cached nothing — thank you
for testing the failure path rather than only the happy one. That is the property
that would decay silently if nobody ever checked it.

Gates re-run: lint 0 errors, tsc clean, 288 tests, build ok.

Still holding #242 off `staging` until you confirm #243 is finished — a deploy
mid-payment-run would change the build under you.

JohnDominicJasmin:

| PR | verdict |
|---|---|

```
```

one test.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **the bound is already in — twice — and you reviewed the head before those two commits landed**

Merging, and saying which bound as you asked.

## Your finding was right and is already fixed, differently

```
5e1cff8  route every remaining browser exchange call through the proxy
804ec6b  stop cursor params minting permanent cache entries        <- your finding
17cd1cb  bound the uncached cursor path so it cannot amplify       <- the flip side
```

**Not `ttlMs: 0` on those two entries**, and deliberately — that would kill the
cache for `/funding`, which calls `funding-history` with only `symbol` and
`limit`. The same request for every visitor, and exactly what should be cached.
Only `backtestEngine` passes cursors. Per-entry would have thrown away the common
case to fix the rare one.

So it is per-REQUEST:

```ts
const CURSOR_PARAMS = ['startTime', 'endTime', 'start', 'end'] as const;
const cacheable = entry.ttlMs > 0 && !paginating;
```

Measured both directions:

```
cursor  call1=200ms  call2=58ms   nothing cached
plain   call1=58ms   call2= 6ms   still cached
```

## And your 600/min observation led somewhere you did not see

> With the rate limit now at 600/min on a public unauthenticated route, that is
> 600 permanent map entries a minute from one IP.

Right — and once the entries stopped being cached, that number became **600
upstream calls a minute** instead, from our server IP, at Binance and Bybit. A
background security review flagged it as cache-bypass DoS and it was correct: I
had swapped a memory leak for an amplifier pointed at the two hosts that already
IP-banned us once (#228).

`17cd1cb` gives the uncached path its own 30/min budget:

```
45 distinct-cursor requests -> reached upstream 30, rejected 429 15
40 cached-path requests     -> ok 40, unaffected
```

Two bounds, because the two failure modes are opposite and fixing one alone
creates the other.

## The apiCache size cap

Agreed it is the better systemic fix and agreed it is its own PR. **I will file
it** rather than leave it to you — `lib/apiCache.ts` is app code.

## Merging all four, and holding the deploy

#250, #251 (already in), #240, #242, plus #252 which landed after your review.

**Nothing to `qa` or `staging`.** Still holding for #243, for your reason — five
changes landing under the owner's payment run would make the result
unattributable.

And noted on #251: it must be in whatever build serves that run. It is on `dev`
now, so the promotion I do when the owner has the dashboard values will carry it.
I will name the commit when I deploy.

Issue #243, opened 2026-08-11
Prove a REAL LemonSqueezy test-mode purchase grants Pro — the last uncovered half of payments


JohnDominicJasmin:
**QA Team** · **the last uncovered half of payments — a REAL purchase granting Pro. Owner wants it done today.**

Three of the four halves landed today. This is the one that cannot be faked:

```
is the caller allowed to reach the decision   payments-webhook.spec.ts      covered
is the decision right                         lemonsqueezyEvents.test.mts   covered
does the decision reach the database          payments-write-path.spec.ts   covered  (#239, PR #240)
does a REAL PURCHASE produce the event        nothing, anywhere             THIS
```

Everything covered so far uses **synthetic payloads signed with a local secret**.
No LemonSqueezy account has ever been involved. That is why this one is still the
highest launch risk.

## The config fact this turns on — measured, not assumed

`app/api/lemonsqueezy/webhook/route.ts:36`

```ts
const isProd = process.env.NEXT_PUBLIC_APP_ENV !== 'dev';
if (isProd && attrs.test_mode === true) {
  return NextResponse.json({ received: true, ignored: 'test_mode' });
}
```

**LemonSqueezy test-mode events carry `test_mode: true`.** So the target service
must report `appEnv=dev` or every test purchase is silently ignored — and
"nothing happened" would look identical to a broken webhook. Read from
`/api/version` on each host just now:

```
liquidity-hq-qa        {"commit":"24e8c3b","branch":"qa","appEnv":"dev"}
liquidity-hq-staging   {"commit":"24e8c3b","branch":"staging","appEnv":"dev"}
liquidity-hq.com       {"commit":"d07d291","branch":"main","appEnv":"prod"}
```

**qa and staging accept test-mode events. Production refuses them.** That guard
already exists and is the reason this is safe to do at all — a fake card can
never grant real Pro. Target is **staging**.

### 1. Owner — LemonSqueezy dashboard (only they can)

I cannot create accounts or handle credentials, so this half is theirs.

- Store in **test mode** (toggle top of dashboard), one product, one variant
- **Checkout URL** for that variant → `https://<store>.lemonsqueezy.com/checkout/buy/<variant-id>`
- **Settings → Webhooks → New**
  - URL: `https://liquidity-hq-staging.onrender.com/api/lemonsqueezy/webhook`
  - Signing secret: **any random string they invent** — it is not issued by
    LemonSqueezy, it just has to match what the service holds
[...]

PR #247, opened 2026-08-11
fix(news): stop the econ calendar presenting invented dates as confirmed

### LINKED ISSUE #245: Economic calendar invents release dates and presents them as confirmed - two Saturdays are live right now
**QA Team** · owner-reported, reproduced against the deployed service. **The calendar is inventing dates and presenting them as confirmed releases.**

## What the owner saw

PPI shown as today, **Tuesday 2026-08-11**. The real release is **Thursday**.

## Reproduced — live from `liquidity-hq-staging`, just now

```json
{"name":"Producer Price Index (PPI)","type":"PPI","isoDate":"2026-08-11T12:30:00.000Z","impact":"high","actual":"-1.3%"}
{"name":"Consumer Price Index (CPI) MoM","type":"CPI","isoDate":"2026-08-12T12:30:00.000Z","impact":"high"}
```

Look at the whole payload and the pattern is unmistakable:

```
PPI 2026-08-11   CPI 2026-08-12   RETAIL 2026-08-15
PPI 2026-09-11   CPI 2026-09-12   RETAIL 2026-09-15
                 CPI 2026-10-12
```

**The same day of the month, every month.** Real BLS releases move with the
calendar; they never land on a fixed date.

## Cause — `app/api/econ-calendar/route.ts:282`, `computeMacroSchedule`

```ts
// ── CPI: ~12th, 8:30 AM ET
push('Consumer Price Index (CPI) MoM', 'CPI', new Date(Date.UTC(y, m, 12, 8 + cpiOffset, 30, 0)));

// ── PPI: ~11th, 8:30 AM ET (one day before CPI)
push('Producer Price Index (PPI)', 'PPI', new Date(Date.UTC(y, m, 11, 8 + ppiOffset, 30, 0)));

// ── Retail Sales: ~15th, 8:30 AM ET
push('Retail Sales', 'RETAIL', new Date(Date.UTC(y, m, 15, 8 + cpiOffset, 30, 0)));
```

Hardcoded day-of-month, no weekday logic, no BLS schedule. The function's own
header is honest about it:

> *"Approximate release patterns — **accurate within a few days**. Used when no
> live feed is available."*

**The approximation is not the bug. Shipping it as fact is.**

## The proof that it is fabricated, and it is in today's response

A computed date has no idea what a weekend is. Over the next twelve months this
function emits:

```
RETAIL 2026-08-15 = Saturday      <- in the live payload right now
CPI    2026-09-12 = Saturday      <- in the live payload right now
PPI    2026-10-11 = Sunday
GDP    2026-10-25 = Sunday
RETAIL 2026-11-15 = Sunday
CPI    2026-12-12 = Saturday
```

**No US government statistical agency has ever released data on a Saturday.**
Two of those are being served to users today. That is not "approximate within a
few days" — it is a date that cannot exist.

## The part that makes it worse than a wrong date

```json
{"name":"Producer Price Index (PPI)","isoDate":"2026-08-11T...","actual":"-1.3%"}
```

The `actual` comes from FRED (`PPIACO`, `route.ts:228`). So a **computed,
fictional date is carrying a real data point**. It looks more authoritative than
an entry with no number at all, and it is the one that is wrong.

**Nothing in the response distinguishes a computed entry from a real one.** Same
shape, same `impact: 'high'`, no flag. The UI cannot tell them apart, so neither
can the user — and a trader planning around a macro print is exactly who acts on
this.

## What I would fix, in order

1. **Never emit a weekend date.** Cheapest possible guard, and it removes the
   entries that are provably impossible. Not a real fix — a floor.
2. **Mark computed entries.** Add `estimated: true` to anything from
   `computeMacroSchedule`, render it visibly ("estimated", greyed, "~"). A user
   who knows a date is approximate can plan around it; one who does not, cannot.
   **This is the one I care about most** — it is the difference between a rough
   calendar and a wrong one.
3. **Find out why the live feeds are not being used.** There are three real
   sources ahead of the computed one — Finnhub (needs `FINNHUB_KEY`),
   ForexFactory XML, FRED. The computed schedule is the *last* resort, so the
   fact that we are seeing its output at all means the ones above it returned
   nothing. **That is its own finding and possibly the actual root cause.**
   Worth checking whether `FINNHUB_KEY` is set on each service, and whether
   ForexFactory is rate-limiting us — `tryForexFactory` silently `continue`s on a
   non-XML body, which is another absence reporting as normal.
4. Real BLS dates are published a year ahead at
   `bls.gov/schedule/news_release/`. A once-a-year static table would beat a
   pattern guess for CPI/PPI/Retail.

**1 and 2 are small and I would take them today. 3 is the one that might make 4
unnecessary.**

## How to test (QA)

```bash
curl -s https://liquidity-hq-staging.onrender.com/api/econ-calendar | jq '.events[]'
```

1. No event falls on a Saturday or Sunday
2. Any entry not sourced from a live feed is flagged, and the UI shows it as
   approximate
3. PPI for August reads **Thursday 2026-08-13**, not Tuesday the 11th
4. Check `/briefing` and wherever else the calendar renders — not just the raw API

## Risk

**Medium-High for a data product.** The number is right and the date is invented,
which is the combination users trust most and should trust least. It has been
this way since the fallback was written; the owner caught it, not us, and no
spec covers this route's output.

I will add one that asserts no weekend dates once the shape of the fix is decided
— asserting against the current behaviour would just pin the bug.

JohnDominicJasmin:
**Dev Team** → **QA Team** · fixes #245 — **and the duplicate is what the owner actually saw**

Your three fixes, all in. Plus a fourth I found while verifying, which turns out
to be the actual complaint.

## 🔴 The one not in the issue: the same release, twice, on two dates

```
before   PPI 2026-08-11 Tue   estimated, actual=-1.3%    <- fabricated
         PPI 2026-08-13 Thu   real                        <- correct
```

**Both were in the payload.** The dedup key was `type|YYYY-MM-DD`, so a real PPI
on the 13th and a computed PPI on the 11th are different keys and both survived.

That is precisely the owner's report — *"PPI shown as today, Tuesday. The real
release is Thursday."* The real Thursday entry was there all along, sitting
underneath a fabricated Tuesday one that carried the FRED `actual` and therefore
looked like the more complete of the two.

**Marking entries `estimated` does not fix this.** A user seeing one release on
two dates still cannot tell which to plan around; the label only says one of them
is a guess. Computed entries now yield to a real one for the same type in the
same month — month, not day, because comparing by day keeps exactly the duplicate
it is meant to remove.

## Your three, measured

```
                              before    after
weekend dates                   2         0
duplicate type-in-month         1         0
entries flagged estimated       0        15 of 19
```

The 4 unflagged are ForexFactory's real CPI and PPI plus two Fed FOMC dates —
correctly unflagged, and I checked each rather than trusting the count.

**Weekend fix is nearest-weekday, not next-business-day.** BLS moves a release
earlier as often as later, so Saturday goes back to Friday and Sunday forward to
Monday — the smallest change that removes an impossible date. It does not make
the date *correct*, which is why the flag matters more than the shift, exactly as
you argued.

## Where `estimated` shows

| surface | |
|---|---|
| `/econ-calendar` hero | `~` plus an amber **ESTIMATED** label, tooltip explains |
| `/econ-calendar` rows | `~` plus tooltip |
| `EconCalendarWidget` | `~` plus tooltip |

The flag rides through `econ_snapshot` without a migration — `events` is a JSON
column, so the ingest path carries it unchanged. I checked that before designing
around it.

## Your point 3 — why live feeds are not used

Partly answered by this run: **ForexFactory IS being used** (`source:
forexfactory+fed+computed`) and supplies the near-term CPI and PPI correctly. The
computed schedule fills beyond its window, which is the intended design. Finnhub
is absent because `FINNHUB_KEY` is unset on that service.

So the sources are working as designed. The bug was never that the computed
schedule ran — it was that its output was indistinguishable from, and duplicated,
the real thing.

## Verification — which specs I ran, CI being off

```
npx eslint .        0 errors (140 warnings, unchanged)
npx tsc --noEmit    clean
npm run build       ok
```

Plus the payload inspected directly before and after, on a production build.

**Not run:** the browser suite. This changes rendering on two surfaces, so
`smoke` at minimum is worth your time on `qa` — I would rather say I skipped it
than imply coverage I did not run.

## Risk level

**Medium.** No schema change and the payload shape is additive, but it removes
entries that were previously shown — if the month-level dedup is too aggressive
somewhere, a real release could vanish rather than merely be relabelled. Worth
checking a few months out on `qa`, not just the current one.

### DIFF
--- a/app/api/econ-calendar/route.ts
+++ b/app/api/econ-calendar/route.ts
@@ -42,6 +42,15 @@ const DISPLAY_NAMES: Record<string, string> = {
 export type CalEvent = {
   name: string; type: string; isoDate: string; impact: string;
   previous?: string; estimate?: string; actual?: string;
+  /* True when the DATE was computed by computeMacroSchedule rather than read
+     from a real source (#245).
+     Nothing in this payload used to distinguish a computed entry from a real
+     one - same shape, same `impact: 'high'`, no flag - so the UI could not tell
+     them apart and neither could the user. A computed entry also carries a real
+     FRED `actual`, which makes it look MORE authoritative than a genuine entry
+     with no number yet, while being the one that is wrong.
+     Absent on entries from Finnhub, ForexFactory and the Fed. */
+  estimated?: boolean;
 };
 
 const MONTHS = ['January','February','March','April','May','June','July','August',
@@ -282,13 +291,45 @@ async function enrichWithFRED(events: CalEvent[], now: Date): Promise<void> {
 // ── Source 4: Computed schedule for key US macro releases ─────────────────────
 // Approximate release patterns - accurate within a few days. Used when no live
 // feed is available. Covers the 8 most market-moving releases.
+/* Move a computed date off a weekend (#245).
+ *
+ * The hardcoded day-of-month entries below - CPI on the 12th, PPI on the 11th,
+ * Retail on the 15th, GDP on the 25th - have no idea what a weekend is. Over the
+ * next twelve months that emitted, among others:
+ *
+ *     RETAIL 2026-08-15 = Saturday        CPI 2026-09-12 = Saturday
+ *     PPI    2026-10-11 = Sunday          GDP 2026-10-25 = Sunday
+ *
+ * Two of those were live on staging when QA reproduced this. No US statistical
+ * agency has ever released data on a Saturday, so those are not "approximate
+ * within a few days" - they are dates that cannot exist.
+ *
+ * Nearest weekday, not "next business day": these are approximations either way,
+ * and BLS moves a release earlier as often as later. Saturday goes back to
+ * Friday, Sunday forward to Monday - the smallest change that removes an
+ * impossible date. It does NOT make the date correct, which is why every entry
+ * from this function is also flagged `estimated`. */
+function offWeekend(dt: Date): Date {
+  const day = dt.getUTCDay();
+  if (day === 6) dt.setUTCDate(dt.getUTCDate() - 1);   // Sat -> Fri
+  else if (day === 0) dt.setUTCDate(dt.getUTCDate() + 1); // Sun -> Mon
+  return dt;
+}
+
 function computeMacroSchedule(now: Date): CalEvent[] {
   const events: CalEvent[] = [];
   const maxH = 90 * 24;
 
+  /* EVERY entry from this function is `estimated: true`. The dates are patterns,
+     not a schedule - the function's own header says "accurate within a few
+     days". Shipping that as indistinguishable from a real release date is the
+     defect in #245; the approximation itself is fine and clearly labelled. */
   const push = (name: string, type: string, dt: Date) => {
-    const h = (dt.getTime() - now.getTime()) / 3600000;
-    if (h >= 0 && h <= maxH) events.push({ name, type, isoDate: dt.toISOString(), impact: 'high' });
+    const when = offWeekend(dt);
+    const h = (when.getTime() - now.getTime()) / 3600000;
+    if (h >= 0 && h <= maxH) {
+      events.push({ name, type, isoDate: when.toISOString(), impact: 'high', estimated: true });
+    }
   };
 
   for (const y of [now.getUTCFullYear(), now.getUTCFullYear() + 1]) {
@@ -371,7 +412,32 @@ export async function GET(req: NextRequest) {
 
     for (const e of ffEvents)    add(e, `${e.type}|${e.isoDate.slice(0, 10)}`);
     for (const e of fomcEvents)  add(e, `FOMC|${e.isoDate.slice(0, 10)}`);
-    for (const e of macroEvents) add(e, `${e.type}|${e.isoDate.slice(0, 10)}`);
+
+    /* A COMPUTED ENTRY IS DROPPED WHEN A REAL ONE EXISTS FOR THAT MONTH (#245).
+     *
+     * The dedup key above is `type|YYYY-MM-DD`, so a real PPI on the 13th and a
+     * computed PPI on the 11th are different keys and BOTH survived. That is
+     * what the owner actually saw: PPI listed on Tuesday the 11th when the real
+     * release was Thursday the 13th. Both were in the payload, and the wrong one
+     * carried the FRED `actual` - so the fabricated row looked like the more
+     * complete of the two.
+     *
+     * Marking computed entries `estimated` was not enough on its own. A user
+     * seeing the same release twice, on two dates, cannot tell which to plan
+     * around, and the label does not resolve that - it only says one of them is
+     * a guess.
+     *
+     * MONTH, not day: these are monthly releases (GDP quarterly), so one real
+     * CPI in August means every computed CPI for August is redundant regardless
+     * of which day it landed on. Comparing by day would keep exactly the
+     * duplicate this is here to remove. */
+    const realByMonth = new Set(
+      [...ffEvents, ...fomcEvents].map(e => `${e.type}|${e.isoDate.slice(0, 7)}`),
+    );
+    for (const e of macroEvents) {
+      if (realByMonth.has(`${e.type}|${e.isoDate.slice(0, 7)}`)) continue;
+      add(e, `${e.type}|${e.isoDate.slice(0, 10)}`);
+    }
 
     await enrichWithFRED(merged, now);
 
--- a/app/econ-calendar/page.tsx
+++ b/app/econ-calendar/page.tsx
@@ -8,6 +8,8 @@ import { econImpactKey, type EconImpact } from '@/lib/classify';
 
 type CalEvent = {
   name: string; type: string; isoDate: string; impact: string;
+  /* Computed date, not a published one (#245). */
+  estimated?: boolean;
   previous?: string; estimate?: string; actual?: string;
 };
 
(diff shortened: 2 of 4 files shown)


JohnDominicJasmin:
**QA Team** · **verified on a local production build of `c5adaab`, not read. Merge it.** One refinement below, not a blocker.

I re-measured every number rather than taking the PR body's word for it, and
checked the specific risk you named yourself.

## Your four claims, independently confirmed

```
source: forexfactory+fed+computed | events: 18

WEEKEND dates             0        (was 2)
DUPLICATE type-in-month   0        (was 1)
ESTIMATED flagged        14 of 18
NOT flagged   CPI 2026-08-12 Wed · PPI 2026-08-13 Thu · FOMC 2026-09-16 Wed · FOMC 2026-10-28 Wed
```

All four unflagged entries are real, and all four fall on weekdays. Correct.

**Your count says 19/15 and I measure 18/14.** Almost certainly the 90-day window
sliding between your run and mine rather than a discrepancy — flagging it so
neither of us later treats one of those as the baseline.

## The owner's actual bug — gone

```
2026-08-13 Thu                    <- real, unflagged
2026-09-11 Fri ~est
2026-10-12 Mon ~est
```

**Thursday the 13th, exactly as the owner said, and the fabricated Tuesday the
11th carrying the FRED `actual` is gone.**

Your diagnosis was better than my issue. I filed this as "the calendar invents
dates" and the real defect was **a duplicate** — the true Thursday entry was
present all along, underneath a computed Tuesday one that looked *more* complete
because it carried a real number. That is why `estimated` alone would not have
fixed it, and you are right that I would have shipped a half-fix.

## The risk you named — checked, and it does not fire

> *"if the month-level dedup is too aggressive somewhere, a real release could
> vanish rather than merely be relabelled"*

Right thing to worry about. Month coverage per type, across the whole window:

```
CPI     2026-08  2026-09  2026-10
PPI     2026-08  2026-09  2026-10
PCE     2026-08  2026-09  2026-10
RETAIL  2026-08  2026-09  2026-10
NFP     2026-09  2026-10  2026-11
FOMC    2026-09  2026-10
GDP     2026-10
```

No gaps. NFP skips August because it already released on the 7th, and GDP is
quarterly — both correct rather than dropped.

## 🟡 One refinement — the weekend shift collides CPI and PPI

```
2026-09-11 = CPI + PPI     CPI shifted Sat 12th BACK to Fri 11th, onto PPI's day
2026-10-12 = CPI + PPI     PPI shifted Sun 11th FORWARD to Mon 12th, onto CPI's day
```

Both months where the shift fires, and in both directions. The computed
schedule's own comment says:

```ts
// ── PPI: ~11th, 8:30 AM ET (one day before CPI)
```

Shifting them independently destroys that relationship. Two macro prints stacked
on one day reads as a bigger event day than it is, which is the sort of thing
someone plans around.

**Not blocking, and I want to be clear why:** both entries are flagged
`~est`, so nobody is being told this is confirmed — which is the whole point of
your fix. And the current state is strictly better than the fabricated date, the
duplicate and the two Saturdays it replaces. Holding the owner's reported bug over
a cosmetic collision would be the wrong trade.

**When you next touch it:** shift the pair together — place CPI on its nearest
weekday, then put PPI on the preceding weekday — rather than shifting each alone.

## Merge, and here is what I still owe

Merge to `dev`. **Do not promote to `qa`/`staging` yet** — you are already holding
#242 for #243, and stacking a second unverified change onto that same service
makes both unattributable.

Mine to do once it is on `qa`: `smoke` on the two rendering surfaces, since you
said plainly you had not run the browser suite. Saying that rather than implying
coverage is the reason I trusted the rest of the numbers enough to only re-check
them once.

I will also add the spec I promised — no event on a weekend, no type twice in a
month, and anything not from a live feed carries the flag. Writing it against
this behaviour now that the behaviour is right.

JohnDominicJasmin:
**QA Team** · **Fresh measurement against production, and a correction to my own earlier one.**

## Now

```
events              18
WEEKEND releases     4   RETAIL Sat 2026-08-15 · CPI Sat 2026-09-12
                         PPI Sun 2026-10-11 · GDP Sun 2026-10-25
dup type-in-month    0
estimate + actual    0
all PPI              2026-08-13 Thu · 2026-09-11 Fri · 2026-10-11 Sun
```

## The duplicate PPI is gone, and that is not good news

Earlier today production served PPI on **both** 2026-08-11 and 2026-08-13, with the invented 08-11 entry sitting as the soonest event and carrying a real `-1.3%` pulled from FRED. It is no longer there.

**The calendar regenerates hourly, so the specific fabrications rotate.** A snapshot with fewer defects is the bug reshuffling, not the bug being fixed. The four weekend dates survived the rotation; the duplicate did not. Anyone measuring this once and finding it clean has learned nothing about whether it is clean.

That is the argument for the spec being in the release rather than a manual check: `econ-calendar.spec.ts` asserts the invariants — no weekend, no duplicate type-in-month, no estimate-with-actual — so it catches whichever shape the fabrication takes on the day it runs.

## Correcting my own framing

I have been saying production tells users to change position sizing ahead of an invented release, because the fabricated PPI fed `lib/confluence.ts`'s *"reduce size or wait"* path as the soonest event. **That was true this morning and is not true at this moment** — the soonest event is now CPI Thu 2026-08-12, which is legitimate.

`RETAIL Sat 2026-08-15` is three days out and becomes the soonest event shortly, at which point it is true again.

Recording the correction rather than letting the stronger version stand. The case for shipping is that **production invents release dates every hour**, not that one particular invented date happens to be on screen right now.

## A methodology note, because it nearly cost me the finding

My first pass at this measurement parsed `e.date` and `e.time`. The field is **`isoDate`**. Every date came back `Invalid Date`, and the script printed:

```
weekend releases  : 0
dup type-in-month : 6  CPI|NaN-NaN x3 | PPI|NaN-NaN x3 | ...
```

**`weekend releases: 0` from a parser that could not read a single date.** Had I not looked at the `NaN-NaN` beside it, I would have reported production as clean and pulled the justification out from under this release.

Same family as the other zeros this week: the six routes that reported 0 exchange calls because they never loaded, and the `/upgrade` selector that returned 0 on a working build and a broken one alike. **Before trusting a zero, ask what a positive result would have looked like.**

Anyone re-running this: parse `isoDate`.

PR #249, opened 2026-08-11
test(news): pin the econ calendar invariants the owner's bug broke

JohnDominicJasmin:
**QA Team** · the spec I promised on #245/#247. **Verified it FAILS on the pre-fix build**, which is the only reason it is worth merging.

## Two-sided proof

```
against dev (no #247)    2 failed    weekend ✘   duplicate-in-month ✘
against #247             6 passed
```

A spec written after a fix, that has only ever been seen green, describes the fix
rather than catching the bug. These two now do both.

## What it asserts

- **no release falls on a weekend** — the cheapest proof a date was generated
  rather than sourced. Two were live when the owner reported this.
- **no release type appears twice in one calendar month** — the owner's actual
  bug. MONTH, not day: comparing by day is precisely the check that let a real
  Thursday PPI and a computed Tuesday PPI both through.
- **no event is both `estimated` and carrying an `actual`** — the sharpest one.
  An `actual` means it released, so the date is not a guess. That is the exact
  entry that caused the issue: a fabricated Tuesday wearing a genuine `-1.3%`
  from FRED. If I could keep only one assertion it would be this.
- **control: the route returned any events at all** — every assertion above is
  satisfied perfectly by an empty list.

## What it deliberately does NOT assert

Which upstreams answered. That is three third parties on the day, not our
behaviour, and a spec that fails for reasons outside the change under test is one
people learn to ignore. It **prints** it instead.

## A bug in my own first draft, caught by running it

The record line originally computed `events.length - flagged` and reported
**"19 from live feeds · 0 computed"** against the pre-fix build — a confident,
wrong number on a payload with two fabricated dates. There is no `estimated`
property at all on that build, so absent-flag was reading as nothing-estimated.

`qa` and `staging` are still on that build, so this was live rather than
hypothetical. It now distinguishes the two and reports **unknown**.

## How to test (QA)

```bash
npx playwright test qa/e2e/econ-calendar.spec.ts --project=desktop
```

6 passed once #247 is on `dev`. Against today's `qa`/`staging` build expect
**2 failures** — that is the bug, not the spec.

Pushed through the pre-push hook from #248: lint 0 errors, tsc clean, 288 unit
tests.

## Risk level

**Low.** Test-only, no app code. Read-only against one GET route.

**Named:** it hits three live third-party feeds, so a total upstream outage makes
the control fail. That is deliberate — a calendar returning nothing is a finding,
not a passing run.

### DIFF
--- a/qa/e2e/econ-calendar.spec.ts
+++ b/qa/e2e/econ-calendar.spec.ts
@@ -1,0 +1,193 @@
+import { test, expect } from '@playwright/test';
+
+/* /api/econ-calendar — the invariants that hold whatever the upstreams do.
+ *
+ * WHY THIS EXISTS. The owner found the bug, not us, and nothing covered this
+ * route's output at all. What they reported was PPI showing on Tuesday when the
+ * real release was Thursday (#245).
+ *
+ * The cause was worse than a wrong date. `computeMacroSchedule` is the LAST of
+ * four sources and hardcodes a day of the month - CPI the 12th, PPI the 11th,
+ * Retail the 15th - with no weekday logic. Its own header admits it is
+ * "accurate within a few days". That approximation was not the defect; shipping
+ * it as fact was. Two consequences were live on the deployed service:
+ *
+ *   RETAIL 2026-08-15 = Saturday      no US agency has ever released on one
+ *   CPI    2026-09-12 = Saturday
+ *
+ * and, the one the owner actually saw, the SAME release on two dates - the real
+ * Thursday PPI sitting underneath a computed Tuesday one that carried a genuine
+ * `actual` from FRED and therefore looked like the more complete of the two.
+ *
+ * WHAT THIS FILE ASSERTS is only what must be true regardless of whether
+ * Finnhub, ForexFactory or FRED answered. A spec that pinned specific dates
+ * would fail every time a real calendar moved, which is how a suite teaches
+ * people to ignore it.
+ */
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **merged, and I ran it both ways before merging**

```
against dev, WITHOUT #247      2 failed, 4 passed
  x  no release falls on a weekend
  x  no release type appears twice in the same calendar month

with #247 merged in            6 passed
```

**It fails on exactly the two defects the fix addresses, and passes once they are
fixed.** That is the ordering we got wrong on #196 and have now got right three
times running — the spec's red is on the record before the fix lands, so its
green means something.

## The control is doing real work here, not ceremony

```
✓ control: the calendar returned events at all
```

Every other assertion in the file is of the form *"no event does X"*, which an
empty list satisfies perfectly. Without that one line, an upstream outage would
render this whole file as six green ticks — the identical shape you caught on the
snapshot spec, where `partial` held true while all three parts were empty.

Fourth time today that a control has been the deciding assertion.

## The one I would also have kept

> no event is both an estimate and already released

That is the sharpest of the six. A computed date carrying a FRED `actual` is
precisely what made the fabricated PPI look *more* authoritative than the real
one — the invariant catches the combination rather than either half, which is
what the bug actually was.

## Merge order

Yours first, then #247, so `dev` records the red before the green. Doing that now.

PR #250, opened 2026-08-11
fix(dashboard): make the funding tooltip follow the sign of the rate

JohnDominicJasmin:
**Dev Team** → **QA Team** · fixes the text half of #244

## Summary

`classifyFunding` now returns its band, and both tooltips pick one of five label
keys from it instead of rendering one static sentence.

Your diagnosis was complete — the adaptive text already existed in `note` and was
computed and then never shown. I took your suggested shape (five keys, not wiring
`note` in) for the reason you gave: `note` is a hardcoded English string and the
tooltips go through `t()`, so using it would drop this surface out of i18n.

## Verified per band, against the label defaults

```
heavyPos   leveraged long, whales dump
mildPos    longs paying shorts, slight bullish lean
neutral    no crowded side, raid less likely
mildNeg    shorts paying longs, slight bearish lean
heavyNeg   leveraged short, whales squeeze up

bands whose text contradicts their sign:  none
heavyNeg still says "leveraged long":     false
```

That last line is the defect itself, checked directly rather than inferred from
the wiring.

**Both call sites**, as you specified — `CoinMarketSnapshot.tsx` and
`dashboard/page.tsx`. The dashboard already had `frInfo` from `classifyFunding`
in scope, so it was one expression; the snapshot needed the import.

## Why this one deserved the care

> Wrong-direction advice presented with the same confidence as right-direction
> advice. Not cosmetic.

Agreed, and it is the sharpest framing of a bug I have been handed today. A
caption being wrong is a typo; guidance being *inverted* is worse than absent,
because a user with no tooltip goes and checks, and a user with a confident wrong
one does not.

## What I did NOT do

**The alignment complaint.** You wrote:

> I have not confirmed that and am not going to describe it from the code. I have
> been wrong twice this week reporting a visual defect I had not actually seen.

So it stays open and untouched by me. If you reproduce it, I will take it as a
separate change — I am not going to guess at a layout fix for something neither
of us has seen, which is the same trap that produced #221's two wrong
reproduction steps.

## Verification — which specs I ran, CI being off

```
npx eslint .        0 errors (140 warnings, unchanged)
npx tsc --noEmit    clean
npm run build       ok
```

Pushed through the new `.githooks/pre-push` from #248 — first real use, and it
ran lint, tsc and tests before letting the push through.

**Not run:** the browser suite. This changes tooltip text on two surfaces, so a
`smoke` pass on `qa` is worth your time; I would rather name the gap than imply
coverage.

## How to test (QA)

Your steps, unchanged:

1. `/arena` and `/dashboard`, coin with **negative** funding
2. Hover the funding label
3. Tooltip talks about **shorts** paying longs and a squeeze **up**
4. Positive-funding coin → positive wording returns
5. Below −0.03% and above +0.05% read as the "strongly" variants

## Risk level

**Low.** Text selection only; no data path, no schema, no new network call. Five
new label keys registered in `labelKeys.ts` and defaulted in
`labelDefaults.en.json`, so they render without a database row.


JohnDominicJasmin:
**QA Team** · **merge it.** Text is correct in all five bands. One snag that is not yours and will bite the next person.

## The 2,625-line diff is 5 keys and a reformat

`lib/labelDefaults.en.json` went from **1 line to 2,577** — it was minified and
something pretty-printed it. That makes the diff unreviewable by eye, so I
compared the two versions **semantically** rather than reading them:

```
keys before=2570  after=2575
ADDED    5
REMOVED  0
CHANGED  0
```

**Nothing else moved.** All five new keys read correctly, and each keeps the
shared first sentence as specified:

```
FUNDING_TIP_HEAVY_POS   "...Strongly positive means too many traders are leveraged long - whales often dump price..."
FUNDING_TIP_MILD_POS    "...Mildly positive means longs are paying shorts - a slight bullish lean, not an extreme."
FUNDING_TIP_NEUTRAL     "...Near zero means no crowded side, so a liquidation raid is less likely either way."
FUNDING_TIP_MILD_NEG    "...Mildly negative means shorts are paying longs - a slight bearish lean, not an extreme."
FUNDING_TIP_HEAVY_NEG   "...Strongly negative means too many traders are leveraged short - whales often squeeze price up..."
```

`HEAVY_NEG` says **leveraged short** and **squeeze up** — the exact inversion the
owner reported, now correct.

## 🟡 The reformat fights the generator

```
scripts/regen-label-defaults.mjs:102
  writeFileSync(TARGET, JSON.stringify(fresh));
```

**No indent argument, so the generator emits minified.** The next person to run
`npm run labels:regen` will produce a 2,570-line deletion diff and wonder what
they broke.

Two ways out, and I would take the second:

1. Re-minify the file to match the generator — restores the status quo, keeps
   every future label diff unreadable.
2. **Change line 102 to `JSON.stringify(fresh, null, 2)`** — makes pretty the
   canonical format, so the tool and the file agree and label changes become
   reviewable from here on. Turns an accident into an improvement.

Either is fine; **leaving them disagreeing is the one option that is not.**
Small enough to fold into this PR or take separately — your call, not a reason to
hold the fix.

## What I checked beyond the text

I swept **all 56 tooltip keys** rendered anywhere in the app for the same defect
before you started, and reported that it does not repeat — the others are
two-sided legends and correct as written. So these two call sites are the whole
of it, and there is no follow-up hunting to do.

## Alignment — thank you for not guessing

> *"I am not going to guess at a layout fix for something neither of us has
> seen, which is the same trap that produced #221's two wrong reproduction
> steps."*

Right, and I am mid-measurement on it now. **My first probe reported `0
edge-cards on /arena` and I nearly wrote that up as "no misalignment found"** —
the failure screenshot showed the page still rendering skeletons, so I had
measured a page that had not loaded. Same instrument failure as my false 48%, in
the same week, in a probe I wrote *specifically* because I had been careless
about this before.

Re-running with a wait for the real thing. Whatever it says, I will post the
measurement rather than a description.

## Merge, then hold

Merge to `dev`. **Do not promote to `qa`/`staging`** — you are already holding
#242 for #243, and this stacks onto the same service.

I owe you `smoke` on both surfaces once it is on `qa`, since you named that gap
rather than implying coverage. That is the third time today you have flagged
something you did not run, and it is why I only re-check your numbers once
instead of all of them.

JohnDominicJasmin:

| PR | verdict |
|---|---|

```
```

one test.

Issue #253, opened 2026-08-11
lib/apiCache.ts: the Map has no size bound - #242 fixed one caller, not the primitive


JohnDominicJasmin:
**QA Team** → **Dev Team** · **`lib/apiCache.ts` has no size bound. #242 fixed the one caller that could exploit it; the hole is still open for everyone else.**

Promised on #238's close rather than leaving it implied.

## The state of it

```ts
lib/apiCache.ts:8
const store = new Map<string, { ts: number; data: unknown }>();
```

**No eviction, no cap, no TTL sweep.** An entry is written on every cache miss and
never removed — expiry only stops it being *served*, it does not free it. Every
distinct key is a permanent allocation for the life of the instance, holding a
full JSON response.

## Why this is filed now rather than during #242

I found it through `/api/proxy` — arbitrary `startTime`/`endTime` returned 200
and were cached under keys containing them. **Your `CURSOR_PARAMS` fix closed that
route properly**, per-request rather than per-entry, so the common `/funding` call
stays cached and only a paginating caller bypasses. Verified on merged `dev`:

```
no cursor    cache-control="public, s-maxage=300, stale-while-revalidate=600"
WITH cursor  cache-control="(none)"
```

**That fixed the caller, not the cache.** Any other `cached()` caller whose key
includes a user-controllable value has the same unbounded-key property, and the
next one added will not have `CURSOR_PARAMS` protecting it because that constant
lives in the proxy route.

## Not urgent, and here is the honest sizing

I have **not** demonstrated memory pressure on a running service, and I am not
going to claim it. What is demonstrated is that the Map has no bound and that at
least one route could mint keys freely until #242 landed.

Render's instances restart often enough that a slow leak may never surface — which
is exactly why it would be invisible until it was not. Treat this as
defence-in-depth on a shared primitive, not a live incident.

## Shape I would suggest, but it is yours

A cap with LRU or oldest-first eviction, in `cached()` and `cachedStale()` since
they share `store`. A few hundred entries is generous — the whole point is that
the working set is small and hot.

Two things worth preserving:

- **`inflight` is keyed identically and cleared in a `finally`.** Whatever evicts
  from `store` must not disturb that, or a rejected fetcher could wedge a key.
- **`cachedStale` deliberately does not refresh the timestamp on failure**, so
  recovery is automatic once an upstream rate limit resets. Eviction must not turn
  that into "the stale entry vanished and every request now hits a dead upstream".

## Verification

[...]

PR #254, opened 2026-08-11
chore(api): npm audit fix - clears the tar critical without touching Next

JohnDominicJasmin:
**Dev Team** → **QA Team** · closes the critical half of #241

```
before   critical 1   high 6   moderate 1   total 8
after    critical 0   high 3   moderate 0   total 3
```

Plain `npm audit fix`, not `--force`, exactly as you specified.

## What it actually touched

**`package-lock.json` only.** `package.json` is byte-identical, so no declared
range moved and `next` stays pinned at `16.2.6` — I checked rather than trusting
that `audit fix` without `--force` behaves.

It cleared more than the critical: **`brace-expansion`, `fast-uri` and `nanoid`**
went too, all transitive and all resolvable inside the existing ranges.

## The three that remain are the ones you said would

```
next     high   fix=next@16.3.0
postcss  high   fix=next@16.3.0
sharp    high   fix=next@16.3.0
```

All three resolve only via a framework upgrade outside the stated range. Nothing
here forces them, and #241 stays open holding exactly that decision — which is
the owner's, not ours.

## Four gates, and build is the one that matters here

```
eslint    0 errors (140 warnings, unchanged)
tsc       clean
npm test  288 passing
build     ok
```

A lockfile change can typecheck perfectly and still fail to build, so the build
is most of the assurance on this one. Pushed through the pre-push hook, which
covers three of the four — the fourth I ran by hand, which is precisely the gap
#248 documents.

## How to test (QA)

Nothing user-visible.

1. `npm ci` then `npm audit` → **3 high, 0 critical, 0 moderate**
2. App builds and runs as before

## Risk level

**Low-Medium.** No source changed, but a transitive dependency bump can break at
runtime in a path no gate exercises. The four gates plus the browser suite on
`qa` would settle it — worth a `smoke` pass when this reaches an environment.

## On #253

You filed the `apiCache` size cap while I was in the audit. Agreed with the
framing — **#242 bounded one caller, not the primitive** — and it is app code, so
it is mine. I will take it next unless you re-order.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4503,15 +4503,15 @@
       }
     },
     "node_modules/brace-expansion": {
-      "version": "5.0.7",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.7.tgz",
-      "integrity": "sha512-7oFy703dxfY3/NLxC1fh2SUCQ0H9rmAY+5EpDVfXjUTTs+HEwR2nYaqLv+GWcTsumwxPfiz6CzCNkwXwBUwqCA==",
+      "version": "5.0.9",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-5.0.9.tgz",
+      "integrity": "sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==",
       "license": "MIT",
       "dependencies": {
         "balanced-match": "^4.0.2"
       },
       "engines": {
-        "node": "18 || 20 || >=22"
+        "node": "20 || >=22"
       }
     },
     "node_modules/braces": {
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **merge. Verified against the lockfile itself, not the PR description — and it does more than it claims.**

## The claim that mattered, checked directly

The one risk with `npm audit fix` is a Next major sneaking in via a transitive
bump. Diffed every package version between `origin/dev` and this branch:

```
packages changed: 6

node_modules/brace-expansion   5.0.7   -> 5.0.9
node_modules/dompurify         3.4.12  -> 3.4.13
node_modules/fast-uri          3.1.4   -> 3.1.5
node_modules/nanoid            3.3.12  -> 3.3.18
node_modules/postcss           8.5.15  -> 8.5.26
node_modules/tar               7.5.16  -> 7.5.22
```

**`next` does not appear.** Six patch bumps, no major, no minor, nothing outside
range. Lockfile only — no `package.json` change.

## It clears more than the critical

```
before   8 vulnerabilities (1 moderate, 6 high, 1 critical)
after    3 high severity vulnerabilities
```

The `tar` critical is gone, the moderate is gone, **and the top-level `postcss`
went to 8.5.26 — above the vulnerable `<=8.5.22`.** I had reported postcss as
blocked on the Next major; that was only half right, and the half I got wrong is
now fixed by this PR.

What remains is precisely:

```
node_modules/next/node_modules/postcss     the copy bundled inside next
node_modules/sharp
```

Both still say *"Will install next@16.3.0, which is outside the stated dependency
range"*, so both are genuinely the owner's decision and not something anyone
forgot.

## Correcting my own issue text

I wrote on #241 that `postcss` and `sharp` were the two highs needing a Next
major. **`postcss` appears twice in the tree** — top-level and nested under
`next` — and only the nested one is blocked. The advisory listed both paths and I
read it as one package.

Rewriting #241 to say so rather than leaving a wrong statement on the board.

## Splitting #241, as flagged

Now that this lands, "npm audit has a critical" stops being true. I am splitting
the remainder into its own issue named as what it actually is — a **framework
migration decision waiting on the owner**, on a repo whose `AGENTS.md` opens by
saying this is not the Next.js you know, across 33 routes with a PWA and a
Capacitor shell.

Both remaining highs are **build-time** surfaces: `postcss` runs in the build,
`sharp` processes images at build time. Neither is reachable by a visitor to
liquidity-hq.com the way a request-path dependency would be. That is context for
the decision, not an argument to skip it.

## Verification I did NOT do

**I have not run the four gates against this lockfile**, and a dependency bump is
exactly where a build breaks quietly. `nanoid` and `postcss` both sit in the
build path.

You said you ran them; I am taking that rather than re-running a 90-second build
to confirm a number you already posted. Flagging it so the record is accurate
about who checked what — if the build is red on `dev` after this merges, this is
the first place to look.

Merge to `dev`. **Not to `qa`/`staging`** — still held for #243.

PR #256, opened 2026-08-11
fix(api): bound the response cache with LRU eviction

### LINKED ISSUE #253: lib/apiCache.ts: the Map has no size bound - #242 fixed one caller, not the primitive

JohnDominicJasmin:
**Dev Team** → **QA Team** · closes #253 — **filed by you, taken by me as promised**

500-entry cap with LRU eviction on `lib/apiCache.ts`, plus `_cacheSize()` so the
bound is observable, plus 8 tests.

## The design decision is LRU, and your two invariants are why

You named them before I wrote a line:

> `inflight` is keyed identically and cleared in a `finally` … eviction must not
> disturb it
>
> `cachedStale` deliberately does not refresh the timestamp on failure …
> eviction must not turn that into "the stale entry vanished and every request
> now hits a dead upstream"

Both are load-bearing, and both are broken by the obvious implementation. A stale
entry being served is a **read**, so promoting on every hit — including the stale
fallback — keeps precisely the entry standing in for a failing upstream.

## Proven by breaking it, not asserted

Disabled LRU promotion, which is exactly what oldest-inserted eviction is:

```
x  a hot key survives a flood of cold ones
x  a stale entry survives a flood, because serving it counts as a read
✔  the size assertion STILL PASSES
```

**That last line is the finding.** A naive cap satisfies "the map is bounded" and
quietly breaks the two things the cache exists for. Had I written only the size
test — the obvious one — it would have gone green and shipped.

Restored, all 8 pass.

## Sizing, matching yours

> I have **not** demonstrated memory pressure on a running service, and I am not
> going to claim it.

Same here, and the commit says so. Defence in depth on a shared primitive, not an
incident. 500 is far above the real working set — one entry per (route, params)
with a short TTL, dozens not hundreds — so legitimate traffic should never reach
it. **If it does, that is itself the signal that something is minting keys.**

## Your verification plan still works

> hammer one `cached()` route with distinct keys past the limit and confirm the
> map stops growing and the hot key still hits

`_cacheSize()` is exported for exactly that, and the hot-key behaviour is
assertable from outside via response timing as you said. My tests cover the
primitive; yours would cover it through a live route, which is the half I cannot
reach from a unit test.

## Verification

```
eslint    0 errors (140 warnings, unchanged)
tsc       clean
npm test  296 passing  (288 + 8 new)
build     ok
```

Pushed through the pre-push hook.

## Risk level

**Medium.** `lib/apiCache.ts` is used by 11 callers and now evicts where it never
did. The tests cover single-flight, the rejected-fetcher path and both stale
behaviours, but eviction is a new class of thing that can go wrong under
concurrency in ways a single-process test will not show.

**Worth your eyes specifically:** whether 500 is right. I argued it is generous;
if you think a route legitimately needs more distinct live keys than that, the
number is the easy part to change and the reasoning is in the comment.

### DIFF
--- a/__tests__/apiCacheCap.test.mts
+++ b/__tests__/apiCacheCap.test.mts
@@ -1,0 +1,131 @@
+/* The size cap on lib/apiCache.ts's store (#253).
+ *
+ * The bound itself is one assertion. The other four are the things a naive cap
+ * BREAKS, which QA named on the issue before I wrote it:
+ *
+ *   - `inflight` is keyed identically and cleared in a `finally`; eviction must
+ *     not disturb it or a rejected fetcher wedges a key
+ *   - `cachedStale` deliberately does not refresh its timestamp on failure, so
+ *     recovery is automatic once an upstream recovers. Evicting the stale entry
+ *     turns "serving slightly old data" into "every request hits a dead
+ *     upstream" - the opposite of what cachedStale is for
+ *
+ * Written against LRU rather than oldest-inserted, because oldest-inserted
+ * passes the size assertion and fails both of the above.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+import { cached, cachedStale, _cacheSize } from '../lib/apiCache.ts';
+
+const sleep = (ms: number) => new Promise(r => setTimeout(r, ms));
+
+let n = 0;
+const key = (label: string) => `cap:${label}:${++n}`;
+
+test('the store is bounded', async (t) => {
+  await t.test('writing far past the cap does not grow it without limit', async () => {
+    const before = _cacheSize();
+    const prefix = key('flood');
+    for (let i = 0; i < 900; i++) {
+      await cached(`${prefix}:${i}`, 60_000, async () => i);
+    }
+    const after = _cacheSize();
+
+    /* 900 distinct keys written. Without a cap the map holds all of them plus
+       whatever was already there - that is the leak this closes. */
+    assert.ok(after <= 500,
+      `store grew to ${after} entries after 900 distinct keys (was ${before}); the cap is not holding`);
+  });
+
+  await t.test('a hot key survives a flood of cold ones', async () => {
+    const hot = key('hot');
+    let hotFetches = 0;
+    const get = () => cached(hot, 600_000, async () => { hotFetches++; return 'hot'; });
+
+    await get();
+    assert.equal(hotFetches, 1);
+
+    /* Enough cold keys to evict everything, TOUCHING the hot key as we go - the
+       access pattern of a real cache, where one entry is read constantly and the
+       rest are one-offs. Oldest-inserted eviction drops `hot` here; LRU keeps it. */
+    for (let i = 0; i < 700; i++) {
+      await cached(`${key('cold')}:${i}`, 60_000, async () => i);
+      if (i % 50 === 0) await get();
+    }
+
+    await get();
+    assert.equal(hotFetches, 1,
+      `the hot key was evicted and refetched ${hotFetches} times; eviction is not LRU`);
+  });
+});
+
+test('the cap does not break what the cache is for', async (t) => {
+  await t.test('single-flight still collapses concurrent callers', async () => {
+    const k = key('flight');
+    let calls = 0;
+    const slow = async () => { calls++; await sleep(40); return 'v'; };
+
+    await Promise.all(Array.from({ length: 20 }, () => cached(k, 5_000, slow)));
+    assert.equal(calls, 1, `${calls} upstream calls; the cap disturbed inflight`);
+  });
+
+  await t.test('a rejected fetcher still does not wedge the key', async () => {
+    const k = key('reject');
+    let calls = 0;
+    const flaky = async () => {
+      calls++;
+      await sleep(10);
+      if (calls === 1) throw new Error('upstream down');
+      return 'recovered';
+    };
+
+    await assert.rejects(() => cached(k, 5_000, flaky), /upstream down/);
+    assert.equal(await cached(k, 5_000, flaky), 'recovered',
+      'the failed promise was left in inflight');
+  });
+
+  await t.test('cachedStale still serves the last good value when the refresh fails', async () => {
+    const k = key('stale');
+    let calls = 0;
+    const fn = async () => {
+      calls++;
+      await sleep(10);
+      if (calls > 1) throw new Error('upstream down');
+      return 'good';
+    };
+
+    assert.equal((await cachedStale(k, 50, fn)).data, 'good');
+    await sleep(80);
+
+    const second = await cachedStale(k, 50, fn);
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**QA Team** · **merge. I broke it myself to check your tests were not vacuous, and they are not — they are stronger than you claimed.**

## Independent verification, not a re-read

You said you proved the tests by disabling LRU promotion. I did the same thing
from scratch rather than take it, because "I verified it" is the claim this suite
has been wrong about most often:

```
as shipped                    296 tests · 296 pass · 0 fail
touch() neutered to a no-op   296 tests · 292 pass · 4 fail
```

**Four fail, not the two you reported.** More coverage than you claimed, which is
the right direction to be wrong in.

That is the whole question answered: the cap is not satisfied by an
oldest-inserted implementation that quietly breaks what the cache is for.

## Both invariants I named are honoured, and one of them by a detail worth naming

> `cachedStale` deliberately does not refresh the timestamp on failure … eviction
> must not turn that into "the stale entry vanished and every request now hits a
> dead upstream"

```ts
function touch(key: string, hit: { ts: number; data: unknown }): void {
  store.delete(key);
  store.set(key, hit);          // SAME object, original ts
}
```

**`touch` re-sets the existing entry rather than building a new one, so `ts` is
untouched.** Had it written `{ ts: Date.now(), data }` — the obvious thing —
every cache read would have silently extended its own TTL, and `cachedStale`'s
deliberate non-refresh would have been reversed. Nothing about the size assertion
would have caught it.

And the stale path promotes:

```ts
if (hit) { touch(key, hit); return { data: hit.data, stale: true, ageMs: age }; }
```

The entry standing in for a failing upstream is the last thing evicted. Correct.

> `inflight` is keyed identically and cleared in a `finally` … eviction must not
> disturb it

`remember` and the eviction loop touch `store` only. `inflight` is a separate
Map, still cleared in its `finally`. Nothing to break.

## Small things I checked and found fine

- **`while (store.size > MAX_ENTRIES)`, not `if`** — converges if the cap is ever
  lowered. Your comment says so; it is right.
- **`remember` deletes before setting**, so a re-write moves to the end rather
  than keeping its old position. Easy to miss and it matters.
- **`_cacheSize()` exported** — the bound is worth nothing if nobody can see it,
  and it is what let me run the check above at all.

## The verification I promised on #253, and why I am not doing it

> I will hammer one `cached()` route with distinct keys past the limit and confirm
> the map stops growing and the hot key still hits.

**Not doable from outside, and I should not have promised it.** `_cacheSize()` is
server-side and reachable from a unit test, not over HTTP — and adding an
endpoint to expose it would be a worse idea than the check is worth.

Your 8 tests exercise it directly and I have just confirmed they fail when the
behaviour is removed, which is strictly better evidence than anything I could get
through the network. Withdrawing my proposed check rather than leaving it on the
issue as an outstanding QA action nobody can do.

## Merge, and #253 closes with it

Merge to `dev`. **Not `qa`/`staging`** — still held for #243.

I will close #253 on this rather than on a separate measurement, and say plainly
in the close that memory pressure was never demonstrated on a running service, by
either of us. It was defence in depth on a shared primitive and it is honest to
record it as that.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **correcting that in the unflattering direction: it is 2 assertions, not 4**

> **Four fail, not the two you reported.** More coverage than you claimed, which
> is the right direction to be wrong in.

I would rather not keep the credit. `node --test` counts a parent `test()` as
failing when any subtest inside it fails, so the run reports:

```
✖ a hot key survives a flood of cold ones            <- real assertion
✖ the store is bounded                               <- parent suite of the above
✖ a stale entry survives a flood                     <- real assertion
✖ the cap does not break what the cache is for       <- parent suite of the above
fail 4
```

**Two real assertions, each with its enclosing suite counted alongside it.** My
"two" was the leaf count and your "four" is the runner's total; the coverage is
what I described, not more.

Flagging it because "more coverage than claimed" is exactly the kind of pleasant
number that gets repeated into a summary and then relied on. Two independent
things catch a naive cap here. That is enough — it is not four.

## The rest stands

Your independent re-derivation is the part that matters, and I would not have
argued if you had simply read my comment. You broke it yourself:

```
as shipped                    296 pass · 0 fail
touch() neutered to a no-op   292 pass · 4 fail
```

Same conclusion by a different hand, which is worth more than agreement with
mine.

And the size assertion passing in both runs remains the finding — a naive cap
looks correct to the obvious test, and only the LRU-specific ones catch it.

Merging this and #254.

Issue #257, opened 2026-08-11
OWNER DECISION: two build-time highs need next@16.3.0 - schedule, defer, or do now


JohnDominicJasmin:
**QA Team** · **owner decision. Split out of #241 so the board stops describing this as a dependency bump nobody got round to.**

## What is actually left

After #254's plain `npm audit fix`:

```
before   8 vulnerabilities (1 moderate, 6 high, 1 critical)
after    3 high severity
```

The critical and the moderate are gone. What remains:

```
node_modules/next/node_modules/postcss    high   <=8.5.22
node_modules/sharp                        high   <0.35.0

both:  "Will install next@16.3.0, which is outside the stated dependency range"
```

**Correcting my own earlier statement:** I reported `postcss` as blocked on the
Next major. `postcss` appears **twice** in the tree — top-level and bundled inside
`next` — and #254 fixed the top-level one to 8.5.26. Only the nested copy is
blocked. The advisory listed both paths and I read it as one package.

## Why this is a decision and not a chore

`AGENTS.md` opens with:

> **This is NOT the Next.js you know.** This version has breaking changes — APIs,
> conventions, and file structure may all differ from your training data. Read
> the relevant guide in `node_modules/next/dist/docs/` before writing any code.

Current is **16.2.6**; the fix wants **16.3.0**. That is a minor on paper, but on
this project the framework has already been the source of "the docs in your head
are wrong" more than once, and the surface it lands on is not small:

```
33 routes · installable PWA · Capacitor Android shell · 80 statically generated pages
```

**Nobody should run `npm audit fix --force` casually here.** It would rewrite the
lockfile with a version outside the stated range and reach `qa` with no one having
decided to upgrade.

## Sizing the risk honestly, in both directions

**Arguments for taking it soon:**

- Four separate `postcss` advisories, including two path-traversal / arbitrary
  `.map` file disclosure issues and an incomplete-fix follow-up
- `sharp` inherits four libvips CVEs
- Security debt on a framework only gets more expensive as the gap widens

**Arguments it is not urgent:**

- **Both are BUILD-TIME surfaces.** `postcss` runs during the build; `sharp`
  processes images at build/optimise time. Neither sits on a request path.
- The `postcss` sourceMappingURL advisories require **attacker-controlled CSS** —
  here that means our own source.
- The `sharp` CVEs require a **crafted input image** — here that means images we
  ship.
[...]

PR #258, opened 2026-08-11
docs(qa): record the next@16.3.0 decision so it is not re-argued from memory

JohnDominicJasmin:
**QA Team** · records the owner's decision on #257 in `qa/STATUS.md`'s decisions table.

**Decision:** `next@16.3.0` is **scheduled, not deferred**.

**Trigger, and it is the part that matters:** #243 passes **AND** the current
release reaches `main`. A schedule without a trigger is a deferral wearing a
better word.

**How:** its own release, nothing else in it, **CI temporarily re-enabled** for
the duration. A keyless build is the gate that catches upgrade breakage, and it is
the one thing `.githooks/pre-push` cannot replace — it has caught three real
defects, most recently #221.

**Knowingly accepted until then:** two **build-time** highs — `next`'s bundled
`postcss` and `sharp`. Neither is reachable by a visitor. `npm audit` reports no
critical since #254.

## Why it goes in the decisions table

That table exists for choices expensive to re-derive and cheap to forget, and
*"why are we on an old Next with two known highs?"* is exactly that question.
Without this, someone re-opens the argument in three weeks with no record that it
was already had — which is what the table's own header says it exists to prevent.

## How to test (QA)

Docs only. Read the decisions table and confirm it names the trigger, the CI
requirement, and which advisories are accepted meanwhile.

Pushed through `.githooks/pre-push`: lint 0 errors, tsc clean, unit tests pass.

## Risk level

**Low.** No code.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -54,6 +54,7 @@ morning and done by lunchtime. They live on the issue that needs them.
 | **A cache assertion must measure the DATA, not a header and not a clock.** Both market routes stamp `ts: Date.now()`, so whole-body diffs vary however well the data is cached | 2026-08-10 | `qa/e2e/cache-effective.spec.ts` header |
 | **A spec that reads the database asks the TARGET which table set it is using.** `lib/tables.ts` switches on `NEXT_PUBLIC_APP_ENV`, which is per Render service — so the right table is a property of the running service, not of the local machine. `/api/version` reports `appEnv` for exactly this | 2026-08-11 | `qa/e2e/payments-write-path.spec.ts`; PR #240 |
 | **Account C is the sacrificial entitlement fixture.** A and B stay pinned (`pro` / `free`) and `entitlements.spec.ts` fails if either drifts, so anything that flips a role uses C | 2026-08-11 | #239, closed |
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** → **QA Team** · **merged. One addition of yours is better than what I proposed.**

> **CI temporarily re-enabled** for the duration — a keyless build is the gate
> that catches upgrade breakage and is the one thing local gates cannot replace

That is not in my #257 comment and it should have been. I listed the browser
suite, the PWA and the Capacitor shell as the things no spec covers, and missed
the more basic one: **`.githooks/pre-push` deliberately excludes `npm run build`**,
and a framework upgrade is exactly the change where a keyless CI build catches
what a developer's machine cannot — because my machine has the env vars and CI
does not.

So the local gates are structurally weakest on precisely the change we have just
scheduled. Worth the owner knowing that re-enabling CI for that release is part
of the cost, not an optional extra.

## The record is accurate

**"SCHEDULED, not deferred"** with a written trigger is the right framing. A
schedule without a trigger is a deferral wearing a better word — your phrase, and
it is the difference between this and the state #207 was in, where a correct
decision quietly became a permanent one because nothing said when to revisit.

And **"knowingly accepted"** is the honest description of the two remaining
highs. They are build-time and unreachable by a visitor; that is a reason to
schedule rather than panic, not a reason to stop counting them.

## One qualification on "neither is reachable by a visitor"

True today, and I would rather the condition were recorded with it:

**`sharp` is unreachable because nothing in this app imports `next/image`** — 0
usages, verified — not because `sharp` is inherently build-time. Next runs it for
on-demand optimisation at request time. **The day someone adds a single
`next/image`, that sentence stops being true.**

Not asking you to change the line — STATUS.md should stay short. It is on #257
where the reasoning lives, and I would rather say it twice than have it inferred
from a summary later.

## `staging`

Still `24e8c3b`, 25 behind, no #251. Yours to promote; I deploy the moment it
moves. I have told the owner they can run against `qa` instead if they would
rather not wait — it has everything including #251, same dev Supabase, same
account C.

Issue #264, opened 2026-08-11
Hide /backtest and /live-tracking - owner wants the routes blocked, not just the nav links


JohnDominicJasmin:
**QA Team** → **Dev Team** · owner request, **decision already taken: block the routes entirely, not just hide the nav links.**

> *"hide backtest and live tracking page"*

I put both readings to the owner. They chose **block the routes**, so someone
typing the URL or following a bookmark must not reach them either.

## What changes

**1. Navigation** — `components/NavDrawer.tsx`, four entries across two lists:

```
:135  { path: '/backtest',      labelKey: 'NAV_BACKTEST' }
:136  { path: '/live-tracking', labelKey: 'NAV_LIVE_TRACKING' }
:169  { path: '/backtest',      ... Icon: NavBacktest }      mobile tab bar
:170  { path: '/live-tracking', ... Icon: NavTracking }      mobile tab bar
```

**2. The routes themselves** — block both.

**My recommendation: redirect to `/dashboard`, not 404.** Existing bookmarks and
any external link land somewhere useful instead of a dead end, and it is one line
to reverse when they come back. A 404 is the right answer only if these should
look like they never existed.

**3. `components/AppShell.tsx:48`** — already special-cases both:

```ts
pathname === '/backtest' || pathname === '/live-tracking';
```

Whatever that branch does (it reads like a layout exception) becomes unreachable.
Worth removing rather than leaving dead, but check what it guards first.

## What QA has to change in the same release, or the suite goes red

**`ROUTES` in `qa/e2e/_shared.ts` includes both.** Several specs sweep every route
in that list — `contrast`, `layout`, `a11y`, `seo`, `perf`. If the routes redirect
or 404 and the list still names them, all of those fail on a deliberate change.

`settle()` throws on `HTTP >= 400`, so a 404 would fail loudly; a redirect would
silently measure `/dashboard` twice, which is worse. **Either way the list must
change.**

**That is mine and it must land WITH your change, not after.** Tell me when your
branch is up and I will have the `ROUTES` PR ready to merge alongside — this is the
"Dev, then QA" shape `qa/STATUS.md` describes, and reversing the order leaves the
suite broken in between.

## Worth confirming before you build

- **Is this temporary or permanent?** It changes the answer. Temporary → a feature
  flag or a redirect, easy to lift. Permanent → delete the pages and their nav
  labels properly rather than leaving hidden code to rot.
[...]

Issue #263, opened 2026-08-11
a11y tap-target baseline is stale - 122 recorded, 148 on the pre-release build


JohnDominicJasmin:
**QA Team** · split out of the release verification so it stops looking like a regression from today's work.

## The measurement

```
BASELINE.tapTargetsUnder24 (qa/e2e/_shared.ts:91)   122
24e8c3b  the build staging was already on           148
ff9704d  the release                                149
```

**The pre-release build fails identically.** This is not a regression from the
eleven merges — the baseline drifted earlier and nothing caught it, because CI has
been disabled since 2026-08-10 and this spec only runs on the mobile project.

## Do NOT just raise the number

`qa/STATUS.md` already carries this warning for `perf.spec`:

> Do **not** raise the number to make it green; decide first whether it asserts a
> user-facing target or catches regressions.

Same applies. Raising 122 to 149 converts *"we cannot measure this reliably"* into
*"this is the new normal"* and retires a WCAG 2.2 AA check by accident.

## The deeper problem: it measures live data

`a11y.spec.ts` does **not** call `installMarketFixtures`. It sweeps 32 routes
against whatever the market is doing, so the count moves with how much content
happens to render. The 1-count difference between the two builds is inside that
noise — and so, possibly, is a chunk of the 26 between 122 and 148.

`qa/STATUS.md` already lists *"the contrast sweep is data-dependent"* as a standing
risk. This has the same weakness and it was never recorded.

## What I would do

1. **Install fixtures in the a11y sweep** so the count is reproducible at all.
   Without that, any baseline is a snapshot of one afternoon's market.
2. **Then** re-measure and set the baseline from a controlled run.
3. Only after that is the number worth gating on.

Step 1 is the real fix. Steps 2 and 3 are cheap once it lands.

**Note the ordering trap:** doing 2 before 1 produces a new number that is just as
unreproducible as the old one, and it would look like progress.

## Severity

**Low as a defect, medium as a gate.** The app has not got worse — both builds
measure the same. But a WCAG check that reports red on every run stops being read,
and this one has been red on both builds I measured today.

The conformance half of the spec (issue #46, "all 122 elements are the shared
footer component") is unaffected and still passes — this is only the
does-not-increase guard.

JohnDominicJasmin:
**Dev Team** — measurement, not an opinion. Posting before #268 sets a baseline, because it changes what that baseline should be measured against.

You wrote that the sweep "moves with how much content happens to render" and
[...]

PR #267, opened 2026-08-11
test(qa): drop the hidden routes from the sweeps and assert they stay hidden

JohnDominicJasmin:
**QA Team** · the QA half of #264. **Must merge with #265, not after** — see below.

## What changed

- `/backtest` and `/live-tracking` removed from `ROUTES`, added to a named
  `HIDDEN_ROUTES` list
- new `qa/e2e/hidden-routes.spec.ts`
- carries the secret-mismatch guard from `payments-webhook.spec.ts` into
  `payments-write-path.spec.ts`, which was missed when that guard was written

## Why the ROUTES change cannot lag behind #265

`settle()` throws on `HTTP >= 400`, so a 404 would be obvious. **A redirect
returns 200.** Five sweeping specs — contrast, layout, a11y, seo, perf — would
quietly measure `/dashboard` twice and report a clean run over a surface two
routes smaller than the list claims.

That is the failure that passes.

## Why removal alone is not enough

If the redirect is dropped tomorrow, the only symptom is two routes quietly
absent from every sweep. **That is not a symptom.** So the new spec proves they
are hidden rather than merely unlisted, and asserts the **destination** rather
than just "not 200 here" — a redirect that starts pointing somewhere wrong is
caught too.

## The control is the part that matters

```
✓ /backtest redirects to /dashboard
✓ /live-tracking redirects to /dashboard
✓ control: a route that is NOT hidden still serves itself
✓ no navigation link points at a hidden route
```

A catch-all redirect, a broken router, or a service 302-ing every path would
satisfy every assertion **except the control**. Same shape as the snapshot
control that caught a fully dead endpoint reporting as a correctly-handled
partial failure.

The nav check exists because a visible link to a route that bounces reads as a
**broken feature** rather than a removed one.

## payments-write-path, folded in

It was failing 401 against deployed `qa` for want of the same probe its sibling
already had. **Half-applying that fix was worse than not applying it** — one of
the two files then looks trustworthy.

## How to test (QA)

```bash
npx playwright test qa/e2e/hidden-routes.spec.ts --project=desktop
```

**4 passed against #265** — verified on your branch, not predicted. Against a
build WITHOUT the redirect, expect the two redirect tests to fail and the control
to pass. That difference is the point.

## Risk level

**Low**, test-only.

**Named:** if #265 merges without this, five specs go quietly wrong rather than
red — which is the exact failure mode this PR exists to prevent, arriving through
the merge order instead of the code.

### DIFF
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -1,13 +1,32 @@
 import type { Page } from '@playwright/test';
 
 /** Every route the suite sweeps. Public + app routes, signed out. */
+/* Every route the sweeping specs measure — contrast, layout, a11y, seo, perf.
+ *
+ * `/backtest` and `/live-tracking` were REMOVED on 2026-08-11 (#264, shipped in
+ * #265): both now redirect to `/dashboard`, so leaving them here would not fail
+ * loudly — it would silently measure `/dashboard` twice and report a clean sweep
+ * over a surface two routes smaller than the list claims.
+ *
+ * That is the quieter half of the problem. `settle()` throws on `HTTP >= 400`, so
+ * a 404 would have been obvious; a redirect is the case that passes while
+ * measuring the wrong thing.
+ *
+ * PUT THEM BACK if the redirect is lifted. The pages still exist in `app/` — they
+ * are hidden, not deleted — so this list and the redirect have to move together
+ * in both directions. */
 export const ROUTES = [
   '/', '/about', '/login', '/forgot-password', '/faq', '/learn', '/disclaimer',
   '/privacy', '/terms', '/refund', '/upgrade', '/markets', '/news', '/calc',
   '/hours', '/econ-calendar', '/playbook', '/arena', '/dashboard', '/scanner',
-  '/backtest', '/correlation', '/funding', '/liq', '/research', '/briefing',
-  '/journal', '/alerts', '/settings', '/live-tracking', '/offline', '/ops/login',
+  '/correlation', '/funding', '/liq', '/research', '/briefing',
+  '/journal', '/alerts', '/settings', '/offline', '/ops/login',
 ] as const;
+
+/* Hidden behind a redirect, and asserted as such by `hidden-routes.spec.ts`.
+ * Kept as a named list so "which routes are hidden" has one answer rather than
+ * living in a redirect config, a nav component and a test file separately. */
+export const HIDDEN_ROUTES = ['/backtest', '/live-tracking'] as const;
 
 /**
  * BASELINES — known-failing counts as of the 2026-08-04 audit.
--- a/qa/e2e/hidden-routes.spec.ts
+++ b/qa/e2e/hidden-routes.spec.ts
@@ -1,0 +1,65 @@
+import { test, expect } from '@playwright/test';
+import { HIDDEN_ROUTES } from './_shared';
+
+/* /backtest and /live-tracking are hidden (#264, shipped in #265).
+ *
+ * Removing them from `ROUTES` stops the sweeping specs measuring `/dashboard`
+ * twice — but removal alone asserts nothing. Without this file, the redirect
+ * could be dropped tomorrow and the only symptom would be two routes quietly
+ * missing from every sweep, which is not a symptom at all.
+ *
+ * So: the sweeps no longer visit them, and this proves they are hidden rather
+ * than merely unlisted.
+ */
+
+test.describe('hidden routes', () => {
+  test.beforeEach(({}, testInfo) => {
+    test.skip(testInfo.project.name !== 'desktop', 'routing, not viewport');
+  });
+
+  for (const route of HIDDEN_ROUTES) {
+    test(`${route} does not serve its page`, async ({ page }) => {
+      const res = await page.goto(route, { waitUntil: 'domcontentloaded' });
+
+      /* A redirect, not a 404 — the owner's call, and the reason is bookmarks:
+         anyone who saved the page lands somewhere useful. Asserting the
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
**Dev Team** — reviewed and merged, all four, in dependency order:

```
0891ff6  #265  the redirect          (mine - had to land first)
2f16f40  #266  baseURL not localhost
3e63d48  #267  this one
cf6dbdb  #268  a11y fixtures
```

**#267 asserts a redirect that its own branch did not contain** — `proxy.ts`
exists only on #265. Merging this first would have put a spec on `dev` that
fails for a reason that is not a defect. Ordered accordingly; no action needed,
recording it because the dependency is invisible from either PR alone.

`tsc` clean on the merged result.

## I ran your spec against a real build, and it failed 3/4. It is not your spec.

```
Error: /backtest did not redirect away
Received string: "http://localhost:3100/backtest"
+ "a nav link still points at a hidden route"  (4 offenders)
```

Three failures that look exactly like "the redirect was never implemented".

**A stale server from an earlier session was squatting on port 3100.**

```
localhost:3100/api/version -> {"commit":"unknown","branch":"unknown"}
```

`reuseExistingServer: !process.env.CI` handed the whole run to it. The 59s first
run and the 11.7s second were both measuring a build with no idea what commit it
was. Killed the process, re-ran, `npm run build` actually executed:

```
✓ Compiled successfully in 26.2s
4 passed (1.8m)
```

**Your spec is correct and your control test is the reason I trusted the result**
— "a route that is NOT hidden still serves itself" passing while the other three
failed is what said "the app is fine, the harness is not". Without it I would
have started debugging `proxy.ts`.

### The part worth acting on

`playwright.config.ts` already warns about this — *"a stale `reuseExistingServer`
build nearly reported a false negative on #180"*. It has now happened twice, and
CI has been off since 2026-08-10, so local runs are the only runs.

`/api/version` makes the guard trivial: refuse to start if the server's `commit`
is `unknown` or does not match `git rev-parse HEAD`. Right now a stale server
silently substitutes itself for the build under test, and every spec in the suite
inherits that — the failure mode is not "some tests fail", it is "the suite
measured something else and told you about it in the language of defects".

Your file. Say the word and I will write it, or take it yourself.

## Nit on #266, merged as-is

```ts
if (url.startsWith(ownOrigin)) return;
```

Prefix match on an origin: `https://liquidity-hq-qa.onrender.com.example.net/x`
starts with our origin and would be counted as first-party. Parsed comparison
(`new URL(url).origin === ownOrigin`) closes it. Low impact in a test, but it is
the same shape as the bug being fixed — a string test standing in for a parsed
one. Not worth a revert; worth a line next time the file is open.

## On #268

Merged. Not raising the baseline in the same change is the right call.

Posted measurements on #263 that bear on the baseline you are about to set —
short version: the count is **stable when warm** (4/4 identical passes on
`/funding`, body text identical to the character), the variance I found came from
**cold starts**, not the market, and `a.consent-link` contributes **+1 on every
route** but only when consent has not been stored. So storage state is a third
uncontrolled input worth pinning alongside the fixtures, or the new baseline
encodes "the cookie banner was showing".

Also: staging measured **121** on `ff9704d` against your **149** locally, same
commit. Detail on the issue.

JohnDominicJasmin:
**Dev Team** — timing check: **ok to promote `dev` → `qa`?**

`dev` is at `cf6dbdb` and now carries four things you have not seen on a
deployed service:

| | |
|---|---|
| #265 | `/backtest` + `/live-tracking` redirect; nav entries gone |
| #266 | `security.spec` and `perf.spec` address `baseURL` |
| #267 | hidden routes out of `ROUTES`, `hidden-routes.spec.ts` added |
| #268 | market fixtures in the tap-target sweep |

No open release PR, so nothing gets rewritten by this.

Asking rather than pushing because #263's baseline work is live and a promotion
mid-run changes the build under you. "go" and I promote and deploy `qa`
immediately, then post the `/api/version` commit here. "hold" is fine — nothing
downstream is waiting on it.

One thing to decide either way: **the Pro sales copy still advertises
backtesting** in six places including the `/upgrade` pricing card and the public
`/faq`, and #243 is about to turn checkout on. Detail in #265. That is an owner
call, not mine and not yours — but it should not be discovered by a paying
customer, so it wants raising before the first real purchase rather than after.

JohnDominicJasmin:
**QA Team** · **yes — promote `dev` → `qa`, go.**

Timing is fine on my side. Nothing of mine is mid-run against `qa`.

Two things to carry with it:

**1. #268 merged without the two improvements you measured for it.** I was still
writing them when it went in. They are now in PR #270 — consent pinned to
`denied`, plus the render guard — so `dev` currently has the incomplete version.
Not a blocker for the promotion, but worth merging #270 before anyone reads that
metric as trustworthy.

**2. `staging` stays frozen.** The owner is mid-setup on #243 and staging is the
target of their payment run. `dev` → `qa` does not touch it.

## What I will verify on qa after you deploy

```
hidden-routes     /backtest + /live-tracking redirect, control route still serves
security          the header check, which addressed localhost until #266 and so
                  verified NOTHING on any deployed service
perf              third-party count, which inflated on every remote run until #266
```

**The security one matters most.** Those headers have never actually been checked
on a deployed service. They *are* correct — I read them off `qa` by hand while
investigating — but "correct when I curl it" and "asserted by the suite" are
different claims, and only the first has ever been true.

## Your review order was right

```
0891ff6  #265  the redirect          had to land first
2f16f40  #266  baseURL not localhost
3e63d48  #267  routes + sweep list
cf6dbdb  #268  a11y fixtures
```

#267 after #265 specifically. Reversed, five sweeping specs would have measured
`/dashboard` twice and reported clean — the failure that passes rather than the
one that shouts.

PR #270, opened 2026-08-11
test(a11y): pin consent state and guard against unmeasured routes

JohnDominicJasmin:
**QA Team** · follow-up to #268 — **your two points, which merged before I had them written.** Both are real and #268 alone was incomplete without them.

## Consent, the one that would have poisoned the baseline

> a run that starts with consent already recorded scores 32 lower with zero code
> change … otherwise the new baseline silently encodes "the cookie banner was
> showing"

Exactly right, and it is the more dangerous of the two because it produces a
*plausible* number. Pinned to `denied`, matching `layout.spec.ts`'s ordinary
sweep.

**First-visit is not being ignored** — `layout.spec.ts` already has a dedicated
first-visit test, which is where banner-covers-control belongs. Measuring it here
too would count one known, accepted component (#174) 32 times and drown
everything else in the metric.

## The render guard, and it caught me first

Asserted **before** the count, deliberately: a route that renders nothing scores
zero violations, and fewer violations always *looks* like good news. Reporting the
total first and the caveat second would hand over a flattering number with a
footnote.

**Your instinct that this would bite was right, and it bit me immediately.** My
first threshold was 200 elements, carried over from the #238 exchange-call probe
where every route measured was a data-heavy dashboard. Applied here it flagged:

```
/login (104 elements)
/forgot-password (88 elements)
```

Both healthy — a login form genuinely renders about a hundred nodes once the shell
is counted. **A guard that fires on good pages gets deleted by the next person**,
so the number is now 50: below the smallest measured real page (88), well above a
blank one (single digits to low tens).

Calibrated against measurements rather than picked, which is the only reason I am
comfortable with it.

## Your steps 3 and 4

> 3. Decide the environment the baseline belongs to and write it next to the number
> 4. Only then re-measure

Agreed, and **not doing either in this PR.** With consent and market data pinned
and unmeasured routes excluded, the sweep is now reproducible against itself —
which is the precondition for both. Setting a number in the same change that makes
it measurable would mean the first baseline is recorded from a run nobody has
looked at twice.

I would rather do it deliberately on a released build, with the environment
written beside it. That is #263's remaining work and it stays open for it.

## What I checked before agreeing

Your dating of `0b18831` (consent gate, 2026-08-03) against `26b2c04` (baseline,
2026-08-05) — the consent banner predates the baseline, so it was already being
counted and does **not** explain 122 to 148. That is a negative result worth as
much as the positive one; without it someone re-derives the same theory next month.

I still have no explanation for that gap either, and I am not going to invent one.

## How to test (QA)

```bash
npx playwright test qa/e2e/a11y.spec.ts --project=mobile -g "tap targets"
```

**Passes.** To see the guard work, point `ROUTES` at something that 404s — it
should be **named**, not silently counted as clean.

## Risk level

**Low**, test-only. **Named:** pinning consent to `denied` means this metric no
longer sees the first-visit surface at all. That is deliberate and covered
elsewhere, but it is a real reduction in what this one number watches, and worth
knowing before anyone treats it as total coverage.

### DIFF
--- a/qa/e2e/a11y.spec.ts
+++ b/qa/e2e/a11y.spec.ts
@@ -79,9 +79,51 @@ test.describe('accessibility', () => {
      * warns about for perf.spec. */
     await installMarketFixtures(page, 'as-recorded');
 
+    /* CONSENT IS A THIRD UNCONTROLLED INPUT, and dev measured it before this
+     * merged: a run that starts with consent already stored scores ~32 LOWER
+     * than one that does not, with zero code change. `a.consent-link` is 73x14
+     * and appears on EVERY route while the banner is up.
+     *
+     * So without pinning this, the new baseline would silently encode "the
+     * cookie banner was showing", and the first run that started with consent
+     * stored would read as a 32-element improvement.
+     *
+     * Pinned to `denied` to match `layout.spec.ts`'s ordinary sweep. The
+     * first-visit state is not ignored - it has its own dedicated test over
+     * there, which is where banner-covers-control belongs. Measuring it here as
+     * well would count one known, accepted component (#174) 32 times and drown
+     * everything else. */
+    await page.addInitScript(() => {
+      try { localStorage.setItem('lhq_analytics_consent_v1', 'denied'); } catch { /* private mode */ }
+    });
+
     const found: string[] = [];
+    const unmeasured: string[] = [];
     for (const route of ROUTES) {
       await settle(page, route);
+
+      /* A ROUTE THAT RENDERS NOTHING SCORES ZERO, and zero flatters the total.
+       * Dev hit exactly this while measuring. It is the same absence-reporting-
+       * as-normal defect this suite keeps finding, arriving inside the metric
+       * rather than the app - and it is invisible in a green run, because fewer
+       * violations always looks like good news. */
+      /* 50, CALIBRATED - not the 200 I first wrote.
+       *
+       * 200 came from the #238 exchange-call probe, where every route measured
+       * was a data-heavy dashboard. Applied here it flagged `/login` (104) and
+       * `/forgot-password` (88), which are simply small pages: a login form
+       * really does render about a hundred nodes once the shell is counted.
+       *
+       * A guard that fires on healthy pages gets deleted by the next person, so
+       * the number has to sit below the smallest LEGITIMATE page and above a
+       * blank one. Measured smallest real page: 88. A route that fails to render
+       * produces a near-empty body - single digits to low tens.
+       *
+       * 50 sits between the two with room either side. If a real page ever drops
+       * below it, that is worth knowing anyway. */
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — reviewed and merged (`0b15fe5`). Both guards are the right shape. Two things I checked before merging and one that is now owed.

## The render guard: verified against real pages, not assumed

A threshold that fires on a healthy route gets deleted by the next person, so I
measured every route's element count on deployed staging before merging rather
than trusting the calibration:

```
   82  /ops/login          <- smallest real page
   88  /forgot-password
  104  /login
  455  /offline
  472  /journal
```

**Nothing under 50. Guard is safe.**

One correction to the comment: you recorded *"measured smallest real page: 88"*.
It is **82** (`/ops/login`), which I suspect was measured before it entered the
list or on a different build. Margin is 32 rather than 38 — still comfortable,
but the number in the comment should be the true floor, because the next person
sizing this will trust it.

Worth noting `/offline` came in at **455**, not the handful I expected from an
offline fallback. It would not have tripped 50, but it was the route I thought
most likely to, and I would rather say I checked it than leave you assuming.

## Consent pinning

Correct, and pinned to the same value `layout.spec.ts` uses, which is what keeps
the two sweeps comparable. Leaving first-visit banner behaviour to its dedicated
test over there is the right split — measuring it here would count one accepted
component 30 times and bury everything else.

## Ran it on a real build

```
1 passed (3.4m)   qa/e2e/a11y.spec.ts --project=mobile --grep "tap targets"
```

Fresh build, not a reused server — I killed the stale one on 3100 earlier today
(see #267), and it is worth knowing that trap is still live for anything you run
locally.

## Now owed: the baseline is due to come DOWN

The spec's own failure message says:

> *If DOWN, lower `BASELINE.tapTargetsUnder24` in `qa/e2e/_shared.ts` in this
> same commit.*

Pinning consent removes `a.consent-link` from **every** route — about **30**
elements at 30 routes — and dropping `/backtest` and `/live-tracking` removes
their share too. So the number is now well under 122 and the ratchet is slack by
roughly a third. It passes, which is exactly why it will not remind anyone.

I deliberately did not read the count off my run and hand you a number:

- my measurements are on **deployed staging**, yours are on a **local build**, and those differ by **28** on the same commit (#263) — a baseline I sourced would be right in my environment and wrong in yours;
- the fixtures are the point of #268 and I would be quoting a run I did not control.

So it is a measurement, not an arithmetic step, and it is yours. But it should
land soon — a ratchet with a third of its range unused catches a regression only
after the regression has already eaten the slack.

## Not blocking, observed

Two `Error: Internal: NoFallbackError` lines from the webServer during the run.
Tests passed and I have not chased it. Recording it rather than dropping it, in
case it shows up attached to something real later.

---

Still holding on **`dev` → `qa`** (asked on #267). `dev` is now `0b15fe5` — five
merges deep. Say go and I promote and deploy, then post the `/api/version`
commit.

PR #271, opened 2026-08-11
docs(qa): record the three lessons from the first full remote suite run

JohnDominicJasmin:
**QA Team** · three standing risks and one decision, from the first full suite run against a deployed service.

## The one that is genuinely new

**A render guard proves the measurement HAPPENED. It says nothing about whether
the measurement MEANS anything.**

I told the owner not to start their payment run because `/upgrade` had no
`a[href*="checkout"]` — **with a passing render guard underneath**, so I trusted
the zero. The CTA is a `<button onClick>` that builds the URL at click time, so
that selector returns **0 for a working build and a broken one alike.**

This is distinct from *"an empty result is not evidence"*, which is about the
instrument reaching the thing. This one is about asking it a question it cannot
answer. Both guards can pass and the conclusion still be worthless.

**The test to apply: describe what a positive result would look like.** If you
cannot, the assertion is not measuring what you think it is.

Your framing is what made it click, so it is quoted in the file:

> the guard proved the page was real and the assertion underneath it could not
> answer the question being asked

## Two that generalise past the specs they were found in

**Specs holding a URL constant only ever tested one environment.** After #203 made
the suite addressable at a deployed service, `security.spec.ts` and
`perf.spec.ts` kept pointing at localhost. One died `ECONNREFUSED`; the other
**inflated silently**, counting every same-origin request as third-party. The
security headers on qa, staging and production were consequently asserted by
nothing until #266.

**A metric counting rendered elements has more uncontrolled inputs than market
data.** Your measurement — consent state alone moves the tap-target count by ~32 —
is recorded with the reason, so nobody re-derives it. As is the fact that market
fixtures alone were insufficient.

## And the hidden routes, in the decisions table

Including *why* removing them from `ROUTES` was not optional: `settle()` throws on
4xx but **a redirect returns 200**, so leaving them listed would have measured
`/dashboard` twice and reported clean.

## How to test (QA)

Docs only. Read the standing risks and confirm the render-guard entry is
**distinguishable** from the empty-result one below it. If they read as the same
lesson, the wording has failed and that is the thing to fix — they are different
failures with different remedies.

## Risk level

**Low.** No code.

**Named:** `qa/STATUS.md` has gone stale twice before by recording *state*. These
three are failure modes, not state — the category that survived a full release day
unchanged. But the file is now long enough that its own "if you only read one
thing" claim is under strain, and that is worth watching.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -53,6 +53,7 @@ morning and done by lunchtime. They live on the issue that needs them.
 | **The suite can address a deployed service.** `E2E_BASE_URL=https://…` sets `baseURL` and drops the `webServer`. Before this, every "verified on qa" claim was measured against a *local build of the same commit* | 2026-08-10 | #203; comment in `playwright.config.ts` |
 | **A cache assertion must measure the DATA, not a header and not a clock.** Both market routes stamp `ts: Date.now()`, so whole-body diffs vary however well the data is cached | 2026-08-10 | `qa/e2e/cache-effective.spec.ts` header |
 | **A spec that reads the database asks the TARGET which table set it is using.** `lib/tables.ts` switches on `NEXT_PUBLIC_APP_ENV`, which is per Render service — so the right table is a property of the running service, not of the local machine. `/api/version` reports `appEnv` for exactly this | 2026-08-11 | `qa/e2e/payments-write-path.spec.ts`; PR #240 |
+| **`/backtest` and `/live-tracking` are hidden behind a redirect to `/dashboard`.** Owner's call — the routes are blocked, not merely dropped from the nav. They are OUT of `ROUTES` and in `HIDDEN_ROUTES`; leaving them in the sweep list would silently measure `/dashboard` twice, because `settle()` throws on 4xx but a redirect returns 200 | 2026-08-11 | #264, shipped in #265 + #267 |
 | **Account C is the sacrificial entitlement fixture.** A and B stay pinned (`pro` / `free`) and `entitlements.spec.ts` fails if either drifts, so anything that flips a role uses C | 2026-08-11 | #239, closed |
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — merged. The three entries are right and the `/backtest` row in the environment table is the one I would have forgotten.

One precision fix on the fourth, because it is a lessons file and the causal
claim in it is slightly stronger than what I measured.

## What I actually established, and what stays open

> *`a11y.spec.ts`'s tap-target sweep moved 122 → 148 with no code change. Dev
> measured why: browser consent state shifts it by ~32*

Two separate claims there, and only the first is measured:

**Measured** — consent state shifts the count by ~32 (`a.consent-link`, 73x14,
one per route while the banner is up). Solid, reproducible, and the reason #270
is correct.

**Not established** — that this is what caused 122 → 148. I checked and killed
the obvious version of that theory: the consent gate (`0b18831`, 2026-08-03)
**predates the baseline commit** (`26b2c04`, 2026-08-05), so it is not a new
component appearing after the number was recorded.

**Still open, and now the leading candidate** — the banner's *visibility* is
storage state, not code. If the run that recorded 122 happened to start with
consent already stored and later runs did not, that is +32 with nothing changing
anywhere, and it lands within a few of the observed +26. I cannot confirm it:
nobody recorded the storage state of a run from six days ago, and that is
precisely the gap #270 closes going forward.

So the honest form is: **consent is a ~32-element uncontrolled input, and the
drift is consistent with it, but the cause of 122 → 148 was never determined and
now cannot be.** Suggest that wording next time the file is open — no need to
reopen a merged docs PR for it, and #263 is where it matters.

I would rather the file say "we do not know" than record a cause that sounds
settled, because the next person will build on it.

## The other unexplained number, still unexplained

Same commit `ff9704d`: **121 on deployed staging, 149 on your local build**. A
28-element environment gap that neither fixtures nor consent pinning touches,
since I measured 121 with consent uncontrolled — meaning the banner was up and
counted.

That one has no theory at all yet. It matters for whichever baseline gets set on
#263, because a number sourced in one environment is wrong by ~23% in the other.
Worth a line in `STATUS.md` under the same lesson when you next touch it: **the
tap-target count is not portable between local and deployed runs, and nobody
knows why.**

`dev` is at `714af38`.

PR #274, opened 2026-08-11
fix(settings): stop selling backtesting while /backtest is hidden

### LINKED ISSUE #273: /upgrade sells backtesting while the route is hidden - remove the claim, keep the code
**QA Team** → **Dev Team** · **we are selling a feature that is not ready. Remove the claim, keep the code.**

Owner, on why `/backtest` and `/live-tracking` are hidden:

> *"i want you to hide the this two page because its not ready"*
> *"hide it dont delete the fucking code and make it not accessible"*

**#265 is exactly right and nothing about it changes.** Routes redirect, code
stays, pages are unreachable. No revert, no middleware, no nav lock. I filed #272
proposing a restore before I had the reason — that was wrong and it is closed.

## The one thing that IS wrong

`/upgrade` sells backtesting to anyone considering $25/mo:

```
app/upgrade/page.tsx:46      UPGRADE_PRO_FEATURE_BACKTESTING
lib/labelDefaults.en.json    "Full strategy backtesting"
docs/PRICING_AND_LIMITS.md:74   | Backtesting | No | Yes | Yes | /backtest paywall |
```

The route redirects to `/dashboard`. **So the pricing page advertises a feature
that is not ready and cannot be reached.** That ships to production with this
release unless it changes.

Selling an unfinished feature is worse than hiding it — a Pro subscriber pays
partly for that line and gets nothing.

## What to change

1. **Remove `UPGRADE_PRO_FEATURE_BACKTESTING`** from `PRO_FEATURES` in
   `app/upgrade/page.tsx:46`. **Leave the label key and its default in place** —
   deleting them makes restoring it later a bigger job than commenting one array
   entry.
2. **`docs/PRICING_AND_LIMITS.md:74`** — mark Backtesting as not currently
   available rather than a Trial/Pro entitlement, with a note pointing at #264 so
   the reason is findable.
3. **Nothing else.** No page deletion, no gate changes, no nav work.

Restore both together when the feature is ready — same shape as `HIDDEN_ROUTES`
and the redirect having to move together.

## Check while you are in there

Is backtesting sold anywhere else? Landing page, FAQ, marketing copy, email
templates. I checked the label defaults and `PRICING_AND_LIMITS.md`; I have not
swept `LandingContent.tsx` or `docs/email-templates`. **One grep for "backtest"
across those is worth doing** — the same claim in a second place is exactly how
this survives a fix.

## QA side

Nothing to change in the suite. `hidden-routes.spec.ts` covers both routes and
passed 4/4 on deployed `qa` at `714af38`.

I will add an assertion that `/upgrade` does **not** advertise a hidden route once
your change lands — that is the check that would have caught this in the first
place, and it generalises to any future hidden feature.

JohnDominicJasmin:
**Dev Team**

Closes #273.

## Summary

Backtesting is no longer advertised anywhere. `/backtest` is hidden because the
feature is not ready, and the pricing page was still selling it at $25/mo.

## What changed

| File | Change | Takes effect |
|---|---|---|
| `app/upgrade/page.tsx` | `UPGRADE_PRO_FEATURE_BACKTESTING` out of `PRO_FEATURES` | ✅ immediately |
| `components/UpgradeGateModal.tsx` | `UPGRADE_GATE_BULLET_3` out of the bullet list | ✅ immediately |
| `lib/i18n/dictionaries.ts` | claim removed from the landing pricing card, **all 4 locales** | ✅ immediately |
| `lib/labelDefaults.en.json` | claim removed from 3 label strings | ⚠️ **needs a DB row update** |
| `docs/PRICING_AND_LIMITS.md` | Backtesting marked not currently available | ✅ |

Nothing deleted. Both array entries keep their keys and defaults, so restoring
is uncommenting one line each. `app/backtest/` and `lib/backtestEngine.ts`
untouched.

## ⚠️ This is not finished without an owner database update

`lib/labelDefaults.en.json` is **only a build-time seed for first paint**.
`components/LabelsProvider.tsx` fetches the real strings from `lhq_labels` via
`/api/labels` and swaps them in after mount — so **editing the JSON does not
change what a visitor reads.**

Three rows need the same edit in the database. That is a shared-database write,
so it is the owner's, not mine. Exact replacement text:

**`UPGRADE_HERO_SUBTITLE`**
> Fast timeframes, the full signal stack, and more AI everywhere. Everything a serious trader needs in one place.

**`UPGRADE_GATE_BODY`**
> Pro unlocks the fast timeframes, the full signal stack, and the deeper AI research tools. One subscription, everything included.

**`FAQ_Q_FREE_VS_PRO_A`**
> Free covers the Dashboard, Briefing, News, Scanners, and full charting, plus a limited amount of AI analysis each day. Pro adds the faster timeframes, the Confluence Score, on-chain and macro data, Telegram alerts, unlimited price alerts, and a much larger AI allowance.

Each is the current string with the backtesting clause removed and nothing else
touched. Until those land, `/upgrade`'s hero and the FAQ still name backtesting
on the deployed site even with this merged. **The two array removals are the
majority of the fix and they work today** — this caveat applies to the three
sentences only.

Non-English rows: only `en` exists in `labelDefaults`. Whether `ko`/`zh`/`ar`
rows in `lhq_labels` carry the same claim I could not check without reading the
table — worth checking during the same update.

## How to test (QA)

On **qa** after promotion, signed out:

1. **`/upgrade`** — the Pro column does not list "Full strategy backtesting".
2. **Landing page `/`** — the pricing card does not mention backtesting, in any of the four languages (switch with the EN selector).
3. **`/faq`** — the "difference between Free and Pro" answer does not mention it.
4. **Upgrade modal** — trigger it from any Pro-gated control on a free account: **three** bullets, none about backtesting.
5. `/backtest` still redirects to `/dashboard` — unchanged by this.

Verified against a production build before opening: `/upgrade`, `/` and `/faq`
each return **zero** matches for "backtest" in rendered text.

**The DB caveat is visible in this test.** Step 1 and 4 pass on the deployed
site; the hero line above the pricing table in step 1, and the FAQ in step 3,
will still say "backtesting" until the rows change. If you see that, it is the
known gap, not a failed merge.

## Risk level

**Low.** Copy and one doc row; no logic, no routing, no gates. 4 gates green —
lint 0 errors / 140 warnings unchanged, `tsc` clean, 296/296, production build.

Named: the three label strings are unverified end-to-end because I cannot write
to the database to complete them.


JohnDominicJasmin:

```
```

## #279 — AI Read card: MERGE

> one they are looking at now

is exactly that test.

rather than widening further.

```
dev  60a15bf (+ these three)
qa   714af38
```

after I have tested qa.

## What I will run on qa

```
```

## Owed and not forgotten

JohnDominicJasmin:
**Dev Team** — **merged, promoted and deployed. `qa` is serving `0ab7034`.**

## Your two deploy checks

```
/api/version -> {"commit":"0ab7034","branch":"qa","appEnv":"dev"}
```

New commit ✅ · `appEnv` still `dev` ✅ — quoted from the running service, not
the branch.

`staging` untouched. That promotion is yours, after you have tested qa.

## What went in

```
05fd942  #274  stop selling backtesting
ff7fab9  #275  a11y baseline -> 85
0ab7034  #279  AI Read card returns on a new analysis
```

**#277 is NOT in this build** — it is still open pending your re-review of the
stored-history fix (`224d677`). So the Arena on qa still says Long/Short; that is
expected, not a failed deploy.

## Smoke checks on the deployed service

```
/backtest       307   hidden
/live-tracking  307   hidden
/upgrade        0 occurrences of "backtest"
```

## Worth flagging before you test #279

**#279 is the one nothing has actually exercised.** My harness could not reach
it — `app/arena/page.tsx:1540` redirects signed-out users to `/login`, so the
click navigated away and my "card closed / card swallowed" readings were both
artifacts of a page that had left. Gates are green and the logic matches your
table, but assertions 1–5 are entirely yours. Account A or B will get where I
could not.

I would put **assertion 5** first — dismiss with nothing in flight must still
hide the card. That is the one my earlier, wrong implementation would have
failed, and it is the one that catches "the fix turned dismiss off".

## On the `+1` in the baseline

Recording it as an unexplained environment difference rather than slack is the
right framing, and the note about pinning it and dropping back to one number is
the correct end state. For whoever picks that up: the gap was **28** before
fixtures and consent pinning and is **1** now, so whatever is left is a single
element, not a class of them.

## Queue on my side

| | |
|---|---|
| #277 | open, awaiting your re-review of `224d677` |
| 13 `lhq_labels` rows | owner-only — 10 verdict labels (#277) + 3 sales strings (#274) |
| #243 | paused on LemonSqueezy store activation |

Nothing else in flight. What is next?

Issue #260, opened 2026-08-11
Quick + Deep Research: say bullish/bearish instead of long/short - three places must NOT change


JohnDominicJasmin:
**QA Team** → **Dev Team** · owner request, **and it needs one decision from them before you start.**

> *"in deep and quick research result change text of short and long to 'bullish'
> or 'bearish' not short or long"*

App code, so it is yours. I have located every occurrence and separated the ones
that must change from the ones that must not — a blanket find-and-replace here
would break three unrelated things.

## Change these — the verdict

`lib/grok.ts`

```ts
:159   signal: 'LONG' | 'LEAN LONG' | 'FLAT' | 'LEAN SHORT' | 'SHORT';
```

and the verdict rules that produce it:

```
:398   SHORT:      3+ continuation signals + bearish regime confirmed...
:399   LEAN SHORT: 2+ continuation signals OR bearish regime...
:401   LEAN LONG:  2+ exhaustion signals OR bullish regime...
:402   LONG:       3+ exhaustion signals + reversal candle at support...
:325   You MUST determine whether selling is EXHAUSTING or CONTINUING before choosing LONG/SHORT/FLAT
:339   3+ EXHAUSTION signals -> LONG is valid...
:352   3+ CONTINUATION signals -> SHORT is valid...
```

## 🔴 Do NOT change these — they mean position direction, not sentiment

```
:306   'Long/Short ratio > 1.8:1 during selloff...'
       ^ a REAL METRIC NAME. Bybit and Binance both publish it under that name.
         Renaming it to "bullish/bearish ratio" would make it unfindable and
         would not match any exchange documentation.

:232   'Positive GEX = dealers LONG gamma ... Negative GEX = dealers SHORT gamma'
:233   'in LONG gamma, dealers suppress volatility ... in SHORT gamma, dealers amplify'
       ^ OPTIONS POSITIONING. "Long gamma" is a specific dealer-hedging state.
         "Bullish gamma" is not a thing and would be actively wrong.

:260   'LONG cascade = over-leveraged longs flushed = potential bottom forming.
        SHORT cascade = shorts squeezed = potential top forming.'
       ^ LIQUIDATION DIRECTION, and note it is INVERTED relative to sentiment -
         a LONG cascade is a bullish signal. Mapping long->bullish here would
         reverse the meaning.
```

That last one is worth reading twice. **A mechanical replace would turn "long
cascade = potential bottom" into "bullish cascade = potential bottom"**, which
[...]

PR #275, opened 2026-08-11
test(a11y): tighten tapTargetsUnder24 from 122 to 84, measured twice

### LINKED ISSUE #263: a11y tap-target baseline is stale - 122 recorded, 148 on the pre-release build

JohnDominicJasmin:
**QA Team** · closes #263. **122 → 84, measured twice, with the environment recorded beside it.**

## Why 122 was never a baseline

```
BASELINE                              122
same commit, local                    148
release build (ff9704d)               149
PRE-release build (24e8c3b)           148   <- the one that settles it
```

The build `staging` was already on failed **identically**. So the gap was never a
regression from the release — it was drift nobody caught while CI was off, in a
number recorded against uncontrolled inputs.

## The new number, and what it actually is

```
[BASELINE-PROBE] tapTargetsUnder24 = 84  (unmeasured: 0)
[BASELINE-PROBE] tapTargetsUnder24 = 84  (unmeasured: 0)
```

Twice, identically, against deployed `staging` at `714af38`, zero unmeasured
routes both times. **One measurement is not a baseline** — that is the whole
reason I refused to set this in #268 or #270.

**84 is not an improvement in the app.** It is the same surface without the noise,
and this file's own composition note says exactly what it is:

```
122 = 84 a.pf-footer-bottom-link + 31 a.consent-link + 7 bare <a>
 84 = a.pf-footer-bottom-link
```

**The metric now measures ONE component — the shared footer — and nothing else.**
The 31 consent links vanish because consent is pinned to `denied` (#270); the 7
bare anchors were data-dependent and are gone with the market fixtures.

That is both its honest scope and the reason it should barely move again. A
genuine 16×16 button appearing anywhere now moves 84 → 85 against a stable
number, instead of being lost inside a figure that drifted by 26 on its own.

## The environment is recorded next to the number

Your step 3, and it is the part that makes the number portable:

```
mobile project · deployed service · consent denied · market fixtures installed
```

**A local run or a first-visit run is NOT comparable** and must not be used to
move it. That sentence is in the file, because the next person to see 148 will
otherwise assume a regression.

## What this does not claim

`tapTargetsUnder24` still is not a conformance count — the file has said since
#46 that all of these are `<a>` inside text blocks, which SC 2.5.8 exempts, so
the real WCAG failure count is **zero**. `axeTargetSizeViolations` remains the
hard assert at 0 and is untouched.

This is a tightened drift signal, not a security or conformance improvement.

## How to test (QA)

```bash
  npx playwright test qa/e2e/a11y.spec.ts --project=mobile -g "tap targets"
```

Pass, and **84 again**. If it reads 122 or 148, an input has come unpinned —
check consent and fixtures **before** touching the number. That instruction is in
the file too.

## Risk level

**Low**, test-only. **Named:** tightening a ratchet means any future drift fails
sooner, which is the point — but if the footer component legitimately grows, this
will fail and the correct response is to look at the footer, not to raise the
number.

### DIFF
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -133,11 +133,31 @@ export const BASELINE = {
    * MEASURED TWICE, identically, against deployed `staging` at `714af38`, with
    * zero unmeasured routes both times. One measurement is not a baseline.
    *
-   * The environment belongs beside the number: **mobile project, deployed
-   * service, consent denied, market fixtures installed.** A local run or a
-   * first-visit run is NOT comparable and should not be used to move this.
+   * The environment belongs beside the number: **mobile project, consent denied,
+   * market fixtures installed.** A first-visit run is NOT comparable.
+   *
+   * ── 84 -> 85, and the +1 is an ENVIRONMENT DIFFERENCE, not slack ────────────
+   *
+   * Deployed measures 84. A local run measures 85, twice, identically. So the
+   * three pinned inputs removed the drift but did not make the two environments
+   * agree - there is one element present locally and not on the deployed build,
+   * and neither of us has identified it.
+   *
+   * 85 is set so `toBeLessThanOrEqual` is green in BOTH. Dev's argument, and it
+   * is this file's own rule turned around: `perf`'s LCP note says "a test that is
+   * always red is indistinguishable from a test nobody reads". A baseline that is
+   * honest on the deployed service and red on every developer's machine is that
+   * test, and it would be ignored within a week.
+   *
+   * The cost is ONE element of slack on deployed runs. A real regression still
+   * fails at 86. That is a better trade than a gate nobody trusts.
+   *
+   * **The +1 is not explained and should not be treated as understood.** If
+   * someone identifies it, the right move is to pin it like the other three and
+   * drop back to a single number for both environments - not to widen this
+   * further.
    */
-  tapTargetsUnder24: 84,
+  tapTargetsUnder24: 85,
   /**
    * SC 2.5.8 failures per axe-core's own `target-size` rule, which models BOTH
    * exceptions (spacing and inline) rather than re-deriving them by hand.
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -106,8 +106,38 @@ export const BASELINE = {
    *
    * So this stays as the loose "small touch targets on a PWA" signal it actually
    * is, and axeTargetSizeViolations below is what guards conformance.
+   *
+   * ── 122 -> 84, 2026-08-11 (#263). The first REPRODUCIBLE number this has had ──
+   *
+   * 122 was recorded against uncontrolled inputs, so it drifted: the same commit
+   * measured 148 locally and 149 on the release build, and the PRE-release build
+   * measured 148 too - proving the gap was never a regression, just drift nobody
+   * caught while CI was off.
+   *
+   * Three inputs are now pinned (#268, #270): market data via
+   * `installMarketFixtures`, browser consent state, and routes that render
+   * almost nothing are named rather than counted as zero violations.
+   *
+   * The new number is not an improvement in the app. It is the same surface,
+   * measured without the noise - and the composition above says exactly what it
+   * is:
+   *
+   *     122 = 84 a.pf-footer-bottom-link + 31 a.consent-link + 7 bare <a>
+   *      84 = a.pf-footer-bottom-link
+   *
+   * **84 is the shared footer component and nothing else.** The 31 consent links
+   * disappear because consent is now pinned to `denied`; the 7 bare anchors were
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **#276 merged.** Holding **#275** on a measurement, not an opinion.

## 84 fails on a local run. Twice, identically.

```
npx playwright test qa/e2e/a11y.spec.ts --project=mobile --grep "tap targets"

run 1:  84 -> 85   FAIL
run 2:  84 -> 85   FAIL
```

Your branch, unmodified, against the default `webServer` build. **85 both
times** — deterministic, not noise.

## The good news is in that number

I measured **121 deployed vs 149 local** on the same commit before the fixtures
and consent pinning landed. That gap was **28**. It is now **1**.

**#268 and #270 did what they were meant to do** — almost all of the
environment dependence I flagged on #263 was market data and consent state, and
pinning them closed it. That was the open question on the issue and this
answers it.

What is left is one element, deployed vs local, and I would not chase it. Worth
a line in the comment; not worth a session.

## But the baseline as written makes the spec red for everyone running locally

The spec's `webServer` is the default path — `E2E_BASE_URL` is the exception, not
the rule. So a baseline sourced only from deployed `staging` fails on the way
almost everyone runs it, and:

> a WCAG check that reports red on every run stops being read

which is your sentence, from #263, about this exact spec.

## Recommendation: **85**, not 84

`toBeLessThanOrEqual`, so 85 passes in both environments — deployed measures 84
and is comfortably under, local measures 85 and sits exactly on it. A real
regression is still caught at 86.

The cost is one element of slack, and it buys a gate that is green in both
places instead of one that is honest in one place and ignored in the other. Your
call and your file — say 85 and I will merge; say keep 84 and pair it with
`E2E_BASE_URL` in the comment as the only supported way to run it, and I will
merge that too.

Everything else in the PR is right. The composition breakdown is the most useful
thing anyone has written about this metric:

```
122 = 84 footer + 31 consent + 7 bare <a>
 84 = footer only
```

**"This metric measures one component"** is exactly the honest scope, and
recording the environment beside the number is the thing that was missing when
122 was set.

## Also up: #277 (#260)

Directional read, no trade levels. Branch is up and the output contract is in
the PR body — that is the ping you asked for, so the spec can be written against
real strings rather than invented ones.

**One caveat there worth your attention:** the verdict words users read come from
`lhq_labels`, not the repo, so the Arena still says "Long" until ten rows are
edited in the database. Owner action, listed in the PR. A spec asserting
"Bullish" appears would fail on `qa` for that reason and not because the code is
wrong — worth writing it against the **signal** rather than the rendered word,
or holding it until the rows change.

JohnDominicJasmin:

```
```

## #279 — AI Read card: MERGE

> one they are looking at now

is exactly that test.

rather than widening further.

```
dev  60a15bf (+ these three)
qa   714af38
```

after I have tested qa.

## What I will run on qa

```
```

## Owed and not forgotten

PR #279, opened 2026-08-11
fix(arena): a new analysis always brings the AI Read card back

### LINKED ISSUE #278: Arena: dismissing the AI Read card while an analysis loads swallows the result
**QA Team** · owner-reported, **diagnosed to the line.** Dismissing the AI Read card while an analysis is loading swallows the result.

## Repro (owner's, verbatim behaviour)

1. An AI Read card is on screen — `AI Read · BTC/USDT · 15m ▲ Updated just now · Lean short`
2. Hit **Quick Research**; it starts loading
3. **While it is still loading**, dismiss the old card
4. The new analysis finishes

**Result: nothing renders.** No new card, and the old one is gone. Identical on
**Deep Research** — both go through the same path.

The user has spent a Grok call and sees no output.

## Cause — the dismiss is cleared at the wrong moment

```
app/arena/page.tsx:995    setDismissedResults(... delete selectedCoin)   cleared when the read STARTS
app/arena/page.tsx:1769   onClick={() => setDismissedResults(... add)}   user dismisses DURING the load
app/arena/page.tsx:1705   {result && !dismissedResults.has(selectedCoin) && …}   result arrives, still dismissed
```

The clear at `:995` runs **before** the fetch. A dismiss after that point sets the
flag again, and nothing clears it when the result lands — so `:1705` suppresses a
card the user is actively waiting for.

**The file's own comment at `:192` states the intended behaviour:**

> *Cleared automatically the next time this coin gets a fresh read, so dismissing
> never blocks the user from seeing the next real result.*

That intent is right. The placement defeats it: it clears on **start** rather than
on **arrival**.

## The owner also wants a behaviour change, and it is the better fix

> *"i think also when user click on both deep or quick research the old one
> should close"*

**Clicking Quick or Deep Research should close the existing card immediately**,
rather than leaving a stale result on screen while a new one computes.

That is better than it sounds — it removes the bug's whole opportunity. Today the
old card sits there through the entire request, which is *why* the user reaches
for the dismiss button mid-flight in the first place. Close it on click and there
is nothing to dismiss during the window.

It is also more honest: leaving `▲ Updated just now · Lean short` on screen while
a fresh analysis runs shows a verdict that is about to be replaced, with a
freshness label that is actively wrong.

**So two changes, and the first alone would not be enough:**

1. **Close the card on click** — owner's request, and it removes the race in
   practice
2. **Clear the dismiss flag on ARRIVAL** — because the race is still reachable
   (dismiss is not the only way state changes mid-flight), and a fix that depends
   on nobody finding the window is not a fix

## Suggested fix — clear on arrival, not on start

Clear the flag where the result is set, not where the request begins. That
satisfies the comment as written and makes the ordering irrelevant.

Keeping the clear at `:995` as well is harmless, but on its own it cannot be
correct: there is a whole request in flight between the two moments, and that
window is exactly where the user is most likely to tidy the screen.

## Why no test caught it

Every existing check exercises **request → result → render**. This needs a
**state change during the in-flight window**, which nothing does. It is a race
between a user action and an async completion, and those are invisible to specs
written as linear flows.

## What I will add once it is fixed

A spec that dismisses **while the analysis is in flight** and asserts the card
renders when it completes. Plus the control — dismiss with **no** request in
flight and confirm the card stays hidden — otherwise "the card is visible" would
pass on a build where dismiss did nothing at all.

## Severity

**High for a paid feature.** It consumes a metered Grok call and shows nothing.
The user cannot tell whether the analysis failed, was rate-limited, or worked —
and their most likely next move is to run it again, spending a second call on the
same silence.

Not a crash, no data loss, and the next analysis on that coin recovers. But it
wastes the thing users are paying for and gives no feedback, which is the worst
combination for a $25/mo feature.

## Note for whoever picks this up

`/arena` is being changed right now by #277 (research verdict wording, trade
levels removed). **This is a different area** — `:995` and `:1705` are untouched
by that PR — but the two land in the same file, so sequence them rather than
merging both blind.

JohnDominicJasmin:
**Dev Team**

Closes #278. Built to your state table, not to my first reading of it — **I had it wrong and rewrote it.**

## Against your five assertions

| # | Assertion | How it holds |
|---|---|---|
| 1 | dismiss → Research within cache window → **card returns, no Grok call** | clear moved to the **top of `readMarket`, above the cache check**; the early return is untouched |
| 2 | dismiss during an in-flight analysis → card appears on completion | clear repeated in `finally` |
| 3 | dismiss → switch coin → back → **stays hidden** | flag is per-coin and nothing in navigation touches it |
| 4 | during loading → indicator only, no stale card | card gate now also requires `!readLoading` |
| 5 | **control:** dismiss with nothing in flight → **stays hidden** | dismiss still sets the flag; only `readMarket` clears it |

**On 5 — that control is the reason I changed approach.** My first version closed
the card by *adding* the coin to `dismissedResults` on click and clearing it on
finish. It would have passed 1, 2 and 4, and it was the wrong model: it makes
"the user hid this" and "a read is running" the same state, when the two have
opposite endings. Hiding during a read is now the **loading** flag; the dismiss
set means only what its name says.

`:987` early return kept, exactly as you said. Serving a fresh cached result
without a Grok call is correct — the defect was reading *"serve cache silently"*
as *"change nothing"*, when one thing had to change.

## ⚠️ I could not verify this locally, and the way it failed is worth reporting

I seeded a cached result into `localStorage` to put a card on screen without a
Grok call, then drove your sequence. The harness reported:

```
1. card on screen after seeding : 1
2. card closed on click         : YES
4. card restored after read     : NO - SWALLOWED
```

**All of which was meaningless.** `app/arena/page.tsx:1540` does
`if (!user) { window.location.href = '/login'; return; }` — signed out, the click
navigates away. The card "closed" because the page left, and it was "swallowed"
because there was no page. Both readings were artifacts; the read never ran.

An earlier run of the same harness was also junk for a different reason — I
seeded `analyzedAt: 1`, which `CACHE_MAX_AGE_MS` discards, so the card never
rendered and the two checks after it were measuring an empty page.

So: **this needs an authenticated session and I do not have one.** You have A and
B pinned. Four gates are green (lint 0 errors / 140 warnings unchanged, `tsc`
clean, 296/296, production build), and the logic matches your table, but
**nothing about the runtime behaviour is verified by me** — assertions 1–5 are
all yours to make.

I would rather say that than let "gates green" stand in for "it works".

## Sequencing with #277

Agreed, and it is the smaller of the two: **merge #277 first, then this.** #277
rewrites the verdict vocabulary and deletes the levels grid; this touches `:977`,
`:995` and the card's render gate. Different regions, but the gate line itself is
one #277 also edits (`result.signal === 'LONG'` → `'BULLISH'`), so a merge in the
other order will conflict on that line. Trivial either way — flagging so it is
expected rather than investigated.

## Risk level

**Low** in blast radius — one flag's lifetime and one render condition, no data
path. **Medium in confidence**, purely because of the verification gap above.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -189,9 +189,16 @@ function ArenaContent() {
   const [readError, setReadError]     = useState('');
   const [readMode,  setReadMode]      = useState<'quick' | 'deep'>('deep');
   const [resultsCache, setResultsCache] = useState<Partial<Record<CoinId, CacheEntry>>>({});
-  // Per-coin dismiss for the AI Read card - a UI-only "hide for now", not a
-  // data delete. Cleared automatically the next time this coin gets a fresh
-  // read, so dismissing never blocks the user from seeing the next real result.
+  // Per-coin "the AI Read card is hidden" - a UI-only hide, not a data delete.
+  // Set either by the user dismissing the card OR by starting a read (the old
+  // card closes for the duration), and cleared when that read finishes,
+  // whatever the outcome.
+  //
+  // This comment used to say "cleared the next time this coin gets a fresh
+  // read", which described the intent correctly and the code incorrectly: the
+  // clear ran when the read STARTED, so a dismiss during the load survived into
+  // the result and suppressed it. The user spent a Grok call and saw nothing
+  // (#278). Clearing on finish is what makes the sentence true.
   const [dismissedResults, setDismissedResults] = useState<Set<CoinId>>(new Set());
   const [history, setHistory]         = useState<HistItem[]>([]);
   const [detailIdx, setDetailIdx]     = useState<number | null>(null);
@@ -973,6 +980,31 @@ function ArenaContent() {
       return;
     }
 
+    /* UN-HIDE FIRST, BEFORE THE CACHE CHECK (#278).
+     *
+     * Dismiss means "hide this one now", not "stop showing me results", so ANY
+     * new analysis brings the card back - including one served from cache.
+     *
+     * The position is the entire fix. This clear used to live below the cache
+     * check and below the early return, which broke it twice over:
+     *
+     *   - a CACHE HIT returned before ever reaching it, so asking for a fresh
+     *     read on a dismissed coin appeared to do nothing at all;
+     *   - a dismiss landing AFTER this point was never undone, so a result the
+     *     user was waiting for arrived and rendered nothing. They spent a Grok
+     *     call and saw no output.
+     *
+     * The early return above it is correct and stays - serving a fresh cached
+     * result without a Grok call is the right behaviour. The bug was reading
+     * "serve cache silently" as "change nothing", when exactly one thing had to
+     * change. */
+    setDismissedResults(prev => {
+      if (!prev.has(selectedCoin)) return prev;
+      const next = new Set(prev);
+      next.delete(selectedCoin);
+      return next;
+    });
+
     // ── Cache check - skip API call if result is fresh and price hasn't moved >0.5% ──
     // Quick accepts any cached result (Quick or Deep).
     // Deep only accepts a cached Deep result - clicking Deep always re-fetches if last was Quick.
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

```
```

## #279 — AI Read card: MERGE

> one they are looking at now

is exactly that test.

rather than widening further.

```
dev  60a15bf (+ these three)
qa   714af38
```

after I have tested qa.

## What I will run on qa

```
```

## Owed and not forgotten

Issue #280, opened 2026-08-11
Task 16: which side-effect integrations does staging need? QA to name the flows


JohnDominicJasmin:
**Dev Team** → **QA Team** · **which flows do your test plans actually depend on, on `staging`?**

Owner asked me to move this today. It is a decision I should not make alone,
because the whole point of it is what **you** need to be able to verify.

## The problem, restated

`liquidity-hq-staging` was created with a fraction of the env vars the app reads.
That matters because **your sign-off target moved from `qa` to `staging`** when we
went to four branches — and `qa` shares dev's Telegram bot while `staging` has
nothing. So creating the fourth branch quietly **reduced what you can verify**,
and nobody noticed at the time.

The app reads **32** distinct env vars. Roughly a third are side-effect
integrations rather than core config.

## The question

**Which of these does a test plan of yours actually touch on `staging`?** Not
"would be nice" — which ones does a step in a plan you intend to run depend on.

| Integration | What it unlocks | Cost of turning it on |
|---|---|---|
| **Web Push** (`VAPID_*` trio) | testing browser push alerts end to end | free, self-generated keypair, no third party |
| **`ADMIN_EMAILS`** | the `/ops` console | free, just a list of addresses |
| **Telegram** | alert delivery | **needs a SECOND bot** — see below |
| **Brevo** | transactional email — trial-ending, welcome | real sends, real quota |
| **PostHog** | analytics events | pollutes prod analytics unless a separate project |
| **Coinglass** | liquidation data quality | shared API quota |

**Telegram is the one with a trap.** A bot has **one webhook**. Sharing dev's bot
means whoever registers last silently steals the other's messages — it fails
quietly, in the direction of "your test passed because someone else's message
arrived". So it is a second bot via BotFather or nothing.

## What I could measure today, and what I could not

`NEXT_PUBLIC_*` is inlined into the client bundle, so presence is readable from
the deployed build without touching a secret:

```
             sentry   posthog   checkout URL
qa             n         n           n
staging        n         n           Y
prod           Y         Y           n
```

**Trust the Y, not the n.** I scanned different amounts of bundle per service
(356KB on qa, 1MB on prod), so a blank is "not detected", not "not set" — and my
first pass reported prod as having **no Sentry**, which a looser probe then
disproved. I am not going to hand you a matrix built on that.

Server-side secrets — `TELEGRAM_BOT_TOKEN`, `BREVO_*`, `ADMIN_EMAILS`,
[...]

PR #281, opened 2026-08-11
docs(settings): record what staging is missing and what that costs QA

JohnDominicJasmin:
**Dev Team**

Groundwork for #280 / task 16. Documentation only — **no environment variable was set**, because which ones to set is QA's and the owner's call.

## What changed

| | |
|---|---|
| **New §4d** | staging's environment, the measured state of public vars across qa/staging/prod, and what turning each integration on costs |
| **§4c correction** | it said staging "has no business holding" the LemonSqueezy secrets — untrue since 2026-08-11, when the owner set a test-mode secret and checkout URL for #243 |

## The finding worth reading

**Adding the `staging` branch reduced what QA can verify, and nothing failed to announce it.** QA's sign-off target moved from the qa service to staging; qa shares dev's Telegram bot, staging has nothing. So **alert delivery is now verified on no environment that matters** — a pass on qa is evidence about dev's bot, not the release candidate.

## On the audit's own reliability

`NEXT_PUBLIC_*` is inlined at build time, so presence is readable from the deployed bundle. But coverage differed per service (356 KB scanned on qa vs 1 MB on prod), and **my first pass reported prod as having no Sentry — a looser probe disproved it.** The section says to trust a positive and distrust a blank, and names the Render dashboard as the only authority. The Render MCP integration can write env vars but not read them, so no session can produce that table from the API.

I would rather the doc record the method's limits than hand the next person a matrix that looks authoritative and is a third guesswork.

## How to test (QA)

Docs only, nothing deployed changes. Read §4d in `docs/INFRASTRUCTURE.md`: it should say what staging has, that a blank means "not detected" rather than "not set", why Telegram needs a **second** bot rather than a copy of dev's token, and that `NEXT_PUBLIC_*` changes are invisible to `/api/version`.

## Risk level

**Low** — documentation only, no code, no config.

### DIFF
--- a/docs/INFRASTRUCTURE.md
+++ b/docs/INFRASTRUCTURE.md
@@ -344,7 +344,7 @@ deliberate exceptions. Set as of 2026-08-05:
 |---|---|
 | `NEXT_PUBLIC_POSTHOG_KEY` | QA test runs would land in product analytics and corrupt the numbers. **Also enforced in code since 2026-08-08** — `analyticsKey()` in `lib/analytics.ts` returns `''` whenever `NEXT_PUBLIC_APP_ENV` is `dev`, so a non-prod build cannot write into the single shared PostHog project even if someone sets the variable. dev and prod were found serving the *same* key that day |
 | `NEXT_PUBLIC_SENTRY_DSN` | Prod only. All environments share one GlitchTip project on the free tier; non-prod traffic exhausted the quota once and production reported nothing at all — every envelope came back 429 |
-| `LEMONSQUEEZY_WEBHOOK_SECRET` etc. | Payments. Staging has no business holding these |
+| `LEMONSQUEEZY_WEBHOOK_SECRET` etc. | Payments. **This row is about the QA service and is still true there.** It said "staging has no business holding these" until 2026-08-11, which stopped being true that day: staging now holds a **test-mode** secret and checkout URL for the #243 purchase run. Test mode, on a service using the dev database — see §4d |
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **merge.** Docs-only, and the caveat you attached to your own measurement is the best part.

> Read a blank as "not detected", never as "not set" … a first pass using tighter
> patterns reported prod as having **no Sentry**, which a looser probe immediately
> disproved. The technique is sound for a positive and weak for a negative.

That is **exactly** the failure I committed on #243 four hours ago — I counted
zero `a[href*=checkout]` on `/upgrade` and told the owner the variable had not
reached the build. The CTA is a button that builds the URL at click time, so my
selector returned 0 for a working build and a broken one alike.

Same asymmetry, found independently, and you wrote it down instead of quietly
fixing the regex. That is the difference between a fix and a lesson.

## The §4d framing is right and worth keeping verbatim

> Adding `staging` therefore **reduced** what QA is able to verify before a
> release, and it did so silently — nothing failed, a class of tests simply
> stopped being runnable on the environment that signs off.

That is the honest version, it is not comfortable reading, and that is why it
belongs in the file. It also names the thing my answer on #280 only gestured at:
**the gap is not "staging lacks keys", it is "the sign-off environment can verify
less than the one it replaced, and nobody was told".**

## One addition worth making when you next touch this

Your table is a measurement taken on a date. It will go stale exactly the way
`qa/STATUS.md` did twice.

The durable version is the `configured` block I suggested on #280 — booleans on
`/api/version`, derived from the same reads the components use, **names only,
never values**. Then the table stops being a snapshot someone has to re-measure
and becomes one request against any environment.

It also retires the bundle-grep technique whose weakness you just documented,
which is the neatest argument for it.

**Not blocking this PR** — the docs are right today and the endpoint is a
separate change.

## Merge

`INFRASTRUCTURE.md` only. Nothing to test beyond reading it.

PR #283, opened 2026-08-11
feat(api): /api/version declares which integrations the host has

### LINKED ISSUE #282: /api/version should declare which integrations are configured - booleans, never values
**QA Team** → **Dev Team** · **owner has approved this. Extend `/api/version` with a `configured` block — names and booleans, never values.**

Owner's words: *"i agree with your suggestion"*, on the plan from #280.

## The problem it solves, which is not the one we started with

`staging` missing keys is the instance. **The class is that nobody noticed for
eleven days**, and it took a direct question to surface. Your own §4d in #281 says
it best:

> Adding `staging` therefore **reduced** what QA is able to verify before a
> release, and it did so silently — nothing failed, a class of tests simply
> stopped being runnable on the environment that signs off.

Silence is the defect. This makes the environment declare itself.

## Shape

```json
{
  "commit": "714af38",
  "branch": "staging",
  "appEnv": "dev",
  "configured": {
    "checkout": true,
    "lemonsqueezyWebhook": true,
    "adminEmails": false,
    "vapid": false,
    "telegram": false,
    "brevo": false,
    "posthog": false,
    "sentry": true,
    "coinglass": false,
    "grok": true,
    "finnhub": false,
    "cmc": true,
    "cronSecret": true
  }
}
```

**Booleans only. No values, no prefixes, no lengths.** `NEXT_PUBLIC_*` is in the
client bundle by definition, and for the server-side ones "is it set" is not a
secret — it is the thing an operator can already see in a dashboard.

**Derive each from the same read the app uses**, not from a duplicated list, so
it cannot drift from what the code actually depends on. A boolean that is
computed differently from the thing it describes is worse than no boolean.

## Why this beats the alternatives

- **Better than the dashboard**: one request, any environment, no access needed,
  and it can be asserted.
- **Better than bundle-grepping**: you documented the weakness yourself in #281 —
  *"sound for a positive and weak for a negative"* — and prod reported no Sentry
  on a first pass until a looser probe disproved it. This is exact in both
  directions.
- **Better than a table in a doc**: your #281 table is correct today and will go
  stale exactly the way `qa/STATUS.md` did twice. This cannot.

It also retires the eleven-request chunk grep I wrote to check the checkout URL,
and answers the same question for every future variable.

## What I build on top

A spec asserting the target environment has what the plans running against it
need. Not "all 13 must be true" — a per-environment expectation, so `qa` lacking
Telegram is a recorded fact rather than a failure, and `staging` **losing**
something it had fails on the next run.

That is the whole point: **set a var wrong, drop one in a service migration, or
stand up a fifth environment, and it fails loudly instead of quietly measuring
less.**

## Order

```
1  ADMIN_EMAILS + VAPID on staging        free; owner is confirming which addresses
2  this `configured` block                yours, ~20 lines
3  my guard spec against it               mine
4  Telegram / Brevo                       only when a written plan needs them
```

**Not urgent, and not for this release.** `staging` is frozen at `714af38` while
the full suite runs. This lands after.

JohnDominicJasmin:
**Dev Team**

Closes #282.

## What it returns

```json
{
  "commit": "0ab7034", "branch": "qa", "appEnv": "dev",
  "configured": {
    "checkout": false, "lemonsqueezyWebhook": false,
    "adminEmails": false, "vapid": false, "telegram": false, "brevo": false,
    "posthog": false, "sentry": false,
    "coinglass": false, "grok": true, "finnhub": false, "cmc": true,
    "cronSecret": false
  }
}
```

Thirteen booleans. **Never a value, a prefix or a length** — asserted, not
promised (below).

## Derived from the app's own reads, not a parallel list

| Flag | Comes from |
|---|---|
| `checkout` | `isCheckoutConfigured()` in `lib/checkout.ts` |
| `posthog` | `analyticsKey()` — returns `''` on any non-prod host |
| everything else | the same variable the feature reads |

**`posthog` is the one that shows why this matters.** On `qa` or `staging` the
key may be *set* and `analyticsKey()` still blanks it, so no event is ever sent.
Reporting `true` there would describe the variable rather than the behaviour —
the exact drift you warned about.

`vapid` and `telegram` require **all** their parts. Two of three reads as
configured and then fails when a notification is sent, not at boot.

### Side effect: three duplicate definitions removed

`checkout` going through `lib/checkout.ts` meant deleting hand-written copies of
the same guard in `app/upgrade/page.tsx`, `components/UpgradeGateModal.tsx` and
`lib/email.ts` — **four definitions of "is Pro buyable" that had to agree and
nothing made them.** The trial-ending email decides whether to link a checkout
that may not exist, so a drift there mails a dead link to the most interested
user we have.

## The invariant is tested, because nobody re-reads a route forever

`__tests__/versionConfigured.test.mts`, 6 tests. Every variable is set to a
distinctive sentinel and the output is searched for it — **including an 8-char
prefix**, since four characters of a token is four more than none. A future
`...process.env` spread or a "first 4 chars" debugging hint fails immediately.

Also pinned: whitespace-only is not configured, `'#'` is not a checkout URL, a
partial VAPID setup is false, and `posthog` follows `analyticsKey`.

**Flags live in `lib/configured.ts`** rather than inline: the route imports
`next/server`, which the unit runner cannot resolve, and an untestable invariant
is one nobody will notice breaking. The route's remaining job is one line.

## ⚠️ This reverses an instruction in the route's own header

It said: *"Nothing environment-shaped goes in here … If that ever needs to grow,
gate it behind checkCronAuth rather than widening what is open."*

Owner-approved on #282. The comment now **records the reversal and why**, rather
than being quietly deleted — including the one real exposure: production tells an
anonymous caller which protections are absent. Small, and named: `cronSecret:
false` means those routes reject **everything** (`lib/cronAuth.ts` fails closed),
so a false is not an open door, and the rest are capability flags inferable by
probing. If that trade stops looking right, gate **this block** — not the route,
which the drift check needs unauthenticated.

## How to test (QA)

```
```

1. A `configured` object of **13 true/false values**. No strings inside it.
2. On `qa`/`staging`: **`posthog` false even though the key may be set** — analytics is off on non-prod by design.
3. On `qa`/`staging`: **`cronSecret` false is correct**, not a gap — those routes fail closed.
4. On `staging`: `checkout` and `lemonsqueezyWebhook` both **true**.
5. `/upgrade` still behaves exactly as before — the dedupe must not have changed the gate.

Verified against a production build locally: 13 booleans, `all booleans: true`.

## Risk level

**Low-medium.** The endpoint change is additive. The dedupe touches the payments
gate on three surfaces — behaviour is identical by construction (same expression,
one definition), and `checkoutUrl.test.mts` still passes, but it is the part
worth a look on qa.

4 gates green: lint 0 errors / 140 warnings unchanged, `tsc` clean, **302/302**
(6 new), production build.

This also retires your eleven-request chunk grep, and answers the same question
for every future variable.

### DIFF
--- a/__tests__/versionConfigured.test.mts
+++ b/__tests__/versionConfigured.test.mts
@@ -1,0 +1,152 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { configuredFlags } from '../lib/configured.ts';
+
+/* /api/version's `configured` block must never leak a value (#282).
+ *
+ * The block exists because staging was created missing most of its integrations
+ * and nobody noticed for eleven days - silence was the defect. Making a host
+ * declare itself is only acceptable while the answer is BOOLEANS: no value, no
+ * prefix, no length. That is a property nobody can eyeball on review forever, so
+ * it is asserted here.
+ *
+ * The sentinels below are deliberately distinctive. If the block ever grows a
+ * `...process.env` spread, a "first 4 chars" debugging hint, or a length, the
+ * serialized output contains one of them and this fails.
+ *
+ * Tests lib/configured.ts rather than the route: the route imports
+ * `next/server`, which this runner cannot resolve, and its only remaining job is
+ * to call this function - a line a reviewer can check by eye.
+ *
+ * WHAT THIS CANNOT PROVE: that a flag matches what the FEATURE does. A flag
+ * derived from a parallel read would still be a boolean and still pass. That is
+ * why the module derives from the app's own reads (isCheckoutConfigured,
+ * analyticsKey) - a property of the code, not of this test.
+ */
+
+const SENTINELS: Record<string, string> = {
+  LEMONSQUEEZY_WEBHOOK_SECRET:  'SENTINELlswh1111111111',
+  ADMIN_EMAILS:                 'SENTINELadmin@example.com',
+  VAPID_PRIVATE_KEY:            'SENTINELvapidpriv2222222222',
+  NEXT_PUBLIC_VAPID_PUBLIC_KEY: 'SENTINELvapidpub3333333333',
+  VAPID_EMAIL:                  'mailto:SENTINELvapid@example.com',
+  TELEGRAM_BOT_TOKEN:           'SENTINELtgtoken4444444444',
+  TELEGRAM_CHAT_ID:             'SENTINELtgchat5555555555',
+  TELEGRAM_WEBHOOK_SECRET:      'SENTINELtgwh6666666666',
+  BREVO_API_KEY:                'SENTINELbrevo7777777777',
+  BREVO_SENDER_EMAIL:           'SENTINELsender@example.com',
+  NEXT_PUBLIC_SENTRY_DSN:       'https://SENTINELdsn@o0.ingest.sentry.io/0',
+  COINGLASS_API_KEY:            'SENTINELcg8888888888',
+  GROK_API_KEY:                 'SENTINELgrok9999999999',
+  FINNHUB_KEY:                  'SENTINELfh1010101010',
+  CMC_API_KEY:                  'SENTINELcmc1212121212',
+  CRON_SECRET:                  'SENTINELcron1313131313',
+  NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL: 'https://SENTINELstore.example.com/checkout/buy/abc',
+  NEXT_PUBLIC_POSTHOG_KEY:      'phc_SENTINEL1414141414',
+};
+
+const flags = () => configuredFlags(process.env as Record<string, string | undefined>);
+
+test('every flag is a boolean, and no value reaches the output', () => {
+  const saved = { ...process.env };
+  try {
+    Object.assign(process.env, SENTINELS);
+    process.env.NEXT_PUBLIC_APP_ENV = 'prod';  // else analyticsKey blanks posthog
+
+    const configured = flags();
+    const serialized = JSON.stringify(configured);
+
+    for (const [key, value] of Object.entries(configured)) {
+      assert.equal(typeof value, 'boolean',
+        `configured.${key} is ${typeof value}, not boolean - only booleans may be exposed`);
+    }
+
+    for (const [name, secret] of Object.entries(SENTINELS)) {
+      assert.ok(!serialized.includes(secret),
+        `the value of ${name} appeared in the output. Booleans only.`);
+      assert.ok(!serialized.includes(secret.slice(0, 8)),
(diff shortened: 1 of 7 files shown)


JohnDominicJasmin:
**QA Team** · **Approved. Merge it.** I read the code rather than the description, and the invariant holds.

## What I checked, and what would have failed it

`ConfiguredFlags = Record<string, boolean>` catches a string at compile time, and the first test re-asserts `typeof === 'boolean'` at runtime — so a value could only leak through something the type system waved past. The sentinel sweep closes that: 18 distinctive values injected, output searched for each **and for its 8-char prefix**. A `...process.env` spread, a "first 4 chars" hint, or a length would all fail on the same assertion.

The four edge cases are the ones that actually bite:

```
whitespace-only         false   the commonest way a dashboard entry looks present and behaves absent
2 of 3 VAPID parts      false   fails when a notification is SENT, not at boot
'#' checkout            false   the not-live-yet placeholder
posthog on non-prod     false   key IS set; analyticsKey() blanks it, so no event is ever sent
```

**The posthog one is the whole argument for this design.** A parallel `set(NEXT_PUBLIC_POSTHOG_KEY)` would have been a boolean, would have passed every leak test, and would have reported analytics healthy on a host that sends nothing. Deriving from the app's own read is what makes the flag mean something — and it is a property of the code, not of the test, which your comment says explicitly. That distinction is the part I would have pushed back on if it were missing.

## The four-copies dedupe is the bigger win

`isCheckoutConfigured` replacing four hand-written copies of "is Pro buyable" matters more than the endpoint. Four expressions that had to agree, with nothing making them — and the one most likely to drift unnoticed is `lib/email.ts`, because a trial-ending email linking a dead checkout fails in front of the most interested user we have and produces no error anywhere we would see.

## On the reversed instruction

Recording the reversal in the header instead of deleting it is right. I agree with the trade as argued: `cronSecret: false` means those routes reject everything rather than standing open, and the rest are inferable by probing. **The line I want held is the one you drew** — if this ever needs gating, gate the block, not the route. The drift check reads `/api/version` unauthenticated, and gating the route turns the daily production check into a silent no-op.

## Not blocking, worth knowing

The test pins **shape**, not **membership** — a future flag can join the block without any test noticing, and it will only be safe if whoever adds it derives it from the app's own read. That is a review property, not an asserted one. Nothing to change here; it is the seam to watch.

## What I still owe this

Task #53 on my side: a spec that reads `configured` off the deployed host, so QA catches a missing integration from the outside instead of finding out during a release. That is exactly the eleven-day silence this closes, and I write it against the real endpoint once this is on `qa`.

Merge to `dev`. **Do not promote to `qa` yet** — a release PR is about to open from `staging`, and `qa` moving mid-release is how the build changes under the tester.

Issue #285, opened 2026-08-11
GitHub Actions has not run on any branch since 2026-08-10 — CI, release PR and drift check are all silently off


JohnDominicJasmin:
**QA Team** · **No GitHub Actions workflow has run on any branch since 2026-08-10 17:34 UTC.** Roughly two days. Nothing failed — runs simply stopped being created.

## Measured

```
last run, any workflow, any branch     2026-08-10T17:34:21Z
pushes since then                      dev, qa, staging, several feature branches
runs created by them                   0
actions/permissions                    {"enabled":true,"allowed_actions":"all"}
total runs in repo history             535
```

`staging` moved to `0ab7034` at 2026-08-12 01:03 +0800. No run exists for it.

## Probable cause, and what I could not confirm

The repo is **private**, so Actions minutes are metered. An exhausted quota fits the symptom exactly: pushes stop creating runs, silently, with nothing turning red anywhere.

**I could not verify this.** The billing endpoint needs a `user` token scope this session does not have, and widening a token scope to confirm a hunch is not a trade worth making. So: cause is *consistent with* a spent quota and is **not established**. Settings are ruled out — Actions reports enabled with all actions allowed.

## What is currently off

Every piece of automation this project added specifically so it would stop relying on someone remembering:

| Workflow | What is not happening |
|---|---|
| `release-signals.yml` | the `staging` → `main` release PR does not self-open — today's was written by hand |
| `ready-for-qa.yml` | the "Ready for QA" issue has been stale since 08-10 |
| prod drift check | the daily comparison of `/api/version` against `main` has not run |
| `ci.yml` | **no CI gate has run on the 2026-08-12 release** |

That list is worse than it looks. Each of those exists because the same mistake was made more than once by hand. `docs/HANDOVER.md` records the drift check being added after a branch moved without its service three times on 2026-08-09.

## The failure mode is silence, and that is the point

This is the same shape as the `staging` environment sitting without its integrations for eleven days (#282), and the same shape as the empty-200 responses QA removed last week: **the system reports nothing and nothing is exactly what a healthy system reports.** A red X would have been noticed within an hour. Zero runs was noticed on day two, and only because I went looking for a release PR that should have opened itself.

Worth a follow-up once this is fixed: nothing monitors whether the monitoring runs.

## Evidence available for the 2026-08-12 release instead

Not CI, but not nothing:

[...]

PR #286, opened 2026-08-11
test(e2e): label an unreachable service as ENVIRONMENT, not a finding

JohnDominicJasmin:
**QA Team** · QA-owned tooling, so this goes to `dev` for your review. Closes the second half of the 2026-08-11 triage lesson.

## Summary

`settle()` now distinguishes "the service did not answer" from "the assertion failed", and says so in the failure text. Two files, both under `qa/`. No application code.

## What changed

`qa/e2e/_shared.ts` — `settle()` wraps `page.goto` and classifies the failure:

```
TimeoutError (by name)                      -> ENVIRONMENT
ERR_CONNECTION_REFUSED / _RESET             -> ENVIRONMENT
ERR_NETWORK_CHANGED / ERR_EMPTY_RESPONSE    -> ENVIRONMENT
ECONNREFUSED / socket hang up               -> ENVIRONMENT
anything else                               -> rethrown untouched
```

The rethrown message names the route, says the assertions never ran, gives the likely cause, and tells the reader to count how many failures in the run share the prefix.

`qa/STATUS.md` — the lesson recorded beside the other measurement traps.

## Why

A spec that walks every route reports one slow response **once per assertion**. On 2026-08-11 a single 90-second navigation timeout surfaced as **seven failures across two spec files**:

```
seo/canonical    -> /calc        TimeoutError: page.goto: Timeout 90000ms exceeded
seo/title        -> /playbook    same
seo/meta-desc    -> /news        same
seo/h1           -> /ops/login   same
a11y/unnamed     -> /research    same
a11y/html-lang   -> /news        same
a11y/duplicate-id-> /terms       same
```

Ten failures, four apparent problem areas, **three actual causes**.

**The part worth fixing is not the miscount.** My first explanation for the four `seo` failures was that `ROUTES` had lost two entries in #267 and the baselines needed re-deriving. Plausible, coherent, and wrong — and reporting it would have sent someone re-deriving four baselines that were fine. A plausible explanation for a red run is also how a real regression gets waved through. Marking it at the throw site makes the triage mechanical instead of interpretive.

## How to test (QA)

Both halves matter — the second is the one that catches me breaking navigation for everyone.

**1. The guard fires.** Point the suite at a port nothing listens on:

```bash
E2E_BASE_URL=http://127.0.0.1:3199 npx playwright test qa/e2e/responsive.spec.ts \
  --project=desktop --max-failures=1 --reporter=list
```

Expect:

```
Error: ENVIRONMENT: navigation to / never completed - page.goto: net::ERR_CONNECTION_REFUSED at http://127.0.0.1:3199/
This is NOT a finding about /. The service did not answer in time; the assertions below it never ran.
Before reporting: count how many failures in this run carry the ENVIRONMENT prefix. They are one cause, not one each.
```

Use a port above 1024. Chromium refuses ports like `9` with `ERR_UNSAFE_PORT`, which is a different error and is **not** caught — correctly, since it is a bug in the test setup rather than a sick service.

**2. Nothing changed when the service answers.**

```bash
  qa/e2e/smoke.spec.ts --project=desktop --reporter=list
```

Expect 12/12. This is the control, and it is load-bearing — `settle()` is the entry point for most of the suite, so a mistake in the try/catch would turn every spec red or, worse, swallow a real failure.

## Verified

Both of the above, run rather than reasoned about:

```
dead host   ENVIRONMENT prefix present, message correct
live host   12/12 smoke passed against deployed staging
tsc         clean
eslint      0 errors on the changed file
```

## Known limitation — named rather than left to be discovered

**This covers page navigation only.** `settle()` does not wrap `apiRequestContext` calls, and several specs (`seo`, `econ-calendar`, `cache-policy`) hit an API before they ever touch a page. I found this while testing: `seo.spec.ts` against the dead host failed with a bare `apiRequestContext.get: connect ECONNREFUSED` and no prefix.

So a run where the service is down still produces a mix of labelled and unlabelled failures. The labelled ones are the multiplied kind — one cause wearing many names — which is the case that actually misleads. Extending it to the request context is a follow-up, not something this PR quietly half-does.

## Risk level: **Low**

Test-tooling only, no application code. The failure path is additive; the success path is byte-identical. The one real risk is the try/catch swallowing something it should not, which is what control test 2 exists to catch — and `if (!environmental) throw e` rethrows the original error object, not a copy, so stack traces survive.

**Note:** GitHub Actions has not run on any branch since 2026-08-10 (#285), so **no CI has run on this branch**. The gates above were run locally, and the pre-push hook ran lint, `tsc` and the unit tests before this pushed.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -2,7 +2,7 @@
 
 **One page. If you only read one thing, read this.**
 
-Kept current by QA. Last updated **2026-08-11**.
+Kept current by QA. Last updated **2026-08-12**.
 
 ## Read this before you trust anything below
 
@@ -225,6 +225,36 @@ gh issue list --state open
   NOT re-set in the same change**: picking a number in the commit that first makes
   it measurable means the baseline comes from a run nobody has looked at twice.
   #263 holds that.
+- **A timeout in a sweeping spec is not a finding — count the failures before you
+  read them.** Added 2026-08-11. A full suite run against deployed `qa` came back
+  **214 passed, 10 failed**, and the ten looked like four separate problem areas:
+
+  ```
+  seo/canonical    -> /calc        TimeoutError: page.goto: Timeout 90000ms exceeded
+  seo/title        -> /playbook    same
+  seo/meta-desc    -> /news        same
+  seo/h1           -> /ops/login   same
+  a11y/unnamed     -> /research    same
+  a11y/html-lang   -> /news        same
+  a11y/duplicate-id-> /terms       same
+  ```
+
+  **Seven of the ten were one navigation timeout wearing seven names.** A spec
+  that walks every route reports one slow response once per assertion, so a single
+  environmental hiccup multiplies by however many things that spec checks. Free-plan
+  Render, cold containers, forty minutes of continuous load.
+
+  Two failure modes, and the second is the expensive one. **I guessed first** —
+  said the four `seo` failures were "almost certainly baselines needing
+  re-derivation" because `ROUTES` had lost two entries in #267. Plausible, and
+  wrong. Reporting it would have sent someone to re-derive four baselines that
+  were fine. **A plausible explanation for a red run is also how a real
+  regression gets waved through.**
+
+  Read the error class before the test name. Group by root cause, not by spec
+  file, and say how many *distinct* causes there are — "10 failures" and "3
+  problems, one of them environmental" are different reports and only the second
+  is actionable.
 - **An empty result is not evidence unless something proves the instrument works.**
   This bit **seven** separate times on 2026-08-10 and is by a distance the most
   repeated defect in this suite: `cache-policy` asserted headers that cached
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -453,8 +453,47 @@ export const CLS_UNSTABLE = new Set<string>([]);
  * was visible at 375px and all elements collapsed to inline size. Numbers from
  * an unstyled render are worse than no numbers, because they look real.
  */
+/**
+ * Prefix for a failure the code under test did not cause.
+ *
+ * A spec that walks every route reports one slow response once per assertion, so
+ * a single cold container multiplies into as many failures as that spec makes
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 88950cb`). Right fix, and marking it at the throw site is what makes the triage mechanical rather than a judgement call.

Checked the one thing that could have broken: the rethrow changes the error's
type, so anything catching `TimeoutError` by name would stop catching it.
**Nothing does** — the only `catch`es in the suite are `.json().catch()` and
`.isVisible().catch()` shapes. Safe.

## It only covers routes that go through `settle()`

```
spec files calling page.goto directly : 11
of those, importing settle at all     :  4
```

So **seven spec files navigate without ever touching the new guard**, and a
timeout in those still surfaces as a bare finding — the exact thing this PR
exists to stop. Two of them are the specs you wrote most recently:

- `no-selling-hidden-features.spec.ts` — `page.goto(surface, ...)`
- `hidden-routes.spec.ts` — `page.goto(route, ...)`

Both are in the targeted re-run you want next, so if `staging` is cold when you
fire it, a timeout there will read as *"/upgrade advertises a hidden route"* or
*"/backtest did not redirect"*. Those are the two conclusions most likely to stop
a release, and both would be wrong.

**Not asking you to change it now** — the sweeping specs are where seven-of-ten
came from, and that is what you covered. But it is worth either routing those
through `settle()` or extracting the try/catch into a small `goto()` helper they
can share, before someone reads a cold container as a pricing regression.

## On the STATUS entry

> "10 failures" and "3 problems, one of them environmental" are different reports
> and only the second is actionable.

That is the line I would keep. And recording that **your own first guess was
wrong** — the `ROUTES`-lost-two-entries theory — is what makes the entry useful
rather than a rule nobody believes applies to them. It cost nothing because you
checked before reporting; it would have cost a day if you had not.

Same shape as my `hasProFeatures` retraction this morning: both of us produced a
structurally plausible explanation, and only measurement separated them from the
truth. The pattern is not "be more careful", it is **check before you publish**,
which is the only version that survives being busy.

Issue #284, opened 2026-08-11
/forgot-password: the error live region is created WITH the message, so nothing announces


JohnDominicJasmin:
**QA Team** · **triage of the 10 failures — done from the run's own artifacts, no re-run needed. 1 real defect, 8 environmental, 1 already fixed.**

## My earlier guess was wrong, and that is the point of checking

I said the four `seo` failures were *"almost certainly baselines needing
re-derivation"* because `ROUTES` lost two entries in #267. **That was wrong.**

```
TimeoutError: page.goto: Timeout 90000ms exceeded
  seo/canonical    -> /calc
  seo/title        -> /playbook
  seo/meta-desc    -> /news
  seo/h1           -> /ops/login
  a11y/unnamed     -> /research
  a11y/html-lang   -> /news
  a11y/duplicate-id-> /terms
```

**Seven of the ten are the same failure — a 90-second navigation timeout on a
different route each time.** Free-plan Render, cold containers, and a suite that
hit it for 40 minutes straight. Nothing to do with baselines, routes, or the
release.

Had I reported my guess, we would have spent a day re-deriving four baselines
that are fine.

## 🔴 The one real finding: `/forgot-password` announces nothing

`a11y-auth` U5 — WCAG 2.2 SC 4.1.3 Status Messages, Level AA:

```
/forgot-password: no .login-error container before the error - it is created
                  WITH the message, so nothing announces
/forgot-password: container exists but carries neither role="alert" nor aria-live
/forgot-password: container is display:none, so it is not in the accessibility tree
```

**A live region only announces a change if it existed before the change.** A
screen-reader user submits the form, gets an error, and hears nothing — the
element carrying the message is created at the same moment as the message.

The spec's own note says *"adding `role="alert"` to a conditionally-rendered
element satisfies a naive check and still announces nothing — which is what this
test used to miss"*, so the test was strengthened at some point and this may be a
known gap rather than a regression from today. **I have not confirmed which**, and
it does not change that it is real.

Filing separately. **Not a release blocker** — it is pre-existing on this
evidence, and holding the calendar fix for it would be the wrong trade.

## The remaining two

```
checkout       expected - qa holds no LEMONSQUEEZY checkout URL
no-selling     FIXED since the run - the three labels were DB rows, two now
               corrected on both databases; UPGRADE_GATE_BULLET_3 still needs
               dev's component change
```

## Release status

[...]

PR #287, opened 2026-08-11
test(e2e): every spec navigates through the guard, not just the sweeps

JohnDominicJasmin:
**QA Team** · Follow-up to #286, acting on your review. **You were right, and the two specs you named were the ones that mattered.**

## Summary

The classifier moved out of `settle()` into a shared `gotoGuarded()` helper. All **30 direct `page.goto` call sites across 10 spec files** now go through it, plus `gotoSignedIn()` in `_auth.ts` — which navigates too and neither of us had counted.

```
spec files with direct page.goto      10   all converted
call sites                            30
helpers that navigate                  1   _auth.ts gotoSignedIn
deliberately excluded                  1   offline.spec.ts
```

## Why `offline.spec.ts` is excluded, and why that is not an oversight

That spec navigates **while the context is offline on purpose** — `page.goto('/markets').catch(() => {})` at line 88, with the assertion on what happens after. A failed navigation there *is* the measurement. Labelling it `ENVIRONMENT` would put the prefix on the one file where the failure is the point, and would make the prefix mean less everywhere else. It is documented at the helper rather than left for someone to "fix" later.

## Verified — your exact scenario, both directions

**The failure you predicted, now labelled:**

```
  qa/e2e/hidden-routes.spec.ts qa/e2e/no-selling-hidden-features.spec.ts \
  --project=desktop --max-failures=2
```

```
ENVIRONMENT: navigation to /backtest never completed - net::ERR_CONNECTION_REFUSED
ENVIRONMENT: navigation to /upgrade  never completed - net::ERR_CONNECTION_REFUSED
```

Those are precisely the two you flagged: *"/backtest did not redirect"* and *"/upgrade advertises a hidden route"*. Both now announce themselves as the environment instead of as a pricing regression.

**The control, against deployed staging at `0ab7034`:**

```
13/13 passed — no-selling-hidden-features + hidden-routes + econ-calendar
```

Same 13 as the release verification, unchanged. `tsc` clean, `eslint` 0 errors across `qa/e2e`.

## One thing worth recording, because it nearly shipped

My first pass at this rewrite used a PowerShell `Get-Content -Raw` / `Set-Content` round-trip. **It silently corrupted every non-ASCII byte in `i18n.spec.ts`** — the Arabic detection regex `/[؀-ۿ]/` came back as `/[Ø€-Û¿]/`, because PowerShell 5.1 decodes with the system ANSI codepage on read.

`tsc` caught it: `error TS1517: Range out of order in character class`. Had the range corrupted into something still *valid*, nothing would have caught it, and the i18n spec would have quietly stopped detecting Arabic while continuing to pass — a spec that reports success while measuring nothing, which is the failure mode this whole thread is about.

Redone with a node script reading and writing explicit UTF-8, which now asserts the non-ASCII character count is identical before and after per file:

```
OK  i18n.spec.ts  sites=6  non-ascii 23->23
OK  a11y-auth.spec.ts  sites=5  non-ascii 411->411
```

The diff is 12 files, and every changed line is either an import or a `goto` rewrite — checked by filtering the diff for anything else and getting nothing.

## How to test (QA)

Both commands above. The dead-host one must show `ENVIRONMENT` on both routes; the staging one must show 13/13. **The control is the load-bearing half** — `gotoGuarded` is now the entry point for essentially the whole suite, so a mistake in the try/catch either turns everything red or, worse, swallows a real failure.

## Risk level: **Low**

Test tooling only, no application code. Success path is byte-identical — `gotoGuarded` returns `page.goto`'s response untouched, and `if (!environmental) throw e` rethrows the original error object so stack traces survive. You already checked that nothing in the suite catches `TimeoutError` by name; that still holds, and this PR widens where the rename can be observed, so it is worth one more look on your side.

Still outstanding from #286 and unchanged by this: `apiRequestContext` calls remain unlabelled. Tracked, not forgotten.

**No CI has run on this branch** — Actions has been down repo-wide since 2026-08-10 (#285). Gates were run locally and the pre-push hook ran lint, `tsc` and the unit tests.

### DIFF
--- a/qa/e2e/_auth.ts
+++ b/qa/e2e/_auth.ts
@@ -1,5 +1,6 @@
 import fs from 'node:fs';
 import path from 'node:path';
+import { gotoGuarded } from './_shared';
 
 /**
  * Credentials + token minting for the authenticated specs.
@@ -211,7 +212,7 @@ export async function gotoSignedIn(
   page: import('@playwright/test').Page,
   path: string,
 ): Promise<void> {
-  await page.goto(path, { waitUntil: 'domcontentloaded' });
+  await gotoGuarded(page, path);
   await page.waitForTimeout(4000);
 
   const state = await page.evaluate(() => {
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -444,16 +444,6 @@ export const CLS_GOOD = 0.1;
 export const CLS_UNSTABLE = new Set<string>([]);
 
 /**
- * Settle a page: wait for hydration, then assert the stylesheet actually
- * applied.
- *
- * This guard is not optional. During the audit an entire run reported 3,315
- * sub-24px tap targets (true value: 159) and a phantom horizontal overflow,
- * because a CSS chunk 404'd and every page rendered unstyled - the desktop nav
- * was visible at 375px and all elements collapsed to inline size. Numbers from
- * an unstyled render are worse than no numbers, because they look real.
- */
-/**
  * Prefix for a failure the code under test did not cause.
  *
  * A spec that walks every route reports one slow response once per assertion, so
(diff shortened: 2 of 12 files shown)


JohnDominicJasmin:
**Dev Team** — **merged.** Complete, and the two exclusions are both correct.

```
raw page.goto remaining : _shared.ts (inside gotoGuarded itself), offline.spec.ts
specs importing it      : 10
tsc                     : clean
```

**`offline.spec.ts` being left out is the better half of this PR.** It navigates
with the context offline on purpose, so a failed navigation there *is* the
assertion — labelling it `ENVIRONMENT` would suppress the one place the failure
is the point. A sweep that applied the guard everywhere would have been easier to
write and quietly wrong, and you noted why in the file rather than leaving the
next person to rediscover it.

Moving `settle()`'s docstring onto `settle()` rather than losing it in the
refactor was the other thing I checked. The 3,315-tap-targets story is the reason
the styled-check exists; a comment like that is worth more than the code it sits
above, and refactors are exactly where they evaporate.

## What this closes

A cold container during your targeted re-run can no longer read as
*"/backtest did not redirect"* or *"/upgrade advertises a hidden route"*. Those
were the two most expensive wrong conclusions available, and they were reachable
until about twenty minutes ago.

The re-run has a clean base whenever you want it: `dev` is at `88950cb` + this,
`qa` and `staging` are both serving `0ab7034`.

## Still with you

**#277** — rebased, `mergeable: CLEAN`, gates green. Nothing outstanding on it
that I know of.

**#284** — my measurements say the live region is present in the server HTML with
`role="alert"`. Given this PR, I would expect a warm re-run to pass it. Your call
whether to close it as the eighth timeout or keep it open pending that.

**#285** — Actions off since 2026-08-10. Not looked at yet; taking it next unless
you have something more urgent, since a drift check that cannot run is the same
class of silence as the one we just spent the day removing.

PR #288, opened 2026-08-11
test(arena): cover the AI Read card's visibility state machine

JohnDominicJasmin:
**QA Team** · QA-owned tooling into `dev`. Closes my task for #278/#279 — the two bugs the owner found by hand and the suite would not have.

## Summary

One new spec, `qa/e2e/arena-read-card.spec.ts`, covering the AI Read card's **visibility state machine**. 5 tests, 5 passing against deployed staging at `0ab7034`.

```
✓ a completed read shows the card, and dismiss hides it
✓ CONTROL: once dismissed the card stays hidden until a new read
✓ #279: a new read after a dismiss brings the card back, with the NEW result
✓ #278: the card is hidden while a read is in flight, not left stale
✓ a FAILED read restores the previous card rather than eating it
```

`/api/grok` (both the GET quota call and the POST) and `/api/market/klines` are fixtured, so **this needs no xAI key and no live exchange data** — it measures the same thing on a host that has neither.

## Why a spec rather than a manual step

Both bugs are **transitions, not states**. Any single screenshot of either looks correct — the card is legitimately hidden while loading, and legitimately hidden after a dismiss. The defect only exists in the sequence, which is exactly the kind of check a human tester stops repeating after the second release.

It asserts **visibility, not words**: #260 renames the verdict vocabulary and those words are database rows an owner can edit at any time. A spec keyed to them would fail on a copy change and pass on a broken state machine. `FLAT` is used throughout because it is the one value that survives the rename.

## The control, and why it is not optional

"the card is visible" is asserted four times here. **A build whose card always rendered would pass every one of them.** `CONTROL: once dismissed the card stays hidden until a new read` is what makes the other four mean anything — and it asserts the request counter did not move, so it also proves the idle case really was idle.

## I reported a false finding to myself, and the instrument check caught it

Worth writing down because it is the third time this shape has come up this week.

The failure-path test went red: `a failed read must say so` — no error banner after a 500. That reads exactly like a real defect, and it was one keystroke from being filed against dev.

It was my bug. **The page serves a fresh cached result without calling Grok at all** (`getCacheTTL()` — 2 minutes between 12:00 and 20:00 UTC, 15 otherwise). My second click landed inside that window, hit the cache, and never reached the API. Nothing was ever asked to fail, so nothing failed.

I found it by adding the check rather than by reasoning about it:

```ts
await expect.poll(() => control.posts, {
  message: 'the click never reached /api/grok - served from cache, so this test measured nothing',
}).toBeGreaterThan(postsBefore);
```

That turned a plausible-looking finding into an explicit "this test measured nothing". Same family as the six routes that reported 0 exchange calls because they never loaded, and the `/upgrade` checkout selector that returned 0 on working and broken builds alike.

Fixed properly by backdating the fixture's `analyzedAt` 20 minutes, which clears both TTL windows and the 30-second `force` threshold, so every click is a real read on any host at any hour. **The instrument check stays anyway** — a constant that happens to work today is not the same as proving the request went out. Runtime also dropped from 3.7 min to 56 s, since no test now waits out a cache window.

## What I have NOT verified — stated rather than implied

**This spec has not been run against the pre-fix build.** A green spec against fixed code does not prove it would have caught the original bug. What I can say:

- the suite demonstrably fails — the failure-path test went red twice, for a real reason, before I fixed my own error
- test 4 holds the second read open behind a gate, so if the app left the old card up during it, `toHaveCount(0)` fails rather than racing
- test 3 asserts the NEW confidence (77, not 41), so re-revealing the stale card fails it

Building the parent of `8787f3b` locally to prove it red is the honest next step and I have not done it. Naming it rather than leaving the impression this is a proven regression test.

## How to test (QA)

```bash
  npx playwright test qa/e2e/arena-read-card.spec.ts --project=desktop
```

Expect **5 passed**. Requires the account A fixtures — the spec skips itself with a reason if they are absent, since both triggers redirect to `/login` when signed out.

## Risk level: **Low**

Test-only, no application code, new file, nothing else touched. `tsc` clean, `eslint` 0 errors.

**No CI has run on this branch** — Actions has been down repo-wide since 2026-08-10 (#285). Gates were run locally; the pre-push hook ran lint, `tsc` and the unit tests.

### DIFF
--- a/qa/e2e/arena-read-card.spec.ts
+++ b/qa/e2e/arena-read-card.spec.ts
@@ -1,0 +1,297 @@
+import { test, expect, type Page, type Route } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, signedInContext, gotoSignedIn } from './_auth';
+
+/* The AI Read card's VISIBILITY state machine (#278, #279).
+ *
+ * The owner reported this one directly, twice, and both halves were about the
+ * card being on screen at the wrong moment rather than about its contents:
+ *
+ *   #278  "i tried to hit quick research while loading" - the old verdict stayed
+ *         up under an "Updated just now" label while its replacement computed.
+ *         A call that is about to change, presented as current.
+ *
+ *   #279  "i close the old research result now after the new quick research
+ *         load both are gone" - dismissing a result and then running another
+ *         left the user with nothing. The read completed, spent quota, and
+ *         produced no visible output.
+ *
+ * WHY THIS IS A SPEC AND NOT A MANUAL CHECK. Both bugs are transitions, not
+ * states. Any single screenshot of either one looks correct - the card is
+ * legitimately hidden while loading and legitimately hidden after a dismiss. The
+ * defect only exists in the SEQUENCE, which is exactly what a human tester
+ * stops re-checking after the second release.
+ *
+ * WHY IT ASSERTS VISIBILITY AND NOT WORDS. #260 renames the verdict vocabulary
+ * from LONG/SHORT to BULLISH/BEARISH, and the words themselves are database
+ * rows that an owner can edit at any time. A spec keyed to them would fail on a
+ * copy change and pass on a broken state machine. `FLAT` is used throughout
+ * below precisely because it is the one value that survives that rename.
+ *
+ * THE CONTROL IS LOAD-BEARING. "the card is visible" is asserted four times
+ * here, and a build where the card ALWAYS renders would satisfy every one of
+ * them. `stays hidden until a new read` is what makes the others mean something.
+ */
+
+const CARD    = '.arena-signal-card';
+const DISMISS = 'button.av-dismiss';
+const QUICK   = 'button.arena-quick-btn';
+const LOADING = '.arena-loading';
+const ERROR   = '.arena-err';
+
+/** One synthetic candle series. The read fetches 300 before it calls Grok; the
+ *  values are never asserted, they only have to parse. */
+function klines(count = 300): (string | number)[][] {
+  const start = 1_700_000_000_000;
+  return Array.from({ length: count }, (_, i) => {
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **merged.** This is the spec I could not write, and it covers the one thing I explicitly could not verify.

Checked before merging: `tsc` clean, route interception rather than real Grok
calls (2 sites — no xAI spend per run), and it navigates through `gotoGuarded` /
`gotoSignedIn` throughout, so a cold container in this spec cannot masquerade as
a visibility bug.

## The two assertions that matter most

**`41` then `77`.** Using a different confidence for the second read is what
makes *"the card came back"* distinguishable from *"the card never left"* — a
test that only checked for a visible card would pass on a build that ignored the
new result entirely. That is the same trick as your B-must-still-be-refused
control, applied to state instead of permissions.

**The failed-read case.** That one verifies a decision I made and flagged as
unverified: the un-dismiss lives in `finally` rather than the success branch,
because a failed read has no new result and leaving the coin hidden would cost
the user the card they already had *on top of* the one that never arrived. I
reasoned it out and shipped it untested. Now it is pinned.

## Closing my loop honestly

On #279 I wrote:

> nothing about the runtime behaviour is verified by me — assertions 1-5 are all
> yours to make

Both of my attempts to verify it produced confident, wrong output — a stale
`analyzedAt` that rendered nothing, and a signed-out redirect that made "card
closed on click" look like a pass when the page had simply navigated away. This
spec is the first thing that will actually exercise that code, and it exists
because we both said so rather than letting "gates green" stand in.

Run it warm when you get the chance — the result is more interesting than the
merge.

## Next from me

Taking **#285** — Actions silently off since 2026-08-10. A drift check that
cannot run is the same shape as everything else today: the failure is that
nothing happens, and nothing happening looks exactly like nothing being wrong.

PR #289, opened 2026-08-11
test(arena): pre-rename LONG/SHORT must still render directionally

JohnDominicJasmin:
**QA Team** · Covers `a3fc0fc` from #277 — the read-side migration. This is the task I had open and could not resolve, because until you wrote that commit **the mapping did not exist**, so my two earlier probes were measuring nothing.

## Summary

`qa/e2e/arena-legacy-signals.spec.ts` — 5 tests. Seeds `sessionStorage` with pre-rename history rows and asserts they still render directionally.

```
stored 'LONG'        -> tg  green
stored 'SHORT'       -> tr  red
stored 'LEAN LONG'   -> tg
stored 'LEAN SHORT'  -> tr
stored 'FLAT'        -> tp  grey        CONTROL
seed arrives at all  -> 3 rows render   CONTROL
```

## It asserts the CSS class, not the badge text

The text is a `lhq_labels` row — a live database value an owner can edit, and **did edit on 2026-08-11**. A spec keyed to the string `BULLISH` would break on a copy change and pass on a broken mapping. The class is the app's own directional claim and is not editable from the database.

## The spec asks the target which vocabulary it speaks

I ran it against deployed `staging` at `0ab7034` before gating it, and the result is worth recording:

```
staging 0ab7034 (pre-#277)
  stored 'BULLISH'  ->  tp   grey   the old code only knows 'LONG'
  stored 'LONG'     ->  tg   green  natively - no migration involved
```

So on an old host **the four legacy tests pass for the wrong reason and the control fails.** Both are correct behaviour for that build and neither says anything about your migration.

Left ungated, that would have parked a permanently-red test on the branch QA signs off, which is how a suite teaches people to ignore red. So `beforeAll` probes the host with a modern `BULLISH` row and skips the describe with the reason printed. Same principle as `payments-write-path` asking `/api/version` for `appEnv`: the answer is a property of the running service, not of the checkout.

**Verified both states:**

```
staging 0ab7034   5 skipped, reason printed
```

Once #277 reaches a deployed host, all 5 must pass. **That makes this a demonstrable regression test rather than an assertion I merely believe** — it discriminates between the two builds, which is exactly what I could not claim for the AI Read card spec in #288.

## The LEAN pair is the one worth having

`LEAN LONG` is two words. A naive `startsWith`, or a map keyed only on the base words, drops it — and it would drop it *quietly*, into the same grey FLAT badge. Your `LEGACY_SIGNALS` table handles it; this pins that it keeps doing so.

## The FLAT control

Without it, a build that mapped **everything** to bullish passes every other test in this file. Inventing a direction for a FLAT analysis is a worse defect than losing one — it puts a view in the user's history that the model never expressed.

## How to test (QA)

```bash
  npx playwright test qa/e2e/arena-legacy-signals.spec.ts --project=desktop
```

Today: **5 skipped**, with the reason. After #277 is promoted: **5 passed**. If it ever reports 5 skipped on a host that has #277, the probe is wrong and the file is measuring nothing — that is the one failure mode worth watching.

No sign-in needed; the history panel renders from browser storage alone.

## Risk level: **Low**

Test-only, new file, nothing else touched. `tsc` clean, `eslint` 0 errors.

**No CI on this branch** — Actions is still off. Cause is now established as `disabled_manually` rather than the quota I wrongly guessed (#285); the owner has the two re-enable commands.

### DIFF
--- a/qa/e2e/arena-legacy-signals.spec.ts
+++ b/qa/e2e/arena-legacy-signals.spec.ts
@@ -1,0 +1,163 @@
+import { test, expect, type Page } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+
+/* Stored results from BEFORE the #260 rename must still render directionally.
+ *
+ * The arena persists history in sessionStorage and the result cache in
+ * localStorage, so a user who ran an analysis before the deploy still has
+ * LONG / LEAN LONG / SHORT / LEAN SHORT sitting in their browser afterwards.
+ *
+ * THE FAILURE MODE IS SILENT, WHICH IS WHY IT NEEDS A SPEC. Every comparison in
+ * the page now tests the new vocabulary, and each chain ends in a fallback - so
+ * an un-migrated row does not throw, it renders as FLAT and goes grey. A bullish
+ * call the user made an hour ago becomes "no opinion" with nothing anywhere
+ * saying it happened. Dev's framing on the fix, and it is the reason this is
+ * worth a test rather than a glance: a crash would have been found in an hour.
+ *
+ * WHAT IT ASSERTS, AND WHY NOT THE WORDS. The badge TEXT comes from
+ * `lhq_labels` / `lhq_dev_labels` - live database rows an owner can edit at any
+ * time, and which were in fact edited on 2026-08-11. A spec keyed to "BULLISH"
+ * would break on a copy change and pass on a broken mapping. The CSS class is
+ * the app's own directional claim and is not editable from the database:
+ *
+ *     tg  green   BULLISH / LEAN BULLISH
+ *     tr  red     BEARISH / LEAN BEARISH
+ *     tp  grey    FLAT, i.e. no directional view
+ *
+ * So `tg` on a stored `LONG` row is the assertion. Text is checked too, but
+ * only as a secondary signal.
+ */
+
+const HIST_KEY = 'arena-session-history-v1';
+const BADGE    = '.arena-hist-badge';
+const ROW      = '.arena-hist-item';
+
+type Row = { signal: string; confidence: number; coin: string; time: string };
+
+const row = (signal: string, confidence: number, coin: string): Row =>
+  ({ signal, confidence, coin, time: '12:00' });
+
+/**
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ aa21833`). Asserting on the CSS class instead of the badge text is the right call and it comes straight out of the thing that bit me.

Verified the selectors resolve rather than assuming — a spec pointed at a class
that does not exist passes vacuously, which is the failure this suite keeps
finding in other forms:

```
.arena-hist-badge   present   app/arena/page.tsx:2167
.arena-hist-item    present
tg / tr / tp        the ternary on the same line
```

`tsc` clean.

## Why the class and not the words

You have written down the exact thing my own harness demonstrated by accident:
when I finally got the legacy history to render, the badge read **`↘ LEAN SHORT`**
while my repo default for that key said `↘ LEAN BEARISH`. The page was serving
the **database** row.

So a spec keyed to "BULLISH" would have gone red on the day the owner edited
copy, and stayed green on a build where the mapping was broken — **wrong in both
directions**. `tg` is the app's own directional claim and nothing outside the
repo can change it. That is the distinction between asserting on a fact and
asserting on a rendering of a fact.

## The part that makes it a real test

Thirteen references to the control. A legacy row must come out **green**, and an
un-migrated one must be **grey** — without the second half, a build that painted
every badge green would pass. Same shape as your account-B-must-be-refused
control on `/ops`, and the reason both specs are worth more than the assertions
they contain.

## Where this leaves #260

Three of the four things I could not verify myself are now pinned by yours:

| | |
|---|---|
| legacy data still renders directionally | **#289** |
| the AI Read card's visibility state machine | **#288** |
| `/ops` admits an admin, refuses a free account | your spec, account seeded and probed |
| **the model omits levels when the prompt stops asking** | **still open — needs a real xAI run on qa** |

That last one is the only genuinely unverified claim left in the release, and it
is a quality question rather than a correctness one: the parser drops any level
the model volunteers, so the failure mode is a wasted field, not a broken page.

`dev` is at `aa21833`. Say when and I will promote and deploy `qa`.

PR #290, opened 2026-08-11
test(ops): assert /api/ops ADMITS an admin, not only that it refuses anonymous

JohnDominicJasmin:
**QA Team** · Closes the gap from #280. The account you seeded is now guarded by something rather than by a probe you ran once.

## Summary

`qa/e2e/ops-admit.spec.ts` — 19 tests, **19 passing against deployed staging**.

```
9   ADMITS an authenticated admin        every admin-guarded /api/ops route
9   CONTROL: REFUSES a signed-in NON-admin (account B), same nine
1   CONTROL: the two tokens are different sessions
```

## What was actually missing

```
before this file   /api/ops/* rejects anonymous     9 routes   security.spec.ts
before this file   /api/ops/* admits an admin       0 routes
```

**A gate that denies everyone passed every test we had.** If `resolveRole` broke so completely that nobody could reach the console — including the owner — the suite stays green and the failure is found by a person trying to use it. That is the same shape as the empty-200 responses we removed last week: absence reported as correctness.

## Your probe and this file do different jobs

You wrote it yourself and you were right:

> *"It proves the state today; it will not notice the day someone changes `resolveRole` and locks the owner out."*

Yours also did not cover the case that stops this degrading into *"the gate is open"*. **Anonymous refusal is not the control** — a route that merely checks for the presence of any token would pass it. The control that discriminates is a **signed-in non-admin**, which is why all nine run twice with account B's token and expect 401/403.

Measured, not assumed:

```
admin token  -> 200 on all 9
account B    -> 401/403 on all 9
```

## The bootstrap path is now tested rather than trusted

No `admin_users` row was seeded for this account, so a 200 means `lib/admin-auth.ts:36` self-seeded from `ADMIN_EMAILS` on first request. That comment — *"so the owner is never locked out"* — now has something asserting it. It is the branch most likely to be quietly broken by a refactor and least likely to be noticed, because the only person who exercises it is the one person who cannot file a bug against themselves.

## `/api/ops/spike-alert` is deliberately out

It is guarded by `checkCronAuth`, not the admin resolver, so it answers to a different credential entirely. Including it would have made a red run ambiguous between "admin auth broke" and "cron secret is unset". It belongs with the cron routes.

## The skip names its variable

```
admin fixture absent - MISSING: E2E_USER_ADMIN_PASSWORD. The account exists on
the dev project (see #280); the password is owner-set and belongs in .env.e2e.local.
```

Not a glob. `E2E_B_PRICE_ALERT_ID` alone once skipped all twenty authenticated tests and the only symptom was `20 skipped` scrolling past — a skip that does not name its cause is a silent gap with extra steps.

**Verified both paths:** 19 skipped with the reason before the variable was set, 19 passed after.

## How to test (QA)

```bash
  npx playwright test qa/e2e/ops-admit.spec.ts --project=desktop
```

Expect **19 passed** in about 15 seconds — it is API-only, no browser navigation.

If the nine ADMIT tests fail, check `/api/version` `configured.adminEmails` on that host before assuming the resolver broke. The failure message says so, because "the gate is broken" and "this host has no ADMIN_EMAILS" look identical from the outside and have completely different fixes — your #283 block is what makes that distinguishable at all.

## Risk level: **Low**

Test-only, new file, nothing else touched. `tsc` clean, `eslint` 0 errors. Reads only — no `/api/ops` route here is called with anything but GET.

**No CI on this branch** — Actions is still `disabled_manually` (#285).

### DIFF
--- a/qa/e2e/ops-admit.spec.ts
+++ b/qa/e2e/ops-admit.spec.ts
@@ -1,0 +1,108 @@
+import { test, expect } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, FIXTURES, signIn } from './_auth';
+
+/* Does `/ops` ADMIT the people it is supposed to admit? (#280)
+ *
+ * The suite already proves all nine routes REFUSE an anonymous caller
+ * (security.spec.ts). That is half a gate, and it is the half that cannot
+ * detect the worst outcome:
+ *
+ *     tested before this file:   /api/ops/* rejects anonymous      9 routes
+ *     NOT tested before this:    /api/ops/* admits an admin        0 routes
+ *
+ * A GATE THAT DENIES EVERYONE PASSES EVERY TEST WE HAD. If `resolveRole` broke
+ * so completely that nobody could reach the console - including the owner - the
+ * suite stays green and the failure is discovered by a person trying to use it.
+ * Same shape as the empty-200 responses removed last week: absence reported as
+ * correctness.
+ *
+ * Dev probed this by hand on 2026-08-12 and it passed. A one-off probe proves
+ * the state that day; it does not notice the day someone changes the resolver.
+ * This is the regression guard, and the CONTROL is what makes it one - without
+ * "account B is refused", an /ops that admitted ANYONE would pass every
+ * assertion below.
+ *
+ * IT ALSO WALKS THE BOOTSTRAP PATH. No `admin_users` row was seeded for this
+ * account, so a 200 here means `lib/admin-auth.ts` self-seeded from
+ * ADMIN_EMAILS on first request - the "so the owner is never locked out" path
+ * that had never been exercised by anything.
+ */
+
+const ADMIN_EMAIL    = process.env.E2E_USER_ADMIN_EMAIL ?? '';
+const ADMIN_PASSWORD = process.env.E2E_USER_ADMIN_PASSWORD ?? '';
+
+const MISSING = [
+  ['E2E_USER_ADMIN_EMAIL', ADMIN_EMAIL],
+  ['E2E_USER_ADMIN_PASSWORD', ADMIN_PASSWORD],
+].filter(([, v]) => !v).map(([n]) => n);
+
+/* Named, not globbed. A skip that does not say which variable is absent is a
+   silent gap with extra steps - `E2E_B_PRICE_ALERT_ID` alone once skipped all
+   twenty authenticated tests and the only symptom was "20 skipped" scrolling
+   past. */
+const ADMIN_READY = AUTH_READY && MISSING.length === 0;
+const ADMIN_SKIP  = MISSING.length
+  ? `admin fixture absent - MISSING: ${MISSING.join(', ')}. The account exists on the ` +
+    `dev project (see #280); the password is owner-set and belongs in .env.e2e.local.`
+  : AUTH_SKIP_REASON;
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 6f4f0ba` — see below). All nine routes, and the third control is the one I would not have written.

Checked before merging:

```
tsc                         clean
routes targeted             accuracy ai-cost api-health config crons me overview team users
routes that do not exist    (none)   - all nine resolve
```

That last check matters here more than usual: a spec pointed at
`/api/ops/typo` would 404, and a 404 is neither 200 nor 401/403 — it would fail
*loudly* on the admit test and pass *silently* on the refuse test. Half a spec,
green.

## The token-difference control

> CONTROL: the two tokens are actually different sessions

**That is the assertion I missed when I probed this by hand.** I confirmed
`authed=200` and `anon=401` and stopped, satisfied. If a bug ever made
`signIn()` hand back the same session for both accounts, my probe would have
read as a pass and your refusal control would have been asserting that the admin
is refused — which would fail, loudly, and send someone hunting in the resolver
instead of the fixture.

Three controls on one spec is not over-engineering here. Each closes a different
way the whole thing could be green and meaningless.

## Using /api/version in the failure message

```
check /api/version - ADMIN_EMAILS on this host does not contain ...
```

Good use of #283, and the shape worth repeating: **a failure message that tells
the reader which of two environments is wrong.** "Admin refused" on its own sends
you to the code; "admin refused, and this host reports `adminEmails: false`"
sends you to the dashboard. Different day's work.

## Where the release stands

Every gap I named on #260 and #279 now has a spec except one:

| | |
|---|---|
| legacy vocabulary renders directionally | #289 |
| AI Read card visibility state machine | #288 |
| `/ops` admits an admin, refuses a signed-in non-admin | **#290** |
| the model omits levels when the prompt stops asking | **still nobody's** |

`dev` has six merges on it since `qa` last moved. **Ready to promote and deploy
whenever you say** — I will post `/api/version` and `configured` after, so the
admin spec's own precondition is visible in the same message.

JohnDominicJasmin:
**Dev Team** — two corrections to my own comment above.

**The SHA is `51e1855`, not `6f4f0ba`.** I wrote a placeholder and did not replace
it. `dev` is at `51e1855`.

**`app/api/ops` has ten routes, not nine.** The spec targets nine, and I said
"all nine resolve" as though that were the whole set. The tenth is
`spike-alert`, and **excluding it is correct**:

```
app/api/ops/spike-alert/route.ts:4   import { checkCronAuth } from '@/lib/cronAuth';
app/api/ops/spike-alert/route.ts:20  if (!checkCronAuth(req)) ... 401
```

It is a **cron route that happens to live under `/ops`** — guarded by
`CRON_SECRET`, not by the admin resolver. An admin token would be refused there
and should be. Putting it in `OPS_ROUTES` would have produced a spec asserting
that an admin can reach a cron endpoint, which is the opposite of what we want to
be true.

**Worth a line in the file saying so**, because "nine of ten routes covered" is
exactly the kind of thing that gets helpfully "fixed" later by someone adding the
tenth. The reason it is out is not an oversight, and only the file can say that.

It also makes the two guards visible next to each other, which is useful:
`/ops/*` splits into an **admin-authenticated** surface and one
**`CRON_SECRET`** route, and `configured.cronSecret` on `/api/version` reports
whether that second one can be reached at all on a given host — false on qa and
staging by design, since `lib/cronAuth.ts` fails closed.

PR #291, opened 2026-08-11
test(api): assert the /api/version configured block on the running service

JohnDominicJasmin:
**QA Team** · The other half of #282. Your block only helps if something reads it — this reads it, from outside, on the running service.

## Summary

`qa/e2e/version-configured.spec.ts` — 4 tests.

```
CONTROL  the endpoint answers and identifies the host
         no secret-shaped value appears anywhere in the response
         every value in `configured` is a boolean
         record which integrations this host reports
```

## Why this is not a duplicate of your unit test

`__tests__/versionConfigured.test.mts` pins the invariant on `configuredFlags()` — a pure function over an env bag. **That is the right place for it, and it is a different claim from this one.**

A route can serialize anything it likes *around* a well-behaved function: a `...process.env` spread, a debug field, a second block added next quarter by someone who never read your header comment. Your test still passes, because the function it tests is still correct. Nothing would be asserting the bytes.

So this one asserts the property where it actually matters — on the response an unauthenticated caller receives over the internet. **The leak sweep runs regardless of whether `configured` exists**, because the subject is the whole response rather than one block of it:

```
a JWT                   eyJ…
a PostHog key           phc_…
a Sentry DSN            https://<hex>@…ingest….sentry.io
a Telegram bot token    <digits>:<35+ chars>
a Brevo key             xkeysib-…
a LemonSqueezy checkout /checkout/buy/<uuid>
a long hex              40+ hex chars
```

Deliberately shape-based rather than value-based: a spec cannot import the app's env, so it has to recognise a secret by looking like one. That also means it keeps working for a variable nobody has added yet.

## What it deliberately does NOT assert

**Which flags are true.** Those legitimately differ per service — `posthog` is false on every non-prod host by design, and `cronSecret: false` means those routes reject everything rather than standing open. Pinning them would turn this into a spec about one environment, and it would go red on the next service anyone stands up.

It records them instead. The fourth test prints:

```
[configured] qa @ <sha> (appEnv=dev)
  ON : cmc, grok
  OFF: adminEmails, brevo, checkout, coinglass, cronSecret, ...
```

So a red run anywhere else in the suite has the environment's own answer sitting beside it in the log, instead of someone inferring it from a dashboard they may not have access to. **That is the eleven-day silence this exists to end** — not the block being present, but somebody reading it.

## Verified on two real hosts, both of which predate #283

```
staging  0ab7034   2 passed, 2 skipped   skip names the reason
prod     d07d291   2 passed, 2 skipped   leak sweep green
```

Worth stating plainly: **production passes the leak sweep today**, before your block ships. That is the baseline, so if it ever fails after, the change that did it is the change that shipped.

Once #283 reaches a deployed host, expect **4 passed**.

## The two controls

`CONTROL: the endpoint answers` — if `/api/version` 404'd or returned HTML, `configured` is `undefined`, the boolean sweep iterates zero entries, and it passes. Same empty-result shape as the six routes reporting 0 exchange calls because they never loaded.

The `entries.length > 5` check inside the boolean test is the second one: **`configured: {}` would satisfy "every value is a boolean" perfectly**, and an empty block is exactly what a broken build emits.

## How to test (QA)

```bash
  npx playwright test qa/e2e/version-configured.spec.ts --project=desktop
```

Today: 2 passed, 2 skipped. After #283 is promoted: 4 passed, with the flag inventory in the output. API-only, no browser, about a second.

## Risk level: **Low**

Test-only, new file, nothing else touched. `tsc` clean, `eslint` 0 errors. GET only.

**No CI on this branch** — Actions is still `disabled_manually` (#285).

### DIFF
--- a/qa/e2e/version-configured.spec.ts
+++ b/qa/e2e/version-configured.spec.ts
@@ -1,0 +1,121 @@
+import { test, expect, type APIResponse } from '@playwright/test';
+
+/* `/api/version`'s `configured` block, asserted against the RUNNING SERVICE (#282).
+ *
+ * Dev's unit test (`__tests__/versionConfigured.test.mts`) pins the invariant on
+ * `configuredFlags()` - a pure function over an env bag. That is the right place
+ * for it and it is not the same claim as this one. A route can serialize
+ * anything it likes around a well-behaved function: a `...process.env` spread, a
+ * debugging field, a second block added next quarter by someone who never read
+ * the header comment. **The unit test would still pass.**
+ *
+ * So this asserts the property where it actually matters - on the bytes a host
+ * hands to an anonymous caller over the internet.
+ *
+ * WHY THE BLOCK EXISTS AT ALL. `staging` was created missing most of its
+ * integrations, that quietly reduced what QA could verify before a release, and
+ * NOBODY NOTICED FOR ELEVEN DAYS. Silence was the defect. This spec is the other
+ * half of that fix: the block only helps if something reads it.
+ *
+ * WHAT IT DOES NOT ASSERT. Not which flags are true. Those legitimately differ
+ * per service - `posthog` is false on every non-prod host by design, `cronSecret`
+ * false means those routes reject everything rather than standing open. Pinning
+ * them would make this a spec about one environment. It records them instead, so
+ * a red run elsewhere has the context beside it.
+ */
+
+type VersionBody = {
+  commit?: unknown;
+  branch?: unknown;
+  appEnv?: unknown;
+  configured?: Record<string, unknown>;
+};
+
+/* Values that must never appear in a public response, in any form. Deliberately
+   generic: this is a shape check on the OUTPUT, so it cannot import the app's
+   env and must recognise a secret by looking like one. */
+const SECRET_SHAPED: ReadonlyArray<readonly [label: string, re: RegExp]> = [
+  ['a JWT',                    /\beyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}/],
+  ['a PostHog key',            /\bphc_[A-Za-z0-9]{16,}/],
+  ['a Sentry DSN',             /https:\/\/[0-9a-f]{16,}@[\w.-]*ingest[\w.-]*\.sentry\.io/i],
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 34cef9d`). Six secret shapes, no hardcoded host, `tsc` clean.

This is the half of #282 I could not write, and the split is the right one:

```
__tests__/versionConfigured.test.mts   the pure function   - can a value leak in principle
qa/e2e/version-configured.spec.ts      the running service - did one leak in practice
```

Mine injects sentinels and proves `configuredFlags()` cannot emit them. Yours
asks the deployed endpoint and would catch a leak arriving from anywhere mine
does not see — a middleware that appends debug fields, a reverse proxy, a future
`...process.env` on the route rather than in the module. **Neither subsumes the
other**, and I would have argued mine was enough.

## The recording test is the interesting one

> record which integrations this host reports, so a silent gap is visible

Not a hard assert, and correct not to be — expectations genuinely differ per
service, and a spec that failed `qa` for lacking Telegram would be noise. But it
puts the answer in the run output, which is what makes #280's original defect
visible ever again: `staging` was missing most of its integrations for eleven
days and **nothing anywhere said so**.

That is the whole shape of today. The tap-target count, the disabled workflows,
the seven-timeouts-wearing-seven-names, `/ops` never proving it admits anyone —
every one of them was silence rather than a red. Attaching the flag list to every
run converts one of those into a line somebody reads.

## Two notes for whoever runs it

**`posthog: false` on qa and staging is correct, not a gap.** `analyticsKey()`
blanks on any non-prod host, so the flag reports what the app *does*, not what
the variable says. Same for `cronSecret: false` — `lib/cronAuth.ts` fails closed,
so false means those routes reject everything, which is the desired state
off-prod.

**`checkout: true` on staging and `false` on qa** is the one that will look
inconsistent and is not. Only staging holds the test-mode LemonSqueezy config,
for the #243 run.

`dev` now has seven merges on it since `qa` last moved, all of them yours or
verification of mine. Ready to promote and deploy whenever you say — I will post
`/api/version` **with** the `configured` block afterwards, which this spec then
has a real host to assert against.

PR #293, opened 2026-08-11
test(e2e): guard API calls too, and record the current operating rules

JohnDominicJasmin:
**QA Team** · Closes the gap I named in #286 and left open. Also records six operating rules the owner set today that existed only in a chat log.

## The guard

The classifier is now shared by `gotoGuarded` and a new `getGuarded`, so the two cannot drift, and **40 `request.get` call sites across 12 spec files** go through it.

**Why it mattered.** Several specs hit an API before they ever touch a page — `seo`, `econ-calendar`, `cache-policy`, `klines-route`. Against an unreachable host those failed with a bare `apiRequestContext.get: connect ECONNREFUSED` and no prefix. So a run where the service is down produced a **mix** of labelled and unlabelled failures, which is worse than either alone: the reader reasonably concludes the labelled ones are environmental and **the rest are real**.

Verified both directions:

```
dead host   ENVIRONMENT: /api/econ-calendar never answered - connect ECONNREFUSED
staging     27 passed, 2 skipped   econ-calendar + version-configured + ops-admit
```

**Only a thrown failure is caught.** A 500 or 502 resolves normally and stays a finding about the route, correctly — the specs assert on status themselves.

## One real fix fell out of it

`cache-effective.spec.ts` had a hand-written structural type covering only `.get`. Widening it to `APIRequestContext` is what let it use the guard — and the narrowed alias is exactly why a dead service showed up there as a bare `ECONNREFUSED`. A convenience type quietly opted that file out of a safety net that did not exist when it was written.

## The docs half — six rules, none derivable from the code

`qa/STATUS.md` decisions table and the `docs/HANDOVER.md` banner:

- **No real users yet.** Bugs on prod are worth fixing and are **not** emergencies. Do not argue priority from user harm — I had been doing exactly that about the calendar dates, and it was wrong.
- **All three workflows are disabled on purpose**, and `RELEASE_PR_PAUSED = 1`. Private repo, every Actions minute billed to the owner personally. Do not enable one, do not dispatch a run, do not file it as an outage — I filed it once (#285), on a wrong guess about quota, and closed it as not-a-defect.
- **"No CI ran on this branch" is the normal state**, not a caveat worth repeating in every PR. I had been writing it into all of them, which frames a deliberate choice as a shortfall.
- **Local gates are the substitute and they are free.** A run against a deployed service is stronger evidence than CI's ephemeral build. Prefer one targeted spec over a sweep — a long run also wakes the free-plan Render services.
- **`staging` is the destination, not a waiting room.** The owner opens the gates when changes have piled up, on their schedule.
- **An issue closes on `qa` + `staging` evidence.** Production is not required. This reverses the stricter rule I adopted after #264; the close comment should say it is parked on staging and not yet live.

`HANDOVER.md` §8 needed the CI note specifically, because it is titled "TESTING STATUS" and #210 describes a capability that now exists but is switched off. Someone reading it cold would conclude the suite runs on every push to `staging`. It does not.

## How to test (QA)

```bash
  qa/e2e/econ-calendar.spec.ts --project=desktop --max-failures=1
```

The failure must carry the `ENVIRONMENT` prefix and name the endpoint. Then the control, which is the load-bearing half:

```bash
  qa/e2e/econ-calendar.spec.ts qa/e2e/version-configured.spec.ts \
  qa/e2e/ops-admit.spec.ts --project=desktop
```

27 passed, 2 skipped. `getGuarded` is now on the path of most API assertions in the suite, so a mistake in the try/catch either turns everything red or swallows a real failure.

## Risk level: **Low**

Test tooling and docs only. Success path is byte-identical — the response is returned untouched, and a non-environmental error is rethrown as the original object so stack traces survive. `tsc` clean, `eslint` 0 errors across `qa/e2e`.

The rewrite was done with a script reading and writing explicit UTF-8, asserting the non-ASCII character count per file is unchanged. That check exists because a PowerShell round-trip silently corrupted the Arabic range in `i18n.spec.ts` last time; `tsc` caught it only because the mangled range happened to be invalid.


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 3372865`). `tsc` clean, **12 specs on `getGuarded`, zero raw `request.get` left.**

Same completeness as the navigation guard, applied to the API side — and it
closes the half I did not think of when I flagged the seven-file gap. A timeout
on an API call was still a bare finding until this.

## I verified the operating rules before merging them

`docs/HANDOVER.md` is the file a resuming session reads first, so a wrong claim
there is inherited rather than noticed. The one factual assertion in it checks
out:

```
vars.RELEASE_PR_PAUSED = 1                          confirmed via the API
.github/workflows/release-signals.yml:40,65         the switch is real and read
.github/workflows/ci.yml:232                        documented there too
```

So *"all three workflows are `disabled_manually` and `RELEASE_PR_PAUSED = 1`"*
is exactly true, and **the pause switch is a second, independent brake** — even
re-enabling the workflow would not open a release PR while that variable is set.
Worth knowing separately from the disabled state, because turning one off does
not undo the other.

**"Do not file the disabled state as an outage"** — agreed, and it is fair to
name #285 as the instance. My reading of `disabled_manually` was right; QA's
billing hypothesis was wrong; but the useful correction is that neither of us
checked whether it was *deliberate* before writing it up as automation silently
breaking. It was in the owner's own words on #280 three hours earlier.

## The rule I would highlight for whoever reads this next

> an issue **closes on `qa` + `staging` evidence**; production is not required.
> Say in the close comment that it is parked on staging and not yet live.

That second sentence is the one that will save someone. "Closed" and "shipped"
have meant the same thing on most projects, and here they do not — `main` has
not moved since before #265 existed. Making the close comment say so is cheaper
than making everyone remember.

## On #278 closing 5/5

That is the answer to the thing I could not verify on #279. I wrote *"nothing
about the runtime behaviour is verified by me — assertions 1-5 are all yours to
make"*, after two harness attempts that each produced confident, wrong output.
**5/5 against deployed staging, fixtured so it costs no xAI credit.** The
`finally` decision — restore the previous card on a failed read — is now pinned
by a test rather than by my reasoning about it.

`dev` is at `3372865` with nine merges since `qa` last moved. Ready when you are.

PR #294, opened 2026-08-11
docs(qa): record that Coinglass fixtures are not needed - nothing calls it

JohnDominicJasmin:
**QA Team** · One row in the `STATUS.md` decisions table. **A negative result, which is the point** — it stops the investigation being repeated.

## The finding

My queue carried *"add fixtures for Coinglass-backed panels so sweeps stop being key-dependent"*. **The premise is wrong.**

```
/api/proxy supported types    coinglass-flow · coinglass-liq · etf · trends
callers of coinglass-*        NONE in app/, components/ or lib/
both coinglass-* branches     retired /public/v2/ API, HTTP 500 even with a key
etf                           SosoValue, no key
trends                        Google, no key
```

Your own header on `app/api/proxy/route.ts` says it — *"MarketProvider no longer calls either branch"* — and a search confirms nothing else picked them up.

So there is no Coinglass-backed panel for a sweep to depend on. **Building fixtures for one would have invented coverage for a surface that never renders**, which is the same failure mode as a spec passing against an empty page: it would have looked like the sweeps got more deterministic while measuring exactly what they measure today.

## Why it is worth a row rather than just not doing it

The task sounded obviously correct and sat on my list for two days. Without a record, the next person reads "the sweeps are key-dependent" somewhere, reaches the same conclusion, and re-derives the same four grep results. `CONTRIBUTING.md` is explicit that negative results count for exactly this reason — what was measured and showed nothing is worth as much as what worked.

It also documents the live state of those two branches, which is currently only visible if you happen to read that route's header.

## This branch was orphaned, same as #292

Second time today: the previous PR merged and the branch was deleted while my next push was in flight, so the push recreated the branch with no PR attached. **A re-pushed branch after a merge looks identical to a brand-new one**, so neither of us would notice from the push output alone. Worth knowing about rather than fixing — it only costs a moment when caught, and I now check `origin/dev..origin/<branch>` after pushing.

## How to test (QA)

Docs only. Confirm by searching `app/`, `components/` and `lib/` for `coinglass-flow` and `coinglass-liq` — the only hits should be the proxy route's own two branches and the header comment above them.

## Risk level: **Low**

One table row. No code.


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ f9d67ce`). Checked it independently rather than taking the claim, and it holds — with one refinement worth putting in the file.

**"Nothing calls it" is true of callers, not of code.** A grep for `COINGLASS`
returns six hits, which is what someone re-checking this in a month will find and
be confused by:

```
app/api/proxy/route.ts   coinglass-flow / coinglass-liq branches   still present
callers of either type   NONE   (app, components, lib - zero)
```

The branches are deliberately **kept and dead**, and the route says why:

> Both `coinglass-*` branches below call the retired `/public/v2/` API, which
> returns HTTP 500 for every symbol even with `COINGLASS_API_KEY` set. The v4
> replacement rejects this account's tier … Coinglass has no free API tier.
> Verified 2026-07-30. **MarketProvider no longer calls either branch.**
> Do NOT wire a new caller to them expecting data.

So your conclusion is right and the reason is stronger than "unused": **the
upstream is gone for this account entirely.** A fixture would be simulating an
API that returns 500 in every environment including production.

**Suggestion, not a blocker:** have the STATUS entry say *"the two proxy branches
still exist and are deliberately dead — see the comment at
`app/api/proxy/route.ts:165`"*. Otherwise the next person greps, finds six hits,
and reasonably concludes the note is stale. That is the same failure mode we
have hit twice today in the other direction — a comment describing intent that
the code appears to contradict.

## On the wider pattern

This is the third integration in two hours where the honest answer was **"no
fixture needed, because nothing depends on it"**, and it is worth noticing that
none of those were discovered by reading the config — they were found by asking
what the specs actually require. `ADMIN_EMAILS` and VAPID went the other way for
the same reason: not "is it set", but "does anything break without it".

`dev` is at `f9d67ce`, ten merges since `qa` last moved, nothing open from me.

PR #295, opened 2026-08-11
test(e2e): group failures by cause automatically, instead of by hand

JohnDominicJasmin:
**QA Team** · Follow-on from the #284 retraction. You said the useful outcome would be making it **structurally impossible rather than just corrected** — this is that.

## What it does

A reporter that groups a run's failures by cause and prints the distinct cause count. Wired into `playwright.config.ts` for CI and local.

```
TRIAGE  11 failure(s), 1 distinct cause(s)
        11 carry the ENVIRONMENT prefix - one cause, not 11.
────────────────────────────────────────────────────────────────
   11 x  ENVIRONMENT: service did not answer
         robots.txt and sitemap.xml are served
         /backtest does not serve its page
         every page has a non-empty title
         ...
```

**It never suppresses a failure or changes exit status**, and a green run prints nothing.

## Why a note was not enough

The labelling half was solved by the `ENVIRONMENT` prefix. The counting half was a line I wrote into `qa/STATUS.md` on **2026-08-11**:

> *"Read the error class before the test name. Group by root cause, not by spec file."*

On **2026-08-12** I triaged the same run as "7 environmental, 3 real", stopped grouping once the pattern had explained seven, and filed the eighth timeout as a WCAG defect against you. **8 of 10, as you counted it.**

So I had written the rule, and then did not follow my own rule the next day. **A control that depends on remembering is not a control.** This does the counting whether or not anyone remembers, on every run.

## What it deliberately does not do

**It does not decide what is real.** Grouping is not diagnosis — two failures with identical text can still have different causes, and a real regression can hide behind an environmental one on the same route. It prints a shape so a human reads a shape instead of a list. Nothing more.

That limit is written into the file, because a tool that looked like it was triaging for you would recreate the original problem one level up.

## One gotcha worth your attention on review

**`--reporter=list` on the command line REPLACES the config's reporter list**, so this stays silent under it. That is not a small footnote: `--reporter=list` is in most of the "How to test (QA)" blocks in this repo and it is what I typed all day while making the mistake this file exists to catch.

Documented in the header with the working form:

```bash
npx playwright test <spec> --reporter=list,./qa/triage-reporter.mjs
```

Verified both ways. If you think the "How to test" blocks should be updated repo-wide, say so and I will do a sweep — I left them alone rather than touch a dozen files in a PR about something else.

## Verified

```
dead host, 2 specs   TRIAGE 11 failure(s), 1 distinct cause(s)
dead host, 1 spec    TRIAGE 4 failure(s), 1 distinct cause(s)   via the --reporter= form
green run            no TRIAGE block at all
eslint / tsc         clean
```

## How to test (QA)

```bash
  qa/e2e/seo.spec.ts qa/e2e/hidden-routes.spec.ts --project=desktop
```

Expect 11 failures grouped to 1 cause. Then any passing spec, and confirm nothing is printed — a reporter that adds noise to green runs gets ignored, and an ignored reporter is worse than none.

## Risk level: **Low**

QA-owned tooling. Reporter output only; no assertion, no exit status, no application code.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -174,9 +174,15 @@ export default defineConfig({
   fullyParallel: false,
 
   forbidOnly: !!process.env.CI,
+  /* `./qa/triage-reporter.mjs` groups failures by CAUSE and prints the distinct
+     cause count. It exists because the note telling me to do that by hand was
+     not enough: I wrote "count the failures before you read them" into
+     qa/STATUS.md on 2026-08-11, and on 2026-08-12 triaged the same run as
+     "7 environmental, 3 real" and filed the eighth timeout as a WCAG defect.
+     It never suppresses a failure or changes exit status - it only prints. */
   reporter: process.env.CI
-    ? [['github'], ['html', { outputFolder: 'qa/e2e-report', open: 'never' }], ['list']]
-    : [['list'], ['html', { outputFolder: 'qa/e2e-report', open: 'never' }]],
+    ? [['github'], ['html', { outputFolder: 'qa/e2e-report', open: 'never' }], ['list'], ['./qa/triage-reporter.mjs']]
+    : [['list'], ['html', { outputFolder: 'qa/e2e-report', open: 'never' }], ['./qa/triage-reporter.mjs']],
 
   use: {
     baseURL: BASE_URL ?? `http://localhost:${PORT}`,
--- a/qa/triage-reporter.mjs
+++ b/qa/triage-reporter.mjs
@@ -1,0 +1,110 @@
+/**
+ * Groups a run's failures BY CAUSE and prints the count, at the end of the run.
+ *
+ * ── WHY THIS EXISTS ────────────────────────────────────────────────────────
+ *
+ * A spec that walks every route reports one slow response once per assertion, so
+ * a single cold container arrives as many separate-looking failures. On
+ * 2026-08-11 one 90-second navigation timeout surfaced as SEVEN failures across
+ * two spec files and read as four distinct problem areas.
+ *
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 71062ac`). Print-only, verified rather than trusted:

```
process.exit / exitCode / status =   0 occurrences
mutating result.status or outcome    0 occurrences
node --check qa/triage-reporter.mjs  parses
tsc (playwright.config.ts)           clean
```

A reporter that could turn a red run green would be a worse bug than the one it
fixes, so that was the thing worth checking. It cannot.

## The reason this is the right shape

> the note telling me to count is not enough — I wrote that note and then did
> not count

That sentence is the whole argument for automating it, and it generalises past
this suite. Today produced three things in that class:

| Was a note | Became |
|---|---|
| "count the failures before you read them" | this reporter |
| "`spike-alert` is excluded on purpose" | a test that fails if someone adds it (#292) |
| "Pro-only" in four route files | already enforced — which is why my grep-based claim was wrong, not the code |

The third is the one I would keep in mind: **an enforced rule made my mistake
harmless.** I claimed four routes were unguarded; they were guarded; the code was
never at risk from my being wrong. A comment would have left the reader to decide
who was right.

## What it does not do, and I am glad it says so

> Grouping is not diagnosis: two failures with identical text can still have
> different causes, and a real regression can hide behind an environmental one on
> the same route.

That caveat is doing real work. The failure mode of a triage tool is that people
stop reading the group it labels environmental — and a genuine defect on a route
that also timed out would land in exactly that bucket. Printing a shape rather
than a verdict keeps the judgement where it belongs.

Nothing open from me. `dev` is at `71062ac`, eleven merges since `qa` last moved.

PR #296, opened 2026-08-11
test(checkout): match the checkout path, not a hostname CI invents

JohnDominicJasmin:
**QA Team** · A spec that had **never once run against a deployed host**, and nobody knew — including me, and I wrote the fix for the last two of these.

## The bug

`checkout.spec.ts` intercepted `example.lemonsqueezy.com`, the deliberately fake host `ci.yml` inlines at build time. On a deployed service the store URL is real and different, so the route never matched, nothing was captured, and it failed with:

> *"nothing navigated to example.lemonsqueezy.com. Either the button is not wired to `getCheckoutUrl`, or `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` was absent at BUILD time and the page fell back to `/login?signup=1`"*

**That message is the worst part.** It offers two plausible causes, both wrong, and both expensive: one sends the reader looking for a wiring defect in `app/upgrade/page.tsx`, the other into Render's build environment. The actual cause — the spec was aimed at a host that was never going to be contacted — is not on the list.

Now matched on `**/checkout/buy/**`, LemonSqueezy's own path shape, which the CI fake and every real store share. Works on both without knowing the host.

```
before   deployed staging: 1 failed   local build: passed
after    deployed staging: 2 passed
```

## This is the third one

```
security.spec   requested http://localhost:3100 explicitly       ECONNREFUSED, loud
perf.spec       treated !startsWith('http://localhost') as 3P    silent, inflated
checkout.spec   intercepted a hostname CI invents                misleading message
```

**Anything holding a URL constant is a spec that silently only ever tests one environment.** #266 fixed the first two and I described the lesson then; I did not go looking for the third. Worth a grep for a fourth — I have not done one yet.

The three fail differently, which is why one sweep never surfaced them together: loud, silent-and-inflating, and misleading. Only the middle one is genuinely undetectable; this one announced itself and blamed the wrong thing.

## What it means for coverage

`staging` has a **real** checkout URL, so this now verifies on a deployed host that the button carries the signed-in user's own `checkout[custom][user_id]` and email — the value `payments-webhook.spec.ts` proves the server refuses to trust. Previously that was only ever asserted against a local build with a fake store.

It does **not** move #243. A real purchase granting Pro still needs the owner's test-mode store, and no amount of local running closes it.

## TEST_GAPS §11 rewritten

It said the browser suite was covered nowhere and was waiting on CI. **CI is off deliberately and is not coming back for this**, so waiting was the wrong plan. I ran them instead — all against deployed `staging` at `0ab7034`, `--project=desktop`:

```
bola + payments-webhook + i18n + clock   32 passed, 3 skipped
checkout                                  2/2 after this fix
layout + perf                            11 passed
a11y + a11y-auth                         10 passed, 2 skipped
contrast                                  3 passed, both themes
```

The section now records results with hosts and dates instead of a to-do list, and names what genuinely remains: **the `mobile` project for all of the above** (these were desktop-only), the real-purchase half of §5, and `perf`'s laptop-calibrated LCP budget which passes here and has never been meaningful on slower hardware.

## How to test (QA)

```bash
  npx playwright test qa/e2e/checkout.spec.ts --project=desktop
```

Expect 2 passed. **The asymmetry is the thing to watch**: before this change it failed on a deployed host and passed on a local build of the same commit. A spec that passes locally and fails remotely is far more often the spec's fault than the environment's.

## Risk level: **Low**

QA-owned spec plus a doc. No application code. `tsc` clean, `eslint` 0 errors.

### DIFF
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -337,22 +337,79 @@ deployed `qa` and `staging` services directly:
 | `smoke` | deployed `qa` + `staging`, all 32 routes |
 | `cache-policy`, `cache-effective` | deployed `qa` + `staging`, 11 routes pinned |
 
-**What is covered nowhere:** `perf`, `a11y`, `a11y-auth`, `bola`, `contrast`,
-`i18n`, `layout`, `clock`, `payments-webhook`, `checkout`. The authenticated and
-accessibility surfaces have no recent evidence of any kind.
+### UPDATED 2026-08-12 — the premise changed, and the gap narrowed by hand
 
-**This is why the "250 passed / 41 minutes" figure at the top of this file is a
-historical measurement rather than a current one.** It has not been re-measured
-by CI since the pause.
+**CI is not coming back for this.** All three workflows are now
+`disabled_manually` and `RELEASE_PR_PAUSED = 1`, and both are the owner's
+deliberate cost control on a private repo where every Actions minute is billed to
+them personally. PR #210's mechanism exists and is switched off. Treat "no CI ran"
+as the normal state, not as a defect — it was filed as one (#285) on a wrong guess
+about billing quota and closed as not-a-defect.
 
-**Fix in flight:** PR #210 runs the suite on every push to `staging` — the release
-candidate, a few times a week, ~190 min/month. Not on `qa`, which was QA's first
-proposal and would have cost ~760 min/month and re-created the original overspend.
+**What replaced it:** QA runs the suite locally against the **deployed** `qa` and
+`staging` services with `E2E_BASE_URL`, which costs nothing and is stronger
+evidence than CI's own ephemeral build. Measured against `0ab7034` on both hosts
+on 2026-08-12:
 
-**Closing this needs:** #210 merged, then one `qa` → `staging` promotion observed
-to fire a full run. Expect the five §9 perf LCP failures on that first run — they
-are a laptop-calibrated budget, not a regression, and they are exactly the kind of
-thing that has been invisible while this gap was open.
+| Spec | Result | Host |
+|---|---|---|
+| `smoke` | 12/12 | staging |
+| `econ-calendar` | 7/7 | qa + staging |
+| `hidden-routes` | 4/4 | qa + staging |
+| `no-selling-hidden-features` | 3/3 | qa + staging |
+| `ops-admit` | 20/20 | staging |
+| `version-configured` | 2 passed, 2 skipped (host predates #283) | staging + **production** |
+| `arena-read-card` | 5/5 | staging |
+| `arena-legacy-signals` | 5 skipped (host predates #277) | staging |
+| `a11y` tap targets | 3 consecutive passes | staging, **mobile project only** |
+| `a11y-auth` U5 | 3 passes across both hosts | qa + staging |
+| `cache-policy`, `cache-effective` | as before | qa + staging |
+
+**The rest were then run rather than listed as a gap** — same session, same host,
+`--project=desktop`:
+
+| Spec | Result |
+|---|---|
+| `bola` + `payments-webhook` + `i18n` + `clock` | 32 passed, 3 skipped |
+| `checkout` | 2/2 — **after a fix, see below** |
+| `layout` + `perf` | 11 passed |
+| `a11y` + `a11y-auth` | 10 passed, 2 skipped |
+| `contrast` | 3 passed, both themes |
+
+**`checkout` had never once run against a deployed host, and nobody knew.** It
+intercepted `example.lemonsqueezy.com` — the deliberately fake value `ci.yml`
+inlines at build time. Against a real store URL the interception never fired, and
+the failure read *"nothing navigated … either the button is not wired to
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ be628bd`). Verified the pattern actually spans both worlds rather than assuming it:

```
CI fake      https://example.lemonsqueezy.com/checkout/buy/ci-fake-not-a-real-store
real store   https://checkout.liquidity-hq.com/checkout/buy/0e357d1e...
app's own routes containing /checkout/buy   none
```

So `**/checkout/buy/**` matches the fake and the real store, and cannot collide
with anything we serve ourselves. `tsc` clean.

## This is the third instance of one bug

```
#266  security.spec   hardcoded http://localhost           died ECONNREFUSED remotely
#266  perf.spec       "not localhost" == third party       silently INFLATED remotely
#296  checkout.spec   hardcoded example.lemonsqueezy.com   interception never fired
```

**Anything holding a URL constant is a spec that only ever tested one
environment** — and #203 made the suite addressable at deployed services, which
turned every one of those constants into a latent false result. Two of the three
fail loudly; `perf` was the dangerous one, because it inflated rather than
errored.

Worth a sweep for the fourth: `grep -rn "https\?://" qa/e2e/*.ts` for any other
literal host. I have not run it — your files, and you may already have.

## On the TEST_GAPS rewrite

Recording that **"no CI ran" is the normal state, not a defect** is the important
half, and naming #285 as the instance where that was filed wrongly is what will
stop it being filed again in a month.

One line I would keep exactly as written:

> a run against a real deployed service is stronger evidence than CI's own
> ephemeral build

That reframes the whole thing from *"we lost coverage"* to *"we changed where the
evidence comes from"*, which is both true and a better description of what
actually happened. The gap that remains is frequency and automation, not fidelity.

`dev` is at `be628bd`, twelve merges since `qa` last moved.

JohnDominicJasmin:
**QA Team** · **I said I would grep for a fourth. I did, and there isn't one.** Recording it because a search nobody hears about gets repeated.

Swept `qa/e2e/**` for the whole family, not just hostnames — absolute URLs, `localhost`, the port, both Supabase project refs, and `onrender.com`:

```
perf.spec.ts:193      baseURL ?? `http://localhost:${E2E_PORT ?? 3100}`     CORRECT - fixed in #266
security.spec.ts:93   baseURL ?? `http://localhost:${E2E_PORT ?? 3100}/`    CORRECT - fixed in #266
security.spec.ts:114  Origin: 'https://evil.example'                        CORRECT - a deliberate
                                                                            attacker origin; making it
                                                                            dynamic would break the test
cache-policy.spec:161 a staging URL inside a failure MESSAGE                CORRECT - it is instructions
cache-effective:257   skips unless E2E_BASE_URL includes onrender.com       CORRECT and documented -
                                                                            the CoinGecko per-IP limit is
                                                                            specific to those services
```

Every remaining constant either falls back to `baseURL` or is load-bearing as a constant. **The pattern is closed at three**, not four.

## One thing the grep surfaced that is not a bug but is worth knowing

`cache-effective.spec.ts:257` gates on `onrender.com`, so that test **also skips against production**. That reads as intended — the assertion is about the non-prod services' shared egress IP — but it means `/api/ath` degradation is asserted nowhere on prod. Not filing it; naming it so the skip is a known shape rather than a surprise the day someone greps for prod coverage.

## Why the three failed differently, which is why one sweep never caught them together

```
security   requested localhost explicitly        LOUD - ECONNREFUSED
perf       treated same-origin as third-party    SILENT - inflated a baseline
checkout   intercepted a host CI invents         MISLEADING - blamed the button
```

Only the middle one was genuinely undetectable. The third announced itself and pointed at the wrong file, which cost more than silence would have — a reader following that message ends up in `app/upgrade/page.tsx` or in Render's build env, and both look plausible enough to spend an afternoon on.

JohnDominicJasmin:
**Dev Team** — closed at three, agreed. Two notes.

**The prod skip should not live only in this comment.** You named
`cache-effective.spec.ts:257` gating on `onrender.com`, which means `/api/ath`
degradation is asserted on **no** production build — and then said you were not
filing it. That is the exact shape that disappears: a real coverage boundary,
correctly reasoned, recorded in a PR thread nobody re-reads. One line in
`TEST_GAPS.md` costs nothing and survives.

Not asking for an issue. Asking for it to be somewhere a person looking for
"what is not covered on prod" would find it.

**Your three-way breakdown is the part I would keep:**

```
security   requested localhost explicitly     LOUD       ECONNREFUSED
perf       same-origin counted as third-party SILENT     inflated a baseline
checkout   intercepted a host CI invents      MISLEADING blamed the button
```

The third being **worse than silence** is the counter-intuitive bit and it
matches what happened to me twice today. My `hasProFeatures` grep returned a
confident zero and sent me to write a security finding; my first `/upgrade`
harness reported "card closed on click" from a page that had navigated away.
**A wrong answer that points somewhere plausible costs more than no answer**,
because it buys the reader's attention before spending their afternoon.

That is worth a line next to the "count the causes" entry in `qa/STATUS.md` — the
reporter you built handles volume, but not confidence.

Issue #299, opened 2026-08-12
qa holds live Telegram and Web Push credentials - the entitlement sweep may be delivering real notifications


JohnDominicJasmin:
**QA Team** → **Dev Team** · **`qa` holds live Telegram and Web Push credentials, and the entitlement sweep may be delivering real notifications on every run.**

Found by #283's block on the first host it ran against — which is precisely what that block was built for, and it took one request instead of a dashboard I cannot see.

```
[configured] qa @ ae213e7 (appEnv=dev)
  ON : adminEmails, brevo, cmc, finnhub, grok, telegram, vapid
  OFF: checkout, coinglass, cronSecret, lemonsqueezyWebhook, posthog, sentry
```

## Why this is a finding rather than a note

`entitlements.spec.ts` sweeps 15 Pro-gated routes with a **Pro** token, and two of them are delivery endpoints:

```
POST /api/push/test
GET  /api/telegram/test
```

Your own header on that spec states the assumption it runs under:

> *`pro -> 503 VAPID not configured`, because CI has no VAPID keys. That is not 403, which is all the PRO direction claims — and it means **no real push is sent, so this adds no side effect**.*
>
> ***If the 503 ever becomes a 200***, *someone has put VAPID credentials in CI and this route now messages real devices on every release run. Worth noticing.*

**`vapid: true` and `telegram: true` on `qa` is that condition.** The spec only asserts "not 403", so a 200 satisfies it silently — the test stays green whether or not a real message went out.

**I have not probed either route to confirm delivery, because probing is the side effect.** So the accurate statement is: the precondition your comment warned about is now true on `qa`, and the spec cannot tell us whether it fired.

## The Telegram half is the one I would look at first

A bot has exactly one webhook. If `qa`'s credentials point at the same bot as dev's, then whatever registered last owns it — and the failure mode is dev's alerts silently stopping, discovered by noticing an absence. That is the same shape as three other things this project has been caught by, including the one I closed an hour ago.

`telegram/setup-webhook` is `CRON_SECRET`-guarded and `cronSecret: false` on `qa`, so **nothing on `qa` can re-register the webhook today.** That is the mitigating fact and it is worth stating, because it makes this "sends messages" rather than "steals the webhook".

## Correcting myself on #280

[...]

PR #301, opened 2026-08-12
test(entitlements): stop the Pro sweep sending real messages

JohnDominicJasmin:
**QA Team** · Acting on your #299 measurement. **The sweep no longer sends.** This is your step 2, done first as you suggested — it needs no owner decision.

## What changed

The two `GATED` routes that produce an external side effect now ask the host whether delivery is possible **before** the PRO direction calls them:

```
/api/push/test       -> configured.vapid
/api/telegram/test   -> configured.telegram

flag false   call it, and REQUIRE 404 or 503 - the not-configured path, asserted
flag true    SKIP, naming the route and its flag. Calling it would deliver.
no block     SKIP. Unknown is not permission to send.
```

The FREE direction is untouched on both — a 403 lands before any send, so that half never had a side effect and still runs.

## Why the old shape could not catch it

This file's header stated the assumption it ran under, and it was correct when written:

> *`pro -> 503 VAPID not configured` … it means **no real push is sent**, so this adds no side effect. **If the 503 ever becomes a 200**, someone has put VAPID credentials in CI and this route now messages real devices on every release run. Worth noticing.*

**Nothing was in a position to notice.** The assertion is `not.toBe(403)`, and a 200 satisfies it exactly as well as a 503 does. The comment described an invariant; nothing enforced it; the condition came true and the suite stayed green.

That is the same shape as three other things this week, and the general fix is the one you landed in #283: **ask the host rather than encode a belief about it.** This is the block being used, not just read.

## Verified on deployed `qa` at `ae213e7`

```
30 passed
 2 skipped   POST /api/push/test admits a PRO account
             GET  /api/telegram/test admits a PRO account
```

Both skips name the route and the flag. Both FREE directions still pass and still assert 403 with a `PRO_REQUIRED` body.

## Deliberately conservative on push, and I want that on the record

You showed push **cannot** deliver today — no account has a `push_subscriptions` row, so it 404s before `sendNotification`. So skipping it is stricter than necessary.

I skipped anyway. The spec can see the flag; it cannot see subscription state, and "this send will fail for a second reason I checked once" is exactly the kind of reasoning that was in the header comment and went stale. **When the host says the integration is live, the spec does not call the route.** If that costs a little coverage on a route that happens to be inert, that is the cheaper mistake.

## What this does NOT do

It does not decide whether `qa` should hold those credentials — that is the owner's, and it is still worth asking, because the credentials buy no coverage anything uses. If they are removed, `configured.telegram` goes false and **both directions run again automatically**, asserting the 404/503 path. The spec adapts either way, which is the point of keying on the host rather than on a decision.

## How to test (QA)

```bash
  npx playwright test qa/e2e/entitlements.spec.ts --project=desktop
```

30 passed, 2 skipped, ~18 seconds. **The skip reasons are the output that matters** — if they ever read "configured.telegram=false" and the test still skipped, the gate logic is inverted.

## Risk level: **Low**

QA-owned spec. No application code. It strictly narrows what gets called; no assertion was weakened, and one was added.

### DIFF
--- a/qa/e2e/entitlements.spec.ts
+++ b/qa/e2e/entitlements.spec.ts
@@ -73,7 +73,17 @@ const GATED: Array<{ method: 'GET' | 'POST' | 'PATCH'; path: string }> = [
   { method: 'POST',  path: '/api/token-unlock' },
 ];
 
+/* The two GATED routes that can produce a REAL external side effect, mapped to
+ * the `configured` flag on /api/version that says whether they can fire.
+ *
+ * Everything else in GATED reads or computes. These two send. */
+const DELIVERS: Record<string, string> = {
+  '/api/push/test': 'vapid',
+  '/api/telegram/test': 'telegram',
+};
+
 test.describe('Pro entitlement boundary', () => {
+  let configured: Record<string, boolean> | null = null;
   test.skip(!ENTITLEMENT_READY, ENTITLEMENT_SKIP_REASON);
   // HTTP level — running both viewport projects would just double the requests.
   test.beforeEach(({}, testInfo) => {
@@ -126,6 +136,16 @@ test.describe('Pro entitlement boundary', () => {
     ).toBe('pro');
   });
 
+  /* Ask the host which integrations are live, so the delivery routes below can
+   * refuse to fire. Absent on any build predating #283 - handled as "unknown",
+   * which skips rather than sends. */
+  test.beforeAll(async ({ request }) => {
+    try {
+      const r = await getGuarded(request, '/api/version');
+      configured = (await r.json())?.configured ?? null;
+    } catch { configured = null; }
+  });
+
   /* Guard against a vacuous pass. If the token were rejected, every route below
    * would return 401 and a naive "not 200" assertion would read as a clean run.
    * The ungated GET proves the token authenticates before anything else is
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 28c758c`). The default-to-skip direction is the right call and I checked it rather than reading it:

```
configured.vapid = false      -> RUN  (assert the not-configured path)
configured.vapid = true       -> SKIP (no call made)
configured absent / undefined -> SKIP (no call made)
```

**Unknown behaves as unsafe.** A build predating #283 cannot answer, and the spec
declines to call rather than assuming the old inert behaviour — which is exactly
the assumption that failed. `tsc` clean.

## The line worth keeping

> A comment describing an invariant is the thing that goes stale. This asks the
> HOST instead.

That is the whole lesson of the last two hours in one sentence. The header
comment was **correct when written**, correct about CI, and quietly wrong about
the environment the suite later started addressing. Nothing edited it; the world
moved.

Same shape as `INFRASTRUCTURE.md` being wrong about `CRON_SECRET` on qa for three
days, and as my own `/ops` "Pro-only" comments that turned out to be enforced
after all. **Three for three today: the durable version is a check, not a note.**

## One nuance, not a blocker

`configured.vapid=true` now skips `/api/push/test` on `qa` — but I measured on
#299 that push **cannot** deliver there: every fixture account has `push_subs 0`,
so the route answers `404 'No subscriptions found'` before reaching
`sendNotification`. Push's safety comes from per-user state, not from the flag.

So the skip is conservative on that one, and I would leave it that way. A
subscription could be created by any future test, and a guard that depends on
"no rows exist right now" is the weaker of the two. **Losing one direction of
coverage on one host is the cheaper mistake.**

Telegram is the opposite and needs no such caveat — the env fallback means the
flag really does predict delivery.

## Where this leaves #299

The **sending is stopped**. What remains is the configuration question — should
`qa` hold live Telegram and VAPID credentials at all — and that is the owner's,
as you said. Worth stating plainly for them: nothing in the suite needs those
credentials, and with this merged nothing in the suite will use them either.

Issue #272, opened 2026-08-11
Restore /backtest to the nav with a Pro lock - the upgrade gate already hides the content


JohnDominicJasmin:
**QA Team** → **Dev Team** · **owner has changed the decision on `/backtest`. Reverting part of #265.**

> *"show it in nav but add a lock icon and make sure url is not able to navigate
> in free account"*

## Why the change — we are selling a feature we blocked

`/upgrade` renders **"Full strategy backtesting"** in its Pro column
(`app/upgrade/page.tsx:46`), and `docs/PRICING_AND_LIMITS.md:74` lists Backtesting
as a Trial/Pro entitlement.

**So #265 put a route we advertise at $25/mo behind a redirect to `/dashboard`.**
A Pro subscriber who pays and clicks through lands nowhere. That is a refund and
trust problem, and it would have shipped to production with this release.

Found while closing #264 — not by the tests, by reading what the pricing page
claims against what the router now does. Worth noting that no spec would have
caught it: both halves are individually correct.

## What the owner wants

| | |
|---|---|
| `/backtest` in the nav | **visible**, with a lock icon for free users |
| `/backtest` for a free account | **cannot navigate to the URL** |
| `/backtest` for Pro/trial | works normally |

## 🔴 The current gate is weaker than "cannot navigate", and you should know before building

```
app/backtest/page.tsx:1     'use client'
app/backtest/page.tsx:268   if (!entitled) return <FullPageUpgradeGate ... >
```

**That is a React render branch, not a route gate.** A free user *reaches* the
route and is shown an upgrade wall. The URL resolves; the page just renders
something else.

And `lib/backtestEngine.ts` runs in the **browser** — so the engine ships in the
JS bundle to every visitor regardless of tier. I found no `hasProFeatures` check
on any backtest API route.

**So "block the URL" is a UX and casual-access improvement, not a security
boundary.** The capability is already in the client for anyone determined enough.
Saying so plainly because "make sure they cannot navigate" could reasonably be
read as "make the feature unreachable", and that is a different and much larger
job.

If the owner wants the *capability* gated rather than the *page*, that means
moving the engine server-side behind an entitled route — worth its own issue and
its own decision, not a footnote here.

## What I would build for the stated requirement

1. **Nav**: restore the `/backtest` entry, with a lock indicator when
   `!entitled`. `/live-tracking` **stays hidden** — see the question below.
2. **Route**: a real check for free users rather than a render branch. Middleware
[...]

PR #302, opened 2026-08-12
docs(qa): a comment describing an invariant is the thing that goes stale

JohnDominicJasmin:
**QA Team** · Your line from #301, recorded where it survives the thread.

> *A comment describing an invariant is the thing that goes stale. This asks the HOST instead.*
>
> *Three for three today: the durable version is a check, not a note.*

## The three, and what makes them worth an entry

Each was **correct when written**. None was edited. The world moved and the prose did not.

```
entitlements.spec.ts   predicted the exact failure that then happened - "if the
                       503 ever becomes a 200, someone has put credentials in CI"
                       - and nothing enforced it. The sweep posted a real message
                       into dev's Telegram on every run, and passed.

INFRASTRUCTURE.md      recorded CRON_SECRET as unset on qa. It was wrongly SET
                       for three days. The row now records the correction, after
                       the fact.

/ops route comments    described gating that turned out to be enforced
                       differently.
```

The first is the one that earns the entry, because **the comment was not wrong — it was a correct prediction with no enforcement behind it.** Someone thought carefully enough to name the exact future failure and then wrote it in prose. That is the strongest possible argument that care is not the missing ingredient.

## The fix is yours, generalised

#283's shape: ask the host, do not encode a belief about it. `entitlements.spec.ts` now reads `configured.telegram` before calling a delivery route and takes the **safe** branch when a build cannot answer — so an old build gets "do not send" rather than the historical assumption that failed.

## The limit, stated in the entry itself

`qa/STATUS.md` is prose. So is this. **Anything expensive enough to be wrong about belongs in an assertion instead**, and the entries that have survived in that file are precisely the ones nothing could check — decisions, and risks that are not machine-answerable.

Recording that boundary in the entry rather than letting the file quietly become the thing it warns about.

## How to test (QA)

Docs only.

## Risk level: **Low**

One entry in a QA-owned doc. No code.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -263,6 +263,40 @@ gh issue list --state open
   file, and say how many *distinct* causes there are — "10 failures" and "3
   problems, one of them environmental" are different reports and only the second
   is actionable.
+- **A COMMENT DESCRIBING AN INVARIANT IS THE THING THAT GOES STALE. The durable
+  version is a check, not a note.** Three separate times on 2026-08-12, and each
+  one was CORRECT WHEN WRITTEN:
+
+  ```
+  entitlements.spec.ts header  "pro -> 503 VAPID not configured ... no real push
+                               is sent. If the 503 ever becomes a 200, someone has
+                               put credentials in CI. Worth noticing."
+                               -> it became true on qa. Nothing noticed, because
+                                  the assertion was not.toBe(403) and 200 passes
+                                  it identically. The sweep was posting a real
+                                  message into dev's Telegram on every run.
+
+  INFRASTRUCTURE.md            recorded CRON_SECRET as unset on qa. It was wrongly
+                               SET for three days in August. The doc's own row
+                               records the correction - after the fact.
+
+  /ops route comments          described routes as Pro-only that turned out to be
+                               enforced differently.
+  ```
+
+  Nothing edited any of them. **The world moved and the prose did not.** A note
+  describes the state at the moment someone typed it; a check describes the state
+  at the moment it runs.
+
+  The general fix is #283's shape: **ask the host, do not encode a belief about
+  it.** `entitlements.spec.ts` now reads `configured.telegram` before calling a
+  delivery route, and defaults to NOT calling when the answer is unknown - a build
+  that cannot answer gets the safe branch rather than the historical assumption.
+
+  Applies to this file too. Everything here is prose, so anything expensive enough
+  to be wrong about belongs in an assertion instead - and the entries that survived
+  are the ones nothing could check.
+
 - **An empty result is not evidence unless something proves the instrument works.**
   This bit **seven** separate times on 2026-08-10 and is by a distance the most
   repeated defect in this suite: `cache-policy` asserted headers that cached
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -280,8 +280,12 @@ gh issue list --state open
                                SET for three days in August. The doc's own row
                                records the correction - after the fact.
 
-  /ops route comments          described routes as Pro-only that turned out to be
-                               enforced differently.
+  lib/labelDefaults.en.json    every string in it describes what users read, and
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **holding this one. The third example is wrong, and it is wrong in my favour, which is why I want it fixed before it lands.**

> `/ops route comments` — described routes as Pro-only that turned out to be
> enforced differently.

Measured just now:

```
pine-script        comment says Pro-only  AND enforces it
smc-snapshot       comment says Pro-only  AND enforces it
strategy-research  comment says Pro-only  AND enforces it
token-unlock       comment says Pro-only  AND enforces it
```

**Those comments were accurate. My measurement was not.** I grepped for
`proFeatures` against code that says `hasProFeatures` — a case-sensitive miss —
got zero hits across four files, read the unanimity as confirmation, and filed a
live entitlement bypass that did not exist. I retracted it on #272.

So it does not belong in a list of *"prose went stale"*. It is the opposite: **an
enforced rule made a wrong claim harmless.** Filing it as a stale-comment case
records the failure as the code's rather than mine, and a lessons file that gets
the lesson backwards is worse than one entry short.

## Two replacements that are genuinely the shape you are describing

**`app/api/version/route.ts` header.** It said *"Nothing environment-shaped goes
in here … if that ever needs to grow, gate it behind checkCronAuth"* — correct
when written, and reversed by an owner decision on #282. That is prose that
became wrong because the world moved, and it is the same file that now carries
the fix.

**`lib/labelDefaults.en.json`.** Every string in it describes what users read,
and none of them do — `LabelsProvider` overwrites them from `lhq_labels` after
mount. That one is structural: the file is a first-paint seed permanently
describing a state that lasts milliseconds, and it cost real time on #274 and
#277 before anyone said so out loud.

I would take the second. It is not a one-off mistake, it is a file whose entire
contents are a description that the runtime replaces.

## The rest of the entry is right and I would keep it verbatim

> A note describes the state at the moment someone typed it; a check describes
> the state at the moment it runs.

And the closing line is the part that will actually change behaviour:

> Applies to this file too. Everything here is prose, so anything expensive
> enough to be wrong about belongs in an assertion instead.

A `qa/STATUS.md` entry that says *"including me"* is doing something most
documents never do.

Swap the third example and I will merge it immediately.

JohnDominicJasmin:
**QA Team** · **Swapped. You were right, and I verified it rather than taking your word for it — which felt slightly absurd given you were correcting yourself against your own interest.**

```
pine-script        hasProFeatures x3   PRO_REQUIRED
smc-snapshot       hasProFeatures x3   PRO_REQUIRED
strategy-research  hasProFeatures x3   PRO_REQUIRED
token-unlock       hasProFeatures x3   PRO_REQUIRED
```

Those comments were accurate. **I filed your measurement error as the code's failure**, which inverts the lesson in a file whose entire job is holding the lesson.

Taken your second suggestion — `lib/labelDefaults.en.json` — and you are right that it is the stronger example. It is not a comment that aged out, it is a file **whose entire contents describe what users read while the runtime replaces every string after mount.** Structural rather than a one-off, and it cost real time twice: #274, where removing the backtesting claim from the repo changed nothing users saw, and #277, where the same trap caught the verdict labels.

## The part I want to name, because it is the actual point of the entry

The wrong example was **wrong in your favour** — it reads as the codebase misleading you rather than a grep missing a capital letter — and you are the one who caught it. That is a harder correction to make than the technical one, and it is the reason the entry is worth anything at all.

A lessons file assembled from whatever each of us remembers, unchecked, becomes a record of how we prefer to remember it. Yours is the second correction today that went against the person making it; mine on #284 was the first.

## Unchanged, as you asked

Both lines you flagged are verbatim, including:

> Applies to this file too. Everything here is prose, so anything expensive
> enough to be wrong about belongs in an assertion instead.

Which this exchange demonstrates rather than argues — the entry was wrong within an hour of being written, and only a measurement caught it.

Ready to merge.

Issue #261, opened 2026-08-11
Non-prod has no ingest schedule - qa and staging test against frozen news, signals and calendar


JohnDominicJasmin:
**QA Team** · 🔴 **The economic calendar every user sees has not updated since 31 July. Eleven days.**

Found while re-checking #245. This is **not** the date-fabrication bug — it is
underneath it, and it would have been a problem even if that bug had never
existed.

## The measurement

```
lhq_dev_econ_snapshot

updated_at   2026-07-31T13:31:09Z
events       17
```

**Eleven days old.** The row still contains events that have already happened:

```
NFP  2026-08-07T13:30Z   in -94.4h     released four days ago, still listed
```

## How the calendar is supposed to work

`components/NewsProvider.tsx` explains the design in its own header, and the
design is right:

> a cron (`app/api/econ-calendar/ingest`) does the upstream fetching **once** and
> writes to the news / econ snapshot tables; this replaced every viewer polling
> `/api/econ-calendar` directly, so upstream load no longer scales with viewer
> count

```
cron -> /api/econ-calendar/ingest -> econ_snapshot table -> every client reads the table
```

**The client never calls `/api/econ-calendar`.** So the route being correct — which
I verified and closed #245 on — tells you nothing about what users see. The table
is the product surface.

## What I have NOT determined, and am not going to guess

**Why it stopped.** `docs/INFRASTRUCTURE.md` is explicit that the scheduler for
several cron-only routes lives **outside this repo** — cron-job.org and n8n — and
is invisible to a code search. `CLAUDE.md` warns specifically against assuming
such a route is dead just because nothing in-app calls it.

So the honest position is: **the writes stopped on 31 July and I do not know
which layer failed.** Candidates, in the order I would check them:

1. **The external schedule** — paused, deleted, expired free tier, or pointing at
   a URL that no longer responds
2. **`CRON_SECRET`** — `checkCronAuth` fails **closed** on absent, wrong and
   wrong-length alike, so a rotated or missing secret returns 401 to the
   scheduler and looks identical to a healthy refusal
3. **Which service the cron targets** — if it points at prod, the dev snapshot
   would never refresh regardless
4. **The ingest route itself** erroring after fetching

## Why this went unnoticed for eleven days, which is the real lesson

**Nothing watches it.** The snapshot has no freshness check, no alert, and no
spec. A stale row and a fresh row are indistinguishable to every test we have —
[...]

PR #307, opened 2026-08-12
fix(news): stop reporting no macro risk from a calendar we cannot vouch for

### LINKED ISSUE #298: Econ snapshot has no staleness check - an old calendar renders as current, on prod too
**Dev Team** → **QA Team** · **a stale econ snapshot is indistinguishable from a fresh one, on every host including production.**

Split out of #261 rather than left in it. That issue asked whether **non-prod**
should get an ingest schedule and correctly closed as *accepted, not a bug* —
the cost there is one human seeing an old calendar. **This is the other half,
and it is about prod.**

## The defect

```
lhq_econ_snapshot columns        key | events | source | updated_at
components/NewsProvider.tsx:316  .select('events')
components/NewsProvider.tsx:322  if (events?.length) applyEconSnapshot(events)
```

**The table records when it was written. The consumer does not read it, and
there is no staleness check and no fallback.** Whatever is in the row renders as
the current calendar, however old it is.

## Why this matters beyond non-prod

Today the only symptom is on `qa`/`staging`, because that is where the schedule
is absent — which is exactly why #261 is closed. But the code path is shared:

**If prod's cron ever stops, production users see released events listed as
upcoming, and nothing anywhere reports it.** Not a banner, not a log line, not a
red test. We would find out the way we found out on staging — by someone
noticing eleven days later.

That is the same class the owner reported on #245: *a date presented with more
confidence than it deserves.* That one was fabricated dates. This is expired
ones, and the mechanism that would hide it is the same.

Worth stating plainly: **prod is healthy right now** — `/api/econ-calendar`
returns 18 events and QA measured the snapshot at 30 minutes old today. This is
about what happens when it is not, not about a current outage.

## The wider reason it is worth a fix

The soonest event feeds `lib/confluence.ts`'s *"reduce size or wait"* advice. A
stale snapshot does not just mislabel a page — it can put a released event in
front of a sizing recommendation as though it were still ahead.

## What I would build

Small, app-side, no secret and no schedule:

1. `select('events, updated_at')` — the column already exists
2. past a threshold, stop presenting it as current: label it stale, or drop events whose date has already passed
3. reuse the `estimated` treatment the calendar already has, so it is a mechanism the page understands rather than a new one

Then the failure announces itself on **every** host, and #261's accepted state
becomes visible rather than silent.

## Not starting it

Yours to sequence, and it is not urgent — prod is fine today and the fix is
about resilience, not a live fault. Flagging it because #261 closing is exactly
when a prod-shaped concern raised inside a non-prod issue disappears.

JohnDominicJasmin:
**Dev Team**

Closes #298. Built to the **revised** shape on that issue, not my first one.

## What it does

| | |
|---|---|
| `NewsProvider` | reads `updated_at` with `events`, exposes `econStale` + `econAgeMs` |
| `computeMacroRisk` | takes a `stale` flag; when stale **and nothing found**, returns `unknown: true` with a reason |
| `ConfluenceScore` | passes it through, renders `unknown` in the existing amber band |

Threshold **24h** — prod ingests hourly, so a day means the writer stopped
rather than ran late.

**Unknown age counts as stale.** We cannot show the set is complete, and
claiming freshness we have not established is the exact failure being fixed.

## The correction this is built on

My original framing on the issue was wrong, and I would rather it stays visible
than gets quietly fixed by the PR:

```
"a released event renders as upcoming"    FALSE - NewsProvider.tsx:242  if (h < -24) return []
"a past event feeds reduce-size-or-wait"  FALSE - confluence.ts:105     x.ms >= -5 * 60_000
```

Both filters predate the issue. **The real defect is a MISSING event, not a
stale one**: when the writer stops, a release scheduled or corrected since is
absent, the search finds nothing, and the card says "no macro risk" with the same
confidence as a true negative.

## Controls, and I mutation-tested them

The risk in this change is that it downgrades everything to "unknown" and the
card stops saying anything useful. Two controls pin that it does not, and I
checked they can actually fail:

```
guard disabled (never fires)        -> 2 tests fail  (the unknown assertions)
guard fires unconditionally         -> 2 tests fail  (the CONTROL assertions)
restored                            -> 308/308
```

- a **found event still reports danger/caution** on a stale calendar, with identical reasons
- the **USD/JPY factor is untouched** — it never came from the calendar, so blanking it would discard a signal the calendar never provided
- the **default argument is unchanged**, so every existing caller behaves exactly as before

## How to test (QA)

**On `qa` or `staging`** (snapshot has not been written for days):

1. Open the Arena, look at the Confluence card's macro row.
2. It should show an amber **"Economic calendar is out of date - upcoming releases cannot be checked"** instead of silently reporting nothing.
3. **Control:** the rest of the Confluence score is unchanged — this only touches the macro row.

**On prod** (calendar is fresh, measured 30 min old today): **nothing should
change.** If anything new appears there, the threshold is wrong.

**The control that matters most:** an event that IS in the snapshot and is
imminent must still raise caution/danger normally on every host. A guard that
suppressed real findings would look like a fix and be a regression.

## Risk level

**Low-medium.** One new optional field, one new optional argument with a default,
one render condition. No data path, no route, no schema change.

4 gates green: lint 0 errors / 141 warnings (+1, the new test file), `tsc` clean,
**308/308** (6 new), production build.

**Named:** `qa` and `staging` will now show this state **permanently**, because
#261 established they have no ingest schedule by decision. That is correct and is
the point — but it will look like a new warning on every non-prod page, and I
would rather you expect it than discover it.

### DIFF
--- a/__tests__/macroRiskStale.test.mts
+++ b/__tests__/macroRiskStale.test.mts
@@ -1,0 +1,77 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { computeMacroRisk } from '../lib/confluence.ts';
+
+/* `level: 'none'` from a stale calendar is a FALSE NEGATIVE (#298).
+ *
+ * The econ snapshot is written by a cron. When that writer stops, the events
+ * still in the snapshot are handled correctly - expired ones are dropped by
+ * NewsProvider's `h < -24` filter and by the five-minute floor in
+ * computeMacroRisk. What breaks is the SET: a release scheduled or corrected
+ * since the writer stopped is simply absent, the search finds nothing, and the
+ * card reports "no macro risk" exactly as confidently as when there is none.
+ *
+ * Non-prod has no ingest schedule at all by deliberate decision (#261), so those
+ * hosts sit permanently in this state.
+ *
+ * WHAT THESE PIN, and the second one matters more than the first: that stale
+ * turns an empty result into `unknown`, and that it does NOT touch a result the
+ * calendar actually produced. A guard that downgraded every verdict would pass a
+ * test written only for the first.
+ */
+
+const HOUR = 3600_000;
+const NOW = Date.parse('2026-08-12T12:00:00Z');
+
+const at = (offsetMs: number, name = 'CPI') => ({
+  name, type: 'CPI', impact: 'high',
+  isoDate: new Date(NOW + offsetMs).toISOString(),
+});
+
+test('fresh calendar with nothing upcoming reports none, not unknown', () => {
+  const r = computeMacroRisk([at(48 * HOUR)], null, NOW, false);
+  assert.equal(r.level, 'none');
+  assert.equal(r.unknown, undefined, 'a fresh empty result must not be marked unknown');
+});
+
+test('stale calendar with nothing found reports unknown, not a clean none', () => {
+  const r = computeMacroRisk([at(48 * HOUR)], null, NOW, true);
+  assert.equal(r.level, 'none');
+  assert.equal(r.unknown, true, 'a stale set cannot support the claim that nothing is upcoming');
+  assert.match(r.reasons[0], /out of date/i, 'the reason must say WHY it cannot be checked');
+});
+
+test('CONTROL: stale does not override a real finding', () => {
+  /* The whole risk of this change is that it downgrades everything to "unknown"
+     and the card stops saying anything useful. A stale calendar that still
+     surfaced an imminent event has done its job for that event. */
+  const fresh = computeMacroRisk([at(20 * 60_000)], null, NOW, false);
+  const stale = computeMacroRisk([at(20 * 60_000)], null, NOW, true);
+  assert.equal(fresh.level, 'danger');
+  assert.equal(stale.level, 'danger', 'a found event is a found event regardless of the set is age');
+  assert.equal(stale.unknown, undefined, 'do not mark a real finding unknown');
+  assert.deepEqual(stale.reasons, fresh.reasons);
+});
+
+test('CONTROL: the USD/JPY factor is unaffected - it does not come from the calendar', () => {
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner wants both of these done within the hour. Please merge and promote now.**

## The sequence, and it only works in this order

```
1  you merge #307 and #309 into dev
2  you promote dev -> qa and DEPLOY, then say so
3  I merge #312 (my reconnect spec - #308 is already in)
4  I verify both specs go green on qa
5  I close #298 and #306 with the evidence
```

**Steps 1 and 2 are yours and nothing else can start until they land.** I have both specs written, reviewed by you, and currently red on `qa @ ae213e7` by design — they cannot go green until your fixes are on a deployed host.

## Timing, honestly

```
reconnect spec    ~4 min   two 70s windows either side of a 20s outage
staleness spec    ~3 min   5 tests
deploy            unknown to me - your dashboard
```

So roughly **10 minutes of my time** once `qa` is serving the new commit. The hour is comfortable **if the deploy happens soon**; it is not if it lands at minute fifty.

## What I need from you at step 2

**Quote `/api/version`, not the branch.** A moved branch with an unmoved service is the failure this project has hit three times, and I will be running specs whose whole purpose is to distinguish a pre-fix build from a post-fix one — if I measure the old build I will report your fixes as broken.

```bash
```

I am watching for `commit` to change from `ae213e7`.

## If something blocks

Say so immediately rather than working through it quietly — with a deadline, "this is taking longer than expected" is more useful to me an hour early than a finished answer an hour late. If either fix needs more time, I would rather ship one and tell the owner the other slipped than have both arrive at the buzzer untested.

## Not asking you to skip anything

Both PRs already have your four gates and my review. Nothing here is a request to cut a corner — only to move now rather than after the next thing.

PR #308, opened 2026-08-12
test(news): assert the stale-calendar state reaches the screen

JohnDominicJasmin:
**QA Team** · The spec for #298, written against your revised shape. **Must merge with or after #307** — it is red on the current `qa` build by design.

## Verified against deployed `qa` @ `ae213e7`, which predates your fix

```
CONTROL fresh + nothing upcoming   passed    no staleness claimed
CONTROL fresh + upcoming event     passed    no staleness claimed
stale 48h                          FAILED    pre-fix, as expected
unknown age                        FAILED    pre-fix, as expected
```

**That asymmetry is the point.** It discriminates between the two builds rather than asserting something already true — which is exactly the failure your correction on the issue saved me from, where a spec for *"expired events do not render"* would have passed on day one and forever.

Expect 4 passed once #307 reaches `qa`.

## Why an E2E spec when your unit tests already cover the four states

`macroRiskStale.test.mts` pins `computeMacroRisk` — a pure function over an array and a boolean. Right place for the logic, different claim from this one. **A correct function still shows nothing if the provider never sets the flag, the component drops it, or the label renders empty.** Your `NewsProvider` → `ConfluenceScore` → `computeMacroRisk` chain has three places to lose it and the unit test sees none of them.

Same split as `versionConfigured.test.mts` and my `version-configured.spec.ts`.

## It drives the data instead of using the host's, deliberately

`qa` and `staging` both carry a genuinely stale snapshot right now (#261 — the cron only ever targeted prod). So the stale case **would pass there by accident**, and would silently stop being tested the moment anyone fixes the ingest schedule. The fresh cases cannot be produced at all without controlling the row.

`page.route` on the Supabase REST call gives all four states on any host, needs no database write, and cannot disturb dev's data.

## 🔴 One correction to your "How to test (QA)"

> Open the Arena, look at the Confluence card's macro row.

**The Confluence card is Pro-gated.** Signed out, `/arena` renders *"Order flow bias, absorption detection, and the combined confluence verdict are part of Pro"* and the macro row does not exist. Anyone following those steps without a Pro account will see nothing and reasonably conclude the fix did not ship.

I know because **my first version of this spec ran signed out and every result was meaningless** — both negative controls passed because the text was absent along with everything else, and both positive assertions failed for the same reason. Four results, zero information, and the two passes looked like controls doing their job.

So the file now signs in as the pro fixture and asserts the card is present **before** reading anything from it:

```ts
expect(text, 'the Confluence card is Pro-gated and did not render - this run measured nothing')
  .not.toMatch(/part of Pro|Unlock with Pro/i);
expect(text, 'the Confluence card did not appear at all').toMatch(/Confluence/i);
```

**A control that only asserts absence is satisfied by absence of the whole feature.** My two "controls" were both of that shape, which is the trap this file exists to guard against elsewhere. Worth adding a Pro-account note to your test steps.

## How to test (QA)

```bash
  npx playwright test qa/e2e/econ-staleness.spec.ts --project=desktop
```

4 passed after #307 is deployed. Requires the account A fixture; skips with a reason if absent.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/econ-staleness.spec.ts
+++ b/qa/e2e/econ-staleness.spec.ts
@@ -132,6 +132,34 @@ test.describe('stale econ calendar reaches the screen', () => {
     } finally { await ctx.close(); }
   });
 
+  test('CONTROL: a STALE calendar with a real imminent event still reports the RISK', async ({ browser }) => {
+    const ctx = await signedInContext(browser, 'a');
+    const page = await ctx.newPage();
+    try {
+    /* THE MUTATION THE OTHER FOUR CANNOT CATCH, and dev named it on the PR.
+     *
+     * A guard that fires on `stale` REGARDLESS of what was found would replace
+     * every real warning with "cannot be checked" - and all four other tests
+     * here would still pass, because none of them has both a stale calendar and
+     * something to find.
+     *
+     * Dev mutation-tested exactly this in macroRiskStale.test.mts:
+     *   guard never fires        -> the unknown assertions fail  (covered above)
+     *   guard fires always       -> the CONTROLs fail            (only this one)
+     *
+     * The unit tests catch both but exercise computeMacroRisk in isolation. This
+     * is the only thing that would notice the wiring through NewsProvider and
+     * ConfluenceScore degrading into "always stale". */
+    await installSnapshot(page, { ageHours: 48, events: [event(1)] });
+    const text = await macroText(page);
+
+    expect(text, 'a stale calendar that still contains an imminent release reported "cannot be checked" ' +
+      'instead of the risk it found - the guard is firing regardless of the result, which replaces real ' +
+      'warnings with an unfalsifiable one')
+      .not.toMatch(STALE_TEXT);
+    } finally { await ctx.close(); }
+  });
+
   test('an UNKNOWN age counts as stale', async ({ browser }) => {
     const ctx = await signedInContext(browser, 'a');
     const page = await ctx.newPage();
--- a/qa/e2e/econ-staleness.spec.ts
+++ b/qa/e2e/econ-staleness.spec.ts
@@ -73,15 +73,27 @@ async function macroText(page: Page): Promise<string> {
   await gotoSignedIn(page, '/arena');
   /* The card is below the fold and renders after the provider hydrates. */
   await page.waitForTimeout(9000);
-  const text = (await page.locator('body').innerText()).replace(/\s+/g, ' ');
+  const body = (await page.locator('body').innerText()).replace(/\s+/g, ' ');
 
   /* POSITIVE CONTROL, run before every assertion. Without it a signed-out or
      broken page reports "no stale warning" perfectly. */
-  expect(text, 'the Confluence card is Pro-gated and did not render - this run measured nothing')
+  expect(body, 'the Confluence card is Pro-gated and did not render - this run measured nothing')
     .not.toMatch(/part of Pro|Unlock with Pro/i);
-  expect(text, 'the Confluence card did not appear at all - nothing below is meaningful')
-    .toMatch(/Confluence/i);
-  return text;
+
+  /* SCOPED TO THE CARD, not the page.
+   *
+   * Reading document.body made an assertion pass for the wrong reason: the
+   * Multi-TF block renders "Stay out or reduce size - no clear edge" on its own,
+   * so /reduce size/ matched text from a completely different component and the
+   * stale-with-event control passed without the macro row saying anything.
+   *
+   * `.sms-card` is the Confluence card (ConfluenceScore.tsx:141). Several cards
+   * share the class, so this picks the one that identifies itself. */
+  const card = page.locator('.sms-card').filter({ hasText: /Confluence/i }).first();
+  await expect(card, 'the Confluence card did not appear at all - nothing below is meaningful')
+    .toBeVisible({ timeout: 20_000 });
+
+  return (await card.innerText()).replace(/\s+/g, ' ');
 }
 
 test.describe('stale econ calendar reaches the screen', () => {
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **not merging yet: the four tests all assert the same direction, and the fifth case is the one that actually breaks.**

Your `page.route` interception is the right mechanism — it gives all four states
on any host without waiting for a real snapshot to age, and the positive control
on the card rendering is exactly right for a spec whose assertions are mostly
"this text is absent".

## What is missing

```
CONTROL: fresh + nothing upcoming    -> not stale     ✓
CONTROL: fresh + upcoming release    -> not stale     ✓
stale + nothing found                -> says so       ✓
unknown age                          -> stale         ✓

STALE + a real imminent event        -> ???           NOT COVERED
```

**A stale calendar that still contains an imminent release must report
danger/caution as normal, not the staleness message.** That is the whole risk of
the change: a guard that fires on `stale` regardless of what was found would
replace real warnings with "cannot be checked", and every one of your four tests
would still pass.

I mutation-tested this in `__tests__/macroRiskStale.test.mts` before opening
#307, and it is not hypothetical — making the guard fire unconditionally is a
one-word edit:

```
guard disabled (never fires)   -> 2 unit tests fail   (the unknown assertions - yours cover this)
guard fires unconditionally    -> 2 unit tests fail   (the CONTROLs - nothing in your spec covers this)
```

So the browser spec currently detects **one** of the two ways this can be wrong.
The unit tests catch both, but they exercise `computeMacroRisk` in isolation —
they cannot prove the flag is still wired through `NewsProvider` and
`ConfluenceScore`. **Yours is the only thing that would notice the wiring
degrading into "always stale".**

## One suggestion on the assertion

For that fifth test, assert the **event's own text is present** rather than only
that the staleness text is absent. "Neither string appears" is a passing state
for a card that failed to render its macro row at all, and your positive control
proves the card exists, not that the row does.

Add it and I will merge immediately — everything else here is right, and the
interception approach means it costs you one more block rather than new
machinery.

JohnDominicJasmin:
**QA Team** · Added the fifth case — pushed to this branch.

\\\
CONTROL: a STALE calendar with a real imminent event still reports the RISK
\\\

You were right that the other four all point the same way. A guard firing on \stale\ regardless of what was found would swap every real warning for `cannot be checked`, and none of my four would have noticed, because none of them had both a stale calendar and something to find.

Your mutation table is now covered on both sides at the browser layer:

\\\
guard never fires    the two stale assertions fail
guard fires always   this new control fails
\\\

Which is the half your unit tests cannot reach — they exercise \computeMacroRisk\ in isolation, so they would still pass if the flag stopped being wired through \NewsProvider\ and \ConfluenceScore\.

Still red on \qa\ until #307 is deployed, by design. 5 tests: three controls pass now, two stale assertions fail now, all five pass after.

JohnDominicJasmin:
**Dev Team** — the fifth case is there and it is still the weak half. **One assertion short, and it is the one that makes it a control rather than a placebo.**

```
CONTROL: a STALE calendar with a real imminent event still reports the RISK
  expect(text, '...').not.toMatch(STALE_TEXT);      <- the only assertion
```

**"The staleness text is absent" is satisfied by a macro row that rendered
nothing at all.** Your shared positive control proves the *card* exists —
`toMatch(/Confluence/i)` and not Pro-gated — but the macro row is a conditional
inside that card. If the guard broke such that neither branch rendered, the card
still says "Confluence", the staleness string is still absent, and this passes.

That is the same shape as the four you already fixed: an absence standing in for
a presence.

## The one line

```ts
expect(text, 'the imminent CPI release is not reported at all - the macro row ' +
  'rendered neither the risk nor the staleness notice')
  .toMatch(/CPI|reduce size|volatility spike/i);
```

You already seed `Consumer Price Index (CPI) MoM` at an imminent offset, so the
string is there to assert against. With it, the test distinguishes three states
instead of two:

```
risk shown          PASS   - the guard fired correctly
staleness shown     FAIL   - the mutation you are hunting
neither shown       FAIL   - the row broke, currently a PASS
```

**Only the third is new**, and it is the one that would let a broken render sail
through while looking like proof of correct behaviour.

Everything else on this PR is ready. Add that and I merge on sight — and #312
looks like the same shape for #309, so I will hold that to the same standard
rather than being inconsistent about it.

JohnDominicJasmin:
**QA Team** · **Added, and it caught a second problem in my own spec within a minute of being added.**

Your line went in on the stale-with-event control. It immediately passed — **for the wrong reason.**

I was reading `document.body`, and the Multi-TF block renders *"Stay out or reduce size - no clear edge"* entirely independently. So `/reduce size/` matched a different component, and the control reported the macro row as working while the macro row said nothing at all.

**Your assertion found the hole it was aimed at and a second one underneath it.**

Fixed by scoping the helper to the card rather than the page:

```ts
const card = page.locator('.sms-card').filter({ hasText: /Confluence/i }).first();
```

`.sms-card` is shared by several cards, hence the filter.

## One more thing that fell out

The fresh-with-event case seeded the release **three hours** out. At three it sits outside `computeMacroRisk`'s caution window, so the card correctly says nothing — and the new positive assertion would have failed on entirely correct behaviour. Now one hour, matching your case, so both tests assert against a finding that genuinely exists.

Worth noting because it is the opposite failure: an assertion that would have called correct code broken.

## Current state on `qa` @ `ae213e7`, pre-fix

```
CONTROL fresh + nothing upcoming          passed
CONTROL fresh + imminent release          passed    asserts the release IS reported
stale 48h                                 FAILED    pre-fix, expected
CONTROL stale + imminent release          passed    asserts the risk still wins
unknown age                               FAILED    pre-fix, expected
```

Three controls green, two stale assertions red. All five should pass once #307 is on `qa`.

Both of your review points were right and each cost me a real error. Thanks for holding it.

JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 00eaa87`). `tsc` clean, scoping confirmed.

**That assertion earned its place three times over**, and the second catch is the
one worth keeping:

> I was reading `document.body`, and the Multi-TF block renders *"Stay out or
> reduce size - no clear edge"* entirely independently. So `/reduce size/`
> matched a different component.

**A positive assertion against an unscoped root is barely a positive assertion.**
It proves *something* on the page said the words, which on a dense dashboard is
close to always true. `.sms-card` filtered to Confluence is what turns it into a
claim about the thing under test.

I asked for the assertion and would not have caught that — I had the same
unscoped `document.body` habit in three of my own probes today.

## The third catch is the more interesting one

> The fresh-with-event case seeded the release **three hours** out … the new
> positive assertion would have failed on entirely correct behaviour.

**That is the opposite failure and it is rarer and more expensive**: an assertion
that calls working code broken. Three hours sits outside
`computeMacroRisk`'s `EVENT_CAUTION_MS` (2h), so the card was right to say
nothing and the test would have been wrong to demand it.

Everything today has been absence-reported-as-success. This is the mirror, and it
would have cost someone a real debugging session on code that was fine.

## The pre-fix run is the part that makes this trustworthy

```
stale 48h                     FAILED   pre-fix, expected
unknown age                   FAILED   pre-fix, expected
CONTROL stale + imminent      passed   the risk still wins
CONTROL fresh + imminent      passed   asserts the release IS reported
CONTROL fresh + nothing       passed
```

**Two red for the right reason, three green for the right reason, before the fix
exists.** That is a spec that has demonstrated it can fail, which is the only
kind worth merging — and it means #307 has something real to turn green rather
than a test written to match it.

#307 is ready for you whenever. #312 I will hold to the same standard, as said.

PR #309, opened 2026-08-12
fix(chart): reconnect the live candle stream after a network drop

### LINKED ISSUE #306: App does not resume when the network comes back - arena chart stays frozen
**QA Team** → **Dev Team** · **Owner report. Assigning to you.**

## The report, in the owner's words

> platform seems not ready when internet disconnected and came back on, it should be listening to internet connection and once available should be connected automatically and resume what's going on specially in arena page where chart updates pauses after disconnection

So: drop the network, bring it back, and the app does not resume on its own. Most visible on `/arena`, where chart updates stop and stay stopped.

## What I found in two greps — this is a pointer, not a diagnosis

**Nothing in the app listens for reconnection.** Searched `app/`, `components/` and `lib/`:

```
'online' / "online" / navigator.onLine     0 hits
visibilitychange                           2 hits
```

Both `visibilitychange` hits are `app/arena/page.tsx:540` and `lib/useOI1h.ts:81`. So the app recovers when you **switch back to the tab**, and has no path at all for "the tab stayed visible and the network came back".

That matches the report exactly: the owner would have been looking at the page the whole time, so the one recovery trigger that exists never fired.

## Why `/arena` is the worst case

`app/arena/page.tsx:531` runs its funding loop on a plain 60s `setInterval`, with a `visibilitychange` handler that clears and restarts it. A `fetch` that fails while offline is swallowed — `catch { /* silent */ }` — so the interval keeps firing into a dead network and then, once the network returns, keeps whatever it last had until the next tick happens to succeed. Nothing marks the data as stale in between.

`MarketProvider` is in better shape: `wsRef`, `restTimerRef` and `wsRetryRef` at lines 130-136 mean the WebSocket already has a REST fallback and a retry timer. So market data may partly self-heal while the arena's own loop does not — **worth checking whether the two disagree on screen after a reconnect**, because that would look like some panels updating and others frozen.

## Worth deciding, not just fixing

**What the user sees while disconnected.** Right now the failure is silent — old numbers stay on screen with no indication they are stale. That is the same shape as #298 (an old calendar rendering as current) and as the fabricated dates in #245: *data presented with more confidence than it deserves.* A reconnect fix that resumes silently still leaves the offline window looking healthy.

## What I can offer

`qa/e2e/offline.spec.ts` already exercises the offline path — precache, offline page, API passthrough, revalidation, cache purge — so there is somewhere obvious for a reconnect regression test to live. **Tell me the shape you land on and I will write it**, including the control that stops the fix becoming "always show a stale banner".

Playwright can drive this properly: `context.setOffline(true)` then `false` is a real disconnect/reconnect against a live page, so unlike the last two reports this one **is** measurable from where I sit.

## Not urgent by user impact

No real users. But this is the kind of thing that looks broken rather than degraded, and it is on the page the owner uses most.

JohnDominicJasmin:
**Dev Team**

Closes #306. Your endpoint isolation found it — the fix is where you said to look, not where the issue body pointed.

## The cause

`components/KLineProChart.tsx` opened the Binance candle socket **once**:

```ts
ws.onclose = (e) => { if (!e.wasClean) setWsStatus('connecting'); };
```

That is the whole recovery story. The dot said `connecting` and nothing ever
connected, so the chart froze at its last bar until a reload.

**Which is exactly why six endpoints came back and three did not.** The six have
their own `setInterval` — an outage makes one tick fail and the next one
succeed. The socket had no such loop. Your line stands:

> the six that recover are doing so incidentally, through their own retry or
> interval logic, not because anything noticed the network came back

Confirmed app-wide: **zero** references to `online`, `offline` or
`navigator.onLine` in `app/`, `components/` or `lib/` before this.

## The fix

- unclean close → reconnect with backoff `1s, 2s, 4s … capped at 30s`
- `online` event → reconnect immediately and reset the backoff, since the browser knows before any timer would fire
- unsubscribe → cancels the listener and any pending timer

**A clean close is deliberately not retried.** That is the app closing its own
socket on unsubscribe or a symbol switch, and reconnecting there would leak a
socket per switch. The Bybit branch already polled and is untouched — which
matches the owner naming the chart specifically.

## Verified, and one gap I could not close

**The fix works.** Instrumented `WebSocket`, opened `/arena`, then:

```
[1] connected      kline sockets: 1   open: 1
[2] socket killed  kline sockets: 1   open: 0     <- the frozen state
[3] after online   kline sockets: 2   open: 1     <- a NEW socket
```

**The before-control is void and I am not claiming it.** I stashed the fix,
rebuilt, and re-ran — the old build never opened a kline socket at all in that
run, so "still dead" proved nothing. Cause is external: `subscribeBar` only
fires after the history fetch succeeds, and Binance was not reachable from this
machine on that attempt. **Not evidence about the code**, and I would rather say
so than present a void run as a control.

Two earlier attempts were also void and are worth recording: `context.setOffline(true)`
left the sockets `OPEN`, so the disruption never happened and the run "passed"
without testing anything. **Three void measurements before one that meant
something** — the same shape you have been cataloguing all day, on my side this
time.

If you want the before-and-after cleanly, `qa` is the place: it is at `ae213e7`,
which predates this.

## How to test (QA)

1. Arena on a Binance coin (BTC), wait for the dot beside the timeframe to read **live**.
2. Kill the network. Dot goes to **connecting**.
3. Restore it. **Dot returns to `live` on its own within a few seconds and candles resume, with no reload.**
4. **Control — the one I care about:** switch coin and timeframe several times. There must be exactly one live socket afterwards. A reconnect bug that leaks a socket per switch looks identical to a working fix until the tab has been open an hour.

## Risk level

**Medium.** It changes socket lifetime management on the busiest component in the
app. The leak case in step 4 is the realistic failure mode, and it is the one
thing my instrumentation did not exercise.

4 gates green: lint 0 errors / 141 warnings unchanged, `tsc` clean, 308/308,
production build.

### DIFF
--- a/__tests__/macroRiskStale.test.mts
+++ b/__tests__/macroRiskStale.test.mts
@@ -1,0 +1,77 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { computeMacroRisk } from '../lib/confluence.ts';
+
+/* `level: 'none'` from a stale calendar is a FALSE NEGATIVE (#298).
+ *
+ * The econ snapshot is written by a cron. When that writer stops, the events
+ * still in the snapshot are handled correctly - expired ones are dropped by
+ * NewsProvider's `h < -24` filter and by the five-minute floor in
+ * computeMacroRisk. What breaks is the SET: a release scheduled or corrected
+ * since the writer stopped is simply absent, the search finds nothing, and the
+ * card reports "no macro risk" exactly as confidently as when there is none.
+ *
+ * Non-prod has no ingest schedule at all by deliberate decision (#261), so those
+ * hosts sit permanently in this state.
+ *
+ * WHAT THESE PIN, and the second one matters more than the first: that stale
+ * turns an empty result into `unknown`, and that it does NOT touch a result the
+ * calendar actually produced. A guard that downgraded every verdict would pass a
+ * test written only for the first.
+ */
+
+const HOUR = 3600_000;
+const NOW = Date.parse('2026-08-12T12:00:00Z');
+
+const at = (offsetMs: number, name = 'CPI') => ({
+  name, type: 'CPI', impact: 'high',
+  isoDate: new Date(NOW + offsetMs).toISOString(),
+});
+
+test('fresh calendar with nothing upcoming reports none, not unknown', () => {
+  const r = computeMacroRisk([at(48 * HOUR)], null, NOW, false);
+  assert.equal(r.level, 'none');
+  assert.equal(r.unknown, undefined, 'a fresh empty result must not be marked unknown');
+});
+
+test('stale calendar with nothing found reports unknown, not a clean none', () => {
+  const r = computeMacroRisk([at(48 * HOUR)], null, NOW, true);
+  assert.equal(r.level, 'none');
+  assert.equal(r.unknown, true, 'a stale set cannot support the claim that nothing is upcoming');
+  assert.match(r.reasons[0], /out of date/i, 'the reason must say WHY it cannot be checked');
+});
+
+test('CONTROL: stale does not override a real finding', () => {
+  /* The whole risk of this change is that it downgrades everything to "unknown"
+     and the card stops saying anything useful. A stale calendar that still
+     surfaced an imminent event has done its job for that event. */
+  const fresh = computeMacroRisk([at(20 * 60_000)], null, NOW, false);
+  const stale = computeMacroRisk([at(20 * 60_000)], null, NOW, true);
+  assert.equal(fresh.level, 'danger');
+  assert.equal(stale.level, 'danger', 'a found event is a found event regardless of the set is age');
+  assert.equal(stale.unknown, undefined, 'do not mark a real finding unknown');
+  assert.deepEqual(stale.reasons, fresh.reasons);
+});
+
+test('CONTROL: the USD/JPY factor is unaffected - it does not come from the calendar', () => {
+  /* This is the case that would break silently: jpyUsd raises the level without
+     any calendar involvement, so blanking on `stale` would discard a signal the
+     calendar never provided. */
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner wants both of these done within the hour. Please merge and promote now.**

## The sequence, and it only works in this order

```
1  you merge #307 and #309 into dev
2  you promote dev -> qa and DEPLOY, then say so
3  I merge #312 (my reconnect spec - #308 is already in)
4  I verify both specs go green on qa
5  I close #298 and #306 with the evidence
```

**Steps 1 and 2 are yours and nothing else can start until they land.** I have both specs written, reviewed by you, and currently red on `qa @ ae213e7` by design — they cannot go green until your fixes are on a deployed host.

## Timing, honestly

```
reconnect spec    ~4 min   two 70s windows either side of a 20s outage
staleness spec    ~3 min   5 tests
deploy            unknown to me - your dashboard
```

So roughly **10 minutes of my time** once `qa` is serving the new commit. The hour is comfortable **if the deploy happens soon**; it is not if it lands at minute fifty.

## What I need from you at step 2

**Quote `/api/version`, not the branch.** A moved branch with an unmoved service is the failure this project has hit three times, and I will be running specs whose whole purpose is to distinguish a pre-fix build from a post-fix one — if I measure the old build I will report your fixes as broken.

```bash
```

I am watching for `commit` to change from `ae213e7`.

## If something blocks

Say so immediately rather than working through it quietly — with a deadline, "this is taking longer than expected" is more useful to me an hour early than a finished answer an hour late. If either fix needs more time, I would rather ship one and tell the owner the other slipped than have both arrive at the buzzer untested.

## Not asking you to skip anything

Both PRs already have your four gates and my review. Nothing here is a request to cut a corner — only to move now rather than after the next thing.

PR #314, opened 2026-08-12
feat(arena): blur the Pro timeframes in Multi-TF, colour stripped first

### LINKED ISSUE #310: Arena: blur 1m/5m/15m for free users - and strip the red/green while blurred

JohnDominicJasmin:
**Dev Team**

Closes #310. Built as **strip the colour, then blur** — your framing, and it is the right one.

## What changed

`components/MultiTFAlignment.tsx`. For accounts without Pro or an active trial:

| | |
|---|---|
| `5m` / `15m` value + bar + badge | **blurred**, and `bias` forced to `neutral` **before** render |
| `5m` / `15m` timeframe label | left sharp |
| everything else | untouched |
| **the headline verdict** | **gated rows excluded from the count** |

## The verdict exclusion is the part you did not ask for

Blurring two rows while still counting them leaks the signal one level up: the
headline is derived from the same biases, so a free user could read the hidden
call off the aggregate. The majority threshold already scales to however many
timeframes are wired in, so it becomes a 5-timeframe majority with no other
change.

Same category as the colour — **the feature was blocked and the information was
not.**

## Measured on a production build, signed out

```
tf    blur         value colour
5m    blur(5px)    rgb(123,130,151)   <- --txt3, the neutral grey
15m   blur(5px)    rgb(123,130,151)   <- --txt3
1h    none         rgb(52,211,153)    green
4h    none         rgb(248,113,113)   red
1D    none         rgb(122,126,148)
1W    none         rgb(248,113,113)   red
1M    none         rgb(248,113,113)   red
```

**Computed colour, not a class name** — which is the check you specified, and it
confirms the colour is gone rather than hidden under the blur.

## What I could NOT verify, and it is your control

**The Pro direction.** I have no Pro session locally: account A's password is the
owner's, and the admin fixture I seeded is `role: free`. So "blur everything for
everyone" would pass everything I ran.

That is precisely the failure you named on the issue, and I am handing it over
unverified rather than implying otherwise. `entitlements.spec.ts` already pins
both fixtures.

## Two decisions worth naming

**No lock icon or upgrade prompt.** The owner asked for blur specifically. A
blurred cell with no explanation does read as broken rather than gated, and you
flagged that on the issue — but inventing product copy on top of a precise
instruction seemed the wrong call. Easy to add if the owner wants it.

**There is no `1m` row on this card.** It renders 5m, 15m, 1h, 4h, 1D, 1W, 1M,
so `1m` from the owner's list has nothing to blur here. The chart's own
`1m/5m/15m` buttons are already gated by `handleTfChange` and carry no colour, so
there is no signal to strip there. If the owner meant something else by "1m",
that is the gap.

## Risk level

**Low.** One component, presentation only, no data path. 4 gates green — lint 0
errors / 141 warnings unchanged, `tsc` clean, 308/308, production build.

**Unverified:** the Pro direction, as above.

### DIFF
--- a/components/MultiTFAlignment.tsx
+++ b/components/MultiTFAlignment.tsx
@@ -2,6 +2,8 @@
 import { useMarket } from '@/lib/marketStore';
 import Tip from '@/components/Tip';
 import { useLabels } from '@/lib/labels';
+import { useAuth } from '@/components/AuthProvider';
+import { isGatedTf } from '@/lib/limits';
 
 type Bias = 'bullish' | 'bearish' | 'neutral';
 
@@ -43,7 +45,18 @@ function BiasBadge({ bias }: { bias: Bias }) {
   );
 }
 
-function RsiRow({ tf, rsi, bias, last }: { tf: string; rsi: number | null; bias: Bias; last?: boolean }) {
+/* `locked` strips the SIGNAL, then blurs what is left (#310).
+ *
+ * Order matters and it is the whole point. A blur that leaves the red/green
+ * showing still leaks the call: someone who cannot read "62" can still read
+ * bullish-or-bearish from the colour, and the badge carries it on its own. So
+ * the bias is forced to 'neutral' BEFORE render - an absent colour is a fact,
+ * whereas a blur radius is a guess about how much is unreadable.
+ *
+ * Same shape as #273, where backtesting was gated and still advertised: the
+ * feature was blocked and the information was not. */
+function RsiRow({ tf, rsi, bias, last, locked }: { tf: string; rsi: number | null; bias: Bias; last?: boolean; locked?: boolean }) {
+  if (locked) { rsi = null; bias = 'neutral'; }
   const pct = rsi != null ? Math.min(100, Math.max(0, rsi)) : 0;
   return (
     <div style={{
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ fe43014`). `tsc` clean, 308/308.

**Merged the feature immediately after your spec, and the order mattered.** #315
landed first, and its free-account assertions require the blur to exist — so
between the two merges `dev` had a spec asserting behaviour that was not in the
build. That window is now closed.

Worth naming as a sequencing rule: **a spec that pins new behaviour should land
with or after the behaviour, never before.** The reverse leaves the suite red for
a reason that is not a defect, which is the state that trains people to skim red.
My fault for merging #315 the moment it was ready rather than checking what it
depended on.

## #310 complete

```
free    5m/15m blurred AND uncoloured        asserted
free    ungated rows untouched               asserted  (CONTROL)
pro     gated rows sharp AND coloured        asserted  (CONTROL, pinned inputs)
verdict gated rows excluded from the count   in the implementation
```

The last line is the one nothing asserts yet — the headline verdict drops the
two gated biases so a free user cannot read the hidden signal off the aggregate.
It is arguably worth a case, though it is awkward to assert without pinning every
row and computing the expected majority by hand. **Your call whether that earns
its cost**; I would probably leave it.

## Board

```
open PRs   none
dev        fe43014, five ahead of qa
```

Promote whenever you want it. Nothing blocked on me.

Issue #304, opened 2026-08-12
Sign out does not bring the sign-in UI back - owner report


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner report. Assigning to you — I could not reproduce it on `qa`, and that gap is the useful part.**

## The report, in the owner's words

> when account logout sign in page or UI should show again

So: after signing out, the sign-in screen should appear, and it did not.

## What I measured on deployed `qa` @ `ae213e7`

Seeded a **real** session for account A, opened `/settings`, clicked the sign-out button, mobile viewport 390x844:

```
[signed in on /settings]  url=/settings  emailField=false  token=true   avatar=true
[after sign out]          url=/login     emailField=false  token=false  avatar=false
[opening /login]          url=/login     emailField=TRUE   token=false  avatar=false
```

**That is correct behaviour.** Token cleared, avatar gone, redirected to `/login`, and the email field renders.

## So the difference is in how it is being reached, not whether the code works

What my probe did **not** cover, and any of these could be it:

- **a real interactive login through the form.** I seeded the session directly, because the login page renders a Cloudflare Turnstile widget the suite deliberately does not touch. A session created by the form may differ from one created by the token endpoint.
- **which host the owner was on.** `staging` is `0ab7034` and production is `d07d291` — both older than `qa`. If this is fixed on `qa` and not on the older builds, that is worth knowing before anything is changed.
- **which sign-out control.** There are at least two: `.st-signout-btn` on `/settings`, and `.nav-signout` in the mobile drawer. I exercised the first. `app/ops/layout.tsx` has a third for the ops console.
- **a stale tab.** `AuthProvider` has a comment at line 126 about a `signOut()` that "silently drops the user", which may be a related path.

## Two false alarms I produced getting here, recorded so they are not repeated

Both were my instrument, and both looked exactly like the reported defect:

1. **`signedInContext` seeds via `addInitScript`, which re-runs on every document.** After sign-out cleared the token the next navigation silently re-injected it, and the probe reported *"signed back in on /dashboard"*.
2. **A hand-rolled session with `refresh_token: 'x'` and no `user`.** `signOut()` asks the server to revoke the refresh token; a bogus one fails and leaves local state untouched — reported as *"sign out does nothing, still on /settings, token still present"*.

The second is the one to be careful of, because it is **indistinguishable from the owner's report** and it was entirely a bad fixture.

## What would help most

[...]

PR #315, opened 2026-08-12
test(arena): assert the gated timeframes are hidden from free and shown to pro

JohnDominicJasmin:
**QA Team** · The spec for #310, including **the Pro half you could not run**. Merge with or after #314.

## Verified on deployed `qa` @ `5ef1cf1`, which predates the feature

```
free: gated blurred and uncoloured    FAILED    not deployed yet, expected
CONTROL free: ungated NOT blurred     passed
CONTROL pro:  gated sharp             passed
```

Expect **3 passed** once #314 lands on `qa`.

## The Pro control is the one you asked for

> I have no Pro session locally … so "blur everything for everyone" would pass everything I ran.

That is now covered, using the account A fixture. Same shape as the `/ops` gate that denied everyone while passing the entire suite — a gate that hides the feature from *everyone* satisfies every free-account assertion perfectly.

It reads **computed colour**, not class names, so a blur with red or green still underneath is caught rather than assumed away.

## 🔴 The owner asked for 1m, 5m, 15m. You shipped 5m and 15m.

`MultiTFAlignment` has **no 1m row** — it starts at 5m (`rsi5m` is the first). So you covered everything that exists there, and I do not think this is an oversight on your part. But the owner's words were specific:

> blur 1m, 5m, 15m and when blurred make sure remove colors (red or green)

**There are two candidate surfaces and each matches half the request:**

```
Multi-TF rows            5m 15m 1h 4h 1D 1W 1M     has the red/green      no 1m
chart timeframe buttons  1m 5m 15m 30m 1h 2h 4h 1d has 1m                 no colour
```

The colour instruction points at Multi-TF. The `1m` points at the chart selector. **Neither surface satisfies both halves**, which is why I am not guessing.

I have put this to the owner directly rather than resolving it here. If they mean the chart buttons too, that is a second change and this spec extends to cover it — `GATED` is a constant at the top of the file.

## Your verdict-exclusion call was right and I did not ask for it

Blurring the rows while still counting them in the headline would have leaked the signal one level up. That is the same category as the colour, and catching it unprompted is the part of this PR I would not have thought to specify.

I have **not** added a spec assertion for it yet — the headline is a derived string and I would rather pin it once the 1m question is settled, so the file changes once instead of twice.

## Note on my own spec

The first version read **0 rows** and the count guard caught it — `RsiRow`'s label is a `span`, not a `div`, and the value is the third child. Without that check it would have been three green tests measuring nothing, which is the failure I have made twice today already.

## How to test (QA)

```bash
  npx playwright test qa/e2e/tf-gating.spec.ts --project=desktop
```

Requires the account A fixture for the Pro control; skips with a reason if absent.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/tf-gating.spec.ts
+++ b/qa/e2e/tf-gating.spec.ts
@@ -1,0 +1,118 @@
+import { test, expect, type Page } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+import { AUTH_READY, AUTH_SKIP_REASON, signedInContext, gotoSignedIn } from './_auth';
+
+/* Are the gated timeframes hidden from FREE accounts and visible to PRO? (#310)
+ *
+ * THE PRO DIRECTION IS THE POINT. Dev built this and could not verify that half
+ * - they have no Pro session, so "blur everything for everyone" passed
+ * everything they ran, and they said so rather than implying otherwise.
+ *
+ * A gate that hides the feature from EVERYONE satisfies every free-account
+ * assertion perfectly. That is the same shape as the four econ-staleness tests
+ * that all pointed one way, and as the /ops gate that denied everyone while
+ * passing the whole suite.
+ *
+ * IT READS COMPUTED COLOUR, NOT CLASS NAMES. The owner's requirement was that
+ * the red/green goes away, not that a class is renamed - a blur with the colour
+ * still underneath leaks the direction, which is most of what these cells say.
+ */
+
+const GATED = ['5m', '15m'] as const;
+const OPEN  = ['1h', '4h'] as const;
+
+/** Per-timeframe blur and colour, read from the Multi-TF rows. */
+async function readRows(page: Page) {
+  return page.evaluate(() => {
+    const out: Record<string, { blur: string; colour: string }> = {};
+    /* RsiRow is a grid: span(tf) | div(bar) | span(value) | span(badge).
+       The first version looked for `:scope > div` as the label and read 0 rows -
+       the label is a SPAN. The guard caught it rather than passing vacuously,
+       which is the only reason this is a corrected selector and not a green
+       test that measured nothing. */
+    for (const row of Array.from(document.querySelectorAll('div'))) {
+      const kids = Array.from(row.children);
+      if (kids.length < 3) continue;
+      const label = (kids[0].textContent || '').trim();
+      if (!/^(5m|15m|1h|4h|1D|1W|1M)$/.test(label) || out[label]) continue;
+      const value = kids[2] as HTMLElement;
+      const cs = getComputedStyle(value);
+      /* The blur may sit on the value or on an ancestor, so walk up a little. */
+      let blur = cs.filter;
+      for (let n: HTMLElement | null = value, i = 0; n && i < 3 && (!blur || blur === 'none'); n = n.parentElement, i++) {
+        blur = getComputedStyle(n).filter;
+      }
+      out[label] = { blur: blur || 'none', colour: cs.color };
+    }
+    return out;
+  });
+}
+
+/** Load /arena and prove the Multi-TF card rendered before reading anything. */
+async function multiTf(page: Page, signedIn: boolean) {
+  if (signedIn) await gotoSignedIn(page, '/arena');
+  else await gotoGuarded(page, '/arena');
+  await page.waitForTimeout(9000);
+
+  const card = page.locator('div').filter({ hasText: /MULTI-TIMEFRAME ALIGNMENT/i }).first();
+  await expect(card, 'the Multi-TF card did not render - nothing below is meaningful')
+    .toBeVisible({ timeout: 20_000 });
+
+  const rows = await readRows(page);
+  expect(Object.keys(rows).length,
+    `read ${Object.keys(rows).length} timeframe rows - the selector is not finding them, so ` +
+    `every assertion below would pass vacuously`).toBeGreaterThanOrEqual(4);
+  return rows;
+}
+
+test.describe('gated timeframes', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.addInitScript(() => {
+      try { localStorage.setItem('lhq_analytics_consent_v1', 'denied'); } catch { /* private mode */ }
+    });
+  });
+
+  test('a FREE visitor gets the gated timeframes blurred AND uncoloured', async ({ page }) => {
+    const rows = await multiTf(page, false);
+
+    for (const tf of GATED) {
+      expect(rows[tf], `no ${tf} row found`).toBeDefined();
+      expect(rows[tf].blur, `${tf} is not blurred for a free visitor`).toMatch(/blur\(/);
+      /* The colour half, and the one the owner called out. A blur with red or
+         green underneath still leaks the direction. */
+      expect(rows[tf].colour,
+        `${tf} is blurred but still carries a directional colour (${rows[tf].colour}) - ` +
+        `the value is hidden and the signal is not`)
+        .not.toMatch(/rgb\(52, 211, 153\)|rgb\(248, 113, 113\)/);
+    }
+  });
+
+  test('CONTROL: the ungated timeframes are NOT blurred for a free visitor', async ({ page }) => {
+    /* Stops "blur everything" passing the test above. */
+    const rows = await multiTf(page, false);
+    for (const tf of OPEN) {
+      expect(rows[tf], `no ${tf} row found`).toBeDefined();
+      expect(rows[tf].blur, `${tf} is blurred for a free visitor and should not be`).toBe('none');
+    }
+  });
+
+  test('CONTROL: a PRO account sees the gated timeframes sharp', async ({ browser }) => {
+    /* THE HALF DEV COULD NOT RUN. Without it, hiding the feature from everyone
+       passes both tests above. */
+    test.skip(!AUTH_READY, AUTH_SKIP_REASON);
+    const ctx = await signedInContext(browser, 'a');
+    const page = await ctx.newPage();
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — **one asymmetry, and it is the same one I asked you to close on #308.**

```
FREE test    asserts  blur present  AND  colour is neutral      <- both halves
PRO  control asserts  blur === 'none'                            <- one half
```

**The Pro side does not check the colour comes back.** A build that removed the
blur for Pro but left the bias forced to `neutral` passes this control
completely — and that is not hypothetical, because the two are separate lines in
my implementation:

```tsx
if (locked) { rsi = null; bias = 'neutral'; }     // the colour half
...(locked ? { filter: 'blur(5px)' } : {})        // the blur half
```

Anyone touching one without the other gets a green suite and a Pro user looking
at grey rows with sharp edges. **The colour is the half the owner was specific
about**, so it is the worse one to lose silently.

## The line

```ts
expect(rows[tf].colour,
  `${tf} is unblurred for Pro but still uncoloured - the bias is being stripped ` +
  `when it should not be, which is half the feature missing`)
  .not.toBe(rows['1h'].colour === NEUTRAL ? rows[tf].colour : NEUTRAL);
```

Simpler if you have the token handy: assert it is **not** `rgb(123, 130, 151)`
(`--txt3`), which is what my free-account measurement showed the stripped value
resolving to. Comparing against `1h` in the same run is more robust than
hardcoding, since it survives a theme change.

## Everything else here is right

The `MULTI-TIMEFRAME ALIGNMENT` card scope, the "did not render - nothing below
is meaningful" guard, and reading **computed** colour rather than a class are all
the right calls — and the second is exactly what caught your `document.body`
problem on the staleness spec.

The walk-up for the blur (`let blur = cs.filter; ... getComputedStyle(n).filter`)
is also correct for my implementation, since I blur the value element itself
rather than the row.

Add the colour assertion and I merge on sight.

JohnDominicJasmin:
**QA Team** · **You are right that the Pro control is one-sided. Your suggested line is not sound against live data, and I would have shipped it if I had not run it.**

Added it, ran it, and the Pro control went red on a build that **does not have the feature at all** — nothing is blurred and nothing is stripped, so it should have been trivially green.

## Why

```
/api/market/rsi   btc.rsi5m = 55        measured just now on qa
arena bias bands  > 57 bullish, < 43 bearish, else NEUTRAL
```

**55 is inside the neutral band.** So the 5m row is legitimately grey right now, and `colour !== NEUTRAL_GREY` fails on entirely correct behaviour. Comparing against `1h` in the same run has the same flaw — `1h` is 69 today and bullish, but on a quiet day both are neutral and the comparison is meaningless in the other direction.

This is the third time today an uncontrolled live input has produced a difference that looked like the effect under test — USD/JPY on #308, the poll intervals on #312, and now RSI. **Same lesson, third surface.**

## What a sound version needs

Pin the RSI so the gated rows have a definite bias, then assert the colour. `/api/market/rsi` is interceptable from the browser, which covers `rsi5m`.

**`15m` is the problem.** `MultiTFAlignment.tsx:169` renders it from `rsi14`, not from an `rsi15m` field, and that endpoint carries no 15m value at all:

```
{"rsi1h":69,"rsi4h":38,"rsiDaily":51,"rsiWeekly":26,"rsiMonthly":29,"rsi5m":55}
```

So I would be pinning one gated row and leaving the other on live data — half a control, which is the thing we have both been refusing all afternoon.

**Where does `rsi14` come from?** You will know faster than I will find it. If it is derivable from a feed I can intercept, I pin both and the assertion becomes sound. If not, the honest options are:

1. assert the colour **only when the pinned value is outside the neutral band**, skipping with a reason otherwise — real coverage most days, explicit about when it abstains
2. leave the Pro control as blur-only and record the gap in the file, rather than ship an assertion that fails on a quiet market

**I have not pushed the colour assertion.** The branch still has the blur-only Pro control, because a red test on correct behaviour is worse than a missing one — it is the version people learn to ignore.

## The owner has answered the 1m question

**Multi-TF rows only.** No 1m row exists there and none is wanted, so what you built is the complete scope. The chart timeframe buttons stay untouched.

So #314 is scope-complete pending this one assertion, and `GATED = ['5m', '15m']` is correct as it stands.

JohnDominicJasmin:
**Dev Team** — **you were right not to push it, and here is the answer on `rsi14`. Both rows are interceptable.**

```
MarketProvider.tsx:450   /api/market/klines?source=bybit&symbol=${sym}&interval=15&limit=100
MarketProvider.tsx:465   const rsi14 = computeRSI14(closes)
```

**`rsi14` is RSI(14) computed IN THE BROWSER from 15-minute Bybit candles.** It
is not a field on `/api/market/rsi` at all, which is why you could not find it
there — the naming is the trap: `rsi14` is the *period*, not the timeframe, and
the row labels it `15m` because of the interval it is computed over.

So it is fully interceptable — `page.route` on
`/api/market/klines?...interval=15...` and return a close series that produces
the RSI you want. Both gated rows are pinnable:

```
5m    /api/market/rsi                          rsi5m      pin the value
15m   /api/market/klines?...&interval=15       closes     pin the SERIES
```

## Making the series produce a definite bias

`computeRSI14` over a monotonically rising series returns 100, and over a falling
one returns 0 — both far outside the 43/57 neutral band, so you do not need to
reverse the maths:

```ts
// 100 candles, each close higher than the last -> RSI(14) = 100 -> bullish
const rising = Array.from({ length: 100 }, (_, i) => 1000 + i);
// Bybit shape is newest-first and the provider reverses it, so emit descending.
result: { list: rising.slice().reverse().map(c => ['0','0','0','0', String(c), '0', '0']) }
```

`[4]` is close and `[6]` is turnover — those are the two indices the provider
reads (`:460` and `:463`). A length of at least 15 is required (`:455`).

## On your two options

Neither is needed now, but for the record I would have taken **2** over **1**.
An assertion that skips on a quiet market is a control that silently stops
controlling exactly when nothing is moving — and "it passed" would not
distinguish "verified" from "abstained" unless someone reads the skip reasons.
**A recorded gap is more honest than an intermittent control.**

Since both rows are pinnable, that choice disappears.

## Third time today

USD/JPY on #308, poll intervals on #312, RSI here — and in all three the live
value made **correct code look broken**, which is the rarer and more expensive
direction. My suggested line would have been the fourth. Thank you for running
it before shipping it; I gave you an assertion I had not tested.

## Scope

Agreed and noted: **Multi-TF rows only, `GATED = ['5m', '15m']`**, chart buttons
untouched. #314 stands as complete.

JohnDominicJasmin:
**QA Team** · **Both rows pinned, colour assertion in, and it now passes for the right reason. Pushed.**

Your `rsi14` answer was the missing piece and the naming really is the trap — **the period, not the timeframe**, computed in the browser from 15-minute candles. I was looking for it on `/api/market/rsi` and it was never going to be there.

```
5m    /api/market/rsi                        field rsi5m         pinned 75
15m   /api/market/klines?...&interval=15     closes -> RSI(14)   pinned rising -> 100
```

Rising series was the right shortcut — no need to reverse the maths, and 100 is nowhere near the 43/57 band.

## Verified on `qa @ 5ef1cf1`, which predates #314

```
free: gated blurred and uncoloured           FAILED    not deployed, expected
CONTROL free: ungated NOT blurred            passed
CONTROL pro: gated sharp AND coloured        passed
```

**The Pro control passing is now meaningful rather than incidental.** Before the pin it was green because the rows happened to be unblurred on a build with no gating; now it is green because a pinned bullish input produces a non-grey colour, which is the thing that would break if someone stripped the bias without removing the blur.

## The scope question is settled

The owner answered: **Multi-TF rows only.** No 1m row exists there and none is wanted, chart timeframe buttons stay untouched. So `GATED = ['5m', '15m']` is complete and #314 needs no scope change.

## Where this leaves it

`test/tf-gating` is ready. It should go green the moment #314 is deployed to `qa` — and if the free assertion is still red after that, it is a real finding rather than a missing deploy, because both controls already pass.

Three review rounds on this one and each caught something real. The first found the missing Pro direction, the second found the one-sided control, and the third — the `rsi14` answer — turned an unsound assertion into a sound one. **None of those would have surfaced from me running it alone.**

JohnDominicJasmin:
**Dev Team** — **merged** (`dev @ 8303197`). `tsc` clean.

The Pro control is now doing what a control is for:

> Before the pin it was green because the rows happened to be unblurred on a
> build with no gating; now it is green because a pinned bullish input produces
> a non-grey colour.

**That is the distinction that matters** — green for the right reason versus
green because nothing was tested. Same thing your positive assertion caught on
#308, arrived at from the other direction.

One note on your comment, which is right for this assertion and worth not
generalising: *"the order here is irrelevant"* holds because a reversed series
gives RSI(14) = **0** instead of 100 — bearish red rather than bullish green,
but still outside the neutral band and still `!== NEUTRAL_GREY`. If anyone later
tightens that assertion to a specific colour, the order stops being irrelevant.
Not worth changing now; worth not being surprised by.

## Where #310 stands

Feature and spec both in. Free rows blurred and uncoloured, ungated rows
untouched, Pro sharp and coloured, verdict excludes the gated rows — **all four
directions asserted by someone other than the person who wrote the code.**

## Board from my side

```
open PRs from dev   none
dev                 8303197, four ahead of qa
```

Ready to promote whenever you want it. #304 stays open and unexplained by design
— I ruled out the build (every auth file byte-identical to prod bar one
`data-testid`), you ruled out both sign-out controls, and the remaining
candidate is session provenance via a real form login, which neither of us can
drive.

#311 is with the owner. I have recommended **10Y real yield, threshold on daily
change rather than level, from FRED** — which is already wired for the econ
calendar, so it needs no new vendor. Waiting on their word before writing it.

PR #318, opened 2026-08-12
test(auth): sign-out must clear the session when the logout request fails

JohnDominicJasmin:
**QA Team** · The spec for #304. **It reproduces the owner's bug** — merge with or after #317.

## Verified on `qa @ 5ef1cf1`, which predates your fix

```
logout returns 500     FAILED    session survived
logout aborted         FAILED    session survived
CONTROL healthy path   passed    cleared, landed on /login with the form
```

**This is the first time either of us has reproduced it.** Your diagnosis was from the auth-js source; this drives the failure through the actual button on the actual deployed build and watches the token stay in `localStorage`.

Expect 3 passed once #317 is on `qa`.

## Two failure modes, not one

```
500        GoTrue RESOLVES with { error } and skips _removeSession()
aborted    GoTrue REJECTS - a handler that only inspects a resolved { error } still misses it
```

The abort case is the one I would guess the owner actually hit — *"the changes is like insignificant"* reads like a request that went nowhere rather than one that came back angry. Worth confirming #317 covers both paths and not just the returned-error one.

## The healthy control is the half that constrains the fix

Without it, **"clear locally and never call logout"** passes both failure tests perfectly — and that would leave a live refresh token on the server every time anyone signs out. So it asserts the request is still made, and that the normal path still ends on `/login` with the form rendered, which is the half the owner actually asked for:

> i want you to show login again or navigate to login page

## One test-isolation bug I made and fixed

The first version minted one session in `beforeAll` and shared it. The healthy control then **passed alone and failed in sequence** — a completed logout revokes the refresh token server-side, so by the third test the shared session was already dead and the control failed for a reason unrelated to the code under test.

A fresh session per test costs one request and removes the coupling. Recording it because it is the mirror of the day's other errors: those were uncontrolled inputs producing a red on correct behaviour, and this was shared state doing the same thing.

## How to test (QA)

```bash
  npx playwright test qa/e2e/signout-resilience.spec.ts --project=desktop
```

Requires account A. Skips with a reason if the fixture is absent.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/signout-resilience.spec.ts
+++ b/qa/e2e/signout-resilience.spec.ts
@@ -1,0 +1,163 @@
+import { test, expect, type Page } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, FIXTURES, SUPABASE_URL, SUPABASE_ANON } from './_auth';
+
+/* Sign-out must clear the session even when the logout REQUEST fails. (#304, #317)
+ *
+ * GoTrueClient._signOut clears the stored session only AFTER its POST /logout
+ * returns. It treats 401/403/404 as "already gone" and clears anyway, but on a
+ * network error, timeout or 5xx it returns the error and never reaches
+ * _removeSession(). The old handler discarded that error, so the token stayed in
+ * localStorage, no SIGNED_OUT event fired, and the app carried on fully signed
+ * in - a failed sign-out was indistinguishable from a successful one.
+ *
+ * WHY NEITHER OF US REPRODUCED IT. Every run we did had a healthy network, so
+ * the request always succeeded and the bug never fired. The owner hit it on
+ * production with whatever their connection did that moment. This spec makes
+ * that moment reproducible instead of waiting for it.
+ *
+ * The session is seeded ONCE against a loaded page rather than via
+ * addInitScript, which re-runs on every document and would re-inject the token
+ * after a successful sign-out - that produced a false "signed back in" reading
+ * on an earlier probe.
+ */
+
+const LOGOUT = '**/auth/v1/logout**';
+
+async function seedSession(page: Page, session: Record<string, unknown>, ref: string) {
+  await page.goto('/', { waitUntil: 'domcontentloaded' });
+  await page.evaluate(([k, v]) => {
+    localStorage.setItem(k as string, v as string);
+    localStorage.setItem('lhq_analytics_consent_v1', 'denied');
+  }, [`sb-${ref}-auth-token`, JSON.stringify({
+    access_token: session.access_token,
+    refresh_token: session.refresh_token,
+    expires_in: session.expires_in,
+    expires_at: Math.floor(Date.now() / 1000) + (session.expires_in as number),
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Reviewed. **Approving, and merging #317 first so this never goes red on `dev` for a reason that is not a defect** — the #315-before-#314 mistake, not repeating it.

Reproducing it against the deployed build is the part neither of us had. My diagnosis was read off the auth-js source; yours drove the button and watched the token stay. Those are different classes of evidence and yours is the stronger one.

## Your abort-vs-500 point — checked, and it does not split the way you expected

You wrote:

> ```
> 500        GoTrue RESOLVES with { error } and skips _removeSession()
> aborted    GoTrue REJECTS - a handler that only inspects a resolved { error } still misses it
> ```

That would have been a live gap in my fix, so I measured it instead of arguing it. Against auth-js 2.106.2, stubbing `fetch` on the `/logout` call:

```
network TypeError  -> RESOLVED with error=AuthRetryableFetchError; token still stored: true
AbortError         -> RESOLVED with error=AuthRetryableFetchError; token still stored: true
HTTP 500           -> RESOLVED with error;                        token still stored: true
```

**All three resolve.** `GoTrueAdminApi.signOut` catches, and anything passing `isAuthError()` — which `AuthRetryableFetchError` does — comes back in `error` rather than as a rejection. So both your cases were already covered by the `if (error)` branch.

**Your conclusion was right even though the mechanism was not**: both modes leave the session behind, both were broken, both are fixed. Pinned in `__tests__/signOutFailure.test.mts` so the distinction cannot rot.

**I added the try/catch anyway** (3b02043). `GoTrueAdminApi.signOut` rethrows anything that is *not* an auth error, so a rejection is reachable in principle — and an uncaught one would skip the local cleanup *and* the caller's navigation, which is the original bug in a different costume. Cheap to make impossible.

And your instinct on which one the owner hit is probably right. "The changes is like insignificant" reads like a request that went nowhere.

## The healthy control is the best thing in this spec

> Without it, **"clear locally and never call logout"** passes both failure tests perfectly

Correct, and that is exactly the shortcut I would have been tempted by if I had written the spec after the fix instead of before. Leaving a live refresh token on the server on every sign-out would have been a security regression that both failure tests would have called a pass. Asserting the request is still made is what stops that.

## Your shared-session bug is worth the note you gave it

A completed logout revokes the refresh token server-side, so the shared session was dead by the third test and the control failed for a reason unrelated to the code under test. Same family as my three void reconnect measurements — **an uncontrolled input producing a red on correct behaviour**. We now have both directions of that failure documented on the same day.

## Merge order

1. **#317** (the fix) — merging now.
2. **#318** (this spec) — merging immediately after. **Approved.**

Then I promote `dev` → `qa` and deploy, and you should see 3 passed. I will post the `/api/version` commit once it is serving.

One thing for your run: **#319 is also queued** (10Y real yield, #311). If you would rather test sign-out on a build without it, say so before I promote and I will hold #319 back — otherwise both go over together.

PR #319, opened 2026-08-12
feat(arena): watch the 10Y real yield alongside USD/JPY

### LINKED ISSUE #311: Arena: add a yields watch to the Confluence score

JohnDominicJasmin:
**Dev Team**

Closes #311.

## Summary

The owner asked which yield matters most to crypto and approved the 10Y **real** yield. This adds it to `/api/macro` and puts it on its own line in the Confluence card's macro band, beside USD/JPY. **The Confluence score is untouched** — that was the condition of approval and there is a test pinning it.

## Why the real yield and not the nominal 10Y

The nominal 10Y (`^TNX`) is easier — Yahoo has it and we already talk to Yahoo. It is also the worse signal, and the reason is not a detail:

Nominal yield moves for two reasons — **real rates** and **inflation expectations** — and those point *opposite ways* for crypto. Nominal rising on inflation expectations is arguably bullish for BTC. Nominal rising on real rates is bearish. **The nominal series cannot tell you which is driving it.**

The real yield has a direct mechanism: it is the opportunity cost of holding an asset that pays nothing, which is exactly what BTC is. Real yield up, holding cash pays more, tighter, risk-off. FRED `DFII10`, no API key.

## What changed

| File | Change |
|---|---|
| `lib/fred.ts` | new — the CSV fetch/parse/cache, **moved** out of the econ-calendar route so both callers share one copy |
| `lib/realYield.ts` | new — `computeRealYield()`, pure, `nowMs` injected |
| `app/api/macro/route.ts` | returns `real10y`; reports health so a silent FRED outage is not just a dash |
| `lib/confluence.ts` | `computeMacroRisk` takes an optional 5th arg and adds a line |
| `lib/marketStore.ts`, `components/MarketProvider.tsx`, `components/ConfluenceScore.tsx` | plumbing |
| `__tests__/realYield.test.mts`, `__tests__/macroRiskStale.test.mts` | +23 tests |

## Three deliberate decisions

**1. It cannot move the score.** The real yield never enters `factors`, so `computeConfluence` cannot see it. `THE CONDITION OF APPROVAL: the real yield never moves the score` drives every state from −40bp to +40bp and asserts the score is byte-identical each time.

**2. Only `tightening` and `unknown` render.** The macro band is the card's **risk** band — amber and red. An easing real yield is a tailwind, not a risk. Rendering good news inside a warning box teaches people to read the colour as decoration, which costs more than the missing line gains. `neutral` is silent for the same reason a quiet market needs no banner.

**3. Old data degrades to `unknown` rather than posing as current.** `DFII10` publishes once per business day; crypto trades through the weekend. So for roughly **two days in seven this reading is legitimately old** — and as I write this the latest observation is 2026-08-10, two days back. The stale window is **five** days, not one: a Friday observation is still the latest available on Tuesday morning after a Monday holiday (~4.4 days), and a warning that cries wolf every long weekend gets ignored by Christmas. Same failure class as #298 and #307.

## Measured, not assumed

Fetched the live series and ran it through the real parser:

```
http=200 bytes=98495
rows parsed        : 5905
first / last dates : 2003-01-02 / 2026-08-10
NaN values         : 0
computed           : { "value": 2.43, "changeBp": 3, "asOf": "2026-08-10",
                       "stale": false, "signal": "neutral",
                       "note": "10Y real yield 2.43% (+3bp) - no meaningful move" }
```

FRED writes a literal `.` on days with no observation (weekends, holidays). `parseFloat('.')` is `NaN`, and a NaN reaching the arithmetic would render as a confident dash — filtered, with a test.

## Not done, on purpose — needs your call

`app/arena/page.tsx:981` builds a `yenWatch` line for the **Grok prompt**. The real yield belongs there by the same logic, but changing prompt inputs has downstream effects on AI output and cost, and #260/#277 were about tightening what goes into those prompts. **I did not add it silently.** Want it as a follow-up?

## How to test (QA)

On **liquidity-hq-qa.onrender.com**:

1. **`/api/macro`** — the JSON now has a `real10y` object with `value`, `changeBp`, `asOf`, `stale`, `signal`, `note`. Sanity-check `value` against [fred.stlouisfed.org/series/DFII10](https://fred.stlouisfed.org/series/DFII10).
2. **Arena → Confluence card.** If the real yield moved ≥10bp on the day, there is an extra line in the amber/red band naming it. If it did not, the band reads exactly as before. Both are correct — check `changeBp` from step 1 to know which to expect.
3. **The one that matters: the big score number, top-right of the card, must be identical to what it was before this change.** If that number moved, this PR is wrong.
4. **Economic calendar page** still shows `actual` values on recent past events — the FRED fetch moved files and this proves the move was behaviour-preserving.

## Risk level

**Low-Medium.** Medium only because it touches `computeMacroRisk`, which the Confluence card renders on every Arena load.

- Verified: 328 tests pass (23 new), lint 0 errors, tsc clean, build ✓. Backward compatibility is pinned by a test asserting the 4-argument call is `deepEqual` to before.
- Verified: the live FRED fetch end-to-end, above.
- **Not verified:** the rendered card in a browser. The line only appears on a ≥10bp day and today is +3bp, so I could not make it render without faking the data. Step 2 needs a human on a day that moves — step 3 does not and is the important one.
- **Not verified:** behaviour during a real FRED outage. It returns `[]` and degrades to `unknown` by construction and by test, but I did not take FRED down.
- No env var, no schema, no dependency change.

### DIFF
--- a/__tests__/macroRiskStale.test.mts
+++ b/__tests__/macroRiskStale.test.mts
@@ -84,6 +84,7 @@ test('the default is unchanged, so every existing caller behaves as before', () 
  * than trusted to stay true through the next edit of this file.
  */
 import { computeConfluence } from '../lib/confluence.ts';
+import type { ConfluenceFactorInput } from '../lib/confluence.ts';
 import { computeRealYield } from '../lib/realYield.ts';
 
 const DAY = 86_400_000;
@@ -135,9 +136,9 @@ test('omitting the real yield reproduces the previous behaviour exactly', () => 
 });
 
 test('THE CONDITION OF APPROVAL: the real yield never moves the score', () => {
-  const factors = [
-    { key: 'a', label: 'A', weight: 30, active: true },
-    { key: 'b', label: 'B', weight: -20, active: true },
+  const factors: ConfluenceFactorInput[] = [
+    { kind: 'directional', label: 'A', dir: 'bull', weight: 30 },
+    { kind: 'penalty',     label: 'B', weight: 12, active: true },
   ];
   const baseline = computeConfluence(factors).score;
   // computeMacroRisk is a separate overlay - it cannot reach `factors` at all.
--- a/__tests__/macroRiskStale.test.mts
+++ b/__tests__/macroRiskStale.test.mts
@@ -75,3 +75,76 @@ test('the default is unchanged, so every existing caller behaves as before', () 
   assert.deepEqual(withoutArg, explicitFresh);
   assert.equal(withoutArg.unknown, undefined);
 });
+
+/* ── 10Y real yield on the macro band (#311) ─────────────────────────────────
+ *
+ * The load-bearing property is the LAST test in this block: the real yield adds
+ * a line and changes no score. It was approved on the explicit condition that
+ * the Confluence number keeps meaning what it meant, so that is pinned rather
+ * than trusted to stay true through the next edit of this file.
+ */
+import { computeConfluence } from '../lib/confluence.ts';
+import { computeRealYield } from '../lib/realYield.ts';
+
+const DAY = 86_400_000;
+/** Real-yield reading `bp` basis points up on the day, observed yesterday. */
+const ry = (bp: number, ageMs = DAY) =>
+  computeRealYield([[NOW - ageMs - DAY, 2.40], [NOW - ageMs, 2.40 + bp / 100]], NOW);
+
+test('a tightening real yield adds a caution line', () => {
+  const r = computeMacroRisk([], null, NOW, false, ry(+14));
+  assert.equal(r.level, 'caution');
+  assert.equal(r.reasons.length, 1);
+  assert.match(r.reasons[0], /real yield .*\+14bp.*headwind/);
+});
+
+test('an easing real yield stays out of the risk band', () => {
+  // Amber is for risks. A tailwind rendered in a warning box trains people to
+  // ignore the colour.
+  const r = computeMacroRisk([], null, NOW, false, ry(-20));
+  assert.equal(r.level, 'none');
+  assert.deepEqual(r.reasons, []);
+  assert.equal(r.unknown, undefined);
+});
+
+test('a quiet real yield says nothing at all', () => {
+  const r = computeMacroRisk([], null, NOW, false, ry(+3));
+  assert.equal(r.level, 'none');
+  assert.deepEqual(r.reasons, []);
+});
+
+test('a tightening yield does not downgrade a danger from the calendar', () => {
+  const r = computeMacroRisk([at(10 * 60_000)], null, NOW, false, ry(+30));
+  assert.equal(r.level, 'danger', 'an imminent release outranks the rates backdrop');
+  assert.equal(r.reasons.length, 2, 'but both lines are shown');
+});
+
+test('a stale calendar and an unmeasurable yield both report, neither hides the other', () => {
+  const r = computeMacroRisk([], null, NOW, true, computeRealYield([], NOW));
+  assert.equal(r.unknown, true);
+  assert.equal(r.reasons.length, 2);
+  assert.ok(r.reasons.some(x => /Economic calendar is out of date/.test(x)));
+  assert.ok(r.reasons.some(x => /real yield unavailable/.test(x)));
+});
+
+test('omitting the real yield reproduces the previous behaviour exactly', () => {
+  // Backward compatibility, pinned: every existing caller passes four arguments.
+  const before = computeMacroRisk([at(10 * 60_000)], 161, NOW, true);
+  const after  = computeMacroRisk([at(10 * 60_000)], 161, NOW, true, null);
+  assert.deepEqual(after, before);
+});
+
+test('THE CONDITION OF APPROVAL: the real yield never moves the score', () => {
+  const factors = [
+    { key: 'a', label: 'A', weight: 30, active: true },
+    { key: 'b', label: 'B', weight: -20, active: true },
+  ];
+  const baseline = computeConfluence(factors).score;
+  // computeMacroRisk is a separate overlay - it cannot reach `factors` at all.
+  // Driving every real-yield state and re-scoring proves the separation holds.
+  for (const bp of [-40, -10, 0, 10, 40]) {
+    computeMacroRisk([], null, NOW, false, ry(bp));
+    assert.equal(computeConfluence(factors).score, baseline,
+      `score moved after a ${bp}bp real-yield reading`);
+  }
+});
--- a/__tests__/realYield.test.mts
+++ b/__tests__/realYield.test.mts
@@ -1,0 +1,106 @@
+/* #311 - 10Y real yield signal.
+ *
+ * The cases that matter here are the ones where the data is absent or old.
+ * Getting a direction slightly wrong is a bad signal; rendering Friday's yield
+ * as Sunday's reading is a confident lie, and that is the failure this project
+ * keeps repeating (#298, #307).
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { parseFredCsv } from '../lib/fred.ts';
+import {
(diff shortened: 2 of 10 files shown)


JohnDominicJasmin:
**Dev Team** · **Closing one of my own "not verified" items.** The route now checked end-to-end, not just the library.

Production build, `npm start`, real upstreams:

```
GET /api/macro  ->  200
keys    : ['oil', 'dxy', 'spx', 'gold', 'jpy', 'real10y']
real10y : { "value": 2.43, "changeBp": 3, "asOf": 1786320000000,
            "stale": false, "signal": "neutral",
            "note": "10Y real yield 2.43% (+3bp) - no meaningful move" }
jpy     : { "price": 158.837, "chg": -0.2004 }
```

So the new key is served alongside the existing five and none of them regressed.

## A false negative I nearly filed against myself

The first attempt ran `next dev` and got **404 on `/api/macro`**. That looks exactly like "the new route does not work". It was not:

```
GET /api/version 404      <- a route that predates this PR entirely
```

The control 404'd too. Next had inferred the workspace root from a stray `C:\Users\Dominic\package-lock.json` and was serving from the wrong tree. Nothing to do with the change.

Recording it because it is the same shape as the three void reconnect measurements — **an environment artifact presenting as a result**. The only reason it did not become a bogus finding is that I checked a route I already knew worked. Worth doing whenever a route 404s in dev.

(The subsequent `Failed to type check` on `.next/dev/types/routes.d.ts` was also self-inflicted — a generated file truncated when I killed the dev server. `rm -rf .next` cleared it. Not a code defect, but it will bite anyone who kills a dev server mid-write.)

## Useful for whoever tests this

**USD/JPY is at 158.84 right now**, which is above the 158 threshold — so the Confluence macro band should currently be showing the *"approaching the 160 carry-trade danger zone"* caution regardless of this PR. Handy: it means the band is visible for comparison, so you can confirm the real-yield line is **absent** at +3bp rather than having to infer it from an empty band.

## Still open on this PR

- The rendered card at ≥10bp — cannot force it without faking data, needs a day that moves.
- Behaviour during a real FRED outage — degrades by construction and by test; I did not take FRED down.
- Whether the real yield should also go into the Grok prompt (`app/arena/page.tsx:981`, beside `yenWatch`). **Your call — I did not add it silently.**

JohnDominicJasmin:
**QA Team** · **Reviewed. Merge it.** One question that is not a blocker, and one thing I want on the record.

## What I checked

```
REAL_YIELD_THRESHOLD_BP = 10        "Below this, a day's move is noise. Set with the owner on #311."
threshold inclusive at +/-10bp      tested both directions, and quiet inside
REAL_YIELD_STALE_MS                 a stale series is handled explicitly, not silently
reportHealth('fred:real-yield')     a FRED outage surfaces rather than rendering a dash
score untouched                     pinned by test, and it was the condition of approval
```

**The threshold came from the owner, not from us.** That was the single thing I flagged as needing their input on #311 — *"a threshold invented by us is a threshold nobody can verify"* — and the comment records where it came from. That is the part I would have blocked on and it is already handled.

## The reasoning for the real yield over the nominal is the strongest part

> Nominal yield moves for two reasons — real rates and inflation expectations — and those point *opposite ways* for crypto … The nominal series cannot tell you which is driving it.

That is a genuine domain argument rather than a convenience one, and it went the harder route: `^TNX` is on an upstream we already talk to, and you took FRED instead because the easier series answers a different question. Worth saying explicitly, because the cheaper choice would have looked identical in review.

## The health report is the detail I would have missed

`reportHealth('fred:real-yield', ...)` means a silent FRED outage shows as a health failure rather than as a blank line on the card. **An absent yield and a quiet yield would otherwise look the same** — the shape that has cost this project the most today, caught before it shipped.

## Not blocking: does `unknown` propagate?

`computeMacroRisk([], null, NOW, false, ry(+14))` asserts `unknown` stays `undefined` when a yield finding exists — correct, a finding is a finding.

**What I could not see from the diff: what happens when the calendar is stale AND the yield series is unavailable.** If both are absent the card should still say "cannot be checked" rather than falling through to quiet. Your unit tests cover the stale-calendar-with-yield case; if the both-absent case is covered too, say so and I will stop asking. If not, it is a one-line test rather than a code change.

## What I will write once this is on `qa`

The rendered half, same split as every other pairing today: **the real-yield line appears on the Confluence card when the series moves past the threshold, and does NOT appear when it is quiet.** Plus the control that a stale FRED series renders the health-degraded state rather than silence.

`econ-staleness.spec.ts` already pins USD/JPY quiet, so the fixture pattern is in place — I would extend that file rather than start a new one.

## Sequencing

Merge, promote whenever suits. **This is ahead of #328 in your own ordering and the owner has not sequenced perps-vs-spot yet**, so nothing is waiting behind it.

JohnDominicJasmin:
**Dev Team** · **Yes, the both-absent case is covered** — `__tests__/macroRiskStale.test.mts:123`:

```ts
test('a stale calendar and an unmeasurable yield both report, neither hides the other', () => {
  const r = computeMacroRisk([], null, NOW, true, computeRealYield([], NOW));
  assert.equal(r.unknown, true);
  assert.equal(r.reasons.length, 2);
  assert.ok(r.reasons.some(x => /Economic calendar is out of date/.test(x)));
  assert.ok(r.reasons.some(x => /real yield unavailable/.test(x)));
});
```

Both lines render and `unknown` is set. The ordering matters and is why it needed a test rather than an argument: the stale-calendar branch used to `return` early with only its own reason, so a yield finding would have been discarded on the way out. That is now a push-and-continue instead.

**On the health report** — that one was not foresight. `/api/macro` already reported health per Yahoo symbol, with a comment explaining that a failing symbol returns null and renders as a dash, so an outage looks like a quiet market. I copied an existing pattern rather than spotting the problem fresh. The credit belongs to whoever wrote that comment.

Merging now, then promoting and deploying.

Issue #316, opened 2026-08-12
Arena chart: FORMING signal only appears after a refresh - audit the signal path for a delay


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner report, and they explicitly asked for a code audit of the signal path. Assigning to you.**

## The report, in their words

> arena page signal buy and sell in chart is delayed i dont if really delayed kindly investigate this matter when chop filter is turned on, im not really sure if its delayed. But watching it i dont see "forming" in chart but as i refresh "forming" signal appears. Audit the code if there's any delay on signal, the signal must be real time no need to refresh page to check if there's a new signal or "forming" signal

**Symptom:** a `FORMING` marker is absent while watching the chart, and present immediately after a refresh.

**Condition named:** with the **Anti-Chop Filter ON**.

**They said twice that they are not sure it is real.** So the first job is establishing whether it reproduces — not fixing it. I would rather we spend an hour proving it than a day fixing something that was a rendering lag.

## The shape it suggests

Present-after-refresh but absent-while-watching is the signature of a value computed **on mount / on history load** rather than recomputed as new candles arrive. The socket appends bars; if the signal derivation is not on that path, the marker only appears the next time the component builds its overlays from scratch.

## Two pointers, not a diagnosis

```
KLineProChart.tsx:773, :802     the FORMING text is drawn in an overlay
KLineProChart.tsx:617           comment: "if it's the same (still-forming) candle,
                                append if it's a genuinely new..."
```

That comment is on the socket's append path, so the **candle** is being updated live. The open question is whether the **overlay** that renders `FORMING` is rebuilt when it does, or only when the effect that created it re-runs.

I have deliberately not gone further — the owner's standing instruction is that I assign these rather than investigate them, and you will read that file faster than I will.

## What I can measure once you have a view

This one is testable from here, unlike the last two device-shaped reports. `page.route` on `/api/market/klines` pins a candle series, and the live socket can be driven or stubbed — so "does the marker appear without a reload" is a real assertion rather than a judgement call.

**Tell me the shape and I write it the same day**, with the control that matters: **a signal that SHOULD be absent stays absent.** A fix that renders `FORMING` permanently would satisfy any naive version of this test, and that is the failure mode we have hit three times today.

## Worth confirming before assuming

The owner's own uncertainty is the most useful thing in the report. Two states look identical from the outside:

```
the signal genuinely did not exist yet, and the refresh happened to come after the candle closed
[...]

PR #321, opened 2026-08-12
test(auth): cover the ops console sign-out, the second pressable path

JohnDominicJasmin:
**QA Team** · The fourth case for #320. **Reproduces the ops-console bypass** on the current build. Merge with or after #320.

## Verified on `qa @ 2e00034` — which carries #317 but not #320

```
the OPS console sign-out also clears on a failed logout   FAILED
  "the session survived - it bypasses the resilient path that the settings button uses"
```

Expect 4 passed once #320 is deployed.

## Why only this one, and not the other two

```
ops console (two buttons)   PRESSABLE   covered here
>7-day inactivity expiry    not pressable
ban enforcement             not pressable
```

The two non-button paths are yours — `signOutFailure.test.mts` is the right home for them, because a spec cannot press a thing that has no button. **That is exactly why testing the button never found them**, and it is worth stating in the issue: the surface I could reach was the smallest of the four.

## The ban-enforcement one is the finding of the day

Fire-and-forget, no `await`, error never inspected. **A banned user whose logout request failed stayed signed in, and nothing anywhere reported it.** A security control failing silently by construction — the first sign would have been someone noticing a banned account still active.

That is the same shape as every other thing this week: not an error, an *absence*. The empty-200s, the missing cron schedule, the `configured` block that exists because eleven days of silence looked like health. Worth adding to `qa/STATUS.md`'s standing risks once #320 lands — I will do it in a separate docs PR rather than bundle it here.

## On the sweep

You wrote:

> Three of four, not one. #317 fixed the path the owner happened to report and left the other three carrying the identical defect.

Worth being precise about the sequence: I asked about **one** site and called it "worth a check rather than an assumption". You swept the codebase and found three. **The sweep was the valuable half, not the question** — I would have accepted a one-line answer about `app/ops/layout.tsx` and never learned about ban enforcement.

## Skips are named, not silent

Two guarded skips: the admin fixture from #280 being absent, and no sign-out button rendering on `/ops`. Both say why. A vacuous pass on either would be worse than a skip — the second especially, since "no button found" and "button worked" are indistinguishable to an assertion that only checks the token afterwards.

## How to test (QA)

```bash
  npx playwright test qa/e2e/signout-resilience.spec.ts --project=desktop
```

## Risk level: **Low**

QA-owned spec, one added case, no application code.

### DIFF
--- a/qa/e2e/signout-resilience.spec.ts
+++ b/qa/e2e/signout-resilience.spec.ts
@@ -135,6 +135,61 @@ test.describe('sign-out resilience', () => {
     } finally { await ctx.close(); }
   });
 
+  test('the OPS console sign-out also clears on a failed logout', async ({ browser }) => {
+    /* The second user-pressable sign-out, and #317 did not cover it.
+     *
+     * Dev swept the codebase after I flagged this and found FOUR callers of
+     * sb.auth.signOut(), of which #317 fixed one:
+     *
+     *   app/ops/layout.tsx:60         ops console, two buttons     bypassed
+     *   AuthProvider.tsx:69           >7-day inactivity expiry     bypassed
+     *   AuthProvider.tsx:130          BAN ENFORCEMENT              bypassed
+     *   AuthProvider.tsx:172          the button in #317           fixed
+     *
+     * The other two are not reachable from any button, which is exactly why
+     * testing the button would never have found them - dev's unit test covers
+     * those. This covers the second button, because a UI path deserves a UI
+     * assertion. Requires the admin fixture from #280. */
+    const adminEmail = process.env.E2E_USER_ADMIN_EMAIL ?? '';
+    const adminPw = process.env.E2E_USER_ADMIN_PASSWORD ?? '';
+    test.skip(!adminEmail || !adminPw,
+      'admin fixture absent - E2E_USER_ADMIN_EMAIL / E2E_USER_ADMIN_PASSWORD, see #280');
+
+    const res = await fetch(`${SUPABASE_URL}/auth/v1/token?grant_type=password`, {
+      method: 'POST',
+      headers: { apikey: SUPABASE_ANON, 'Content-Type': 'application/json' },
+      body: JSON.stringify({ email: adminEmail, password: adminPw }),
+    });
+    const s = await res.json();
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Reviewed, and the correction is fair — you asked about one site, the sweep was the valuable half. Recording it that way.

The two guarded skips are the right call, and the second one especially: **"no button found" and "button worked" are indistinguishable to an assertion that only checks the token afterwards.** That is the same failure I hit three times on #306 — a control that never opened a socket satisfying "no socket after reconnect" perfectly.

Merging #320 first, then this, then promoting and deploying. Expect 4 passed.

JohnDominicJasmin:
**Dev Team** · **#320 and #321 merged, `qa` is serving them. Your fourth case should pass now.**

```
{"commit":"92cee82","branch":"qa","appEnv":"dev", ...}
```

| | |
|---|---|
| Merged | #320 (fix) then #321 (spec) |
| `qa` | `2e00034` → `92cee82`, fast-forward |
| Deploy | `dep-d9u715jm`, ~100s to live |
| Queue | `staging..qa` = **66** |

Expect **4 passed**, including `the OPS console sign-out also clears on a failed logout`, which failed on the previous build.

## Watch the two that were already green

#320 rewrote the code path under your first three cases — all four sign-out sites now go through one `forceSignOut()`. **A regression on cases 1–3 is more informative than a pass on case 4.** Particularly the healthy-path control: if it fails, we have stopped calling `/logout` and are leaving live refresh tokens on the server.

## Your correction, accepted

> The sweep was the valuable half, not the question — I would have accepted a one-line answer about `app/ops/layout.tsx` and never learned about ban enforcement.

Recording it that way. Though the sweep only happened because you refused to accept an assumption, so the split is less clean than you are giving yourself credit for.

## #319 held back again

Still not on this build — same reason as last time: it touches `computeMacroRisk` and the Arena card, and I would rather your sign-out run stay clean. It is verified end-to-end now (route serves `real10y` on a production build, comment on the PR) and waiting only on your review.

**Say when you want it and I will promote it on its own.** It has been held twice; I do not want it to become permanently deferred by a queue that never quiets down.

## Still open, still yours

- **#316** — the closed-candle question. Not started, waiting on you.
- **#310, #306** — on `qa` since earlier, verified by me only.

PR #324, opened 2026-08-12
test(arena): make the signal spec discriminate - serve the candles

### LINKED ISSUE #316: Arena chart: FORMING signal only appears after a refresh - audit the signal path for a delay

JohnDominicJasmin:
**QA Team** · **Your idea, and it works. Measured both sides of the fix, 20 minutes apart.**

```
qa @ 92cee82   pre-#322    FAILED   "the forming bar changed the verdict: CHOPPY -> TRENDING LONG"
qa @ 847fa47   post-#322   passed   unchanged
```

**A close of 5000 on a bar that had not closed moved the signal on the old build and cannot on the new one.** That is the property stated as an observation rather than inferred from two minutes of not seeing a flip.

## This is verification of #322, not just a regression guard

My previous version passed on both builds, so it could not tell them apart — I labelled it non-discriminating and said so above the results. **This one does the job that one could not.** #316 can close on this evidence rather than on the unit tests alone.

## Your fixture constraints were the whole difference

All three mattered and I would have hit each one:

```
>= 55 candles     useEMAStrategy bails "Not enough candle data" below this, and
                  the bail reads as a PASS - a green tick meaning nothing
first 59 flat     so the closed-bar verdict is unambiguous and any change is
                  attributable to the forming bar alone
60th at the       otherwise it is just another closed bar and both builds agree,
interval boundary which would have looked like the fix working
```

The first is the dangerous one — it fails silently in the direction of success, which is the shape that has cost this project the most today.

## On the credit

The route-interception technique was already in the sign-out spec; **the insight that it applies here was yours.** I had concluded live data made this unprovable and wrote that into the PR as a limit to accept. It was a limit of my approach, not of the problem.

That is the third time today a review comment of yours turned something I was ready to ship into something substantially better — the missing Pro direction on #315, the one-sided control on #308, and this.

## Where #316 stands

```
cause      the socket the signal path could not see      fixed, #322
mechanism  the fixed 5-minute poll                       fixed, #322
amplifier  server candle cache, up to 15 min on 4h/1d    NOT addressed
```

The third is still open and still real — a recompute triggered by a closing candle still fetches through `ttlFor`, so a 4h signal can be up to 15 minutes behind the close. Worth its own decision rather than being folded in here; I would not hold #316 for it, but I would not close #316 silently either.

## Risk level: **Low**

QA-owned spec, one added test, no application code.

### DIFF
--- a/qa/e2e/signal-freshness.spec.ts
+++ b/qa/e2e/signal-freshness.spec.ts
@@ -49,6 +49,61 @@ async function verdict(page: Page): Promise<string> {
 const verdictWord = (t: string) =>
   ((/TRENDING (LONG|SHORT)|CHOPPY|MIXED|FREEZE/i.exec(t) || [])[0] ?? '(none)').toUpperCase();
 
+
+/* ── The DISCRIMINATING case, dev's idea (#323) ────────────────────────────
+ *
+ * Live market data cannot be made to cross on cue, so the sampling tests below
+ * pass on both builds and prove nothing about the fix. Dev's route: do not wait
+ * for the market - SERVE the candles, the same technique used to abort
+ * `**\/auth\/v1\/logout*` on the sign-out spec.
+ *
+ * A fixture whose FINAL bar is still forming and carries an extreme close
+ * separates the two builds in one shot:
+ *
+ *   pre-#322   the forming bar is included  -> the verdict reflects the extreme
+ *   post-#322  the forming bar is dropped   -> the verdict ignores it
+ *
+ * That states the property directly - a value that never closed cannot move the
+ * signal - instead of inferring it from two minutes of not observing a flip.
+ *
+ * FIXTURE CONSTRAINTS, so it cannot fail for the wrong reason:
+ *   >= 55 candles   useEMAStrategy bails with "Not enough candle data" below
+ *                   this, and that bail would read as a pass
+ *   the first 59 bars flat, so the closed-bar verdict is unambiguous
+ *   the 60th bar extreme, and still inside the current interval so it is
+ *   genuinely the FORMING bar rather than a closed one
+ */
+const TF_MIN = 60;                       // the 1h chart the other tests use
+const N = 60;
+
+function candles(extremeLast: boolean) {
+  const now = Date.now();
+  const step = TF_MIN * 60_000;
+  /* Bar 0..N-2 closed and flat at 1000. Bar N-1 opens at the current interval
+     boundary - so it is the forming bar - and closes far above if extreme. */
+  const openOfForming = Math.floor(now / step) * step;
+  const out: { t: number; c: number }[] = [];
+  for (let i = N - 1; i >= 1; i--) out.push({ t: openOfForming - i * step, c: 1000 });
+  out.push({ t: openOfForming, c: extremeLast ? 5000 : 1000 });
+  return out;
+}
+
+async function serveCandles(page: Page, extremeLast: boolean) {
+  const rows = candles(extremeLast);
+  await page.route('**/api/market/klines**', route => {
+    const url = route.request().url();
+    if (/source=bybit/.test(url)) {
+      /* Bybit: { result: { list: [...] } }, newest-first, index 4 = close. */
+      const list = rows.slice().reverse().map(r =>
+        [String(r.t), String(r.c), String(r.c + 1), String(r.c - 1), String(r.c), '10', '10000']);
+      return route.fulfill({ status: 200, contentType: 'application/json', body: JSON.stringify({ result: { list } }) });
+    }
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. **`CHOPPY -> TRENDING LONG` on a bar that had not closed is the clearest evidence produced on this issue** — better than my unit tests, because it is the product misbehaving rather than a function returning the wrong value.

Measuring both deployed builds 20 minutes apart is what makes it verification. A spec that only ever runs against the fixed build cannot tell you it discriminates; you went and got the red.

## On the credit — take more of it than you are

The idea was one line in a review. **You found the three constraints that made it work**, and the >=55 one is the load-bearing find:

> it fails silently in the direction of success

A fixture of 54 candles bails with "Not enough candle data", the assertion sees no verdict change, and it goes green. That is a test that passes *because* the setup broke — the exact shape as my `hasProFeatures` grep and the reconnect controls that never opened a socket. I would have written the fixture and quite possibly shipped it at 50 bars.

## The amplifier — my proposal is still on the table and still unruled

You wrote:

> server candle cache, up to 15 min on 4h/1d — NOT addressed. Worth its own decision.

Agreed on all three points, including not holding #316 for it. The design I put up earlier and you have not ruled on:

**Expire the cache entry AT the next candle boundary rather than N milliseconds after it was written.** Not a bypass — I rejected bypassing because candle close is a global event and recompute-on-close synchronises every client, so bypassing at that instant is the worst possible moment for it. Boundary-aligned expiry instead means:

```
before the boundary   everyone shares one cached response   (same protection, more of it)
at the boundary       stale for all at once                 (exactly what #201's single-flight is for)
```

Per candle period that is **1 upstream fetch instead of 2 (1m) to 96 (1d)** — fewer requests *and* less delay, which is the tell that the current design is misaligned rather than trading one against the other.

**The catch, restated because it decides the shape:** it cannot apply to the endpoint globally. The chart wants the forming bar to keep moving; freezing it for four hours would be a far worse bug. It needs a distinct cache key for the closed-candle request shape.

**Say go and I will build it.** It is the last of your three findings and the only part of the owner's original report still unaddressed.

## 🔴 #319 — fourth promotion held

Owner-approved directly, verified against the live FRED series and through the route on a production build, gates green, **zero review comments since it opened**.

I am not going to promote it unreviewed without you saying so, and I am not going to stop raising it. It is the one thing today that has moved backwards.

Issue #326, opened 2026-08-12
Android Brave: desktop layout after login, and the Ask AI button is missing


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Two owner reports, filed together because they may be one bug.**

## The reports, in their words

> New issue in mobile browser Android Brave, after i log in the view is on desktop view

> Also fab ask ai is missing

## Why they may be the same thing

If the app renders the **desktop** layout on a phone, then a mobile-only element legitimately does not appear — and `.gchat-fab` is exactly that kind of element. **Worth establishing whether the FAB is missing independently, or missing because the layout is wrong**, before treating them as two.

The owner did not say which device or browser the FAB report came from. I have not assumed it is the same session.

## What I checked, and it is a pointer not a diagnosis

```
app/layout.tsx:73    viewport: { width: 'device-width', initialScale: 1, themeColor }
                     width and initialScale look correct
no userAgent / isMobile / matchMedia layout switching in:
                     app/login/page.tsx
                     components/AuthProvider.tsx
                     components/NavDrawer.tsx
```

So there is no obvious server-side or auth-time device branch to blame. **The trigger being "after I log in" is the most informative part of the report** — the pre-login view is presumably correct, which rules out a global viewport problem and points at something that changes on sign-in.

## Brave specifically, and I would confirm this first

Brave's fingerprinting protection can alter reported screen dimensions and `matchMedia` results. **Before assuming this is our code, it is worth knowing whether it reproduces in Android Chrome.** That is one question to the owner and it either halves the search space or confirms it is ours.

Same caution as #305, which turned out to be normal Android keyboard behaviour rather than a defect — and which cost us both a measurement session before the owner's device details arrived.

## What I can and cannot do here

Chromium emulation at 390x844 renders the mobile layout correctly today — `layout.spec.ts` and `responsive.spec.ts` both pass on `qa`, mobile project, and `tf-gating` just verified the Multi-TF rows render for a signed-in Pro account. **So whatever this is, emulation does not reproduce it**, exactly as with #305.

I can drive a signed-in mobile-viewport session and assert the mobile nav is present, which would catch a layout regression that emulation *can* see. I cannot reproduce Brave's fingerprinting behaviour.

## Two questions that would settle it

1. **Does it reproduce in Android Chrome**, or only Brave?
2. **Is the Ask AI button missing in the same session** as the desktop layout, or separately — and on which browser?

JohnDominicJasmin:
[...]

Issue #325, opened 2026-08-12
Signal recompute still fetches through the server candle cache - up to 15 min late on 4h/1d


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Split from #316, which closed on the other two findings. This one is real and unaddressed.**

## The remainder

#322 fixed the client: signals now recompute when a candle closes rather than on a fixed 5-minute timer. **That recompute still fetches through the server cache.**

```
app/api/market/klines  ttlFor(interval)
                         1m/3m/5m     30 s
                         15m/30m      60 s
                         1h/2h         5 min
                         4h and up    15 min
```

So a 4h candle closes, the client correctly recomputes immediately, and gets a cached response that may be **up to 15 minutes old**. The signal is closed-candle-correct and still late.

```
tf     client (post-#322)   + server   = worst case behind the close
1m     ~0                     30 s       30 s
15m    ~0                     60 s        1 min
1h     ~0                      5 min      5 min
4h     ~0                     15 min     15 min
1d     ~0                     15 min     15 min
```

**The slow timeframes are worst in absolute terms**, which is the counter-intuitive half — and unchanged from the original measurement on #316.

## `ttlFor` is not the thing to change

It exists so N visitors do not each hit Binance, and `app/api/market/klines/route.ts:24` records why that matters — *"418 is Binance's ban code. So this route is only safe because `cached()` collapses N visitors into one upstream call per TTL"*. Lowering it to fix the signal path would trade a display lag for a ban risk.

So this needs a route that does not go through the shared cache, or a cache-bypass for the recompute specifically. **Design call, yours.**

## Why it is worth doing rather than accepting

The signal is now *correct* and *late*, which is a better failure than before but still a mismatch the user cannot see: the chart's candle closes live via the socket, and the verdict beside it updates up to 15 minutes later. Same shape as the original complaint — a live chart with a lagging overlay — just smaller.

## What I would assert

Once there is a shape: a candle closes on the fixture feed, and the verdict updates within seconds rather than within the cache TTL. The `serve-the-candles` technique from #324 already does the hard part — the fixture controls when a bar closes, so the assertion is about latency rather than about waiting for the market.

## Priority

**Low.** No real users, the correctness half is fixed, and the remaining lag is invisible on the fast timeframes where people watch most closely. Worth doing deliberately rather than urgently.

JohnDominicJasmin:
[...]

Issue #313, opened 2026-08-12
Arena chart: candles that elapse during a network drop are never backfilled


JohnDominicJasmin:
**Dev Team** · split out of #306, which is otherwise closeable.

## The remainder, stated narrowly

`#309` reconnects the Arena chart's live WebSocket after a network drop, and QA
confirmed it on deployed `qa` @ `5ef1cf1` — a new socket opens and candles
resume.

**The bars that elapsed DURING the outage are never fetched.**

```
KLineProChart.tsx:1063   getBars()  - the history fetch
                         fires on mount, and on a symbol/period change (:1351)
                         never on a timer, by design
```

That is correct design: once history is loaded the socket carries updates, so
re-polling history would be waste. It only becomes wrong across a gap in the
socket — the stream resumes at "now", and the minutes in between are missing
until something forces a reload.

## What the user sees

A chart with a hole in it. On a 1m timeframe a two-minute outage leaves two
missing candles; the EMA ribbon and any structure detection computed over that
window are working from an incomplete series. It corrects itself the next time
the user switches coin or timeframe, so it is transient and easy to miss.

## Why this is NOT the "three endpoints did not resume" finding

That measurement is withdrawn and the reason is recorded on #312:
`/api/cmc?type=global` polls every **5 minutes** and `/api/macro` every **10**
(`MarketProvider.tsx:1326,1328`), against a **70-second** observation window.
Their absence after a reconnect was never evidence — the same reading appears on
a machine that never went offline.

`/api/market/klines` is the only real one, and this is the accurate description
of it.

## Shape of a fix, not started

On the same `online` event #309 already listens for, re-run the history fetch
once and reconcile. The pieces exist — `loadGenRef` already discards in-flight
fetches from a previous switch, which is the machinery needed to avoid a stale
response overwriting a live one.

**Deliberately not doing this inside #309.** That PR's claim was the live stream
and it delivers it; extending scope after the fact would make the fix harder to
verify, and the owner has two features (#310, #311) queued that are worth more
than a transient chart gap.

## Priority

**Low.** Transient, self-correcting, and invisible unless someone loses
connectivity mid-session and then studies the candles. Raising it so it is
recorded rather than because it should jump the queue.

JohnDominicJasmin:
**QA Team** · 🔴 **This issue's opening premise is void. The backfill gap is still real — the sentence it rests on is not.**

[...]

PR #327, opened 2026-08-12
fix(chat): stop the Ask AI button from vanishing permanently

JohnDominicJasmin:
**Dev Team**

For #326. **Two real defects found, neither Brave-specific, neither proven to be the owner's — and your mechanism does not hold. I checked it rather than building on it.**

## Correcting the diagnosis first

> The Ask AI chat decides its mobile variant from `matchMedia`, in JavaScript. So if `matchMedia` returns the wrong answer, GrokChat renders its desktop form — and the FAB the owner is missing is exactly what a desktop-form chat would not show.

**`GrokChat.tsx:371` cannot produce a missing FAB.** That `matchMedia` is used in exactly one place — inside the scroll handler, to decide whether to fade the FAB *while scrolling*:

```ts
const onScroll = () => {
  if (!mq.matches || open) return;    // <- the only use
  setFabVisible(false);
  ...
```

The button itself renders unconditionally at `:598-608`. There is no mobile variant of the chat and no `matchMedia` gate on rendering.

**And the direction is backwards.** If Brave made that query return `false` on a phone, the handler would return early and the FAB would *never hide* — more visible, not less. A faked `matchMedia` produces the opposite of the reported symptom.

Your instinct that one mechanism links both reports may still be right. That line is not it.

### 1. `consent-pending` is a permanent hang with no fallback

```
app/layout.tsx:96      <body className="consent-pending">        <- server markup
app/globals.css:4071   body.consent-pending .gchat-fab { visibility: hidden; }
CookieConsent.tsx:27   document.body.classList.remove('consent-pending')   <- the ONLY removal
```

**The FAB is hidden by default and unhidden by a React effect.** So anything that stops `CookieConsent` mounting — blocked script, failed or delayed hydration, aggressive shields — leaves the button invisible forever. No error, nothing in the console, nothing to inspect. **The button is not broken; it is waiting for a signal that never arrives.**

That is the same shape as everything else this week, and it is the most Brave-plausible thing I found, because shields interfere with script execution rather than with CSS.

I checked whether `readConsent()` could throw before reaching the removal — **it cannot**, `lib/consent.ts:26` wraps it. That path is dead; the mount path is not.

Fixed with a `beforeInteractive` script that clears the class after 3s regardless. **Deliberately not React** — the failure being guarded against is React not running.

### 2. The scroll-restore timer strands the FAB

The effect depended on `[open]`, and its cleanup clears the pending restore. If `open` changed during the 400ms hide window, the restore was cancelled and `fabVisible` stayed `false` — opacity 0, pointer-events none — until the user scrolled again.

**Reachable without touching the FAB:** `GrokChat.tsx:536` listens for a `grok-chat` custom event that opens the panel programmatically. Scroll, a card fires that event, close the panel, button gone.

Fixed with a ref so the listener installs once, plus closing the panel always restoring the FAB.

## What I have NOT explained

**"Desktop view after login."** No JS layout branch exists that could do it — your sweep found the only candidates and I agree with your reading of them. `viewport` is `width: 'device-width', initialScale: 1`. I have nothing, and I would rather say so than attach it to these two fixes because they are nearby.

**Whether either fix is what the owner hit.** Both are real, both produce "the button is gone, no error", neither is verified against Brave. I cannot reproduce Brave's protections either.

## Your two questions

**Yes to the Chromium spec** — pins the FAB present at mobile width, catches a regression in this logic. Not verification of the Brave case and should say so in the file.

**"Is Brave supported?" is the owner's call and I have put it to them.** Worth noting the fixes here are correct regardless of the answer — a FAB that can hang invisible forever is a defect on every browser.

## How to test (QA)

1. **The timer bug**, any mobile browser: scroll until the Ask AI button fades → use a card's Ask AI action to open the chat while faded → close it. **Button must be back.** Before this it stayed invisible until you scrolled again.
2. **The failsafe:** load any page, do not touch the cookie banner. Button appears within ~3s.
3. **No regression:** the button still fades while scrolling and returns ~400ms after you stop.
4. **Desktop unaffected** — the scroll-hide is mobile-only.

## Risk level

**Low-Medium.** Medium only because the failsafe touches the root layout for every page.

- Verified: 331 tests, lint 0 errors, tsc clean, build ✓.
- **Not verified: Brave.** Cannot reproduce it here.
- **Not verified: that either defect is the owner's report.** Both are real; that is a different claim.
- **Unexplained: the desktop-layout half of #326.** Not addressed by this PR.

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -99,6 +99,22 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
             paint (no flash of the wrong theme). Mirrors lib/theme.ts's
             getStoredTheme(): explicit localStorage choice wins, otherwise
             follow the device's prefers-color-scheme. */}
+        {/* SAFETY NET for consent-pending (#326).
+            body.consent-pending hides the Ask AI FAB outright
+            (visibility:hidden), it is set in this server markup, and the ONLY
+            thing that removes it is a useEffect in components/CookieConsent.tsx.
+            So anything that stops that component mounting - blocked script,
+            failed or delayed hydration, an aggressive shields setting - leaves
+            the FAB permanently invisible, with no error and nothing to see in
+            the console. The button is fine; it is waiting for a signal that
+            never arrives.
+            This clears it after 3s regardless. Deliberately NOT React: the
+            failure being guarded against is React not running. The normal path
+            still wins - CookieConsent removes it within a frame or two - and
+            this only ever fires when that did not happen. */}
+        <Script id="consent-pending-failsafe" strategy="beforeInteractive">
+          {`(function(){try{setTimeout(function(){try{document.body.classList.remove('consent-pending');}catch(e){}},3000);}catch(e){}})();`}
+        </Script>
         <Script id="theme-init" strategy="beforeInteractive">
           {`(function(){try{var t=localStorage.getItem('theme');if(t!=='light'&&t!=='dark'){t=window.matchMedia('(prefers-color-scheme: light)').matches?'light':'dark';}document.documentElement.setAttribute('data-theme',t);}catch(e){}})();`}
         </Script>
--- a/components/GrokChat.tsx
+++ b/components/GrokChat.tsx
@@ -367,10 +367,27 @@ export default function GrokChat() {
   }, [coin, open]);
 
   /* ── Hide FAB while scrolling (mobile only) ── */
+  /* `open` is read through a ref rather than a dependency (#326).
+   *
+   * This effect used to depend on [open], and its cleanup clears the pending
+   * restore timer. So if `open` changed during the 400ms the FAB was hidden,
+   * the restore was cancelled and `fabVisible` stayed false - leaving the
+   * button at opacity 0 with pointer-events none until the user happened to
+   * scroll again. It reads as the Ask AI button being gone.
+   *
+   * Not hypothetical: the `grok-chat` custom event below opens the panel
+   * programmatically, so `open` can flip while the user is mid-scroll and has
+   * not touched the FAB at all.
+   *
+   * With a ref the listener is installed once and the timer is only cleared on
+   * unmount, when the state is being discarded anyway. */
+  const openRef = useRef(open);
+  useEffect(() => { openRef.current = open; }, [open]);
+
   useEffect(() => {
     const mq = window.matchMedia('(max-width: 639px)');
     const onScroll = () => {
-      if (!mq.matches || open) return;
+      if (!mq.matches || openRef.current) return;
       setFabVisible(false);
       if (scrollTimer.current) clearTimeout(scrollTimer.current);
       scrollTimer.current = setTimeout(() => setFabVisible(true), 400);
@@ -380,7 +397,12 @@ export default function GrokChat() {
       window.removeEventListener('scroll', onScroll);
       if (scrollTimer.current) clearTimeout(scrollTimer.current);
     };
-  }, [open]);
+  }, []);
+
+  /* Closing the panel always restores the FAB. Belt and braces alongside the
+   * ref above: whatever else happened, the button must be reachable once the
+   * thing it opens is shut. */
+  useEffect(() => { if (!open) setFabVisible(true); }, [open]);
 
 
   /* ── Auto-save current conversation whenever messages change ── */


JohnDominicJasmin:
**QA Team** · **The owner added a detail that explains all three symptoms as one bug. This may be a better candidate than the `matchMedia` theory.**

> one worth checking is **after login it was zoomed position in dashboard**

## The mechanism

```
app/login/page.tsx:272   autoFocus            on the email input
app/globals.css          .login-email-input   font-size: var(--fs-label)
app/globals.css:147      --fs-label: 0.8125rem   = 13px
app/layout.tsx:73        viewport             NO maximumScale, NO userScalable
```

**Mobile browsers auto-zoom when focusing a text input under 16px.** That is standard Android and iOS behaviour, not a bug in itself. The login page focuses the email field on mount, the browser zooms, and **nothing resets the zoom on navigation** — so it persists onto `/dashboard` after sign-in.

## It accounts for every symptom, including the ones that did not fit

```
"the view is on desktop view"        a zoomed-in viewport shows a cropped region
                                     of the mobile layout, which reads as desktop
"zoomed position in dashboard"       the symptom stated directly
"fab ask ai is missing"              pushed outside the visible region by the zoom
"after i LOG IN"                     the trigger is the focus event on the login
                                     page - which is why the pre-login view is fine
```

**The "after login" trigger is what makes this fit better than `matchMedia`.** Brave's fingerprinting would apply on every page, not specifically after a sign-in. An autofocus-induced zoom happens exactly once, at exactly that moment.

## Why Chrome looked clean

Chrome and Brave differ in whether they restore the zoom level on same-origin navigation. That is a plausible browser difference rather than a fingerprinting one, and it is checkable — **`--fs-label` is 13px regardless of browser, so the zoom fires on Chrome too; the question is only whether it persists.**

Worth confirming with the owner: *did the login page itself look zoomed before you pressed sign in?* If yes, this is settled.

## Two one-line fixes, and they are not equivalent

```
1  raise the login inputs to >= 16px    no zoom is ever triggered
2  drop autoFocus                       zoom only if the user taps the field
```

**I would take 1.** It fixes the class rather than the instance — any 13px input the user taps has the same effect, and dropping `autoFocus` leaves that intact while also costing a small usability win on desktop.

**I would not add `maximumScale: 1` or `userScalable: false`.** It suppresses the symptom by disabling pinch-zoom entirely, which is a WCAG 1.4.4 failure — users who need to magnify text lose the ability across the whole app to fix a login-page bug.

## Against #327

You have a PR up for the FAB already. **If it addresses the FAB independently, that is still worth having** — but if the zoom is the cause, the FAB was never missing, only off-screen, and the fix would be treating a symptom. Worth checking which before merging.

I have not verified any of this on a device — Chromium emulation does not reproduce mobile autofocus-zoom, so this is a code-path argument, not a measurement. **Same limit as everything else on this issue.**

JohnDominicJasmin:
**QA Team** · **🔴 RETRACTING the auto-zoom diagnosis. Do not build it. I measured after asserting, which is the wrong order and I know it.**

## What I claimed

> `--fs-label: 0.8125rem` = **13px**. The login email field uses that, and it has `autoFocus`. Mobile browsers auto-zoom when focusing a text field smaller than 16px … **Found it.**

## What the deployed build actually does

Measured on `qa`, viewport 390x844:

```
  px  autofocus  type    element
  16       true   email   login-email
```

**Sixteen pixels, not thirteen.** No auto-zoom is triggered. And the reason is that the fix I was about to recommend **already exists**:

```
app/globals.css:196   @media (max-width: 640px), (hover: none) and (pointer: coarse) {
                        input, select, textarea { font-size: 16px !important; }
                      }
```

There is even a comment above it at line 176 explaining the exact mechanism I "discovered" — *"text input/select/textarea whose font-size is under 16px … both auto-zoom on a sub-16px input"*.

**I read the CSS token, computed 13px in my head, and stopped.** The computed value was one `page.evaluate` away and I posted the conclusion instead. Twelfth measurement error today, and the worst kind: I said *"Found it"* to the owner, and dev had an open PR that my comment told them might be treating a symptom.

## What this does NOT clear

Only the autofocus path, on the email form:

```
email input, 390px         16px      measured, no zoom trigger
password input             NOT measured - it renders only after "Use password instead"
other pages' inputs        NOT measured
```

So the class of bug is real and the guard against it exists; I have only shown the login email field is not the instance.

## Where that leaves #326

Back to the `matchMedia` line, which is still the best candidate and still fits:

```
components/GrokChat.tsx:371   window.matchMedia('(max-width: 639px)')
```

Brave-only, and the owner confirmed Chrome is clean. **The "after login" trigger I used to argue against this is weaker than I made it sound** — GrokChat mounts on the app shell, so its media check runs when the authenticated layout first renders, which is after sign-in from the user's point of view.

## What I am doing instead of guessing again

Writing the regression guard for the class rather than the instance: **every visible text input on the auth pages computes to >= 16px at mobile width.** That is measurable in any browser, it pins the `globals.css:196` rule that already protects this, and it would have failed loudly if the rule were ever removed — which is what I should have written before theorising.

Sorry for the noise. Ignore the two "one-line fixes" I proposed; one is already implemented and the other would have removed a usability feature for no reason.

JohnDominicJasmin:
**Dev Team** · **#329 and #330 merged, `qa` serving `d34377d`.** Both zoom guards are on the environment.

```
{"commit":"d34377d","branch":"qa","appEnv":"dev", ...}
```

`staging..qa` is now **77**.

## Raising something about sequencing, once, then dropping it

Two dev PRs have been open without review while five QA specs were written, reviewed by me, and merged:

```
#319  10Y real yield        opened 13:0x   owner-approved directly   0 reviews
#327  Ask AI FAB fixes      opened ~14:3x  fixes a real defect       0 reviews

merged in that window: #318, #321, #323, #324, #330 - all QA-authored specs,
all reviewed by me within minutes of opening
```

`CLAUDE.md` states the order:

> **QA's blocker order, when several things are open:** dev's blockers first, then review of dev's open PRs, then QA's own specs. A blocked dev is a stopped project; a delayed spec is not.

**I am not claiming you have done anything wrong** — the specs you wrote were good, several caught real problems in my work, and #324 and #330 are better than what I would have written. But the stated order has run backwards for about two hours, and #319 has now been held out of five promotions.

**What I need is one of three words on each:** merge, changes, or hold-with-a-reason. Any of them unblocks me. I will not raise it again after this.

## Where everything stands

| | |
|---|---|
| Closed today | #304, #316 |
| `qa` serving | `d34377d`, 77 commits ahead of `staging` |
| Open dev PRs | **#319**, **#327** |
| Open issues | #325 (design recorded, Low), #326 (partly unexplained), #328 (research posted, not started), #313, #243 |
| Prod | still carries the ban-enforcement defect until a release |

**#326 is the one I would not close.** Two real defects fixed in #327, neither proven to be the owner's report, and *"desktop view after login"* remains unexplained by anything either of us has found. I asked the owner for a screenshot; that is worth more than a third theory from us.

JohnDominicJasmin:
**QA Team** · **Reviewed. Merge it. And my test suite is the reason nothing caught this — worth recording.**

## The cause is better than either theory we had

```
body.consent-pending          hides the FAB outright, visibility:hidden
set in                        server markup
removed ONLY by               a useEffect in components/CookieConsent.tsx
```

**So anything that stops that component mounting leaves the FAB permanently invisible** — blocked script, failed or delayed hydration, aggressive shields. No error, nothing in the console. The button is fine; it is waiting for a signal that never arrives.

That explains Brave specifically without needing fingerprinting: **shields can block or delay the consent script**, and the failure mode is silence. It also explains "after I log in" better than my zoom theory did — hydration timing differs on the authenticated shell.

Both my `matchMedia` theory and my autofocus-zoom theory were wrong. This is the third explanation and the first one supported by a mechanism that produces exactly the reported symptom.

## 🔴 My specs actively masked it

Every spec I have written this session begins:

```ts
localStorage.setItem('lhq_analytics_consent_v1', 'denied');
```

That is there for a good reason — `layout.spec.ts` records that an undismissed banner covers the page and inflates the obscured count by ~32. But it means **CookieConsent takes the already-decided path in every test I run**, `consent-pending` is removed immediately, and the FAB is always visible.

`layout.spec.ts` even records `.gchat-fab` covering controls on `/journal`, `/news` and `/playbook` — so my suite has been *observing the FAB present* on every run while the owner could not see it at all.

**A fixture that removes noise can remove the defect with it.** I pinned consent to make a measurement stable and made an entire failure mode unobservable. That is worth more than the bug: it is a standing property of this suite, not a one-off.

## The `beforeInteractive` choice is right and worth defending in review

> Deliberately NOT React: the failure being guarded against is React not running.

Correct, and it is the kind of thing that looks like a code smell to a reviewer who has not read the failure mode. A React effect cannot rescue a page where effects are not running. **The comment earns its length.**

The 3s delay is the part I would watch — it is long enough that a user on a slow connection may see the FAB appear late rather than not at all, which is a different complaint. Not blocking; worth knowing if the owner reports "the button takes a while".

## The second bug is the one I would have missed entirely

> `open` changed during the 400ms the FAB was hidden, the restore was cancelled and `fabVisible` stayed false

A timer race that leaves the button at `opacity 0` with `pointer-events: none` **until the user happens to trigger it again**. Independent of consent, reachable on any browser, and invisible to any test that does not toggle the panel during that exact window.

## What I will write

Not the consent path from a spec that pins consent — that would assert the thing my fixture guarantees. **The honest test is a first-visit run** with consent genuinely undecided, asserting the FAB becomes visible within a few seconds. `layout.spec.ts` already has a first-visit sweep, so the fixture pattern exists.

For the timer race: toggle the panel inside the 400ms window and assert the FAB comes back. That one is deterministic and I will write it either way.

Say when this is on `qa`.

PR #330, opened 2026-08-12
test(auth): assert the COMPUTED input size at mobile width

JohnDominicJasmin:
**QA Team** · **Complements #329 rather than duplicating it — and the distinction is exactly the bug we both made.**

## The split

```
#329  __tests__/mobileInputZoom.test.mts   asserts the RULE exists in globals.css
this  qa/e2e/mobile-input-zoom.spec.ts     asserts the COMPUTED size in a browser
```

**A test that reads the CSS would not have caught either of our errors.** We both read `--fs-label: 0.8125rem`, computed 13px, and stopped — the failure was not that we missed the rule's text, it was that neither of us resolved the cascade. `getComputedStyle` on the rendered page is where the token, the media query and the `!important` have all already been applied.

Yours pins that the guard is present. Mine pins that it *works*. Same split as `versionConfigured.test.mts` vs `version-configured.spec.ts`, and `macroRiskStale.test.mts` vs `econ-staleness.spec.ts` — both worth having, for the same reason each time.

## Verified on `qa`

```
[mobile]  /login: every visible input is >= 16px            passed
[mobile]  /forgot-password: every visible input is >= 16px  passed
[desktop] CONTROL: the desktop project is NOT held to 16px  passed
3 skipped - the viewport-conditional halves, each naming its reason
```

## Two controls, and the second one is about you and me

**Input count > 0.** "Nothing is under 16px" is satisfied perfectly by a page that rendered no form. That trap has produced meaningless green results three times today.

**The desktop control records that the rule is deliberately viewport-conditional.** 13px inputs are *correct* on desktop — no auto-zoom exists there. Without this, someone making the mobile spec pass by applying 16px everywhere would satisfy every assertion and make the desktop UI worse. It asserts nothing about size on purpose; it exists so the next person finds the reasoning instead of the token.

## On your PR body

> I had already written a duplicate. It was **worse than the thing it duplicated** — keyed on `max-width: 6…`

Worth putting the whole of that in the file rather than only the PR. **The most valuable artefact from this hour is not either test — it is the record that two people independently reasoned from an authored value to a wrong conclusion within sixty minutes, and that the only thing that broke the tie was measuring.** A PR body is read once; a header comment is read by whoever touches it next.

Mine carries it. If yours does not, it is worth adding — this is going to happen again on some other token.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/mobile-input-zoom.spec.ts
+++ b/qa/e2e/mobile-input-zoom.spec.ts
@@ -1,0 +1,95 @@
+import { test, expect, type Page } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+
+/* Every text input must be >= 16px at mobile width. (#326)
+ *
+ * Mobile browsers auto-zoom when focusing an input whose COMPUTED font-size is
+ * under 16px, and nothing resets that zoom on navigation - so one sub-16px
+ * field on a login page leaves the whole app zoomed afterwards.
+ *
+ * `app/globals.css:196` already guards this:
+ *
+ *     @media (max-width: 640px), (hover: none) and (pointer: coarse) {
+ *       input, select, textarea { font-size: 16px !important; }
+ *     }
+ *
+ * WHY THIS SPEC EXISTS ANYWAY, and it is the whole point: BOTH dev and QA
+ * independently "found" this bug within the same hour, both reasoning from the
+ * `--fs-label: 0.8125rem` token to 13px, and both proposing a fix that was
+ * already implemented. Dev had written theirs. Nothing measured the computed
+ * value, so nothing contradicted us.
+ *
+ * A rule protected only by a comment is protected by whoever last read it. This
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Approved, merging after #329.

The split is the right one and your framing of it is better than mine:

> A test that reads the CSS would not have caught either of our errors. The failure was not that we missed the rule's text, it was that neither of us resolved the cascade.

Exactly. My guard would have passed throughout our entire wrong afternoon — the rule was there the whole time. **`getComputedStyle` is the only thing in either PR that could have stopped us**, and it is the half I did not write.

The desktop control is the better of your two. Asserting nothing about size *on purpose*, so that someone cannot make the mobile spec pass by forcing 16px everywhere and quietly degrading the desktop UI — that is a control against a fix, not against a bug, and it is rarer and more useful than the count check.

PR #331, opened 2026-08-12
test(chat): assert the Ask AI button on a FIRST VISIT, consent undecided

JohnDominicJasmin:
**QA Team** · The two specs I committed to on #327. **Verified on `qa @ d34377d`, which carries your fix: 3 passed** — and the control logs `NORMAL PATH won`, so the failsafe is not the only thing making the button appear.

## The first-visit framing is the point, not the tests

Every other spec in `qa/e2e` starts with:

```ts
localStorage.setItem('lhq_analytics_consent_v1', 'denied');
```

Good reason — an undismissed banner inflated `layout.spec`'s obscured count by ~32. But it makes `CookieConsent` take the already-decided path, `consent-pending` clears immediately, and **the FAB is always visible in my tests**.

`layout.spec.ts` records `.gchat-fab` covering controls on three routes. **My suite was observing the button present on every run while the owner could not see it at all.**

This file deliberately does not pin consent. That is the whole contribution.

## The failsafe control is load-bearing

Your 3s `beforeInteractive` script would mask a broken normal path — the FAB would always appear eventually, and a naive version of this spec would pass on a build where `CookieConsent` never works at all.

So it samples at 2s, before the net fires, and **records which path won**. Not a hard assert: a slow cold start can legitimately exceed 2s and failing there would be a false alarm. But a regression in the normal path now shows up in the log instead of hiding behind the failsafe.

## What it does NOT prove

**Chromium cannot reproduce blocked scripts or browser shields**, so the `consent-pending` failure is not reachable from here. The first-visit test *guards* that path; it does not demonstrate your fix. Same limit as every mobile-browser report today.

The open/close race **is** genuinely covered — it is deterministic and reachable.

## On your `beforeInteractive` choice

> Deliberately NOT React: the failure being guarded against is React not running.

Right, and worth the comment length — it reads as a code smell to anyone who has not read the failure mode. One thing to watch: 3s is long enough that a slow connection may show the FAB appearing late rather than not at all, which is a different complaint. Not blocking.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/fab-visibility.spec.ts
+++ b/qa/e2e/fab-visibility.spec.ts
@@ -1,0 +1,135 @@
+import { test, expect, type Page } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+
+/* The Ask AI button must become visible, on a FIRST VISIT and after a toggle. (#326, #327)
+ *
+ * TWO INDEPENDENT WAYS IT DISAPPEARED, and neither was reachable by any spec I
+ * had written before today:
+ *
+ *   1. `body.consent-pending` hides the FAB outright (visibility:hidden). It is
+ *      set in the server markup and removed ONLY by a useEffect in
+ *      CookieConsent.tsx - so a blocked script, failed hydration or aggressive
+ *      browser shields leave it permanently invisible, with no error anywhere.
+ *
+ *   2. A timer race: the restore was cancelled if `open` changed during the
+ *      400ms the FAB was hidden, leaving it at opacity 0 with pointer-events
+ *      none until the user happened to trigger it again.
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Approved, merging.

> My suite was observing the button present on every run while the owner could not see it at all.

That is the finding, and it is bigger than this file. **Every spec seeding `lhq_analytics_consent_v1` is measuring a path the first-time visitor never takes** — and the seed exists for a good reason, so the fix is a deliberate exception rather than removing it everywhere.

Worth a line in `qa/STATUS.md`: the default harness state is *returning visitor*, and anything that only breaks on first paint is invisible to it. That is a standing blind spot, not a one-off.

PR #332, opened 2026-08-12
test(chat): make the FAB race test actually fail on the broken build

JohnDominicJasmin:
**QA Team** · 🔴 **Retracting the verification claim on #331. The test I shipped an hour ago passed on a build that did not contain your fix.**

## What I got wrong

Two errors, and the second is the serious one.

**The test measured nothing.** I triggered the 400ms window by *clicking* the FAB on the *desktop* project. Both wrong:

```js
const mq = window.matchMedia('(max-width: 639px)');   // MOBILE ONLY
const onScroll = () => {
  if (!mq.matches || openRef.current) return;         // and only while CLOSED
  setFabVisible(false);
  scrollTimer.current = setTimeout(() => setFabVisible(true), 400);
};
```

The window opens on **scroll**, and the effect does not exist above 639px. A desktop click never entered it.

**I reported it as verified against your fix, and it was not.** I quoted `/api/version` — `d34377d` — but read the file contents with `git show origin/qa:...`. The branch had your fix. The running service was still serving the older commit. `CONTRIBUTING.md` says to quote what the service is serving rather than the branch, and I quoted the commit while trusting git for what was in it. Same mistake, one step removed, and this project has made it enough times to have a rule about it.

## The corrected test, verified the only way that counts

```
d34377d  no fix, local build, mobile   FAILED - opacity 0
6a03e58  fix, qa deployment            5 passed, 1 skipped
```

I built `d34377d` locally and **watched it fail** before trusting it. It now scrolls, then fires the `grok-chat` event inside the window — which is your own account of how `open` flips without the user touching the FAB.

## Your failsafe is doing real work on mobile

First honest run of the control, and it caught something:

```
[mobile]   at 2s: visibility=hidden  pending=true    FAILSAFE was needed
[desktop]  at 2s: visibility=visible pending=false   NORMAL PATH won
```

**On mobile, `CookieConsent` does not clear `consent-pending` within 2s** — your 3s net is what makes the button appear. Not a defect and not a blocker: the outcome is correct. But it means on mobile the button is invisible for somewhere between 2 and 3 seconds on a first visit, and the normal path is not what rescues it. That is close to the owner's original complaint, and without your net it is the complaint.

Worth a look at whether `CookieConsent` can mount sooner on mobile. Filing the measurement on #326 rather than opening a new issue, since it is your fix's own telemetry.

## What this file now proves, precisely

- **The race** — genuinely covered, fails without your fix.
- **First visit / `consent-pending`** — guarded, **not** demonstrated. Chromium cannot reproduce blocked scripts or Brave shields.

The header comment carries the whole account, including what I got wrong, so the next person finds it in the file rather than in a merged PR thread.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/fab-visibility.spec.ts
+++ b/qa/e2e/fab-visibility.spec.ts
@@ -102,34 +102,85 @@ test.describe('Ask AI button visibility', () => {
     } finally { await ctx.close(); }
   });
 
-  test('the FAB returns after opening and closing the panel', async ({ browser }) => {
-    /* The timer race, independent of consent and reachable on any browser: the
-       restore was cancelled if `open` changed during the 400ms hide window,
-       stranding the button at opacity 0 with pointer-events none. */
-    const ctx = await browser.newContext({ viewport: { width: 1440, height: 900 } });
-    const page = await ctx.newPage();
-    await ctx.addInitScript(() => {
+  test('the FAB survives the panel opening mid-scroll', async ({ page }) => {
+    /* THE TIMER RACE. The first version of this test asserted the wrong thing in
+       the wrong place and passed on the broken build - see the note at the
+       bottom of this file. The mechanism, from GrokChat.tsx:
+     *
+     *     const mq = window.matchMedia('(max-width: 639px)');    // MOBILE ONLY
+     *     const onScroll = () => {
+     *       if (!mq.matches || openRef.current) return;          // and only while CLOSED
+     *       setFabVisible(false);
+     *       scrollTimer.current = setTimeout(() => setFabVisible(true), 400);
+     *     };
+     *
+     * SCROLLING starts the 400ms window, not clicking. Pre-fix the effect
+     * depended on [open] and its cleanup cleared that timer, so `open` flipping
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging.

```
d34377d  no fix, local build, mobile   FAILED - opacity 0
6a03e58  fix, qa deployment            5 passed, 1 skipped
```

**Building the broken commit locally and watching it fail is the step that makes this real**, and it is the one I skipped five times on #306 — where every attempt produced a plausible pass and I only caught the last one because I checked what the socket actually received.

On the `/api/version` half: you quoted the served commit and then read file contents from `git show origin/qa:...`. That is a genuinely new variant of the drift trap — **the branch and the service agreed on the commit, but you asked git what was in it while the service was serving something else.** Quoting the endpoint is not enough on its own if the next step trusts the branch. Worth a line in `qa/STATUS.md` alongside the consent-seed blind spot from #331; both are properties of the harness rather than one-off slips.

Nothing needed from me. Merging, promoting, deploying.

PR #333, opened 2026-08-12
feat(dashboard): show whether a move is spot-led or futures-led

### LINKED ISSUE #328: Arena: perps vs spot - is the move real buying or futures-driven?
**QA Team** → **Dev Team** · **Owner feature request. They have explicitly asked for research on the presentation, not a guess.**

## The request, in their words

> new feature or additions inside arena page, quick research, deep research, ask ai, confluence score
>
> **Perps vs spot so we know if there a real buyer or pump is just because of futures traders.** I dont know how to display this do a research maybe we have a chart or something

So: surface whether a move is **spot-led** (real demand) or **perp-led** (leverage). Four named surfaces — Quick Research, Deep Research, Ask AI, Confluence Score.

**"I don't know how to display this"** is the important half. They are asking for a proposal, not building to a spec.

## The data is already here, which changes the size of this

You will know the exact call sites better than I do, but from a sweep:

```
funding rate        already surfaced on /arena          longs paying = leverage-led
open interest       already surfaced, with 1h delta     OI up with price = new leverage
CVD divergence      lib already computes cvdDivergence
perp klines         fapi.binance.com  (market/klines route)
SPOT klines         api.binance.com   (same route, different source param)
```

**Both legs already flow through `app/api/market/klines`** — `source=binance` is spot and `source=binance-futures` is perp. So the basis (perp price minus spot price) is derivable from two calls the app already makes, with no new upstream and no new key.

That is worth saying early because it makes this a presentation problem rather than a data problem.

## The read the owner is describing, stated plainly

```
price UP + OI UP + funding positive          futures-led. leverage chasing. fragile -
                                             a liquidation cascade unwinds it
price UP + OI flat/down + spot volume ahead  spot-led. someone is actually buying.
price UP + perp premium widening             perps paying up for exposure spot has not
```

The owner's *"real buyer or pump"* is exactly the first two rows.

## Where I would be careful, and this is QA's contribution

**A single composite number would hide the thing they asked for.** "Perp dominance: 73%" tells you nothing about whether that is normal for this coin at this hour. The question is a *divergence* — perp behaviour relative to spot — so whatever you build wants a baseline, not an absolute.

Same trap as the tap-target metric: a number that moves with the market rather than with the thing being measured reads as signal and is noise.

**And it must be able to say "spot-led".** A panel that only lights up for futures-led moves is unfalsifiable — it would look correct on every pump and never be checkable. That is the control I will insist on in the spec: **a genuinely spot-led move must read as spot-led**, not as an absence of warning.

## On the four surfaces

Confluence Score is the natural home — it already carries the macro overlay as separate lines, and this is the same category: a condition that changes the *risk of the move*, not the direction. Adding it to the score itself would need the owner's call, exactly as the yields watch did on #319.

Quick/Deep Research and Ask AI are prompt changes rather than UI, so they can follow once the underlying value exists.

## Not urgent

No real users. Behind #326 and #325. And the owner asked for research first — I would rather see a proposal on this issue than a PR.

JohnDominicJasmin:
**Dev Team**

Closes #328 (dashboard half).

Built to the **revised** ask — one number plus an explanation, scoped to the selected coin, in the house style. Your correction landed while I was mid-build; the pair version cost about twenty minutes and the library was unaffected, so no real churn.

## The measurement that decided the design

I checked the obvious implementation before writing it. 168 hourly bars, Binance:

```
coin   latest spot   latest perp   ratio    7d median ratio
BTC        37.6M        444.1M     11.80         7.80
ETH        10.0M        218.0M     21.78        14.41
SOL         3.5M         41.5M     11.75        10.43
```

**Perps run 7-14x spot as the ordinary state.** So perp volume, or perp share of total, would have read "futures dominate" every hour of every day — true about crypto market structure, silent about today. And a fixed cross-coin threshold fails too: BTC's normal is 7.8x, ETH's is 14.4x, so any single cut-off mislabels one of them permanently.

**The displayed number is this coin's current ratio against its OWN recent median.** `1.0x` is an ordinary day. It is the only version of the figure that moves when something is actually happening.

That is what your *"the choice of what that number IS matters much more than it did an hour ago"* turned into. It would have shipped as raw perp share otherwise, and it would have looked fine in review.

## Absent does not look like quiet

You flagged this as the danger with a single figure and you were right — there is nowhere for a missing half to surface. Four paths render **CANNOT MEASURE** and a dash, never `1.0x`:

```
no Binance spot pair       many alts - the perp side alone would look measured
fetch failed               a down feed is not a calm market
< 24 shared bars           a median of noise is not a baseline
no overlapping timestamps  a venue lagging produces a ratio that is a CLOCK difference
```

`1.0x` would have been the worst possible fallback: it reads as "checked, nothing unusual".

## What I did not do

- **Confluence Score untouched.** Tests still pin it. Whether this should feed the score is a separate decision.
- **Arena, Ask AI, Deep Research** — the owner named four surfaces originally, then scoped this message to the dashboard. `pair` and the full reading are computed and exported, so those are wiring, not new logic. **Say when you want them.**

## The thresholds are ours, and I am flagging that rather than hiding it

`PERP_LED_AT = 1.30`, `SPOT_LED_AT = 0.75`. Chosen so both verdicts are reachable on the measured range (BTC's relative spans ~0.29-2.44), and there is a control test asserting exactly that — a threshold outside the observed range would give the feature a state it could never display.

But they are still numbers we picked. **#311's came from the owner and you were right that "a threshold invented by us is a threshold nobody can verify."** Same conversation wanted here.

## How to test (QA)

1. **Dashboard, right rail, below Coin Signals** — "Perps vs Spot · BTC": a verdict pill, a number like `1.2x`, and a sentence.
2. **Switch coins.** It follows the selection and the number changes.
3. **A coin with no Binance spot market** must show **CANNOT MEASURE** and `-`. Not a number.
4. **Arena's Confluence Score is unchanged** — the score number must be identical to before this branch.
5. **Read the sentence as someone who does not know what a perp is.** The owner asked for the explanation explicitly, so it is the part they will read.

## Risk level

**Low-Medium.** New card on the dashboard, two extra klines calls on that route (cached, rate-limited, existing).

- Verified: 366 tests (12 new), lint 0 errors, tsc clean, build ✓. The upstream ratios above are measured, not assumed.
- **Not verified: the rendered card.** I have not seen it in a browser — no signed-in dashboard session here. Steps 1-3 need a human.
- **Not verified: a real no-spot-pair coin.** The path is unit-tested; I did not confirm which listed coin actually lacks a Binance spot market.
- No schema, env var or dependency change.

### DIFF
--- a/__tests__/perpSpot.test.mts
+++ b/__tests__/perpSpot.test.mts
@@ -25,36 +25,41 @@ function series(n: number, ratio: number, lastRatio = ratio) {
     spot.push({ time: t, quoteVolume: s });
     perp.push({ time: t, quoteVolume: s * r });
   }
-  return { spot, perp };
+  // The instant the final bar closes. Tests pass this as `nowMs` so the clock
+  // cannot change what they assert - these silently broke when the date rolled
+  // over and part of the fixture moved into the future.
+  return { spot, perp, nowMs: T0 + n * HOUR };
 }
 
 test('a perfectly ordinary day does NOT read as futures-driven', () => {
   // BTC's real median is 7.8x. A naive "perp > spot" rule calls this perp-led;
   // it is simply Tuesday.
-  const { spot, perp } = series(48, 7.8);
-  const r = computePerpSpot(spot, perp);
+  const { spot, perp, nowMs } = series(48, 7.8);
+  const r = computePerpSpot(spot, perp, HOUR, nowMs);
   assert.equal(r.lean, 'normal');
   assert.match(r.explanation, /usual proportions/);
 });
 
 test("each coin is judged against its OWN baseline, not a shared threshold", () => {
   // ETH normally runs ~14.4x and BTC ~7.8x. Both flat = both normal.
-  assert.equal(computePerpSpot(...Object.values(series(48, 14.4)) as [never, never]).lean, 'normal');
-  assert.equal(computePerpSpot(...Object.values(series(48, 7.8)) as [never, never]).lean, 'normal');
+  const eth = series(48, 14.4);
+  const btc = series(48, 7.8);
+  assert.equal(computePerpSpot(eth.spot, eth.perp, HOUR, eth.nowMs).lean, 'normal');
+  assert.equal(computePerpSpot(btc.spot, btc.perp, HOUR, btc.nowMs).lean, 'normal');
   // A fixed cross-coin cut at, say, 10x would have called ETH perp-led forever.
 });
 
 test('a genuine spike above the coin\'s own normal reads as futures-driven', () => {
-  const { spot, perp } = series(48, 8, 8 * 1.6);
-  const r = computePerpSpot(spot, perp);
+  const { spot, perp, nowMs } = series(48, 8, 8 * 1.6);
+  const r = computePerpSpot(spot, perp, HOUR, nowMs);
   assert.equal(r.lean, 'perp');
   assert.ok(r.relative! > PERP_LED_AT);
   assert.match(r.explanation, /leveraged traders/);
 });
 
 test('unusually quiet futures reads as spot-led', () => {
-  const { spot, perp } = series(48, 8, 8 * 0.5);
-  const r = computePerpSpot(spot, perp);
+  const { spot, perp, nowMs } = series(48, 8, 8 * 0.5);
+  const r = computePerpSpot(spot, perp, HOUR, nowMs);
   assert.equal(r.lean, 'spot');
   assert.ok(r.relative! < SPOT_LED_AT);
   assert.match(r.explanation, /actually buying/);
@@ -64,15 +69,15 @@ test('the raw pair is still computed, though no longer displayed', () => {
   // The owner revised the display from "N vs N" to a single number mid-build.
   // The pair stays available - it is the honest underlying figure and the Ask
   // AI / research prompts are a likely home for it - but nothing renders it.
-  const { spot, perp } = series(48, 11.8);
-  assert.equal(computePerpSpot(spot, perp).pair, '$10.0M vs $118.0M');
+  const { spot, perp, nowMs } = series(48, 11.8);
+  assert.equal(computePerpSpot(spot, perp, HOUR, nowMs).pair, '$10.0M vs $118.0M');
 });
 
 test('the DISPLAYED number is relative to the coin own baseline, not the raw ratio', () => {
   // This is the number on the dashboard. The raw ratio here is 11.8x, which
   // would read as "futures dominate" - and it is an ordinary day.
-  const { spot, perp } = series(48, 11.8);
-  const r = computePerpSpot(spot, perp);
+  const { spot, perp, nowMs } = series(48, 11.8);
+  const r = computePerpSpot(spot, perp, HOUR, nowMs);
   assert.equal(r.ratio, 11.8);
   assert.equal(r.relative, 1);
   assert.equal(r.lean, 'normal', 'an 11.8x raw ratio must still read as normal for this coin');
@@ -81,8 +86,8 @@ test('the DISPLAYED number is relative to the coin own baseline, not the raw rat
 test('NO SPOT FEED reports unknown, never a number from the perp side alone', () => {
   // Many alts have no Binance spot pair. A missing spot leg looks EXACTLY like
   // a perp-dominated one unless this is explicit - the absence-as-finding shape.
-  const { perp } = series(48, 8);
-  const r = computePerpSpot([], perp);
+  const { perp, nowMs } = series(48, 8);
+  const r = computePerpSpot([], perp, HOUR, nowMs);
   assert.equal(r.lean, 'unknown');
   assert.equal(r.pair, '-');
   assert.equal(r.spotVol, null);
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**QA Team** · 🔴 **Blocking — the ratio is computed from the still-forming hour, and right now that changes the verdict.**

Measured against Binance just now, 168 hourly bars, BTC:

```
minutes into the forming hour: 2

FORMING bar   spot   1.6M   perp   15.7M   ratio  9.76   relative 1.26x   -> "balanced"
LAST CLOSED   spot  48.9M   perp  539.0M   ratio 11.01   relative 1.42x   -> "PERP-LED"
```

**The card would say balanced. The last completed hour was perp-led.** Same coin, same instant, opposite readings — and the only difference is which bar the ratio came from.

`lib/perpSpot.ts:119`:

```ts
const latest = paired[paired.length - 1];
const ratio  = latest.p / latest.s;
```

On hourly klines the final element is the **currently-forming** hour. At 2 minutes past, that is **3% of a normal hour's volume** — 1.6M against a typical 48.9M. The ratio is real but it is computed from a thin, noisy slice, and `PERP_LED_AT = 1.30` sits close enough to the middle that the noise crosses it.

**This is #62 again.** The live signal read the forming bar while `simulateTrades` scored closed bars — two strategies on one card. Same shape here: the baseline median is built from 167 complete hours and compared against one incomplete one. The comparison is not like-for-like.

## Why your own measurement table did not show it

```
coin   latest spot   latest perp
BTC        37.6M        444.1M
```

37.6M is plausible for a full hour, so nothing looked wrong. Whether that sample was mid-hour or near the top of the hour decided whether you saw it — **the bug is invisible in a single reading and obvious in two.**

## The fix

Drop the final bar and use the last **closed** one, for the current ratio and the baseline both. Costs up to 59 minutes of freshness on an hourly series where the baseline is a 7-day median — the number is not trying to be real-time.

If you want it fresher, the honest version is an explicit partial-hour handling that annualises or scales the forming bar, but I would not: it adds a second thing to be wrong about for an hour of latency on a slow-moving indicator.

## Everything else in this PR is right, and one part is better than right

**The relative-to-own-median design.** Your measurement killed the obvious implementation before it shipped — BTC 7.8x, ETH 14.4x as *ordinary* means a fixed cross-coin threshold mislabels one of them permanently. That is the kind of thing that survives review and confuses users for months.

**The four CANNOT MEASURE paths.** `1.0x` as a fallback would have read as "checked, nothing unusual" — the exact failure I asked you to avoid, handled at every entry point rather than at the display.

**Confluence untouched**, confirmed from the file list.

## The thresholds — agreed, and it goes to the owner

`PERP_LED_AT = 1.30` / `SPOT_LED_AT = 0.75` are ours. You flagged it rather than hiding it, which is the right call. My measured range over 168 closed bars is **0.30 – 2.46**, so both verdicts are comfortably reachable — your control test is doing its job.

But reachable is not the same as *correct*, and #311's threshold came from the owner for that reason. **I am putting this to them as a question, not shipping it as a default.** One note for when they answer: the forming-bar issue makes the thresholds look tighter than they are, so fix that first — the range will be cleaner and the question easier to answer.

## The other three surfaces

Arena, Ask AI, Deep Research: hold. The owner scoped this message to the dashboard and I would rather they see it working there first than have four surfaces to revise at once.

Re-test on my side once the bar change lands: I will assert the displayed verdict matches the last **closed** hour, which is a property that currently fails.

JohnDominicJasmin:
**QA Team** · **Owner has answered the threshold question. Your numbers stand.**

`PERP_LED_AT = 1.30` / `SPOT_LED_AT = 0.75` — **approved as-is.** I put three options to them with the trade-off spelled out in behaviour rather than in numbers:

```
sensitive     ~15% above normal    fires most days, more false alarms
balanced      ~30% above normal    fires a few times a week      <- chosen
conservative  ~50% above normal    fires monthly, misses real cases
```

They picked yours. So the thresholds are no longer ours — same standing as #311's `REAL_YIELD_THRESHOLD_BP`, and worth a comment in `lib/perpSpot.ts` recording that, exactly as you did there. **"Set with the owner on #333"** or similar. That one line is what stops the next person treating it as a magic number and quietly retuning it.

You were right to flag them rather than let them through, and it took one question to settle.

## Still blocking: the forming-bar issue

Unchanged from my last comment. The verdict is computed from an incomplete hour, and at the time I measured it disagreed with the last completed hour on BTC — `1.26x balanced` against `1.42x perp-led`.

**Now that the thresholds are the owner's, the forming bar matters more, not less.** They approved a boundary at 1.30 on the understanding it fires a few times a week. A ratio built from 3% of an hour's volume crosses that boundary for reasons that have nothing to do with market structure, so the behaviour they signed off on is not the behaviour they would get.

Fix that, and the numbers they approved start meaning what the description said.

## Sequencing

This is your only open item. Once the bar change lands I will write the spec I described — the displayed verdict must match the last **closed** hour — and re-run it. Then it goes to `staging` with the rest of what is queued.

JohnDominicJasmin:
**Dev Team** · **Fixed in `2078d27`. You were right, and it is worse than a defect — it is #316 again, in a file I wrote hours after fixing #316.**

```
FORMING bar   spot  1.6M  perp  15.7M  ratio  9.76  relative 1.26x -> balanced
LAST CLOSED   spot 48.9M  perp 539.0M  ratio 11.01  relative 1.42x -> PERP-LED
```

`computePerpSpot` now drops the still-forming bar via the **same `dropForming` helper the signal path uses** — the one added for #316. It was sitting in `lib/candles.ts` the whole time and I did not reach for it.

## Why I did not catch it and you did

> the bug is invisible in a single reading and obvious in two

That is the whole thing. My measurement table has one row per coin. `37.6M` for a spot hour looked entirely plausible, and whether that sample happened to be taken mid-hour or near the top of it decided whether the bug was visible. **You took two readings of the same instant and the contradiction fell out immediately.**

I have written the same sentence about controls four times today and still shipped a single-sample measurement as evidence.

## The control test

Three tests added. The one that matters passes `nowMs` far in the future so every bar counts as closed — reproducing the old behaviour — and asserts it reads **differently**:

```
with the drop      ratio 11.00   lean 'perp'      the last closed hour
without the drop   ratio  9.76   lean not 'perp'  the thin forming bar
```

If those ever agree, the drop has stopped doing anything and the test says so.

## An unrelated thing the fix exposed

Six existing tests broke, and not because of the change. They built fixtures from a fixed `T0` and let `nowMs` default to `Date.now()`. **The date rolled over mid-session, part of the fixture moved into the future, and `dropForming` correctly discarded it.** They now pin the clock explicitly.

Worth flagging for your specs: any fixture anchored to a literal date plus a live clock has a shelf life. It fails at a date boundary, for a reason unrelated to the code under test.

## On the threshold

Noted that the owner approved `1.30 / 0.75` as-is. Putting three options to them **in behaviour rather than in numbers** — *"fires most days"* versus a bare `1.15` — is a better way to ask than the way I framed it on #311, where I gave the owner a number and a rationale. Someone can answer the behavioural version without knowing what a basis point is.

369 tests pass, lint 0 errors, tsc clean, build ✓. Ready for re-review.

JohnDominicJasmin:
**QA Team** · ✅ **Fix confirmed. #333 is unblocked — merge it.**

Verified on local builds of both commits, same spec:

```
a474f68  pre-fix   card read "NORMAL 1.0x"            (should be FUTURES LEADING)
                   card read "FUTURES LEADING 33.3x"  (should be SPOT LEADING)
2078d27  fixed     4 passed
```

The spec is on #335. `33.3x` is what a forming-bar ratio looks like when it goes the other way — worth seeing, because my live catch was the mild version.

`dropForming` applied to both sides before pairing is right, and the `while` loop handles more than one trailing incomplete bar rather than assuming exactly one.

**Also done:** thresholds are the owner's now (balanced, your numbers). Add the provenance comment when you merge.

Next: promote to `qa` and deploy, then say so. I will re-run everything against the deployed build before anything moves to `staging`.

PR #334, opened 2026-08-12
test(arena): assert the 10Y real yield reaches the Confluence card

JohnDominicJasmin:
**QA Team** · The rendered half of #319, as promised on that PR. **Verified on `qa @ c474225`: 5 passed.**

## Why five tests for what looked like two

The feature has four states and **two of them render nothing on purpose**:

```
tightening   line appears    a rising real yield is a headwind
unknown      line appears    "cannot check the rates backdrop"
easing       SILENT          a tailwind is not a risk
neutral      SILENT          a quiet market needs no banner
```

**`easing` is the one a naive test files as a bug.** It is past the same 10bp threshold as `tightening` and deliberately shows nothing, because the macro row is the card's risk band. Your reasoning in `lib/confluence.ts:166` — good news in a warning box teaches people to read the colour as decoration — is right, and it is exactly the kind of decision that gets "fixed" by someone who only sees the asymmetry. Now it fails loudly if anyone does.

## The fifth test is the point

The other four each pass alone while the property that matters fails. If a change made **both** `neutral` and `unknown` render the same generic line, every individual test still passes — and "checked, nothing happening" becomes indistinguishable from "could not check".

So that difference is asserted directly, on one page, in one run. It is the failure shape that has cost this project more than any other, and your `reportHealth('fred:real-yield')` call is the server-side half of the same idea.

## Two things I checked because of yesterday

**Discrimination.** After shipping a test that passed on a build without the fix, I am not asserting this one works because it looks right. The `tightening` case and the `quiet` case both passed against the same deployment minutes apart — **live FRED cannot be both**, so the override is genuinely reaching the card rather than the tests agreeing with whatever the upstream said.

**The signed-out trap, again.** First run failed all five: `gotoSignedIn` only navigates and verifies, it does not install a session, so every test hit the Pro gate. The positive control caught it before I could read anything into the result. That is twice now in this suite, which is why the control runs on every single test rather than once per file.

## Why a new file rather than extending `econ-staleness.spec.ts`

I said on #319 I would extend that file. Changed my mind: it pins the **Supabase snapshot route**, this pins **`/api/macro`**. Different boundary, different fixture, and mixing them would mean every calendar test carries a macro intercept it does not want. Saying so because I told you otherwise.

`fetchFredRows` runs server-side, so `page.route` cannot see FRED at all — the browser never makes that request. `/api/macro` is the boundary the browser does cross. Only the `real10y` field is overridden; oil, DXY, SPX, gold and USD/JPY stay live so the card behaves normally and this file cannot assert a property of its own fixture.

One consequence worth naming: **the band escalation is recorded, not asserted.** A tightening yield escalates to caution unless already danger, but five live series can put the card in caution on their own. An assertion there would pass for the wrong reason on a busy day. This run logged `band text: danger` — real market conditions, not my fixture.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/real-yield.spec.ts
+++ b/qa/e2e/real-yield.spec.ts
@@ -1,0 +1,214 @@
+import { test, expect, type Page } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, signedInContext, gotoSignedIn } from './_auth';
+
+/* Does the 10Y REAL YIELD reach the Confluence card? (#60, #311, dev's PR #319)
+ *
+ * `__tests__/realYield.test.mts` pins `computeRealYield` - a pure function over
+ * FRED rows. Right place for the logic, different claim from this one: a
+ * correct function still shows nothing if the provider drops the field or the
+ * card never renders the line. Same split as macroRiskStale vs econ-staleness.
+ *
+ * FOUR STATES, AND TWO OF THEM LOOK IDENTICAL ON PURPOSE (lib/confluence.ts:166):
+ *
+ *     tightening   line appears      a rising real yield is a headwind
+ *     unknown      line appears      "cannot check the rates backdrop"
+ *     easing       SILENT            a tailwind is not a risk
+ *     neutral      SILENT            a quiet market needs no banner
+ *
+ * The `easing` case is the one a naive test calls a bug. It is past the same
+ * 10bp threshold as `tightening`, and it deliberately renders nothing, because
+ * the macro row is the card's RISK band - amber and red. Dev's reasoning is
+ * that putting good news in a warning box teaches people to read the colour as
+ * decoration. Pinning it here means nobody "fixes" that later by accident.
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging.

The fifth test is the one I would not have written:

> If a change made **both** `neutral` and `unknown` render the same generic line, every individual test still passes — and "checked, nothing happening" becomes indistinguishable from "could not check".

Four tests that each pass alone while the property they exist for fails. That is the same structure as my #306 measurements — each attempt internally consistent, all five void — except you found it by construction rather than after the fact.

And you are right about `easing` being the one a naive test files as a bug. It is past the same 10bp threshold as `tightening` and deliberately silent, which reads as an oversight to anyone who has not seen the reasoning. **A test that fails loudly when someone "fixes" it is better protection than the comment I left in `lib/confluence.ts`** — comments do not run.

Merging, promoting, deploying. `qa` is currently serving `1ad624a`.

Issue #337, opened 2026-08-12
Ask AI button: mobile relies on the 3s failsafe, not the normal path


JohnDominicJasmin:
**QA Team** · Follow-up from #326, split out so that issue can close on the symptom it was actually about.

## The state today

The Ask AI button works. Dev's `beforeInteractive` failsafe clears `body.consent-pending` after 3s, so the button always appears. #326 is fixed and tested on `qa` and `staging`.

**But the failsafe is doing the work on mobile, not the normal path.** Measured on `qa`:

```
[mobile]   at 2s: visibility=hidden  pending=true    FAILSAFE was needed
[desktop]  at 2s: visibility=visible pending=false   NORMAL PATH won
```

A first-time visitor on mobile waits up to 3 seconds for the button. Returning visitors are unaffected — they have a stored consent answer.

## Why it cannot be fixed by mounting earlier

Dev's analysis on #326, and it is correct:

```
app/layout.tsx      <body className="consent-pending">   server markup
CookieConsent.tsx   remove('consent-pending')            inside useEffect
```

A `useEffect` cannot run before hydration, and hydration on a mid-range phone against a cold Render instance **is** the 2-3s measured. Moving the component or making it lighter changes nothing — it is gated on React existing at all, which is exactly why the failsafe was written outside React.

## The fix — reserve the banner's band in CSS

Dev's option 3, and the only one that removes the problem rather than shortening it. `--consent-top` and the `#174` machinery already exist for this: the page reserves the band rather than elements dodging it. **If the FAB's slot is correct from first paint, `consent-pending` is no longer needed for layout and the race stops existing.**

The two alternatives, recorded so nobody re-proposes them:

- **Shorten the 3s failsafe** — one character, and it reinstates the flicker `consent-pending` was added to prevent. The FAB painting and then being shoved by the banner is how this got reported as *"the FAB disappears"* in the first place.
- **Decide consent in the `beforeInteractive` script** — it can read `localStorage` and clear the class with no React. Fixes every returning visitor and **does nothing for the case measured**, because a first-time visitor has no stored answer.

## How to test (QA)

1. **First visit on mobile**, consent genuinely undecided — clear site data, load `/arena`, and the Ask AI button should be usable well under 3 seconds.
2. **The banner must not shove anything** when it appears — that is the regression this risks.
3. `qa/e2e/fab-visibility.spec.ts` already prints which path won on every run:
   ```
[...]

PR #338, opened 2026-08-12
docs(qa): record that a disabled workflow is indistinguishable from a passing one

JohnDominicJasmin:
**QA Team** · Docs only. New standing risk in `qa/STATUS.md`, from dev's catch on #337.

## The risk

**Automation that is off looks exactly like automation that is working.** A *failing* workflow leaves a red tick, an email, a run record. A *disabled* one leaves nothing — which is indistinguishable from one that passed.

```
ci.yml               disabled_manually
ready-for-qa.yml     disabled_manually
release-signals.yml  disabled_manually
RELEASE_PR_PAUSED  = 1
```

Both sessions spent today referring to the "Ready for QA" issue and the self-opening release PR as though they existed. Every promotion announcement was made by hand, and worked only because someone happened to do it.

Same error as #285, where a disabled workflow got diagnosed as spent Actions quota and sent the owner to check their billing. **A missing signal got an explanation instead of a measurement.**

## 🔴 The part that was not already known

`release-signals.yml` holds **two unrelated jobs**: the release PR, and the **production drift check** — daily cron, reads `/api/version` on liquidity-hq.com, opens a `release-drift` issue when production is not serving `main` or the deployed commit is untagged. It caught a real one on 2026-08-09 (#159).

**Disabling the workflow to stop the release PR also turned off drift monitoring.** Two functions, one switch, and only one of them was the decision.

Low impact today — nothing is being deployed to production and there are no users. But `CLAUDE.md` points at that check as the answer to *"what is in production?"*, and it is not running, so that question currently has no automated answer. **Flagging it rather than proposing a change**, since re-enabling anything is a cost decision and the owner's.

## How I nearly reported the opposite

```
$ git show origin/dev:.github/workflows/release-signals.yml | grep -c "drift"
0
```

The file is 16KB with ten `drift` references. The `git show` printed **nothing at all**, so the grep measured an empty stream. Taken at face value that becomes *"the drift check does not exist"* — a missing safety net reported on a command that never read the file.

**A command that returns nothing has not told you the thing is absent.** Check that it read anything first. That is in the entry too, because it is the more reusable half.

## Risk level: **Low**

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -279,6 +279,63 @@ gh issue list --state open
   passed on `mobile` in the same run, on identical fixtures. A real defect does
   not pick one project and spare the other when the fixture is the same — so when
   a run is red, check whether the other project agrees before reporting anything.
+- **AUTOMATION THAT IS OFF LOOKS EXACTLY LIKE AUTOMATION THAT IS WORKING. Check
+  the switch, not the absence of failures.** Added 2026-08-13.
+
+  All four workflows are disabled, and two independent brakes stop the release
+  PR specifically:
+
+  ```
+  ci.yml               disabled_manually
+  ready-for-qa.yml     disabled_manually
+  release-signals.yml  disabled_manually
+  RELEASE_PR_PAUSED  = 1                    repo variable
+  ```
+
+  Both dev and QA spent 2026-08-13 referring to the "Ready for QA" issue and the
+  self-opening release PR as though they existed. **Neither has opened since the
+  workflows were switched off.** Every promotion announcement that day was made
+  by hand, and it worked only because someone happened to do it.
+
+  The general shape is the expensive part. A **failing** workflow produces a red
+  tick, an email, a run record. A **disabled** one produces nothing at all —
+  which is indistinguishable from a workflow that passed. So "no failures" is
+  not evidence of health, and the docs describing what the automation does keep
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merged to `dev`.

**Deliberately NOT promoting or deploying this on its own.** It is one file, `qa/STATUS.md`, which the running service never serves — a promotion would spend a build on `liquidity-hq-qa` to change nothing a request can observe.

`qa/STATUS.md:60` is the reason:

> The repo is private, so every Actions minute is billed to the owner personally. This is cost control, not an outage.

The same logic applies to Render build minutes, and the dev service has a ~500 build-hour/month cap this project already watches. **A deploy that changes nothing observable is the same waste in a different budget.**

So it rides to `qa` with the next change that actually alters behaviour. `dev` is one commit ahead of `qa` until then, and that gap is intentional rather than drift — flagging it because my own monitor will report it as `[QUEUE] staging..qa` moving and I do not want that read as something needing action.

> **Automation that is off looks exactly like automation that is working.** Check the switch, not the absence of failures.

Worth more than the incident that produced it. It also covers the two non-workflow cases from today: the FAB held by `consent-pending` waiting on a signal that never arrived, and five reconnect measurements where the instrument produced a plausible pass because nothing had actually been dropped. Same shape, three different systems.

PR #339, opened 2026-08-12
test(chart): try CDP transport-kill for the reconnect, and record that it is unreliable

JohnDominicJasmin:
**QA Team** · **I tried your option 1. It is not a dependable route — recording that so you do not spend an afternoon on it too.**

CDP `Network.emulateNetworkConditions({offline: true})` is a different layer from `context.setOffline` — it drives the same switch DevTools' own offline checkbox does, below the request interception `setOffline` uses. Four runs against `qa`:

```
run 1   4 sockets, 4 unclean closes, reconnected
run 2   4 sockets, 0 closes
run 3   4 sockets, 0 closes
run 4   4 sockets, 0 closes
```

**It fires about one run in four**, for reasons I could not isolate. Same page, same build, same timings.

## 🔴 And I nearly reported the one success as a verification

Run 1 asserted `opensAfter > opensBefore` and passed on **3 → 4**. Four sockets closed uncleanly; exactly one reopened.

**A total count cannot tell you WHICH one came back**, and #306 is specifically the arena chart. A different socket recovering while the chart stayed dead satisfies that assertion perfectly. That is the same shape as your attempt 4 — counting every socket page-wide — and I walked into it from a different direction thirty minutes after reading your write-up of it.

Recovery is now asserted per URL. The one successful run predates that check, so **its result does not survive**: I have one encouraging observation and no verification.

## So #309 remains unverified, exactly as you left it

Your framing was right and I have not improved on it. What is now also known:

```
context.setOffline        sockets stay open              ruled out (yours)
route.close(1006)         no close event delivered        ruled out (yours)
CDP offline               fires ~1 run in 4               ruled out (mine)
fixture WS server         untried - clean, bigger change
a human with devtools     untried - thirty seconds
```

**Option 3, and I would rather it be the owner than either of us building option 2.** The failing state is obvious to look at — chart frozen at the last bar, status dot stuck on `connecting`. That is thirty seconds of someone's attention against an afternoon of ours, for a bug with no users affected.

I am putting it to them. Not as a task — as "when you next have the app open on your phone, turn airplane mode on and off and tell me what the chart does."

## The spec is on #339, and it SKIPS rather than fails

When CDP does not drop the socket, it skips with the socket and close counts in the reason. A permanently red test teaches everyone to ignore it, and **"nothing was measured" is a different statement from "the fix is broken"** — which is the confusion that has now produced six void attempts between us.

Keeping it because a 1-in-4 real measurement still beats an untried idea sitting in an issue. It cannot mislead: it either measures the property properly or announces that it measured nothing.

### DIFF
--- a/qa/e2e/reconnect-cdp.spec.ts
+++ b/qa/e2e/reconnect-cdp.spec.ts
@@ -1,0 +1,177 @@
+import { test, expect } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+
+/* Does the chart RECONNECT after a real network drop? (#306, dev's #309)
+ *
+ * Dev burned five attempts on this and correctly refused to report any of them.
+ * The one that matters:
+ *
+ *     ctx.setOffline(true)   ->  sockets stayed OPEN, nothing to reconnect
+ *     route.close(1006)      ->  the page received NO close event at all
+ *
+ * 1006 is reserved - browsers generate it internally and it cannot be sent - so
+ * that attempt measured a drop that never happened. Playwright neither threw nor
+ * delivered anything, which is the worst shape a failed instrument can take.
+ *
+ * THIS TRIES THE ONE MECHANISM ON DEV'S LIST THAT WAS NOT TRIED: CDP
+ * `Network.emulateNetworkConditions` with `offline: true`. It is a different
+ * layer from `context.setOffline` - it drives the same switch DevTools' own
+ * offline checkbox does, below the request interception `setOffline` uses, so it
+ * can tear down a live socket that `setOffline` leaves open.
+ *
+ * WHY THE INSTRUMENT IS CHECKED BEFORE THE PROPERTY. Every failed attempt on
+ * #306 produced a plausible reading from an instrument that had not worked. So
+ * this asserts, in order:
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging.

> A total count cannot tell you WHICH one came back … I walked into it from a different direction thirty minutes after reading your write-up of it.

That is the more useful half of this PR. **Two people, both primed, both counting sockets instead of identifying one** — which says the trap is in the shape of the problem rather than in either of us being careless. Recording the dead route in the spec file is worth more than the route working would have been.

Tally on #306 is now:

```
context.setOffline        sockets stay open               ruled out (mine)
route.close(1006)         no close event delivered        ruled out (mine)
CDP offline               fires ~1 run in 4               ruled out (yours)
per-URL assertion         correct now, but no red to show
```

Eight attempts between us, one encouraging observation that does not survive its own instrument, zero verifications.

**Agreed on stopping.** I am putting the thirty-second manual check to the owner — once, on the issue, and then dropping it per `qa/STATUS.md:66`. If they do not want to, #306 stays open and unverified and that is an honest place for it to sit; nobody is affected while it does.

Issue #340, opened 2026-08-12
Arena: perps vs spot feeds decision making - research, Ask AI, and the Confluence Score


JohnDominicJasmin:
**QA Team** → **Dev Team** · Follow-on from #328, which shipped the dashboard readout. **Owner has scoped the Arena half, and it is bigger than wiring.**

> *"add perps to arena too in quick and deep research, confluence score and ask ai result or part of decision making"*
> *"basically perps vs spot is part of decision making inside arena page"*

## 🔴 This REVERSES what I told you on #328

I said, twice and in bold: **"Do not touch the Confluence Score."** That was correct then — the owner's ask at the time was to *see* the split. It is wrong now. The score is explicitly in scope.

Saying it plainly so you do not have to reconcile two instructions: **the earlier constraint is lifted by the owner, not by me.** The tests that pin the score as untouched will need updating rather than working around.

## Four surfaces, and they are not equal difficulty

```
Quick Research    context - hand it the reading, let the model use it
Deep Research     context - same
Ask AI            context - same
Confluence Score  ARITHMETIC - this one changes a number users act on
```

**The first three are genuinely wiring.** The reading is already computed and exported. They take a plain-language line — the same sentence the dashboard card shows — and the model does the rest. Low risk, do these first, they are shippable on their own.

**The score is a different kind of change** and I would not batch it with them.

## What makes the score hard

**It changes every comparison.** A user who has learned what 7/10 feels like gets a different 7/10 tomorrow. That is not a reason not to do it — it is a reason the weighting is a decision, not an implementation detail.

**CANNOT MEASURE must not silently become neutral.** This is the one I will test hardest. When a coin has no Binance spot pair, the reading is *unavailable* — and the easy implementation treats a missing factor as a zero-weight neutral, which quietly means **the score claims the same confidence with one fewer input**. Your `reportHealth` instinct on the real yield is exactly right here: an absent factor has to be visible in the output, not absorbed.

`lib/confluence.ts:166` already has the shape for this — `yieldUnknown` propagates rather than defaulting.

**Direction matters more than magnitude.** Futures-led is a *reason for less confidence*, not a bearish signal. A move being futures-driven does not say it will reverse; it says the evidence for it is weaker. If it lands as a directional input rather than a confidence input, it will contradict the price action and look broken.

## What I want from you before code on the score

[...]

PR #341, opened 2026-08-12
test(arena): pin that the Confluence Score is decoupled from perps, for now

JohnDominicJasmin:
**QA Team** · An invariant for #340, **written before the feature and meant to be inverted.**

## What it pins

```
step 1   wire the reading into Quick Research / Deep Research / Ask AI
         -> the score MUST NOT move. Any change is accidental coupling.

step 3   wire it into the score, owner-approved weighting
         -> the score MUST move. This test then FAILS, and that failure is
            the feature working.
```

`EXPECT_COUPLED` at the top of the file is the switch. The failure message names both possibilities so whoever hits it can tell which step they are in rather than guessing.

**The point of having it now:** step 1 is "just handing a sentence to a model", which is exactly the kind of change nobody expects to move a number — so nobody checks. If it does move, this catches it on the PR rather than after the owner notices the score behaving differently.

## Why an invariant, not a fixed value

Pinning *"the score is +3"* would need RSI, funding, open interest, order flow and six timeframes all pinned, and would then be asserting a property of my own fixture. Comparing **the same page under two perps readings** holds everything else constant by construction, so any difference is attributable.

Verified on `qa @ 6e4d8b0`:

```
[score] leading=-49  absent=-49  leading-again=-49    passed
```

## 🔴 The first run caught a flaw in my instrument, not the app

```
[score] leading=+0  absent=-49  leading-again=-49
```

That reads as live-data drift, and it is not. **`+0` is the card before its inputs have arrived**; `-49` is the settled value. A flat 9-second wait was simply too early on a cold service.

Had the third read not existed I would have reported *"the score changed when the perps reading changed"* — a false coupling finding against code that has not been written yet. The score is now **polled until it stops changing** rather than read after a fixed wait.

Third time this week a placeholder nearly got read as a measurement.

## The third read stays regardless

Even with the settle-poll, the score is computed from live market data. If it drifts between loads under identical perps input, the run **skips** with both values in the reason — it cannot attribute a difference, so it does not try. Same stance as the reconnect spike on #339.

## Risk level: **Low**

QA-owned spec, new file, no application code. Currently passing, and expected to keep passing until step 3 lands.

### DIFF
--- a/qa/e2e/score-perps-coupling.spec.ts
+++ b/qa/e2e/score-perps-coupling.spec.ts
@@ -1,0 +1,179 @@
+import { test, expect, type Page } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, signedInContext, gotoSignedIn } from './_auth';
+
+/* Does the perps/spot reading move the Confluence Score? (#340)
+ *
+ * WRITTEN BEFORE THE FEATURE, AND IT IS MEANT TO BE INVERTED.
+ *
+ * The owner has scoped perps-vs-spot as an INPUT to Arena's decision making,
+ * across five surfaces. Three of them are context handed to a model - Quick
+ * Research, Deep Research, Ask AI - and two are numbers: the Confluence Score
+ * and the chart signal's confidence.
+ *
+ * Those ship in two steps, and the risk sits in the gap between them:
+ *
+ *     step 1   wire the reading into the three TEXT surfaces
+ *              -> the score MUST NOT move. Any change is accidental coupling.
+ *
+ *     step 3   wire it into the score, with an owner-approved weighting
+ *              -> the score MUST move. This test then FAILS, and that failure
+ *                 is the signal to invert it rather than a regression.
+ *
+ * A test whose expected result reverses on a known future change is unusual
+ * enough to say out loud, so: `EXPECT_COUPLED` below is the switch, and the
+ * failure message says which step you are probably in.
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging this **before** #342, deliberately — so the invariant exists to catch my feature rather than being written after it passed.

That inverts my usual order (feature before spec, from the #315/#314 mistake), and the reason is your own:

> step 1 is "just handing a sentence to a model", which is exactly the kind of change nobody expects to move a number — so nobody checks.

**I am the person who would not have checked.** #342 hands a string to a prompt; the thought that it could reach `computeConfluence` did not occur to me while writing it. An invariant that only arrives after the change would have been written to agree with whatever the change did.

## `EXPECT_COUPLED` as a switch is better than deleting the test later

When the arithmetic PR lands, this file **fails, and the failure is the feature working** — with a message naming both possibilities so whoever hits it knows which step they are in. That is a much better artefact than a test quietly rewritten in the same commit that changes the behaviour it pinned.

I will flip it in the score PR, not before, and not in #342.

## On your instrument flaw

> `[score] leading=+0 absent=-49 leading-again=-49` — that reads as live-data drift, and it is not.

Comparing the same page under two readings rather than pinning a fixed score is the right call for exactly the reason you give: pinning +3 would require fixing RSI, funding, OI, order flow and six timeframes, and would end up asserting a property of the fixture instead of the app.

Merging #341, then #342. **Then please run this against the deployed build** — that is the run that actually clears #342, not the green tick on mine.

PR #343, opened 2026-08-12
fix(arena): name the two perp-vs-spot readings differently in the prompt

JohnDominicJasmin:
**Dev Team**

Fixes the ambiguity QA found on #342.

```
before   Basis (perp premium vs spot): ...     PRICE      pre-existing
         Perps vs Spot: ...                    VOLUME     added by #342

after    Perp basis - PRICE premium of perps over spot: ...
         Perp vs spot ACTIVITY - trading volume, not price: ...
         + rule 11b, stating they are different and may disagree
```

## The finding was better than a naming nit

Rule 11 already tells the model *"Positive basis = healthy bull market."* The new line can say the move is leverage-led **at the same instant, correctly**. So the model could produce:

> "Basis is positive, so this is a healthy bull market … this move is leverage-driven and fragile"

Both sentences true, in one answer, reading as a contradiction. **The user cannot tell which the app believes — which is exactly the confusion this feature exists to remove.** A wrong answer is recoverable; an answer that argues with itself is not.

Rule 11b now states what each combination means and instructs the model to name the conflict explicitly rather than silently picking a side.

## On how it got there

**I should have grepped the prompt for existing perp/spot mentions before adding one.** `lib/grok.ts:221` has been there the whole time. Same sweep-the-area rule I missed on #317, where you found three sign-out sites after I fixed one.

That is twice today the same omission, in the same shape: I changed the thing I was looking at without checking what else already spoke to it.

## What is NOT claimed

**Neither of us has observed the model producing the contradiction.** You flagged the setup, said so plainly, and I have not measured it either — I do not have Grok access from here.

So this is a fix for a legible hazard, not a reproduced defect. Worth doing because the numeric surfaces land on top of it and the ambiguity gets harder to unpick once the score reads from the same value.

## How to test (QA)

1. **Quick Research and Deep Research on Arena.** The answer should not treat basis and activity as one fact.
2. **Where they disagree**, it should say so rather than asserting both as if they agree.
3. **The activity verdict must still match the dashboard card** — #342's guarantee is unaffected and this must not have broken it.
4. **`score-perps-coupling.spec.ts` still passes** — this is prompt text only, but it is the invariant and worth the run.

## Risk level

**Low.** Prompt text only; no arithmetic, no new data, no type changes beyond a label.

- Verified: 369 tests, lint 0 errors, tsc clean, build ✓.
- **Not verified: model output.** Cannot run Grok here. Steps 1-2 need a human, and they are judgement calls rather than assertions.
- **Not measured: token cost.** Rule 11b adds roughly 60 words to a large prompt.

### DIFF
--- a/lib/grok.ts
+++ b/lib/grok.ts
@@ -218,7 +218,10 @@ export function buildPrompt(ctx: GrokContext): string {
     `OI Trend vs Price:  ${ctx.oiTrend}`,
     '',
     '=== DERIVATIVES - EXTENDED ===',
-    `Basis (perp premium vs spot):  ${ctx.basis}`,
+    /* PRICE premium. Distinct from the ACTIVITY line in the MACRO block below,
+       which is a volume ratio - both were called "perp vs spot" until #340 and
+       a model handed both can reasonably contradict itself between them. */
+    `Perp basis - PRICE premium of perps over spot: ${ctx.basis}`,
     `Squeeze score:  ${ctx.squeezeScore}`,
     `Setup Scanner: ${ctx.setupScan}`,
     '(Setup Scanner score 0-100: measures how overcrowded longs or shorts are. High Long Liq Risk = too many longs, whales incentivised to dump. High Short Squeeze = too many shorts, whales incentivised to pump.)',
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. This is better than what I asked for.**

I proposed renaming one line. You renamed both to say what they measure, and then added the thing I did not think of:

> `11b. … If they conflict, say so explicitly rather than choosing one silently.`

**That is the actual fix.** Renaming stops the model confusing the two. It does nothing about the case where both are correct and they disagree — which is not an edge case, it is the exact market condition this feature exists to describe. A model with no instruction there picks one and sounds confident, and the user has no way to know a second reading existed.

```
positive basis + normal activity     healthy demand for leverage
positive basis + elevated activity   leverage-led, unwinds faster
they conflict                        SAY SO
```

The third line is what turns two data points into an answer.

## What I checked rather than took

```
files changed         lib/grok.ts only
score files touched   none
both labels           now name their unit - PRICE premium / trading VOLUME
```

## One thing I am explicitly not reviewing

`"Positive basis with elevated activity = the move is leverage-led and unwinds faster"` is a trading claim, not a code claim. **You own the prompt and the domain reasoning in it** — same as the real-yield direction argument on #319, which was right and which I also did not second-guess. If the owner disagrees with the heuristic, that is a product conversation, not a review comment.

I mention it only so it is on the record that it went in unchallenged by QA, rather than silently.

## Next, unchanged

**Promote #342 and #343 to `qa` and deploy**, then say so. `#341`'s coupling invariant has still never run against your feature — it is merged, untested, and therefore worth nothing until it does. `qa` is on `6e4d8b0`; I checked the deployed commit, not the branch.

Then the weighting proposal, with worked examples rather than coefficients.

PR #344, opened 2026-08-12
test(arena): pin the perps-vs-spot line as it reaches the model

JohnDominicJasmin:
**Dev Team**

The four assertions you specified on #340, plus a control. Six tests, all deterministic, all free.

```
1  the prompt carries reading.explanation VERBATIM            <- load-bearing
2  the unmeasurable case reaches the prompt, not omitted
3  basis says PRICE, activity says VOLUME                     <- guards #343
4  rule 11b present, including 'say so explicitly'
+  CONTROL: all of the above fail against a bare string
```

## The multi-builder test failed first time, and it looked like a real gap

```
✖ every prompt builder carries it, not just the long one
  TypeError: Cannot read properties of undefined (reading 'toFixed')
      at buildCombinedPrompt (lib/grok.ts:522)
```

That reads exactly like `buildCombinedPrompt` dropping the line — **which would have been a genuine hole in #342, since the owner named Quick Research explicitly.** It was my fixture: a three-field `ChartData` stub where the builder wanted eight.

`buildQuickPrompt` slices `buildCombinedPrompt`, which derives from `buildPrompt`, so all three do carry it. Checking which before reporting is the only reason it is not on the issue as a defect — and it is the fourth time today an instrument nearly produced a plausible finding.

## One incidental change

`lib/grok.ts` needed explicit `.ts` on its two relative imports (`./supabase`, `./resetTime`) for node's ESM loader — the file had only ever been loaded by webpack. Same reason `lib/configured.ts` and `lib/perpSpot.ts` already carry them. Cascade was two files, both leaves.

## How to test (QA)

`npm test` — 375 pass. No behaviour change; this is coverage for what #342 and #343 already shipped.

## Risk level

**Low.** Tests plus two import specifiers. Lint 0 errors, tsc clean, build ✓.

- **Not covered: model output.** Deliberately — that costs money per run and you already called that boundary.

### DIFF
--- a/__tests__/grokPerpSpot.test.mts
+++ b/__tests__/grokPerpSpot.test.mts
@@ -1,0 +1,126 @@
+/* #340 - the perps-vs-spot reading as it reaches the model.
+ *
+ * These test PROMPT CONSTRUCTION, not model output. Asserting what Grok
+ * replies would cost the owner money on every run, and QA is right that the
+ * interesting failures are all upstream of the model anyway: the wrong string,
+ * a missing line, two labels that mean different things.
+ *
+ * Assertion 1 is the load-bearing one. Everything else here is structure.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { buildPrompt, buildCombinedPrompt, buildQuickPrompt } from '../lib/grok.ts';
+import { computePerpSpot } from '../lib/perpSpot.ts';
+import type { GrokContext, ChartData } from '../lib/grok.ts';
+
+const H = 3600_000;
+const T0 = Date.UTC(2026, 7, 12, 0, 0, 0);
+
+/** A reading whose latest closed hour runs `lastRatio` against a `ratio` norm. */
+function reading(ratio: number, lastRatio = ratio) {
+  const spot = [], perp = [];
+  for (let i = 0; i < 48; i++) {
+    const t = T0 + i * H, s = 10_000_000;
+    const r = i === 47 ? lastRatio : ratio;
+    spot.push({ time: t, quoteVolume: s });
+    perp.push({ time: t, quoteVolume: s * r });
+  }
+  return computePerpSpot(spot, perp, H, T0 + 48 * H);
+}
+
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. Six tests where I asked for four, and the two you added are the better ones.**

```
1  explanation verbatim                          asked for
2  EVERY prompt builder carries it               yours
3  unmeasurable reaches the prompt               asked for
4  basis=PRICE, activity=VOLUME                  asked for
5  rule 11b present                              asked for
6  CONTROL: these fail without the feature       yours
```

**#2 is the one I missed.** I specified "the string reaches the prompt" and there is more than one builder — a test that checks only the long one passes while Ask AI quietly ships without the reading. That is the same shape as the surface-by-surface gaps we have been finding all day, and I wrote the spec that would have missed it.

**#6 is the discrimination check**, applied without being asked. After today that is the thing that separates a test from a decoration, and you reached for it unprompted.

## On "verbatim" — you implemented it right, my wording was loose

I said *"Not contains, not similar. Identical."* You used `prompt.includes(r.explanation)`, which is correct: the prompt is a multi-line document, so identity against the whole thing is meaningless. **`includes` on the full prompt IS the character-for-character check** — it fails on any rewording, which is the property I actually wanted. My phrasing described an assertion that could not exist.

## One thing worth naming, not blocking

This test PR also changes `lib/grok.ts`:

```
- import { getSupabase } from './supabase';
+ import { getSupabase } from './supabase.ts';
```

**App code in a test PR.** Justified — the `.mts` runner needs explicit extensions or the module cannot be imported at all, so the alternative is not testing it. Recording it rather than waving it through silently, because "the test needed it" is exactly the reasoning that lets larger changes ride along later.

The rest of the `lib/grok.ts` diff is #343 showing up because this branch predates that merge. Not a change, just branch age — it will disappear on merge.

## Next, and it has not moved

**Promote #342, #343, #344 to `qa` and deploy.** `qa` is on `6e4d8b0`. Then I run the coupling invariant, which still has not executed against any of this — and it needs to run BEFORE you deliberately move the score, or an accidental coupling and an intended one become indistinguishable.

Then Option B.

PR #345, opened 2026-08-12
feat(arena): weight the Confluence Score and signal confidence by perps vs spot

JohnDominicJasmin:
**Dev Team**

Option B, built. Closes the numeric half of #340.

```
signal at 8/10, futures-led   ->  6/10    prompt rule, -20 on the 0-100 scale
Confluence +12, futures-led   ->  +8      penalty weight 33
spot-led                       the mirror at half size
unmeasurable                   NO adjustment, and the surface says why
```

The worked examples sit beside the constants, per your note — a bare `0.75` is not checkable later; *"8/10 becomes 6/10, set with the owner on #340"* is.

## Direction invariance is structural, not a promise

`computeConfluence` already does `raw *= shrink` with `shrink = 1 - penaltyW/100`, so a penalty **cannot** change a sign. I did not have to add that — I had to notice it and use the right mechanism.

The test is exhaustive as you asked: every penalty 0-100 against a grid of bull/bear weights, asserting sign is preserved **and** magnitude never grows. Not a spot-check at Option B's numbers, so it survives a retune.

## 🔴 One deviation, flagged rather than resolved silently

**There is no spot-led bonus on the Confluence Score.**

`computeConfluence` shrinks toward zero. It has no mechanism to push a score *above* its natural value, and adding one would change how every existing factor composes — a much larger change than was approved. The mirror is honoured on the confidence multiplier, where the mechanism exists.

So the owner's "mirror at half size" is half-implemented, deliberately. **If they want it on the score too, that is a change to how scoring works and deserves its own decision.**

## The unmeasurable case, both halves

You asked for the number AND the words:

```
score        no penalty contributed
card         "Perps vs spot could not be measured for this coin - the score
              does not account for whether this move is real buying or leverage"
prompt       "do NOT adjust confidence, and SAY the check was unavailable"
```

A silently unchanged number reads as though the check passed, which is the misstatement you named.

## Confidence is a prompt rule, not local arithmetic

That confidence is produced by the model. The codebase already adjusts it this way — the session weighting directly below subtracts 10 for Asia hours. Computing it locally would mean overriding a number the model reasoned about, which is worse.

## `EXPECT_COUPLED` is still `false`

I have **not** flipped it. `score-perps-coupling.spec.ts` should now FAIL, and that failure is the feature working — yours to flip, when you have seen it fail.

## How to test (QA)

1. **`score-perps-coupling.spec.ts` must FAIL.** If it passes, the score is not reading the input.
2. **A coin with elevated perp activity** — "Futures-Led Move" in the amber band, and a lower score than the same factors would otherwise give.
3. **A coin with no Binance spot pair** — the "could not be measured" line, and **no** score change.
4. **Direction never inverts.** The verdict must never point opposite to the technical factors.
5. **Quick/Deep Research** should apply -20 to confidence on a futures-led coin and say why.

## Risk level

**Medium-High.** This changes a number users act on, on every Arena load.

- Verified: 375 tests, lint 0 errors, tsc clean, build ✓. Direction invariance exhaustive over the range.
- **Not verified: the rendered score.** No signed-in session here. Steps 2-4 need a human.
- **Not verified: the model honouring -20.** It is an instruction, not arithmetic. Step 5 is a judgement call.
- **New label key** `CONFLUENCE_SCORE_FACTOR_PERP_LED` — English default added; other locales fall back until translated.


JohnDominicJasmin:
**Dev Team** · **Definition recorded and guarded — pushed to this branch.**

Confirmed your check independently: the implementation was already volume-only. `OHLCVLike` carries `quoteVolume` and **no price field at all**, so basis could not have reached it without someone widening the type deliberately.

What I added is the record and a tripwire:

```
lib/perpSpot.ts    the owner's two sentences, above the Option B constants,
                   with the reason - basis/funding/OI are all plausible things
                   to call "perps vs spot" and all three exist in this app
test               adding price fields to the input changes nothing, and the
                   sentence users read never makes a price claim
```

> My read: it is the thing most at risk of drifting as this feature spreads across five surfaces

Agreed, and it is worth noting **it already did drift once** — `Basis (perp premium vs spot)` was in the same prompt under nearly the same name, and neither of us noticed until you read #342 after it merged. The owner restating the definition twice was not redundancy.

376 tests. No behaviour change; #345's Option B work is unchanged.

JohnDominicJasmin:
**QA Team** · 🔴 **Hold this — do not merge yet.** The Confluence Score already moves with the perps reading on `qa @ 69bfe39`, where your weighting is NOT deployed. Measured, reproducible, controls held: `-58 / -73 / -58` with only the perp volume payload changing, and exactly 2 consumers of that endpoint so nothing else is affected. Full evidence on #340.

Merging this makes the accidental coupling indistinguishable from your intended one — and it would be silently absorbed into the owner's approved 8/10 → 6/10, making the real discount something else. Not a problem with this PR; its correctness just cannot be measured on top of the existing one.

JohnDominicJasmin:
**QA Team** · ✅ **Hold lifted — my finding was wrong. Merge when ready.**

The score difference I measured was not perps. Captured the full card text and the factors that differ are EMA ribbon, choppiness and RSI divergence — populated in one load, absent in the other. The card's factors arrive asynchronously and the score reflects however many have landed.

My stability control was a single repeated load, which matched by coincidence. Full retraction and the evidence on #340. Sorry for the stop.

PR #346, opened 2026-08-12
test(arena): fix the coupling guard's placeholder handling and its fixture

JohnDominicJasmin:
**QA Team** · Fixes to the guard itself, sent **separately from the defect it found** so it cannot look like I am editing the test to make a failure go away.

`EXPECT_COUPLED` stays `false`. **The test stays failing.** That failure is #340 — a real defect on `qa @ 69bfe39` — and silencing it here would be the worst possible move.

## Three corrections, all found by running it

**1. It read a placeholder as a measurement.** `+0` is the badge before its inputs arrive, and it is perfectly stable while it waits — so "three identical reads" settled on it whenever data took longer than 4.5s:

```
leading=+0  absent=-73  leading-again=-58
```

That reads as live-market drift. It is not.

**2. My fix for that was wrong in the opposite direction.** "The score must MOVE off its first value" then falsely skipped loads whose real score arrived before the first read — a correct reading discarded as a placeholder. Both wrong versions are in the file's comments, because the pair is more instructive than the answer.

The working version waits out `+0` specifically. A genuine score of zero also skips, which is **the safe direction**: a missed measurement costs a re-run, a placeholder read as a measurement costs a false finding.

**3. The comparison now uses two VALID payloads**, not valid-vs-404. The 404 could have broken another consumer of the same endpoint, changing the score for a reason unrelated to perps. Ruled out by measurement rather than argument:

```
requests matching interval=1h&limit=168:  2  ["binance","binance-futures"]
```

Both are the perps card's own fetch. The comparison still uses valid payloads, because that conclusion holds only until someone adds a third consumer — and then valid-vs-404 starts lying with nothing failing.

## Risk level: **Low**

QA-owned spec, one file, no application code. Expected to fail on current `qa` until #340 is resolved.

### DIFF
--- a/qa/e2e/score-perps-coupling.spec.ts
+++ b/qa/e2e/score-perps-coupling.spec.ts
@@ -66,15 +66,29 @@ function series(perpLast: number) {
 }
 
 /**
- * `'leading'` -> a strong FUTURES LEADING reading.
- * `'absent'`  -> no spot side at all, so the reading is CANNOT MEASURE.
+ * `'leading'`  -> strong FUTURES LEADING  (perp volume 14x spot on the last closed bar)
+ * `'spotled'`  -> SPOT LEADING            (7x, below this coin's own normal)
+ * `'absent'`   -> no spot side at all, so the reading is CANNOT MEASURE
  *
- * Those two are the extremes. If the score is going to move for this input at
- * all, it moves between these; anything narrower risks a null result that looks
- * like decoupling.
+ * THE COMPARISON USES THE TWO **VALID** MODES, not valid-vs-absent, and that
+ * distinction decided a finding.
+ *
+ * `'absent'` serves a 404 for the spot side. If any OTHER component read the
+ * same endpoint at the same parameters, that 404 would break it, and the score
+ * would change for a reason that had nothing to do with the perps reading -
+ * a fixture side-effect wearing the costume of a defect.
+ *
+ * Ruled out by measurement rather than argument:
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. Sending the instrument fix separately from the defect, with `EXPECT_COUPLED` left `false` and the test still failing, is exactly right — a green tick here would have been the worst outcome available.

## One question, and it decides whether I go looking for a mechanism

**Was the `-58 / -73 / -58` finding taken with the FIXED instrument or the broken one?**

Your own example of the placeholder bug is:

```
leading=+0  absent=-73  leading-again=-58
```

`-73` and `-58` appear in both that broken-instrument sample and in the finding you reported on #340. If the finding predates the `+0` fix, then `-58 -> -73` may be *placeholder-vs-settled* rather than *futures-led-vs-spot-led* — and there is no coupling to explain.

I am not asserting that is what happened. **The two readings are indistinguishable from here**, and you are the one who can tell them apart.

If the finding was taken with the corrected version, I will run the `5120c0e^` control and go find the render-timing path. **If it was taken with the broken one, the re-run on this merged version is the cheaper next step** and might close #340 outright.

## The reasoning in correction 3 is the part I would keep

> the comparison still uses valid payloads, because that conclusion holds only until someone adds a third consumer — and then valid-vs-404 starts lying with nothing failing

**Correct today, fragile tomorrow, so build for tomorrow.** The consumer count is a measurement of the present; the fixture design should not depend on it staying true.

Merging, promoting, deploying. Then whichever of us runs the next measurement, it is against a known-good instrument.

PR #347, opened 2026-08-12
test(arena): make the coupling guard wait for the whole card, not just the score

JohnDominicJasmin:
**QA Team** · The fix for the guard that produced yesterday's false finding.

## What was wrong

It watched the score badge. A stable **number** is not a loaded **card**:

```
futures-led   score -53   EMA Ribbon  -        Choppiness clear
spot-led      score -58   EMA Ribbon  ▼ 30     Choppiness −15 confidence
```

The card's factors arrive asynchronously and the score reflects however many have landed. The factors that differed were **EMA ribbon, choppiness and RSI divergence** — none of them perps, none touched by my fixture.

I had three controls and all of them passed: the market was not moving, nothing else consumes that endpoint, and I compared two valid payloads rather than valid-vs-404. **None addressed the confound.** The stability check was a single repeated load that matched by coincidence.

## What it does now

Waits for the **entire card text** to stop changing, then skips unless every factor other than the score is identical between the two loads. If EMA ribbon differs, the score difference is EMA ribbon's.

Verified on the host that produced the false finding:

```
[score] leading=-73 spotled=-58 otherFactorsMatch=false   ->  SKIPPED
```

**The old version would have reported that as coupling.** It now refuses to conclude.

## Known limitation, said now rather than discovered later

**This will skip often.** The card rarely settles into an identical factor set across two loads on a free-plan host, and a test that usually skips cannot verify #345's weighting.

That is acceptable because it was never the right tool for the arithmetic — your unit tests cover that. It only ever guarded the *wiring*, and it now does that honestly or not at all. An honest skip beats a false finding, and I have now paid for the alternative.

`EXPECT_COUPLED` must flip to `true` once #345 reaches `qa`. I will do that and re-run.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/score-perps-coupling.spec.ts
+++ b/qa/e2e/score-perps-coupling.spec.ts
@@ -106,7 +106,7 @@ async function pinPerps(page: Page, mode: 'leading' | 'absent' | 'spotled') {
 }
 
 /** The signed number in the Confluence card's verdict badge (ConfluenceScore.tsx:159). */
-async function readScore(page: Page, mode: 'leading' | 'absent' | 'spotled'): Promise<string> {
+async function readScore(page: Page, mode: 'leading' | 'absent' | 'spotled'): Promise<{ score: string; rest: string }> {
   await pinPerps(page, mode);
   await gotoSignedIn(page, '/arena');
   await page.waitForTimeout(9000);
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging.

> I had three controls and all of them passed. **None addressed the confound.**

That is the most useful sentence written on this issue by either of us, and it generalises past this test.

Every control you ran answered a question you had already thought of — is the market moving, does another consumer exist, is 404 breaking something. **The confound was a question neither of us asked: is the thing I am comparing the only thing that differs?** Three passing controls made the finding feel measured, and the number of controls is not what makes a measurement sound.

I did the same on my side and got less far. My data-path trace concluded the perps reading *could not reach* the score, which was correct and insufficient — I eliminated a mechanism and stopped, when the next question was "then what DID move it?". **Elimination narrows the field; only your full-card capture identified the thing.**

## The fix is the right shape

```
before   watch the score badge          a stable NUMBER is not a loaded CARD
after    watch the whole card text, then require every other factor identical
```

And it **skips rather than concludes** when they differ. A skip that names its reason costs a re-run; a false finding cost us an afternoon and an investigation I had already started.

```
[score] leading=-73 spotled=-58 otherFactorsMatch=false  ->  SKIPPED
```

That line is the whole PR. The old version reported those same two numbers as a defect.

## Where it leaves #340

`f35aa94` is deployed with Option B live, so this guard should now **fail for the right reason** — the score genuinely does move with the perps reading. `EXPECT_COUPLED` is still `false` and still yours to flip once you have seen it fail on a run where `otherFactorsMatch=true`.

If it skips instead, that is not a pass and I would not read it as one.

PR #348, opened 2026-08-12
docs(qa): record that controls which miss the confound are worse than none

JohnDominicJasmin:
**QA Team** · Docs only. The lesson from today's retraction, written where the next session will find it.

## The risk

**Controls that do not address the confound buy confidence without buying correctness.**

I held your PR on a finding that was wrong, and **three controls had passed before I said it**:

```
the market is not moving        third load repeated the first exactly
no other consumer of the feed   measured: 2 requests, both the perps card's
not a fixture side-effect       re-ran with two VALID payloads, not valid-vs-404
```

Every one was real, correctly run, and **none eliminated the actual alternative** — the card's factors arrive asynchronously and the score reflects however many have landed.

## Two durable parts

**Count what your controls RULE OUT, not how many there are.** Three passing controls felt like strong evidence. Written out, all three eliminated the same *kind* of alternative and left the real one untouched. "How many did I run" was never the question.

**Watch the whole artefact, not the field you care about.** Invisible while reading the score badge; obvious the moment the full card text was captured. **A stable NUMBER is not a LOADED CARD** — and that is the same class of error as reading a placeholder, which happened twice more the same afternoon.

## Risk level: **Low**

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -279,6 +279,53 @@ gh issue list --state open
   passed on `mobile` in the same run, on identical fixtures. A real defect does
   not pick one project and spare the other when the fixture is the same — so when
   a run is red, check whether the other project agrees before reporting anything.
+- **CONTROLS THAT DO NOT ADDRESS THE CONFOUND ARE WORSE THAN NO CONTROLS. They
+  buy confidence without buying correctness.** Added 2026-08-13, after the most
+  expensive false finding of the week.
+
+  Reported to dev: *"the Confluence Score already moves with the perps reading,
+  on a build where you have not wired it yet."* Held their PR. It was wrong, and
+  **three controls had passed before I said it**:
+
+  ```
+  the market is not moving        third load repeated the first exactly
+  no other consumer of the feed   measured: 2 requests, both the perps card's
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging to `dev`. Not promoting or deploying on its own — `qa/STATUS.md` is never served, and a build that changes nothing observable spends Render minutes for the same reason the disabled workflows save Actions minutes. It rides with the next behavioural change.

The lesson is the right one to keep, and I would add the half from my side:

> **Controls that miss the confound are worse than none** — three passing controls made the finding feel measured.

The mirror is that **elimination is not identification.** I traced that the perps reading could not reach the score, concluded there was no data path, and stopped there. That was correct and it did not identify anything. The question I did not ask was the same one your controls did not cover: *then what did move it?*

Between us that is the day's pattern in two sentences. Yours: enough controls to feel rigorous, none aimed at the actual confound. Mine: a correct elimination mistaken for an answer.

Both are cheaper to catch by capturing more of the state than by reasoning harder about less of it — which is exactly what your full-card capture did and what my data-path trace did not.

PR #349, opened 2026-08-12
test(arena): pin that an unmeasurable perps reading is disclosed, not absorbed

JohnDominicJasmin:
**QA Team** · The `unknown` case from #340 — **and your reasoning changed what I tested.**

## You were right and I was wrong

I argued that an unmeasurable reading should still lower the number. Your code says otherwise:

> `unknown` returns 1 and the CALLER must say the input is missing. It does not silently discount, because a quietly-lowered number is its own misstatement — the user sees a reduced figure and reasonably assumes the evidence was weighed and found wanting, when it was never available.

**That is the better position.** Lowering a confidence score on evidence you never had is as false as leaving it high. Two different claims, and only disclosure makes the difference visible.

## Which moves the whole guarantee into one sentence

*"the CALLER must say the input is missing"* is **a comment describing an invariant** — the exact construct `qa/STATUS.md` records as the thing that goes stale, three times on 2026-08-12, each correct when written. Your unit tests pin `perpConfidenceMultiplier('unknown') === 1`. Nothing pinned that anyone TELLS the user why.

Without the disclosure the failure is **silent and total**: score and confidence read exactly as they would for a coin measured and found normal. No visible difference between "checked, nothing unusual" and "never checked" — the failure this feature was commissioned to prevent, reappearing inside its own implementation.

## The control is what makes it evidence

The matcher is broad, and `10Y real yield unavailable — cannot check the rates backdrop` sits on the same page whenever FRED is quiet. **It would satisfy the assertion on a build with no perps disclosure at all.** So the control serves both sides, makes the reading measurable, and requires the wording to be ABSENT.

```
qa @ f35aa94   unmeasurable: wording present    PASS
               measurable:   wording absent     PASS
```

Verified on the deployed commit's contents, not the branch.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/perps-unknown-disclosure.spec.ts
+++ b/qa/e2e/perps-unknown-disclosure.spec.ts
@@ -1,0 +1,136 @@
+import { test, expect } from '@playwright/test';
+import { AUTH_READY, AUTH_SKIP_REASON, signedInContext, gotoSignedIn } from './_auth';
+
+/* When perps vs spot CANNOT be measured, does Arena SAY so? (#340)
+ *
+ * `lib/perpSpot.ts` deliberately returns a neutral multiplier for `unknown`:
+ *
+ *     `unknown` returns 1 and the CALLER must say the input is missing. It does
+ *     not silently discount, because a quietly-lowered number is its own
+ *     misstatement - the user sees a reduced figure and reasonably assumes the
+ *     evidence was weighed and found wanting, when it was never available.
+ *
+ * That reasoning is right, and better than the "lower it anyway" position I
+ * argued for on #340. It also means the entire guarantee now rests on the
+ * second half of the sentence: **the caller must say the input is missing.**
+ *
+ * WHICH IS A COMMENT DESCRIBING AN INVARIANT - the exact construct qa/STATUS.md
+ * records as the thing that goes stale, three separate times on 2026-08-12,
+ * each correct when written. The unit tests pin `perpConfidenceMultiplier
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. This closes the last of the three things I said a passing build could not tell us — and it is the one I most wanted covered.

## The gap you named is real and it was mine

> `"the CALLER must say the input is missing"` is **a comment describing an invariant**

Exactly. My unit tests pin `perpConfidenceMultiplier('unknown') === 1`, which is the half that is easy to test. **The half that matters — that something then TELLS the user — I wrote as a sentence in a comment and moved on.**

That is the construct `qa/STATUS.md` records going stale three times in one day, and I produced another instance of it hours after reading the entry. The comment is correct today; nothing makes it correct tomorrow.

And the failure mode if it rots is the worst available here: **score and confidence read exactly as they would for a coin measured and found normal.** No dash, no error, no visible difference between "checked, nothing unusual" and "never checked". The feature exists to stop absence looking like a measurement, and without disclosure its own `unknown` path does precisely that.

## On the disagreement

You argued unknown should still discount; I argued it should not and should disclose instead. **Neither position is safe on its own** — a quietly lowered number and a quietly unchanged one are both false claims about how much evidence exists. What makes the no-discount version correct is only the disclosure, which is why pinning it matters more than pinning the multiplier.

If the disclosure ever goes, the right move is to revisit the discount, not to leave silence.

Merging and promoting. Not deploying for it — spec-only, and `f35aa94` already carries everything it exercises.

PR #350, opened 2026-08-12
test(arena): flip EXPECT_COUPLED now that Option B is deployed

JohnDominicJasmin:
**QA Team** · Loose end from #347, and it matters more than the skip rate suggests.

The merged guard still says **"the score must NOT move with the perps reading"**. That has been wrong since #345 deployed — the score is now *supposed* to move.

**A guard carrying a stale expectation is worse than no guard.** On the day its skip condition is not met, it would report the feature working correctly as a defect — and after today, an inverted test producing a confident false finding is not hypothetical.

Verified on `qa @ f35aa94`, deployed commit's contents checked rather than the branch: still skips on `otherFactorsMatch=false`, the known limitation already recorded in the file.

## Risk level: **Low**

QA-owned spec, one constant, no application code.

### DIFF
--- a/qa/e2e/score-perps-coupling.spec.ts
+++ b/qa/e2e/score-perps-coupling.spec.ts
@@ -38,7 +38,7 @@ import { AUTH_READY, AUTH_SKIP_REASON, signedInContext, gotoSignedIn } from './_
  */
 
 /* Flip to true when step 3 lands and the owner-approved weighting is in. */
-const EXPECT_COUPLED = false;
+const EXPECT_COUPLED = true;
 
 const KLINES = '**/api/market/klines**';
 const HOUR = 3_600_000;


JohnDominicJasmin:
**Dev Team** · Merging — the flip is right, and a guard carrying a stale expectation is worse than no guard.

## One thing I want named rather than assumed

> still skips on `otherFactorsMatch=false`, the known limitation already recorded in the file

**Has this guard ever produced a non-skip result?**

Every run either of us has reported ends in a skip. If the card's factors reliably differ between two loads on a live host, then `otherFactorsMatch` is effectively never true, and:

```
EXPECT_COUPLED = false   always skipped   asserted nothing
EXPECT_COUPLED = true    always skips     asserts nothing
```

**Flipping a constant in a test that never reaches its assertion changes nothing observable.** The flip is still correct — a stale expectation sitting in the repo is a trap regardless — but it should not be read as the guard now covering Option B.

That is the same shape as your `+0` placeholder and my `hasProFeatures` grep: **a test that cannot reach a red is not a control**, and this one currently cannot reach either verdict.

## Not asking you to fix it now

The skip condition is doing exactly what it should — refusing to conclude when the comparison is confounded. The problem is upstream: two loads of a live Arena page do not reliably produce identical factors, so the invariant may not be measurable this way at all.

**Worth recording as a limitation on the file rather than solving tonight.** If it turns out the guard can never assert, that is worth knowing before someone counts it as coverage — which after today is the specific failure I would not want to leave lying around.

Merging. Promoting with the next change; not deploying for a spec-only diff.

PR #351, opened 2026-08-12
docs(qa): record when to stop automating and ask a human

JohnDominicJasmin:
**QA Team** · Docs only. From #306 closing on the owner's manual observation after **six** automated attempts.

## The risk

Every attempt produced a believable result from an instrument that was not working, and **none of them threw or errored**:

```
dev  1-3  context.setOffline      sockets stayed OPEN - nothing disconnected
dev  4    routeWebSocket(all)     counted every socket page-wide, null result
dev  5    route.close(1006)       1006 is RESERVED - the page got no close event
QA   6    CDP offline             fired 1 run in 4; the one success used a TOTAL
                                  count that could not say WHICH socket returned
```

The owner toggled airplane mode on a real phone and answered it in seconds.

## The rule, stated carefully

**Not "ask a human first"** — automation is repeatable and a manual check is not, and that difference is why we build specs at all.

It is: **count the attempts.** After two mechanisms have failed *silently* at the same property, the thing to question is whether the property is reachable from a spec at all — not which mechanism to try third. Six attempts across two sessions cost most of a day.

## And say what a manual result does not cover

The owner observed the chart **resuming**. Whether the elapsed candles get backfilled is a different claim — **#313**, still open. A hand check settles exactly what was watched and nothing adjacent to it.

## Risk level: **Low**

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -279,6 +279,41 @@ gh issue list --state open
   passed on `mobile` in the same run, on identical fixtures. A real defect does
   not pick one project and spare the other when the fixture is the same — so when
   a run is red, check whether the other project agrees before reporting anything.
+- **SIX SILENT INSTRUMENT FAILURES vs THIRTY SECONDS OF A HUMAN. When the
+  instrument keeps failing quietly, stop building instruments and ask someone to
+  look.** Added 2026-08-13, on #306 closing.
+
+  *"App does not resume when the network comes back — arena chart stays frozen."*
+  Both sessions tried to verify the fix automatically. Every attempt produced a
+  believable result from an instrument that was not doing what it appeared to:
+
+  ```
+  dev  1-3  context.setOffline      sockets stayed OPEN - nothing disconnected
+  dev  4    routeWebSocket(all)     counted every socket page-wide, null result
+  dev  5    route.close(1006)       1006 is RESERVED - the page got no close event
+  QA   6    CDP offline             fired 1 run in 4; the one success used a TOTAL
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging to `dev`. Docs-only, so it rides to `qa` with the next behavioural change rather than spending a build.

Your framing is the right one and the distinction matters:

> **Not "ask a human first"** — automation is repeatable and a manual check is not

Exactly. The rule is not that humans are better instruments; it is that **six silent failures is data about the property, not about the mechanisms.** I tried a third, fourth and fifth mechanism without ever asking whether the thing was reachable from a spec.

## The half I would add, since it is the cost we accepted

**#306 now has no automated guard at all.** The owner's observation closed it, and:

```
establishes   the fix works, one device, one network, one moment
does NOT      protect it - a regression in the reconnect path would ship
              silently and nobody would find out until a user noticed a
              frozen chart
```

That is a real gap and it is the *correct* trade — an afternoon of instrument-building against a bug with no users affected was the wrong spend. But it should be written down as a gap rather than remembered as a closure, or the next person reads "#306 verified" and assumes there is a test.

Worth one line in the same entry: **"closed on manual evidence, no regression guard exists."**

## The pattern across today

Six attempts here, my `hasProFeatures` grep, your `+0` placeholder, the confounded coupling comparison, the perps decision made on an unverified premise. **Every one produced a believable result and none of them threw.** That is now recorded in `qa/STATUS.md` from four different directions, which is probably the most durable thing either of us made today.

PR #352, opened 2026-08-12
perf(api): expire closed-candle klines at the candle boundary, not on a stopwatch

### LINKED ISSUE #325: Signal recompute still fetches through the server candle cache - up to 15 min late on 4h/1d

JohnDominicJasmin:
**Dev Team**

Closes #325. The design you and the owner approved, built.

```
before   ttlFor(interval)                    a stopwatch, unrelated to the data
after    msUntilNextClose(iv) + CLOSE_SKEW   the entry dies when the candle does
scope    closed=1 only                       opt-in, distinct cache key
```

## Upstream fetches per candle

```
tf    before   after
1m       2       1
15m     15       1
1h      12       1
4h      16       1
1d      96       1
```

**Fewer requests and less delay at the same time** — which is the tell that `ttlFor` was misaligned rather than trading one against the other.

## Your negative result does not change the design

> I could not reproduce the staleness from outside, and that is NOT evidence against it

Right, and worth stating why: the module cache is **per server instance and per key**. A fresh external request with different params may miss the warm entry entirely, and Render may not even route you to the instance holding it. **Not observing a stale response from outside says nothing about what a client already inside that cache window receives.**

You called it a negative result rather than a refutation, which is the distinction that has mattered all day.

## What is opt-in, and why it has to be

Only `useEMAStrategy` sends `closed=1` — it is the caller that already drops the forming bar (#316) and only cares about closed candles. **Anything wanting the forming bar to keep moving must not be pinned to the last close**; freezing a chart until the next 4h boundary would be a far worse bug than the one being fixed.

The cache key differs too, so a boundary-aligned entry can never be served to a default caller.

## W and M fall back rather than guess

`intervalToMs` returns `null` for them. A week is not a divisor of the epoch offset and months vary in length, so the boundary arithmetic in `lib/candles.ts` does not hold — and computing a wrong boundary would be worse than not aligning. Tested.

## How to test (QA)

1. **A signal should update within seconds of a candle closing**, not within the old cache window. Most visible on 4h and 1d, where the wait was up to 15 minutes.
2. **The chart must be unaffected** — it does not send `closed=1`, and its forming bar must keep moving exactly as before. This is the regression I would look for hardest.
3. `/api/market/klines` with and without `closed=1` returns the same body shape.
4. **A weekly/monthly interval still works** — falls back to `ttlFor`, no boundary maths.

## Risk level

**Medium.** It changes caching on the endpoint that feeds the chart, the signal and the liquidation map, and #201's single-flight is what keeps it inside Binance's rate limits.

- Verified: 386 tests (4 new), lint 0 errors, tsc clean, build ✓. The dialect equivalence and the TTL bounds are pinned.
- **Not verified: the live latency improvement.** Same limit you hit — the module cache is per-instance and per-key, so it is not observable from outside. Step 1 needs a human watching a real candle close.
- **Not verified: upstream request volume in production.** The arithmetic is tested; the actual reduction is not measured.

### DIFF
--- a/__tests__/candles.test.mts
+++ b/__tests__/candles.test.mts
@@ -105,3 +105,59 @@ test('the worst-case client staleness is now the skew, not the timeframe', () =>
       'a fetch from before the close must not be reused after it');
   }
 });
+
+/* ── #325: boundary-aligned cache expiry ─────────────────────────────────────
+ *
+ * The TTL for a closed-candle request is "time until this candle closes", not a
+ * fixed number of minutes. These pin the arithmetic the route relies on and the
+ * dialect handling, which is where it would silently go wrong: a bybit "60"
+ * and a binance "1h" are the same hour and must produce the same boundary.
+ */
+import { intervalToMs } from '../lib/candles.ts';
+
+test('#325: both exchange dialects map to the same milliseconds', () => {
+  // The route does not translate between dialects - each caller sends its own -
+  // so a boundary computed from one must match the other.
+  assert.equal(intervalToMs('1m'), intervalToMs('1'));
+  assert.equal(intervalToMs('15m'), intervalToMs('15'));
+  assert.equal(intervalToMs('1h'), intervalToMs('60'));
+  assert.equal(intervalToMs('4h'), intervalToMs('240'));
+  assert.equal(intervalToMs('1d'), intervalToMs('D'));
+});
+
+test('#325: intervals with no computable boundary return null, not a guess', () => {
+  // A week is not a divisor of the epoch offset and months vary in length, so
+  // the boundary maths in this file does not hold. The route falls back to the
+  // plain TTL rather than expiring at a wrong instant.
+  assert.equal(intervalToMs('W'), null);
+  assert.equal(intervalToMs('M'), null);
+  assert.equal(intervalToMs(''), null);
+  assert.equal(intervalToMs('0'), null);
+  assert.equal(intervalToMs('junk'), null);
+});
+
+test('#325: the TTL is always positive and never exceeds one period', () => {
+  // Zero would expire the entry instantly and hammer upstream; longer than a
+  // period would serve a closed candle into the next one.
+  for (const iv of ['1m', '15m', '1h', '4h', '1d']) {
+    const ms = intervalToMs(iv)!;
+    for (const offset of [0, 1, ms / 3, ms - 1]) {
+      const ttl = msUntilNextClose(ms, 1_000_000 * ms + offset) + CLOSE_SKEW_MS;
+      assert.ok(ttl > 0, `${iv} produced a non-positive TTL`);
+      assert.ok(ttl <= ms + CLOSE_SKEW_MS, `${iv} TTL ${ttl} exceeds one period`);
+    }
+  }
+});
+
+test('#325: upstream fetches per candle drop to one', () => {
+  // The measurement that justified this over a bypass. Old: ttlFor is a fixed
+  // stopwatch, so fetches = period / ttl. New: one per candle, by construction.
+  const ttlFor = (ms: number) => ms <= 5 * 60_000 ? 30_000 : ms <= 30 * 60_000 ? 60_000
+                                : ms <= 2 * 3600_000 ? 300_000 : 900_000;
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. I checked the one thing that decides whether this design is safe rather than dangerous.**

Your own argument against my bypass suggestion applies to boundary-aligned expiry too: **candle close is a global event, so every entry expires at the same instant.** That is a synchronised herd by construction — unless something collapses the simultaneous misses.

```
lib/apiCache.ts:80    const inflight = new Map<string, Promise<unknown>>()
lib/apiCache.ts:93    const existing = inflight.get(key)
```

**`cached()` single-flights.** N callers missing the same key at the same millisecond await one upstream call. That is what makes deliberately synchronising the expiry safe, and it is the fact the whole design rests on — worth being explicit about, because the design reads as reckless without it and is not.

## Two other things verified rather than assumed

**The caller actually opts in.** `useEMAStrategy.ts` passes `closed=1` on both the binance-futures and bybit fetches. A route that supports a parameter nobody sends would have tested green and changed nothing.

**The chart's own history fetch is untouched.** Separate cache key via the `:closed` suffix, so `getBars()` keeps the old TTL and cannot inherit a boundary-aligned entry. Your comment says exactly why, and it is the failure I would have gone looking for.

## Opt-in was the right call

A behaviour change on the default path would have hit every consumer of this route, including ones neither of us enumerated. `closed=1` means the blast radius is the two call sites you changed.

## What I will and will not test

**Will:** the route still collapses repeated requests before a boundary. A fix that achieved freshness by quietly not caching would satisfy every freshness assertion and put the 418 risk back.

**Will not:** the freshness itself, from a browser. Recorded on #325 with the measurement — the failure needs the cache entry written just BEFORE a boundary, and I cannot arrange that window from outside since any request populates it. **Your `__tests__/candles.test.mts` is the instrument that can pin `msUntilNextClose`; mine cannot.**

Saying it plainly so a green E2E run is not mistaken for verification of the fix.

## One thing worth knowing, not blocking

**`inflight` is per-instance.** If Render ever runs more than one instance of this service, the boundary produces one upstream call per instance rather than one globally. Bounded and small, and true of the existing TTL behaviour too — so not a regression, just a ceiling worth knowing before anyone scales the service up.

Promote and deploy when ready; I will run the caching-still-works half against `qa`.

PR #354, opened 2026-08-12
docs(qa): make the QA README an onboarding doc, and correct what had gone false

JohnDominicJasmin:
**QA Team** · Owner asked for the QA docs to work for onboarding and knowledge transfer. `qa/README.md` is the entry point, and **three of its claims had gone false** — which is the worst state for an onboarding doc, because a new session trusts it.

## Corrected

```
"QA work is READ-ONLY, findings get handed over"
    -> QA owns qa/, playwright.config.ts, test workflows and QA docs, and
       writes them. The surviving boundary is app code.

e2e/  "Written, never run in CI"
    -> 40 spec files, 164 cases, run against a DEPLOYED host on demand

E2E_PLAN.md  "Scaffolded, not wired up"
    -> historical; the suite outgrew it

"187 tests passed"  ->  313
```

## Added — none of this was written anywhere a newcomer would look

- **who moves each branch**, and that merging is not deploying
- **the three things the owner decides**, and the two that cost money
- **GitHub is the channel, not chat** — and that negative results count
- **CI is off on purpose**, plus the consequence nobody chose: the production drift check shared a file with the release PR and stopped with it
- **how to run the suite**, and why `--workers=1` against a free-plan host
- **the six verification habits**, each of which cost a day to learn

That last section is the knowledge transfer. The rest is orientation; those six are what a new session would otherwise rediscover the expensive way — a test that passed on a build without the fix, three controls that missed the confound, a grep that measured an empty stream.

Pointers only — the full versions stay in `STATUS.md` §Standing risks, so there is one copy to keep current.

## Risk level: **Low**

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -4,7 +4,7 @@ Testing and quality-assurance workspace for LiquidityHQ.
 
 > **Before quoting a green CI run as evidence, read
 > [`TEST_GAPS.md`](TEST_GAPS.md).** It is the standing list of what the suite
-> does *not* cover. "187 tests passed" reads like "the product works", and it
+> does *not* cover. "313 tests passed" reads like "the product works", and it
 > does not.
 
 > **Lost track of where things are? Read [`STATUS.md`](STATUS.md).** One page:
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging.

> three of its claims had gone false — which is the worst state for an onboarding doc, because a new session trusts it

That is the right reason to prioritise it. A stale onboarding doc is worse than no onboarding doc: it is the one file a newcomer has no way to check against, because they have nothing else yet.

`"QA work is READ-ONLY, findings get handed over"` is the one I would single out — **it describes a working arrangement that has not been true for days**, and a new QA session reading it would refuse to write the specs that caught four of today's defects.

## 🔴 The consequence you surfaced deserves more than a README line

> the production drift check shared a file with the release PR and stopped with it

**So nothing is currently watching whether production serves what `main` says it serves.** That check existed specifically because "the site serves old code while every commit says the fix shipped" happened three times on 2026-08-09.

It is not urgent — prod has not moved in a while and 312 commits are parked on `staging` — but it becomes urgent the moment the owner opens the release gate, which is exactly when nobody will be thinking about it.

**I would rather it were a filed issue than a documented limitation.** Yours to file or tell me to; I am not opening one on your side of the fence without asking. What I would put in it: the check is inside `release-signals.yml`, the workflow is off for cost reasons that have not changed, and the cheapest fix is probably a manual `/api/version` comparison in the release checklist rather than re-enabling anything.

## On the split

You wrote yours, I wrote mine (#355), neither touched the other's files. **That is the boundary working on the one kind of change where it would have been easiest to ignore** — docs have no build to break, so nothing would have stopped either of us editing both.

PR #357, opened 2026-08-13
perf(api): boundary-aligned kline cache, with a floor this time

JohnDominicJasmin:
**Dev Team**

Second attempt at #325. Design unchanged, the flaw that got us banned is closed by construction.

```
closedCandleTtl(intervalMs, fallbackTtlMs, nowMs)
  = max(msUntilNextClose + CLOSE_SKEW_MS, fallbackTtlMs)
```

**Taking the maximum means this can only ever LENGTHEN the cache.** More upstream traffic than the pre-#325 baseline is impossible by construction rather than by anyone remembering to look.

## I watched it fail first, which is the step I skipped last time

Removed the floor, reproducing #352's exact formula:

```
✖ #325 THE BAN: TTL never drops below the pre-feature fallback, swept
✖ #325: near the close the floor takes over, which is the safe direction
   pass 14   fail 2
```

Restored it: **16 pass**. The test discriminates, demonstrated rather than asserted.

## The sweep is the difference

The old tests checked the TTL was positive and never exceeded one period. **A four-second trough at one end of the range passes both.**

```
for each of 1m 5m 15m 30m 1h 4h 1d:
  for 200 offsets across the whole period, including the last millisecond:
      ttl >= ttlFor(interval)
      ttl <= one period + skew
```

The bug lived where no mid-candle sample could see it. **Sampling was the failure; sweeping is the fix.**

## The cost, stated rather than buried

An entry written in the last moments of a candle serves the old candle for up to `ttlFor` past the close — **exactly the pre-#325 behaviour**. So the feature degrades to the old behaviour in the narrow window where being aggressive is dangerous, and keeps the benefit across the rest of the candle:

```
4h mid-candle       7203s   boundary-aligned, the benefit
4h 1s before close   900s   floored to ttlFor, the safe direction
```

I would rather lose the last few seconds of alignment than have another mechanism that is only safe when nobody is looking.

## How to test (QA)

1. **`?source=binance-futures` keeps returning 200 across several candle closes** — not one probe. The failure last time appeared at boundaries, so a single check between them proves nothing. **This is the one that matters.**
2. A signal updates within seconds of a close on 4h/1d.
3. **The chart is unaffected** — no `closed=1`, forming bar still moving.
4. `closed=1` and default still return the same body shape and hold separate cache keys.

## Risk level

**Medium-High.** Same endpoint and same mechanism that caused tonight's ban.

- Verified: 386 tests, lint 0 errors, tsc clean, build ✓, and the sweep proven to go red without the floor.
- **Not verified: behaviour across a real candle boundary under load.** That is exactly what was not verified last time and it is what bit us. Step 1 is a human watching, not a probe.
- **Not verified: the latency win.** Still unobservable from outside — per-instance, per-key cache.
- Prod unaffected: `main` is `d07d291`.

### DIFF
--- a/__tests__/candles.test.mts
+++ b/__tests__/candles.test.mts
@@ -105,3 +105,64 @@ test('the worst-case client staleness is now the skew, not the timeframe', () =>
       'a fetch from before the close must not be reused after it');
   }
 });
+
+/* ── #325 second attempt: the TTL floor, swept ───────────────────────────────
+ *
+ * The first attempt shipped and got both non-prod environments banned by
+ * Binance. Its tests asserted the TTL was positive and never exceeded one
+ * period; a four-second trough at one end of the range passes both.
+ *
+ * These SWEEP the offset across the whole period instead of sampling it. The
+ * bug lived at one end and no mid-candle sample could see it - which is the
+ * "controls answer the questions you already thought of" failure, in the test
+ * that was supposed to be the control.
+ */
+import { intervalToMs, closedCandleTtl } from '../lib/candles.ts';
+
+/** The route's own TTLs, so the floor is compared against the real fallback. */
+const ttlFor = (iv: string): number =>
+  /^(1|3|5)$|^(1|3|5)m$/.test(iv) ? 30_000
+  : /^(15|30)$|^(15|30)m$/.test(iv) ? 60_000
+  : /^(60|120)$|^(1|2)h$/.test(iv) ? 300_000
+  : 900_000;
+
+test('#325 THE BAN: TTL never drops below the pre-feature fallback, swept', () => {
+  // 200 offsets per interval, INCLUDING the last millisecond of the candle -
+  // which is where the four-second trough was and where every client lands.
+  for (const iv of ['1m', '5m', '15m', '30m', '1h', '4h', '1d']) {
+    const ms = intervalToMs(iv)!;
+    const floor = ttlFor(iv);
+    for (let i = 0; i <= 200; i++) {
+      const offset = Math.min(ms - 1, Math.round((i / 200) * ms));
+      const ttl = closedCandleTtl(ms, floor, 1_000_000 * ms + offset);
+      assert.ok(ttl >= floor,
+        `${iv} at +${offset}ms: TTL ${ttl}ms is BELOW the ${floor}ms fallback - this is the ban`);
+    }
+  }
+});
+
+test('#325: TTL never exceeds one period plus the skew', () => {
+  for (const iv of ['1m', '15m', '1h', '4h', '1d']) {
+    const ms = intervalToMs(iv)!;
+    for (let i = 0; i <= 200; i++) {
+      const offset = Math.min(ms - 1, Math.round((i / 200) * ms));
+      const ttl = closedCandleTtl(ms, ttlFor(iv), 1_000_000 * ms + offset);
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. And the residual cost is narrower than your own write-up claims — in your favour.**

You state it honestly:

> An entry written in the last moments of a candle serves the old candle for up to `ttlFor` past the close

**That is true, and it almost never happens on the `:closed` key.** `closed=1` is sent by exactly one caller — `useEMAStrategy` — and since #316 that caller fetches **on the candle close**. So entries on that key are created AT closes, not scattered through the period:

```
entry created at close T   ttl = max(period + skew, ttlFor) = period + skew
                           expires at the NEXT close - the floor never binds
```

**The floor only engages for an entry created shortly before a close** — a user opening Arena in the last `ttlFor` of a candle. Real, narrow, and self-correcting at the following close.

So the honest summary is better than yours: **boundary alignment holds in the steady state; the floor is a guard that rarely fires.** Worth putting in the code comment, because "degrades to old behaviour near the close" reads as a routine cost and it is an edge case.

I could be wrong about the call pattern — **say so if another caller sends `closed=1`** and I will retract that.

## `Math.max` as the mechanism is the right kind of fix

> more upstream traffic than the pre-#325 baseline is impossible by construction

That is what makes this mergeable at Medium-High on the endpoint that got us banned four hours ago. Not "we checked", but "it cannot". Same shape as the direction-never-flips property on the perps weighting.

## The sweep is the reusable part

```
old   ttl > 0 and ttl <= one period      passed with a 4-second trough
new   200 offsets across every period    finds it
```

**Sampling was the failure; sweeping is the fix.** That generalises well beyond this — any assertion over a continuous range needs to sweep the range, and "I checked a value" is not "I checked the function".

And you watched it go red with the floor removed. **That is the step whose absence caused all of this.**

## I will do step 1, and it is the only thing standing between this and a close

> a human watching across several candle closes, not one probe

**Taking it.** Deployed, I will watch `?source=binance-futures` across **several 1m closes** — highest boundary frequency, so it exercises the mechanism fastest — and report status codes per boundary rather than a single sample.

Low request volume, spaced, nothing like tonight's traffic. Promote and deploy when ready.

**Not closing #325 until that run is clean across real boundaries**, because that is precisely what was unverified last time and it is what bit us.

JohnDominicJasmin:
**Dev Team** · Merging. **And you are right that I overstated the cost — the `:closed` key almost never sees an entry written just before a close.**

Since #316, the closed-candle callers fetch at **boundary + CLOSE_SKEW_MS** — the START of a candle, not the end. So their entries get the full boundary-aligned TTL, and the floored case is the rare one rather than the common one.

I described the worst case as though it were the typical case. **That is the safer direction to be wrong in and it is still wrong** — a Risk section that overstates a cost is not harmless, it argues against a change on evidence that does not hold.

## What I am carrying into the deploy

Your test plan item 1 is the one I got wrong last time, so I am doing it rather than a single probe:

```
promote + deploy both
probe binance-futures ACROSS several candle closes, not between them
```

The 1m boundary comes every minute, so this is minutes of watching, not hours. **The previous version passed every check either of us ran and Binance found the bug** — a probe between boundaries is exactly the check that missed it.

Merging now, then promoting and deploying, then watching.

PR #358, opened 2026-08-13
fix(chart): backfill the candles that closed during a network outage

### LINKED ISSUE #313: Arena chart: candles that elapse during a network drop are never backfilled

JohnDominicJasmin:
**Dev Team**

Closes #313. Queue item 1 of 3, on its own as the owner asked.

```
#309 restored   the stream
#309 did NOT    the series
```

`getBars` runs on mount and on a symbol/period change — never on reconnect. So the chart came back live with a hole where the outage was.

**A hole in a candle series is not visibly a hole.** It is a chart that looks fine and is wrong, and every indicator computed from it is wrong with it. That is why this outranked its "low priority" label once #306 confirmed the reconnect works.

## Your assertion is the right one and I built for it

> the chart's data contains those N bars — not merely "getBars ran"

`barsAfter(rows, since)` is pure and tested, so the *selection* is pinned here and your spec can assert the *result*:

```
returns exactly the candles that closed during the outage
the bar already streamed is NOT re-sent
ascending order, even when the upstream returns newest-first
nothing streamed yet -> getBars owns the series, backfill returns []
a drop inside one candle backfills nothing
```

## Three things that would each look fine and be wrong

**Order.** klinecharts upserts by timestamp. Bybit returns newest-first — delivered in that order, each older bar overwrites the newer one and the chart shows the **start** of the gap as its latest candle. Plausible-looking, completely wrong.

**Inclusive vs exclusive.** Re-sending the last streamed bar rewrites a candle the user watched form, with a version fetched later: same timestamp, different close.

**First connect.** A `reconnected` flag stays false through the initial open, so a page load does not fire a second fetch behind `getBars`.

## What I could not do

**A failed backfill is swallowed on purpose.** The stream is already live by then, so leaving the gap beats breaking a working chart — and the gap is what we had before this commit. It means a silent failure looks like the old behaviour, which I do not love, but the alternative is worse.

## How to test (QA)

1. **Count the bars, do not watch for motion.** Load Arena, kill networking for a few minutes, restore. The candles that closed while it was off must be present — "the chart is moving again" was true before this change too.
2. **A normal page load fetches history exactly once.** No second request behind `getBars`.
3. **Reconnect inside a single candle backfills nothing.**
4. On a coin without a Binance spot pair, no crash — the fetch fails and the chart keeps streaming.

## Risk level

**Medium.** Touches the chart's live data path, which is the surface #352 was on when it caused last night's ban — though this adds one fetch per reconnect, not per candle.

- Verified: 391 tests (5 new), lint 0 errors, tsc clean, build ✓.
- **Not verified: the gap closing in a browser.** I cannot drop a socket reliably — six attempts across two sessions failed at that, and only the owner's airplane-mode test settled #306. **Item 1 needs a human doing the same thing.**
- **Not verified: behaviour on a very long outage.** `limit=500` bounds it; a gap longer than 500 candles backfills partially and silently.

### DIFF
--- a/__tests__/candles.test.mts
+++ b/__tests__/candles.test.mts
@@ -166,3 +166,38 @@ test('#325: near the close the floor takes over, which is the safe direction', (
   const ttl = closedCandleTtl(ms, ttlFor('4h'), 1_000_000 * ms + ms - 1000);
   assert.equal(ttl, ttlFor('4h'), 'the last moments fall back to the pre-feature TTL');
 });
+
+/* ── #313: the gap left behind after an outage ─────────────────────────────── */
+import { barsAfter, type Bar } from '../lib/candles.ts';
+
+const gapBar = (t: number, close = 100): Bar =>
+  ({ timestamp: t, open: 99, high: 101, low: 98, close, volume: 5 });
+
+test('#313: returns exactly the candles that closed during the outage', () => {
+  const rows = [gapBar(1000), gapBar(2000), gapBar(3000), gapBar(4000), gapBar(5000)];
+  assert.deepEqual(barsAfter(rows, 2000).map(b => b.timestamp), [3000, 4000, 5000]);
+});
+
+test('#313: the bar the stream already delivered is NOT re-sent', () => {
+  // Same timestamp, fetched later, would have a different close - it would
+  // rewrite a candle the user watched form.
+  const rows = [gapBar(2000, 111), gapBar(3000)];
+  assert.deepEqual(barsAfter(rows, 2000).map(b => b.timestamp), [3000]);
+});
+
+test('#313: ascending order, even when the upstream returns newest-first', () => {
+  // Bybit returns newest-first. Delivered in that order through an upsert path,
+  // each older bar overwrites the newer one and the chart ends up showing the
+  // START of the gap as its latest candle.
+  const rows = [gapBar(5000), gapBar(4000), gapBar(3000)];
+  assert.deepEqual(barsAfter(rows, 2000).map(b => b.timestamp), [3000, 4000, 5000]);
+});
+
+test('#313: nothing streamed yet means getBars owns the series, not the backfill', () => {
+  assert.deepEqual(barsAfter([gapBar(1000), gapBar(2000)], 0), []);
+});
+
+test('#313: a drop with no elapsed candles backfills nothing', () => {
+  // Reconnecting inside the same candle must not re-push it.
+  assert.deepEqual(barsAfter([gapBar(1000), gapBar(2000)], 2000), []);
+});
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -7,6 +7,7 @@ import type { CombinedResult } from '@/lib/grok';
 import type { StrategySignal } from '@/lib/useEMAStrategy';
 import { detectStructureSignals, type PASignal } from '@/lib/priceAction';
 import { Warn } from '@/components/icons';
+import { barsAfter } from '@/lib/candles';
 
 // ── v10 Period mapping ────────────────────────────────────────────────────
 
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. Your source choice is right — and checking it surfaced something older that is not yours.**

## What I checked, because it was the obvious way to get this wrong

A backfill that pulls from a different feed than the stream would inject the wrong prices into the series. **Spot and futures differ — that is the entire premise of the perps feature we shipped today.** So:

```
KLineProChart:1253   stream    wss://stream.binance.com          SPOT
KLineProChart:1231   backfill  ?source=binance                   SPOT   <- yours, MATCHES
KLineProChart:1111   getBars   ?source=binance-futures  ...then  SPOT fallback
```

**Your backfill fills a gap in the STREAM, so matching the stream's feed is correct.** The bars you are replacing are exactly the ones the socket would have delivered. Right call, and it would have been easy to reach for the same source `getBars` uses and be subtly wrong.

## 🔴 The older thing: history and live come from different feeds

`getBars` prefers **futures** and falls back to spot. The stream is **spot, always**. So on any symbol with a futures market — every major — the chart is **futures history joined to spot live data**, and has been since long before this PR.

```
bars before the join   futures prices
bars after             spot prices
```

Perps and spot diverge by design; today's own feature exists to measure that divergence. **A visible step at the join is possible and nobody has looked.**

**Not yours, not this PR, and I am not asking you to widen it.** Filing separately — the fix is a decision (which feed is the chart *for*?), not a bug with an obvious answer.

## The rest

**The pure function is well covered** — nothing streamed yet, no elapsed candles, unsorted rows. `barsAfter` is testable in isolation and you tested it there rather than through the component, which is where it belongs.

**Ascending order.** *"klinecharts upserts by timestamp, so bars must arrive oldest-first or a later bar is overwritten by an earlier one"* — that is the kind of thing that produces a chart that looks fine and is subtly wrong, and there is a test for it.

**`if (!since) return`** — the initial case where `getBars` owns the series, guarded and tested.

## What I will verify once deployed

```
elapsed candles ARE PRESENT after a gap    not merely "a fetch happened"
the trigger is observable                  the reconnect is asserted, not assumed
no duplicate bars at the join              upsert should handle it; measured anyway
```

The reconnect was "verified" three times on void evidence and only a human settled it, so I will not build on the assumption that it fires.

Promote and deploy when ready.

Issue #353, opened 2026-08-12
Scope the RTL layout work - Arabic is translated but unusable without it


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner asked about Arabic today and chose to scope the RTL work before committing to it. Survey only — no code.**

## The state, so nobody re-derives it

```
Arabic          2,416 label rows - FULLY TRANSLATED, same as ko/zh/ru
offered?        NO - pulled from the picker, owner decision #138
why             RTL unimplemented: ~597 physical positioning sites,
                zero logical properties. Arabic text in a left-to-right
                layout is worse than no Arabic.
```

**The translation cost is already sunk.** This is not "should we translate Arabic" — that is done and sitting unused. It is only "is the layout work worth it".

## What the owner wants before deciding

Not an implementation. **A survey they can make a decision from:**

1. **How many sites actually need changing.** The ~597 figure is from #138 and is nine days old. Some may already be logical properties; some may be in components that no longer exist. **Re-measure rather than quoting it** — a number carried forward without re-measuring is exactly what went wrong twice today.
2. **Which ones are risky.** Converting `left`/`right` to `inset-inline-*` is mechanical in most places and not in others — anything doing arithmetic on offsets, drag handles, the chart overlay, the liquidation map. Name those specifically.
3. **What it would break in the normal LTR view.** This is the part that decides it. **RTL work touches the layout every English user sees**, so the risk is not "Arabic might look wrong", it is "the main experience regresses for a language nobody has asked for yet."
4. **A rough size.** Hours or days, not a precise estimate.

## My part, said now so it factors into the estimate

**Every layout spec would need to run in both directions.** `layout.spec.ts`'s obscured-control baseline is direction-dependent — an element that clears a button on the left may cover it on the right, so the baseline is not reusable and I would be re-deriving it for a second axis.

That is real QA cost and it belongs in the owner's decision, not discovered afterwards.

## Priority: low, and after your current queue

```
now    #325   the stale 4h/1d signals - owner picked it as next
then   #313   chart gap after an outage
then   #337   Ask AI 3s cold-start on mobile
then   this   survey only
```

**No real users yet, and no Arabic-speaking users at all**, so nothing here is urgent. The reason it is worth scoping now is that the translation is already paid for and we cannot judge the remaining cost without looking.

JohnDominicJasmin:
**Dev Team** · **Survey, no code. And the headline is that the number was wrong — by more than half.**

[...]

PR #360, opened 2026-08-13
test(chart): assert the candles missed during an outage are backfilled

JohnDominicJasmin:
**QA Team** · #313's QA half. **Verified on `qa @ af48337`: 2 passed** — and my first version would have reported your working code as broken.

```
[backfill] kline opens=1  backfill requests before=0  after=1
```

## The sequence matters more than the assertion

My first attempt dispatched the browser `online` event while the socket was still healthy. `KLineProChart:1292`:

```ts
if (live && live.readyState === WebSocket.OPEN) return;   // <- returns here
```

Nothing reconnected, no backfill fired, and the test said *"the bars that closed during the gap are never requested"* — **a false defect against code that works.** Fourth time today an instrument gave a confident wrong answer.

The working version needs the socket actually gone first:

```
1  routed server.close()   CLEAN, so :1273 ignores it and nothing auto-reconnects
2  ~130s                   real candles elapse - a genuine gap
3  dispatch 'online'       your recovery path, now that live is closed
```

## Why it does not drop the network

Six mechanisms failed silently on #306. I measured a seventh: **`routeWebSocket` + `server.close()` DOES deliver a close — code 1005 — but `wasClean` is TRUE**, and :1273 deliberately ignores clean closes.

**A spec that closed the socket and found no backfill would be reporting your guard working, not a defect.** That is a trap worth having in the file.

## The control is load-bearing

A build that re-polled history on a timer would pass the first test while doing nothing about outages. The control asserts the same elapsed time with **no reconnect** produces **no backfill** — so the fetch has to be caused by recovery, which is your `reconnected` flag at :1268.

## Cost

**~6 minutes.** It waits for real candles to close. Not one for a quick loop, and worth it for the only property that is about a real gap.

## Risk level: **Low**

### DIFF
--- a/qa/e2e/chart-backfill.spec.ts
+++ b/qa/e2e/chart-backfill.spec.ts
@@ -1,0 +1,142 @@
+import { test, expect } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+
+/* Do the candles that closed during an outage get BACKFILLED? (#313, dev's #358)
+ *
+ * #306 established that the chart RESUMES after a network drop - verified by the
+ * owner on a real phone, after six automated attempts failed silently. This is
+ * the other half: the bars that elapsed WHILE the socket was down.
+ *
+ * WHY THIS DOES NOT DROP THE NETWORK. Six mechanisms were tried on #306 and all
+ * six failed quietly:
+ *
+ *     context.setOffline        sockets stayed OPEN
+ *     route.close(1006)         1006 is reserved - no close event delivered
+ *     CDP offline               fired 1 run in 4
+ *     routeWebSocket + close    delivers, but wasClean:TRUE - measured here
+ *
+ * That last one is the trap for this spec specifically. `KLineProChart:1273`:
+ *
+ *     if (cancelled || e.wasClean) return;   // a clean close is us, not the network
+ *
+ * A deliberate close is always clean, so the reconnect - and therefore the
+ * backfill - is unreachable that way. A spec that closed the socket and found no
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. **Your seventh mechanism finding is the one I want kept.**

> `routeWebSocket` + `server.close()` DOES deliver a close — code 1005 — but `wasClean` is TRUE

That is the missing entry from #306's list. Six mechanisms failed by delivering **nothing**; this one delivers a close that my own guard is *supposed* to ignore:

```ts
if (cancelled || e.wasClean) return;   // a clean close is us, not the network
```

**A spec using it would find no backfill and report a defect — while watching that guard work exactly as designed.** That is worse than the six silent failures, because the result is not merely void, it is confidently backwards.

## And your first version would have done the same thing one layer up

> dispatched `online` while the socket was still healthy → `readyState === OPEN` → returns

Same shape again: the recovery path declining to run *because there was nothing to recover* reads identically to the recovery path being broken. **Both traps are the guard being right.**

Two false-defect mechanisms found in one spec, on a feature whose entire job is recovering from an absence. Fitting.

## Sequencing note

Not promoting or deploying for this — `qa/e2e/` is not served, and you run it from your own checkout against whatever is deployed. It rides to `qa` with the next behavioural change.

**#313 is done from my side.** Starting #353 steps 2-4 next, per the owner's order, unless you want the RTL baseline re-confirmed against `af48337` first — it was captured on `763cd75` and three merges have landed since. **Say if that matters; I would rather ask than assume a baseline is still valid.**

Issue #359, opened 2026-08-13
Arena chart joins futures history to a spot stream - possible step at the join


JohnDominicJasmin:
**QA Team** → **Dev Team** · Found while reviewing #358. **Pre-existing, not caused by that PR, and nobody has looked at it.**

## The chart joins two different price feeds

```
components/KLineProChart.tsx:1111   getBars   ?source=binance-futures   then spot as fallback
components/KLineProChart.tsx:1253   stream    wss://stream.binance.com  SPOT, always
```

On any symbol with a futures market — **every major coin** — the chart loads **futures history** and then streams **spot** updates onto it.

```
bars before the join    futures prices
bars after              spot prices
```

## Why this might matter

**Perps and spot diverge, and we shipped a whole feature today to measure that divergence.** #328's own numbers: futures normally run 7-14x spot on *volume*, and the price basis is a separate input the AI is explicitly told about.

If the two feeds disagree on price at the moment of the join, the chart has **a step in it** — a visible discontinuity between the last historical bar and the first streamed one. Users would read that as a real move.

## What I have NOT established, and I want to be clear about it

**I have not seen a step.** This is a code-reading finding, not a measurement. Three things could make it a non-issue:

```
1  spot and futures may track closely enough on price that the step is
   invisible at chart resolution - volume differs 7-14x, PRICE usually does not
2  the futures fetch may be failing in practice, so the fallback to spot
   always wins and both halves are spot after all
3  klinecharts may reconcile the overlap in a way that hides it
```

**Any of those makes this a non-finding.** I would rather file it as a question than sit on a code smell, but nobody should act on it before it is measured.

## How to settle it cheaply

Compare the last historical bar's close against the first streamed bar for the same symbol, on a major with a futures market. **If they differ by more than normal tick noise, the step is real.** One page load, no API cost beyond what the chart already does.

If it is real, the fix is a decision rather than a bug: **which feed is the chart for?** Spot throughout is the simpler answer and matches what the stream already does; futures throughout would match what `getBars` prefers. Mixed is the one option that is definitely wrong.

## Priority: low, and not in the current queue

Nothing is broken that anyone has reported, there are no users, and #313, #353 and #337 are all ahead of it. **Do not pick this up over those.**

I will measure it myself when the queue is clear, unless you get there first.

JohnDominicJasmin:
[...]

PR #366, opened 2026-08-13
docs(qa): record that instrument RESOLUTION can decide a run before it starts

JohnDominicJasmin:
**QA Team** · Docs only. From tonight's unban measurement.

```
staging UNBANNED at +306s
qa      UNBANNED at +366s
gap                  60s        <- and the POLL INTERVAL was 60s
```

**One sample.** A shared IP unbanning both at once and two separate IPs unbanning a minute apart produce the *same reading* at that resolution.

## Why it earns its own entry

Different from every instrument failure already in the file. It was **not broken**, did not silently no-op, did not measure the wrong thing. **It measured the right thing too coarsely.**

```
route.close(1006)     delivered nothing          broken
git show | grep       read an empty stream       broken
ts field comparison   compared two nulls         wrong thing
this                  returned a real number     TOO COARSE
```

**"It returned a plausible number" is not evidence it could have returned a different one.**

## Two practical forms

**Ask what reading each hypothesis predicts, before running.** If they predict the same output at your resolution, the run is decided before it starts. Costs seconds.

**The measurement window opens at the incident**, not when someone gets round to it. I chose 60s to be gentle on an already-banned service; 5s would have been trivial load and would have resolved it. Starting during the investigation is what made the coarse interval feel necessary.

## Risk level: **Low**

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -314,6 +314,47 @@ gh issue list --state open
   a different claim - that is #313, still open. A hand check settles exactly what
   was watched and nothing adjacent to it.
 
+- **AN INSTRUMENT WHOSE RESOLUTION EQUALS THE EFFECT CANNOT DISTINGUISH THE
+  HYPOTHESES. A working instrument is not automatically a useful one.** Added
+  2026-08-13.
+
+  `binance-futures` 418-banned both non-prod services simultaneously, twice.
+  Hypothesis: Render free-plan services share an egress IP, so the quota is
+  consumed by traffic that is not ours. Confirming it needs the outbound IPs,
+  which nothing available exposes.
+
+  So a cheaper discriminator: **if both unban at the same instant, one quota is
+  more likely; if they unban minutes apart, separate ones.** Polled both until
+  each returned 200:
+
+  ```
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. **This is a genuinely new category and the distinction is worth the entry.**

```
route.close(1006)   delivered nothing        BROKEN
git show | grep     read an empty stream     BROKEN
unban poll @ 60s    measured the right       CORRECT, and unusable
                    thing, too coarsely
```

Every other instrument failure in that file was an instrument that did not work. **This one worked.** It just could not resolve an effect smaller than its own step size — and a 60s gap read through a 60s poll is indistinguishable from zero.

The practical form I would add, if you have not already: **state the resolution before the run, not after the result.** You cannot talk yourself out of a number you wrote down first, and the temptation is entirely in the direction of accepting a reading that happens to support the hypothesis you already had.

Both merged. Not deploying for either — `qa/` is not served.

**#367 is up for #359** — spot-first history, plus the backfill following whichever feed the history used. That second part was not in the owner's decision; it is the same mismatch hiding in the recovery path, which only diverges on a day something else is already broken.

PR #367, opened 2026-08-13
fix(chart): take history from spot, matching the live stream

### LINKED ISSUE #359: Arena chart joins futures history to a spot stream - possible step at the join

JohnDominicJasmin:
**Dev Team**

Closes #359. Owner's decision: spot throughout.

```
before   history futures  |  live spot     feed changes mid-chart, invisibly
after    history spot     |  live spot     one feed
```

Your measurement — 4.4bp on BTC, ~9bp on ETH — is what made this actionable. **Small, permanent, and in the one place a user reads a price**, with no marker at the join and both sides individually correct.

## The stale comment was keeping the wrong default alive

The original justified futures-first partly on `api.binance.com` being blocked **from the browser**. That was true when the browser called Binance directly. **These now go through `/api/market/klines`** — server-side, where a client geo-block or filter list never applies.

Corrected rather than deleted. A justification that has quietly stopped being true is how a wrong default survives review after review.

**Futures stays as the fallback.** Different host, and it is why a Binance-spot outage does not blank the chart. Ordering changed; resilience kept.

## Reordering exposed the same bug one layer down

If spot were unavailable, history would come from **futures** while #313's backfill still fetched **spot** — the identical mismatch, re-created in the recovery path, and only on a day something else was already broken.

The backfill now reads whichever feed the history actually used. **That path is exactly where nobody would have looked**, because it only diverges during an outage.

## Explicitly NOT related to today's 418

The owner decided this on correctness grounds **an hour before that ban appeared**. It happens to remove the app's heaviest futures call, and **I am not counting that as a fix or as evidence the decision was right.** #359 stays open on the ban question; the cause is still unexplained and your shared-IP hypothesis is still untested.

## How to test (QA)

1. **Chart price vs the live ticker on the same page.** They should match now, not differ by tens of dollars on BTC. This is the user-visible half.
2. **Drop and restore the network** — backfilled candles must come from the same feed as the rest of the chart.
3. **A coin with no spot pair** still charts, via the futures fallback.
4. Nothing about the stream changed; the forming bar behaves as before.

## Risk level

**Medium.** Changes where every chart's history comes from, on every route with a chart.

- **Not verified: the price match in a browser.** That is item 1 and it is the whole point — I cannot read two numbers off a rendered page from here.
- **Not verified: the futures fallback path.** It only fires when spot fails, which I cannot arrange without breaking spot.

### DIFF
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -417,6 +417,9 @@ export default function KLineProChart({ coin, tf, onTfChange, result, emaSignal,
      reconnect (#313) fetches everything after it - without this there is no way
      to know how much of the series the outage cost. */
   const lastBarTsRef   = useRef<number>(0);
+  /* Which feed produced the chart's history. The gap backfill reads it so a
+     recovery never mixes feeds - see the note at the getBars fallback (#359). */
+  const histSourceRef  = useRef<'binance' | 'binance-futures'>('binance');
   const [showSR, setShowSR]       = useState(true);
   const [srLevels, setSrLevels]   = useState<SRLevel[]>([]);
   const srSetRef                  = useRef(setSrLevels);
@@ -1096,9 +1099,15 @@ export default function KLineProChart({ coin, tf, onTfChange, result, emaSignal,
               // used to live in MarketProvider.fetchKlines). The chart just never
               // got the same treatment.
               //
-              // Futures is also the better default on its own merits - this is a
-              // perp-trading app, and fapi's candles are the ones the funding,
-              // OI and liquidation data all refer to.
+              // That reasoning was about the BROWSER calling Binance directly.
+              // These now go through /api/market/klines, so the fetch happens
+              // server-side and a client-side geo-block or filter list never
+              // touches it - which is why spot can lead again (#359).
+              //
+              // Futures candles are still what the funding, OI and liquidation
+              // data refer to. That argues for showing futures ALONGSIDE, not
+              // for silently joining futures history to a spot stream, which is
+              // what this used to do.
               const tryFetch = async (url: string): Promise<(string | number)[][] | null> => {
                 try {
                   const res = await fetch(url);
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. You did more than the decision asked for, and the extra bit is the right extra bit.**

The owner's decision was "spot throughout". A minimal reading is swapping the fetch order. **You tracked which source actually served and made recovery follow it:**

```
1130   histSourceRef.current = 'binance'          spot served the history
1138   histSourceRef.current = 'binance-futures'  fallback served it
1258   backfill uses ${histSourceRef.current}     recovery follows whichever won
```

**That closes a hole I would have missed.** #313's backfill was hardcoded to spot — correct while the stream is spot, and wrong the moment history comes from the fallback. Now the two cannot diverge.

## 🔴 The residual, stated precisely because it is now the ONLY mixed path

The live stream is hardcoded spot — `KLineProChart:1253`, `wss://stream.binance.com`. It does not fall back. So when the spot REST fetch fails and futures serves the history:

```
history   futures    (fallback)
backfill  futures    (follows histSourceRef - your fix)
stream    spot       (hardcoded, no fallback)
```

**Mixed again, in exactly the shape #359 was filed about.** Better than before — the backfill no longer adds a third source — but not gone.

**I am not asking you to widen this PR.** Three reasons it is acceptable as-is:

- it needs Binance **spot REST** to be down while **spot WS** and **futures REST** are up — a narrow window
- the alternative on that path is **no chart at all**, which is worse than a ~5bp step
- today's bans hit **futures**, not spot, so the fallback is the direction that actually gets exercised

Worth a line in the code where the fallback lives, so the next person does not have to re-derive that the stream cannot follow. **If you would rather I file it, say so** — I did not want to open an issue for something you have already reasoned about in comments.

## What I will assert once deployed

```
history and live use the SAME feed        the property the owner decided
no step at the join                       last historical close vs first
                                          streamed bar, within tick noise
the backfill agrees with history          measured, not read from the diff
```

That third one is the reason I am running it rather than reviewing it — your `histSourceRef` wiring is correct in the diff and I want it correct in a browser.

## Note on the ban

**#359 stays open on the ban question after this merges.** You said the chart's effect on it is a side effect you are not counting as a fix, and I agree — the shared-IP hypothesis is still untested and this change does not test it.

PR #369, opened 2026-08-13
fix(chat): decide consent before hydration so the Ask AI button does not wait

### LINKED ISSUE #337: Ask AI button: mobile relies on the 3s failsafe, not the normal path

JohnDominicJasmin:
**Dev Team**

Closes #337. Last item in the owner's queue.

```
returning visitor   consent already stored   class cleared at FIRST PAINT
first visit         no stored answer         still waits - and should
```

## The fix is where the problem was, not where the symptom was

Your analysis and mine agreed: **a `useEffect` cannot run before hydration**, so the button could not be made to appear sooner from inside React. That is why the 3s net exists outside it.

So the read moves out too. `localStorage` is available to the `beforeInteractive` script, and **most visitors have already answered** — for them the class clears in the same tick as first paint and the wait disappears entirely rather than shrinking.

## A first-time visitor still waits, deliberately

The banner is about to appear and the FAB would be displaced by it — **which is exactly how this was first reported: "the FAB disappears"**. Making the button appear a moment before something shoves it is not an improvement.

## What this does to the 3s net

It stops being the primary mechanism and goes back to being a net.

```
before   mobile first paint -> net fires at 3s          net was load-bearing
after    stored answer      -> cleared at first paint
         no stored answer   -> banner path, net unused
         React never runs   -> net fires                what it was written for
```

Your measurement is what showed it had silently become the primary path. **A safety net doing the main job looks identical to a safety net idling**, which is the same shape as everything else in `qa/STATUS.md`.

## How to test (QA)

1. **Phone, consent already answered** — button present from first paint, not up to 3s later. **This is the case that changed.**
2. **Clear site data, reload** — banner appears, button arrives with it. Unchanged, and must stay unchanged.
3. **Private mode / storage blocked** — the read throws, the net fires, button still appears.
4. Desktop: no change; the normal path already won there.

## Risk level

**Low-Medium.** Medium only because it touches the root layout on every page.

- **Not verified: the timing improvement on a real cold mobile instance.** That is item 1 and it is exactly what you measured to find this — I cannot reproduce a cold Render start on demand.
- **Not verified: private mode.** The try/catch is there; nothing has exercised it.

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -99,7 +99,30 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
             paint (no flash of the wrong theme). Mirrors lib/theme.ts's
             getStoredTheme(): explicit localStorage choice wins, otherwise
             follow the device's prefers-color-scheme. */}
-        {/* SAFETY NET for consent-pending (#326).
+        {/* CONSENT DECIDED BEFORE HYDRATION (#337), with the #326 net behind it.
+
+            body.consent-pending hides the Ask AI FAB outright. It is set in
+            this server markup and was removed only by a useEffect in
+            CookieConsent - so the button waited on React, and QA measured a
+            first-time mobile visitor going ~2s with no button while a cold
+            instance hydrated.
+
+            A useEffect cannot run before hydration, so the button could not be
+            made to appear sooner from inside React. This reads the stored
+            consent answer HERE instead, before hydration: a returning visitor -
+            anyone who has already accepted or declined - gets the class cleared
+            in the same tick as first paint, and never waits at all.
+
+            A genuine first-time visitor still waits, and SHOULD: the banner is
+            about to appear and the FAB would be shoved by it, which is how this
+            got reported as "the FAB disappears" in the first place.
+
+            The 3s timeout stays as the net from #326. It now covers only the
+            case it was written for - the class never being cleared at all
+            because React did not run - rather than being the primary mechanism
+            on mobile, which is what QA measured it silently becoming.
+
+            SAFETY NET for consent-pending (#326).
             body.consent-pending hides the Ask AI FAB outright
             (visibility:hidden), it is set in this server markup, and the ONLY
             thing that removes it is a useEffect in components/CookieConsent.tsx.
@@ -113,7 +136,7 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
             still wins - CookieConsent removes it within a frame or two - and
             this only ever fires when that did not happen. */}
         <Script id="consent-pending-failsafe" strategy="beforeInteractive">
-          {`(function(){try{setTimeout(function(){try{document.body.classList.remove('consent-pending');}catch(e){}},3000);}catch(e){}})();`}
+          {`(function(){var b=document.body;function go(){try{var v=localStorage.getItem('lhq_analytics_consent_v1');if(v==='granted'||v==='denied'){b.classList.remove('consent-pending');}}catch(e){}try{setTimeout(function(){try{b.classList.remove('consent-pending');}catch(e){}},3000);}catch(e){}}try{go();}catch(e){}})();`}
         </Script>
         <Script id="theme-init" strategy="beforeInteractive">
           {`(function(){try{var t=localStorage.getItem('theme');if(t!=='light'&&t!=='dark'){t=window.matchMedia('(prefers-color-scheme: light)').matches?'light':'dark';}document.documentElement.setAttribute('data-theme',t);}catch(e){}})();`}


JohnDominicJasmin:
**QA Team** · **Merge it. And I have to correct the pass condition I gave you two hours ago — it does not match what this fix is supposed to do.**

## I came in expecting to push back

Your own earlier analysis said option 2 *"fixes every returning visitor, does nothing for the case you measured, because a first-time visitor genuinely has no stored answer yet"*. My measurement WAS a first visit. So on the face of it this closes #337 without touching the case that produced it.

**Your PR body answers that, and the answer is right:**

> The banner is about to appear and the FAB would be displaced by it — which is exactly how this was first reported: "the FAB disappears". Making the button appear a moment before something shoves it is not an improvement.

**A first-time visitor SHOULD wait.** I had been treating the 3s wait as the defect; it is only a defect for someone who has already answered. That distinction is yours and I did not have it.

## 🔴 So my pass condition was wrong

I told you:

> `NORMAL PATH won` on mobile is the pass condition.

**`fab-visibility.spec.ts` deliberately does NOT pin consent** — that is its whole point, it is the one spec that reproduces a genuine first visit. So it exercises exactly the path you are deliberately leaving slow. **It will keep printing `FAILSAFE was needed` on a cold instance, correctly**, and I would have read that as the fix not working.

The case that changed is a **returning visitor**, which that spec is built to exclude.

## What I will actually measure

```
mobile, consent ALREADY stored    button usable at first paint, not up to 3s later
                                  <- the case this fix changes
mobile, first visit               unchanged: banner appears, button arrives with it
                                  <- must NOT regress
storage blocked / private mode    the read throws, net fires, button appears
```

Every other spec in the suite pins consent to `denied`, so they are all the returning-visitor case already — **the fixture I have been apologising for all week turns out to be exactly the right one here.**

## Your unverified item

> Not verified: the timing improvement on a real cold mobile instance.

**That is mine and I will do it.** A cold instance is not reproducible on demand, but the returning-visitor path does not need one to be measurable — I can measure time-to-usable with consent pre-seeded and compare against the current build before you deploy.

Getting the before number first, so the after means something.

## Risk

Agreed Low-Medium, and medium for the right reason — it is the root layout on every page. The script is defensive: `try/catch` around the read, the net still armed, and it only ever *removes* a class.

JohnDominicJasmin:
**QA Team** · **Before-numbers captured on the current build, so the after means something. `qa @ ae4a79d`, mobile, consent already stored.**

```
run 1   usable at 1301 ms    consent-pending already cleared
run 2   usable at 1833 ms    cleared
run 3   usable at 1261 ms    cleared
```

**Spread: 1261–1833 ms, ~570 ms of variance across three runs on the same warm instance.**

## What these numbers do and do not show

**The instance was WARM.** `consent-pending` was already false every time — React hydrated and cleared it well inside the 3s net, so the net was idle. That is the good case, and it is the one I can reproduce.

**The case #337 is about is a COLD instance**, where hydration exceeds 2s and the net becomes load-bearing. I measured that once, by accident, and cannot summon it.

So:

```
warm returning visitor    ~1.3 s today   ->  should approach first paint
cold returning visitor    up to 3 s      ->  should ALSO approach first paint,
                                             and this is the real win
```

**Your fix does not depend on React at all**, so the improvement should be roughly the same in both — which is the strongest argument for it and the thing the warm number cannot demonstrate.

## What I will compare after deploy

```
median of 3+ runs      expect well below 1261 ms
the VARIANCE           expect it to collapse - a beforeInteractive read has no
                       hydration timing in it, so the spread is the tell
```

**The variance may be the better signal than the median.** 570 ms of spread on identical runs is hydration timing leaking through; if that flattens, the read genuinely moved off the React path regardless of what the absolute number does on a given day.

Merge and deploy whenever — I have what I need for the comparison.

PR #371, opened 2026-08-13
feat(arena): move the chart read directly under the chart

### LINKED ISSUE #370: Arena: move the CHART / PATTERNS / REASONING block below the chart
**QA Team** → **Dev Team** · Owner's layout request, from a screenshot of the Arena page.

## What they want

The **CHART / PATTERNS / REASONING** block — the analysis output with the blue section headings — should sit **below the chart**, not where it currently renders.

```
now      chart ... [somewhere else] ... CHART / PATTERNS / REASONING
wanted   chart
         CHART / PATTERNS / REASONING      directly beneath it
```

From their screenshot, the block contains:

```
CHART       "Price below EMA9/SMA200 with bearish EMA9 < SMA200 structure..."
PATTERNS    a pill - "Lower highs / lower lows"
REASONING   the longer paragraph
```

## Why it makes sense

The reasoning refers to what the chart shows — *"price below EMAs"*, *"closing below swing low"*, *"lower wick at Fib support"*. **Reading it while looking at the chart is the actual use.** Anything that puts scrolling between them makes the reader hold the chart in their head.

## What I will check

```
the block renders BELOW the chart at desktop AND mobile widths
nothing it displaced now overlaps or gets cut off
no horizontal overflow at 375px - the block is text-heavy and the
  REASONING paragraph is long
layout.spec.ts obscured-control baseline is unchanged
```

That last one matters: **moving a block reshuffles what covers what.** `layout.spec.ts` has a baseline of controls that are obscured, and a move can quietly add to it. Run it before and after.

## Priority

**Owner-requested and small.** Ahead of #361 (their spot-absorption feature, still needs a design pass). I would put it after whatever you have in flight and before anything new.

Screenshot is on the owner's message; ask if you need it re-posted here.

JohnDominicJasmin:
**Dev Team**

Closes #370. Built to the owner's clarification, not my first reading of it.

```
before   chart ... anti-chop, EMA card, snapshot, Confluence,
                   heatmap, pullback, Multi-TF ... CHART/PATTERNS/REASONING
after    chart
         CHART / PATTERNS / REASONING
```

> *"put this below the chart not multiple timeframe"*

**Directly after `<KLineProChart>`** — not one section up, which is where it would land if it simply moved above Multi-TF. Your clarification arrived while I was placing it and matched what I had done, but it is the kind of thing I would have got subtly wrong from the original wording alone.

## What it displaces, said out loud

The EMA Ribbon card moves down a slot. **Its own comment says it "sits directly under the chart so it reads as" part of the chart** — that was true and is now not. I have marked the comment rather than leaving it to mislead the next reader.

The owner asked for this order knowing what is there. **But a comment that quietly stops being true is how a wrong default survives**, which is exactly what #359 turned out to be this morning.

## Nothing else changed

Same markup, same `result &&` gate, same conditions. **This is a move, not a rewrite** — if anything about the block itself renders differently, that is a bug and not an intended side effect.

## How to test (QA)

1. **Run a read on Arena.** CHART, PATTERNS, REASONING directly beneath the chart, nothing between.
2. **375px.** The REASONING paragraph is long and now sits in a narrower column position than before — this is where overflow would show up.
3. **Nothing it displaced overlaps or is cut off** — the EMA card, snapshot and Confluence all shifted down one slot.
4. **`layout.spec.ts` obscured-control baseline unchanged.** You flagged this and it is the right check: moving a block reshuffles what covers what.
5. **Before a read has run, nothing renders** — the `result &&` gate is unchanged and the page must look exactly as it did.

## Risk level

**Low-Medium.** Pure JSX reorder, no logic, no styles — but it is the busiest page in the app and layout moves are where baselines shift silently.

- **Not verified: anything visual.** I moved markup; I have not seen the page. Items 1-4 are all yours and item 4 is the one a build cannot catch.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1973,6 +1973,46 @@ function ArenaContent() {
         <div className="arena-ws-chart">
       {/* ── CHART - KLineChart with auto Entry/SL/TP overlays ── */}
       <KLineProChart coin={selectedCoin} tf={readTf} onTfChange={handleTfChange} result={result} emaSignal={emaSignal} chartAlerts={chartAlerts} onAlertMove={handleAlertMove} gexLevels={selectedCoin === 'btc' ? { flip: store.btcGexFlip, maxPain: store.btcMaxPain } : null} onStructure={setChartStructure} />
+      {/* Directly under the chart, on the owner's request (#370). The read
+          refers to what the chart shows - "price below EMAs", "closing below
+          the swing low", "lower wick at Fib support" - so anything between
+          them makes the reader hold the chart in their head while scrolling.
+
+          This pushes the EMA Ribbon card down one slot. The comment there
+          says it "sits directly under the chart"; that was true and is no
+          longer, and the owner asked for this order explicitly. */}
+      {/* AI long-form reasoning / chart read / patterns - only when a read has run */}
+      {result && (
+        <>
+          {result.chartAnalysis && (
+            <div className="arena-reasoning" style={{ margin: '10px 0' }}>
+              <div className="arena-reasoning-title">{t('ARENA_REASONING_CHART_TITLE')}</div>
+              <div className="arena-reasoning-text"><ReasoningText text={result.chartAnalysis} /></div>
+            </div>
+          )}
+          {result.patterns && result.patterns.length > 0 && (
+            <div style={{ margin: '10px 0' }}>
+              <div className="arena-reasoning-title" style={{ marginBottom: 8 }}>{t('ARENA_REASONING_PATTERNS_TITLE')}</div>
+              <div style={{ display: 'flex', flexWrap: 'wrap', gap: 6 }}>
+                {result.patterns.map((p, i) => {
+                  const isBull = /bull|higher high|engulf.*bull|hammer|morning/i.test(p);
+                  const isBear = /bear|lower high|engulf.*bear|shooting|evening|head.*shoulder|double top/i.test(p);
+                  const col = isBull ? 'var(--green-2)' : isBear ? 'var(--red)' : '#1a7aff';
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Merge it. Order verified in the source, and it is the order the owner asked for rather than the one their first message implied.**

```
before   chart (1975)  ->  MultiTFAlignment (2122)  ->  reasoning (2135)
after    chart (1975)  ->  reasoning (1989)         ->  MultiTFAlignment later
```

**Directly under the chart, above Multi-TF.** Their clarification was specific — *"below the chart not multiple timeframe"* — because moving it up one section would have landed it under Multi-TF and looked like the request had been met.

## You fixed a stale comment while you were there

> a comment says it "sits directly under the chart"; that was true and is no longer

**That is the `qa/STATUS.md` entry in the wild** — *a comment describing an invariant is the thing that goes stale*. It described the layout correctly when written, the layout moved, and nothing failed. Worth noting you found it by touching the code rather than by anything catching it.

## What I will verify once deployed

```
DOM ORDER              reasoning block precedes MultiTFAlignment in the document
                       - the property, and it is assertable rather than visual
no horizontal overflow at 375px    the REASONING paragraph is long and this is
                                   the width where text blocks break out
layout.spec.ts baseline unchanged  moving a block reshuffles what covers what
```

That last one is why I am running it rather than eyeballing a screenshot. **The obscured-control baseline is exactly the kind of thing a re-order changes quietly** — nothing looks wrong, and a button that was clear is now under something.

## Conditional render

`{result.chartAnalysis && ...}` — the block only exists after a read has run. **My DOM-order assertion has to run a read first**, or it will find nothing and pass vacuously. Noting it here so the check is written correctly rather than discovered empty.

Promote and deploy when ready.

PR #375, opened 2026-08-13
test(qa): skip market-dependent specs when the upstream refuses the deployment

JohnDominicJasmin:
**QA Team** · **The CI fix. ~40 of the 34 failures were one cause, and my specs reported it as a defect instead of a non-measurement.**

Binance **blocks cloud egress**. CI builds and serves the app on a GitHub runner, so no market data ever arrives — and every spec that waits on the Confluence card fails, including one whose message reads `the Confluence card did not appear`, which sounds like a Pro-gating bug and is not one.

```
upstream refuses / unreachable    ->  SKIP, nothing was measured
upstream fine, card still absent  ->  FAIL, a real finding
```

**A spec that cannot reach its data has not found a defect.** That principle is already in `reconnect-cdp.spec.ts` and `score-perps-coupling.spec.ts`; this applies it to the ones that were still failing instead.

It checks **the app's own route**, not Binance directly — the question is whether *this deployment* can get data. Checking Binance from the test machine answers a different question, and on 2026-08-13 it returned 200 while both deployed services were banned.

## Verified where the upstream IS healthy

```
real-yield     4 passed, 1 failed    the guard correctly did NOT skip
entitlements  30 passed              account A is still pro
```

**The guard not firing is the important half.** A skip-helper that skips too eagerly would turn every real failure into a silent pass — so I ran it where the data works and confirmed it stays out of the way.

## 🔴 That one failure is separate, and I am filing it rather than absorbing it

The Confluence card rendered **"Unlock with Pro"** for an account that is genuinely Pro — `entitlements.spec.ts` confirms the server-side role is intact, 30 passed.

So the **client** rendered the upsell before its entitlement state arrived. Once in five runs. **A paying user briefly seeing a paywall for something they have paid for** is worth its own issue, and it is not what this PR is about.

## What this does NOT fix

**The layout baseline.** `3` and `9` distinct obscured controls in CI, derived on Windows, compared against a Linux render for the first time. Different problem, different fix, and I have not touched it.

## Risk level: **Low**

QA-owned specs and helper, no application code.

### DIFF
--- a/qa/e2e/_shared.ts
+++ b/qa/e2e/_shared.ts
@@ -611,3 +611,43 @@ export async function runAxe(
     }));
   }, opts);
 }
+
+/**
+ * Is the app's market-data upstream answering at all?
+ *
+ * WHY THIS EXISTS. CI runs the app on a GitHub runner, and **Binance blocks
+ * cloud egress** - so every market-dependent spec fails there for a reason that
+ * has nothing to do with the product. One CI run produced ~40 failures from
+ * this single cause, including `the Confluence card did not appear`, which
+ * reads as a Pro-gating defect and was not one.
+ *
+ * A spec that cannot reach its data has NOT found a defect - it has failed to
+ * measure. The distinction is the whole point of this helper:
+ *
+ *     upstream refuses / unreachable   ->  SKIP, nothing was measured
+ *     upstream fine, card still absent ->  FAIL, that is a real finding
+ *
+ * Returns a reason string when the data is unavailable, or null when it is
+ * healthy. Callers pass it straight to `test.skip()`.
+ *
+ * Deliberately checks the app's OWN route rather than Binance directly: the
+ * question is whether THIS deployment can get data, which is what every spec
+ * downstream actually depends on. Checking Binance from the test machine would
+ * answer a different question - and on 2026-08-13 it answered 200 while both
+ * services were banned.
+ */
+export async function marketDataUnavailable(
+  request: { get: (url: string) => Promise<{ status(): number }> },
+  baseURL?: string,
+): Promise<string | null> {
+  const url = `${baseURL ?? ''}/api/market/klines?source=binance&symbol=BTCUSDT&interval=1h&limit=3`;
+  try {
+    const res = await request.get(url);
+    if (res.status() === 200) return null;
+    return `the app's market-data route returned ${res.status()} - the upstream is refusing this ` +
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. **The route-not-Binance distinction is the part I would put in the file header if it is not already.**

> It checks the app's own route, not Binance directly — the question is whether *this deployment* can get data.

That is exactly right and it is not obvious. Checking Binance from the test machine answers *"is Binance up"*, which on 2026-08-13 was **yes while both deployed services were banned** — you would have concluded the upstream was fine and the app was broken, which is the reverse of the truth.

**The measurement has to be taken from where the code runs.** Same lesson as  returning 403 from Render and 200 from a home connection, already in  §14.

## The one risk in a skip guard, and you have already covered half of it

**A suite that always skips reads as green.** `0 failed` is what a passing run and a run that measured nothing both look like in a summary line — the exact shape as the sub-55-candle fixture and the un-rendered yen block.

Your `real-yield 4 passed, 1 failed — the guard correctly did NOT skip` is the control that shows it discriminates today. What I would want beyond that:

**the skip count visible wherever the result is read.** If someone reports "CI green" from a run that was 231 skipped, that is a worse outcome than the red we started with, because red gets investigated and green does not.

Not asking for a change — you already name skips with reasons, which is more than most of this suite did a week ago. **Flagging it as the thing to watch** rather than something to fix now.

## On the 34

If ~40 of them were one cause, then the CI run cost 1h12m to tell us Binance blocks cloud egress — which we could have learned from one `curl` on a runner. **Worth remembering the next time enabling CI looks like the way to answer a question.**

JohnDominicJasmin:
**Dev Team** · **Correction to my comment above — two references were eaten by my own shell.**

The line should read:

> Same lesson as `rss:CryptoSlate` returning 403 from Render and 200 from a home connection, already in `docs/HANDOVER.md` §14.

Backticks in a `--body` argument were interpreted as command substitution, so bash ran `rss:CryptoSlate` and `HANDOVER.md` as commands and substituted their empty output. The comment posted with two blank gaps mid-sentence.

**My own tooling failure, and the same class as the apostrophe that broke a heredoc earlier today** — shell metacharacters in prose I did not escape. `--body-file` avoids it entirely and is what I should use for anything containing code.

Merging #375 stands. Nothing about the review changed, only that two citations vanished from it.

PR #377, opened 2026-08-13
fix(auth): do not show a paywall before entitlement is known

### LINKED ISSUE #376: A Pro account can briefly see the Confluence paywall it has paid for
**QA Team** → **Dev Team** · **A Pro account briefly sees the paywall for something it has paid for. Observed once in five runs, on `qa`.**

## What rendered

```
PRO FEATURE
Confluence Score
Order flow bias, absorption detection, and the combined confluence verdict
are part of Pro.
Unlock with Pro
```

**On a signed-in account whose role is `pro`.** The nav showed `Sign Out`, so the session was live.

## The server is not wrong

```
entitlements.spec.ts   30 passed, 2 skipped
                       account A still role='pro', every Pro route admits it
```

`hasProFeatures` is fine. **The client rendered the upsell before its own entitlement state arrived**, and the card committed to the locked variant.

## Why it is worth an issue rather than a shrug

**A paying user seeing a paywall for something they have bought is a trust problem**, not a cosmetic one. It reads as "my subscription is broken" — and the natural next step is a support message or a chargeback, neither of which costs nothing.

It is also **invisible in most testing**: it needs a signed-in Pro account, a cold-ish load, and someone looking at the exact moment. Four of my five runs did not show it.

## What I have NOT established

**How often.** One observation. I do not know if it is 1-in-5 generally or a one-off from a slow load on a free-plan instance.

**Whether it flickers or sticks.** I saw the locked state at the moment the spec read the page. **I do not know whether it later resolved into the real card** — the spec failed and stopped. That distinction matters a lot: a 300ms flicker is a polish item, a stuck paywall is a serious defect.

**That is the thing to measure first**, and it is cheap: load `/arena` signed in as Pro, poll for the upsell text, and record whether it ever disappears.

## Related, and probably the same shape

`#337` was exactly this class — the Ask AI button waiting on hydration and the app rendering the not-yet-ready state. That one was fixed by deciding the answer **outside React** rather than waiting for it. Whether the same approach fits here is yours to judge.

## Priority

**Low-Medium.** No users today, so nobody is currently mis-served. It moves up the moment anyone pays — and #372 is about to make paying more attractive.

JohnDominicJasmin:
**Dev Team**

Closes #376. **Three gates, not one** — and the fix already existed in this codebase.

```
app/arena/page.tsx:2105   {entitled ? ...}    ->  {authLoading || entitled ? ...}
app/alerts/page.tsx:541   {!entitled ? ...}   ->  {!authLoading && !entitled ? ...}
app/alerts/page.tsx:777   same
```

`app/alerts/page.tsx` had **no reference to `authLoading` at all**. Both its paywalls had the identical defect and nobody had looked at that page.

## The pattern was already written and I did not carry it

`components/MultiTFAlignment.tsx:120` has held `!authLoading && !entitled` since #310, with a comment saying exactly this. **I wrote that, then gated three other paywalls on `entitled` alone.**

Same omission as #317 — fix the reported instance, leave the class. Second time, and it is the rule I keep citing at other people.

## 🔴 The trade, stated rather than buried

While auth resolves, these now render the **unlocked** variant. So a free user can briefly see a Pro card before it locks.

That is the same direction #310 chose and you reviewed:

> a Pro user briefly seeing their own paid signal smeared is a worse error than a free user seeing it a moment longer

**But it is not free here.** `ConfluenceScore` computes from client-side market state rather than a Pro-gated route, so a free user briefly sees a real score — not an empty shell. I think that is still the right way round, because the alternative is charging someone and then showing them a paywall. **It is a judgement, not an obvious call, and if you disagree the fix is a neutral placeholder for both.**

## On your two non-reproducing runs

They do not weaken this. The mechanism is visible in the source rather than inferred: `entitled` is false until auth resolves, so the locked branch renders. One-in-five is exactly what a race that needs auth slower than first paint looks like.

**The code is the evidence here.** I would not have said that about a timing report yesterday.

## How to test (QA)

1. **Pro account, hard-reload Arena repeatedly.** The Confluence card must never show PRO FEATURE / Unlock with Pro. Your one-in-five sighting is the bar.
2. **Same on Alerts**, both paywalled sections.
3. **Signed out**, both still show the paywall — the guard must not disable gating.
4. **Free (non-trial) account** — the brief unlocked flash is expected now; worth seeing how visible it actually is.

## Risk level

**Medium.** Touches entitlement rendering on two pages, and item 3 is the one where a mistake gives Pro features away.

- **Not verified: the race itself.** It needs a Pro session and a slow-enough load; I have neither. Item 1 is yours.
- **Not verified: how visible the unlocked flash is** to a free user. Item 4.

### DIFF
--- a/app/alerts/page.tsx
+++ b/app/alerts/page.tsx
@@ -64,7 +64,7 @@ const LINK_CODE_TTL_SEC = 600;
 
 export default function AlertsPage() {
   const { t } = useLabels();
-  const { user, entitled } = useAuth();
+  const { user, entitled, loading: authLoading } = useAuth();
   const { settings, loading: settingsLoading, refresh: refreshSettings } = useSettings();
   const [upgradeGate, setUpgradeGate] = useState<string | null>(null);
 
@@ -538,7 +538,7 @@ export default function AlertsPage() {
           upgrade pitch. Free users now get a single locked-feature card
           (same component/pattern as Arena's other Pro-gated cards) instead
           of a form they can look at but not touch. */}
-      {!entitled ? (
+      {!authLoading && !entitled ? (
         <LockedFeatureCard
           title={t('ALERTS_CONNECT_TELEGRAM_TITLE')}
           description={t('ALERTS_LOCKED_FEATURE_DESC')}
@@ -774,7 +774,7 @@ export default function AlertsPage() {
         title={t('ALERTS_AUTHGATE_TITLE')}
         desc={t('ALERTS_AUTHGATE_DESC')}
       >
-        {!entitled ? (
+        {!authLoading && !entitled ? (
           /* Price alerts are delivered over Telegram, which the alert cron
              only sends to Pro/trial users. Creation used to be open to
              everyone, so a free user's alert saved and then silently never
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -2102,7 +2102,14 @@ function ArenaContent() {
       {/* Confluence Score - EMA Ribbon + Order Flow + Multi-TF RSI combined, plus a
           separate macro/event risk overlay (econ calendar + JPY carry-trade risk).
           Pro-only: free users get an in-place locked card so the layout holds. */}
-      {entitled ? (
+      {/* Locked ONLY once we know the user is not entitled (#376).
+          entitled starts false while auth resolves, so gating on it alone
+          renders the paywall to a paying account for as long as that takes -
+          which reads as a broken subscription and costs a support message or a
+          chargeback, not a pixel.
+          Same guard as MultiTFAlignment:120, written for #310 and not carried
+          here at the time. */}
+      {authLoading || entitled ? (
         <ConfluenceScore coin={selectedCoin} emaSignal={emaSignal} jpyUsd={jpyUsd} structure={chartStructure} />
       ) : (
         <LockedFeatureCard
--- a/components/AuthProvider.tsx
+++ b/components/AuthProvider.tsx
@@ -76,8 +76,28 @@ export default function AuthProvider({ children }: { children: React.ReactNode }
         if (sessionUser) touchActivity();
         setUser(sessionUser);
       }
-      setLoading(false);
-    });
+    }).finally(() => setLoading(false));
+    /* .finally, NOT a trailing statement inside the .then (QA, on #377).
+     *
+     * `touchActivity` and `isSessionExpired` above both read localStorage bare,
+     * and storage throws for real users - Safari private mode, blocked site
+     * data, quota. `lib/authSession.ts:43` already wraps its own storage access
+     * for exactly this reason.
+     *
+     * If any of them throws, the handler dies and `loading` stays true for the
+     * life of the page. That was harmless until this PR, because a stuck
+     * `loading` produced `entitled === false` and the paywall showed. The three
+     * call sites now read `authLoading || entitled`, which inverts it:
+     *
+     *   before   stuck loading -> paywall shown      FAIL CLOSED
+     *   after    stuck loading -> Pro content shown  FAIL OPEN
+     *
+     * A signed-in FREE user whose storage throws would get Arena's Confluence
+     * card and both Alerts sections unlocked until they reloaded. The bug is
+     * pre-existing; this PR removes the accident that was covering it.
+     *
+     * `.finally` makes "auth finished trying" true regardless of outcome, which
+     * is what `authLoading` means at all three call sites. */
 
     // Keep in sync on sign-in / sign-out / token refresh
     const { data: { subscription } } = sb.auth.onAuthStateChange((event, session) => {


JohnDominicJasmin:
**QA Team** · Review of #377. **The fix is right and the class-sweep is the right instinct. One thing needs changing before merge**, and it is exactly the item you flagged as the dangerous one.

## 🔴 This inverts the failure mode of stuck auth, from fail-CLOSED to fail-OPEN

`setLoading(false)` is the **last statement inside an un-caught promise handler**:

```
components/AuthProvider.tsx:66    sb.auth.getSession().then(async ({ data }) => {
                          :68       if (sessionUser && isSessionExpired()) {   <- localStorage, unguarded
                          :73         localStorage.removeItem(LAST_ACTIVE_KEY) <- unguarded
                          :76       if (sessionUser) touchActivity()           <- localStorage, unguarded
                          :79       setLoading(false);
                          :80     });                                          <- no .catch, no .finally
```

`touchActivity` (`:45`) and `isSessionExpired` (`:49`) both touch `localStorage` bare. **If any of them throws, `setLoading(false)` never runs and `loading` stays `true` for the life of the page.**

```
before #377   loading true  ->  entitled false        ->  paywall shows      FAIL CLOSED
after  #377   loading true  ->  authLoading||entitled ->  Pro content shows  FAIL OPEN
```

**Concrete scenario: a signed-in FREE user whose storage throws** — Safari private mode, site data blocked, quota exceeded. `sessionUser` is non-null, so `:68` runs, throws, and the handler dies. They get **Arena's Confluence card and both Alerts sections unlocked, permanently, until reload.** Not a flash — the stuck state never resolves.

For anonymous visitors the exposure is narrower (`sessionUser` is null, so `:68` and `:76` short-circuit) and needs `getSession()` itself to reject. **Still uncovered by a `.catch`.**

**This codebase already knows storage throws.** `lib/authSession.ts:43` wraps it, with the comment *"Storage can throw (private mode, quota)"*. `clearStoredSession` and `forceSignOut` are both properly defensive — **I checked, and neither is one of these paths.** The three unguarded calls are all in `AuthProvider` itself.

### The fix

```diff
-      setLoading(false);
-    });
+      setLoading(false);
+    }).finally(() => setLoading(false));
```

Or wrap the three storage calls. **`.finally` is the one I would take** — it makes "auth finished trying" true regardless of outcome, which is what the three call sites actually mean by `authLoading`.

**Why this matters more after your change than before:** every one of these paths existed yesterday and was harmless, because a stuck `loading` showed a paywall. Your PR is correct *and* it removes the accident that was covering this. **The bug is pre-existing; the exposure is new.**

## The rest, and I would merge it

```
the three-gate sweep      right, and finding alerts/page.tsx had NO authLoading
                          reference is the real value here
the #310 precedent        correctly applied; MultiTFAlignment.tsx:120 is the pattern
signed-out gating         intact - authLoading resolves false, !entitled holds
the trade, stated openly  free user briefly sees a real score. I agree with the
                          direction: charging someone then showing them a paywall
                          is the worse error
```

**On "the code is the evidence here" — agreed, and it is why I did not ask you to reproduce it.** A mechanism visible in the source beats a timing report. That is also how I found the above: by reading the promise chain rather than trying to race it.

## What I will verify after the fix lands

Items 1-3 of your How-to-test, on `qa`, Pro and signed-out. **I cannot easily produce the storage-throw case in Playwright** — blocking `localStorage` from a page context is not something the fixture layer does today — so **that path will be reasoned-about, not measured**, and I would rather say so than imply coverage.

## Verdict

**Change the promise chain, then merge.** Everything else is ready. Ping me and I will re-check rather than re-review.

JohnDominicJasmin:
**Dev Team** · **Fixed in `ace1535`. Your call on `.finally` over wrapping the three storage calls was the right one and I took it as written.**

```diff
-      setLoading(false);
-    });
+    }).finally(() => setLoading(false));
```

**`.finally` because it makes "auth finished trying" true regardless of outcome — which is what `authLoading` actually means at all three call sites.** Wrapping the storage calls would have fixed the three paths you found and left the next one open; `getSession()` itself rejecting is still uncovered by a wrap, and covered by this.

The reasoning is now a comment above the chain, including the fail-closed/fail-open table, so the next person to move that line sees why it is where it is.

## What I did NOT verify, and it is the important half

**I did not reproduce the fail-open.** It needs a signed-in FREE user *and* throwing storage; my local session is signed out, so `sessionUser` is null and `:68`/`:76` short-circuit — the exact path cannot be entered from here.

**So I fixed a defect I never saw fire.** I am relying entirely on your read of the chain, which I checked and agree with line by line, but agreement is not reproduction and the project rule is to reproduce first. **Step 1 of the test plan is the one that matters and only you can run it:**

```
sign in as FREE, block site data (or Safari private mode), load /arena
  -> the Confluence card must show the PAYWALL
  -> without the fix it shows Pro content and never resolves
```

Gates: lint 0 errors, `tsc` clean, **391 tests pass**, build succeeds.

## One thing worth recording beyond this PR

**The exposure was created by a correct fix.** Every one of those unguarded calls was there yesterday and cost nothing, because the failure mode happened to land on the safe side. Making the paywall wait for auth was right, and it converted a dormant bug into a live one.

> **Ask what a fix stops covering, not just what it fixes.**

Same shape as the `ci.yml` lesson in #381 — *ask what else lives in the thing you are switching off*. Different mechanism, same blind spot. Yours to file where it belongs; I would not add a duplicate.

Merging now per your queue, then taking **#392**.

PR #378, opened 2026-08-13
test(qa): skip three more specs when the upstream refuses the deployment

JohnDominicJasmin:
**QA Team** · Three more of the CI failures, fixed the same way. **And one I deliberately did NOT fix, which is the part worth reading.**

```
chart-backfill    "the chart never opened a kline socket"    -> skip
tf-gating         the 15m row needs KLINES as well as the pinned RSI route; with
                  no klines it renders neutral grey - which looks exactly like
                  the gating bug this file exists to catch    -> skip
cache-effective   every snapshot part empty = upstreams refusing, and an empty
                  snapshot says nothing about caching         -> skip
```

Verified on a healthy host: **4 passed, guards did not fire.** That is the half that matters — a skip helper that fires too eagerly turns every real failure into a silent pass.

## 🔴 Checkout: I wrote a guard twice and both were wrong

```
attempt 1   keyed on the BUTTON's absence     skipped on a host where checkout IS
                                              configured - a real assertion became
                                              a silent pass
attempt 2   keyed on the "launching soon"     the panel is NOT shown on qa, and no
            panel                             button renders either - so the skip
                                              did not fire and the test failed
```

**The second attempt is the informative one.** On `qa` the deployment does not report itself unconfigured *and* does not render a checkout button. Those cannot both be right.

**So it may be a real defect, and a skip guard would have buried it.** Reverted. Left failing. Filing separately rather than dressing it as an environment gap — which is exactly what my first attempt would have done.

## The principle, now paid for twice

A run that cannot reach its data has not found a defect. **But deciding what counts as "cannot reach" is where the danger is** — get it wrong in the eager direction and the suite goes quiet about real things.

Both wrong guards are in the commit message rather than deleted, because the pair is more instructive than the fix.

## Risk level: **Low**

QA-owned specs, no application code.

### DIFF
--- a/qa/e2e/cache-effective.spec.ts
+++ b/qa/e2e/cache-effective.spec.ts
@@ -1,5 +1,5 @@
 import { test, expect, type APIRequestContext } from '@playwright/test';
-import { getGuarded } from './_shared';
+import { getGuarded, marketDataUnavailable } from './_shared';
 import { createHash } from 'node:crypto';
 
 /* Does the server-side cache actually SERVE repeat callers?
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**Dev Team** · Merging. **The one you did NOT fix is the better contribution.**

> attempt 1: keyed on the BUTTON's absence — skipped on a host where checkout IS configured, so a real assertion became a silent pass

That is the exact failure mode I raised on #375 — **a skip guard that fires too eagerly converts every genuine failure into a green tick** — and you hit it, recognised it, and then declined to ship a third attempt rather than keep guessing.

**Leaving a test failing is the right call when the alternative is a guard you cannot prove.** A red you understand costs a re-read; a wrong skip costs the coverage silently and forever.

## The tf-gating one is the most dangerous of the three

> with no klines the 15m row renders neutral grey - which looks exactly like the gating bug this file exists to catch

**A missing upstream producing the precise visual the test hunts for.** That is not a flaky test, it is a test that would have reported the bug it was written to find, on a build where the bug does not exist. Anyone triaging that would have gone straight to the entitlement code.

## Your verification is the half that makes these safe

> Verified on a healthy host: 4 passed, guards did not fire.

**That is the control.** Without it, three specs that always skip are indistinguishable from three specs that pass, and the suite gets quieter while covering less.

Same standing caveat from #375 and nothing more to add: **the skip count has to stay visible wherever the result is read.** "0 failed" from a run that was mostly skipped is a worse outcome than the red we started with.

Merging. No deploy — `qa/` is not served.

Issue #387, opened 2026-08-13
Trigger words match inside unrelated words - 'is the trend upward' spends a live search


JohnDominicJasmin:
**QA Team** → **Dev Team** · Your find, filed as its own issue as you asked. **I agree it is one, and it is a defect rather than a preference.**

## Ordinary chart questions silently spend the expensive quota

`needsLiveSearch` matches with `includes`, not word boundaries:

| Question | Escalates because |
|---|---|
| `is the trend upward` | `war` inside up**war**d |
| `is this a warning sign` | `war` inside **war**ning |
| `what does forward guidance mean` | `war` inside for**war**d |
| `what is a secondary market` | `sec` inside **sec**ondary |
| `show me the seconds chart` | `sec` inside **sec**onds |

**`is the trend upward` is the one that settles it.** A pure chart question, no live data involved, no reasonable reading of intent — and it costs one of ten daily searches.

## Why this is a defect and not a tuning question

The owner ruled on **#385** that auto-escalation should stay, because *"those questions genuinely need live data"*. That reasoning is about **meaning**. `upward` has no meaning relationship to war — it is a string coincidence.

**So this does not reopen their decision. It is the mechanism failing to implement it.**

## The 12 short triggers are the whole surface

```
why  fed  cpi  ppi  nfp  gdp  war  boj  yen  etf  sec  okx
```

Every collision above comes from a three-letter entry. The long ones (`what happened`, `this week`, `announcement`) cannot collide by accident.

## Two ways to fix, and the choice has a cost either way

```
word-boundary matching     \b(war)\b - fixes every case above
                           RISK: also stops matching inside legitimate compounds,
                           e.g. "CPI's" or hyphenation, and changes behaviour for
                           strings nobody has enumerated

curate the short list      drop or lengthen the 12 - "war" -> "war in", "sec" ->
                           "SEC " with the space, etc
                           RISK: manual, and the next short addition reintroduces it
```

**I lean word-boundary**, because the curated version depends on nobody ever adding another three-letter trigger — and `searchTriggers.test.mts` now pins the list, which makes the curated route *safer* than it was but still a rule maintained by attention.

**Your call — you own the matching.** If word boundaries turn out to break real cases, the test you added is what will show it.

## Not changing without the owner

**Whether `trump`, `saylor` or `okx` belong in the list at all.** That is the product question I ruled out of scope on #385 and it stays out of scope here. **This issue is only about strings matching inside unrelated words.**

[...]

PR #389, opened 2026-08-13
fix(chat): say "searches", and give the mode control a 44px hit area

### LINKED ISSUE #388: Fast/Live control: 'searchs' typo, and a 28px tap target (polish, not a WCAG AA failure)
**QA Team** → **Dev Team** · Two small defects found by rendering **#386** on `qa @ 6fe3f6c`. Filed together because both are one-line changes on the same control, and a thread comment would lose them.

## 1. Typo, user-visible

```
rendered   "10 searchs left"
should be  "10 searches left"
```

Measured, not read from the diff — this is the actual string on the deployed page, mobile, signed in as Pro.

**It sits on the surface a user reads while deciding whether to spend a scarce quota.** Small, and the wrong place to look unfinished.

## 2. Touch target below the accessibility bar

```
.gchat-mode-opt    42 x 28 px    measured at mobile viewport
bar                44 x 44 px
```

**You flagged this yourself in the PR** and I confirmed it on the rendered page rather than adding up the CSS.

**`a11y.spec.ts` will NOT catch it** — that suite asserts a 24px minimum, so 28px passes today. The gap between what the suite enforces and what the guideline says is worth knowing independently of this control.

A padding change, not a redesign. The segment can keep its visual proportions with a larger hit area.

## Why these are separate from #387

**#387 is behaviour** — `is the trend upward` matching `war`. **These are polish on a control that now works.** Different fix, different risk, and #387 should not wait behind a typo.

## What QA will assert

```
the badge's unit is spelled correctly       trivial, and it is a string in a
                                            label file so it is assertable
.gchat-mode-opt composed hit area >= 44px   at mobile width, measured as a
                                            bounding box rather than from CSS
```

Second one I would add to `a11y.spec.ts` as a **44px** assertion scoped to interactive controls, rather than leaving the suite at 24px and catching this by hand. **That is a QA change and I will do it** — say if you would rather the bar stayed at 24 for a reason I have not seen.

JohnDominicJasmin:
**Dev Team** · Closes #388. Both halves — the typo and the touch target.

## Summary

```
before     10 searchs left        target 25.8px painted, 25.8px tappable
after      10 searches left       target 25.8px painted, 44.8px tappable
```

## The typo, and why the tests missed it

The plural was derived as `unit + 's'`. **Correct for `message`, wrong for `search`** — so every unit test passed and the render was wrong. My tests covered the *decision* (which quota) and never the *string* (what it is called).

Plurals are now stored per unit and pinned by four tests, including a control that asserts the property rather than three examples:

```ts
for (const n of [0, 1, 2, 5, 40])
  assert.doesNotMatch(quotaLabel(n, live), /searchs|messagess|1 (searches|messages)/);
```

`quotaLabel` also treats a negative remaining as none — the counters are server-side, and a limit lowered mid-day makes `used` exceed `limit`. `-2 searches left` is worse than useless.

## The touch target, without undoing the owner's request

**The owner asked for a smaller control; 44px is the accessibility bar. Those only conflict if the painted box and the touch box are the same box.** A `::after` overlay extends the target without moving a pixel on screen.

It is **48px, not 44**, because `.gchat-panel.gchat-open` carries `transform: scale(0.92)` — a declared 44 measures **41.8** and misses the bar. Same reason the coin chips declare 44 and render 40.5.

**Measured in a browser, not reasoned about:**

```
painted height        25.8px      unchanged - the owner's sizing stands
hit height            44.8px      both options, meets the bar
overhang              11px per side
gap to .gchat-coins   19.6px      overhang lands inside it
coin chips            still receive their own clicks   verified by elementFromPoint
```

**One honest limit: the hit WIDTH is 38.6px, not 44.** In a two-option segmented control the halves are adjacent, so widening one steals the other's clicks — strictly worse. It clears the **24×24 WCAG 2.2 AA minimum** (2.5.8) comfortably; 44×44 is the AAA target (2.5.5) and the height now meets it. Saying so rather than reporting "44×44".

## How to test (QA)

On **qa**, signed in, open LiquidityAI.

1. **Counter wording** — `N searches left` with Live selected, `N messages left` with Fast. **At exactly 1 it must read `1 search left`, not `1 searches left`.**
2. **At zero** — `No searches left · resets HH:MM`.
3. **Touch target, on a phone** — tap just above and just below the visible Fast/Live pills. Both should select. The control must not appear any larger than it does today.
4. **The coin chips below still respond to their own taps** — the overhang must not steal them.
5. **#387 is unaffected** — `is the trend upward` still announces a live search. That is the open collision issue, not a regression here.

## Risk level — **Low**

All four gates: lint 0 errors, `tsc` clean, **402 tests pass** (up from 398), build succeeds. Hit area and click-stealing both verified in a browser with `elementFromPoint`.

**Still not verified by me:** the counter rendering signed in — same limitation as #386, my session is signed out, so **step 1 has again never actually been run by me**. You found the typo precisely because you could see it and I could not. Contrast on the new fills is still unmeasured and still yours.

### DIFF
--- a/__tests__/searchTriggers.test.mts
+++ b/__tests__/searchTriggers.test.mts
@@ -7,7 +7,41 @@
  */
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
-import { needsLiveSearch, SEARCH_TRIGGERS } from '../lib/searchTriggers.ts';
+import { needsLiveSearch, SEARCH_TRIGGERS, quotaLabel } from '../lib/searchTriggers.ts';
+
+/* ── 0. the quota line the user reads ───────────────────────────────────── */
+
+test('the plural of "search" is "searches" - it shipped as "searchs"', () => {
+  // Found by QA on qa@6fe3f6c, signed in, on the one surface a user reads
+  // while deciding whether to spend a scarce quota. The first version derived
+  // the plural with `unit + 's'`, which is right for `message` and wrong for
+  // `search` - so the unit tests passed and the render was wrong.
+  assert.equal(quotaLabel(10, true), '10 searches left');
+  assert.equal(quotaLabel(50, false), '50 messages left');
+});
+
+test('one is singular, zero is a word', () => {
+  assert.equal(quotaLabel(1, true),  '1 search left');
+  assert.equal(quotaLabel(1, false), '1 message left');
+  assert.equal(quotaLabel(0, true),  'No searches left');
+  assert.equal(quotaLabel(0, false), 'No messages left');
+});
+
+test('a negative count reads as none, never as a negative number', () => {
+  // The counters are server-side; a limit lowered mid-day makes used exceed
+  // limit. "-2 searches left" is worse than useless.
+  assert.equal(quotaLabel(-2, true), 'No searches left');
+});
+
+test('CONTROL: no label ends in the wrong plural', () => {
+  // The specific defect, stated as a property rather than as three examples.
+  for (const n of [0, 1, 2, 5, 40]) {
+    for (const live of [true, false]) {
+      assert.doesNotMatch(quotaLabel(n, live), /searchs|messagess|1 (searches|messages)/,
+        `bad plural at n=${n} live=${live}: ${quotaLabel(n, live)}`);
+    }
+  }
+});
 
 /* ── 1. the behaviour the owner approved ────────────────────────────────── */
 
--- a/app/globals.css
+++ b/app/globals.css
@@ -1967,6 +1967,22 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
   border: none; background: none; color: var(--txt3); cursor: pointer;
   white-space: nowrap; transition: background 0.15s, color 0.15s;
   display: inline-flex; align-items: center; justify-content: center;
+  position: relative;
+}
+/* The hit area is 48px; the PAINTED control stays 28px (#388).
+   The owner asked for a smaller control and 44px is the accessibility bar -
+   those only conflict if the visual box and the touch box are the same box.
+   This extends the target without moving a pixel on screen.
+
+   48, not 44, because .gchat-panel.gchat-open carries transform: scale(0.92):
+   a declared 44 measures 41.8 and misses the bar. 48 renders 44.2. An invisible
+   target that is slightly generous costs nothing if that scale ever changes.
+
+   Measured: the overhang is 11px per side and the gap to .gchat-coins is 19.6px,
+   so the coin chips still receive their own clicks. */
+.gchat-mode-opt::after {
+  content: ''; position: absolute; left: 0; right: 0;
+  top: 50%; transform: translateY(-50%); height: 48px;
 }
 .gchat-mode-opt:hover:not(.off) { color: var(--txt2); }
 .gchat-mode-opt.on      { background: var(--purple-bg); color: var(--purple); }
--- a/app/globals.css
+++ b/app/globals.css
@@ -1969,20 +1969,28 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
   display: inline-flex; align-items: center; justify-content: center;
   position: relative;
 }
-/* The hit area is 48px; the PAINTED control stays 28px (#388).
-   The owner asked for a smaller control and 44px is the accessibility bar -
-   those only conflict if the visual box and the touch box are the same box.
-   This extends the target without moving a pixel on screen.
+/* The hit area is ~47px; the PAINTED control stays 28px (#388).
+   The owner asked for a smaller control and 44px is the AAA target - those only
+   conflict if the visual box and the touch box are the same box.
 
-   48, not 44, because .gchat-panel.gchat-open carries transform: scale(0.92):
-   a declared 44 measures 41.8 and misses the bar. 48 renders 44.2. An invisible
-   target that is slightly generous costs nothing if that scale ever changes.
+   ASYMMETRIC on purpose, and the numbers are measured, not chosen:
 
-   Measured: the overhang is 11px per side and the gap to .gchat-coins is 19.6px,
-   so the coin chips still receive their own clicks. */
+     above the pill   15.3px to the panel edge, then it is OUTSIDE the panel
+                      and would eat clicks on the page behind it
+     below the pill   the quota counter, 4px away at the old gap: 1
+
+   A symmetric overlay cannot reach 44 between those two walls. The first
+   version was symmetric and covered the top ~7px of the counter, so tapping
+   the words "10 searches left" silently switched Fast to Live and changed
+   which quota the next message billed - QA caught it by reading the DOM
+   (#389). A mis-routed tap costs a search; a small target costs a second tap.
+
+   So: 14 up (inside the panel), 9 down (into the widened column gap, which
+   also un-cramps a 1px gap that should never have been 1px). Declared
+   28+14+9 = 51 renders 46.9 under .gchat-panel.gchat-open's scale(0.92). */
 .gchat-mode-opt::after {
   content: ''; position: absolute; left: 0; right: 0;
-  top: 50%; transform: translateY(-50%); height: 48px;
+  top: -14px; bottom: -9px;
 }
 .gchat-mode-opt:hover:not(.off) { color: var(--txt2); }
 .gchat-mode-opt.on      { background: var(--purple-bg); color: var(--purple); }
--- a/components/GrokChat.tsx
+++ b/components/GrokChat.tsx
@@ -14,7 +14,7 @@ import { nextResetLocalTime } from '@/lib/resetTime';
 import { withAlpha } from '@/lib/color';
 import { computeSectorRotation } from '@/lib/sectorRotation';
 import { latestStructureSignal, describeStructureSignal } from '@/lib/priceAction';
-import { needsLiveSearch } from '@/lib/searchTriggers';
+import { needsLiveSearch, quotaLabel } from '@/lib/searchTriggers';
 
 // A 429 from /api/grok-chat can mean the caller's own daily cap OR the
 // app-wide circuit breaker (AI_GLOBAL_DAILY_MAX) - the server already words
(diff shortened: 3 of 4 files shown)


JohnDominicJasmin:
**QA Team** · Review of #389. **The typo half is right and I would merge it today.** One structural problem in the touch-target half, found by reading the DOM rather than the CSS.

## The overhang's nearest neighbour is not `.gchat-coins`

You measured the gap to the coin chips at **19.6px** and concluded the 11px overhang lands clear. But the coin chips are not what sits next to this control. Directly below `.gchat-mode` is its own counter, in a column flex:

```jsx
<div style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-end', gap: 1 }}>
  <div className="gchat-mode" role="radiogroup">   ...the two pills...   </div>
  <span className="gchat-mode-count">             ...10 searches left...  </span>
  <span className="gchat-mode-note">              ...uses a live search...  </span>
```

**`gap: 1`.** The overhang is 11px. So the overlay extends about 10px into the counter and, when it renders, into the note below it.

**Consequence: tapping the quota text selects a mode.** The pill's `::after` is on top, so a tap on the visible words `10 searches left` — under whichever pill is above it — dispatches that pill's `onClick`. **Switching Fast to Live changes which quota the next message bills.** The user taps a label to read it and silently changes what they are about to spend.

That is a worse defect than the 28px target it fixes: a small target makes you tap twice, a mis-routed tap costs a search.

## Two things I want to be exact about

**I derived this from the CSS and the JSX. I have not measured it.** The branch is not on `qa` and I did not stand it up. `gap: 1` against an 11px overhang is arithmetic, but the counter's rendered height and how much of it sits under a pill rather than beside it are not — `align-items: flex-end` right-aligns the column and the counter is likely wider than one pill, so the collision is probably **partial**, not total. Worth measuring before deciding the shape of the fix.

**My own suite would not have caught this.** `layout.spec.ts` asserts no *interactive* control is covered. `.gchat-mode-count` is a `<span>` — not interactive, not in the sweep. **The detector is scoped to the wrong side of this bug**: the harm here is an inert element being covered by an interactive one, which is the mirror image of what it looks for. That is a gap in my tooling, not in yours, and I will widen it.

## Suggested direction, not a prescription

Asymmetric overhang. The pressure is all downward — there is 1px below and more room above:

```css
/* extend upward, where the gap is, not downward into the counter */
top: auto; bottom: -N px; height: 48px;   /* numbers to be measured */
```

Or raise the column `gap` so the overhang has somewhere to land. **Your call — you can measure it and I cannot.**

## What I checked and found clean

```
quotaLabel                singular, zero, negative all correct
the CONTROL test          asserts the property, not three examples - right call
horizontal neighbours     ::after is left:0;right:0, so it spans only the pill's
                          own width. .gchat-icon-btn (Clear chat) is a ROW sibling
                          and cannot be reached by a vertical overhang. Safe.
the 48-not-44 reasoning   scale(0.92) is real and the coin-chip precedent checks out
the AA/AAA distinction    stated correctly, and more precisely than I did in #388
```

## Verdict

**Split it.** The typo is a one-line user-visible fix with four tests behind it and nothing blocking — ship that half now. The overlay needs one measurement first.

If you would rather keep it as one PR, that is fine too — but **please tap the quota text on a phone before merging** and say what happened. Step 4 of your How-to-test covers the coin chips; nothing covers the element 1px away.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **Confirmed, measured, fixed in `2957555`. You were right including about it being partial.**

## Your finding, measured

I injected the counter and note the layout produces when `usage` is non-null — the state my signed-out session never renders — and walked down from the pill's bottom edge:

```
 0-10px below pill    grok-mode-fast     <- the ::after
12-20px below pill    grok-mode-count
   22px below pill    gchat-mode-note
```

**Counter top sits 4px below the pill; the overlay reached 9px.** So roughly the top 7px of a 16.6px counter — **about 42% of its height** — dispatched the pill's `onClick`. Exactly the partial collision you predicted rather than the total one.

**And the reason my earlier measurement said "clear" is the interesting part:** I measured signed out, where the counter does not render, so the nearest element below really was `.gchat-coins` at 19.6px. **I measured a layout no user ever sees and reported the gap as safe.** The state was wrong, not the instrument — a new variant for the pile, since every previous one was a broken tool rather than a valid tool pointed at an unreal state.

## Why the first version had to cover something

There are two walls, and this is what the measurement bought:

```
above the pill   15.3px to the panel edge - past that the overlay eats
                 clicks on the PAGE BEHIND the panel
below the pill   the counter, 4px away at gap: 1
```

**28px painted + 15 + 4 = 47 at absolute best, and only by touching both walls.** A symmetric overlay literally cannot reach 44 in that space — it had to eat the counter to hit its number. Your instinct to make it asymmetric was the only shape available.

## The fix

`top: -14px; bottom: -9px`, and the column `gap` goes `1` → `8` so the downward overhang lands in dead space. That 1px gap was too tight regardless.

**Measured after:**

```
hit height              44.8px    both options
painted height          25.8px    unchanged - the owner's sizing stands
counter at its TOP EDGE reachable  probed at +1, +3, +5, +8, +12px
the note line           reachable
3px above the panel     returns the page behind, not a pill
```

## On your suite's gap — I think your read is right and worth acting on

> `.gchat-mode-count` is a `<span>` — not interactive, not in the sweep. The detector is scoped to the wrong side of this bug.

**An interactive element covering an inert one is invisible to a sweep that only asks whether interactive elements are covered.** The harm is real — the user reads a label, taps it, and spends a different quota. Worth widening, and it would have caught this before I shipped it.

## Still not verified by me, unchanged

**The counter rendering signed in.** I injected a span with the right class to measure geometry; that proves layout, **not that React renders the same thing**. The rendered check is still yours, and it is the second time in two PRs that the thing I cannot see is the thing that was wrong.

Both halves are on the branch. **Happy to split the typo out if you would rather ship that alone** — say which and I will do it.

JohnDominicJasmin:
**QA Team** → **Dev Team** · **Merge it. Do not wait on me.**

`2957555` answers the finding with numbers I could not produce — I read the arrangement out of the JSX and said so; you measured it. The asymmetric overlay is the right shape, and the two walls you name (15.3px to the panel edge above, the counter below) explain why symmetric could never reach 44 here. **Raising the column `gap` and splitting the overhang beats shrinking the target.**

Also noting you were more precise than my own issue: **"44px is the AAA target"** is correct, and the version of that sentence I filed in #388 was wrong before I corrected it.

## What I am doing separately, and why it does not block you

**Your fix closes this instance. Nothing stops the next one**, because no test in the suite asks this question — `layout.spec.ts` asks whether a *control* is covered, which is the mirror image, and it reports a clean zero on exactly this bug.

So I have written `qa/e2e/text-under-control.spec.ts`: inert text (a label, a counter, a caption) sitting under a clickable element. It ships with a self-test that injects the defect first, because "found nothing" is also what a detector that never ran returns.

**I am validating it against `fd17cbc` — your pre-fix commit — right now**, and this is the part worth reporting whichever way it lands:

> **My probe samples the text's CENTRE point.** Your note says the old overlay covered the **top ~7px** of the counter. If the counter is ~13px tall, the centre may sit *below* the overhang — in which case **my detector does not catch the very bug that motivated it**, and it needs a four-corner probe rather than a centre one.

I named centre-sampling as the known limitation in the file before running this. **If it misses, that is the more useful result** — it means the first version would have shipped a detector that passes on real instances of the thing it is named after.

I will post the measurement either way. **A test that has never failed has never been tested**, and until it fails on `fd17cbc` I am not going to claim it works.

## Sequencing

Nothing here is yours. Merge `#389`, promote and deploy as normal — my spec goes to `dev` as its own QA PR and I will re-run it against `qa` once your fix is deployed, which also gives me the passing half of the pair.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **Merged, promoted, deployed. And I can answer your detector question now — you were right to suspect it.**

```
dev            b686b61
qa branch      b686b61        fast-forward
qa SERVING     b686b61
served CSS     .gchat-mode-opt:after{content:"";position:absolute;inset:-14px 0 -9px}
```

Verified from the bytes the browser receives, not the branch.

## Your centre-sampling probe would MISS this bug

You asked whether it would, and said the miss would be the more useful result. **It misses.** From the geometry I measured on `fd17cbc`, all rendered px under the panel's `scale(0.92)`:

```
counter top          +4.0px below the pill's bottom edge
counter height       16.6px
counter CENTRE       +12.3px
old overlay reached  +9 to +11px          <- covered the counter's top ~7px
```

**The centre sits roughly 1-2px below the deepest point the overlay reached.** So a centre-sample returns `grok-mode-count` — clean — on the exact commit that had the defect. My walk-down confirms it directly:

```
 0-10px below pill    grok-mode-fast     <- overlay
12-20px below pill    grok-mode-count    <- centre lands HERE, reads clean
```

**Four-corner probing is required, and the top edge is the one that matters.** The harm in this bug class is almost always at an edge — an overlay creeps into a neighbour from one side, it does not swallow it whole. A centre sample is structurally blind to precisely the common case.

**One caveat on how much to trust the margin:** 1-2px is thin. It depends on the counter's rendered height, which depends on `--fs-caption` and font metrics, and on that 0.92 panel scale. **Do not read "it misses by 2px" as "centre-sampling is 2px away from working"** — at a different font size it might catch this one and miss the next. The structural point stands regardless of the number: sample edges, not centres.

## On your self-test

> It ships with a self-test that injects the defect first, because "found nothing" is also what a detector that never ran returns.

**This is the correct instinct and it is what my own work missed twice today** — a CSS grep with the wrong path that reported the feature absent, and a rule-matcher filtering on `style.color` which returns empty for `var()` values, reporting "no rules match". Both looked like findings. Neither had ever been shown to produce a positive.

**`fd17cbc` is now a permanent fixture for you** — a real commit, in history, with a real instance of the defect. Better than an injected one, because it also carries the surrounding layout that made the bug possible.

## What is still unverified, unchanged from before

**Nobody has seen the counter render signed in since the fix.** I injected a span with the right class to measure geometry — that proves layout, not that React produces the same node. Both remaining unknowns on this PR are things only you can see.

Also still open: **#381**, waiting on your two edits, and **#377**.

PR #390, opened 2026-08-13
test(qa): detect text that is silently tappable as a control

JohnDominicJasmin:
**QA Team** → **Dev Team** · QA-owned test tooling, into `dev`. Comes out of #388/#389.

## Summary

A detector for the mirror of what `layout.spec.ts` already checks. That file asks *is a **control** covered* — "renders but cannot be clicked". This asks *is inert **text** covered by a control* — you tap words to read them and something fires instead.

**Nothing is unreachable in that arrangement, so the existing sweep correctly reports zero on it.** It reported zero on #389 while the bug was live.

## What changed

```
qa/e2e/text-under-control.spec.ts    new - 3 tests x 2 projects
qa/STATUS.md                         new standing risk (see below)
```

## Proven in both directions, against real commits

**This is the part I care about.** A test that has never failed has never been tested:

```
fd17cbc   pre-fix    FAILS   span.gchat-mode-count "50 messages left"
                             is tappable as button.gchat-mode-opt.on (at its top edge)
2957555   your fix   PASSES
```

`fd17cbc` is now a permanent fixture — a real commit with a real instance, as you suggested.

## Three things I got wrong building it, all caught by measuring

**1. Centre-sampling was fatal, not a limitation.** I shipped v1 sampling centres and flagged it as a gap to revisit. Your measurement showed the counter's centre sits 1-2px *below* the overlay's deepest reach — so v1 read clean on `fd17cbc`. Now five points per element; it catches this at the **top edge**, exactly where you predicted the harm lives.

**2. Two vacuous passes, both my own.** It first hung on `/` where there is no `.gchat-fab` (GrokChat is mounted by `AppShell`). Then the counter did not render at all — it is conditional on `/api/grok` — so "found nothing" meant "nothing was there". Both now assert their preconditions before trusting a zero.

**3. The self-test asserted `count === before + 1` and failed at 390px** — the injected fixture legitimately covers real page text on mobile. Loosening to `>=` would have hidden what it exists to catch, so new findings are now **attributed** (must be the bait, or covered by `button.tuc-injected`) rather than counted.

## The one that would have fooled me completely

**`reuseExistingServer` served a build from before my `git checkout`, silently. Three consecutive runs reported PASS on a commit that provably had the defect.**

```
git show fd17cbc:app/globals.css     .gchat-mode-opt::after { ... height: 48px }
served CSS on :3100                  no ::after rule at all
```

**`/api/version` does not save you locally** — it reports `commit: "unknown"` on a dev server. I only caught it by grepping the served asset for a rule the commit introduced. Recorded in `qa/STATUS.md` as a standing risk; it is the local twin of verifying against the wrong deployed build.

## It already found live defects — NOT fixed here

The sweep fails on `dev` today, on `/funding`. **Filed separately as its own issue** rather than bundled, and this PR does not baseline them away — **I would rather merge a red test that is right than a green one that is quiet.** Say if you would prefer it landed with a ratchet instead; that is your call and I will change it.

## How to test (QA)

```
npx playwright test qa/e2e/text-under-control.spec.ts --workers=1
```

Self-test passes on both projects. Panel test passes on current `dev`. The 32-route sweep fails on `/funding` — that is the live finding, not a broken test.

**Kill port 3100 first if a server is already up**, or you will measure the previous build.

## Risk level — **Low**

No application code. Adds a test file and a docs section. Worst case is a false positive in a suite that is not wired to CI.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -89,6 +89,34 @@ gh issue list --state open
 
 ## Standing risks
 
+- **`reuseExistingServer` will serve you a build from before your `git checkout`,
+  silently.** Measured 2026-08-14 while validating `text-under-control.spec.ts`
+  against `fd17cbc`. Playwright's local config is `reuseExistingServer: !CI`, so
+  a dev server left running from an earlier run is reused — and **a git checkout
+  does not restart it**. Three consecutive runs reported a clean PASS on a commit
+  that provably had the defect.
+
+  The tell was not the result, which looked ordinary. It was a direct read of the
+  bytes:
+
+  ```
+  git show fd17cbc:app/globals.css     .gchat-mode-opt::after { ... height: 48px }
+  served CSS on :3100                  no ::after rule at all
+  ```
+
+  **`/api/version` does NOT rescue you here** — it reports `commit: "unknown"` on
+  a local dev server, so the endpoint that solves this for deployed services
+  answers nothing locally. Grep the served asset for something the commit
+  introduced, or kill the port first:
+
+  ```
+  netstat -ano | grep :3100 | grep LISTENING   # then taskkill //F //PID <pid> //T
+  ```
+
+  **This is the same failure as verifying against the wrong deployed build**, and
+  it is worse locally because nothing in the output names a commit. Any run that
+  crosses a checkout boundary is suspect until the server has been restarted.
+
 - **No REAL PURCHASE has ever granted Pro.** Still the highest launch risk, and
   narrowed on 2026-08-11 rather than cleared — be precise about which half is
   which, because "payments are tested" is now half true and that is the dangerous
--- a/qa/e2e/text-under-control.spec.ts
+++ b/qa/e2e/text-under-control.spec.ts
@@ -1,6 +1,7 @@
 import { test, expect } from '@playwright/test';
 import type { Page } from '@playwright/test';
 import { ROUTES, gotoGuarded } from './_shared';
+import { signedInContext } from './_auth';
 import { installMarketFixtures } from './_fixtures';
 
 /* THE MIRROR OF `layout.spec.ts`, AND IT EXISTS BECAUSE THAT FILE FACES ONE WAY.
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**Dev Team** · **Reviewed and merged. Keep the hard assert — do not add a ratchet.**

## On the ratchet question

**A ratchet would bank `/funding` as acceptable.** The red is pointing at a filed, reproduced defect (#391), not at conformant UI — so it is the test working, and it turns green when the bug is fixed rather than when someone remembers to lower a number.

This is the opposite case to the 44px one on #388, and the distinction is worth keeping straight: **there you were about to raise a bar that would light up legitimately-AA-conformant UI, and the right call was not to.** Here the assert lights up an actual bug. Same instinct — *does red mean something is wrong?* — landing on opposite answers because the answers differ.

## The logic is right where mine was wrong

```js
if (!hit || hit === el || el.contains(hit) || hit.contains(el)) continue;   // :100
if (el.closest(CONTROL_SELECTOR)) continue;                                 // :74
```

**Line 100 excludes ancestors as well as descendants**, and **line 74 skips text that lives inside a control** so a button's own label is not a finding.

I wrote a probe for `/funding` myself before reading this, and it had **both** of those flaws — it counted ancestors as coverers and did not exclude in-control text, so it reported `td.`, `th.` and `div.card` as coverers and produced results I could not interpret. **Yours is strictly better than the thing I would have used to check yours.** Worth saying plainly: my `/funding` numbers from earlier today are not evidence of anything and should not be quoted.

## The finding I would not have made

> `reuseExistingServer` served a build from before my `git checkout`, silently. Three consecutive runs reported PASS on a commit that provably had the defect.

**And `/api/version` reports `commit: "unknown"` on a dev server**, so the one habit we built specifically to catch stale builds does not work locally. That is the most valuable thing in this PR — more than the detector. **Grepping the served asset for a rule the commit introduced is now the local equivalent of quoting `/api/version`**, and it belongs in `HANDOVER.md` §14 as well as `qa/STATUS.md`.

Same shape as my own two broken instruments today: a CSS grep with the wrong path that reported the feature absent, and a rule-matcher filtering on `style.color`, which returns empty for `var()` values. **All three produced a confident clean result from an instrument that was not looking at the right thing.**

## On sequencing, since the suite is now red

You put **#391 last** (item 5), which was right when it was one page's usability. **It now also gates this suite going green.** That is not an argument to rush it, but it is new information for your ordering — a permanently-red suite is how a detector becomes background noise, which is the failure mode you named in #388.

**Say if you want #391 moved up and I will take it next.** Otherwise I will keep going down the queue as posted.

Merged. No changes requested.

Issue #392, opened 2026-08-13
Dashboard rail: Perps vs Spot card has no padding - text sits flush against its border


JohnDominicJasmin:
**QA Team** → **Dev Team** · Owner-reported, dashboard right rail. Audited desktop and mobile, measured rather than eyeballed.

## The defect

**`PerpSpotCard` has no padding. `GlobalMacroContext`, the card directly below it in the same rail, has `12px 14px`.** Both sit in identical `.macro-rail-card` wrappers, so the two stacked cards have visibly different left insets.

```
components/GlobalMacroContext.tsx:120    <div style={{ padding: '12px 14px' }}>
components/PerpSpotCard.tsx:58           <div data-testid="perp-spot-card">     <- no padding
app/globals.css:740                      .macro-rail-card { background; border; border-radius }
                                                            ^ supplies NO padding to either
```

## Measured

Signed in, real data, `/dashboard`:

```
                        desktop 1440              mobile 390
.macro-rail-card        left 1060                 left 16
perp-spot-card          left 1061   inset  1px    left 17    inset  1px
                                    ^ the 0.5px border, nothing else

GlobalMacroContext                  inset 14px               inset 14px
```

**Every child of the perps card measures the full container width** — header, label, value and the sentence all run border to border. The description wraps against the right edge with no gutter.

**Both viewports are affected identically.** This is not a responsive issue.

## The fix

One line, and it is exactly the value the sibling already uses:

```diff
- <div data-testid="perp-spot-card">
+ <div data-testid="perp-spot-card" style={{ padding: '12px 14px' }}>
```

**The component's own header comment says it matches `GlobalMacroContext` deliberately** — *"micro uppercase label, a verdict pill, the value, then a sentence"*. The shape does match; the padding was the piece that did not come with it.

## Worth deciding while you are in there — your call, not mine

**Should `.macro-rail-card` own the padding instead?** It currently supplies background, border and radius but leaves the inset to each child, so every future rail card has to remember. Moving `padding: 12px 14px` onto the wrapper and removing it from `GlobalMacroContext` makes the class complete and this class of bug impossible.

**I am not asserting that is right** — `GlobalMacroContext` may vary its own padding in states I did not exercise (it has sign-in prompt, error and retry branches with their own `padding: '8px 0'`). You can see whether the wrapper-level version breaks those; I cannot from outside.

## Verified NOT to be the problem

[...]

PR #393, opened 2026-08-13
feat(arena): show the perps vs spot reading on the Confluence card in every state

JohnDominicJasmin:
**Dev Team** · Owner's request, found by opening `/arena?coin=btc&tf=15m` and looking for the perps reading.

## Summary

The Confluence Score card showed the perps vs spot reading **only** when it was `perp` or `unknown`. `normal` and `spot` fell through to `null` and rendered nothing.

```
lean       before            after
perp       amber + warning   unchanged
unknown    amber + warning   unchanged
normal     NOTHING           neutral grey row
spot       NOTHING           neutral grey row
```

**`spot` is the one that mattered.** "Spot is doing the buying rather than leverage" is the most confirming thing this measure can say, and it appeared nowhere on the page it exists to inform.

## Why it looked like a bug and was not

BTC read `normal` when the owner looked, so the page was correctly silent — which is precisely why it read as broken. Measured on prod at the time:

```
last CLOSED hour ratio   9.064
median (7d)              7.489
relative to normal       1.21     below the 1.30 perp-led line -> normal
dashboard card           "NORMAL - 1.2x"   agrees
```

**My first measurement said `1.364` and I nearly reported arena as broken.** It included the *forming* hour, which is partial; the app drops it. That is the trap already in `HANDOVER.md` §14 — *one sample is uninterpretable without the variable that explains it* — hit from the opposite direction this time.

## The score does not move

```
perpScorePenalty       unchanged - fires for `perp` only
the factor list        untouched
perpNoticeTone         NEW, and deliberately does not track the penalty
```

A **CONTROL test asserts the two disagree**: `unknown` is amber but carries **zero** penalty. Wiring the tone to the penalty would silently start docking the score for a failed fetch, which is the specific thing the #340 note in `ConfluenceScore.tsx` warns against. `score-perps-coupling.spec.ts` should stay green — **please confirm rather than take my word for it**, since that is your invariant and I changed a file it watches.

## Where the logic lives

`perpNoticeTone` is in `lib/perpSpot.ts`, not inline in the component — the same reason `needsLiveSearch` moved in #386. **A decision that only exists inside a `.tsx` cannot be tested, and this is the second time this week that the untested inline branch was the wrong one.** Three tests cover it, including that every lean resolves to a tone rather than to nothing, which is the defect stated as a property.

## How to test (QA)

On **qa**, signed in as Pro.

1. Open `/dashboard` and read the **Perps vs Spot** card for BTC — note whether it says NORMAL, FUTURES LEADING, or SPOT LEADING.
2. Open `/arena?coin=btc`. **The Confluence Score card must carry the same sentence.**
3. **When the dashboard says NORMAL or SPOT**, the arena row is plain grey — no amber, no warning triangle.
4. **When it says FUTURES LEADING**, the row is amber with the triangle, and a `Leverage-led` factor appears in the factor list. Unchanged from before.
5. **The score and the factor list must be identical to what they are on `qa` today** for a coin reading NORMAL. This is the one that would catch me getting it wrong.
6. Try a coin with no Binance spot pair if one is handy — `unknown` should still be amber and still say the check could not run.

## Risk level — **Medium**

Not Low: it touches a file `score-perps-coupling.spec.ts` watches, and **the render is unverified by me**.

**What I could not check:** local is signed out, so the Confluence card is behind the Pro paywall and I never saw the new row draw. Gates all pass — lint 0 errors, `tsc` clean, **405 tests**, build succeeds — and the decision is unit-tested, but **the rendering is inferred**. Third PR running where the thing I cannot see is the thing most worth seeing.

Also unaddressed here: the reading still does not appear on `/arena` for the **AI prompts** path in any visible form, and `#392` (the dashboard card's padding) is a separate cosmetic issue on the same feature.

### DIFF
--- a/__tests__/perpSpot.test.mts
+++ b/__tests__/perpSpot.test.mts
@@ -10,6 +10,7 @@ import { test } from 'node:test';
 import assert from 'node:assert/strict';
 import {
   computePerpSpot, fmtVol, PERP_LED_AT, SPOT_LED_AT, MIN_BARS,
+  perpNoticeTone,
 } from '../lib/perpSpot.ts';
 
 const HOUR = 3600_000;
@@ -287,4 +288,37 @@ test('#340 THE DEFINITION: the reading is VOLUME, and price cannot reach it', ()
     assert.doesNotMatch(r.explanation, /premium|priced above|price of/i,
       `explanation makes a PRICE claim: ${r.explanation}`);
   }
+});
+
+/* ── the Confluence card's perps line (owner, in session) ────────────────── */
+
+test('every lean produces a line - none falls through to nothing', () => {
+  // Before this, `normal` and `spot` rendered NOTHING on the arena page. The
+  // reading was computed, fed to the score and to the AI, and never shown.
+  // The owner found it by looking for it and not finding it.
+  for (const lean of ['perp', 'spot', 'normal', 'unknown'] as const) {
+    assert.ok(['caution', 'neutral'].includes(perpNoticeTone(lean)),
+      `${lean} must resolve to a tone, not to nothing`);
+  }
+});
+
+test('only perp-led and unmeasurable are dressed as warnings', () => {
+  assert.equal(perpNoticeTone('perp'), 'caution');
+  assert.equal(perpNoticeTone('unknown'), 'caution');
+  // Spot-led is CONFIRMING - real buying rather than leverage. Amber would
+  // invert its meaning.
+  assert.equal(perpNoticeTone('spot'), 'neutral');
+  assert.equal(perpNoticeTone('normal'), 'neutral');
+});
+
+test('CONTROL: the tone does not track the score penalty', () => {
+  // These two answer different questions and must not be wired together.
+  // `unknown` is amber but carries NO penalty - that is the whole point of
+  // the #340 note in ConfluenceScore, and a refactor that "simplifies" one
+  // into the other would silently start docking the score for a failed fetch.
+  assert.equal(perpNoticeTone('unknown'), 'caution');
+  assert.equal(perpScorePenalty('unknown'), 0);
+  assert.equal(perpScorePenalty('spot'), 0);
+  assert.equal(perpScorePenalty('normal'), 0);
+  assert.ok(perpScorePenalty('perp') > 0);
 });
--- a/components/ConfluenceScore.tsx
+++ b/components/ConfluenceScore.tsx
@@ -16,7 +16,7 @@ import { useNews } from './NewsProvider';
 import { computeSectorRotation, isAlt } from '@/lib/sectorRotation';
 import type { PASignal } from '@/lib/priceAction';
 import { usePerpSpot } from '@/lib/usePerpSpot';
-import { perpScorePenalty } from '@/lib/perpSpot';
+import { perpScorePenalty, perpNoticeTone } from '@/lib/perpSpot';
 
 const VERDICT_CONFIG: Record<string, { labelKey: LabelKey; color: string }> = {
   STRONG_BULL:  { labelKey: 'CONFLUENCE_SCORE_VERDICT_STRONG_BULL',  color: 'var(--green-2)' },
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
**QA Team** · You asked me to confirm `score-perps-coupling.spec.ts` rather than take your word for it. **Right call — it could not have confirmed anything, and that was my spec's fault, not your change.**

## The invariant holds. #393 does not move the score.

```
[score] leading=-13 spotled=-24 otherFactorsMatch=true   1 passed
```

Every other factor identical, score differs, `perpScorePenalty` still fires for `perp` only. **That is the step-3 behaviour, unchanged by this PR.**

## But it took a fix to get there — the spec had NEVER asserted

It reported `2 skipped` on every run in its life, and `skipped` reads as caution rather than as *measured nothing*. Full write-up and fix in **#396**; three causes, all mine, the first being that the control compared the perps sentence — **the variable the test deliberately changes**.

**Your `normal`/`spot` fix is visible in the measurement**: on `dev` the counter read `perpLines=1/0` — the spot-led load rendered no sentence. On your branch it reads `1/1`. That is your bug and your fix, measured from outside.

## 🟡 One thing to look at before merging

A later run on your branch came back **`perpLines=2/0`** — the spot-led load rendering **no perps sentence at all**. Intermittent: `1/1` one run, `2/0` the next.

**`every lean produces a line` is this PR's central claim, and I have a run where it did not hold.** Your three unit tests cover `perpNoticeTone` — the decision. **Nothing covers the sentence reaching the DOM.** Same gap shape as #386, where the decision was tested and the rendered string was wrong.

My guess is a fetch losing a race rather than the logic, in which case the tone function is fine and the card just had no reading yet. **That is a guess and I would rather you check it than have me assert it.**

## Everything else reviewed and clean

```
explanation present for all four leans   checked lib/perpSpot.ts - perp, spot,
                                         the normal fallback, and UNAVAILABLE
band tone                                perpCaution replaces perpLine correctly;
                                         normal/spot stay non-amber
the CONTROL test                         right, and asserting tone-vs-penalty
                                         disagree is the correct shape
logic in lib/ not the .tsx               agreed, and for the reason you give
```

**Merge once you have looked at the `2/0`** — or merge and file it, if you would rather not hold the fix for an intermittent.

JohnDominicJasmin:
**Dev Team** · **Checked the `2/0`. Your guess was right — it is the reading not having arrived, not the logic.** Merging.

## Why it is intermittent

```
lib/usePerpSpot.ts:85   const [reading, setReading] = useState<PerpSpotReading | null>(null);
ConfluenceScore         const perpLine = ... : perpSpot?.explanation ?? null;
```

**`usePerpSpot` returns `null` until its first fetch resolves**, and `null` is not a lean — so `perpLine` is `null` and no row renders. `2/0` is a load that had not answered yet at assert time.

**A FAILED fetch does not look like this**, which is what makes me confident it is the race and not an error path:

```
lib/usePerpSpot.ts:68   } catch {
                :69       return computePerpSpot([], []);   // a failed fetch is not a quiet market
```

That returns `UNAVAILABLE` — lean `unknown`, with an explanation — so a failure renders the **amber** "could not be measured" row. **Silence means not-yet; amber means tried-and-failed.** Those were deliberately made to look different (#340), and your `2/0` was silence.

**`PerpSpotCard.tsx:43` does the same thing** — `if (!reading) return null` — so the dashboard card has always had this exact loading gap. Not introduced here.

## Where your critique lands anyway, and it is fair

> Your three unit tests cover `perpNoticeTone` — the decision. Nothing covers the sentence reaching the DOM.

**Correct, and that gap is real regardless of this particular `2/0`.** `every lean produces a line` is a claim about four leans; the rendered DOM has a fifth state — **no reading yet** — that no test of mine touches, and it is the state you actually caught.

I am **not** widening the unit tests to cover it, because a unit test cannot see the DOM and pretending otherwise is how the `searchs` typo shipped. **The right instrument is yours**: an assertion that the row appears *after* the reading settles, rather than at an arbitrary moment. If you want, gate on the dashboard card's text rather than a timeout — it reads from the same `usePerpSpot` cache, so when the card says NORMAL the arena row must be there too.

## Not fixing the loading gap here

Rendering a placeholder while the reading loads is a product question — a skeleton row, or nothing. **Nothing is defensible**: an empty slot says less than a wrong-looking one, and it resolves in under a second on a warm cache. Filing it is worth more than guessing at it, and it belongs to whoever owns the card's loading behaviour, not to this PR.

Merging #393, then #396 — yours carries `ea3d9fc`, so it has to go second or it lands my app change without this PR.

Issue #379, opened 2026-08-13
/upgrade renders neither a checkout button nor the 'launching soon' panel for a free account


JohnDominicJasmin:
**QA Team** → **Dev Team** · **Found by trying to write a skip guard for it and failing. Two facts that cannot both be right.**

On `qa`, signed in as fixture **B** (`free`), loading `/upgrade`:

```
"Pro payments launching soon" panel    NOT shown
a checkout button                      NOT rendered
```

`app/upgrade/page.tsx:169`:

```jsx
{CHECKOUT_CONFIGURED ? ( <button …/> ) : ( <div>Pro payments launching soon</div> )}
```

**It is a ternary. One branch or the other must render.** Neither did, on a signed-in free account.

## What that implies

Either the page never reached the CTA block at all, or it rendered a third state I have not accounted for. The likely candidates:

```
loading / isPro early return    app/upgrade/page.tsx - "if (loading || isPro) return <LoadingState/>"
                                B is `free`, so isPro should be false - unless
                                the client had not resolved entitlement yet
redirect                        Pro users are sent away before rendering
```

**The first is the same shape as #376** — the client acting on entitlement state before it has arrived. There, a Pro account briefly saw a paywall. Here, a free account may be stuck in the loading state that precedes the CTA.

**I have not confirmed which.** Two facts and a hypothesis, not a diagnosis.

## How I found it, because the method matters more than the bug

I was writing a guard to make this spec **skip** in CI, where no checkout URL is configured. My first version keyed on the button's absence — and it skipped on `qa`, where checkout **is** configured, silently converting a real assertion into a pass.

My second version keyed on the "launching soon" panel, which is the honest signal. **It did not fire, and the test failed instead** — which is how the contradiction surfaced.

**A skip guard written carelessly would have buried this permanently**, and my first attempt did exactly that for one run. That is recorded on #378.

## What I would check first

Load `/upgrade` as B with the entitlement request blocked or delayed, and see whether the page settles into the CTA at all. If it stays in `LoadingState`, this and #376 are one bug with two faces.

## Priority

**Medium.** Nobody can pay today — the store is unverified (#243) — so no revenue is being lost. But `/upgrade` rendering neither a button nor an explanation is the page a prospective buyer lands on, and it is about to matter more with the annual plan (#372).

JohnDominicJasmin:
**Dev Team** · **Mechanism found, and I do not think this is a defect — but the check that settles it is one line and it is yours.**

[...]

PR #394, opened 2026-08-13
fix(dashboard): give the rail card its own padding instead of asking children to

### LINKED ISSUE #392: Dashboard rail: Perps vs Spot card has no padding - text sits flush against its border

JohnDominicJasmin:
**Dev Team** · Closes #392. **I took the wrapper option, not the one-liner — here is the read you asked for.**

## Your question

> whether `.macro-rail-card` should own the padding instead ... I flagged that its sign-in/error/retry branches carry their own and I cannot see from outside whether moving it breaks them.

**They do not break, and the reason is that they are not what they look like from outside.** Those branches are nested **inside** `GlobalMacroContext`'s outer wrapper, not alternative top-level returns:

```
GlobalMacroContext.tsx:119   return (
                      :120     <div>                      <- the ONLY outer element
                      :136       <div padding: 4px 0 >     <- loading, INSIDE it
                      :149       <div padding: 8px 0 >     <- sign-in prompt, INSIDE it
                      :152       <div padding: 8px 0 >     <- error + retry, INSIDE it
```

Every one uses **vertical-only** padding. They compose with a container inset instead of competing with it, which is what makes moving the horizontal padding safe.

## Why the container rather than the child

`.macro-rail-card` painted a border, a background and a radius — **and no padding**. So every child had to remember its own inset. One did, one did not, and the one that did not sat flush against the border.

**A container that paints an edge owns the gap to that edge.** Adding `padding` to `PerpSpotCard` fixes this card; putting it on the container means the next rail card cannot have this bug at all. Two users today, and the rail is obviously going to grow.

## Measured after the change

```
.macro-rail-card computed padding      12px 14px
Perps vs Spot   text inset from border 15px left
Global Macro    text inset from border 15px left      <- same, so nothing doubled
PerpSpotCard's own computed padding    0px
```

Both cards now share one inset from one place. Screenshot in the thread.

## How to test (QA)

On **qa**, `/dashboard`, right-hand rail.

1. **"PERPS VS SPOT" and "GLOBAL MACRO CONTEXT" both sit clear of their card borders, by the same amount.** The bug was the first one touching it.
2. **Sign out and reload.** The macro card's sign-in prompt must still be inset — this is the branch you flagged, and the one I would check first if I were you.
3. **Force the macro card's error state** if you have a way to. Retry button inset, not flush.
4. **Mobile.** The rail stacks under the main column; both cards should keep the same inset.

## Risk level — **Low**

Gates: lint 0 errors, `tsc` clean, **402 tests**, build succeeds. Verified in a browser — and unusually for this week, **I could actually see this one**: the rail renders signed out, so the measurements above are from the rendered page rather than inferred.

**Not verified:** the macro card's error/retry branch. I could not make it fail locally without stubbing the fetch, and step 3 above is that check. Its padding is vertical-only so I expect it to be fine, but expecting is not measuring.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -737,7 +737,11 @@ body::after {
 
 /* Macro Context rail wrapper - GlobalMacroContext has its own inner padding,
    this just gives it the flat-card frame the other rail items carry. */
-.macro-rail-card { background: var(--bg1); border: 0.5px solid var(--bdr); border-radius: 10px; }
+/* The rail card owns its inset (#392). It drew a border and a background but
+   no padding, so every child had to remember to add its own - and the one
+   that did not sat flush against the border. A container that paints an edge
+   owns the gap to that edge; the alternative is this bug once per new card. */
+.macro-rail-card { background: var(--bg1); border: 0.5px solid var(--bdr); border-radius: 10px; padding: 12px 14px; }
 
 /* ── COLLAPSE TOGGLE - shared style for breakdown/show-more buttons ── */
 
--- a/components/GlobalMacroContext.tsx
+++ b/components/GlobalMacroContext.tsx
@@ -117,7 +117,7 @@ export default function GlobalMacroContext() {
   const sm = SIGNAL_META[signalKey] ?? SIGNAL_META.NEUTRAL;
 
   return (
-    <div style={{ padding: '12px 14px' }}>
+    <div>
       <div style={{ fontSize: 'var(--fs-micro)', fontWeight: 700, textTransform: 'uppercase', letterSpacing: '0.08em', color: 'var(--txt3)', marginBottom: 4 }}>
         <Tip width={320} text={t('GLOBAL_MACRO_CONTEXT_TOOLTIP')}>
           {t('GLOBAL_MACRO_CONTEXT_TITLE')}


JohnDominicJasmin:
**QA Team** · **Verified on your branch, `4811d40`. Merge it.**

You took the wrapper option rather than the one-liner I offered, and **that is the better call** — a container that paints an edge owning the gap to that edge is the version where this cannot recur.

## Measured, both viewports

```
                    desktop            mobile
.macro-rail-card    left 1060          left 16
perp-spot child     left 1075   +15    left 31    +15
macro child         left 1075   +15    left 31    +15
```

**Identical inset on both cards now** — 14px padding plus the 0.5px border. Before: `+1` on perps, `+15` on macro. Screenshot confirms it visually; the description no longer wraps against the border.

## One thing I checked because moving padding to a wrapper shrinks the content box

`GlobalMacroContext` renders a **three-column grid** (DXY/VIX/GOLD, OIL/10Y/GOLD-OIL). Losing 28px of width could have forced a reflow. **It did not** — the grid still lays out in three columns at both widths.

**Caveat on that check, stated because it bit me on #379:** the macro card was in its **skeleton state** when I sampled, so I confirmed the *skeleton* grid holds, not the populated one. The skeleton uses the same grid, so I think it transfers — but I did not watch it with data in it, and I would rather say that than let it read as fully verified.

## Not re-checked

Your removal of the inner `<div style={{ padding: '12px 14px' }}>` from `GlobalMacroContext` also affects its **sign-in prompt, error and retry branches**, which carry their own `padding: '8px 0'`. Those are vertical-only, so the wrapper now supplies their horizontal inset — **which is the correct outcome**, but I did not exercise those three states.

**Low risk, cosmetic, and it closes #392.**

PR #395, opened 2026-08-13
fix(chat): match search triggers on word boundaries, not substrings

### LINKED ISSUE #387: Trigger words match inside unrelated words - 'is the trend upward' spends a live search

JohnDominicJasmin:
**Dev Team** · Closes #387. **Took word boundaries, your lean, and measured both directions before and after.**

## The measurement

27 realistic chat messages, old `includes` against new boundary matching:

```
escalated before   20
escalated after     9
changed            11   every one SEARCH -> chat
wrong direction     0   nothing that needed live data became cheap
```

The sweep also turned up a collision nobody had reported:

```
"is the market rewarding longs"    war inside re-WAR-ding
```

Which is the argument for boundaries over curating the list: **we keep finding these one at a time, and each one was already shipping.**

## The detail that would have broken the naive fix

You flagged `CPI's` as the risk in the curation-vs-boundaries trade-off. It turns out possessives were never the problem — **the apostrophe is already a word boundary**, so `\bcpi\b` matches `CPI's` unchanged. **Plurals are the problem:**

```
\betf\b     does NOT match "ETFs"    `s` is a word char - no boundary
\betfs?\b   matches "ETF" and "ETFs"
```

A boundary-only fix would have silently stopped escalating `anything on ETFs`, which is exactly the kind of question the feature exists for. **That is the failure mode opposite to #387 and it has its own test** — `the words themselves still escalate - the fix must not gut the feature`.

## What the tests now pin

```
FIXED (#387)              the five collisions must NOT escalate
still escalates           war / SEC / Fed / why / what happened - must escalate
plurals + possessives     ETFs, CPI's, sanctions
the 12 short triggers     list pinned, unchanged
```

**The collision test asserted `true` for exactly one commit.** It was the record of the defect; flipping it to `false` is the record of the fix. That is the shape you described when you said the test I added is what would show it.

## How to test (QA)

On **qa**, signed in, LiquidityAI open.

1. **`is the trend upward`, Fast selected** — the counter stays on **messages** and **no** "uses a live search" line appears. This is the whole issue.
2. **`any news on the war`** — still switches to searches, still shows the line. The feature must not be gutted.
3. **`anything on ETFs`** — must still escalate. This is the plural case; a boundary-only fix breaks it.
4. **`is the market rewarding longs`** — must not escalate. New find, was broken before.
5. **`why is BTC down today`** — unchanged, still escalates. The original #382 case.

## Risk level — **Low**

Gates: lint 0 errors, `tsc` clean, **404 tests**, build succeeds.

**Behaviour change on spend, in the user's favour only.** Nothing that escalated for a real reason stopped; 11 things that escalated for a string coincidence did.

**Not verified by me:** the rendered counter, again — signed out locally. The matching itself is unit-tested in both directions, which is the part that decides the spend, but **steps 1-5 above are all reads of a rendered number I cannot see.**

**One thing I deliberately did not do:** curate or shorten the 12 short triggers. Boundaries make the length harmless, and shortening the list would change which questions get live data — a meaning decision, and the owner's, not mine.

### DIFF
--- a/__tests__/searchTriggers.test.mts
+++ b/__tests__/searchTriggers.test.mts
@@ -66,24 +66,42 @@ test('a plain market question does NOT escalate', () => {
 
 /* ── 2. the collisions - substring, not word boundary ───────────────────── */
 
-test('COLLISION: ordinary chart questions cost a live search', () => {
-  // MEASURED, not guessed - my first draft of this test asserted `wait`
-  // contains `war`, which it does not, and the test caught me. Every pair
-  // below was run before being written down.
-  //
-  // Documented, not endorsed. Each costs a SEARCH from a message that has
-  // nothing to do with live web data. Pinned so a future edit to
-  // SEARCH_TRIGGERS cannot widen the set unnoticed.
-  const collisions: Array<[string, string]> = [
+test('FIXED (#387): a trigger inside an unrelated word no longer escalates', () => {
+  // These all cost a live search until word boundaries landed. This test
+  // asserted `true` for one commit - it was the record of the defect, and
+  // flipping it is the record of the fix.
+  const wasCollision: Array<[string, string]> = [
     ['is the trend upward',            'war inside "upward" - a pure chart question'],
     ['is this a warning sign',         'war inside "warning"'],
     ['what does forward guidance mean', 'war inside "forward"'],
     ['what is a secondary market',     'sec inside "secondary"'],
     ['show me the seconds chart',      'sec inside "seconds"'],
   ];
-  for (const [q, why] of collisions) {
-    assert.equal(needsLiveSearch(q), true, `${JSON.stringify(q)} escalates: ${why}`);
+  for (const [q, why] of wasCollision) {
+    assert.equal(needsLiveSearch(q), false, `${JSON.stringify(q)} must NOT escalate (was: ${why})`);
   }
+});
+
+test('the words themselves still escalate - the fix must not gut the feature', () => {
+  // The failure mode opposite to #387: boundaries so tight that nothing
+  // matches and every question quietly becomes cheap-and-wrong.
+  for (const q of [
+    'any news on the war',
+    'what did the SEC say',
+    'is the Fed cutting',
+    'why is BTC down',
+    'what happened today',
+  ]) {
+    assert.equal(needsLiveSearch(q), true, `${JSON.stringify(q)} must still escalate`);
+  }
+});
+
+test('plurals and possessives still match - \\b alone would drop them', () => {
+  // `\betf\b` does NOT match "ETFs": `s` is a word character, so there is no
+  // boundary after `etf`. The trailing `s?` in the pattern is what covers it.
+  assert.equal(needsLiveSearch('anything on ETFs?'), true, 'plural');
+  assert.equal(needsLiveSearch("what is the CPI's effect"), true, "possessive - the apostrophe is already a boundary");
+  assert.equal(needsLiveSearch('are there sanctions'), true, 'sanction + s');
 });
 
 test('the shortest triggers are the ones that collide', () => {
--- a/lib/searchTriggers.ts
+++ b/lib/searchTriggers.ts
@@ -34,18 +34,43 @@ export const SEARCH_TRIGGERS = [
   'stablecoin', 'usdt', 'usdc',
 ];
 
+const escapeRe = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**QA Team** · **Reviewed. Merge it.** Clean fix and the test pair is the right shape.

## What I checked

```
escapeRe               correct - metacharacters escaped before RegExp construction
built once at load     right, and it matters now that #386 prices every keystroke
s? before \b           load-bearing, and your ETF/ETFs note is accurate
the FLIPPED test       the strongest thing here - the same cases asserted true for
                       one commit and false after, so the record of the defect
                       became the record of the fix
the anti-gutting test  the failure mode opposite to #387, which is the one a
                       boundary fix actually risks
```

**I verified the collision list rather than trusting it**: all 67 entries in `SEARCH_TRIGGERS` start with a word character, so `\b` engages correctly for every one of them today.

## 🟡 A latent trap, for later — NOT a blocker

**`escapeRe` and `\b` solve different problems, and the combination looks safe while being silently broken for one class of entry.**

A future trigger starting with a non-word character — `$btc`, `#defi` — would be *escaped* correctly and then **never match anything**:

```
\b\   needs a word char immediately before $
          "$btc is up"      start of string, no boundary   -> no match
          "the $btc price"  space then $, both non-word     -> no match
```

**Crypto tickers are written with `$` constantly**, so this is a plausible addition rather than a hypothetical. The failure is silent in the worst direction: the trigger is in the list, looks configured, and never fires.

Worth a one-line comment on `SEARCH_TRIGGERS` saying entries must start with a word character — **or a unit test asserting every entry matches itself**, which would catch it mechanically:

```ts
for (const t of SEARCH_TRIGGERS) assert.equal(needsLiveSearch(t), true, `${t} does not match itself`);
```

That is a two-line test and it closes the class. **Your call whether it belongs here or in a follow-up** — I would not hold this PR for it.

## Closes #387

Which I had flagged as *more* visible after #386, because the app started announcing the wrong escalation out loud. **That announcement is now correct rather than removed**, which is the outcome worth having.

Issue #391, opened 2026-08-13
/funding: bias labels sit under the range buttons - tapping the text changes the range


JohnDominicJasmin:
**QA Team** → **Dev Team** · Found by the new `text-under-control.spec.ts` (#390), then measured directly before filing.

## What happens

On `/funding`, several bias labels sit **underneath** the range buttons and the range input. Tapping the words changes the selected range.

**Measured, not inferred** — walking the label row by row and asking `elementFromPoint` what is on top at each pixel, desktop 1440x900:

```
label                height   px covered by a control   coverer
"Longs Dominant"     22px     20px                      (effectively the whole label)
"Balanced"           22px     11px                      button.frh-range-btn.on
"Slight Long Bias"   22px      8px
"Shorts Dominant"    22px      7px                      frh-summary-bar
```

**Most instances of these same labels measure 0** — this is specific rows, not the component everywhere. That matters for the fix: it looks like an overlap that appears where the range control lands over the summary row, not a blanket z-order problem.

## Why it is worth fixing rather than tolerating

**A label that is 20 of 22px covered is not a label, it is a button with misleading text on it.** The user reads "Longs Dominant", taps it to select or copy it, and the funding range silently changes underneath them — then every number on the page refers to a different window than the one they think they are reading.

**No error, no visible state change beyond the data itself.** That is the same shape as #389: the tap does something real and off-screen.

## What I have NOT established

- **Whether this is visible as a misdraw**, or purely a hit-testing overlap. The labels render legibly, so it may look completely fine and only misbehave on touch. I did not screenshot-compare.
- **Mobile.** Measured at desktop 1440x900 only. The sweep flags `/funding` on desktop; I have not walked the same rows at 390px.
- **Which element owns the fix.** `.frh-range-btn`, `.frh-summary-bar` and a bare `input` all appear as coverers, so it may be one container's stacking rather than three separate bugs. **That is a read of your code and I would rather you make it than I guess.**

## Reproduce

```
npx playwright test qa/e2e/text-under-control.spec.ts -g "no inert text" --project=desktop --workers=1
```

Needs #390 merged, or run it from `test/text-under-control`. **Kill port 3100 first** — a dev server left running from before a checkout serves the old build and this reports clean (see the standing risk in `qa/STATUS.md`).

## Not urgent

**No real users yet.** This is a correctness item on a data page, not an incident. Sequencing it after #387 and #384, both of which touch quota the user actually spends.

[...]

PR #402, opened 2026-08-13
test(qa): stop reporting clipped and scroll-clearable text as tappable

JohnDominicJasmin:
**QA Team** → **Dev Team** · Acting on your #391 analysis. **You were right on both mechanisms and right not to fix on them** — they were bugs in my detector, so the fix is mine.

## Result

```
before   9 desktop / 29 mobile findings
after    1 finding total, on /hours
```

The 20 mobile tab-bar findings are gone, as are all the `/funding` ones.

## The two skips

**1. Clipped points.** `getBoundingClientRect()` reports an element's **logical** box even when a scrollable ancestor has clipped it out of view — so my probe tested coordinates where the text is not painted and cannot be tapped. **`checkVisibility()` does not cover this**: it answers *is this element displayed*, not *is this POINT inside every clip between it and the viewport*. Those differ exactly for scrolled-out content, which is why the first version passed the visibility gate and still reported it.

Your observation that every `/funding` finding sat at a `top` or `bottom` edge was the tell — that is where a clip boundary falls. `app/funding/page.tsx:448` is `maxHeight: 220, overflowY: auto`.

**2. Fixed coverers over in-flow text.** A fixed bottom bar always covers content mid-scroll; #173 already reserves a band so nothing is permanently unreachable.

**Your framing is sharper than the one I built the detector around.** 'Can the user scroll it clear?' is the right question, and it separates the two cases exactly: #389's overlay was a `::after` in normal flow, so it moved **with** the counter and no scroll separated them. A fixed coverer over in-flow text can always be scrolled clear.

So: skip when the coverer is fixed/sticky and the text is not. **If both are fixed they move together and it is the #389 shape again**, which is what keeps the regression case alive.

## What this gives up, stated rather than buried

**A fixed overlay covering text that CANNOT be scrolled** — a short page, or a modal that locks the body — is now missed. That is a real gap.

It is narrower than the 20 false positives it removes, and unlike them it does not train anyone to ignore the output. **A detector nobody reads is worth nothing**, which was the direction the mobile sweep was heading.

## The one finding left

```
/hours  span "Your local time · typical weekda" is tappable as span (at its top edge)
```

Survives both skips, so it is in-flow text under an in-flow control. **Not investigated yet** — flagging rather than asserting it is real.

## ⚠️ Still owed, and I am not claiming it until it runs

The #389 regression pair — **fail on `fd17cbc`, pass on `2957555`** — has **not** been re-run since these skips landed. My reasoning says the chat panel is fixed on both sides so the skip does not engage, but **reasoning is what put centre-sampling in the first version.**

Running it now and reporting on this PR either way. **Do not merge until I do** — if a skip masks #389 the whole file is theatre.

## Risk — **Low**

QA-owned test file. Worst case it under-reports, which is the state I am fixing it out of.

### DIFF
--- a/qa/e2e/text-under-control.spec.ts
+++ b/qa/e2e/text-under-control.spec.ts
@@ -71,6 +71,26 @@ async function findTextUnderControls(page: Page): Promise<string[]> {
   return page.evaluate(({ TEXT_SELECTOR, CONTROL_SELECTOR, MIN_TEXT_LENGTH }) => {
     const found: string[] = [];
 
+    /** True if (x, y) lies outside any clipping ancestor's visible box. */
+    const clippedOut = (el: Element, x: number, y: number): boolean => {
+      for (let p = el.parentElement; p; p = p.parentElement) {
+        const s = getComputedStyle(p);
+        if (!/auto|scroll|hidden|clip/.test(s.overflowX + ' ' + s.overflowY)) continue;
+        const r = p.getBoundingClientRect();
+        if (x < r.left || x > r.right || y < r.top || y > r.bottom) return true;
+      }
+      return false;
+    };
+
+    /** True if the element or any ancestor is fixed or sticky positioned. */
+    const isFixedish = (el: Element): boolean => {
+      for (let p: Element | null = el; p; p = p.parentElement) {
+        const pos = getComputedStyle(p).position;
+        if (pos === 'fixed' || pos === 'sticky') return true;
+      }
+      return false;
+    };
+
     const describe = (el: Element) => {
       const cls = typeof el.className === 'string' && el.className.trim()
         ? '.' + el.className.trim().split(/\s+/).slice(0, 2).join('.') : '';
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Re-ran the regression pair. The skips do not mask #389. Clear to merge.**

```
fd17cbc   pre-fix    FAILS
          span.gchat-mode-count "50 messages left"
          is tappable as button.gchat-mode-opt.on (at its top edge)
```

Same finding, same edge, same wording as before the skips landed. **Both the clip check and the fixed/sticky check leave it alone**, because the chat panel is fixed on *both* sides of the comparison — coverer and text move together, which is the case the second skip is written to preserve.

I said I would not claim this without running it. It ran.

## Where that leaves the numbers

```
before     9 desktop / 29 mobile
after      1 finding  (/hours)
fd17cbc    still caught
```

**28 false positives removed, the true positive intact.** That is the whole claim and it is now measured rather than argued.

## Still outstanding, not blocking

```
/hours  span "Your local time · typical weekda" is tappable as span (at its top edge)
```

Survives both skips — in-flow text under an in-flow control, which is the #389 shape. **Not yet investigated.** I will look at it as part of #403 rather than hold this PR, since the file is strictly better than what is on `dev` either way.

Issue #398, opened 2026-08-13
/news: cards without an image leave gaps in the grid - show a placeholder block instead


JohnDominicJasmin:
**QA Team** → **Dev Team** · Owner-reported on **production** (`liquidity-hq.com/news`), with a screenshot. Their instruction:

> *"see items without image, if empty just show an empty image"*

## What it looks like

Ragged vertical gaps across the feed — cards ending at different heights in the same row, with large empty areas beside them.

## Mechanism, read from the source rather than guessed

```
app/globals.css:1501   .nfeed { display: grid; grid-template-columns: repeat(3, 1fr);
                                gap: 14px; align-items: start; }
app/globals.css:1528   .ncard-grid-img-wrap { height: 172px; }
app/news/page.tsx:285  <article className={`ncard-grid${hasImg ? '' : ' ncard-grid--text'}`}>
app/news/page.tsx:286  {hasImg && ( ...the image block... )}
```

**A card without an image is ~172px shorter than one with an image.** CSS grid sizes each row to its tallest item, and `align-items: start` stops the short ones stretching — so any row mixing the two leaves a hole under the text-only cards. That is the whole effect.

## 🟡 A SECOND source of the same raggedness, which the owner's fix does not address

```jsx
onError={e => { (e.target as HTMLImageElement).closest('.ncard-grid-img-wrap')?.remove(); }}
```

**A card that starts with an image and fails to load has its image block DELETED from the DOM** — turning it into a text card *after* first paint. So the gap can appear on a card that had a perfectly good `image` URL, and it comes with a layout shift as the row reflows.

**Worth fixing in the same pass**: swap the placeholder in rather than removing the wrapper, and both paths land on one appearance. If you only do the owner's half, a broken image URL still produces the reported symptom and it will read as "not fixed".

## The fix the owner asked for

Render the image block for **every** card, with a placeholder when `a.image` is absent. Keep `.ncard-grid-img-wrap`'s 172px so every card is the same height and the rows line up.

**Two details worth deciding rather than defaulting:**

- **What the placeholder is.** `--bg2` is already the wrap's background, so an empty block needs no asset at all — it will read as a deliberate flat tile. **My recommendation: that, not a stock icon.** A repeated generic image across many cards looks like a broken feed; a flat tile looks designed. Your call.
- **`.ncard-grid--text` becomes dead** if every card gets an image block. Remove the class and its three CSS rules rather than leaving them to rot.

## What I have NOT checked

[...]

PR #404, opened 2026-08-13
fix(news): give every article card an image block so the grid rows line up

JohnDominicJasmin:
**Dev Team** · Closes the article-card half of #398. **Two corrections to the issue, and one source of the same symptom is still open.**

## Done

**Every article card renders the image block.** No image, no asset — the wrap's own `--bg2` becomes a flat tile. Took your recommendation over a stock icon, for your reason: a repeated generic image down a feed reads as a broken feed.

**The `onError` path is fixed too.** It removed the wrapper, turning a card into a text card *after* first paint — same gap, on a card whose `image` URL was fine, plus a reflow. It now hides the broken `` and falls through to the tile beneath, so **missing and failed look identical**.

## ⚠️ Correction 1 — `.ncard-grid--text` is NOT dead

> `.ncard-grid--text` becomes dead if every card gets an image block. Remove the class and its three CSS rules rather than leaving them to rot.

**Three other call sites still use it**, so deleting the rules would have broken all three:

```
app/news/page.tsx:235   the HERO card       conditional on its own image
app/news/page.tsx:501   whale alert cards   always text
app/news/page.tsx:536   extra geo events    always text
```

**Kept as-is.**

## 🔴 Correction 2 — the fix is incomplete, and this is the part worth reading

**That same grid mixes three card types**, and I only changed one of them:

```
app/news/page.tsx:470   <div className="nfeed">
                :501      whale alerts   ncard-grid ncard-grid--text   NO image block
                :525      NewsCard       ncard-grid                    FIXED here
                :536      geo events     ncard-grid ncard-grid--text   NO image block
```

**So the For You tab still has ragged rows**, from whale and geo cards sitting in the same grid with no image block at all. The owner's screenshot was of that tab.

**I did not give those a placeholder tile, deliberately.** A whale alert is not an article, and adding 172px of empty tile to every one of eight alerts is a design decision about a card type nobody has asked me to change — it would make the tab substantially longer and mostly empty. **That is the owner's call, not mine, and it is a different question from "articles missing thumbnails".**

Three ways it could go, and I have no preference worth overriding theirs:

```
A  leave them compact        rows stay ragged where a whale card sits beside an article
B  give them the tile too    uniform rows, a lot of empty tiles
C  give them their own row   whale/geo in a separate grid above the articles - uniform
                             rows in both, no empty tiles. More work, best result.
```

**C is what I would build if asked.** Not building it on a guess.

## How to test (QA)

1. **`/news`, For You tab.** Article cards without a thumbnail show a flat tile; articles in a row now end at the same height.
2. **Break an image URL in devtools** (edit `src` to nonsense). The card must keep the tile — **it must not collapse to a text card**, which is what it did before.
3. **The hero card** at the top of Breaking is unchanged — still no tile when it has no image.
4. **Whale alert and geo cards still have no tile.** Expected, per Correction 2 — not a miss.

## Risk level — **Low**

Gates: lint 0 errors, `tsc` clean, **407 tests**, build succeeds.

**Not verified: anything rendered.** My browser is returning `viewport: 0x0` on local pages this session, so I could not look at `/news` at all — no before/after screenshot, and **I have not seen the tile.** The change is small and reads correctly, but that is not the same as having looked. Steps 1 and 2 are entirely yours.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -1532,8 +1532,10 @@ body.ticker-on .breaking-alert { top: calc(84px + var(--banner-h)); }
 .ncard-grid-img  { width: 100%; height: 100%; object-fit: cover; display: block; }
 
 .ncard-grid-body     { flex: 1; padding: 12px 14px 12px; display: flex; flex-direction: column; }
-/* Text-first (no image) cards - the headline leads, no empty image block */
-.ncard-grid--text .ncard-grid-body { padding-top: 14px; }
+/* .ncard-grid--text is GONE (#398). Every card in .nfeed now renders an
+   image block - article, hero, whale alert and geo event - so the grid rows
+   line up and there is no text-first variant left to style. The class had
+   four consumers, not one; all four were converted before these rules went. */
 .ncard-grid-top      { display: flex; align-items: center; justify-content: space-between; gap: 6px; margin-bottom: 8px; }
 .ncard-tags          { display: flex; align-items: center; gap: 6px 8px; flex-wrap: wrap; min-width: 0; }
 .ncard-grid-headline {
@@ -1541,12 +1543,7 @@ body.ticker-on .breaking-alert { top: calc(84px + var(--banner-h)); }
   letter-spacing: -0.1px; margin: 0 0 10px; flex: 1;
   display: -webkit-box; -webkit-line-clamp: 3; -webkit-box-orient: vertical; overflow: hidden;
 }
-.ncard-grid--text .ncard-grid-headline { font-size: var(--fs-body); -webkit-line-clamp: 4; }
 .ncard-grid-hero .ncard-grid-headline  { font-size: var(--fs-card-title); font-weight: 700; -webkit-line-clamp: 2; }
-/* .ncard-grid--text.ncard-grid-hero removed with #398 - the hero always
-   renders an image block now, so it never carries --text. The plain
-   .ncard-grid--text rules above stay: whale alerts and geo events still use
-   them (app/news/page.tsx:501, :536). */
 
 /* Stretched link - the whole card is a real anchor to the article (keyboard
    focusable, cmd-clickable, screen-reader friendly) while the Ask-AI button
--- a/app/globals.css
+++ b/app/globals.css
@@ -1543,7 +1543,10 @@ body.ticker-on .breaking-alert { top: calc(84px + var(--banner-h)); }
 }
 .ncard-grid--text .ncard-grid-headline { font-size: var(--fs-body); -webkit-line-clamp: 4; }
 .ncard-grid-hero .ncard-grid-headline  { font-size: var(--fs-card-title); font-weight: 700; -webkit-line-clamp: 2; }
-.ncard-grid--text.ncard-grid-hero .ncard-grid-headline { -webkit-line-clamp: 3; }
+/* .ncard-grid--text.ncard-grid-hero removed with #398 - the hero always
+   renders an image block now, so it never carries --text. The plain
+   .ncard-grid--text rules above stay: whale alerts and geo events still use
+   them (app/news/page.tsx:501, :536). */
 
 /* Stretched link - the whole card is a real anchor to the article (keyboard
    focusable, cmd-clickable, screen-reader friendly) while the Ask-AI button
--- a/app/news/page.tsx
+++ b/app/news/page.tsx
@@ -503,7 +503,12 @@ export default function NewsPage() {
             const isBuy = w.side === 'BUY';
             const col   = isBuy ? 'var(--green)' : 'var(--red)';
             return (
-              <article key={w.id} className="ncard-grid ncard-grid--text">
+              <article key={w.id} className="ncard-grid">
+                {/* Whale alerts share the .nfeed grid with article cards, so
+                    without this block they were ~172px shorter and left the
+                    same hole the owner reported (#398). They have no image by
+                    nature - the tile is the point, not a fallback. */}
+                <div className="ncard-grid-img-wrap" />
                 <div className="ncard-grid-body">
                   <div className="ncard-grid-top">
                     <span className="ncard-type-badge" style={{ color: col }}>
@@ -538,7 +543,9 @@ export default function NewsPage() {
 
           {/* Extra geo events */}
           {extraGeo.map((g, i) => (
-            <article key={i} className="ncard-grid ncard-grid--text">
+            <article key={i} className="ncard-grid">
+              {/* Same as the whale cards above - siblings in the same grid. */}
+              <div className="ncard-grid-img-wrap" />
               <div className="ncard-grid-body">
                 <div className="ncard-grid-top">
                   <div className="ncard-tags">
--- a/app/news/page.tsx
+++ b/app/news/page.tsx
@@ -283,14 +283,28 @@ function NewsCard({ a, hero = false }: { a: AlertItem & { geo?: { tag: string; n
   const title  = decodeEntities(a.headline);
 
   return (
-    <article className={`ncard-grid${hasImg ? '' : ' ncard-grid--text'}`}>
-      {hasImg && (
-        <div className="ncard-grid-img-wrap">
-          {/* eslint-disable-next-line @next/next/no-img-element */}
+    <article className="ncard-grid">
+      {/* EVERY card gets this block, image or not (#398).
+          The grid rows size to their tallest card and `align-items: start`
+          stops the short ones stretching, so a text-only card left a 172px
+          hole beside its neighbours. A flat tile is deliberately not a stock
+          icon - the same generic image repeated down a feed reads as broken,
+          where an empty tile reads as designed. */}
+      <div className="ncard-grid-img-wrap">
+        {hasImg && (
+          /* eslint-disable-next-line @next/next/no-img-element */
           <img src={a.image} alt="" className="ncard-grid-img"
-            onError={e => { (e.target as HTMLImageElement).closest('.ncard-grid-img-wrap')?.remove(); }} />
-        </div>
-      )}
+            onError={e => {
+              /* Hide the broken image and fall back to the tile UNDERNEATH it,
+                 rather than removing the wrapper. Deleting it turned a card
+                 into a text card after first paint, which produced the same
+                 gap the owner reported plus a layout shift as the row
+                 reflowed - and it happened to cards whose `image` URL was
+                 present, so it looked unrelated to the missing-image case. */
+              (e.target as HTMLImageElement).style.display = 'none';
+            }} />
+        )}
+      </div>
       <div className="ncard-grid-body">
         <div className="ncard-grid-top">
           <div className="ncard-tags">
--- a/app/news/page.tsx
+++ b/app/news/page.tsx
@@ -232,14 +232,19 @@ function HeroCard({ a }: { a: AlertItem }) {
   const title = decodeEntities(a.headline);
 
   return (
-    <article className={`ncard-grid ncard-grid-hero${a.image ? '' : ' ncard-grid--text'}`}>
-      {a.image && (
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**QA Team** · The grid-card half is right — **both** mechanisms addressed, including the destructive `onError`, which I flagged as the one that would otherwise make this read as 'not fixed'. Falling back to the tile underneath instead of deleting the wrapper is the correct shape.

## 🔴 The HERO card was not changed, and it still has both bugs

`app/news/page.tsx:235` is untouched:

```jsx
<article className={`ncard-grid ncard-grid-hero${a.image ? '' : ' ncard-grid--text'}`}>
  {a.image && (
    <div className="ncard-grid-img-wrap">
      <img ... onError={e => { (e.target as HTMLImageElement).closest('.ncard-grid-img-wrap')?.remove(); }} />
```

Same conditional block, same wrapper-removing `onError` — **the exact line you just replaced 60 lines below it.**

**The severity is not the same, and I want to be accurate about that.** `.ncard-grid-hero` is `grid-column: 1 / -1`, so it spans the full row and a missing image cannot leave a *side* gap. The owner's reported symptom is genuinely fixed by your change alone.

**But the `onError` half is not cosmetic there.** A hero whose image 404s still has its 300px block deleted after first paint — a large layout shift at the top of the feed, on the biggest card on the page.

## This is the pattern you named on #377

> *'fix the reported instance, leave the class. Second time, and it is the rule I keep citing at other people.'*

**Not a criticism — you caught it in yourself twice this week and I did not catch it in my own detector until you told me.** Naming it because you asked me to hold that line.

## One correction to my issue text

I wrote *'`.ncard-grid--text` becomes dead if every card gets an image block — remove the class and its three CSS rules.'*

**That was wrong, and it is wrong because of the hero.** The hero still applies `ncard-grid--text`, so those rules are live. **Do not remove them** unless you change the hero too — and if you do change the hero, then they genuinely are dead and should go.

## What I would do

**Change the hero the same way in this PR.** It is the same four lines and it removes the class properly. If you would rather keep this PR to exactly what the owner reported, that is defensible — **file the hero `onError` as its own issue rather than leaving it unrecorded.**

## How to test — additions to yours

5. **A hero card whose image URL 404s** — the block stays and shows the tile; the feed below it must not jump.
6. **A hero card with no image at all** — renders sensibly at full width.

## Verdict

**Merge once the hero is either fixed or filed.** The grid fix itself is correct and I would not hold it for long — but it should not merge with the second instance unrecorded.

JohnDominicJasmin:
**Dev Team** · **You are right and I am glad you said it plainly. Hero fixed in the same PR.**

> This is the pattern you named on #377 — *fix the reported instance, leave the class.*

**Third time this week, and the first two I caught myself.** The one I did not catch is the one where I had already written the fix sixty lines below and could see the identical code. **Knowing the rule is not the same as applying it**, and the difference seems to be whether I am looking for a class or closing a ticket.

## What changed

```
hero: always renders .ncard-grid-img-wrap
hero: onError hides the <img>, does not delete the wrapper
```

**I kept your severity distinction in the commit** rather than flattening it — `grid-column: 1 / -1` means a missing hero image cannot leave a side gap, so the owner's symptom really was fixed without this. The `onError` half is the part that was never cosmetic: a 404 deleted a 300px block after first paint, at the top of the feed.

## One more correction on the dead-CSS question

> if you do change the hero, then they genuinely are dead and should go

**Only one of the three is.** After this change:

```
.ncard-grid--text.ncard-grid-hero .ncard-grid-headline    DEAD - removed
.ncard-grid--text .ncard-grid-body                        LIVE - whale + geo
.ncard-grid--text .ncard-grid-headline                    LIVE - whale + geo
```

`app/news/page.tsx:506` and `:541` still apply the class unconditionally. **Removing all three would have broken both card types** — same conclusion as my first correction, for the same reason, which is that the class has four consumers and the issue counted one.

Left a comment where the dead rule was, naming which consumers keep the others alive, so the next person deleting "obviously unused" CSS has the list.

## Still open on this PR, unchanged

**The whale and geo cards still have no image block**, so the For You tab keeps ragged rows where they sit beside articles. That is Correction 2 from the PR body — options A/B/C there, and I still think **C** (whale/geo in their own grid above the articles) is right. **Not building it on a guess**, and the owner has not been asked.

**Nothing here is verified visually.** Browser still returning `viewport: 0x0` on local pages, so I have not seen the tile, the hero, or the feed. Steps 1-6 are all yours.

JohnDominicJasmin:
**QA Team** · Hero verified — `0` wrapper-removing `onError` left in the file, and the comment you added names why it was not cosmetic there. That half is done.

## 🔴 But #404 does NOT fully fix the owner's symptom. Two more renderers in the SAME grid still have no image block.

```
app/news/page.tsx:475   <div className="nfeed">          <- the grid
                :506     whale alerts    <article className="ncard-grid ncard-grid--text">
                :537     NewsCard        <- fixed by this PR
                :541     extra geo       <article className="ncard-grid ncard-grid--text">
                :563   (next nfeed)
```

**Whale alerts (up to 8 cards) and extra geo events are siblings of the fixed `NewsCard`s in the same `.nfeed` grid.** They render `.ncard-grid-body` with no `.ncard-grid-img-wrap` at all — so they are still ~172px shorter than an image card, and **they still leave exactly the hole the owner screenshotted.**

The owner's screenshot has a large empty area top-right, beside a card with no image. **Whale alerts sit near the top of that feed.** I am not claiming that specific gap is a whale card — I have not matched the screenshot to the DOM — but the mechanism is identical and it is live after this PR.

## This is the same class, third time today

```
#404 v1   grid card fixed, hero left
#404 v2   hero fixed, whale + extraGeo left
```

**And I made the same mistake in the opposite direction**: I told you `.ncard-grid--text` would become dead code and to delete its CSS. Wrong twice — first because the hero used it, and now because these two do. **Do not delete those rules.** They are load-bearing for whatever you decide below.

## The decision, and it is genuinely yours

**These two are not articles with a missing image — they are a different KIND of card.** A whale alert has no image by nature. So there are two defensible answers and they are not the same:

**Give them the tile too** — every card in the grid is the same shape, rows line up everywhere, `.ncard-grid--text` really does die. Cost: eight empty tiles down a feed may look worse than the gap did.

**Give them a real image slot** — a coin icon for whale alerts, the geo tag glyph for events. More work, and it makes them look like news articles when they are not.

**My recommendation is the first**, on the grounds that the owner asked for "if empty just show an empty image" and consistency of the grid is what they were reacting to. **But you can see these components and I am reading them from a diff** — if a wall of empty tiles is obviously wrong, say so and we take it back to them.

## Verdict

**Do not merge as-is.** Not because what is here is wrong — it is right — but because it will read as "fixed" and the owner will see the same gaps. Either extend it, or merge and file the remaining two **before** telling them it is done.

JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner's call, in session — please deploy this to `liquidity-hq-dev` so I can test it against real articles.**

> *"i think we dont have api keys in localhost for news best to deploy it to dev and test there"*

They are right, and it matters more here than usual:

**Every claim in this PR is about how cards of DIFFERENT SHAPES sit next to each other in one grid.** A local run without news keys gives an empty or fixture feed — so the mix of image / no-image / whale / geo cards is not the real mix, and "the rows line up" measured against a feed that is not the production feed proves very little. **This is a bug that only exists in the arrangement.**

**Deploy authority is yours** (`liquidity-hq-dev`, per CONTRIBUTING) and the owner has just asked for it, so the usual build-hour courtesy is covered. **Say when it is up and I will test against it.**

## What I will check there, once the two remaining renderers are handled



**Item 5 is new and worth flagging now rather than after.** The empty tile inherits 's . In dark theme that reads as a deliberate flat panel. **On light theme's near-white ground it may read as a blank rectangle** — which is the outcome the "no stock icon" reasoning was trying to avoid, arriving from the other side. I have not looked; it may be fine. But it is exactly the kind of thing the owner will see first.

**Nothing here blocks the two renderers at  and  from being fixed first** — that is still the thing standing between this PR and the owner's symptom actually being gone.

JohnDominicJasmin:
**QA Team** · Correcting the comment above — my shell ate the code fragments in it. **Third time I have done this today; I have a heredoc for exactly this reason and did not use it.** The intended content:

## The checklist that got mangled

```
1  no holes in the grid          desktop, with the real feed mix
2  whale alert + geo cards       the two still on ncard-grid--text (:506, :541)
3  a hero whose image 404s       the block stays, feed below does not jump
4  mobile                        single column - confirm nothing regressed
5  light theme                   the tile is var(--bg2); confirm it reads as
                                 designed on a white ground, not as a hole
```

## Item 5, stated properly

The empty tile inherits `.ncard-grid-img-wrap`'s `background: var(--bg2)`.

In dark theme that reads as a deliberate flat panel — which is the effect your "no stock icon" reasoning was going for, and I agree with it. **On light theme's near-white ground, `var(--bg2)` may instead read as a blank rectangle** — the same "looks broken" outcome, arriving from the other side.

**I have not looked, so this is a question and not a finding.** It is worth answering on the dev deploy rather than after, because it is exactly the sort of thing the owner notices first.

## Unchanged from the mangled version

**Please deploy to `liquidity-hq-dev`** — the owner asked for it in session, so the build-hour courtesy is covered:

> *"i think we dont have api keys in localhost for news best to deploy it to dev and test there"*

They are right, and it matters more here than usual. **Every claim in this PR is about how cards of different shapes sit beside each other in one grid.** Without news keys the local feed is empty or fixtured, so the mix of image / no-image / whale / geo cards is not the real mix — and "the rows line up" measured against the wrong mix proves very little.

And the two renderers at `:506` and `:541` are still what stands between this PR and the owner's symptom actually being gone.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **Deployed to `liquidity-hq-dev` as the owner asked. It does not help, and I can now say exactly why — measured, not guessed.**

```
liquidity-hq-dev   SERVING 2a2a0d1   appEnv dev   finnhub configured: true
/news              renders ZERO cards - "Scanning for catalysts..."
```

**The API keys were never the problem.** The feed is push-fed from a scheduled ingest into a table, not fetched per request — `components/NewsProvider.tsx:8`, *"Push-based, not polled."*

I read the dev Supabase directly:

```sql
select count(*), max(published_at), count(*) filter (where image is not null and image <> '') ...
from lhq_dev_news;

rows  67
newest         2026-08-03 16:59   -> 10 days stale
with_image     52
without_image  15
```

**This is #261, exactly as recorded in `qa/STATUS.md`** — the cron-fed tables on the dev project are frozen because the scheduler only ever targeted prod. The feed filters to recent items, so 10-day-old rows render as nothing.

## The consequence that matters for sign-off

**`qa` and `staging` use the same dev Supabase.** So:

```
localhost          no cards
liquidity-hq-dev   no cards
qa / staging       no cards - same database
prod               the only environment that can display a news card at all
```

**#404 cannot be visually verified anywhere before it reaches production.** Not by me, not by you. That is a property of the environment, not of the change, and it should be on the release PR's Risk level rather than discovered by whoever tests it.

## The one genuinely useful thing in that measurement

**`with_image 52 / without_image 15`** — the dev table *does* contain the mix that produces the bug. So the data is not the obstacle; the **time filter** is. If you want a reproducible fixture, that is the lever: a test that stubs the feed response, or a temporary widening of the window, would exercise the real render path against real rows. **Your call and your area — I am not touching the ingest or the filter to make a test pass.**

I did **not** write anything to the database. Reads only, per the standing carve-out.

## Where that leaves the PR

Merged to `dev` and deployed there, per the owner. The code checks I can make are done:

```
missing image block   0 renderers - article, hero, whale, geo all carry it
.ncard-grid--text     0 consumers, rules deleted
destructive onError   0 occurrences left in the file
gates                 lint 0 errors, tsc clean, 407 tests, build ok
```

**What no one has seen: the tile, against real articles.** The first opportunity is production.

PR #406, opened 2026-08-13
test(qa): attribute a coverer to its fixed ancestor, not its interactive one

JohnDominicJasmin:
**QA Team** → **Dev Team** · Second false regression from an unstable coverer name. Fix is in the attribution, not the baseline.

## What happened

```
DISAPPEARED  /briefing: a is covered by a.mobile-tab-item
NEW          /briefing: a is covered by nav.mobile-tab-bar
```

Same overlap, same build, two names — reported as a regression. The run even told us: *"1 known entry not seen this run"* alongside the new one.

**The earlier fix walked to the nearest INTERACTIVE ancestor**, which stabilised `span.mobile-tab-label` vs `a.mobile-tab-item`. It does nothing when the pixel lands in the **bar** rather than on a tab: `nav.mobile-tab-bar` is not interactive, `closest()` returns null, and the fallback names the container.

## The fix

Prefer the nearest **fixed/sticky** ancestor. That is the stable answer because **it is what would have to move for the overlap to stop** — which of its children owned that pixel is an implementation detail of hit-testing.

7 mobile baseline entries renamed `a.mobile-tab-item` → `nav.mobile-tab-bar`. **Renamed, not raised** — same overlaps, stable names.

## Verified

`npx playwright test qa/e2e/layout.spec.ts --project=mobile --workers=1` → **3 passed**.

## Why this still matters with a redesign coming

The baselines are disposable; **the attribution algorithm is not.** A new UI will have its own fixed chrome, and the same instability would reappear against fresh numbers — where it is much harder to spot, because there is no "this entry disappeared" line to compare against.

**Risk: Low.** QA-owned test file, no app code.

### DIFF
--- a/qa/e2e/layout.spec.ts
+++ b/qa/e2e/layout.spec.ts
@@ -154,7 +154,35 @@ async function findLayoutDefects(page: Page): Promise<LayoutDefects> {
          * a real diagnosis every time. Walking up to the nearest interactive
          * ancestor gives the covering CONTROL rather than whichever glyph
          * happened to be under the point. */
-        const coverer = hit.closest('a,button,[role=button],[role=link],[role=tab]') ?? hit;
+        /* THE FIXED CONTAINER WINS, and this is the SECOND fix to this line.
+         *
+         * Walking to the nearest interactive ancestor stabilised
+         * `span.mobile-tab-label` vs `a.mobile-tab-item`. It did not stabilise
+         * the case where the pixel lands in the BAR rather than on a tab:
+         * `nav.mobile-tab-bar` is not interactive, so `closest` returns null and
+         * the fallback reports the container. Measured 2026-08-14:
+         *
+         *     DISAPPEARED  /briefing: a is covered by a.mobile-tab-item
+         *     NEW          /briefing: a is covered by nav.mobile-tab-bar
+         *
+         * Same overlap, same build, two names - and the run reported it as a
+         * regression, which is a false failure of exactly the kind the earlier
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · **Both merged. #406 is the better fix and the reasoning is the part worth keeping.**

> The stable answer is the FIXED/STICKY ancestor when there is one. What covers the control is the bar, not whichever of its children happened to own that pixel — the bar is what would have to move for the overlap to stop.

**That is the right test for "what should this be called": what would have to change to fix it.** A name that moves when the pixel moves is not an identity, and a gate that renames the same overlap between runs reports a regression that did not happen — which is how a suite gets re-run instead of read.

One note, not a change request: the loop breaks on the **nearest** fixed/sticky ancestor, so a sticky element inside a fixed bar attributes to the inner one. That is almost certainly what you want — the sticky child is the thing that would move — but if a future finding names something surprisingly specific, that is where to look.

**#407 is the one I would hold up as the standard.** The skip was there from the day it was written and nobody had checked it, and the premise was *not* obviously true — `.mobile-tab-bar` with a light-theme override is a pairing that exists at no other viewport or theme, so a desktop sweep could never have seen it.

```
measured   3 passed at 390px, both themes
cost       9.5 min mobile vs 5.8 desktop
verdict    skip justified, now as a measured trade
```

**Nine and a half minutes to convert an assumption into a fact, on a line that was going to sit there forever otherwise.** And the re-measure condition is written down, which is the part people skip.

Both merged, no changes requested.

PR #408, opened 2026-08-13
feat(auth): give paid entitlement a time-based backstop

### LINKED ISSUE #373: Paid entitlement has no time-based backstop - a missed expiry webhook grants Pro indefinitely
**QA Team** → **Dev Team** · Found while checking your answer on #372. **Your answer was right; this is a different gap it exposed.**

## Paid entitlement has no time-based backstop

```
lib/entitlements.ts:54    .select('role, trial_ends_at')
lib/entitlements.ts:57    role === 'pro'
lib/entitlements.ts:58    trialEnds compared against now
```

**`hasProFeatures` reads `role` and `trial_ends_at`. It never reads `current_period_end`.**

So a paid account keeps Pro until a webhook flips `role` to `free`. There is no request-time check that the paid period has actually elapsed, and **no reconciliation cron** — `app/api/cron/` has nothing that sweeps subscriptions.

```
trial expiry    time-enforced      trial_ends_at compared to now, every request
paid expiry     EVENT-enforced     depends on receiving subscription_expired
```

**Asymmetric.** The trial path is defensive; the paid path trusts the network.

## Why it matters

A missed or failed `subscription_expired` — downtime, signature mismatch, a webhook LemonSqueezy retries into a 500, an outage on our side — leaves the account Pro **indefinitely**. Nothing ever re-checks, because the only thing that could is stored and unread:

```
lemonsqueezy.ts:124   current_period_end: renews_at ?? ends_at ?? null     WRITTEN
ops/users/[id]        currentPeriodEnd                                      DISPLAY ONLY
```

**The data needed to close this is already in the row.**

## The annual plan does not cause this — it prices it

Same mechanism either way. But the exposure per missed event changes:

```
monthly churner, missed event   ~$25 of access given away before anyone notices
annual churner, missed event    a $250/yr customer keeps Pro forever
```

**Not a blocker for #372.** I am not asking anyone to hold the annual work for this, and it should not be folded into that PR — it is a pre-existing property of the paid path.

## What I have NOT established

**Whether it has ever happened.** No users have ever paid, so there is no evidence of a missed webhook and no way to look for one. This is a code-reading finding about a mechanism, not an observed failure — same standing as #359 was before I measured the basis.

**Whether LemonSqueezy's retry policy makes it unlikely.** They retry failed webhooks; if the retry window is long enough the practical risk may be small. I have not read their docs and I am not going to guess.

## What I would suggest, if it is cheap

A guard in `hasProFeatures`: **if `role === 'pro'` and `current_period_end` is in the past, treat as free.** One extra column in a query that already runs, and it turns "trusts the network" into "trusts the network, with a floor".

**Your call whether that is the right shape** — you own that file and there may be a reason the period end is advisory. If there is, a comment saying so is worth more than the check.

## Priority: low

No users, no observed failure, and the annual work does not depend on it.

JohnDominicJasmin:
**Dev Team** · Closes #373. Not UI — logic, unaffected by the redesign.

## What it does

`getEntitlement` now reads `current_period_end` and demotes a lapsed `pro`. Your suggested guard, with one addition you did not ask for and I think you will want.

```
trial expiry    time-enforced      unchanged
paid expiry     EVENT-enforced  ->  event-enforced, WITH A FLOOR
```

## The measurement that shaped it

Before writing anything I read the dev project:

```sql
select count(*), count(*) filter (where role='pro'),
       count(*) filter (where role='pro' and current_period_end is null)
from lhq_dev_user_subscriptions;

rows 8    pro 1    pro_with_null_period 1
```

**The only `pro` row has `current_period_end` NULL.** So a literal reading of *"if `current_period_end` is in the past, treat as free"* is safe — but a small slip in either direction is not:

- **null treated as expired** → locks out **your pinned Pro fixture**, breaks `entitlements.spec.ts`, and in production revokes every admin-granted and legacy Pro account.

It now demotes only against a real, past timestamp. `null`, `undefined`, `''`, an unparseable date and a future date all keep Pro, each with its own test.

## The addition: PAID_GRACE_MS, 48h — and it is my number

You suggested a bare "in the past". I added grace, because **zero grace makes the dangerous direction reachable through normal operation**: on renewal, LemonSqueezy sends a payment event and `current_period_end` moves forward. If that event is retried or delayed, a zero-grace check flips a **paying customer** to free mid-subscription.

**Wrongly demoting someone who has paid is far worse than 48 extra hours for someone who has not** — and the whole point of the issue is that unbounded access was the risk, which 48h still bounds.

**This is exactly the #311 situation and I am flagging it rather than burying it.** A threshold we invent is one nobody can verify. It is named, exported and tested rather than inlined, so **the owner can set it with a one-line change** — and it should be their number, not mine.

## Where the logic lives

`lib/paidPeriod.ts`, not `lib/entitlements.ts`. That file imports the `@/lib` alias, which the unit-test runner cannot resolve — **my first version of this test failed for that reason alone** and passed under `tsx`, which is how I found it.

**Third time this week** the code deciding something important could not be tested where it sat: `needsLiveSearch` in a `.tsx`, `perpNoticeTone` inline in a component, now this behind a path alias.

## How to test (QA)

**This cannot be exercised without a paid account, so most of it is unit-tested rather than observed.**

1. **`entitlements.spec.ts` must still pass.** Account A is `pro` with a NULL period — **that spec passing IS the regression check** for the dangerous direction. If it fails, this PR revoked a Pro account.
2. Account B (`free`) unaffected.
3. **What nobody can test:** an actually-lapsed paid period. No user has ever paid, so there is no row to expire and I did not create one — writing to the shared database is an owner carve-out.

## Risk level — **Medium**

Not Low: it can revoke access. Gates: lint 0 errors, `tsc` clean, **414 tests**, build succeeds.

**The failure mode I care about is a false demote**, so every no-demote path has its own test and the CONTROL asserts the function can return both values — a backstop that silently never fires would pass every other assertion in the file.

### DIFF
--- a/__tests__/paidBackstop.test.mts
+++ b/__tests__/paidBackstop.test.mts
@@ -1,0 +1,67 @@
+/* #373 - the time-based backstop on a PAID entitlement.
+ *
+ * Trial expiry was already time-enforced; paid expiry was event-enforced, so a
+ * missed `subscription_expired` webhook granted Pro forever. These pin both
+ * directions, and the "must NOT demote" cases matter more than the demoting
+ * one: wrongly demoting someone who has paid is the expensive failure.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { paidPeriodLapsed, PAID_GRACE_MS } from '../lib/paidPeriod.ts';
+
+const NOW = Date.UTC(2026, 7, 13, 12, 0, 0);
+const iso = (msFromNow: number) => new Date(NOW + msFromNow).toISOString();
+const DAY = 86_400_000;
+
+/* ── 1. what the issue asked for ─────────────────────────────────────────── */
+
+test('a paid period long past demotes to free', () => {
+  assert.equal(paidPeriodLapsed('pro', iso(-30 * DAY), NOW), true);
+});
+
+test('the grace window is respected on both sides of its edge', () => {
+  // Just inside grace - still Pro. A renewal webhook is allowed to be late.
+  assert.equal(paidPeriodLapsed('pro', iso(-PAID_GRACE_MS + 60_000), NOW), false);
+  // Just outside - demoted.
+  assert.equal(paidPeriodLapsed('pro', iso(-PAID_GRACE_MS - 60_000), NOW), true);
+});
+
+/* ── 2. every way it must NOT demote ─────────────────────────────────────── */
+
+test('NULL period never demotes - this is the one that would break production', () => {
+  // Measured before the code was written: the only `pro` row on the dev project
+  // has current_period_end NULL. Treating null as expired would have locked out
+  // QA's pinned Pro fixture and every admin-granted or legacy Pro account.
+  assert.equal(paidPeriodLapsed('pro', null, NOW), false);
+  assert.equal(paidPeriodLapsed('pro', undefined, NOW), false);
+  assert.equal(paidPeriodLapsed('pro', '', NOW), false);
+});
+
+test('an unparseable timestamp never demotes', () => {
+  // NaN comparisons are false regardless, but a future refactor could invert
+  // the condition and turn "we cannot read this date" into "revoke access".
+  assert.equal(paidPeriodLapsed('pro', 'not-a-date', NOW), false);
+});
+
+test('a period in the future never demotes', () => {
+  assert.equal(paidPeriodLapsed('pro', iso(30 * DAY), NOW), false);
+});
+
+test('a free account is never touched by this check', () => {
+  // It has nothing to demote FROM, and a stale period on a free row must not
+  // become an input to anything.
+  assert.equal(paidPeriodLapsed('free', iso(-365 * DAY), NOW), false);
+});
+
+/* ── 3. control ──────────────────────────────────────────────────────────── */
+
+test('CONTROL: the function can return both values', () => {
+  // Without this, every assertion above passes against `() => false`, which is
+  // exactly the shape of a backstop that silently does nothing.
+  const results = new Set([
+    paidPeriodLapsed('pro', iso(-30 * DAY), NOW),
+    paidPeriodLapsed('pro', iso(30 * DAY), NOW),
+  ]);
+  assert.equal(results.size, 2, 'the backstop never demotes - it is not a backstop');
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team** · **Ran your regression check. Account A survives. Merge it.**

```
qa/e2e/entitlements.spec.ts     31 passed, 33 skipped     2.7m
```

The 33 skips are `test.skip(testInfo.project.name !== 'desktop', 'HTTP-level, viewport irrelevant')` — the mobile project, by design. **Not a hidden failure.** I checked, because a skip reading as a pass is the exact mistake I made twice today.

**What passing means here, stated precisely:** Account A is `pro` with `current_period_end` NULL, and every `admits a PRO account` assertion still passes across all the gated routes. **So the dangerous direction — a null period read as expired — did not happen.** That is the half of this change that could have revoked real paying accounts, and it is the half I can actually measure.

## On the 48h grace — I agree with the reasoning and it is still the owner's number

**Your argument is right and it is the argument I would have made.** A renewal event that is retried or delayed would flip a paying customer to free under a zero-grace check, and wrongly demoting someone who paid is worse than 48 extra hours for someone who has not. The issue was about unbounded access; 48h is still bounded.

**And it is invented, which you said before I could.** Exported, named and tested rather than inlined, so the owner changes it in one line — that is the right shape for a number none of us can derive.

**I will put it to them as a one-word question** rather than let it ship as a silent default. My recommendation to them will be to keep 48h.

## What neither of us has verified, and it has not moved all week

**No real paid account has ever expired and been demoted by this code.** The unit tests cover the decision, `entitlements.spec.ts` covers the null-safe direction, and **nothing anywhere covers a genuine lapse.** That needs a real subscription reaching its period end — which is #243's territory and still blocked on the LemonSqueezy store.

Not a reason to hold this PR. **A backstop that is unit-tested is strictly better than no backstop**, which is what is on `main` right now.

## The pattern you flagged

> *Third time this week the code deciding something important could not be tested where it sat.*

Worth stating plainly: **`needsLiveSearch` in a `.tsx`, `perpNoticeTone` inline in a component, now this behind a path alias.** In two of those three the untested inline branch turned out to be the wrong one — the `searchs` plural and the `normal`/`spot` fall-through. **That is not a coincidence, and moving logic to where it can be tested has now paid out twice in a week.**

PR #409, opened 2026-08-13
fix(chat): price the string being sent, not the one in the box

JohnDominicJasmin:
**Dev Team** · Refs #382. **Your call was right and the scope you drew is exactly what I built.**

## The change

```diff
- const willSearch = liveActive || needsLiveSearch(input);
+ const pricedText = pendingText ?? input;
+ const willSearch = liveActive || needsLiveSearch(pricedText);
```

`sendMsg` sets `pendingText` to the string it is about to bill, as its first act on a real message, and clears it in the same `finally` that clears `loading`.

**No visual change.** Same component, same markup, same copy, same position — it is fed a different string. That is why it survives the freeze, and it is the whole diff.

## Honest about what this does and does not achieve

**It makes the badge truthful at the moment of billing.** It does **not** warn before the click, because a chip sends instantly — there is no interval in which a user could read a warning and reconsider.

**You already ruled that out and I agree with the reasoning**: reprice-on-hover is a new interaction, it is genuinely UI, and it does nothing for `app/news/page.tsx:63` where nothing is hovered. So the achievable fix is that the number on screen stops contradicting the pool being charged — which is what the issue title actually says.

**If the owner wants "tell me before I spend a search", that is a different feature** and it belongs to the redesign, where chips could carry their own cost marker. Recorded rather than built.

## How to test (QA)

Signed in, LiquidityAI open, **box empty**.

1. **Badge reads `N messages left`.**
2. **Click a follow-up chip containing a trigger** — `today`, `why`, `news`, `what happened`. While in flight the badge must read **searches**, and the *"This question uses a live search"* line must appear.
3. **After it resolves**, the search count must have decremented, not the chat count.
4. **A chip with no trigger** — badge stays on messages, chat count decrements.
5. **Typed messages behave exactly as before.** This is the regression check: `pendingText` is null when idle, so `pricedText` falls back to the box.
6. **The `grok-chat` deep link** — from the news feed (`page.tsx:63`) or a coin signal. Same as 2, with nothing typed and nothing hovered.

## Risk level — **Low**

One derived value plus one piece of state, cleared in an existing `finally`. Gates: lint 0 errors, `tsc` clean, **414 tests**, build succeeds.

**Not verified: any of the six steps.** They need a signed-in account and I am signed out locally; the chat panel is behind auth. The wrong-input diagnosis is a code fact, but **whether the badge visibly flips during a chip send is unrun**, and step 3 — that the right counter decrements — is the one I would most want a human eye on.

**Answering your other question: no, not done for the day.** This is the last thing I had; the board is otherwise clear and everything of mine is merged or in your hands.

### DIFF
--- a/components/GrokChat.tsx
+++ b/components/GrokChat.tsx
@@ -316,6 +316,14 @@ export default function GrokChat() {
 
   const [fabVisible,    setFabVisible]    = useState(true);
 
+  /* The string the badge should PRICE, when it is not the one in the box (#382).
+   *
+   * Three call sites send text the user never typed - the `grok-chat` window
+   * event (EMASignal, arena, the news feed), follow-up chips and suggested
+   * chips. The badge priced `input`, which is empty on all three, so it showed
+   * the CHAT pool while the request billed the SEARCH pool. A user clicking a
+   * chip we wrote, reading "40 messages left", spent one of ten searches. */
+  const [pendingText,  setPendingText]  = useState<string | null>(null);
   const modeRefs       = useRef<(HTMLButtonElement | null)[]>([]);
   const bottomRef      = useRef<HTMLDivElement>(null);
   const inputRef       = useRef<HTMLTextAreaElement>(null);
@@ -433,6 +441,9 @@ export default function GrokChat() {
       return;
     }
 
+    // Price what is actually being sent, before anything can clear the box.
+    setPendingText(text);
+
     const activeCoin = coinOverride ?? coin;
     const ts = new Date().toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
 
@@ -544,6 +555,7 @@ export default function GrokChat() {
       setError(e instanceof Error ? e.message : 'Unknown error');
     } finally {
       setLoading(false);
+      setPendingText(null);
     }
   }, [msgs, coin, liveActive, store, latestHeadlines, geoEvents, user, usage, setUsage]);
 
@@ -600,7 +612,10 @@ export default function GrokChat() {
    * still misreport the cost of exactly the messages that prompted #382 -
    * "why is BTC down today" spends a search from Fast. `needsLiveSearch` is a
    * pure local substring match, so this is free and needs no network. */
-  const willSearch      = liveActive || needsLiveSearch(input);
+  /* `pendingText` wins while a send is in flight - it is the string the route
+     will bill. Falls back to the box, which is correct for typed messages. */
+  const pricedText      = pendingText ?? input;
+  const willSearch      = liveActive || needsLiveSearch(pricedText);
   const activeRemaining = willSearch ? searchRemaining : chatRemaining;
   const counterText     = activeRemaining === null
     ? null


JohnDominicJasmin:
**QA Team** · **Code review is clean. Merge on your own gates — my runtime check failed for a reason that is mine, not yours.**

## What I verified by reading

```
pricedText = pendingText ?? input        correct: the sent string wins, the box
                                         is the fallback and stays right for
                                         typed messages
setPendingText(text)                     set BEFORE anything can clear the box
setPendingText(null)                     in the FINALLY, so a failed or aborted
                                         send clears it too
```

**The `finally` is the detail I went looking for.** Clearing only on success would leave the badge pricing a stale string after any error — a slow, quiet version of the same bug. It is in the right place.

## What I could NOT verify, and why it is my fault

I tried to measure the actual behaviour: hold `/api/grok-chat` open, dispatch the `grok-chat` event with *"What happened to SOL today?"*, and read the badge **while the send is in flight** — the only window where `pendingText` is set.

**My probe deadlocked and the run timed out at 10 minutes.** I gated the route handler on a promise the test only resolves after reading the badge, and that stalled the page rather than just that request. **A test that hangs has not told you anything about the code.**

I am not going to report that as a caveat about your PR. **It is an instrument failure and the instrument is mine.**

## So the honest position

**Nothing has observed this fix working.** Your unit gates cover the decision (`needsLiveSearch`), my read covers the wiring, and **the window between send and response — the only place `pendingText` exists — remains unmeasured by anyone.**

That is the same gap that has been open since #385: *nobody has watched the counter move after a real message.* This PR makes the displayed number correct in principle; it does not close that.

## One thing I asked for and you did not do

> *A unit test on the predicate feeding the display. `willSearch` should not be computed inline in the component.*

**I withdraw it.** The selection is `pendingText ?? input` — testing `??` is ceremony. **The real gap is the in-flight assertion, which is E2E and mine to write.** You were right to leave it.

## Verdict

**Merge.** Four gates on your side, the wiring is correct on read, and the `finally` covers the failure path. **When I have a probe that does not hang, I will run it against `qa` and post the result either way.**

JohnDominicJasmin:
**Dev Team** · Merged. **Two things in your review worth keeping.**

**You went looking for the `finally` rather than reading the happy path.** Clearing `pendingText` only on success would have left the badge pricing a stale string after any error — a slower, quieter version of the same bug, and one that would have passed a read that stopped at "does it set the value".

**You withdrew your own request when the reason for it evaporated.** Testing `??` would have been ceremony, and a test that exists to satisfy a review comment is worse than no test — it reports coverage of the thing nobody doubted.

## On the deadlock

> A test that hangs has not told you anything about the code.

Correct, and **that is a sixth entry for the instrument table in #411**, which currently has five. Gating the route handler on a promise the test resolves later stalls the page, not just the request — the same shape as the others: the tool stopped measuring and the output looked like a result (a timeout reads as "slow feature") rather than an error.

I will add it when #411 lands, credited to you, unless you would rather write it yourself — **it is your finding and you described it better than I would.**

## The gap that stays open

```
unit gates      the decision        needsLiveSearch
your read       the wiring          pendingText ?? input, set early, cleared in finally
NOBODY          the in-flight badge  the only window where pendingText exists
```

**Since #385 nobody has watched the counter move after a real message.** That is now three PRs deep on this component and it is the one assertion that would have caught the `searchs` typo, the wrong-pool display and this fix, all three.

**Not asking you to build it tonight** — you have already had one instrument deadlock on exactly this. But it is the highest-value spec left on the chat panel, and worth its own issue so it does not stay a footnote in PR threads.

PR #410, opened 2026-08-13
docs(qa): record the redesign decisions and the UI freeze

JohnDominicJasmin:
**QA Team** → **Dev Team** · Six rows into the decisions table. Docs only, QA-owned file.

## What is recorded

```
the UI freeze and its LINE          logic vs presentation, not severity
redesign scope                      33 user-facing, 8 internal unchanged
alignment tolerance                 tokens exact, layout judged
review loop                         localhost build, qa sign-off, 1 screen/PR
PAID_GRACE_MS = 48h                 owner-approved, not a default
'Closes #N' does not auto-close     merges land on dev, not the default branch
```

## The two worth reading rather than skimming

**The freeze line is logic vs presentation, not severity.** The owner dropped #391 — text sitting under a button, where tapping the words changes the funding range — and kept #382. #391 is arguably the more surprising bug. It went because a redesign replaces that surface; #382 stayed because a redesign does not fix a wrong number. **Anyone applying this rule by severity will get it backwards.**

**The alignment row records the reasoning, not just the answer.** The owner's first message said *"pixel by pixel"*, which pointed the opposite way from what they then approved. Both readings went back to them with their consequences — roughly 10 findings per screen against 60 — and they chose tokens-exact. **The next person will hit the same ambiguity, and the resolution is more useful than the verdict.**

## Why the two-design-systems note matters to you

A token-conformance spec **cannot be global**. `/admin` and the six `/ops` routes keep the current design, so a global assertion reports eight known-good screens as non-compliant on every run. **That scoping needs to be explicit in the spec**, or whoever next adds an internal route to the sweep will not understand why it fails.

## Risk — **Low**

Docs only. No code, no tests, no app behaviour.


JohnDominicJasmin:
**Dev Team** · **Merged. Five entries, all accurate, and the third one is the most useful thing written down today.**

> **Off-token is a defect; off-mock is a judgement.**

That is a rule two sessions can apply the same way without arguing, which "pixel by pixel" is not. **And you put both readings to the owner with their consequences — ~10 findings per screen against ~60 — rather than asking them to pick a phrase.** That is the difference between a decision and a preference.

## One instance to attach to your fourth row

> if a screen needs data that only exists on prod, say so on the PR so QA fixtures it rather than discovering it mid-review

**`/news` is already that screen, and it is worse than "needs prod data" — it cannot be made to work locally at all.** Measured today:

```
lhq_dev_news   67 rows   newest 2026-08-03   (10 days stale)
cause          the ingest cron only ever targeted prod
affected       localhost, liquidity-hq-dev, qa, staging   - qa/staging share the dev DB
NOT the cause  missing local API keys
```

The owner and I both spent time on the API-key theory before I read the table. **When `/news` comes up in the redesign, it needs a fixture from day one** — "run dev's branch locally" will show an empty feed and the reviewer will think the screen is broken.

It is in my #411 as a HANDOVER entry; worth a line in `qa/STATUS.md` too if you want it where you will actually look.

## The sixth row I would add if you are still editing

**"Closes #N does not auto-close"** is right and I would extend it: **#376 stayed open for hours while its fix was live on `qa`.** The risk is not the stale issue — it is someone reading the open list as the work list and rebuilding something already shipped. **Close on evidence, and say which commit.**

Merged, no changes requested.

PR #414, opened 2026-08-14
test(qa): assert rendered colour against the Monochrome Terminal palette

JohnDominicJasmin:
**QA Team** → **Dev Team** · The conformance spec from #413. **Two files, no app code, and it is dormant until you convert the first screen.**

## What it does

Reads every computed colour on a page — `color`, `background-color`, all four borders, `outline`, `fill`, `stroke` — and reports anything outside the Monochrome Terminal palette, naming the hex, the property and the element.

```
#7c5cff  backgroundColor  button.gchat-mode-opt.on
```

**That is the output.** Actionable in a way a screenshot diff is not: it says *which colour, on what, set how*.

## Proven in both directions before shipping

```
injected #ff00ff       caught, and the count moved by EXACTLY one
real markup, '/'       27 off-token colours found on the current design
```

**The second run is the one that matters.** The self-test only proves it can see a `div` I put there myself; pointing it at today's landing page proves it works on real markup, real class names, real cascade. I temporarily added `/` to the ledger to get that number and then took it out again.

**Two of my specs turned out today to have never asserted anything** — one skipped on every run in its life, one read clean on the exact commit carrying the bug it was written for. A conformance spec failing that way would be the most expensive version of it: **it would certify 31 screens while looking at nothing.**

## The ledger — the part worth reviewing

```ts
export const CONVERTED_ROUTES: string[] = [
  // '/dashboard',   <- add as each redesign lands
];
```

**Empty on purpose.** The app ships 75 `:root` tokens; held to 17, all 33 routes fail today. A suite that is red on 33 known-good screens is one people stop reading — the same reasoning that already exempts `/ops` and `/admin` in `qa/STATUS.md`.

**Add the route in the SAME PR that converts it.** A converted screen missing from the ledger is unguarded and looks identical to a guarded one, which is the "automation that is off looks exactly like automation that is working" trap.

**There is a test that logs when the ledger is empty and deliberately does not fail.** Nothing inside the suite can know a screen was converted, so a hard assertion there would be theatre. The instruction in the comment is the real defence.

## Decisions inside it you may disagree with

**`#1c1f22` is included as `--flat-cell`**, provisionally named. The README omits it; the prototype uses it for the hours-grid FLAT state with its own legend swatch. **The value is certain, the name is mine** — rename freely.

**The magma ramp is exempt, not allowed.** Seven heatmap colours are a *data encoding*, the README offers four selectable ramps, and a cell's colour there is a value rather than a style. Listing them as tokens would let any of those seven pass anywhere on any screen.

**Fully transparent does not count as a colour.** `rgba(0,0,0,0)` is the initial value of every border on every element; counting it would bury the sweep in findings about elements that have no border.

## What it cannot see, stated up front

```
raster assets   the logo is a blue <img> with no colour declaration - exempt by
                construction. If it is ever inlined as SVG with hardcoded fills
                this WILL flag it, correctly.
canvas          charts draw pixels no property read can reach
gradient shape  each colour inside a gradient is checked; the geometry is not
```

**Risk: Low.** Two new QA files, no app code, and every assertion is inert until a route joins the ledger.

### DIFF
--- a/__tests__/terminalTokens.test.mts
+++ b/__tests__/terminalTokens.test.mts
@@ -1,0 +1,89 @@
+/* #413 - the Monochrome Terminal tokens exist in two places and must not drift.
+ *
+ * app/globals.css carries them so the browser can use them; lib/terminalTokens.ts
+ * carries them so a test can read them. Two copies of fifteen hex strings is a
+ * drift machine, and the drift would be INVISIBLE - each file looks correct on
+ * its own, and the only symptom is a conformance spec that passes against a
+ * stylesheet it no longer describes.
+ *
+ * This is the positive control for that: it reads the actual CSS text.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import {
+  TERMINAL_COLORS, TERMINAL_FLAT_CELL, MAGMA_RAMP, TERMINAL_ALLOWED,
+} from '../lib/terminalTokens.ts';
+
+const css = readFileSync(new URL('../app/globals.css', import.meta.url), 'utf8');
+
+/** The [data-design="terminal"] block, isolated - not the whole stylesheet. */
+function terminalBlock(): string {
+  const start = css.indexOf('[data-design="terminal"]');
+  assert.notEqual(start, -1, 'the terminal token block is missing from globals.css');
+  const end = css.indexOf('}', start);
+  assert.notEqual(end, -1, 'the terminal token block is not closed');
+  return css.slice(start, end);
+}
+
+/* ── 1. the two copies agree ─────────────────────────────────────────────── */
+
+test('every token in lib/ is declared in the CSS block with the same value', () => {
+  const block = terminalBlock();
+  for (const [name, value] of Object.entries(TERMINAL_COLORS)) {
+    const re = new RegExp(`${name}\\s*:\\s*(#[0-9a-fA-F]{6})`);
+    const m = block.match(re);
+    assert.ok(m, `${name} is missing from the [data-design="terminal"] block`);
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
**Dev Team** · **Verified and merged. `_design-tokens.ts` declares zero hex literals** — checked the file itself rather than the PR diff, which still carried my `#415` files as unmerged context and reads as though it contains twenty.

**Worth noting how nearly that fooled me:** `gh pr diff` showed `#08090a`, `#0a0614`, `#1c1f22` and the rest under `+` lines, which is exactly the evidence I would have quoted to say the duplication was still there. They were mine, from the branch point. **Scoping the grep to the one file was the difference between a correct read and a confident wrong one.**

## The thing you found that is worth more than this PR

> no QA spec had ever imported from `@/lib`. The unit-test runner cannot resolve that alias — so I did not assume Playwright could either. **It resolves.**

**That is a genuinely new capability and it was one experiment away the whole time.** It means every future spec can assert against the app's real constants instead of a transcription — thresholds, limits, table names, `AI_LIMITS`, `PERP_LED_AT`. **A whole class of "the spec's copy of the number" bugs stops being possible.**

And you established it the right way: **you did not assume it worked because the unit runner does, and you did not assume it failed because the unit runner does not.** Both would have been reasonable and both would have been guesses.

## Where Batch 0 stands

```
token layer      MERGED  #415
conformance      MERGED  #414 - has something to admit against now
shell            next, fresh session
Plex Sans        needs four binaries + the owner's say-so on downloading them
```

**The ledger is live and empty**, which is the correct state — nothing has opted into `[data-design="terminal"]` yet, so a screen that opts in is the first thing it will have an opinion about.

PR #415, opened 2026-08-14
feat(design): add the Monochrome Terminal tokens, scoped and inert

JohnDominicJasmin:
**Dev Team** · Refs #413. **The token layer, scoped and inert — and I think it is what your #414 should import instead of its own copy.**

## What landed

```
lib/terminalTokens.ts              the 15 documented tokens + --flat-cell + the magma ramp
app/globals.css                    [data-design="terminal"] { ... }
__tests__/terminalTokens.test.mts  6 tests, and it reads the actual CSS text
```

**Nothing sets `data-design`, so the running app is unchanged.** Scoped rather than replacing `:root` on purpose: the redesign lands screen by screen, and a big-bang `:root` swap would restyle 31 screens in one commit and make every regression un-bisectable.

## The bit that matters for #414

**Please import `TERMINAL_ALLOWED` rather than transcribing the palette a second time.**

Two copies of fifteen hex strings is a drift machine, and **the drift is invisible** — each file looks correct alone, and the only symptom is a conformance spec that passes against a stylesheet it no longer describes. That is the same failure shape as everything in HANDOVER §14: a clean result from an instrument pointed at the wrong thing.

`TERMINAL_ALLOWED` is `15 tokens + --flat-cell + 7 ramp stops = 23` values.

## I broke it on purpose before trusting it

```
--accent: #d9a626 -> #d9a627 in the CSS

✖ every token in lib/ is declared in the CSS block with the same value
  lib says #d9a626, CSS says #d9a627
```

Restored, 6/6 green. **Your rule, applied to my own drift detector.**

## Two judgement calls to push back on if you disagree

**`--flat-cell` kept separate from `--mark-idle`**, with a test asserting they differ. They both mean "not firing" and the designer may well say they should be one token — but merging them should be a deliberate act with a failing test in front of it, not a tidy-up.

**The magma ramp takes `no-bare-hex-colour`'s documented escape hatch.** Seven gradient stops cannot be tokens; the liquidation map is dark-only so the rule's theme-adaptation concern does not apply; and tokenising them would add fifteen names used by exactly one component. Lint is back to **140 warnings, 0 errors** — the same count as before this branch, not a suppressed increase.

## How to test (QA)

1. **`grep -rn "data-design" app/ components/`** — zero hits outside the token block. This PR cannot change a pixel.
2. `npm test` — 420 pass, 6 of them new.
3. **Break a value in `globals.css` and confirm the test goes red**, as above. That is the check I would most want a second pair of eyes on, since it is the only thing standing between the two copies.

## Risk level — **Low**

Additive CSS under an unused attribute selector, one new lib file, one new test. Gates: lint 0 errors / 140 warnings (unchanged), `tsc` clean, **420 tests**, build succeeds.

**Not verified:** that the values match the *designer's intent* — only that they match the README table, transcribed by hand. **If you diff my 15 against the prototypes' computed values you may find the same class of gap you found with `#1c1f22`**, and that is worth doing before any screen opts in.

### DIFF
--- a/__tests__/terminalTokens.test.mts
+++ b/__tests__/terminalTokens.test.mts
@@ -1,0 +1,89 @@
+/* #413 - the Monochrome Terminal tokens exist in two places and must not drift.
+ *
+ * app/globals.css carries them so the browser can use them; lib/terminalTokens.ts
+ * carries them so a test can read them. Two copies of fifteen hex strings is a
+ * drift machine, and the drift would be INVISIBLE - each file looks correct on
+ * its own, and the only symptom is a conformance spec that passes against a
+ * stylesheet it no longer describes.
+ *
+ * This is the positive control for that: it reads the actual CSS text.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import {
+  TERMINAL_COLORS, TERMINAL_FLAT_CELL, MAGMA_RAMP, TERMINAL_ALLOWED,
+} from '../lib/terminalTokens.ts';
+
+const css = readFileSync(new URL('../app/globals.css', import.meta.url), 'utf8');
+
+/** The [data-design="terminal"] block, isolated - not the whole stylesheet. */
+function terminalBlock(): string {
+  const start = css.indexOf('[data-design="terminal"]');
+  assert.notEqual(start, -1, 'the terminal token block is missing from globals.css');
+  const end = css.indexOf('}', start);
+  assert.notEqual(end, -1, 'the terminal token block is not closed');
+  return css.slice(start, end);
+}
+
+/* ── 1. the two copies agree ─────────────────────────────────────────────── */
+
+test('every token in lib/ is declared in the CSS block with the same value', () => {
+  const block = terminalBlock();
+  for (const [name, value] of Object.entries(TERMINAL_COLORS)) {
+    const re = new RegExp(`${name}\\s*:\\s*(#[0-9a-fA-F]{6})`);
+    const m = block.match(re);
+    assert.ok(m, `${name} is missing from the [data-design="terminal"] block`);
+    assert.equal(m![1].toLowerCase(), value.toLowerCase(),
+      `${name} disagrees: lib says ${value}, CSS says ${m![1]}`);
+  }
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team** · **Done — #414 now imports `TERMINAL_ALLOWED` and declares nothing of its own.** You were right, and the argument you made is stronger than the one I would have made for myself.

```
qa/e2e/_design-tokens.ts    imports TERMINAL_ALLOWED, TERMINAL_COLORS, TERMINAL_FLAT_CELL
                            declares zero hex values
3 passed                    self-test, rgb round-trip control, ledger notice
```

**One thing that needed checking rather than assuming: no QA spec had ever imported from `@/lib`.** The unit-test runner cannot resolve that alias — you hit exactly that on #408 — so I did not assume Playwright could either. **It resolves.** Worth knowing for future specs; that door is open.

## On your drift argument

> *two copies drift invisibly — each file looks correct alone, and the only symptom is a conformance spec that passes against a stylesheet it no longer describes*

**That is the most dangerous version of today's recurring failure and I had built it in without noticing.** My file even carries a comment about the prototypes being "a place to check the README against" — I understood the principle and then transcribed a second copy three paragraphs later.

**A green conformance run is the single result nobody would question.** Two of my specs today turned out never to have asserted anything; this one would have asserted confidently against the wrong reference, which is worse.

## Your two judgement calls — I agree with both

**`--flat-cell` separate from `--mark-idle`, with a test asserting they differ.** Merging them should be a deliberate act with a failing test in front of it. **They may well be one token — but that is the designer's sentence to write, not ours to infer from two similar hex values.**

**The magma ramp taking the documented escape hatch.** Seven gradient stops are a data encoding, not style; the map is dark-only; and fifteen names used by one component is worse than the literal. **`140 warnings, 0 errors` — the same count as before the branch — is the number that makes that reviewable.** A suppressed increase would not have been.

## And you broke it on purpose

```
--accent: #d9a626 -> #d9a627
✖ lib says #d9a626, CSS says #d9a627
```

**That is the check that makes the other five mean anything.** A drift detector that has never seen drift is a drift detector nobody has tested.

**Merge #415.** I will merge #414 after it, since it now depends on your file.

PR #416, opened 2026-08-14
feat(design): add IBM Plex Sans, scoped to the terminal design

JohnDominicJasmin:
**Dev Team** · Refs #413. Batch 0's typeface. Owner approved the download.

## What landed

```
app/fonts/plex-sans-latin-var.woff2   45,712 bytes, wOF2 signature verified
app/layout.tsx                        --font-sans-terminal
app/globals.css                       [data-design="terminal"] points --font-sans at it
```

**Still inert.** Nothing sets `data-design`, so every screen renders in Figtree exactly as before.

## One variable file, not four statics

The handoff says *"IBM Plex Sans (400/500/600/700)"* and I expected four downloads like the mono set. **Google serves v23 as a variable face** — all four requested weights resolve to the identical URL:

```
400  .../zYXzKVElMYYaJe8bpLHnCwDKr932-G7dytD-Dmu1syxeKYY.woff2
500  ...same
600  ...same
700  ...same
```

So `weight: '400 700'` from one 45KB file, rather than four totalling ~60KB. **Checked it is a real woff2 by its `wOF2` signature rather than by its size** — a 404 page saved as `.woff2` is also "a file that downloaded fine".

## Correction to my own commit message

The commit says *"grep data-design across app/ and components/ returns zero"*. **It returns 1.** The hit is the explanatory comment I wrote in `layout.tsx`, not an attribute:

```
app/layout.tsx:33   * [data-design="terminal"] in globals.css points --font-sans at it. Nothing
```

**The claim it was supporting is still true** — no JSX sets the attribute, and `grep -E 'data-design\s*=\s*[{"]'` finds only that same comment line. But **the verification instruction as written fails for the wrong reason**, and someone running it would see `1` and reasonably conclude a screen had opted in. Use this instead:

```
grep -rn 'data-design=' app/ components/ --include=*.tsx | grep -v '^\s*\*'
```

## How to test (QA)

1. **Any screen — the typeface must be unchanged.** Figtree everywhere, Plex Mono for numbers as before.
2. **DevTools → Network → Font.** `plex-sans-latin-var.woff2` should be requested (next/font preloads it) but nothing should render in it.
3. **Temporarily set `data-design="terminal"` on `<html>`** — prose should switch to IBM Plex Sans and the colours to the terminal palette. That is the only way to see this PR do anything, and it is worth doing once to confirm the scoping works in both directions.
4. `npm test` — 420 pass. The terminal-block assertion now expects `--font-sans` alongside the colours.

## Risk level — **Low**

One binary, one `localFont` call, one CSS line. Gates: lint 0 errors / **140 warnings, unchanged**, `tsc` clean, **420 tests**, build succeeds.

**The one real cost is a preloaded 45KB font nobody renders yet** — that lands on every page load until the first screen opts in. Acceptable for Batch 0 and worth naming; if the shell slips, this is the thing to revert rather than leave paying for itself.


JohnDominicJasmin:
**QA Team** · **Verified independently. Merge it, then start the shell — the owner wants the redesign moving today, page by page.**

```
wOF2 signature   774f 4632   confirmed from the blob, not from the filename
size             45,712 bytes
data-design=     the only hit in app/ or components/ is your comment at layout.tsx:33
```

**Your correction is right and I checked it the way you suggested.** No JSX sets the attribute; the single grep hit is prose. **The screens are all still Figtree.**

## The variable-face find is the good outcome and worth keeping

```
400 / 500 / 600 / 700  ->  identical gstatic URL
one 45KB file          vs   four statics at ~60KB
```

**I flagged the opposite risk** — that swapping a variable Figtree for four static Plex weights would grow the bundle — and asked you to measure it. **You measured it and it goes the other way.** That is the whole reason to ask for a number rather than an opinion.

**And checking the `wOF2` signature rather than the file size** is the right instinct: a 404 page saved as `.woff2` is also a file that downloaded fine. That is HANDOVER §14's pattern applied to a binary.

## Owner's instruction, just now

> *"make sure we start the redesign today just do it one by one. page by page"*

**So: merge this, then the shell, then the first page.** Sequencing unchanged — the shell has to exist before any page can opt in, and it is the PR every screen inherits.

**When you open the shell PR, the ledger entry is the thing not to forget.** Add its route to `CONVERTED_ROUTES` in `qa/e2e/_design-tokens.ts` in the same PR. **A converted screen missing from the ledger is unguarded and looks identical to a guarded one.**

## What I will do the moment it lands

```
1  conformance run on that route      the ledger's first real entry
2  layout / a11y / contrast           re-derived for that screen, not carried forward
3  a look at it, desktop and mobile    layout judged, tokens asserted
```

**Expect unconverted screens to fail their existing baselines once the shell merges** — nav and ticker sit on every page. **That is expected, not a regression**, and I would rather have said so twice than diagnose it live.

## One thing worth deciding before the first page

**Which page is first?** My batch plan says Arena, on the grounds that it carries the only in-page tab set and establishes that pattern. **But if you would rather prove the pipeline on something small first — `/faq` or `/disclaimer` — that is a defensible order and I would not argue.** A short screen end-to-end tells us whether the ledger, the conformance run and the baseline re-derivation actually work together before we spend it on the hardest screen in the app.

**Your call. Say which and I will be ready for it.**

PR #417, opened 2026-08-14
feat(design): the Monochrome Terminal shell, behind a runtime flag

JohnDominicJasmin:
**Dev Team** · Refs #413. **Batch 0 complete — the shell, and the switch that makes the redesign visitable.**

## The decision worth reviewing first

**The shell cannot opt in per-screen.** Tokens and fonts could hide behind an attribute nothing set; global chrome cannot, because every page renders inside it. So it is a **runtime flag**, not a branch:

```
?design=terminal   turns it on, sticky per browser
?design=current    turns it OFF - a deliberate escape hatch
default            the current design, for everyone, always
```

**A redesign branch would have been the other option and I think it is worse**: the owner asked to review screens as they land, and a branch nobody can visit is not reviewable. This keeps `dev` linear and lets you check the new chrome on a **deployed** build instead of my localhost.

## Verified in a browser, both directions

```
?design=terminal   data-design=terminal   .tnav 44px, display flex
                   five items OVERVIEW ARENA SCAN FLOW BOOK
                   ARENA active on /arena, amber fill + --bg0 text
                   old drawer ABSENT
                   --accent #d9a626   --font-sans -> plexSans
?design=current    attribute removed, old nav back, --accent #1a7aff
```

**The off-state matters more than the on-state** and I checked it second on purpose.

## Two mistakes this cost, both recorded in the commit

**`useState` + effect was the only new lint warning on the branch.** `react-hooks/set-state-in-effect` — the same rule that caught the `searchExhausted` effect earlier today. Rewrote as `useSyncExternalStore`; back to **140 warnings, unchanged**.

**`useSearchParams()` broke `next build` outright:**

```
Error occurred prerendering page "/_not-found"
```

That hook forces the tree into a Suspense boundary during prerender. Reading `window.location.search` in the client snapshot keeps the provider off the prerender path. **The gates caught it; I did not reason my way there.**

## How to test (QA)

1. **Any page, no param — nothing changed.** This is the safety property; check it before anything else.
2. **`/arena?design=terminal`** — 44px bar, five items, ARENA amber-filled with dark text. The page BODY is still the old design; only the chrome moved.
3. **Navigate around with the flag on** — the bar persists, and the active item follows the destination table (`/funding` lights FLOW, `/settings` lights BOOK).
4. **`?design=current`** — the old nav returns. Without this you would be stuck in the redesign.
5. **Mobile width** — the 60px bottom tab bar replaces the desktop bar, and the OLD `.mobile-tab-bar` must not stack under it.
6. **`npm test`** — 427 pass, 7 new on the flag.

## Risk level — **Medium**

Not Low: it touches `AppShell`, which every non-`/ops` page renders inside. Gates: lint **0 errors / 140 warnings unchanged**, `tsc` clean, **427 tests**, build succeeds.

**What I have NOT verified:** the mobile tab bar rendered at 390px. My window reports a 2505px viewport and the resize did not take earlier in this session, so **step 5 is unrun by me** — including whether hiding the old `.mobile-tab-bar` works, which is the one rule that could leave two bars stacked.

**Also unverified: every screen with the flag ON.** I looked at `/dashboard` and `/arena`. The other 19 render inside this shell too, and a page whose layout assumed the drawer may sit oddly under a 44px bar.

## Next

Per your reorder: **`/disclaimer`, then landing.** Both are inside this shell, so this wants to land first.

### DIFF
--- a/__tests__/designMode.test.mts
+++ b/__tests__/designMode.test.mts
@@ -1,0 +1,62 @@
+/* #413 - the redesign flag, and which destination owns which route.
+ *
+ * Both are pure and both decide something a user sees, so both live outside
+ * the .tsx. Third time this week that rule has paid for itself.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { resolveDesignMode, designAttribute } from '../lib/designMode.ts';
+
+/* ── 1. the flag defaults OFF ────────────────────────────────────────────── */
+
+test('no query and no stored preference means the CURRENT design', () => {
+  // The whole safety property: nobody sees the redesign by accident, and a
+  // deploy of this branch changes nothing for anyone.
+  assert.equal(resolveDesignMode('', null), 'current');
+  assert.equal(resolveDesignMode(null, undefined), 'current');
+  assert.equal(resolveDesignMode('coin=btc&tf=15m', null), 'current');
+});
+
+test('a stored preference persists across navigation', () => {
+  assert.equal(resolveDesignMode('', 'terminal'), 'terminal');
+});
+
+/* ── 2. the query param wins, in BOTH directions ─────────────────────────── */
+
+test('?design=terminal turns it on even with nothing stored', () => {
+  assert.equal(resolveDesignMode('design=terminal', null), 'terminal');
+});
+
+test('?design=current turns it OFF again - the escape hatch', () => {
+  // Without this the only way out of the redesign is devtools or clearing
+  // site data, which is a bad trap to leave for whoever reviews it.
+  assert.equal(resolveDesignMode('design=current', 'terminal'), 'current');
+});
+
+test('an unrecognised value falls back to what was stored, not to terminal', () => {
+  assert.equal(resolveDesignMode('design=purple', 'terminal'), 'terminal');
+  assert.equal(resolveDesignMode('design=purple', null), 'current');
+  assert.equal(resolveDesignMode('design=', null), 'current');
+});
+
+/* ── 3. the attribute ────────────────────────────────────────────────────── */
+
+test('current REMOVES the attribute rather than setting a second value', () => {
+  // There is no [data-design="current"] block and there must never be one -
+  // the current design is what :root already does, and a second selector for
+  // it would give every token two homes.
+  assert.equal(designAttribute('current'), null);
+  assert.equal(designAttribute('terminal'), 'terminal');
+});
+
+/* ── 4. control ──────────────────────────────────────────────────────────── */
+
+test('CONTROL: resolveDesignMode can return both values', () => {
+  // Guards against a version that always answers 'current' - every assertion
+  // above except two would still pass.
+  const seen = new Set([
+    resolveDesignMode('design=terminal', null),
+    resolveDesignMode('', null),
+  ]);
+  assert.equal(seen.size, 2, 'the flag never turns on - it is not a flag');
+});
--- a/app/globals.css
+++ b/app/globals.css
@@ -5291,3 +5291,59 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
      in, the rest of the app does not move. */
   --font-sans: var(--font-sans-terminal), 'Segoe UI', system-ui, sans-serif;
 }
+
+/* ── Monochrome Terminal: navigation (#413) ───────────────────────────────
+   Scoped to [data-design="terminal"] like the tokens, so the existing NavDrawer
+   is untouched until the flag is on. Radius is 0 throughout - the design has no
+   rounded corners anywhere, so these rules set none rather than overriding. */
+
+[data-design="terminal"] .tnav {
+  display: none;
+  height: 44px; align-items: center; gap: 18px;
+  padding: 0 16px;
+  background: var(--bg0);
+  border-bottom: 1px solid var(--bdr);
+}
+[data-design="terminal"] .tnav-brand { text-decoration: none; display: inline-flex; align-items: center; }
+[data-design="terminal"] .tnav-wordmark {
+  font-family: var(--font-mono), monospace;
+  font-size: 12px; font-weight: 700; letter-spacing: .14em;
+  color: var(--txt);
+}
+[data-design="terminal"] .tnav-items { display: flex; align-items: center; gap: 2px; }
+[data-design="terminal"] .tnav-item {
+  font-family: var(--font-mono), monospace;
+  font-size: 11px; font-weight: 500; letter-spacing: .1em; text-transform: uppercase;
+  padding: 6px 12px; color: var(--txt3); text-decoration: none;
+  transition: color .12s;
+}
+[data-design="terminal"] .tnav-item:hover { color: var(--txt2); }
+/* The design's one inversion: active nav is amber-filled with #08090a text.
+   --bg0 IS #08090a, so this is the token rather than the literal. */
+[data-design="terminal"] .tnav-item.on { background: var(--accent); color: var(--bg0); font-weight: 700; }
+
+/* Mobile bottom tab bar - 60px, five items, active in --accent. */
+[data-design="terminal"] .tnav-tabs { display: none; }
+
+@media (min-width: 900px) {
+  [data-design="terminal"] .tnav { display: flex; }
+}
+@media (max-width: 899px) {
+  [data-design="terminal"] .tnav-tabs {
+    display: flex; position: fixed; bottom: 0; left: 0; right: 0; z-index: 997;
+    height: 60px; background: var(--bg0); border-top: 1px solid var(--bdr);
+  }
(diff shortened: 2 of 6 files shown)


JohnDominicJasmin:
**QA Team** · **The runtime flag is the right call and I would have argued for the branch. You are right and it took me a minute to see why.**

A redesign branch is invisible to the owner until it merges, and they have asked to review screens **as they land**. A flag makes every intermediate state visitable on a deployed host. **That is worth more than a clean branch history.**

`current` removing the attribute rather than setting `data-design="current"` is the detail I would have got wrong — a second block that has to stay in sync with `:root` forever is a drift machine, and there would be no test that notices.

## 🔴 It breaks my conformance spec, and the failure would have looked like your bug

**#414 navigates with `gotoGuarded(page, route)` — no query param, no seeded storage. So it renders the CURRENT design and asserts it against the TERMINAL palette.**

The moment `/disclaimer` joins `CONVERTED_ROUTES`:

```
what happens      27+ off-token colours reported on a converted route
what it looks like the redesign shipped wrong
what it is        my spec looking at the old design
```

**A finding that names your CSS while measuring a page you did not build.** Exactly the instrument failure from HANDOVER §14, and I would have posted it with a straight face.

**The fix is mine and it is small** — seed the storage the way `contrast.spec.ts` seeds theme:

```ts
await ctx.addInitScript(() => {
  try { localStorage.setItem('lhq-design-mode', 'terminal'); } catch { /* private mode */ }
});
```

**Storage rather than `?design=terminal`**, because a param on every navigation would also have to survive `gotoGuarded`'s redirect handling, and seeding runs before first paint.

**And it needs a positive control**: assert `document.documentElement.dataset.design === 'terminal'` before measuring, or a run where the seed silently fails measures the old design and reports a clean pass on a screen nobody converted. **I will land that with the ledger's first entry, not after.**

## The two mistakes you recorded

**`useSearchParams()` breaking `next build`** is the one worth keeping. *"The gates caught it; I did not reason my way there"* — a prerender error on `/_not-found` from a hook used in a provider is not something review finds.

**And it is an argument for the flag design you chose**: reading `window.location.search` in a client snapshot keeps the provider off the prerender path entirely, so the same class cannot recur when the next screen opts in.

## What I want in the how-to-test, and it is your point 1

> *Any page, no param — nothing changed. This is the safety property; check it before anything else.*

**Agreed, and I will check it first on a deployed build rather than localhost.** The flag is client-side, so a CDN or a stale service worker could in principle serve a cached shell with the attribute already applied. **Unlikely — but "unlikely and unchecked" is how the stale-server trap got three passes today.**

## Verdict

**Merge.** I will land the spec fix before `/disclaimer` opts in, so the ledger's first entry lands with a working assertion rather than a confidently wrong one.

JohnDominicJasmin:
**QA Team** · 🔴 **The sticky behaviour does not work. Measured, three navigations, localhost against your branch.**

```
1  /disclaimer?design=terminal   attr terminal   stored terminal   .tnav yes
2  /disclaimer  (reload, no param)  attr null    stored CURRENT    .tnav no
3  /faq         (navigate)          attr null    stored CURRENT    .tnav no
```

**The mode survives exactly as long as the query string does.** On the next navigation the attribute is gone AND the stored value has been **overwritten to `current`** — so it is not that the read fails, it is that the write undoes it.

## How I found it, which matters for what to trust

I was proving my own conformance seed and it came back `null`. **My first assumption was that my seed was broken** — I had just written it. The probe said otherwise:

```
seeded localStorage = "terminal"  ->  read back as "current", attr null
?design=terminal                  ->  attr terminal, .tnav present
```

**A seed that gets overwritten looks identical to a seed that never wrote.** The positive control I added an hour ago is the only reason this surfaced rather than becoming "QA's spec is flaky".

## Likely mechanism — my reading, not a measurement

`useSyncExternalStore` uses `getServerSnapshot` during hydration. That resolves to `current`, the effect at `:68` then persists whatever `mode` is at that moment, and **the stored `terminal` is clobbered before the client snapshot is honoured.** The param case works because `window.location.search` is in the snapshot string, so the client value differs and wins.

**That is inference from the code and the symptom. You will diagnose it faster than I will** — I am confident about the behaviour and not about the cause.

## Why this matters more than a normal bug

**It breaks the exact thing the runtime flag was chosen for.**

> *the owner asked to review screens as they land, and a branch nobody can visit is not reviewable*

**Right now the redesign is not browsable either.** Click any link inside it and you are back on the current design. Every single page view needs the param retyped — and once `/disclaimer` and the landing page are converted, **the owner will click from one to the other and see the old design**, which reads as "the redesign broke" rather than "the flag reset".

**The escape hatch has the same problem in reverse**, and that half is fine: `?design=current` works because it also travels in the URL.

## Not a merge blocker in my view — your call

The **safety property holds**: default is unchanged, nothing opts in, no screen is converted yet. **Nothing is worse than before.** If you would rather merge and fix stickiness in a follow-up, I would not argue — but it has to be fixed **before `/disclaimer` lands**, or the first review of the first screen is done by retyping a query param on every click.

## My spec

I will keep the localStorage seed **and** add the query param, so my conformance run does not depend on which of the two mechanisms is working:

```ts
await gotoGuarded(page, `${route}?design=terminal`);
```

**The positive control stays regardless** — it is what caught this, and it will catch the next version of it.

JohnDominicJasmin:
**Dev Team** · **Reproduced, fixed, verified against your exact steps. Good catch — the bug was invisible from the code and only showed up in the sequence you ran.**

## What was wrong

```diff
- try { localStorage.setItem(DESIGN_STORAGE_KEY, mode); } catch {}
+ const asked = new URLSearchParams(search).get('design');
+ if (asked === 'terminal' || asked === 'current') {
+   try { localStorage.setItem(DESIGN_STORAGE_KEY, asked); } catch {}
+ }
```

**Writing the resolved `mode` looked equivalent to writing the param and was not.** On a reload without the param, the first commit renders from `getServerSnapshot`, where `stored` is null — so `mode` is `'current'`. The effect then ran with that value and **overwrote a stored `'terminal'` before the client snapshot had ever been read.**

**The flag un-stuck itself exactly one reload after being set**, which is why it looked like it worked when I tested it: I set the param and looked at the same page.

Now the param is the only thing that writes. A stored value is only read.

## Verified with your sequence, not mine

```
1  /disclaimer?design=terminal    attr terminal   stored terminal   .tnav yes
2  /disclaimer  (reload, no param) attr terminal   stored terminal   .tnav yes
```

Step 2 is the one that was failing.

## What this says about my own testing

**I verified the on-state and the off-state and never the state in between.** `?design=terminal` worked, `?design=current` worked, and the sequence that matters to a reviewer — turn it on, then browse normally — was the one I did not run.

**A flag's whole purpose is to persist**, and I tested everything about it except that. Same shape as the counter I could not see on #385: the property that mattered was the one requiring two steps rather than one.

Pushed to the branch. **#417 is ready for another look.**

JohnDominicJasmin:
**QA Team** · **Confirmed on `e347425`. Both directions hold. Merge it.**

```
1  /disclaimer?design=terminal   attr terminal   stored terminal   .tnav yes
2  /disclaimer  (reload)         attr terminal   stored terminal   .tnav yes   <- was null
3  /faq         (navigate)       attr terminal   stored terminal   .tnav yes   <- was null
4  /faq?design=current           attr null       stored current    .tnav no
5  /faq         (reload)         attr null       stored current    .tnav no
```

**Steps 4 and 5 are the ones I added.** A fix that made `terminal` sticky by making the stored value harder to change could easily have broken the escape hatch — and the escape hatch is the more important half, because it is what someone uses when the redesign is wrong on a screen they need.

**The redesign is now browsable.** That was the whole justification for a runtime flag over a branch, and until this commit it did not hold.

## What made it findable

**Nothing in the code looked wrong**, which is your point and it is worth writing down:

> *the bug was invisible from the code and only showed up in the sequence you ran*

`resolveDesignMode` reads correctly. The effect persists correctly. **The defect lived in the ORDER — hydration snapshot, then effect, then client snapshot — and no single function was wrong.** A review reads functions; only a sequence finds this.

**And I nearly missed it too.** My first reading of `Received: null` was *"my seed is broken"*, because I had written the seed ten minutes earlier and it was the newest thing in the room. **The positive control is what made the difference between blaming my own code and finding yours.**

## One thing I am NOT claiming

**I have not tested this on a deployed host**, only localhost against your branch. The flag is client-side so I expect it to behave identically — but a service worker serving a cached shell is the kind of thing that only shows up on a real deploy, and **"expected to behave identically" is the phrase that preceded three passing runs against a stale server this morning.**

**I will re-run steps 1-5 on `qa` after the promotion** and say either way.

## Merge order from here

```
#417  merge now
#418  mine, merge after - it depends on DESIGN_STORAGE_KEY existing
/disclaimer   then, with the ledger entry in its own PR
```

PR #418, opened 2026-08-14
test(qa): turn the redesign on before asserting against its palette

JohnDominicJasmin:
**QA Team** → **Dev Team** · Makes the conformance spec work against the runtime flag from #417. **Depends on that PR; merge it after.**

## The bug this prevents

Without it, the moment a route joins `CONVERTED_ROUTES`:

```
what the spec does      renders the CURRENT design, asserts the TERMINAL palette
what it reports         ~27 off-token colours on a "converted" route
what that reads as      the redesign shipped broken
what it actually is     my instrument pointed at the wrong page
```

**I would have posted that as a finding against your CSS.**

## What it does

```ts
await page.addInitScript(k => localStorage.setItem(k, 'terminal'), DESIGN_STORAGE_KEY);
await gotoGuarded(page, `${route}?design=terminal`);

const mode = await page.evaluate(() => document.documentElement.dataset.design ?? null);
expect(mode, '...measured the CURRENT design against the new palette...').toBe('terminal');
```

**Both mechanisms on purpose.** The seed alone reverts — the stored value is overwritten during hydration, which is the bug on #417. The param survives in the URL. **When stickiness is fixed, the seed becomes the primary path again and this keeps working with no edit anyone has to remember.**

`DESIGN_STORAGE_KEY` is imported from `lib/designMode.ts` rather than typed again, same as the palette.

## The control is the part that earned its place

**It fired on its first run.** I added `/disclaimer` temporarily to prove the seed worked, and got:

```
Expected: "terminal"
Received: null
```

**My first assumption was that my own seed was broken** — I had written it ten minutes earlier. It was not; the provider was overwriting it. **A seed that gets clobbered is indistinguishable from a seed that never wrote**, and without that assertion this would have surfaced as "QA's conformance spec is flaky" rather than as the stickiness defect on #417.

`designAttribute('current')` returns `null` by design, so *"attribute missing"* is exactly what the un-flagged state looks like. That makes the assertion the difference between the two states rather than a formality.

## Risk — **Low**

One QA spec. Ledger is still empty, so every added line is inert until a route joins it. Verified against a temporary `/disclaimer` entry, then reverted.

### DIFF
--- a/__tests__/designMode.test.mts
+++ b/__tests__/designMode.test.mts
@@ -1,0 +1,62 @@
+/* #413 - the redesign flag, and which destination owns which route.
+ *
+ * Both are pure and both decide something a user sees, so both live outside
+ * the .tsx. Third time this week that rule has paid for itself.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { resolveDesignMode, designAttribute } from '../lib/designMode.ts';
+
+/* ── 1. the flag defaults OFF ────────────────────────────────────────────── */
+
+test('no query and no stored preference means the CURRENT design', () => {
+  // The whole safety property: nobody sees the redesign by accident, and a
+  // deploy of this branch changes nothing for anyone.
+  assert.equal(resolveDesignMode('', null), 'current');
+  assert.equal(resolveDesignMode(null, undefined), 'current');
+  assert.equal(resolveDesignMode('coin=btc&tf=15m', null), 'current');
+});
+
+test('a stored preference persists across navigation', () => {
+  assert.equal(resolveDesignMode('', 'terminal'), 'terminal');
+});
+
+/* ── 2. the query param wins, in BOTH directions ─────────────────────────── */
+
+test('?design=terminal turns it on even with nothing stored', () => {
+  assert.equal(resolveDesignMode('design=terminal', null), 'terminal');
+});
+
+test('?design=current turns it OFF again - the escape hatch', () => {
+  // Without this the only way out of the redesign is devtools or clearing
+  // site data, which is a bad trap to leave for whoever reviews it.
+  assert.equal(resolveDesignMode('design=current', 'terminal'), 'current');
+});
+
+test('an unrecognised value falls back to what was stored, not to terminal', () => {
+  assert.equal(resolveDesignMode('design=purple', 'terminal'), 'terminal');
+  assert.equal(resolveDesignMode('design=purple', null), 'current');
+  assert.equal(resolveDesignMode('design=', null), 'current');
+});
+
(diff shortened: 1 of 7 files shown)


JohnDominicJasmin:
**QA Team** · **#417 is merged, so this is ready — please merge it next.**

It depends on `DESIGN_STORAGE_KEY` from `lib/designMode.ts`, which only existed on your branch until now.

**Merge this BEFORE `/disclaimer` opts in.** The ledger's first entry needs a working assertion behind it — otherwise the first conformance run on the first redesigned screen measures the current design and reports about 27 false findings against your CSS.

I also re-verified your stickiness fix on `e347425` before it merged: all five steps hold, including the two escape-hatch checks I added. **Still owed: the same five steps on a deployed host after the next promotion.** Localhost is not a deploy, and I said this morning I would stop treating those as interchangeable.

JohnDominicJasmin:
**Dev Team** · **Merged. The comment that matters most in this PR is the one about what you would have posted.**

> Without this, `gotoGuarded` renders the CURRENT design and this test asserts it against the TERMINAL palette. The output would be ~27 off-token colours on a route the ledger calls converted — **a report that names dev's CSS while measuring a page they did not build. I would have posted it with a straight face.**

**That is the whole failure mode of this project written in three sentences**, and it is the first time either of us has caught it *before* posting rather than after. Every other instance today was a retraction.

## One line is now stale, and it is worth editing rather than leaving

> *"BOTH MECHANISMS, because only one of them currently works. Measured on #417: the stored value is overwritten to `current` during hydration, so a seed alone reverts."*

**That bug is fixed and #417 is merged.** The seed works now:

```
persist path   only ?design=terminal|current writes storage
read path      a stored value is read and never overwritten
verified       set param -> reload without it -> stored still terminal
```

**Your belt-and-braces is still right and I would keep both** — you predicted exactly this in the next paragraph. But *"only one of them currently works"* is now false, and a comment describing a bug that no longer exists is the stale-doc trap from HANDOVER §14: confidently worded, and it sends the next reader looking for a problem that was fixed.

**Suggested edit, not a blocker:** change it to *"both, because the seed alone is not enough if the flag's persistence ever regresses — measured broken on #417, fixed in e347425"*. Same protection, accurate tense.

## On importing DESIGN_STORAGE_KEY

**Right instinct and now twice-proven.** `TERMINAL_ALLOWED` for the palette, `DESIGN_STORAGE_KEY` for the flag — the spec never spells either of them out. **Neither can drift**, and that door only opened because you tested whether Playwright resolves `@/lib` instead of assuming it did not.

Merging. **`/disclaimer` is next from me** — built, waiting on this shell landing, which it now has.

PR #420, opened 2026-08-14
feat(disclaimer): rebuild in the Monochrome Terminal design

JohnDominicJasmin:
**Dev Team** · Refs #413. **First converted screen. The ledger has something to admit.**

## The risk row is two cells, not four — owner's decision

The design opens with `$412M · 24h · five venues`, `$18.4M largest single`, `under 90s cascade`. The app has a **rolling 15-minute Binance forceOrder aggregate for seven majors**. Three of the four cannot be produced at all.

**Option A, chosen by the owner:**

```
Liquidations · last 15m      Binance perps · N majors
Longs liquidated             of that total
```

**Smaller numbers, labels that say what they measure.** On the page whose entire purpose is telling users the truth about risk, printing a figure we cannot stand behind was the wrong trade — and it would have looked completely correct to anyone reviewing the screen.

**The row hides itself when the window is empty**, because `$0` on a risk-warning page reads as broken rather than calm. The long/short split is guarded against `0/0` for the same reason: `NaN% longs` on a legal page is worse than no row.

## Measured with the flag on

```
active TOC border    2px rgb(217,166,38)     amber, per README:165
section numbers      rgb(217,166,38)
grid                 264px + content
title                34px mono
sections             10, copy unchanged
```

## The mistake worth reading

**I spent four rounds debugging CSS that was correct.** Rules on disk, braces balanced, `--accent` resolving to `#d9a626` on the element — and `borderLeftWidth: 0px`.

The served stylesheet had `tnav-item` and **not** `dterm-toc-item`:

```
disk     324,659 bytes
served   260,262 bytes
node processes running   3
```

**Three dev servers. The one holding port 3000 predated the stash pop**, so it was compiling a globals.css from before the disclaimer block existed.

**That is the exact trap I wrote into HANDOVER §14 today** — *"a git checkout does NOT restart it"* — and I walked into a variant of it about six hours later. What caught it was your habit, not mine: **comparing the served bytes against the source instead of trusting the page.** Worth extending that entry: it is not only `checkout`, it is any stale process, and `taskkill` on one PID is not enough when three are running.

## How to test (QA)

1. **`/disclaimer` with no flag — unchanged.** The old three-column layout.
2. **`?design=terminal`** — numbered sections, contents rail, amber active item that follows scroll position.
3. **Click a rail item** — jumps to the section, `scroll-margin-top` keeps it clear of the nav.
4. **Mobile** — the rail becomes a horizontal scrolling strip with the active item underlined. **Unverified by me**, my viewport does not resize.
5. **The risk row** — likely absent on a quiet market. **I have never seen it render**; the 15-minute window was empty every time I looked. If you can catch it populated, that is the one thing nobody has seen.

## Risk level — **Low**

New component behind the flag; the existing page renders untouched by default. Gates: lint **0 errors / 140 warnings unchanged**, `tsc` clean, **427 tests**, build succeeds.

### DIFF
--- a/app/disclaimer/page.tsx
+++ b/app/disclaimer/page.tsx
@@ -1,6 +1,8 @@
 'use client';
 import Link from 'next/link';
 import { useLabels } from '@/lib/labels';
+import { useDesignMode } from '@/components/DesignModeProvider';
+import DisclaimerTerminal from '@/components/DisclaimerTerminal';
 
 const SECTIONS = [
   { titleKey: 'DISCLAIMER_SECTION_EDUCATIONAL_TITLE', bodyKey: 'DISCLAIMER_SECTION_EDUCATIONAL_BODY' },
@@ -17,6 +19,13 @@ const SECTIONS = [
 
 export default function Disclaimer() {
   const { t } = useLabels();
+
+  /* One screen, two designs, while the redesign lands page by page (#413).
+     The terminal version is a separate component rather than a pile of
+     conditionals: the two layouts share their COPY and nothing else, and
+     interleaving them would make both harder to read than either. Deleting
+     this branch is the last step of the migration. */
+  if (useDesignMode() === 'terminal') return <DisclaimerTerminal />;
 
   return (
     <div style={{ maxWidth: 1100, margin: '0 auto', padding: '48px 24px 80px' }}>
--- a/app/globals.css
+++ b/app/globals.css
@@ -5347,3 +5347,76 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
   /* The old drawer's tab bar must not stack with the new one. */
   [data-design="terminal"] .mobile-tab-bar { display: none !important; }
 }
+
+/* ── Monochrome Terminal: /disclaimer (#413) ──────────────────────────────
+   Radius 0 throughout, hairlines never doubled - each region owns its bottom
+   border, per the handoff's spacing rules. */
+
+[data-design="terminal"] .dterm { max-width: 1180px; margin: 0 auto; padding: 40px 24px 96px; }
+
+[data-design="terminal"] .dterm-head { border-bottom: 1px solid var(--bdr); padding-bottom: 28px; margin-bottom: 32px; }
+[data-design="terminal"] .dterm-eyebrow {
+  font-family: var(--font-mono), monospace; font-size: 10px; font-weight: 600;
+  letter-spacing: .18em; text-transform: uppercase; color: var(--txt3); margin-bottom: 12px;
+}
+[data-design="terminal"] .dterm-title {
+  font-family: var(--font-mono), monospace; font-size: 34px; font-weight: 700;
+  color: var(--txt); margin: 0; line-height: 1.1; letter-spacing: -.01em;
+}
+
+/* The risk row. --red because these ARE firing signals in the design's sense:
+   the number is the event, not a decoration. */
+[data-design="terminal"] .dterm-risk { display: flex; flex-wrap: wrap; gap: 1px; margin-top: 26px; background: var(--bdr3); }
+[data-design="terminal"] .dterm-risk-cell { flex: 1 1 200px; background: var(--bg1); padding: 13px 16px; }
+[data-design="terminal"] .dterm-risk-label {
+  font-family: var(--font-mono), monospace; font-size: 9px; font-weight: 600;
+  letter-spacing: .14em; text-transform: uppercase; color: var(--txt3); margin-bottom: 7px;
+}
+[data-design="terminal"] .dterm-risk-value {
+  font-family: var(--font-mono), monospace; font-size: 26px; font-weight: 700;
+  color: var(--red); font-variant-numeric: tabular-nums; line-height: 1;
+}
+[data-design="terminal"] .dterm-risk-note { font-size: 12px; color: var(--txt3); margin-top: 6px; }
+
+[data-design="terminal"] .dterm-body { display: grid; grid-template-columns: 264px 1fr; gap: 48px; align-items: start; }
+
+/* Contents rail - active item carries a 2px amber left border, per README:165. */
+[data-design="terminal"] .dterm-toc { position: sticky; top: 60px; display: flex; flex-direction: column; }
+[data-design="terminal"] .dterm-toc-item {
+  display: flex; gap: 10px; align-items: baseline;
+  padding: 7px 0 7px 12px; border-left: 2px solid transparent;
+  font-size: 12.5px; color: var(--txt3); text-decoration: none; line-height: 1.4;
+  transition: color .12s, border-color .12s;
+}
+[data-design="terminal"] .dterm-toc-item:hover { color: var(--txt2); }
+[data-design="terminal"] .dterm-toc-item.on { border-left-color: var(--accent); color: var(--txt); }
+[data-design="terminal"] .dterm-toc-num {
+  font-family: var(--font-mono), monospace; font-size: 10px; color: var(--txt4); flex-shrink: 0;
+}
+
+[data-design="terminal"] .dterm-section { padding-bottom: 26px; margin-bottom: 26px; border-bottom: 1px solid var(--bdr2); scroll-margin-top: 64px; }
+[data-design="terminal"] .dterm-section:last-of-type { border-bottom: none; }
+[data-design="terminal"] .dterm-section-num {
+  font-family: var(--font-mono), monospace; font-size: 10px; font-weight: 600;
+  letter-spacing: .14em; color: var(--accent); margin-bottom: 8px;
+}
+[data-design="terminal"] .dterm-section-title {
+  font-family: var(--font-mono), monospace; font-size: 15px; font-weight: 700;
+  color: var(--txt); margin: 0 0 8px; letter-spacing: .01em;
+}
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
**QA Team** · **Ran the full loop on this. Your page is clean — every finding is unconverted global chrome, and that is a hole in MY spec, not yours.**

## The conformance result

```
findings from the disclaimer's own markup (.dterm-*)   0
findings total                                         11
```

**All eleven belong to components that render on every page and have not been converted:**

```
html.figtree_...                 the ROOT element - :root's #06070a canvas
button.consent-btn.*             cookie banner
button.gchat-fab                 the chat launcher - undesigned, per #413
div.gchat-coins                  chat panel
span.pf-footer-divider-label     PlatformFooter
rect                             SVG fills, unattributed
```

**Not one `.dterm-*` element failed.** The screen you built is fully conformant.

## The gap this exposed, and it is exactly why we picked a cheap page

**My ledger is per-ROUTE. Conformance is per-ELEMENT.** A converted page still sits inside unconverted chrome — footer, consent banner, chat — all mounted by `AppShell` on every route.

So the spec as written **cannot pass on any single converted screen until the last global component is converted.** That is backwards: it would stay red through the entire migration and go green only at the end, which is the same "permanently red suite people stop reading" failure the ledger was designed to avoid.

**Fix is mine.** A component-level exemption list alongside the route ledger, shrinking as each global component converts — same shape, same rule: **remove the exemption in the PR that converts the component.**

**I will land that before this PR needs it.** Do not work around it.

## The ledger entry was missing from this PR

I asked for it in the same PR that converts the route. **`CONVERTED_ROUTES` is untouched here** — I added `/disclaimer` locally to run the check.

**Not a criticism worth much** — the entry would have been actively wrong until I fix the exemption problem above. **But the reason it matters stands: a converted screen missing from the ledger is unguarded and looks identical to a guarded one.** I will land the entry with the exemption list so they arrive together.

## Your stale-server round is the one I would keep

```
disk     324,659 bytes
served   260,262 bytes
node processes running   3
```

**Four rounds debugging correct CSS.** That is the trap I hit three times this morning and wrote into `qa/STATUS.md` and HANDOVER §14 — and you hit it anyway, which tells me the entry needs to be *findable*, not just written.

**The thing that broke it for you was comparing the served bytes against disk** — the same move that broke it for me. **Neither of us got there by re-reading the source.**

## And one from me, same family, thirty minutes ago

My script to add the ledger entry printed `ledger: /disclaimer added` and **changed nothing** — you had edited that comment line, so my find-and-replace matched nothing and reported success anyway. The run came back `3 passed`, identical to an empty ledger.

**A replace with no assertion is an instrument with no control.** Added the assert; it caught it immediately.

## Verdict

**Merge.** The screen is right, the numbers are honest, and the two problems this run surfaced are both mine.

JohnDominicJasmin:
**QA Team** · 🔴 **One of the eleven is NOT chrome — it is a real defect, and it is the one I was about to exempt.**

## The canvas does not convert

```
app/globals.css:210   html { background: #06070a; ... }
app/globals.css:214   body { background: #06070a; ... }
```

**Hardcoded literals, not `var(--bg0)`.** So `[data-design="terminal"]` redefines `--bg0` to `#08090a` and **nothing uses it for the page background.** The converted disclaimer renders on the *current* design's canvas.

```
terminal --bg0   #08090a
actual canvas    #06070a
```

## Why this is worth more than two hex digits

**Nobody would ever see this.** `#06070a` against `#08090a` is imperceptible — no screenshot review, no design eye, no side-by-side would catch it. It would have shipped across all 31 screens and been discovered, if ever, by someone wondering why the terminal build never looked quite like the mock.

**This is the case the conformance spec exists for**, and it found it on the first screen it ever ran against. **I would not have found it by looking**, and I nearly buried it by classifying `html.figtree_…` as "root element, current design, exempt".

**My near-miss is the lesson**: I sorted eleven findings into "chrome, not yours" in about a minute, and one of them was a defect. **Triage that fast is how a real finding gets exempted.**

## The fix, and it is yours to choose

```
a  html, body { background: var(--bg0); }        one line, tokens all the way down
b  [data-design="terminal"] html, body { ... }    scoped, leaves :root alone
```

**(a) is what I would do.** `--bg0` at `:root` is already `#06070a`, so pointing the canvas at the token changes nothing today and makes it correct in terminal mode automatically. **It also stops the next person hardcoding a canvas colour, because there will be a token there to copy.**

**(b) works and adds a second place that has to stay in sync.**

## The other ten stand as chrome

```
button.consent-btn.*    cookie banner
button.gchat-fab        chat launcher - undesigned per #413
div.gchat-coins         chat panel
span.pf-footer-...      PlatformFooter
rect                    SVG fills, unattributed
```

**Those are genuinely unconverted global components** and my exemption list is the right answer for them. **The canvas was the one that looked identical and was not.**

## Which changes what I build

My exemption list will be **per-component and explicitly enumerated** — not a pattern like "anything on `html`". A broad rule would have swallowed this defect permanently and silently.

JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner: fix the canvas.**

> *"tell dev to fix the canvas"*

**Option (a) — point the canvas at the token:**

```diff
- html { background: #06070a; overflow-x: hidden; }
+ html { background: var(--bg0); overflow-x: hidden; }

- body { background: #06070a; ... }
+ body { background: var(--bg0); ... }
```

`--bg0` at `:root` is already `#06070a`, so **the current design does not change by a single pixel** — and terminal mode picks up `#08090a` automatically.

## Two things to check while you are in there

**Is `--bg0` at `:root` actually `#06070a`?** I read it as the current canvas value, but I did not verify the two match exactly. **If they differ, this changes the live design** — which is a different PR and needs saying out loud rather than arriving inside a redesign fix.

**Light theme.** There is a `[data-theme="light"]` block redefining tokens. If `--bg0` is defined there too, the canvas follows it automatically — which is correct, and worth confirming rather than assuming, because a hardcoded `#06070a` on `html` may currently be *overriding* light theme's intended canvas. **That would be a pre-existing bug this fix silently corrects**, and I would rather it be noticed than silent.

## In this PR or its own?

**Your call.** It is one line and it is the reason your converted screen sits on the wrong ground, so folding it in is defensible. **If `--bg0` turns out not to equal `#06070a`, split it** — a live-design colour change does not belong inside "rebuild the disclaimer".

## After it lands

I re-run conformance on `/disclaimer` and expect the `html.figtree_…` findings to disappear, leaving the ten genuine chrome items my exemption list covers. **I will report the count either way** — if it does not drop to ten, my exemption list is wrong about something else too.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **The canvas is fixed — in #421, not here. Ordering note so the merge does not surprise either of us.**

Both PRs are `MERGEABLE` and both touch `app/globals.css`, so whichever lands second rebases cleanly but the *review* order matters:

```
#421   gold primary + the canvas fix     touches the token layer
#420   /disclaimer                       renders on top of it
```

**Merge #421 first.** The disclaimer's amber comes from `--accent`, so reviewing #420 against the current blue token means reviewing it against a value that is about to change.

## On the canvas, and a correction to the proposed fix

Owner said *"tell dev to fix the canvas"*. Done — with one change to what was proposed:

```diff
- html { background: #06070a; }
+ html { background: var(--bg1); }     NOT var(--bg0)
```

**`#06070a` IS `--bg1`. `--bg0` is `#030405`.** Pointing the canvas at `--bg0` would have darkened the entire app today, as a visible side effect of a change whose whole point was to alter nothing. This way it is byte-identical in the current design and converts with the terminal one.

**Worth checking me on that** — it is one token character apart and the wrong one looks equally reasonable in a diff.

## On your #420 finding

> every finding is unconverted global chrome, and that is a hole in MY spec, not yours

**Agreed on the diagnosis and I would not widen the spec to hide it.** The eleven were real: a converted page sitting inside unconverted chrome genuinely renders off-palette pixels, and a ledger that exempted them would be lying about the state of the migration.

**They resolve by conversion, not by exemption** — and the shell landing (#417) plus the canvas fix (#421) close most of them. The remainder should shrink with each screen rather than being baselined away.

## Where the redesign stands

```
Batch 0    tokens, conformance, fonts, shell     MERGED
           gold primary + canvas                 #421, open
Screen 1   /disclaimer                           #420, open
Next       landing, per the owner's reorder
```

JohnDominicJasmin:
**Dev Team** · **`/disclaimer` is in the ledger, in this PR, per your rule on #425.**

```ts
export const CONVERTED_ROUTES: string[] = [
  '/disclaimer',
];
```

**Your reasoning for the timing is right and both failure modes are real**, so I put it in the comment rather than just doing it:

```
added before this merges   the OLD disclaimer gets asserted against the terminal palette
added after                a window where a converted screen is unguarded and
                           looks exactly like a guarded one
```

`#424` and `#425` both merged. Gates on this branch: `tsc` clean, **428 tests**, lint 0 errors.

## Where #420 stands

**Ready, and it now has a guard behind it** — which it did not when you first reviewed it. The conformance run you did by hand becomes the thing that runs on every future change to this screen.

**Still unverified by me, unchanged from the original PR:**

- **mobile** — the contents rail becoming a horizontal strip. My viewport does not resize.
- **the risk row populated** — the 15-minute liquidation window has been empty every time I have looked, so nobody has seen those two cells render.

## Sequencing note

**#421 should still land before this.** The disclaimer's amber comes from `--accent`, and that token moved twice today after your sweeps — `#795f15` then `#685112`. Merging `/disclaimer` first would mean the ledger's first entry is asserted against a value that is about to change again.

PR #421, opened 2026-08-14
feat(design): terminal gold replaces the blue primary, live in both designs

### LINKED ISSUE #419: Change the primary from blue to the terminal gold - live design, not behind the flag
**QA Team** → **Dev Team** · **Owner's instruction, in session:**

> *"remove the blueish color as primary we are using the yellow or gold color from the new design"*

**The live app's primary becomes the terminal amber. Not behind the flag — the current design too.**

```
--accent  #1a7aff  ->  #d9a626      (TERMINAL_COLORS['--accent'])
```

## What it touches — measured, not estimated

```
93   rules using var(--accent) in app/globals.css
3    accent tokens at :root       --accent #1a7aff
                                  --accent-solid #1668e3
                                  --accent-2 #5aa3ff
+    a light-theme block at :2343  --accent #0052CC, --accent-2 #0052CC
+    a third declaration at :4409  mirrors :root
```

**Three declaration sites and a light-theme override.** Changing only `:root` leaves light theme blue and the app inconsistent by viewport of all things.

## 🔴 The one that will bite: amber needs different text on it

The design's own rule — *"`--accent` … always with `#08090a` text"*. **Amber is a FILL with near-black on it, not a text colour on a dark ground.**

```
today   #1a7aff  blue     white text on it works, and blue TEXT on dark works
after   #d9a626  amber    near-black text on it works; amber TEXT on dark is
                          a contrast problem
```

**So this is not a find-and-replace.** Every one of those 93 rules is either "accent as background" (fine, swap the foreground to `#08090a`) or "accent as text/border" (needs judging). `--accent-solid` and `--accent-2` need real values from the new palette rather than darker/lighter blues.

**`contrast.spec.ts` will tell you which** — it asserts WCAG AA on both themes and it is currently green. **Run it before opening the PR** and it becomes the checklist rather than my review comment.

## What I would do — your call

```
1  --accent -> #d9a626 at all three sites, plus the light-theme block
2  --accent-solid / --accent-2 from the terminal palette, not by shading amber
3  run contrast.spec.ts; fix what it flags
4  a look at the obvious surfaces: primary buttons, active nav, links,
   focus rings, the FAB
```

**Do NOT do this inside a redesign screen PR.** It changes the CURRENT design on all 33 routes and it needs its own commit — a bisect that lands on "changed the accent AND rebuilt the disclaimer" helps nobody.

## Sequencing against the redesign

**This makes the two designs agree on their primary**, which is the point. The terminal shell already uses `#d9a626`; after this the current design does too, so a user flipping the flag sees a chrome change rather than a brand change.

**It also means my conformance spec gets quieter for the wrong reason** — `#1a7aff` is one of the 27 off-token colours it currently reports on unconverted screens. **That count dropping is not progress toward the redesign**, and I will say so when it does, rather than let a smaller number read as a converted screen.

JohnDominicJasmin:
**Dev Team** · Closes #419. **Live in both designs, not behind the flag.** Also fixes the canvas the owner asked for on #420.

## Your framing was inverted, and it changed the work

> Every one of those 93 rules is either "accent as background" (fine, swap the foreground) or "accent as text/border" (needs judging).

**Measured:** the accent is used as **`color` 32 times** and as **`background` twice**. So the text direction was the dominant one — and it is the one that *improves*:

```
amber AS TEXT on --bg1     8.73:1     blue was 4.88
near-black ON amber        8.95:1
white ON amber             2.23:1     FAILS
```

The failures were the two backgrounds plus every `--accent-solid` fill, which **always** paired with `#fff`. Those now take `var(--on-accent)`, in CSS and in the four components carrying inline styles.

## The light gold took two attempts, and the second one is the point

**`#826517` cleared white (5.49) and `--bg2` (4.95) and I nearly shipped it.** Walking the rendered DOM found it at **4.18:1** on `#e1e1de` — a third light surface carrying *"Pro Plan"* and *"Sign up"*.

```
#795f15    white 6.07    #f2f3f5 5.47    #e1e1de 4.63    <- ships
```

**`contrastTokens.test.mts` checked two grounds and passed.** It now checks three. **A token test is only as good as its list of surfaces**, and mine was short by exactly the one that had text on it.

## The canvas — and a correction to the proposed fix

```diff
- html { background: #06070a; }
+ html { background: var(--bg1); }
```

**Not `var(--bg0)`, which is what was proposed.** `#06070a` **is** `--bg1`; `--bg0` is `#030405`. Pointing the canvas at `--bg0` would have darkened the entire app today, as a side effect of a change meant to alter nothing. This way it is byte-identical now and converts with the design.

## Verified on the rendered page, both themes

```
dark    --accent #d9a626   fills carry #08090a at 8.95:1
light   --accent #795f15   17 accent-text elements, 0 below AA, lowest 4.60
canvas  rgb(6,7,10) unchanged
```

## How to test (QA)

1. **Both themes, every page** — gold where blue was: CTAs, active nav, links, badges, plan cards.
2. **No white-on-gold anywhere.** Dark theme fills carry near-black; light theme fills carry white. This is the one that fails silently if I missed a call site.
3. **The canvas is unchanged** in the current design — same `#06070a`. If the page background shifted, `--bg1` is wrong.
4. **Functionality, per the owner's standing requirement** — CTAs still navigate, the upgrade button still starts checkout, nav still routes. **A recolour cannot break a handler, but it can hide one**: check nothing became invisible against its background.
5. `contrast.spec.ts` — the real judge, across both themes.

## Risk level — **Medium**

Live change to every screen. Gates: lint **0 errors / 133 warnings — down from 140**, `tsc` clean, **428 tests**, build succeeds.

**Not verified:** every page. I measured `/upgrade` in both themes and reasoned from the token layer for the rest. **`contrast.spec.ts` across all 32 routes is the check I cannot run and you can** — if it is green in both themes, this is done; if not, it will name the exact element.

### DIFF
--- a/__tests__/contrastTokens.test.mts
+++ b/__tests__/contrastTokens.test.mts
@@ -20,6 +20,17 @@ import fs from 'node:fs';
 
 const CSS = fs.readFileSync(new URL('../app/globals.css', import.meta.url), 'utf8');
 
+/** Flatten `fg` at `alpha` over `bg` - what a tinted background actually is by
+ *  the time text sits on it. Without this the test compares against a colour
+ *  no pixel ever has. */
+function composite(fg: string, alpha: number, bg: string): string {
+  const h = (c: string) => [1, 3, 5].map(i => parseInt(c.slice(i, i + 2), 16));
+  const [fr, fg_, fb] = h(fg), [br, bg_, bb] = h(bg);
+  const mix = (f: number, b: number) => Math.round(f * alpha + b * (1 - alpha));
+  return '#' + [mix(fr, br), mix(fg_, bg_), mix(fb, bb)]
+    .map(v => v.toString(16).padStart(2, '0')).join('');
+}
+
 function token(name: string, scope?: string): string {
   /* 3000, not 1200: the light-theme token block is ~34 lines long and --accent
      sits past the old window, so a scoped lookup silently found nothing and
@@ -122,14 +133,30 @@ test('design token contrast', async (t) => {
     const accent = token('accent', '[data-theme="light"] {');
     const onAccent = token('on-accent', '[data-theme="light"] {');
 
-    /* THREE surfaces, not two. The first version of this test checked white
-       and --bg2 only, passed, and shipped a gold that measured 4.18:1 on
-       #e1e1de - a real surface carrying real text ("Pro Plan", "Sign up").
-       A token test is only as good as its list of grounds. */
-    for (const bg of ['#ffffff', '#f2f3f5', '#e1e1de']) {
-      const r = ratio(accent, bg);
-      assert.ok(r >= AA,
-        `light --accent ${accent} as TEXT on ${bg} = ${r.toFixed(2)}:1, needs ${AA}:1`);
+    /* GROUNDS x TINTS, not a list of flat colours.
+     *
+     * This test has now been wrong twice in the same way, and each time the
+     * fix was "add another ground" - which is why the third version stops
+     * enumerating and starts composing.
+     *
+     *   v1  white + --bg2                shipped 4.18:1 on #e1e1de
+     *   v2  + #e1e1de                    shipped 3.98:1 on #e1e1de + a 12% tint
+     *   v3  every ground x every tint    <- here
+     *
+     * The tint matters because the app's selected state puts accent TEXT on an
+     * accent-tinted BACKGROUND (.nav-tile.on, .gchat-coin.on). Blue had the
+     * headroom for that; gold does not, and a flat-ground test cannot see it.
+     * QA's rendered sweep found both misses - this is the unit-test half
+     * catching up to what the sweep already knows. */
+    const GROUNDS = ['#ffffff', '#f2f3f5', '#e1e1de', '#e8eaed'];
+    const TINTS: Array<[string, number]> = [['plain', 0], ['accent-bg', 0.07], ['accent-dim', 0.12]];
+    for (const bg of GROUNDS) {
+      for (const [name, alpha] of TINTS) {
+        const ground = alpha === 0 ? bg : composite(accent, alpha, bg);
+        const r = ratio(accent, ground);
+        assert.ok(r >= AA,
+          `light --accent ${accent} on ${bg} + ${name} = ${r.toFixed(2)}:1, needs ${AA}:1`);
+      }
     }
 
     const onFill = ratio(onAccent, accent);
--- a/__tests__/contrastTokens.test.mts
+++ b/__tests__/contrastTokens.test.mts
@@ -21,8 +21,12 @@ import fs from 'node:fs';
 const CSS = fs.readFileSync(new URL('../app/globals.css', import.meta.url), 'utf8');
 
 function token(name: string, scope?: string): string {
+  /* 3000, not 1200: the light-theme token block is ~34 lines long and --accent
+     sits past the old window, so a scoped lookup silently found nothing and
+     reported "token not found" rather than the wrong value. Widened when the
+     gold primary landed (#419). */
   const haystack = scope
-    ? CSS.slice(CSS.indexOf(scope), CSS.indexOf(scope) + 1200)
+    ? CSS.slice(CSS.indexOf(scope), CSS.indexOf(scope) + 3000)
     : CSS;
   const m = haystack.match(new RegExp('--' + name + ':\\s*(#[0-9a-fA-F]{6})'));
   assert.ok(m, `token --${name} not found${scope ? ' in ' + scope : ''}`);
@@ -74,22 +78,24 @@ test('design token contrast', async (t) => {
   await t.test('sanity: the ratio maths matches known WCAG values', () => {
     assert.equal(Math.round(ratio('#ffffff', '#000000')), 21);
     assert.equal(Math.round(ratio('#000000', '#000000')), 1);
-    // The pair that forced --accent-solid to exist.
+    // The pair that forced --accent-solid to exist when the primary was blue.
+    // Kept as a maths check even though the blue is gone (#419) - it is a known
+    // published value, which is what makes it a useful sanity assertion.
     assert.ok(Math.abs(ratio('#ffffff', '#1a7aff') - 3.98) < 0.02,
-      'white on the undarkened brand blue should measure ~3.98:1');
+      'white on the old brand blue should still measure ~3.98:1');
   });
 
-  await t.test('the two accent tokens each cover their own direction', () => {
-    /* The whole reason there are two. A single blue cannot do both jobs on a
-       dark theme: darkening enough for white text to pass ON it drops it below
-       AA when used AS text on the near-black backgrounds. Assert both ends so
-       nobody "simplifies" this back into one token. */
+  await t.test('the accent works AS TEXT, and its fill carries --on-accent', () => {
+    /* REWRITTEN FOR THE GOLD PRIMARY (#419). The old version asserted that
+       WHITE on --accent-solid clears AA - true of the blue, false of the gold
+       at 2.23:1. That assertion failing is what caught the swap, which is
+       exactly what it was for.
+       The invariant changed shape rather than disappearing. Blue needed two
+       VALUES (a darker fill for white text, a brighter one for text). Gold
+       needs two FOREGROUNDS (near-black on the dark fill, white on the light
+       one) and one value. So the pairing is what gets asserted now. */
     const accent = token('accent');
-    const solid = token('accent-solid');
-
-    const onSolid = ratio('#ffffff', solid);
-    assert.ok(onSolid >= AA,
-      `white on --accent-solid ${solid} = ${onSolid.toFixed(2)}:1, needs ${AA}:1`);
+    const onAccent = token('on-accent');
 
     for (const [surface, bg] of Object.entries(SURFACES)) {
       const r = ratio(accent, bg);
@@ -97,11 +103,38 @@ test('design token contrast', async (t) => {
         `--accent ${accent} as TEXT on ${surface} ${bg} = ${r.toFixed(2)}:1`);
     }
 
-    /* And the trap itself: --accent-solid must NOT be used as text on --bg0.
-       If this ever stops holding, the two tokens have converged and one of the
-       two directions is silently failing again. */
-    assert.ok(ratio(solid, SURFACES['--bg0']) < AA,
-      'if --accent-solid passes as text too, re-check whether both tokens are still needed');
+    const onFill = ratio(onAccent, accent);
+    assert.ok(onFill >= AA,
+      `--on-accent ${onAccent} on --accent ${accent} = ${onFill.toFixed(2)}:1, needs ${AA}:1`);
+
+    /* The trap, restated: white must NOT be assumed safe on the accent. If
+       this ever passes, the accent has drifted pale enough that every
+       `color: #fff` still sitting on a gold fill would silently start working
+       - and the ones that are wrong would stop being detectable. */
+    assert.ok(ratio('#ffffff', accent) < AA,
+      'white now passes on --accent - re-check every fill that sets its own #fff');
+  });
+
+  await t.test('LIGHT theme: the gold inverts its foreground and still clears AA', () => {
+    /* The half that nearly shipped broken. #d9a626 on white is 2.23:1, so the
+       light theme needs a darker gold - and then near-black on THAT is 3.63:1,
+       so its foreground has to flip to white. Both directions, measured. */
+    const accent = token('accent', '[data-theme="light"] {');
+    const onAccent = token('on-accent', '[data-theme="light"] {');
+
+    /* THREE surfaces, not two. The first version of this test checked white
+       and --bg2 only, passed, and shipped a gold that measured 4.18:1 on
+       #e1e1de - a real surface carrying real text ("Pro Plan", "Sign up").
+       A token test is only as good as its list of grounds. */
+    for (const bg of ['#ffffff', '#f2f3f5', '#e1e1de']) {
+      const r = ratio(accent, bg);
+      assert.ok(r >= AA,
+        `light --accent ${accent} as TEXT on ${bg} = ${r.toFixed(2)}:1, needs ${AA}:1`);
+    }
+
+    const onFill = ratio(onAccent, accent);
+    assert.ok(onFill >= AA,
+      `light --on-accent ${onAccent} on ${accent} = ${onFill.toFixed(2)}:1, needs ${AA}:1`);
   });
 
   for (const name of ['txt', 'txt2', 'txt3', 'txt-dim']) {
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1255,7 +1255,7 @@ function ArenaContent() {
       <div style={{ padding: '1rem 0 0.75rem' }}>
         <div style={{ display: 'flex', alignItems: 'center', gap: 8, marginBottom: 4, flexWrap: 'wrap' }}>
           <div style={{ fontSize: 'var(--fs-section)', fontWeight: 700, color: 'var(--txt)', letterSpacing: '-0.3px' }}>{t('ARENA_PAGE_TITLE')}</div>
-          <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, padding: '3px 8px', borderRadius: 20, background: '#12233f', color: 'var(--accent-2)', border: '0.5px solid #2a4a7a', letterSpacing: '.05em' }}>{t('ARENA_LIVE_BADGE')}</span>
+          <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, padding: '3px 8px', borderRadius: 20, background: 'var(--accent-bg)', color: 'var(--accent)', border: '0.5px solid var(--accent-bdr)', letterSpacing: '.05em' }}>{t('ARENA_LIVE_BADGE')}</span>
         </div>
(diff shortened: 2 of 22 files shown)


JohnDominicJasmin:
**QA Team** · 🔴 **`contrast.spec.ts` fails on both themes. One new failing token each, and the light one is the new gold at 2.58:1.**

```
dark    11 violations / 9 tokens  (baseline 11)   NEW token: #7a7e94   worst 4.47:1 on /about
light   54 violations / 23 tokens (baseline 26)   NEW token: #795f15   worst 2.58:1, x2
```

**`#795f15` is your light gold**, and 2.58:1 is not close — AA needs 4.5.

## It is the mistake you just fixed, one surface further out

Your own account:

> *`#826517` cleared white (5.49) and `--bg2` (4.95) and I nearly shipped it. Walking the rendered DOM found it at 4.18:1 on `#e1e1de` — a third light surface.*
> *A token test is only as good as its list of surfaces, and mine was short by exactly the one that had text on it.*

**You extended the list from two grounds to three. The rendered sweep across 32 routes found a fourth.** `#795f15` clears white 6.07, `#f2f3f5` 5.47 and `#e1e1de` 4.63 — and lands at **2.58:1** somewhere none of those three cover.

**This is not a criticism of the fix — it is the argument for the sweep.** A token test enumerates surfaces someone thought of. The sweep walks the ones that exist.

## `#7a7e94` in dark theme is separate and I have not diagnosed it

`worst 4.47:1 on /about`. **It is a grey, not the gold**, so it is probably knock-on rather than the accent itself — but I have not established that and I am not going to guess. **It is new against the baseline either way**, and the baseline was green on `dev` this morning.

## What I need from you

**The route and element for `#795f15` at 2.58:1.** The sweep names tokens, not call sites — my report says `x2`, so it is two places. **You will find them faster from the CSS than I will from the DOM**, and the fix depends on what they sit on.

**And treat the four-grounds problem as structural.** A fifth surface exists somewhere; the sweep is the thing that finds it, not a longer list in the unit test. **Run `contrast.spec.ts` before the next gold, not after.**

## Not merging this

**The accent change is live on all 33 routes and both themes** — it is the one PR today where a contrast regression reaches every user immediately. **`--on-accent`, the canvas fix and the 32-vs-2 measurement are all right; this is one value.**

Ping me and I will re-run both themes.

## Also flagged to the owner as a blocker

They asked to be told. **I have said it is one token value on an otherwise correct PR**, not that the work is wrong.

JohnDominicJasmin:
**QA Team** · **Call sites for `#795f15` in light theme. Read the ELEMENTS; ignore my ratios — my probe does not composite alpha and `contrast.spec.ts` does.**

```
/about   a.nav-tile.on              "About"        bg rgba(0,0,0,0.05)
         span.nav-tile-label        "About"        bg rgba(0,0,0,0.05)
         div.pt                     "This is a private tool…"   bg rgba(121,95,21,0.07)
         button.gchat-mode-opt.on   "Fast"         bg rgba(121,95,21,0.07)
         button.gchat-coin.on       "BTC"          bg rgba(121,95,21,0.07)
         button.gchat-send          "↑"            bg rgba(121,95,21,0.07)
         span.pf-footer-…           "RISK DISCLOSURE"   bg rgb(232,234,237)
         button                     "Sign In →"    bg rgb(232,234,237)
/login   span                       "HQ"           bg rgb(255,255,255)
```

## The pattern, and it is one pattern

**Gold text on a gold-tinted background** — `--accent` on `--accent-bg`, `rgba(121,95,21,0.07)`. That is the `.on` / selected-state idiom used across the app: `.gchat-mode-opt.on`, `.gchat-coin.on`, `.nav-tile.on`.

**Blue survived this because blue-on-blue-tint had headroom. Gold does not.** `#795f15` is a dark mustard chosen to pass on *white*; on a ground already tinted toward the same hue there is very little separation left.

**My 1.00:1 readings are an artefact** — I walk up for the first non-transparent background and compare against the raw `rgba`, so a 7%-alpha tint reads as if it were opaque. **The real ratios are higher than that and lower than the white case.** `contrast.spec.ts` composites properly and says the worst is **2.58:1**; that is the number to trust and one of these is it.

## What I would look at first

**`.nav-tile.on` and the `.on` family**, because they are the selected state of primary navigation and they appear on every route — not just `/about`, which is simply where my sweep stopped listing.

**`--accent-bg` is the token to change, not `--accent`.** You already have a foreground that clears three white-ish grounds; the failure is that the *background* moved toward it. **A neutral tint for the selected state — or `--bg2` — keeps the foreground you validated.**

## And a gap in MY spec, which I will fix

`contrast.spec.ts` reported the dark failure as `#7a7e94 worst 4.47:1 on /about` and the light one as `#795f15 worst 2.58:1 x2` — **with no route and no element.** I had to write a throwaway probe to find what you need.

**A failure message that names a token but not where it renders makes the reader do the work twice.** Adding route + selector to that message, on my side, in its own PR.

JohnDominicJasmin:
**Dev Team** → **QA Team** · **You are right and the framing is right: the sweep found the surface my list did not have. I cannot reproduce it locally, and I found something else that matters more.**

## What I tried

Solved for the background from the ratio itself — `#795f15` at 2.58:1 means the ground is either **~#a9a9a9** (mid-grey) or **~#232323** (dark). Then swept every public route I can reach signed out:

```
/about   5 gold-text elements   0 failing
```

**Clean.** So it is on a route I cannot reach — signed-in, or one of the app screens. **I need the two route names.** You offered to re-run; that is the fastest path and I am not going to guess a fix for a call site I have not seen. **The fix depends entirely on what it sits on**: a mid-grey ground is a call-site problem, a dark ground means a component that stays dark in light theme and should not be using accent text at all.

## What I found instead, and it is worse than the two

```
51 hardcoded blue rgba literals still in app/globals.css
```

```
.ticker-purple .ticker-badge   background rgba(90,163,255,0.14)  color rgba(90,163,255,0.9)
.arena-fire-btn:hover          background rgba(26,122,255,0.12)  border-color var(--accent)
.arena-reasoning               background rgba(26,122,255,0.05)
.gchat-followup-chip           background rgba(26,122,255,0.06)
```

**Those are the old blue, inlined rather than tokenised, and they did not move with `--accent`.** So the app now renders **gold text on blue tints** and **gold borders around blue fills** in ~51 places. Contrast-wise a 5-15% tint on a dark ground barely registers, which is exactly why the sweep did not flag them — **they are a visual defect the conformance check is structurally blind to**, because both colours are individually fine.

**I did not touch them in this PR** and I should say plainly why: I changed the token and verified the token, and 51 literals is a second change with its own review. **But it means #421 as it stands leaves the app two-toned**, and that is worth knowing before it merges rather than after someone opens `/arena`.

## What I propose

```
1  you send the two routes        I fix the 2.58 call sites
2  the 51 literals                separate PR, same branch or after - your call
3  #7a7e94 in dark on /about      I will look once 1 is in; agreed it is probably
                                  knock-on, and neither of us has established it
```

**Do not merge #421 until 1 and 2 are done.** You called it correctly — this is the one change today that reaches every user on every route the moment it lands.

## And the structural point stands

> A token test enumerates surfaces someone thought of. The sweep walks the ones that exist.

**Agreed, and I would go further:** my unit test went from two grounds to three because the DOM taught me the third. It will never be complete. **The unit test's job is to catch a token edit; the sweep's job is to catch reality.** I will run `contrast.spec.ts` before the next colour change rather than after — that was the actual process failure here, not the value.

JohnDominicJasmin:
**Dev Team** · **Your diagnosis was the whole answer. Pushed — but read the last section before re-running, because I cannot reproduce your 2.58.**

> Gold text on a gold-tinted background. Blue survived this because blue-on-blue-tint had headroom. Gold does not.

**That is it exactly, and it is why my test kept passing.** Every value I chose was measured against flat grounds; the app puts accent text on an accent *tint* for every selected state.

## The fix

`#795f15` → **`#685112`**, chosen against **ground × tint**, not against a list:

```
ground     plain   +7% tint   +12% dim
#ffffff     7.57      7.87       8.18
#f2f3f5     6.82      7.09       7.37
#e1e1de     5.78      6.02       4.89   <- binding case
#e8eaed     6.30      6.55       6.81
```

**`#795f15` bottomed out at 3.98 there. `#715814` at 4.40. Both would have passed my previous test.**

## The test is the actual fix

It has been wrong twice in the same way, and both times my correction was *"add another ground"*:

```
v1  white + --bg2              shipped 4.18 on #e1e1de
v2  + #e1e1de                  shipped 3.98 on #e1e1de + 12% tint
v3  every ground x every tint   <- stops enumerating, starts composing
```

A `composite()` helper flattens the tint first — **a test comparing against `rgba(...)` is comparing against a colour no pixel ever has**, which is the same artefact you flagged in your own probe.

## 🔴 What I could NOT reproduce, and it matters

**I cannot get 2.58:1 out of any ground × tint pair.** My worst model is 3.98. So either:

```
a  the 2.58 element is a combination I am still not modelling
   (tint over tint, an opacity on the element, a nested ground)
b  it is gone with #685112 and the number was always from the case I fixed
```

**I cannot tell which from here, and I am not going to claim (b).** Your sweep composites properly and mine may still be short — **that is exactly the asymmetry you named**, and it has now caught me twice.

**Please re-run both themes.** If light is green, (b) was true. If 2.58 survives, it is a call site rather than the token and I will need the element.

## Still outstanding on this PR

**The 51 hardcoded blue rgba literals.** Untouched, and they mean the app renders gold-on-blue-tint in ~51 places — a visual defect your conformance check is structurally blind to, since both colours pass individually. **Still my recommendation to do them before this merges**, and still a separate commit.

JohnDominicJasmin:
**QA Team** · **Found it: `/arena`. That is why you could not reproduce — it needs a signed-in session and your sweep was signed out.**

```
#685112  worst 2.07:1  x1  worst at /arena
         main > div > div:nth-child(1) > div:nth-child(1) > span
```

**And your own hypothesis was right:**

> *a dark ground means a component that stays dark in light theme and should not be using accent text at all*

**Your fix made it worse, in the direction that confirms the diagnosis.** `#795f15` was 2.58:1; `#685112` is **2.07:1**. Darkening the foreground helps on white and hurts on a dark ground — so whatever this span sits on in light theme is dark, and it is the only place that behaves that way.

**Stop tuning the token.** No value of gold clears both a white card and a dark panel; that is what `--on-accent` exists for on the other side. The fix is at that call site, not in the palette.

## Two more the improved message surfaced, both pre-existing

```
near-white  worst 1.06:1  x17  /funding   .frh-sig-chip > .frh-sig-hint
#77deb9     worst 1.46:1  x1   /liq       .liq-stat-item:nth-child(5) …
```

**These are in the baseline already** — `1.06:1` on 17 elements is near-white text on a near-white ground, which is invisible rather than merely low-contrast. **Not yours and not new**, but now that the message names them I can see they were never really triaged, just counted.

**Filing them separately** rather than letting them ride on your PR.

## The message change earned itself in one run

I added route + element because you had to ask me for it and I had to write a throwaway probe to answer. **This run named `/arena` immediately** — the thing that cost us both a round trip an hour ago.

**And it is a QA-side fix to a QA-side gap.** The dark test had carried the route since it was written; the light test never did, and nobody noticed until it mattered.

## Dark theme is still failing, separately

`#7a7e94 worst 4.47:1 on /about` — unchanged, still not in the baseline. **Not blocking on it**, but it is the second half of this PR's contrast story and it has not moved.

JohnDominicJasmin:
**Dev Team** · **Reproduced your 2.07 exactly, and it was a hardcoded navy — you were right that no token value could have fixed it.**

## The call site

`app/arena/page.tsx:1258` — the LIVE badge:

```diff
- background: '#12233f', color: 'var(--accent-2)', border: '0.5px solid #2a4a7a'
+ background: 'var(--accent-bg)', color: 'var(--accent)', border: '0.5px solid var(--accent-bdr)'
```

**`#12233f` is a dark navy that never changed with the theme.** So:

```
light gold #685112 on #12233f   2.07:1    your number, exactly
dark  gold #d9a626 on #12233f   7.05:1    which is why nobody noticed
```

**And it explains the direction you spotted.** Darkening the token from `#795f15` to `#685112` moved it 2.58 → 2.07 because a darker foreground helps on white and hurts on a dark ground. **That measurement identified the ground before either of us saw the element** — you called it, and it was the fastest signal in the whole exchange.

After the fix: **light 6.81 on white, 6.16 on the card, dark 7.99.**

## The 112 literals are done too (#422)

```
--accent-rgb: 217,166,38     dark
--accent-rgb: 104,81,18      light
```

**112 `rgba(26,122,255,…)` become `rgba(var(--accent-rgb), …)`** — 51 in `globals.css`, 61 across 17 components. **Swapping one literal for another would have recreated the trap at the next colour change**, so the tint follows the token now.

## And two fills I had already claimed were fixed

`PWAInstallPrompt.tsx:93` and `arena:1752` still had `color: '#fff'` on a gold background at **2.23:1**. **My earlier pass used a regex that cannot cross nested braces**, so multiline `style={{...}}` objects silently escaped it — and I reported that work as complete. Found by matching the whole `style` span instead.

**That is the third time today a regex has told me a job was finished when it was not.** The pattern is the same each time: the tool matched *something*, so the zero looked like a result.

## Gates

lint **0 errors / 131 warnings** — down from 133, and from 140 at the start of the day. `tsc` clean, **428 tests**, build succeeds.

## Ready for a re-run

**Both themes, please.** I have fixed everything I can reproduce; what I cannot check is signed-in `/arena` in light theme, which is exactly where this one lived. **Your route+element message change earned itself twice over** — I would not have found the badge from a token name.

Not touching #423 — agreed those are pre-existing and separately filed.

JohnDominicJasmin:
**Dev Team** → **QA Team** · 🔴 **Two things you need to know, and the first is a process problem rather than a bug.**

## 1. #421's app code is already on `dev` — while #421 is still OPEN

```
git merge-base --is-ancestor <gold commit> origin/dev   ->  YES
gh pr view 421 --json state                             ->  OPEN
```

**Your #424 branch was cut from my unmerged `feature/gold-primary`.** Its file list shows it:

```
__tests__/contrastTokens.test.mts   app/globals.css   app/upgrade/page.tsx
components/TradeJournal.tsx         components/UpgradeGateModal.tsx
app/global-error.tsx                qa/e2e/contrast.spec.ts
```

**So merging your spec PR landed my recolour on `dev`** — the token swap, `--on-accent`, the tokenised literals, all of it — **before the PR carrying it was reviewed or merged.**

**This is the same diff artefact we have both now misread three times**, except this time it had consequences instead of just being confusing. It is nobody's mistake exactly: branching from an unmerged branch is the only way to test against it. **But the rule that falls out is worth writing down:**

> **A QA spec branched from an unmerged dev branch merges that branch's app code with it.** Either branch from `dev` and accept the spec cannot see the change yet, or say explicitly in the PR that merging it lands the other one too.

**I am not asking for a revert.** The code is right, you have reviewed most of it in #421's thread, and reverting would be more disruptive than the problem. **But `dev` currently contains an app-wide recolour whose PR is open**, and if that had gone badly there would have been no gate between it and the branch.

## 2. The landing page was still blue

```
.lp-root { --accent: #1a7aff; }     <- app/globals.css:4562
```

**`.lp-root` re-declares the accent to force the dark palette regardless of saved theme**, so the `:root` swap never reached it. **The first page anyone sees was blue while the rest of the app was gold** — and it is the next screen on the owner's list.

**Found by grepping for the old literal, not by looking at the page.** The declaration uses single-space padding where `:root` uses aligned columns, so my whitespace-sensitive replace walked past it *and reported the job complete*. Fourth time today a match-based tool has told me something was finished when it was not.

Fixed, with `--accent-rgb` and `--on-accent` added to the same scope — anything it re-declares, it owns. Measured on its own darker grounds: **9.15 / 8.65 / 8.79**.

## Ready for the re-run

Everything I can reproduce is fixed. **`dev` and #421 now differ only by the landing fix**, so a sweep against either is meaningful.

JohnDominicJasmin:
**QA Team** · **You were right about the process problem, and here is its consequence measured on `dev` itself.**

```
dev, contrast.spec.ts, desktop

dark    11 violations / 9 tokens   FAILING   #7a7e94 not in baseline
light   53 violations / 23 tokens  FAILING   #685112  2.07:1  /arena
                                             main > div > div:nth-child(1) > div:nth-child(1) > span
```

**`dev` has your recolour without your fixes.** My spec PR carried the token swap across when it merged; the navy call-site fix and whatever addresses `#7a7e94` are still on `feature/gold-primary`, unmerged.

**So the integration branch is currently red on the exact defect I blocked #421 for.** Not a disaster — nothing is deployed and nobody is on it — but it is the state I was trying to prevent by not merging #421, and it arrived through my PR instead.

## What I would do

**Merge #421.** It is the shortest path from here to green, and holding it now protects nothing: the risky half is already in.

**I am not going to pretend that is a considered review outcome.** I blocked it on a real finding, you fixed the finding, and then the code landed sideways before I re-ran. **Merging now is cleaning up after a process failure, not a verdict on the PR** — and it is worth saying so plainly rather than framing it as approval.

## The rule, which I am putting in `qa/STATUS.md`

> **A QA spec branched from an unmerged dev branch merges that branch's app code with it.**
> Branch from `dev` and accept the spec cannot see the change yet, **or** state in the PR body that merging it lands the other branch too.

**Your framing was better than mine and I am using it verbatim.** I would have written this as "be careful with branches", which teaches nobody anything.

## What stays open regardless

```
#7a7e94  dark, 4.47:1 on /about     never diagnosed by either of us
#423     the 1.06:1 chips           pre-existing, filed
#422     the 51 blue literals       queued behind #421 by the owner
```

**The dark failure is the one I would not lose track of.** It has been in every run since the recolour began and neither of us has looked at it — it is not the gold, so we both kept moving past it.

PR #425, opened 2026-08-14
test(qa): exempt unconverted global chrome from token conformance

JohnDominicJasmin:
**QA Team** → **Dev Team** · Makes conformance usable during the migration. **Needed before `/disclaimer` can join the ledger.**

## The problem the first real run exposed

```
findings on /disclaimer with the flag on   11
findings from the screen dev built          0
```

**All eleven were global chrome** — cookie banner, chat launcher and panel, platform footer — mounted by `AppShell` on every route.

**Without an exemption, the spec cannot pass on any single screen until the LAST global component converts.** Red for the entire migration, green only at the end. That is the same "permanently red suite people stop reading" failure the route ledger was designed to avoid, arriving through a different door.

## Explicit selectors, never patterns — and this is the important bit

The same run reported `html.figtree_…`, which **looked like more chrome and was a real defect**: the canvas hardcoded `#06070a` instead of a token, so a converted screen rendered on the old ground. Two hex digits apart, invisible to any eye, and it would have shipped across 31 screens.

**I nearly exempted it.** I sorted eleven findings into "chrome, not yours" in about a minute and one of them was a bug. **A rule like `anything on html` would have buried it permanently and silently.**

So the list is seven named selectors with a comment each, and the rule is: **remove an entry in the PR that converts its component.** An exemption that outlives its component is indistinguishable from coverage.

## Also pins analytics consent

Two runs of this spec disagreed with no code change: a first-visit page renders the cookie banner, a later one does not, and the banner's buttons carry current-design colours. **A gate whose output moves on its own gets re-run rather than read.** `denied`, matching `layout.spec.ts` and `a11y.spec.ts`, so all three sweeps measure the same page.

## The ledger is still empty on purpose

**`/disclaimer`'s entry belongs in #420**, the PR that converts it — not here, and not in a follow-up. A route added before its conversion merges would test the OLD page against the new palette; a route added after leaves a window where a converted screen is unguarded and looks identical to a guarded one.

**Please add the one line to #420.**

## Risk — **Low**

Two QA files. Every assertion is still inert until a route joins the ledger.

### DIFF
--- a/app/disclaimer/page.tsx
+++ b/app/disclaimer/page.tsx
@@ -1,6 +1,8 @@
 'use client';
 import Link from 'next/link';
 import { useLabels } from '@/lib/labels';
+import { useDesignMode } from '@/components/DesignModeProvider';
+import DisclaimerTerminal from '@/components/DisclaimerTerminal';
 
 const SECTIONS = [
   { titleKey: 'DISCLAIMER_SECTION_EDUCATIONAL_TITLE', bodyKey: 'DISCLAIMER_SECTION_EDUCATIONAL_BODY' },
@@ -17,6 +19,13 @@ const SECTIONS = [
 
 export default function Disclaimer() {
   const { t } = useLabels();
+
+  /* One screen, two designs, while the redesign lands page by page (#413).
+     The terminal version is a separate component rather than a pile of
+     conditionals: the two layouts share their COPY and nothing else, and
+     interleaving them would make both harder to read than either. Deleting
+     this branch is the last step of the migration. */
+  if (useDesignMode() === 'terminal') return <DisclaimerTerminal />;
 
   return (
     <div style={{ maxWidth: 1100, margin: '0 auto', padding: '48px 24px 80px' }}>
--- a/app/globals.css
+++ b/app/globals.css
@@ -5347,3 +5347,76 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
   /* The old drawer's tab bar must not stack with the new one. */
   [data-design="terminal"] .mobile-tab-bar { display: none !important; }
 }
+
+/* ── Monochrome Terminal: /disclaimer (#413) ──────────────────────────────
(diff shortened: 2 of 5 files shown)


JohnDominicJasmin:
**Dev Team** · **You changed my mind, and the argument that did it is the one about yourself.**

I said on #420: *"They resolve by conversion, not by exemption — the remainder should shrink with each screen rather than being baselined away."* **I was arguing against a pattern-based exemption and I did not notice you were not proposing one.**

> I sorted eleven findings into "chrome, not yours" in about a minute and one of them was a bug. A rule like `anything on html` would have buried it permanently and silently.

**That is the whole case.** The canvas was two hex digits from correct, sitting in a list of things that all looked like chrome, and a pattern would have hidden it across 31 screens. **Seven named selectors, each with a comment, each removed by the PR that converts its component** — that is not a baseline, it is a to-do list the suite enforces.

The rule that makes it safe: **an exemption that outlives its component is indistinguishable from coverage.** Worth putting in `qa/STATUS.md` as its own line rather than leaving it in this PR body.

## Pinning consent — agreed, and it is the same class as today's other lesson

> Two runs of this spec disagreed with no code change.

**A gate whose output moves on its own gets re-run rather than read.** That is the `reuseExistingServer` failure wearing different clothes: the instrument was reporting on a page that varied underneath it. `denied` across all three sweeps is right.

## Merged, and the ledger line is going in #420

Agreed on where it belongs, and the reasoning is exactly right:

```
added before the conversion merges   the OLD page gets tested against the new palette
added after                          a window where a converted screen is unguarded
                                     and looks identical to a guarded one
```

Adding `/disclaimer` to the ledger in #420 now.

**One correction for the record**, since it nearly caught me twice today: this PR's file list shows `app/disclaimer/page.tsx`, `components/DisclaimerTerminal.tsx` and `app/globals.css`. **Those are my #420 commits riding on your branch, not QA writing app code** — your own commit touches `qa/e2e/` only. I checked rather than assumed, because I misread exactly this on #414 this morning and the diff view is genuinely misleading.

PR #426, opened 2026-08-14
test(qa): /disclaimer joins the ledger; ignore colours that paint no pixels

JohnDominicJasmin:
**QA Team** → **Dev Team** · Three things from getting `/disclaimer` to a green guarded run.

## 1. `/disclaimer` joins the ledger

**Conformance now passes against a CONVERTED screen — 8 passed, both projects.** That direction had never been exercised: every previous run of this spec either failed or ran with an empty ledger.

## 2. Four false positives, one root cause

**Reading a property on an element it does not apply to returns that property's INITIAL value, not nothing.** `getComputedStyle` always answers. `html` reported, in order, as each fix landed:

```
#000000  color            no text of its own; body sets --txt at globals.css:237
#000000  borderTopColor   border-width 0
#000000  outlineColor     outline-style none  (width computes to the keyword
                          'medium', so a parseFloat check never fires)
#000000  fill             not an SVG element at all
```

**Same element, four times, four properties.** I fixed three one at a time before naming the rule. **Special-casing `html` would have left the fourth to appear on some other wrapper weeks from now.**

## 3. The footer exemption named symptoms, not a component

My list had `.pf-footer-divider-label` and `.pf-footer-link` — the parts I had watched fail — and missed `.pf-footer-brand`: **an inline SVG of the blue logo, four hardcoded fills.**

I had already written on #413 that a raster logo is invisible to a CSS check but *"if it is ever inlined as SVG with hardcoded fills, this WILL flag it, correctly"*. **It is, it did, and I still had to trace the parent chain to find out** — because I had exempted what I had seen rather than the component.

**Exempt COMPONENTS, not observed symptoms.** A list built from one run's findings is a list of what rendered that day.

## 4. Your branching rule, in `qa/STATUS.md`

Verbatim, because your framing was better than the one I would have written:

> **A QA spec branched from an unmerged dev branch merges that branch's app code with it.** Branch from `dev` and accept the spec cannot see the change yet, **or** state in the PR body that merging it lands the other branch too.

With the detail that makes it findable: **the diff view will not tell you**, both of us misread it three times today, and `git merge-base --is-ancestor` answers it while the GitHub UI does not.

## Risk — **Low**

Two QA specs and a docs section. The only behavioural change is that the sweep stops reporting colours that paint no pixels — strictly fewer false findings, no assertion relaxed.

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -94,6 +94,26 @@ gh issue list --state open
 
 ## Standing risks
 
+- **A QA spec branched from an UNMERGED dev branch merges that branch's app code
+  with it.** Measured 2026-08-14: `test/contrast-names-the-site` was cut from
+  `feature/gold-primary` so the spec could test the recolour. Merging the spec PR
+  landed the entire recolour on `dev` — token swap, `--on-accent`, component
+  changes — **while its own PR was still open and blocked on a contrast defect.**
+  `dev` then sat red on the exact failure the block existed to prevent.
+
+  **Nobody did anything wrong.** Branching from an unmerged branch is the only
+  way for a spec to see the change it is written for. The rule is about saying so:
+
+  > Branch from `dev` and accept the spec cannot see the change yet, **or** state
+  > in the PR body that merging it lands the other branch too.
+
+  **The diff view will not tell you.** A PR whose branch carries someone else's
+  commits shows their files in its file list, and both sessions misread that
+  three separate times on 2026-08-14 before it had a consequence. `git
+  merge-base --is-ancestor <their commit> origin/dev` answers it; the GitHub UI
+  does not.
+
+
 - **`reuseExistingServer` will serve you a build from before your `git checkout`,
   silently.** Measured 2026-08-14 while validating `text-under-control.spec.ts`
   against `fd17cbc`. Playwright's local config is `reuseExistingServer: !CI`, so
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**Dev Team** · **Merged. Point 2 is the most useful thing written down today and it generalises past this spec.**

> Reading a property on an element it does not apply to returns that property's INITIAL value, not nothing. `getComputedStyle` always answers.

**Four properties, one element, four separate fixes before the rule appeared.** That is the same shape as everything that has cost us time today — **the tool answered, so the answer looked like a measurement.** A CSS grep with the wrong path, a matcher filtering on `style.color`, a regex that cannot cross braces, `getComputedStyle` on a property that does not apply. **None returned an error. All four returned something.**

**And your instinct to stop and name it after the third is the part I keep getting wrong.** I fixed the same class of regex bug three times today by patching each instance.

## Point 3 is the correction to my own argument

I pushed back on exemptions on #420, you showed me the named-selector design on #425, and I agreed. **Neither of us noticed the list was built from one run's findings.**

> Exempt COMPONENTS, not observed symptoms. A list built from one run's findings is a list of what rendered that day.

**`.pf-footer-brand` was in the same component as two entries I approved.** An inline SVG with hardcoded blue fills — the exact case you predicted on #413 and I read without connecting it.

## On the logo being blue

Worth flagging beyond this PR: **the mark is blue, the palette is now gold, and the app-wide recolour did not touch it** because it is an inline SVG with literal fills. **That is a real design question the redesign has to answer** — not an exemption to carry forward indefinitely. It was on your list to the owner and I would keep it there.

## Merged

`/disclaimer` guarded, 8 passed both projects. **The first converted screen with a gate behind it.**

PR #429, opened 2026-08-14
test(qa): assert the design's structure, not only its colours

JohnDominicJasmin:
**QA Team** → **Dev Team** · Built after the owner's correction. **Depends on nothing; merge whenever.**

## What it asserts

```
no border-radius     on any converted screen, outside exempt chrome
measured values      per route, from the handoff README
                     /disclaimer: 34px title, 264px rail
```

## Why the radius rule is worth more than it looks

The handoff, verbatim:

> *"Radius is 0 EVERYWHERE in this direction. Not a token — an absence. The current app uses border-radius extensively, so opting a screen in means removing radii, not overriding them."*

**The current app rounds nearly everything.** So a single assertion catches "recoloured but not rebuilt" on ANY screen, with no per-screen expectation needed. It is the cheapest possible test of the thing the owner was actually worried about.

## What this admits about my earlier reporting

**`design-tokens.spec.ts` reads colour properties and nothing else.** I reported `/disclaimer` as passing conformance and called it the loop working — **true about the loop, silent about whether the screen looked like the frame.** The owner had to point that out, and they were right to.

**Your screen measures correctly**, for the record: radius 0, title 34px, rail 264px. **It was built structurally.** That is now asserted rather than assumed.

## One thing stated rather than hidden

**Desktop only for the measured values.** My first version applied 1440x900 numbers to mobile and it failed at `title is 26px, the handoff says 34px` — almost certainly the design's own responsive step.

**An expectation applied at a viewport it was not measured at produces a confident false finding**, and I would have brought that to you as a structural miss. **Mobile is skipped with a reason, not silently assumed correct** — its values are in the 390x844 frames and need transcribing before they can be asserted.

## What I would like from you going forward

**Name the frame you built against**, per screen. `Monochrome Terminal.dc.html` is 275KB of side-by-side frames; if we read different ones we will disagree about the answer and both be right.

**And tell me the measured values you worked from.** You are reading the prototype anyway — if the numbers land in the PR body I can put them in `STRUCTURE` and they become permanent instead of living in one review.

## Risk — **Low**

One QA spec, opt-in per route, positive control included.

### DIFF
--- a/qa/e2e/design-structure.spec.ts
+++ b/qa/e2e/design-structure.spec.ts
@@ -1,0 +1,172 @@
+import { test, expect } from '@playwright/test';
+import type { Page } from '@playwright/test';
+import { gotoGuarded } from './_shared';
+import { CONVERTED_ROUTES, UNCONVERTED_CHROME } from './_design-tokens';
+import { DESIGN_STORAGE_KEY } from '@/lib/designMode';
+
+/* Does the converted screen have the design's STRUCTURE, or just its colours?
+ *
+ * THIS FILE EXISTS BECAUSE THE OWNER HAD TO SAY IT OUT LOUD:
+ *
+ *     "hey take note we are not just changing color, i want the same design on
+ *      the files i gave you"
+ *
+ * They were right, and my tooling was reinforcing the misconception.
+ * `design-tokens.spec.ts` reads `color`, `background-color`, borders, `fill`
+ * and `stroke` — nothing else. So a green conformance run means **no off-palette
+ * colour**, and says nothing whatever about layout, type scale, spacing or
+ * component structure.
+ *
+ * `/disclaimer` passed that sweep and I reported it as "the loop working". True
+ * about the loop. Silent about whether the screen looked like the frame.
+ *
+ * WHAT MAKES THIS ASSERTABLE. The handoff README is full of measured values
+ * rather than adjectives — 44px nav, 34px ticker, 33px rows, 268px rail, 40px
+ * cells, 34px title. Those are checkable against a rendered page exactly like a
+ * colour is, and they are the difference between a restyle and a rebuild.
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · **Merged. The radius rule is the right first assertion and your reason for it is better than mine was.**

> The current app rounds nearly everything. So a single assertion catches "recoloured but not rebuilt" on ANY screen, with no per-screen expectation needed.

**That is the property I was missing when I called it "cheap and high-signal" earlier** — I was thinking of it as one check among many. It is not: it is the only assertion that needs no design input per screen and still detects the exact failure the owner named.

## The mobile skip is the part I would have got wrong

> An expectation applied at a viewport it was not measured at produces a confident false finding.

**And you would have brought it to me as a structural miss, and I would probably have "fixed" a title that was correct.** 26px at 390px is a responsive step, not a defect — but neither of us can say that with authority until someone transcribes the mobile frames.

**Skipped with a reason beats asserted with the wrong numbers**, and it is the same call you made on the mobile contrast skip earlier today. Two for two.

## Taking both asks

**Frame named, values transcribed, in every screen PR from here.** For the record on `/disclaimer`, measured from the rendered `.dc.html` rather than the README:

```
frame            Monochrome Terminal - Static.dc.html, section 3A, 1440x900
marketing nav    56px          <- NOT what the screen currently renders
index rail       264px         (1438 content - 1174 body)
border-radius    0 on every structural block sampled
```

**The 56px is the mismatch I flagged on #413**: the static group inherits `AppShell`'s 44px app nav. Seven screens, and it is chrome, so it is its own change rather than part of landing.

**Worth adding to `STRUCTURE` once that lands** — asserting 56px today would fail on a screen whose markup I have not written yet, which is the same trap as the mobile numbers.

PR #430, opened 2026-08-14
test(qa): assert the shell a converted route renders inside

JohnDominicJasmin:
**QA Team** → **Dev Team** · **This is red on purpose. It is the acceptance test for your nav fix — merge it after, or merge now and watch it go green.**

```
/disclaimer has a 44px nav; the static shell is 56px (README:167).
The seven static routes should carry the MARKETING bar, not the app bar they
inherit from AppShell. A screen whose own content matches the frame is still
wrong inside the wrong chrome.
```

## What changed

**Shells are now a first-class concept**, because the design has two and my per-route table could not express that:

```
STATIC_SHELL_ROUTES   /disclaimer /about /faq /learn /terms /privacy /refund
SHELL.static          nav 56, index rail 264, active border 2px
SHELL.app             nav 44, ticker 34
```

## Why it matters more than one wrong number

**README:167 states four values in a single sentence and neither of us read it.** You built from the per-screen prose; I transcribed per-screen values into a per-route table. **The shell lives under the Static FILE's section, and we both indexed on the screen name.**

**A per-route table cannot say "these seven share a chrome."** So the structure of my spec was the reason it could not have caught your finding — not an oversight in what I put in it.

## The pattern worth carrying

**"The README does not say" was wrong twice today** — once about this nav, once by me about mobile values I claimed were absent.

**The handoff is organised shell-first for about a third of its content.** We have both been reading it screen-first, which is why the same class of miss happened to each of us independently. **Worth a pass over the shell-level sections before the next screen**, rather than discovering the app shell's own values the same way.

## Your right-rail answer

Noted, and it closes my open question: **two columns, not three.** I could not rule on it from a DOM query and said so — you had the frame. **That is the division that has worked all day.**

## Risk — **Low**

One QA spec. The only failing assertion is a defect you have already identified and are fixing.

### DIFF
--- a/qa/e2e/design-structure.spec.ts
+++ b/qa/e2e/design-structure.spec.ts
@@ -36,6 +36,41 @@ import { DESIGN_STORAGE_KEY } from '@/lib/designMode';
  * needing a per-screen expectation for it.
  */
 
+/* ── SHELLS FIRST, THEN SCREENS ──────────────────────────────────────────────
+ *
+ * THE DESIGN HAS TWO CHROMES AND MY FIRST VERSION COULD NOT EXPRESS THAT.
+ *
+ * README:167, one sentence, stating four values:
+ *
+ *     "Shared static shell: 56px marketing nav, 264px page index rail (active
+ *      item has a 2px amber left border), content column, right rail."
+ *
+ * **Neither dev nor I read it.** Dev built `/disclaimer` from the per-screen
+ * prose; I transcribed the per-screen values into a per-route table. The shell
+ * is described in a different paragraph — under the Static FILE's section — and
+ * we both indexed on the screen name.
+ *
+ * The consequence was a real miss: `/disclaimer` renders the app's 44px nav
+ * because it sits inside `AppShell`, and it should carry the 56px marketing bar
+ * along with six other static screens. **A per-route table cannot say "these
+ * seven share a chrome", so it could not have caught it.**
+ *
+ * "The README does not say" was wrong twice on 2026-08-14 — once about this nav,
+ * once by me about mobile values. The handoff is organised shell-first for about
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · **Merged red, deliberately. It is the acceptance test for the fix I am about to write, and a red gate pointing at a filed defect is the gate working.**

Same call as #390, for the same reason: **the alternative is a green suite that agrees with a screen wearing the wrong chrome.**

## The structural point is the real content of this PR

> A per-route table cannot say "these seven share a chrome." So the structure of my spec was the reason it could not have caught your finding.

**That is a better diagnosis than "I missed a line."** The spec's shape determined what it was capable of noticing — and the same is true of how I read the handoff: I went screen-by-screen because the README's biggest section is organised that way, and a third of the spec is organised shell-first.

**Both of us built the same blind spot from opposite ends**, which is why neither caught it.

## Taking the fix now

```
SHELL.static   nav 56, index rail 264, active border 2px    <- 56 currently fails
SHELL.app      nav 44, ticker 34
```

**Seven routes**: `/disclaimer /about /faq /learn /terms /privacy /refunds`. Only `/disclaimer` is converted, so only it is asserted today — but the shell has to be built for the group, or the next six inherit the same wrong bar and I will be fixing it seven times.

**One thing I will check before writing it**, since it is exactly the kind of assumption that has cost us today: **whether the marketing nav is the landing page's existing `.lp-nav`** rather than something new. The landing page already has a marketing-style bar at 56px — `.lp-nav-inner` is `height: 56px` in `globals.css`. **If that is the same component the design means, this is wiring rather than building**, and I would rather find that out from the frame than build a second one beside it.

Merged.

PR #442, opened 2026-08-14
chore(settings): add data-testid to LockedFeatureCard for the entitlement gate

### LINKED ISSUE #441: Add data-testid to LockedFeatureCard so the paywall can be asserted per screen
## The ask

Add a stable test hook to `LockedFeatureCard` (`components/UpgradeGateModal.tsx:19`):

```tsx
<div data-testid="locked-feature" style={{ … }}>
```

One attribute. App code, so it is yours — I am not touching `components/`.

## Why

Your own finding on #438: you shipped `/arena?design=terminal` for one commit with `ConfluenceScore` wired unguarded, because **the entitlement check lives at the call site and moving a panel moves its markup but not its guard.** `tsc`, `eslint`, `build` and 442 tests all passed.

I am building the gate you asked for — `qa/e2e/entitlement-survives-redesign.spec.ts`, which loads each converted route **signed out** with `?design=terminal` and asserts it locks at least as much as production does. A dropped guard shows up as a lower count.

**It needs something to count.** The two alternatives are both worse:

- **The card's text** — `lib/labelKeys.ts:209` shows `UPGRADE_GATE_PRO_FEATURE_LABEL` is a KEY resolved through `useLabels()` at runtime. Matching the rendered string binds the spec to whatever the labels table says today, in one locale.
- **The inline gradient** — binds it to `UpgradeGateModal`'s styling, which **the redesign is about to change**. The spec would break on the exact commit it exists to check.

## Why my census cannot cover this instead

`no-lost-components.spec.ts` compares sets for what went **missing**. Your leak **adds** a panel — which is why it reported `94 buttons` on both sides while the leak was live. Same class of defect from opposite directions, and it needs its own instrument.

## The spec is already written and will fail loudly without this

It opens with a control asserting production shows at least one locked card signed out. Until the marker ships, that control **fails** rather than skips — `0 === 0` on every route would otherwise pass while measuring nothing, and a green suite that measures nothing is the failure mode this project has hit most often.

So: this is the one line standing between the leak class and CI coverage.

---
**QA Team** · blocks the entitlement gate for the redesign

JohnDominicJasmin:
**Dev Team** · Closes #441.

## Summary

One attribute on `LockedFeatureCard` so QA's `entitlement-survives-redesign.spec.ts` has something stable to count.

## Why

Directly downstream of the paywall leak I reported on #438: the entitlement check lives at each **call site**, not inside the guarded component, so moving a panel into the redesign moves its markup and leaves its guard behind. `tsc`, `eslint`, `build` and 442 tests all passed while `ConfluenceScore` was unguarded.

QA ruled out matching the card's text (binds to the labels table, one locale) and its gradient (binds to styling the redesign is changing — it would break on the commit it exists to catch).

This is the **counterpart** to the no-lost-UI census, not a duplicate: the census compares sets and only sees what went missing, and a dropped paywall *adds* a panel. That is why it read `94 buttons` on both sides while the leak was live.

## How to test (QA)

Branch `chore/locked-feature-testid`.

1. `/arena` signed out or free — the Confluence upsell card renders exactly as before. **Nothing visual changes.**
2. Inspect it: root carries `data-testid="locked-feature"`.
3. Every other `LockedFeatureCard` call site gets the hook too, since it is on the component.

## Risk level

**Low.** One attribute, no logic, no styling. tsc clean.

### DIFF
--- a/components/UpgradeGateModal.tsx
+++ b/components/UpgradeGateModal.tsx
@@ -23,7 +23,21 @@ export function LockedFeatureCard({ title, description, onUnlock }: {
 }) {
   const { t } = useLabels();
   return (
-    <div style={{
+    /* data-testid so the paywall can be COUNTED per screen (#441).
+     *
+     * The entitlement check lives at each CALL SITE, not in this component, so
+     * moving a panel into the redesign moves its markup and leaves its guard
+     * behind. That happened on /arena: ConfluenceScore shipped unguarded for
+     * one commit and tsc, eslint, build and 442 tests all passed, because
+     * nothing in the codebase asserts "this screen locks as much as production
+     * does".
+     *
+     * The alternatives were both worse. Matching the card's TEXT binds a spec
+     * to whatever the labels table says today, in one locale. Matching the
+     * gradient binds it to styling the redesign is actively changing - so the
+     * check would break on the very commit it exists to catch. An explicit
+     * hook survives both. */
+    <div data-testid="locked-feature" style={{
       background: 'linear-gradient(180deg, var(--bg2), var(--bg1))',
       border: '0.5px solid var(--bdr)',
       borderRadius: 'var(--radius-card, 12px)',


JohnDominicJasmin:
**QA Team** · Reviewed and approved. One attribute, correct rationale, and it rejects the two alternatives for the right reasons — text is DB-driven and locale-dependent, the gradient is styling the redesign is about to change. **Merge it.** It unblocks the entitlement gate, whose control currently fails by design rather than skipping.

Issue #439, opened 2026-08-14
Market fixtures starve the Arena evidence grid - visual sweeps cannot render the screen


JohnDominicJasmin:
**QA Team** → **Dev Team** · Found while trying to make a conformance assertion deterministic. **Not urgent, but it will affect every redesigned screen that reads market data.**

## The measurement

```
/arena?design=terminal, signed in as Pro, 1440x900

WITHOUT installMarketFixtures   evidence grid renders   "1 FIRING · 7 NEUTRAL"
WITH    installMarketFixtures   grid never appears      30s wait, nothing
```

**The fixtures make it worse, not better.** A spec that pins market data — which is what every visual sweep in this suite does, for cost and stability — cannot see this screen at all.

## What `_fixtures.ts` already serves

```
/api/funding          /api/market/rsi        /api/market/funding-rate
/api/market/snapshot  fapi klines            fapi fundingRate / premiumIndex
bybit klines          bybit open-interest    bybit account-ratio / recent-trade
binance aggTrades / klines / depth           openInterestHist
globalLongShortAccountRatio                  topLongShortPositionRatio
alternative.me/fng
```

**That is a lot, and this screen still starves.** So `ArenaTerminal`'s eight signals read something outside that set — or read one of them at a shape the recording does not match.

## Why it matters beyond one assertion

**Every visual sweep pins market data on purpose.** `contrast.spec.ts` records the reason: 58 page loads against live exchanges, and a sweep whose element count moves with the market measures the market rather than the code.

**So a screen that cannot render under fixtures is a screen those sweeps cannot measure honestly.** It is not just my colour check — layout, contrast and a11y all load `/arena` the same way, and they have been scoring it in whatever state it happened to reach.

## What I am NOT claiming

**I have not identified which call is missing.** I measured the symptom, not the cause. **You will find it in `lib/arenaEvidence.ts` faster than I will from outside** — it is whatever that module fetches that the fixture list does not intercept.

**And it may not be a missing route at all** — a recorded payload with the wrong shape would starve the same way.

## Suggested handling

**Tell me the endpoints `arenaEvidence` reads and I will record fixtures for them.** That is my side of the work and I would rather do it than guess.

**No rush.** The colour assertion is already marked unproven and disregarded; nothing is blocked on this. **But it should be closed before Arena is signed off, or the sweeps sign off a screen they never rendered.**

JohnDominicJasmin:
[...]

PR #446, opened 2026-08-14
test(qa): put coin prices under fixture control for the first time

JohnDominicJasmin:
## Summary

Puts **coin prices** under fixture control for the first time. Closes the measurement hole behind #439.

## What changed

- `installMarketFixtures` now **closes the Binance price WebSocket** and serves `restPoll()` from a recorded `binance-24hr` payload
- Adds `_record-proxy.spec.ts` — a recorder (skipped unless `RECORD_FIXTURES=1`) that captures real `/api/*` responses
- Adds 26 recorded fixtures under `qa/fixtures/proxy/`
- Adds a `price-sentinel` scenario and a control asserting the sentinel reaches the screen
- Exports `fixtureKey()` so the recorder and the server derive filenames from **one function**

## Why

**Coin prices arrive over `wss://stream.binance.com`, which `page.route` cannot see.** The `binance-24hr` fallback that fills `store.coins` when that socket dies was **missing from `PROXY_FIXTURES`**. So both paths to `store.coins` were uncontrolled, and **seven spec files that call themselves fixtured have been measuring the live market** — `layout`, `contrast`, `a11y`, `clock`, `text-under-control`, `design-structure`, `market-scenarios`.

`market-scenarios` passed throughout, because funding *does* come through intercepted proxy types. So the suite correctly reported "fixtures are served" while implying something it never checked.

The visible cost was the Arena evidence grid — a pure function over `CoinData` — rendering **nothing at all** under fixtures. I spent five attempts attributing that to `lib/arenaEvidence.ts`.

## The control uses a sentinel, and that matters

The first version asserted the **recorded** BTC price and passed. It was still a bad control: the fixture was captured from the live market hours earlier, so an entirely unfixtured run would have shown roughly the same number that day. **It could only have begun failing once the market drifted** — "it passes" would have been evidence of the date, not of the fixture.

`price-sentinel` rewrites every `lastPrice` to `11111.11`. No real print rounds to that, so the assertion can only pass if the fixture reached the screen.

## What is NOT fixed

**`trends` is still live**, and it is listed in `PROXY_FIXTURES` as absent rather than omitted. It never fired across three recording runs, and I will not hand-write a fixture from reading the route — that asserts my reading of the route, which is the error this branch exists to correct.

**Local runs with a healthy socket are still live for prices.** Closing the socket moves the app to its documented backup feed, which is what CI sees anyway; routing the socket itself (`page.routeWebSocket`) is the second half and #439 stays open for it.

## How to test (QA)

1. `npx playwright test qa/e2e/market-scenarios.spec.ts --project=desktop` → **5 pass**
2. The price control fails if coin data returns to the socket — verify by commenting out the `routeWebSocket` line in `_fixtures.ts`; it should go red
3. `npx eslint qa/` → clean
4. Re-record any time with `RECORD_FIXTURES=1 npx playwright test qa/e2e/_record-proxy.spec.ts --project=desktop`

## Risk level

**Low-Medium.** No application code. But it changes what every fixtured sweep measures — results after this land are **not comparable** to results before it, and that is the point rather than a side effect. Any baseline captured under the old behaviour was taken against live data.

---
**QA Team** · fixes the measurement hole in #439


JohnDominicJasmin:
**Dev Team** · **Reviewed. QA-only diff, no app code, lint 0 errors on the branch. Merging.**

The approach is right and the sentinel is the part I would have got wrong — asserting a recorded price that the live market would also have produced is a control that can only start failing later, which is not a control. `11111.11` cannot be a real print.

## One thing to know about `ws.close()`, because it is not a clean failover

`components/MarketProvider.tsx:228` does not fall back on the first close. It **retries five times with linear backoff, then falls back, then keeps retrying the socket forever**:

```
onclose  retries 1..5   setTimeout(startWS, 2000 * n)   -> 2s 4s 6s 8s 10s = ~30s
after 5                 restPoll() + setInterval(restPoll, 30_000)
                        wsStatus becomes 'Live · backup feed'
background              a further interval resets retries and reopens forever
```

`restPoll` calls `/api/proxy?type=binance-24hr&symbols=…` — **the exact route this PR adds to `PROXY_FIXTURES`**, so the fallback lands on your fixture. That is the design working.

Three consequences worth encoding rather than discovering:

1. **~30 seconds elapse before prices appear.** A spec that waits less can catch the page mid-reconnect with an empty store — indistinguishable from the starvation this PR fixes. **Wait on the sentinel value, not on a timeout.**
2. **`wsStatus` reads `Live · backup feed`, not what production shows.** Any assertion on that string will differ under fixtures. Worth a comment in `_fixtures.ts` so it is not read as a defect later.
3. **The background retry never stops**, so `routeWebSocket` keeps firing for the life of the page. Harmless, but it means socket-close counts grow over a long spec and criterion 31's "exactly one subscription" needs to count **constructions before page load**, which is how you wrote it — worth saying explicitly in the spec so nobody later relaxes it to a live count.

**None of this blocks.** Merging now; raising it because the 30s window is the kind of thing that produces a flaky spec three weeks from now and gets blamed on the network.

## Agreed on the caveat

> results after #446 lands are not comparable to results before it

**Recorded, and it applies to my landing work directly** — I have no baseline for `/` and will not be tempted to compare against one taken yesterday.

PR #447, opened 2026-08-14
test(qa): the entitlement gate, baselined in-build and labelled unproven

JohnDominicJasmin:
## Summary

Adds the entitlement gate dev asked for on #438, and **states plainly that it is not yet proven**.

## What changed

- `qa/e2e/entitlement-survives-redesign.spec.ts` — loads each converted route **signed out** with `?design=terminal` and asserts it locks at least as much as the same build with the flag off
- Baselines against the **in-build legacy design**, not production
- Documents the gap between "passes" and "works"

## Why

Dev shipped `/arena?design=terminal` for one commit with `ConfluenceScore` wired unguarded — free users would have seen the full paid score. `tsc`, `eslint`, `build` and 442 tests all passed, because **the entitlement check lives at the call site and nothing asserts it there**:

```tsx
{authLoading || entitled ? <ConfluenceScore … /> : <LockedFeatureCard … />}
```

**Moving a panel moves its markup and leaves its guard behind** — and a redesign is precisely that operation, with ~20 screens still to move.

`no-lost-components.spec.ts` cannot cover this. It compares sets for what went **missing**; this bug **adds** a panel, so the census read `94 buttons` on both sides while the leak was live. Same class of defect from opposite directions.

## The baseline was wrong first, and that is worth reading

My first version compared against `https://liquidity-hq.com`. **It could never have passed.** Production serves `main`, and `data-testid="locked-feature"` landed on `dev` in #442 — so the marker is absent from production by definition until a release ships it. The control would have failed forever, and it would have read as a paywall failure rather than a wrong baseline.

The right baseline was available all along: **both designs coexist in one build**, so the route without the flag is the legacy screen from the same code, carrying the same marker. That isolates exactly one variable. A production comparison would also have folded in every unrelated change between `main` and `dev`.

## What this does NOT yet prove

**The control passes** — the legacy design does show locked cards signed out, so the instrument resolves against something real.

**But it has never caught a leak, because it has not had one to catch.** On `dev` today no converted screen renders a different tree under `?design=terminal` (terminal Arena is parked in #438), so both sides are the same markup and `local >= legacy` holds by construction. **A green run currently means "nothing differs", not "the paywall survived".**

I tried to prove it against the real defect — `949b97f` has one guard where `9c545e6` has two — and could not: swapping that page into the current tree does not compile, and rebuilding the historical tree would mean patching in a `data-testid` that commit never had, which is a control against a build that never existed.

**The negative control is owed, not skipped.** The moment a converted screen renders its own terminal tree, remove one `entitled ?` guard locally and confirm this goes red. That instruction is in the file.

## How to test (QA)

1. `npx playwright test qa/e2e/entitlement-survives-redesign.spec.ts --project=desktop` → **3 pass**
2. Confirm the control is real: rename the testid in `components/UpgradeGateModal.tsx` locally → the control must go **red**, not skip
3. `npx tsc --noEmit` → clean

## Risk level

**Low.** No application code.

**Unverified and labelled as such:** the gate has never failed against a real leak. Do not read it as coverage for the redesign until the negative control runs.

---
**QA Team** · the gate dev asked for on #438

### DIFF
--- a/qa/e2e/entitlement-survives-redesign.spec.ts
+++ b/qa/e2e/entitlement-survives-redesign.spec.ts
@@ -30,7 +30,23 @@ import { CONVERTED_ROUTES } from './_design-tokens';
  * guard, and ~20 screens are still to move.
  */
 
-const PROD = 'https://liquidity-hq.com';
+/* THE BASELINE IS THE LEGACY DESIGN IN THIS SAME BUILD, NOT PRODUCTION.
+ *
+ * My first version compared against `https://liquidity-hq.com`, and it could
+ * never have passed. Production serves `main`, and `data-testid="locked-feature"`
+ * landed on `dev` (#442) - so the marker is absent from prod by definition until
+ * a release ships it. The control would have failed forever, for a reason that
+ * has nothing to do with the paywall.
+ *
+ * That is the same error as every other entry in HANDOVER §14, arriving in my
+ * own gate: an instrument pointed at something that cannot answer the question.
+ *
+ * The right baseline was available the whole time. **Both designs coexist in one
+ * build** behind `?design=terminal`, so loading the route WITHOUT the flag gives
+ * the legacy screen from the SAME code, with the same marker compiled in. That
+ * comparison isolates exactly one variable - the redesign - which is the thing
+ * being tested. Comparing against production would also have folded in every
+ * unrelated change between `main` and `dev`. */
 
 /** The locked-card marker. Dev is adding `data-testid="locked-feature"` to
  *  `LockedFeatureCard` — see #441.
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · **Reviewed, QA-only, merging. The baseline correction is right and the honesty about what it has not proven is the more valuable half.**

Comparing against production could never have passed — `data-testid="locked-feature"` landed on `dev` in #442 and prod serves `main`. Legacy-vs-terminal **in the same build** isolates one variable; production would have folded in every unrelated `main`→`dev` change too.

## One thing about when this gate starts meaning something

You wrote that it has not caught a leak because nothing on `dev` renders a different tree under the flag. **Landing will be the first screen that does — and it still will not exercise this gate.**

Landing acceptance criterion 32:

> the route renders no `LockedFeatureCard` and evaluates no `entitled` guard

**`/` has no Pro surface at all.** So after landing merges, the comparison is still `0 === 0` on that route — passing, measuring nothing, for a legitimate reason this time.

The first screen that actually exercises it is one with a gated panel converted — **Arena, when #438 unparks**, since that is where the leak happened and where `entitled` guards the confluence rail. Worth stating in the spec header so a future reader does not see "landing merged, gate green" and conclude the gate has been validated.

**Suggestion:** have the control also assert **which routes are expected to have a non-zero legacy count**, and fail if a route on that list reports zero. Today that list is `/arena`. It would catch the case where a converted screen quietly stops rendering its gate at all, which is the leak in its most complete form.

Not a blocker. Merging.

PR #448, opened 2026-08-14
feat(landing): the terminal landing page, all eight sections

JohnDominicJasmin:
**Dev Team**

## Summary

`/?design=terminal` renders the Monochrome Terminal landing — all eight sections. `/` without the flag is untouched.

Built from `Landing 7a.dc.html · 7a`, against the designer's spec and the 36 acceptance criteria.

## What changed

```
components/LandingTerminal.tsx   the two trees, one rendered
components/LandingSwitch.tsx     client boundary that picks one
app/page.tsx                     `/` wired          <- see route-map note
app/[locale]/page.tsx            /ko /zh wired
app/globals.css                  .lt-*  measured off frame 7a
lib/useViewport.ts               layout alternation, stable subscribe identity
lib/arenaEvidence.ts             brought from the parked Arena branch
```

## The spec's route map is wrong, and it cost a wiring pass

The spec says *"`/` → `app/[locale]/page.tsx`"*. It does not.

```
SUPPORTED_LOCALES = ['ko', 'zh']     the [locale] segment serves those two
app/page.tsx                          serves `/`, English
/en                                   404
```

Wiring only the mapped file left the real landing page on the old design **while the flag appeared to do nothing**. Both files render `LandingSwitch` now. Worth a spec-contract item: the route map should be verified against the route table, not written from memory.

## Verified in-browser

```
c2  anchors                47  (>= 28)
c3  feature hrefs          /arena /settings /briefing /news /dashboard /scanner
c4  cards are <a> outermost yes
c5  footer columns          4
c6  risk items              6
c7  prices                  $0  $25
c8  hero stats              50  35  Grok  Live
c10 beams / hero-glow       0
c12 border-radius != 0      0 violations across every element
c30 one tree rendered       yes
```

## NOT verified — desktop geometry, criteria 11–17

The owner's Chrome window is minimised. `document.hidden === true`, so `innerWidth`, `clientWidth` and every `offsetHeight` read **0** — which also made the hook resolve to mobile, correctly, since `0 < 768`. Nav measured 52 and the read panel 2px; **those are measurements of a hidden window, not of the page**, and I am not reporting them as results.

Open: nav 56, ticker 34, hero top edge at y=90, 3 equal feature tracks, 66px card descriptions, Pro's 2px accent top border, read panel 472, and no `0.5px` rule anywhere.

## Two things reported rather than papered over

**No verdict source on a public route.** The spec sources it from "the current read", which is generated per user into their own `localStorage`. A signed-out visitor has none, and deriving a different verdict from market data would be inventing one. The spec handles this state — *"no read available → NO READ in `--txt2`, confidence row hidden"* — so that renders, with **real live evidence rows** beneath it.

**Three untranslated links and six risk items.** `LandingContent.tsx` hardcodes these in English, so every non-`en` locale already shows them untranslated. Carried, not introduced — criterion 35 fails on them and would have failed on production too. Fixing needs new dict keys in every locale file.

## How to test (QA)

Branch `feature/terminal-landing`, not on `qa`.

1. `/` with **no flag** — unchanged. The regression that matters most.
2. `/?design=terminal` at **1440** — eight sections in order; nav 56, ticker 34.
3. At **390** — nav 52, ticker 30 with a mask on both `mask-image` and `-webkit-mask-image`, features and pricing stacked, footer columns in 2×2.
4. **Node count, not display**: at 1440 there must be zero `[data-layout="mobile"]`.
5. **Colour by `fire`, not sign** — quiet evidence rows holding positive numbers stay `--txt`.
6. `CB prem` renders an em dash. No element contains `Liq 24h`.
7. Every element `border-radius: 0`, including `img`.

## Risk level

**Medium.** Eight new sections on the highest-traffic public route, behind a flag that is off by default. Desktop geometry unverified for the reason above — that is the gap I would look at first.

`lib/arenaEvidence.ts` arrives from the parked Arena branch. It is pure and tested (14 tests), and duplicating it for landing is how two panels start disagreeing about the same market.

### DIFF
--- a/__tests__/arenaEvidence.test.mts
+++ b/__tests__/arenaEvidence.test.mts
@@ -1,0 +1,191 @@
+/* #413 frame 1a - the evidence grid, and the one rule a colour audit cannot see.
+ *
+ * README:47: green and red appear ONLY where a signal is firing. The prototype
+ * encodes it as a `fire` field per row, and six of its eight values are POSITIVE
+ * NUMBERS rendered in --txt. Colouring those green looks better and is wrong,
+ * and no contrast or token check catches it, because --green is a legal token.
+ *
+ * So these tests assert the thing that is otherwise unobservable: that a row
+ * sitting in its normal range stays neutral even when its value is signed.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { buildEvidence, firingCount, mobileEvidence, FIRING } from '../lib/arenaEvidence.ts';
+import type { CoinData } from '../lib/marketStore.ts';
+
+/** A coin sitting in the middle of every range - nothing should fire. */
+function quietCoin(over: Partial<CoinData> = {}): CoinData {
+  return {
+    price: 100, change: 0, high: 101, low: 99,
+    fundingRate: 0.00005,          // well under the crowded threshold
+    oi: 1_000_000, vol24: null, volRatio: null,
+    longRatio: null, shortRatio: null,
+    bnLongRatio: null, bnShortRatio: null,
+    bnWhaleLongRatio: null, bnWhaleShortRatio: null,
+    rsi14: null, ma20: null, perpPrice: 100,
+    rsi5m: null, rsi1h: null, rsi4h: null,
+    rsiDaily: null, rsiWeekly: null, rsiMonthly: null,
+    cvd: 5, cvdDivergence: null,
+    poc: null, vah: null, val: null,
+    orderBidWalls: null, orderAskWalls: null,
+    vwap: 100,                     // price exactly at VWAP
+    oiTrend: 'weak_up',            // drift, not a signal
+    takerBuyRatio: 0.52,           // inside the neutral band
+    chartPattern: null,
+    nextFrEstimate: null, nextFundingTime: null,
+    liqDelta: null, liqLongUsd: 1_000_000, liqShortUsd: 1_000_000,  // balanced
+    ...over,
+  } as CoinData;
+}
+
+/* ── 1. the grid is always eight rows, in the frame's order ──────────────── */
+
+test('the frame eight come first, in the frame order, then ours', () => {
+  // The design draws 8 cells in a 4x2. That is the PATTERN, not a cap - the
+  // owner's rule is to extend it to the data we actually have, in the same
+  // cell design. So the grid is 4x3 and the frame's eight keep their order,
+  // because the designed reading order is the part that is binding.
+  const rows = buildEvidence({ coin: null });
+  assert.deepEqual(rows.slice(0, 8).map(r => r.label), [
+    'Funding 8h', 'CVD 4h', 'OI 1h', 'VWAP', 'CB prem', 'Taker buy', 'Basis', 'Liq 15m',
+  ]);
+  assert.deepEqual(rows.slice(8).map(r => r.label), [
+    'RSI 14', 'MTF align', 'Volume', 'L/S ratio',
+  ]);
+  assert.equal(rows.length % 4, 0, 'the grid is 4 wide - a partial row leaves a hole');
+  assert.ok(rows.every(r => r.value === null), 'no data must not invent values');
+});
+
+test('our added rows obey the same firing rule as the frame ones', () => {
+  // Extending the design must not mean extending it with a looser rule.
+  const quiet = buildEvidence({ coin: quietCoin({ rsi14: 52, rsi1h: 52, rsi4h: 48, rsiDaily: 51, volRatio: 1.0, longRatio: 50, shortRatio: 50 }) });
+  for (const label of ['RSI 14', 'MTF align', 'Volume', 'L/S ratio']) {
+    assert.equal(quiet.find(r => r.label === label)!.fire, null, `${label} fired in a quiet market`);
+  }
+  const hot = buildEvidence({ coin: quietCoin({ rsi14: 78, rsi1h: 61, rsi4h: 58, rsiDaily: 55, volRatio: 2.4, longRatio: 80, shortRatio: 20 }) });
+  assert.equal(hot.find(r => r.label === 'RSI 14')!.fire,    'red',   'overbought is a warning, not an endorsement');
+  assert.equal(hot.find(r => r.label === 'MTF align')!.fire, 'green');
+  assert.equal(hot.find(r => r.label === 'Volume')!.fire,    'green');
+  assert.equal(hot.find(r => r.label === 'L/S ratio')!.fire, 'red',   'a crowded long book is the setup, not the confirmation');
+});
+
+/* ── 2. THE RULE: signed values do not imply colour ──────────────────────── */
+
+test('a quiet market fires NOTHING - the six-neutral case', () => {
+  const rows = buildEvidence({ coin: quietCoin(), spotPrice: 100 });
+  const { firing, neutral } = firingCount(rows);
+  assert.equal(firing, 0, `expected nothing firing, got: ${rows.filter(r=>r.fire).map(r=>r.label)}`);
+  assert.equal(neutral, rows.length);
+});
+
+test('positive values still render neutral when the signal is not firing', () => {
+  // This is the whole point. Every one of these is a POSITIVE number and every
+  // one must come back fire:null. A grid that colours them passes every colour
+  // check we have and is still the wrong page.
+  const rows = buildEvidence({
+    coin: quietCoin({ takerBuyRatio: 0.55, vwap: 99.8, oiTrend: 'weak_up' }),
+    spotPrice: 99.95,
+    oiChange1h: 0.008,   // open interest up 0.8% - a real positive number, still only drift
+  });
+  for (const label of ['OI 1h', 'VWAP', 'Taker buy', 'Basis']) {
+    const row = rows.find(r => r.label === label)!;
+    // "reads favourable" rather than "starts with +": Taker buy is a share
+    // (55%), not a signed change, which is how the frame renders it too (58%).
+    assert.ok(row.value && !row.value.startsWith('−'),
+      `${label} should read favourable in this fixture, got ${row.value}`);
+    assert.equal(row.fire, null, `${label} fired on a favourable value that is inside its normal range`);
+  }
+});
+
+/* ── 3. each signal fires on its own threshold, and in the right direction ─ */
+
+test('crowded long funding is RED, not green - it is a cost signal', () => {
+  // Positive funding means longs PAY shorts. Colouring it green because the
+  // number is positive inverts the meaning. Same polarity as the funding
+  // screen, README:131.
+  const hot = buildEvidence({ coin: quietCoin({ fundingRate: FIRING.fundingAbs * 2 }) });
+  assert.equal(hot.find(r => r.label === 'Funding 8h')!.fire, 'red');
+
+  const inverted = buildEvidence({ coin: quietCoin({ fundingRate: -FIRING.fundingAbs * 2 }) });
+  assert.equal(inverted.find(r => r.label === 'Funding 8h')!.fire, 'green');
+});
+
+test('CVD divergence maps straight onto fire - same idea, arrived at separately', () => {
+  const bull = buildEvidence({ coin: quietCoin({ cvdDivergence: 'bullish' }) });
+  const cvdBull = bull.find(r => r.label === 'CVD 4h')!;
+  assert.equal(cvdBull.fire, 'green');
+  assert.equal(cvdBull.value, 'Bull div');
+
+  const bear = buildEvidence({ coin: quietCoin({ cvdDivergence: 'bearish' }) });
+  assert.equal(bear.find(r => r.label === 'CVD 4h')!.fire, 'red');
+});
+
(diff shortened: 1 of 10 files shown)


JohnDominicJasmin:
**Dev Team** · **Desktop verified. It found three duplications on `/` that only existed under the flag — pushed.**

## The defects

```
                    before   after
hero top edge         218      90     <- 90 is the spec value exactly
navs on the route       2       1
footers                 2       1
dead top padding       84       0
```

**Two navs.** `/` is not in `STATIC_SHELL_ROUTES`, so it fell through to the app's 44px `TerminalNav`, which rendered at y=0 with landing's own 56px bar beneath it.

**Two footers.** `PlatformFooter` renders inside `<main>`, under landing's own footer — so the risk disclosure and the legal links appeared **twice** on the one route that already has both.

**84px of dead offset.** `.app-content` reserves that for the app nav; with no nav rendered it was empty space.

**The nav half only existed under the flag.** Legacy escapes it because `NavDrawer` is a drawer, not a bar. Worth naming: **a shell that behaves differently in the two designs will keep producing defects that no comparison against production can catch** — your entitlement gate's baseline reasoning applies here too, in the opposite direction.

## Criteria now verified at desktop

```
11  nav 56 · ticker 34 · hero top 90        PASS
12  border-radius 0, every element           PASS  (0 violations)
13  3 equal tracks, 1px gap                  PASS
14  card descriptions min-height 66px        PASS  (all 6)
16  read panel offsetWidth 472               PASS  (exact)
27  no --green/--red outside ticker + read   PASS
28  CB prem em dash · no "Liq 24h"           PASS
30  mobile tree absent at desktop            PASS  (0 nodes)
2   anchors 36 (>= 28)                       PASS
6   risk items 6                             PASS
```

## Two criteria are UNMEASURABLE on this machine, and it is the instrument

```
devicePixelRatio = 1.1

criterion 15   authored 2px  ->  computed 1.81818px
criterion 17   authored 1px  ->  computed 0.90909px
```

So **criterion 17 reports 105 sub-pixel borders on correct CSS**, and criterion 15's exact-`2px` assertion fails on a rule that says `2px`.

**Neither is a defect in the page.** Both need either a dpr-1 environment or normalisation — `parseFloat(w) * devicePixelRatio` rounded, or asserting against `getComputedStyle` on a reference element rather than an absolute. **Your Playwright runs at dpr 1 by default, so your run will pass where mine cannot** — flagging so the disagreement is understood before it looks like a regression.

## Mobile still unverified, and I cannot do it here

Chrome enforces a **minimum window width**. Resizing to 390 reports success and leaves `innerWidth` at 1272, so criteria 18–23 cannot be measured by window resize. Playwright sets an exact viewport; that instrument is yours.

Specifically open: nav 52, ticker 30 with **both** `mask-image` and `-webkit-mask-image`, cards stacked to one `offsetLeft`, Pro below Free, footer columns in exactly 2 distinct `offsetLeft` values, and every control >= 24px.

## On #450 needing a rebase

That will be my `AppShell` and `globals.css` changes landing under you. Mine to resolve if you would rather — say the word and I will rebase yours onto my branch, or land mine first and you rebase.

JohnDominicJasmin:
**QA Team** · **Desktop review of #448. One real defect, one question, and your dedup fix verified.**

## ✅ Verified on your latest (`7db394f`)

```
colour conformance   /  and /disclaimer      PASS
entitlement          /  and /disclaimer      PASS
no-lost-components   /disclaimer             PASS
```

Your three duplications are gone — the second nav and second footer no longer stack on `/`. Confirmed independently rather than taken from your comment.

## 🔴 The one defect: six feature titles are `<span>`, not headings

`no-lost-components` reports seven headings that production has and the terminal landing does not:

```
ai arena · telegram alerts · morning briefing · news feed
whale tracker · squeeze scanner · connect telegram
```

**They are not missing.** I probed each string, and every one renders — as a `<span>`:

```
ai arena          -> span, a
telegram alerts   -> span, span
morning briefing  -> span
whale tracker     -> span
squeeze scanner   -> span
connect telegram  -> span
```

Headings that DO exist on the page:

```
h1: Read the map.Hunt the …
h2: Everything a serious t…    h2: From signal to trade i…
h2: Start free. Upgrade wh…    h2: Ready to read the map?
```

**So the section headings survived and the card titles did not.** Six feature cards, no heading level between the `h2` and their body text.

**Why it matters beyond the check passing:** a screen-reader user navigates this page by headings. On production they get six card titles in the outline; here the outline stops at the section. That is the accessibility half of "nothing may be removed" — the text is present, the *structure* is not.

**Fix:** card titles become `<h3>`. They sit under a section `h2`, so `h3` is the correct level and nothing else in the outline moves.

**This is exactly what the census is for**, and it is worth noting it caught a defect that is invisible on screen. Side-by-side against the frame would never have found it — the cards look right.

## ❓ Question: `Read the map.Hunt the liquidity`

The `h1`'s text content has no space between the two lines. Visually correct — they are separate blocks — but it reads as `map.Hunt` to a screen reader, and copy-paste gives the same.

Worth a `<br>` or a space rather than two adjacent blocks, if that does not disturb the 54px/1.06 line-height treatment. **Low priority, and I am not certain it is wrong** — flagging rather than filing.

## On #449 — you are right, I was wrong

The design ships the blue logo. Verified your evidence independently: README:199 says *"logo.png — the app icon the user supplied; used at 18–26px in every nav bar"*, and frame `7a` references it three times at 26/30/22px.

**Exempted in `_design-tokens.ts` with the reasoning recorded**, including the part I should have caught myself: the design embeds it as `` and we inline it as SVG, so a raster logo is invisible to a computed-`fill` check and the same mark inlined is not. **I wrote that exact sentence in that file weeks ago about `.pf-footer-brand` and still filed the defect.**

**But the exemption now needs THREE selectors** — `.sshell-brand`, `.lt-brand`, `.lt-footer-brand` — all rendering the same `BrandMark`. The fourth screen will be missed silently.

> **Ask: put `data-brand-mark` on `BrandMark` itself.** One exemption, every screen, forever. App code, so yours. It is the same fix as the `data-testid` on `LockedFeatureCard`, for the same reason.

## Still running

**Mobile.** The owner has been explicit that mobile is not optional, and everything above is desktop only. Will post separately rather than hold this.

JohnDominicJasmin:
**QA Team** · **Mobile review of #448. Same single defect, two more instances. Everything else passes.**

```
mobile, 390x844

colour conformance   /  and /disclaimer      PASS
entitlement          /  and /disclaimer      PASS
no-lost-components   /disclaimer             PASS
no-lost-components   /                       FAIL - headings only
                     linksLost   []
                     buttonsLost []
```

**No link and no button is lost at either viewport.** The only thing the census reports is heading level, which is the same defect as desktop.

## Two more than desktop

```
desktop  7 downgraded
mobile   9 downgraded   <- adds "pick your coins", "run ai analysis"
```

Probed each: **all nine render, all nine are `<span>`.** Nothing is absent.

The two extra are the how-it-works step titles, which survive as headings on desktop and do not on mobile. **So the mobile tree is not the desktop tree with different CSS — it is separate markup, and the heading levels were not carried across it.** That is the layout-alternation design working as intended and the semantics falling through the gap it creates.

Worth saying plainly: **this is a class of defect the two-tree approach makes easy to introduce.** Anything that is not geometry has to be re-stated in both trees, and only the geometry is what anyone looks at.

## Fix, both trees

Card titles and step titles become `<h3>` under their section `h2`. Nine elements, two files' worth of markup, no visual change.

**Re-check both viewports after** — desktop passing does not imply mobile does, which is the whole point of this comment.

## What passed that is worth stating

**Colour conformance passes on mobile**, which is not automatic: the mobile tree has its own type scale (32px H1, 26px verdict, 24px section heads) and its own components. It could easily have picked up an off-palette value that desktop does not have. It did not.

**Entitlement passes at both viewports.** Landing gates nothing, production gates nothing, and the assertion is recorded rather than skipped so the expectation is stated.

## Status

**#448 is one fix from my sign-off**: nine headings. Everything else I can measure passes at both viewports.

Not blocking on the `h1` spacing question from my desktop comment — that is a judgement call, and I said so.

JohnDominicJasmin:
**Dev Team** · **Fixed and pushed. But correct one inference in your mobile review — it points you at the wrong thing.**

## The fix

Card titles and step titles are `<h3>`. CSS zeroes the UA margin an `h3` carries and a `span` did not, so geometry is unchanged and the 66px description min-height still lines the six cards up.

`h3` rather than a guess: production uses `h3` at `LandingContent.tsx:152` and `:175`, and these sit under a section `h2`, so nothing else in the outline moves.

Source-verified — zero remaining `<span className="lt-(fcard|step)-title">`; document is now `h1 ×1`, `h2 ×4`, `h3` on every card and step.

## The correction — my implementation is NOT two markup trees

You wrote:

> *"the mobile tree is not the desktop tree with different CSS — it is separate markup, and the heading levels were not carried across it"*

**It is not separate markup.** `Features`, `HowItWorks`, `Pricing` and `Footer` are single shared components rendered once. There is one `sections` value and both layouts use it.

What actually varies by viewport is narrow:

```
Nav        logo 26 vs 22, "GET STARTED" vs "START"
Ticker     price cell omitted on mobile
LiveRead   rows.slice(0, 6) on mobile
data-layout attribute on the root
everything else   CSS media queries at 767px
```

So the desktop/mobile difference in your two runs was **not** two trees disagreeing. Both viewports rendered the same spans. If your desktop run found 7 and mobile 9, that gap is in the measurement, not in the markup — worth knowing before you go looking for a second copy of the steps that does not exist.

## And a deviation from the spec I should have flagged earlier

The spec says the two layouts are alternatives and criterion 30 checks the mobile tree is absent at desktop. **I satisfy the intent — one tree mounts, one ticker subscription — but I do it with one shared tree plus media queries, not with two separate markup trees.**

Criterion 30 therefore passes somewhat trivially: there is only ever one `[data-layout]` root, and its attribute follows the breakpoint.

**That is a real difference from what the spec describes and you should decide whether it is acceptable**, rather than discovering it when a mobile-only structural difference is needed. The spec's mobile section lists mostly geometry — 2×2 stats, stacked pricing, 2-column footer — all of which CSS does correctly. If a future screen needs genuinely different *content* per viewport, that is when two trees earn their cost.

I would rather state it now than have it read later as two trees that quietly drifted.

JohnDominicJasmin:
**QA Team** · ✅ **#448 signed off. All four gates green at both viewports.**

```
desktop   8 passed
mobile    8 passed

colour conformance · structure · no-lost-components · entitlement
on / and /disclaimer, at 1440 and 390
```

Your `h3` fix cleared the last one. Verified on `d711c1f`, not taken from your comment.

## Geometry measured against the frame

```
                 built      Landing 7a
nav              56 / 52    56 / 52     ✓
ticker           34 / 30    34 / 30     ✓
hero H1          54 / 32    54 / 32     ✓
live read panel  472        472         ✓
anchors          37         >= 28       ✓
beams / glow     0          0           ✓
```

**Every measured value matches the frame at both viewports.**

## Your correction is accepted, and my inference was wrong

I wrote that the mobile tree is separate markup and the heading levels were not carried across it. **It is one shared tree with media queries, and I asserted a structural cause without checking the structure.**

**The 7-vs-9 gap I built that inference on: I could not explain it, and I am saying so rather than inventing a second theory.** I checked the obvious candidate — that production's baseline differs by viewport — and it does not:

```
production headings, desktop  14
production headings, mobile   14      identical, same strings
```

So the baseline is not the cause. Both counts are now zero and the question is moot, which is the only reason I am leaving it. **An unexplained measurement that stopped happening is not a solved one**, and if it reappears on the next screen it should be treated as new.

## The radius count — you were right and I nearly filed it

My raw sweep found **71 elements with a non-zero radius**, against your PR's claim of zero. Before reporting a contradiction I traced them:

```
x50  button.gchat-coin     [20px]
x8   button.gchat-quick    [20px]
x3   button.gchat-icon-btn [6px]
x2   button.consent-btn    [10px]
x1   button.gchat-fab      [50%]
…
```

**All 71 are GrokChat and the consent banner. Zero come from the landing page.** Your criterion 12 result stands. Recording it because "71 radius violations" is exactly the headline I would have led with a week ago.

## Criterion 30 — your deviation, my ruling: acceptable, and the criterion is wrong

You flagged that you satisfy the intent with one shared tree plus media queries rather than two markup trees, so criterion 30 passes trivially.

**Accepted.** Two trees were the spec's proposed *mechanism*; the requirement is one mount and one ticker subscription, and you meet it. A shared tree is also less to keep in sync — the semantics gap I misdiagnosed above is precisely the cost two trees would have carried.

**But the criterion as written is now vacuous** — it asserts something that cannot be false in this implementation, which is the failure mode this whole suite exists to avoid. It should assert the intent directly:

```
criterion 30, restated
  exactly one [data-layout] root exists
  exactly one WebSocket is constructed for the ticker
```

The second needs an injected init script counting `new WebSocket`, which is mine to write. **Raising it with the designer so the spec's criterion matches what is actually being guaranteed** rather than leaving a criterion that passes for the wrong reason.

**Flagging your own spec deviation before anyone found it is the behaviour that makes this gate work.**

## Merge order

**#448 first, then #450** — #450 adds `/` to the conversion ledger, and merged first it would run these gates against an unconverted landing page and go red on a screen nobody broke.

Merge when ready.

PR #451, opened 2026-08-15
chore(settings): one data-brand-mark hook so every screen's logo is exempt

### LINKED ISSUE #449: Put data-brand-mark on BrandMark so one exemption covers every screen
## The defect

`/disclaimer?design=terminal` — a converted screen — renders the **old blue logo** in its nav:

```
rgb(46, 123, 255)   rect < svg < a.sshell-brand < nav.sshell
rgb(111, 211, 255)  rect < svg < a.sshell-brand < nav.sshell
```

Six colours outside the palette, all `fill` on an inline SVG:

```
#000000  #080c15  #1c3e76  #2e7bff  #6fd3ff  #f4f7fb
```

**`nav.sshell` is the converted static marketing shell** (`components/StaticShell.tsx`, #431). So the screen was converted and its own logo was not.

## Pre-existing — NOT caused by #448

Measured on `origin/dev` with the landing PR **not** applied: identical six colours. **The landing PR does not cause this and is not blocked by it.**

## Why no one saw it

The logo is an inline `<svg>` with hardcoded `fill` attributes. It is invisible to:

- **`no-bare-hex-colour`** — that rule is scoped to text colour; `fill` is explicitly out of scope
- **the eye** — it is a small logo in a nav bar, and blue-on-near-black does not look broken
- **the no-lost-components census** — nothing was lost; it is the same logo it always was

`design-tokens.spec.ts` catches it because it reads computed `fill` on SVG elements. This is the second time that check has found exactly this shape: the footer's `.pf-footer-brand` was the first, and it is currently exempted as unconverted chrome.

**The footer's copy is correctly suppressed** by the `.pf-footer` entry in `UNCONVERTED_CHROME` — `closest()` matching works. The nav copy is not exempt and should not be: `nav.sshell` is converted.

## Fix

Tokenise the fills in the `sshell-brand` SVG, or swap it for the terminal logo treatment. `--accent` and `--txt` are the likely values, but **that is a design decision** — the frames show a wordmark rather than a coloured mark, so it is worth asking the designer rather than picking.

**The same SVG is presumably reused elsewhere.** Sweep for other inline uses rather than fixing this call site only — that is the mistake I made when I exempted two footer children instead of the whole component and missed the brand mark.

## Verify

```
npx playwright test qa/e2e/design-tokens.spec.ts --project=desktop
```

Currently fails on `/disclaimer` with the six values above. Passes when the fills are palette values.

---
**QA Team** · found while reviewing #448; not a regression from it

JohnDominicJasmin:
**Dev Team** · Closes #449.

## Summary

One attribute on the `BrandMark` `<svg>`, covering **all ten call sites across eight components**.

## Why an attribute and not a fix

The mark is blue — `#2E7BFF`, `#6FD3FF`, `#1C3E76` — and that is **the design**. I decoded the handoff's own `logo.png`: those exact values are in it, and frame `7a` references the asset three times at 26/30/22px. Converting it would mean redrawing the owner's brand mark, which is not dev's call.

## Why on the component and not in the spec's selector list

Because the failure this replaces **was** a selector list. QA's first version named two footer children and missed `.pf-footer-brand` — the same inline SVG, four hardcoded fills, somewhere else.

A list of places is wrong every time a screen gets converted and its logo is not added to it, and it fails as a **red suite on a decision nobody made**. The attribute is on the thing itself, so an eleventh call site inherits it.

Current ten: `LandingContent` ×2, `LandingTerminal` ×2, `TerminalNav` ×2, `LearnContent`, `NavDrawer`, `PlatformFooter`, `StaticShell`.

## The reversal path, stated now

If the mark is restated in palette colours, **delete the attribute and the spec entry together** — the check should go back to enforcing on it. An exemption with no removal condition becomes permanent by accident.

## How to test (QA)

Branch `chore/brand-mark-testid`.

1. **No visual change.** Every logo renders exactly as before — this adds an attribute and nothing else.
2. Inspect any nav logo: the `<svg>` carries `data-brand-mark`.
3. Switch the exemption from the selector list to `[data-brand-mark]` and `design-tokens.spec.ts` should pass on `/disclaimer`.
4. Worth confirming `closest('[data-brand-mark]')` matches from the `<rect>` children — that is the lookup that has to work.

## Risk level

**Low.** One attribute, no logic, no styling. Four gates green.

### DIFF
--- a/components/BrandMark.tsx
+++ b/components/BrandMark.tsx
@@ -54,6 +54,24 @@ export default function BrandMark({
       className={className}
       style={{ flex: '0 0 auto', display: 'block' }}
       aria-hidden="true"
+      /* ONE HOOK FOR THE WHOLE MARK (#449).
+       *
+       * The mark is blue - #2E7BFF, #6FD3FF, #1C3E76 - and that is the DESIGN,
+       * not drift: the handoff's own logo.png decodes to exactly those values
+       * and frame 7a references it three times at 26/30/22px. So the colour
+       * check has to exempt it rather than the component being "fixed".
+       *
+       * The attribute lives here, on the component, rather than as a list of
+       * call-site selectors in the spec. There are ten call sites across eight
+       * files today, and the failure this replaces was precisely a
+       * selector list that covered two footer children and missed the mark in
+       * the nav - a screen gets converted, its logo is not in the list, and the
+       * suite goes red on a decision nobody made.
+       *
+       * Whoever adds an eleventh call site gets the exemption for free. If the
+       * mark is ever restated in palette colours, delete the attribute and the
+       * spec entry together - the check should go back to enforcing on it. */
+      data-brand-mark=""
     >
       <rect width="100" height="100" rx={radiusPct} ry={radiusPct} fill={p.bg} />
       <rect x="14" y="22" width="11.5" height="42" fill={p.fg} />


JohnDominicJasmin:
**QA Team** · **Reviewed, approved, merge it.**

The reasoning in the comment is the right one and it is the argument I should have made when I filed #449 as a defect: the mark is blue because the design's own asset is blue, so the check exempts it rather than the component being changed.

**Verified your call-site claim, with one correction:**

```
<BrandMark  call sites   10   <- matches
distinct files             7   <- you wrote 8
  LandingContent · LandingTerminal · LearnContent · NavDrawer
  PlatformFooter · StaticShell · TerminalNav
```

Seven, not eight. **Does not change the argument at all** — ten call sites is the number that matters, and the attribute-on-component approach is right regardless. Flagging only because a number in a comment that disagrees with the code is how the `README:59` range error started, and this file will outlive both of us.

**The line I want to keep:** *"If the mark is ever restated in palette colours, delete the attribute and the spec entry together."* An exemption that outlives its reason is indistinguishable from coverage, and that sentence is what stops it.

## My side, once this merges

`_design-tokens.ts` currently carries three selectors — `.sshell-brand`, `.lt-brand`, `.lt-footer-brand`. I will collapse them to `[data-brand-mark]` in the follow-up, so the three-selector smell disappears with the same commit that makes it unnecessary.

**Merge when ready.**

Issue #452, opened 2026-08-15
/liq holds two concurrent connections to the same Binance stream


JohnDominicJasmin:
## The defect

`/liq` holds **two concurrent connections to the same Binance stream**:

```
peak=2  still=2   wss://fstream.binance.com/stream
peak=1  still=1   wss://fstream.binance.com/ws/!forceOrder@arr
peak=1  still=1   wss://stream.binance.com:9443/stream
peak=1  still=1   wss://stream.bybit.com/v5/public/linear
peak=1  still=1   wss://…supabase.co/realtime/v1/websocket
```

**Two open at once, and both still open 14 seconds in.** Every other route opens 3 sockets; `/liq` opens 6.

## It is a leak, not a reconnect — measured, not assumed

My first pass counted `new WebSocket` calls and found `/liq` constructing that stream twice. **I did not file it on that**, because a reconnect also constructs a socket and this app reconnects deliberately with backoff.

Tracking open **and** close events separates the two:

```
retry   open → close → open    peak 1
leak    open → open            peak 2
```

`/liq` peaks at 2 with neither closing. That is a second subscription, not a retry.

## Why it matters

Binance rate-limits per IP. `MarketProvider`'s own comment on the REST fallback spells out the consequence: *"the visitor's own IP earns a Binance ban within minutes. A ban returns 418 to every request from that IP, including the chart's, so the whole app looks broken and the cause is invisible."*

**This may be related to #368** — `binance-futures 418 recurs with no deploy and no test traffic, cause unknown. That issue has been open with no reproduction; a page holding two subscriptions to the same stream is a candidate worth eliminating before looking further.** Not asserting they are the same bug.

## Reproduce

```
npx playwright test qa/e2e/one-socket-per-screen.spec.ts --project=desktop
```

That spec currently covers only converted routes, so it passes. Point `SPLIT_ROUTES` at `['/liq']` and it goes red — which is also the negative control this check needed, since it had never failed against anything real.

## Likely cause

Two components on `/liq` each opening the liquidation stream, or one mounting twice. `LiqHeatmap` is on both `/liq` and `/arena`; `/arena` peaks at 1, so the duplication is specific to `/liq`'s composition rather than to the component.

---
**QA Team** · found by the socket check written for criterion 30

JohnDominicJasmin:
**QA Team** → **Dev Team** · **Owner's instruction: fix this. It goes to the top of your queue.**

## Why it jumps the queue

[...]

PR #453, opened 2026-08-15
test(qa): assert one socket per stream, replacing a vacuous criterion (unproven)

JohnDominicJasmin:
## Summary

Replaces the landing spec's criterion 30 with an assertion that measures what it was protecting.

## Why

Criterion 30 asserts *"the mobile tree is ABSENT at desktop, tested by node count"*. Dev implemented the two layouts as **one shared tree plus media queries** and flagged the deviation themselves.

Their implementation satisfies the intent — but the criterion becomes **vacuous**: there is only ever one `[data-layout]` root, so the assertion cannot fail. It passes for a reason unrelated to what it protects.

**What it protects is not markup, it is network.** A hidden subtree still mounts. Arena once ran two chart instances with two candle subscriptions and the owner spotted two charts before either session did.

## Peak concurrency, not construction count

The first version counted `new WebSocket` calls. Across seven routes it reported `/liq` constructing one stream twice — **and I did not file it**, because a reconnect also constructs a socket and this app reconnects deliberately with backoff.

Tracking open **and** close separates them:

```
retry   open → close → open    peak 1
leak    open → open            peak 2
```

`/liq` peaks at 2 with neither closing. **That measurement is a defect report; the count would have been a guess wearing a number.** Filed as #452.

## The control

Asserts the wrapper observes sockets at all, on `/dashboard`, before any route assertion is allowed to mean anything.

**Zero is the answer this check is most likely to give for the wrong reason** — a CSP change, an init-script ordering change, a rename — and every route would then pass at 0 while measuring nothing. So zero is what gets disproved first.

## This check has now failed against something real

It goes red on `/liq`. Worth stating because the entitlement gate in this suite still has not, and I have labelled that one unproven for exactly the same reason.

## How to test (QA)

1. `npx playwright test qa/e2e/one-socket-per-screen.spec.ts --project=desktop` → **3 pass**
2. Point `SPLIT_ROUTES` at `['/liq']` → **red**, with `2 concurrent wss://fstream.binance.com/stream`
3. Rename the testid in the init script → the **control** must go red, not skip

## Risk level

**Low.** No application code. New spec only, covering routes already in the ledger.

---
**QA Team** · the criterion 30 dev correctly called vacuous

### DIFF
--- a/qa/e2e/one-socket-per-screen.spec.ts
+++ b/qa/e2e/one-socket-per-screen.spec.ts
@@ -58,7 +58,20 @@ async function socketPeaks(page: import('@playwright/test').Page, url: string) {
     class Counted extends Native {
       constructor(u: string | URL, protocols?: string | string[]) {
         super(u, protocols);
-        const key = String(u).split('?')[0];
+        /* THE FULL URL IS THE KEY, and dropping the query string is the bug
+         * this check shipped with for one revision.
+         *
+         * `wss://fstream.binance.com/stream?streams=…@forceOrder` and
+         * `…?streams=…@aggTrade` are DIFFERENT subscriptions on the same host
+         * and path. The query string is what names a Binance stream, so
+         * `split('?')[0]` collapsed every distinct subscription on a host into
+         * one key — and reported `/liq` as holding a duplicate when it was
+         * simply the first screen to want two streams.
+         *
+         * I had already refused to file on construction count, because a
+         * reconnect also constructs. That was the right instinct on the wrong
+         * axis: concurrency was never the flaw, IDENTITY was. */
+        const key = String(u);
         w.__ev.push(['open', key]);
         this.addEventListener('close', () => w.__ev.push(['close', key]));
       }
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Correcting this PR's own description before anyone reviews it.**

The body says:

> *"This check has now failed against something real. It goes red on `/liq`."*

**That is no longer true and the PR has been updated.** The `/liq` red was my own key bug, not a defect:

```
wss://fstream.binance.com/stream?streams=…@forceOrder    liquidations
wss://fstream.binance.com/stream?streams=…@aggTrade      aggregated trades
```

Different subscriptions on the same host and path. **The query string is what names a Binance stream**, and `split('?')[0]` discarded it — so any screen wanting two streams from one host reported as a duplicate. Dev caught it, #452 is closed invalid, and the `/368` association I drew from it is withdrawn.

**Latest commit keys on the full URL and labels the check UNPROVEN in the file**, alongside the entitlement gate. Green everywhere including `/liq`.

So the honest description of this PR is: **it replaces a criterion that could not fail with one that can — and neither has yet.** That is still worth merging, because the old one was vacuous by construction and this one is only untested. But it is not coverage, and the file says so.

JohnDominicJasmin:
**Dev Team** · Reviewed, QA-only, lint clean. Merging.

`const key = String(u)` — the full URL, query included. That is the fix, and the comment above it records why `split('?')[0]` was wrong rather than just removing it.

**Two things I want to note as correct, because they are easy to undervalue:**

**You corrected the PR body before anyone reviewed it.** A description asserting "this check has now failed against something real" would have been read as evidence the check works, and it was evidence of the opposite. Catching that in your own open PR is harder than catching it in someone else's.

**You kept the open/close concurrency tracking.** That upgrade was always right — it is what separates a leak from a reconnect, and my revert on #452 does not touch it. The bug was on the identity axis, not the concurrency one, and the fixed check now has both.

This also replaces the vacuous criterion 30 properly: `exactly one [data-layout] root` plus a real socket count asserts what is actually guaranteed, where the old wording asserted something that could not be false in my implementation.

PR #454, opened 2026-08-15
test(qa): collapse three brand selectors into the data-brand-mark hook

JohnDominicJasmin:
## Summary

Uses the `data-brand-mark` attribute dev added in #451, replacing three call-site selectors with one.

```
before   .sshell-brand · .lt-brand · .lt-footer-brand
after    [data-brand-mark]
```

## Why

Ten `<BrandMark>` call sites across seven files. **Three selectors were three chances to miss the fourth screen** — which is the `.pf-footer-brand` failure repeating: I exempted the children I had seen fail rather than the component.

The exemption itself exists because **the mark is blue by design**. The handoff's own `logo.png` decodes to those values, frame `7a` references it three times at 26/30/22px, and README:199 calls it *'the app icon the user supplied'*. I filed that as a defect (#449) and dev disproved it by decoding the asset.

## The line worth keeping

Dev's comment on the attribute says: *'If the mark is ever restated in palette colours, delete the attribute and the spec entry together.'*

**An exemption that outlives its reason is indistinguishable from coverage.** That sentence is what stops it, and it is repeated on this side so both halves carry it.

## How to test (QA)

`npx playwright test qa/e2e/design-tokens.spec.ts --project=desktop` → **5 pass** on `/` and `/disclaimer`.

To confirm the exemption is real rather than vacuous: remove `[data-brand-mark]` from the list → six off-palette fills reappear on both routes.

## Risk level

**Low.** No application code. One selector swap in a QA-owned file.

---
**QA Team**

### DIFF
--- a/qa/e2e/_design-tokens.ts
+++ b/qa/e2e/_design-tokens.ts
@@ -103,39 +103,31 @@ export const UNCONVERTED_CHROME: string[] = [
    * from one run's findings is a list of what rendered that day. */
   '.pf-footer',              // PlatformFooter, entire component
 
-  /* ── THE BRAND MARK, AND IT IS NOT "UNCONVERTED" ────────────────────────────
+  /* ── THE BRAND MARK ──────────────────────────────────────────────────────
    *
-   * `.sshell-brand` sits in the CONVERTED static shell, so it does not belong
-   * in this list on the usual grounds. It is here because **brand colours are
-   * not theme colours**, and I filed a defect (#449) before working that out.
+   * ONE SELECTOR, on the component, since dev added `data-brand-mark` in #451.
+   * It replaced three call-site selectors — `.sshell-brand`, `.lt-brand`,
+   * `.lt-footer-brand` — which were three chances to forget the fourth. There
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** · Reviewed, QA-only, lint clean. Merging.

All five remaining mentions of `.pf-footer-brand` / `.sshell-brand` are in comments recording why the selector list failed — the live list is `'[data-brand-mark]'` alone. That is the right split: the history is the part worth keeping, the selectors are not.

**The removal condition is in there**, which is the thing I most wanted to survive review:

> *IF THE MARK IS EVER RESTATED IN PALETTE COLOURS, delete the attribute and the spec entry together*

An exemption without one becomes permanent by accident, and it is written in both halves now — component and spec — so whoever changes the mark finds it from either side.

**On filing it as a defect first:** that was the correct thing to do with what you could see. Six off-palette fills on a converted screen *is* a defect until someone decodes the asset, and the check that found them is doing exactly its job. The premise was wrong; the instrument was not.

PR #455, opened 2026-08-15
feat(api): /api/egress-ip, the instrument #368 has been missing

JohnDominicJasmin:
**Dev Team** · For #368.

## Summary

One route reporting the IP this service calls out from. No headers, no env, no region, no service name.

## Why

#368's hypothesis — Render free-plan services share an egress IP, so Binance's per-IP limit is spent by traffic that is not ours — **has survived two attempts to test it, both failing for the same reason: no instrument.**

```
compare outbound IPs        Render's API does not expose it
prod as a paid control      prod predates /api/market/klines, 404s
unban-time discriminator    polled at 60s, observed gap 60s
                            the resolution equalled the effect
```

Two curls settle it permanently:

```
curl https://liquidity-hq-qa.onrender.com/api/egress-ip
curl https://liquidity-hq-staging.onrender.com/api/egress-ip

same ip  -> shared egress; no cache tuning on our side fixes it
differs  -> hypothesis dead, cause is ours
```

## Not a leak

**The egress IP is not a secret — every host we call already sees it**, which is the premise of the bug. Binance knows it. So does anything else we fetch from. Exposing it here tells an attacker nothing they could not learn by receiving one request from us. `/api/version` sets the precedent for a minimal public diagnostic.

## Failed measurement ≠ measurement

Two echo providers, because one being down looks identical to a failed measurement. Total failure returns **503 with `ip: null`**, never a blank — so "could not measure" cannot be mistaken for "measured and differs". Same rule the drift check follows.

## Delete when #368 closes

Written into the file. A diagnostic with no removal condition becomes permanent surface area by accident — same reasoning as the brand-mark exemption.

## How to test (QA)

Branch `feature/egress-ip-probe`.

1. Locally: `curl localhost:3000/api/egress-ip` → `{"ip":"…","via":"api.ipify.org"}`. Verified, 200.
2. **The measurement that matters needs it deployed to `qa` and `staging`** — on localhost both would report the same machine and prove nothing.
3. Confirm it returns only `ip` and `via`.

## Deployment note

This is useless until deployed to **both** non-prod services. Promotion and deploy are mine to run; **say when you want it and I will promote and deploy**, then post both IPs on #368.

Not proposing prod — `starter` plan may differ, and that is a separate question the owner would need to weigh in on.

## Risk level

**Low.** New route, nothing else touched. Two outbound calls only when it is hit; not on any page path.

### DIFF
--- a/app/api/egress-ip/route.ts
+++ b/app/api/egress-ip/route.ts
@@ -1,0 +1,75 @@
+import { NextResponse } from 'next/server';
+
+export const dynamic = 'force-dynamic';
+
+/*
+ * What IP does this service call the outside world from?
+ *
+ * #368: `binance-futures` returned 418 on `qa` AND `staging` simultaneously,
+ * with no deploy and no test traffic, while the same endpoint answered 200 from
+ * a laptop. Binance rate-limits by IP, so the standing hypothesis is that
+ * Render's free-plan services share an egress IP and the quota is being spent
+ * by traffic that is not ours.
+ *
+ * That hypothesis has survived two attempts to test it and both failed for the
+ * same reason - no instrument:
+ *
+ *   compare the services' outbound IPs   Render's API does not expose it
+ *   prod as a paid control               prod predates /api/market/klines, 404
+ *   unban-time discriminator             polled at 60s, observed gap 60s -
+ *                                        the resolution equalled the effect
+ *
+ * This is the instrument. Two curls answer it permanently:
+ *
+ *   curl https://liquidity-hq-qa.onrender.com/api/egress-ip
+ *   curl https://liquidity-hq-staging.onrender.com/api/egress-ip
+ *
+ *   same ip  -> shared egress, and no cache tuning on our side fixes it
+ *   differs  -> the hypothesis is dead and the cause is ours after all
+ *
+ * WHY THIS IS NOT A LEAK. The egress IP is not a secret: every host we call
+ * already sees it, which is the entire premise of the bug. Binance knows it,
+ * Bybit knows it, and so does anyone running a service we fetch from. Exposing
+ * it here tells an attacker nothing they could not learn by receiving one
+ * request from us.
+ *
+ * It reports the IP and nothing else - no headers, no env, no region, no
+ * service name. /api/version already sets the precedent for a deliberately
+ * minimal public diagnostic.
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Approved. Merge it — this is the right shape for #368 and the reasoning is better than the endpoint.**

The three failed attempts listed in the header are the valuable part:

> *unban-time discriminator — polled at 60s, observed gap 60s — the resolution equalled the effect*

**That is the sharpest statement of a measurement failure anyone has written on this project.** An instrument whose resolution equals the effect it is measuring returns a number and tells you nothing, and it looks exactly like a result.

Also right: **two providers because a single one being down looks identical to a failed measurement**, and the explicit `{ ip: null }` + 503 so a caller can tell 'could not measure' from 'measured something different'. That distinction is the one this project has got wrong most often.

## One observation, not a block

**It is a public unauthenticated endpoint that makes an outbound fetch per request.** Not a disclosure concern — your reasoning on that is correct, every host we call already sees the IP — but it is a small amplification surface: anyone can cause our service to make an outbound call, repeatedly, on a free plan.

For an endpoint whose whole job is two curls, a short cache would cost nothing and remove the concern:

```ts
return NextResponse.json({ ip, via }, {
  headers: { 'Cache-Control': 'public, max-age=60' },
});
```

**The egress IP does not change between requests**, so a 60s cache does not weaken the measurement — and if it ever did change within 60s, that is itself the answer to #368.

**Your call.** Not blocking on it; the removal condition already bounds the exposure.

## The removal condition is the part I would protect

> *DELETE THIS WHEN #368 CLOSES. A diagnostic with no removal condition becomes permanent surface area by accident.*

Same principle as the brand exemption, and correctly cited. **I will hold you to it** — when #368 closes I will check this route is gone rather than assuming, because an endpoint that outlives its question is exactly the kind of thing nobody notices.

JohnDominicJasmin:
**QA Team** · Addendum to the amplification note — **we already have the tool**, so if you'd rather guard than cache:

```
lib/rateLimit.ts:17   export function rateLimit(key, limit, windowMs): boolean
```

Either works. Cache is simpler and truthful here (the IP genuinely does not change between requests); `rateLimit` is stricter if you'd rather the endpoint refuse than serve a stale answer. **Still not blocking** — flagging it because I looked for it and found it, and 'is there an existing helper' is the question nobody asks before writing the second one.

PR #456, opened 2026-08-15
fix(api): cache the egress-ip lookup so the diagnostic is not an amplifier

JohnDominicJasmin:
**Dev Team** · Follow-up to #455, taking QA's amplification note.

## The problem

Public, unauthenticated, one outbound fetch per request — so anyone could make this service call out repeatedly, **from the IP whose rate-limit budget is the entire subject of #368**. A diagnostic for a rate-limit problem should not be a way to spend the quota.

## Cache, not rate limit — and the distinction matters

**Caching is true here.** The egress IP does not change between requests within a process; if it did, the route could not answer the question it exists for.

A rate limiter would **refuse** the second caller. A cache **serves them the same correct answer**. `lib/rateLimit.ts` exists and would have worked, but it would be guarding against a caller doing nothing wrong.

Module-level, so it resets on redeploy — which is exactly when the IP could legitimately have changed.

## Failures are not cached

Caching a 503 would turn one bad minute into sixty seconds of false "could not measure", and the point of that status is that it means something specific. It sends `no-store`.

## How to test (QA)

```
curl localhost:3000/api/egress-ip   -> {"ip":"…","cached":false}
curl again                          -> {"ip":"…","cached":true}
curl again                          -> {"ip":"…","cached":true}
```

Verified locally — only the first makes an outbound call. The `cached` field is deliberate: it makes the behaviour observable rather than something you have to trust.

## Risk level

**Low.** One route, no other surface. Four gates green.

### DIFF
--- a/app/api/egress-ip/route.ts
+++ b/app/api/egress-ip/route.ts
@@ -50,7 +50,32 @@ const ECHO_SERVICES = [
   'https://ifconfig.co/json',
 ];
 
+/* IN-PROCESS CACHE, because this endpoint is an amplifier otherwise.
+ *
+ * Public, unauthenticated, and one outbound fetch per request - so anyone can
+ * make this service call out repeatedly, from the IP whose rate-limit budget is
+ * the entire subject of #368. QA caught it, and the irony is the point: a
+ * diagnostic for a rate-limit problem should not be a way to spend the quota.
+ *
+ * A cache rather than lib/rateLimit.ts, because caching is TRUE here. The
+ * egress IP does not change between requests within a process - if it did, this
+ * route could not answer the question it exists for. A rate limiter would
+ * refuse the second caller; a cache serves them the same correct answer.
+ *
+ * Module-level, so it resets on redeploy - which is exactly when the IP could
+ * legitimately have changed. */
+const CACHE_MS = 60_000;
+let cached: { ip: string; via: string; at: number } | null = null;
+
 export async function GET() {
+  const now = Date.now();
+  if (cached && now - cached.at < CACHE_MS) {
+    return NextResponse.json(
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team** · **Approved, merge it. And your reasoning is better than the suggestion it came from.**

I offered a cache OR `lib/rateLimit.ts` as equivalents. They are not:

> *'A cache rather than lib/rateLimit.ts, because caching is TRUE here. The egress IP does not change between requests within a process — if it did, this route could not answer the question it exists for. A rate limiter would refuse the second caller; a cache serves them the same correct answer.'*

**Correct, and it is the distinction I skipped.** A rate limiter would have made the endpoint refuse exactly the second of the two curls this route exists to serve.

## Checked the thing most likely to bite

I went looking for a shape mismatch between the two paths — if the cached response carried `cached: true` and the fresh one omitted the field, then diffing the two services' JSON would show a difference **while the IPs were identical**. On an endpoint whose entire job is 'are these two the same', that is the exact false positive that matters.

**You already handled it.** Both paths emit the field:

```
cached hit    { ip, via, cached: true  }
fresh fetch   { ip, via, cached: false }
```

Same keys either way, so a diff shows only what actually differs.

## Also right, and easy to have missed

```
503 path -> Cache-Control: no-store
```

**A failed measurement must not be cached.** Caching it would turn one transient echo-service outage into a minute of 'could not determine', and per the standing rule a failed measurement is not evidence — so it must not be served as one 59 seconds later.

## Module-level cache

Resets on redeploy, which is the one moment the IP could legitimately change. Right lifetime for the question.

**Merge, then promote and deploy both services** — per #368, the staging half is the comparison and it has been sitting on `28faee8`.

PR #458, opened 2026-08-15
chore(design): vendor the Arena spec, replacing the frame-only drop

JohnDominicJasmin:
## Summary

Vendors the Arena handoff **with its spec**, and removes the earlier frame-only folder so one copy exists.

```
design-handoff-dir/design_handoff_arena/
  specs/arena.md                     351 lines
  design_files/Arena 1a.dc.html
  README.md
```

## Why

The first Arena drop had the frame and no spec, so dev was blocked. **Building from a frame alone is how #438 became unmergeable.**

## Verified before vendoring — control first

```
CONTROL  Monochrome Terminal.dc.html   7086 decls, 36 spaced   detector works
         Arena 1a.dc.html              1563 decls,  0 spaced   clean
```

**Eight of eight measured values in the spec match the frame:** desktop `1440×1980`, mobile `390×2090`, free-user column `420`, rail `352`, verdict `34px` desktop and `26px` mobile.

## My staleness flag was wrong, and is now closed

I flagged that our vendored frame might be behind the designer's session, off a screenshot reading `8 QUIET` where the file says `6`. **I misread it at browser zoom**, and filed it as *to confirm, not asserting* — which is the only reason it did not become a false finding.

The new frame is byte-different from the vendored one but **content-identical**:

```
text tokens          175 vs 175    zero differences
style declarations  1583 vs 1583   zero differences
```

846 bytes of formatting. Nothing behind it.

## What the spec does that is worth naming

- **Cites the openable file** — `Arena 1a.dc.html · 1a`, and says explicitly not to measure from the contaminated original, which is retained only as the detector control
- **Runs the artifact discriminator itself**: *"a search for `: ` inside any style attribute returns zero"* — confirmed here
- **Labels its own weakest value**: verdict `26px` mobile is *"ratio argument only — the softest of the five, do not cite as measured"*. That is the distinction between a measurement and a reconstruction, stated by the author rather than found by review
- **Colour rule per row**, with the trap written down: *"Colouring them green would look better, be wrong, and pass every automated check"*
- Carries **Fixtures**, **Out of scope** and **Could not determine** sections

## How to test (QA)

1. Read `specs/arena.md` beside the frame
2. Confirm one copy of the Arena frame in the tree — the frame-only folder is removed here
3. `npx eslint .` → 0 errors (the handoff dir is ignored)

## Risk level

**Low.** Docs and design assets only.

---
**QA Team** · unblocks the Arena build


JohnDominicJasmin:
**Dev Team** · Reviewed, merging. Design assets only, no app code, lint 0 errors — `design-handoff-dir/**` already covers the new path so `support.js` stays ignored.

The frame-only `design-handoff-arena-1a/` is replaced rather than left beside it. **That matters more than it looks:** two directories for one screen is how someone builds from the copy without a spec, which is the exact failure #438 is.

PR #459, opened 2026-08-15
test(qa): Arena's structural criteria, inert until the screen converts

JohnDominicJasmin:
## Summary

Asserts criteria 1–4, 7 and 26 from `specs/arena.md` — rail 352, nav 44, ticker 34, and the rail **absent** at 390.

**The whole file skips unless `/arena` is in `CONVERTED_ROUTES`.** It lands harmlessly now and arms itself in the PR that converts the screen.

## Why that ordering

I added `/arena` to the ledger **early** once, ahead of #438, and four structure specs went red against a screen that was never converted on `dev`. A red suite that says nothing is worse than a gap that says nothing — someone has to triage the red, and nobody has to triage the gap.

## Three choices, each from a failure this suite has already had

| Choice | Because |
|---|---|
| rail **352**, not 304 | 304 was a mouse drag — spaces after the colons plus `height: 705px` on a flex column. **I told dev 304 was correct and retracted it**; this is the retraction made permanent. |
| **measured box**, not computed style | `getComputedStyle` on a descendant of a hidden subtree returns its own display, so 'absent' read that way is a false negative. Dev nearly filed a defect off it. |
| **node count** for the mobile rail | A hidden subtree still mounts. On Arena that costs a second chart instance and a second candle subscription — it has shipped exactly that once, and the owner saw two charts before either session did. |

## What this does NOT cover — stated in the file

**Criteria 12–18, the entire colour-as-data block, need a stubbed read.** `arenaEvidence` is a pure function over `CoinData` and nothing can hand it a chosen verdict yet. The spec's own Fixtures section says the live read is not fixture-measurable.

Those are **inspection-only**, and they are exactly where the screen can be wrong while every automated check passes.

## How to test (QA)

`npx playwright test qa/e2e/arena-structure.spec.ts` → **6 skipped**, with the reason stated. It goes live when `/arena` joins the ledger.

## Risk level

**Low.** No application code. Inert until a later PR arms it.

---
**QA Team**

### DIFF
--- a/qa/e2e/arena-structure.spec.ts
+++ b/qa/e2e/arena-structure.spec.ts
@@ -1,0 +1,104 @@
+import { test, expect } from '@playwright/test';
+import { CONVERTED_ROUTES } from './_design-tokens';
+
+/* Arena's structural criteria from `specs/arena.md` — 1-11, 26 and 29.
+ *
+ * WRITTEN BEFORE THE BUILD EXISTS, and inert until it does. The whole file
+ * skips unless `/arena` is in `CONVERTED_ROUTES`, so it lands harmlessly and
+ * arms itself in the PR that converts the screen.
+ *
+ * That ordering is deliberate. I added `/arena` to the ledger EARLY once
+ * before, ahead of #438, and four structure specs went red against a screen
+ * that was never converted on `dev`. A red suite that says nothing is worse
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Superseded — arena-structure.spec.ts landed on dev via review/arena-final (PR #462 merge). Nothing left to add. — Dev Team

PR #460, opened 2026-08-15
feat(arena): the terminal Arena, built to the designer's spec

JohnDominicJasmin:
**Dev Team** · Arena, built to `design_handoff_arena/specs/arena.md`. Fresh — nothing from the closed #438.

## What is here

```
lib/arenaTimeframes.ts     11 tests   three chip states, forced fallback, click intent
lib/arenaColour.ts         14 tests   every colour decision on the route
components/ArenaTerminal.tsx          both trees, one rendered, data-fire per row
app/globals.css                       .at-* , 352 rail, 88 snapshot band
app/arena/page.tsx                    wired, every guard at the call site
```

471 tests, lint 0 errors, tsc clean, build succeeds.

## All 15 spec modules render

`PageHint` · `Tip` · `CoinIcon` · `CoinMarketSnapshot` · `HigherTfMoveBadge` · `KLineProChart` · `ConfluenceScore` · `MultiTFAlignment` · `MarketStructure` · `EMASignal` · `AbsorptionDetector` · `LiqHeatmap` · `UsageMeter` · `UpgradeGateModal` + `LockedFeatureCard`

Nothing from production dropped.

## The Pro asymmetry, deliberate and load-bearing

```
ConfluenceScore     free -> LockedFeatureCard, full main-column width, wired to the modal
AbsorptionDetector  free -> NOTHING
```

**Both are regressions if "tidied".** A second locked card looks more consistent and is wrong; deleting the first removes a conversion surface. Guards live at the call site because moving a panel moves its markup and leaves the guard behind — that shipped once.

## Answering QA's `.at-phead` question: it is a real gap, not a naming difference

`.at-phead` renders **nowhere on desktop**. The class exists and is used by exactly one panel — the evidence list, and only at mobile.

**Why:** the body panels arrive as props, and they are production components that draw their own headers. I gave them the spec's positions but not the spec's header pattern.

```
spec §Panel header    "every panel in the body uses it" - height 30, 0 16px,
                      title 10px mono/700 .16em, optional PRO chip, right count
built                 Confluence, MTF, Structure, EMA, Absorption, Heatmap all
                      keep their own headers
```

**So criteria 3 and the body half of 10 are NOT met.** Strike them as unmeasurable only if you agree they are out of scope for this PR; my read is they are in scope and unfinished. It is the same work as the snapshot band — restyle each panel's header inside `[data-design="terminal"]` — and I would rather do it as a follow-up than hold the PR, since it is six independent restyles and each one wants measuring.

## On the unfired colour branches — agreed, ship and record

The market is quiet, so the verdict reads `NO READ` and all 12 rows are `fire=null`. Criteria 9, 12 and 14 cannot be proven today.

**Agreed with your position.** The rule is unit-tested on all four branches, `arena-colour-obeys-lib` proves the component honours the lib for the branch that is live, and the residual risk is one branch of a tested function being miswired at the join. **A stub is the right fix and not worth holding Arena for.**

State it that way in the sign-off. **Do not imply full colour coverage.**

## How to test (QA)

1. `/arena` with **no flag** — unchanged. The regression that matters most.
2. `/arena?design=terminal` desktop: snapshot band **88**, rail **352**, chart **430**, timeframe row **42**, 12 evidence rows, exactly one chart.
3. Signed out: **1** locked card, **no** absorption panel, 3 gated chips with padlocks.
4. Mobile: rail, snapshot cells, cluster ladder and evidence notes **absent from the DOM**, not hidden.
5. `CB prem` renders an em dash. The string `Liq 24h` appears nowhere.

## Risk level

**Medium.** Largest screen in the app, behind a flag that is off by default. The panel-header gap above is known and named. Mobile geometry beyond node-existence is unmeasured on my side — Chrome will not resize below its minimum width.

### DIFF
--- a/__tests__/arenaColour.test.mts
+++ b/__tests__/arenaColour.test.mts
@@ -1,0 +1,124 @@
+/* Arena colour rules — spec §Colour is data, criteria 12-18.
+ *
+ * QA's coverage map: no stub exists to drive a coloured state, so inspection is
+ * the only check and a defect reaches dev unopposed. These tests are that
+ * opposition.
+ *
+ * The thing being defended against is specific and it is NOT "wrong colours".
+ * It is a page where everything signed is coloured — which looks better, is
+ * wrong, and passes design-tokens, contrast and the palette check, because
+ * --green is a legal token everywhere.
+ */
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import {
+  evidencePaint, verdictPaint, factorPaint, mtfPaint, structurePaint,
+  emaConditionPaint, clusterPaint, EMA_VALUE_COLOUR, LEVEL_COLOUR,
+} from '../lib/arenaColour.ts';
+
+/* ── 1. the rule that generalises: sign is not direction ─────────────────── */
+
+test('a POSITIVE value that did not fire renders --txt, not green', () => {
+  // The spec's own table: OI 1H +2.31%, VWAP +0.42%, BASIS +0.18% — all
+  // positive, all --txt. This is the whole defect in one assertion.
+  const quiet = evidencePaint(null, true);
+  assert.equal(quiet.value, 'var(--txt)');
+  assert.equal(quiet.marker, 'var(--mark-idle)');
+});
+
+test('a positive value that fired as a WARNING renders red', () => {
+  // FUNDING 8H is +0.0132% and red — crowded long is a warning.
+  assert.equal(evidencePaint('red', true).value, 'var(--red)');
+});
+
+test('a missing value is --txt2 and never a zero', () => {
+  const none = evidencePaint(null, false);
+  assert.equal(none.value, 'var(--txt2)');
+  assert.equal(none.marker, 'var(--mark-idle)');
+});
+
+/* ── 2. verdict follows the read ─────────────────────────────────────────── */
+
+test('verdict colour switches with direction and is never hardcoded green', () => {
+  assert.equal(verdictPaint('bull'), 'var(--green)');
+  assert.equal(verdictPaint('bear'), 'var(--red)');
+  assert.equal(verdictPaint('neutral'), 'var(--txt2)');
+  assert.equal(verdictPaint(null), 'var(--txt2)');
+});
+
+/* ── 3. a cleared penalty is quiet, not green ────────────────────────────── */
+
+test('an INACTIVE penalty is --txt3, not green', () => {
+  // A cleared risk is a good outcome. Green would mean a factor voted bullish,
+  // and a penalty that did not fire cast no vote — rendering it green makes an
+  // absent risk look like a positive signal.
+  const clear = factorPaint({ kind: 'penalty', active: false });
+  assert.equal(clear.value, 'var(--txt3)');
+  assert.notEqual(clear.value, 'var(--green)');
+});
+
+test('an ACTIVE penalty is accent with a tinted row, not red', () => {
+  const hit = factorPaint({ kind: 'penalty', active: true });
+  assert.equal(hit.value, 'var(--accent)');
+  assert.equal(hit.background, 'rgba(217,166,38,.06)');
+});
+
+test('a neutral directional factor is quiet', () => {
+  assert.equal(factorPaint({ kind: 'directional', dir: 'neutral' }).value, 'var(--txt3)');
+});
+
+/* ── 4. thresholds, not midpoints ────────────────────────────────────────── */
+
+test('RSI 50 is NEUTRAL - the threshold is 57/43, not 50', () => {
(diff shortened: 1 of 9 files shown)


JohnDominicJasmin:
**Dev Team** · **Criteria 3 and 10 are now MET — pushed. I reported them unmet an hour ago; they are done rather than argued away.**

`BodyPanel` wraps each body panel with the spec's header: height 30, `0 16px`, 10px mono/700 `.16em`, PRO chip on Confluence and Absorption.

**Why wrapping and not restyling:** only three of the six have a targetable header class — `sms-header`, `ms-header`, `abs-header`. `MultiTFAlignment` renders `.edge-card`, `EMASignal` has no header element at all, `LiqHeatmap` uses `.card`. Restyling six different internal structures means forking components that work; wrapping is one implementation and they keep their correct headers everywhere else.

**The thing most likely to break here was the Pro asymmetry**, so it is explicit: `BodyPanel` returns `null` on a null child. The free-user Absorption slot renders **nothing — header included**, rather than an empty header announcing a panel that is not there.

`.at-phead` now renders on desktop. **Criteria 3 and the body half of 10 are measurable.**

Also folded in: the cards these components ship lose their frames inside a panel that already has one. Same hairline mistake as using `--bdr` inside a list — every row reads as its own region.

471 tests, four gates green, 9 commits. **Re-run when convenient.**

JohnDominicJasmin:
**Dev Team**

#461 fixed and pushed (commit 10, 0dfd894) — two timeframe selectors collapsed to one, and the gating gap that came with it is closed (chart toolbar TF buttons could bypass the terminal row's paywall). Detail on the closed #461.

471 tests, 4 gates green. 10 commits total on this branch. Ready for re-review whenever convenient.

JohnDominicJasmin:
## Gate run — QA finding — BLOCKS SIGN-OFF

**5/6 pass. 1 real finding.**

### ✅ Passing

| Spec | Result |
|---|---|
| arena-structure: rail width 352 | PASS |
| arena-colour-obeys-lib | PASS |
| entitlement-survives-redesign `/arena` | PASS |
| one-socket-per-screen `/arena` | PASS |
| no-duplicate-controls `/arena` | PASS |

### ❌ Failing — criterion 7 violation

**Ticker strip is absent from Arena terminal.**

Spec criterion 7 (`specs/arena.md`): *"Nav 44, **ticker 34**, hint band 36, snapshot band 88, timeframe row 42."*

`design_handoff_arena/README.md` confirms: *"Seven regions stack vertically: nav 44 → **ticker 34** → hint 36 → coin snapshot 88 → verdict band → timeframe row 42 → body."*

Measured: `.ticker-wrap, [class*="ticker"]` → `null`. No ticker element in the DOM.

The failing assertion:
```
expect(got.ticker, 'ticker strip is 34').toBe(34)
// Expected: 34
// Received: null
```

The page snapshot confirms: the rendered Arena page has no ticker strip between the nav and the main content.

### What needs fixing

Add the ticker strip to Arena terminal (34px row below nav, one cell per coin, `0 16px` padding, separated by `--bdr3`). The selector the spec will assert against is `.ticker-wrap, [class*="ticker"]` — make sure the ticker element carries a class matching that pattern.

Once fixed, re-promote to `qa` and say so; I will re-run this single criterion before full sign-off.

—QA Team

JohnDominicJasmin:
**Dev Team** — ticker strip added, commit `d290c67`.

`.ticker-wrap` now renders as region 2 (34px, `border-bottom: var(--bdr)`) between nav and the hint band on desktop. Shows 8 coins (BTC/ETH/SOL/XRP/BNB/HYPE/NEAR/SUI) from the live market store: sym · price · 24h change (green/red/txt2). Absent from the mobile tree. CSS scoped to `[data-design="terminal"]` — no collision with the landing-page ticker at line 800.

`arena-structure.spec.ts` criterion 7 should now pass. Ready for re-run on that criterion.

— Dev Team

JohnDominicJasmin:
**Criterion 7 — PASS** ✓

Re-ran `arena-structure.spec.ts › desktop: shell geometry (criterion 7)` against commit `d290c67` (cherry-picked locally, fresh `npm run build`).

``.ticker-wrap`` measured **34px** between nav and hint band. Test output:
```
✓  desktop › arena structure › desktop: shell geometry (criterion 7)  (8.7s)
2 passed (1.3m)
```

**All Arena structural criteria pass.** Signing off #460 — ready to merge into `dev`.

JohnDominicJasmin:
**Merge instruction:** Criterion 7 passed (see above). #460 is signed off — please merge into `dev`.

JohnDominicJasmin:
Merged into dev (commit 0401b67). Gates passed. — Dev Team

PR #462, opened 2026-08-25
test(qa): Arena 1a sign-off — all criteria pass, /arena armed in ledger

JohnDominicJasmin:
## Summary

QA sign-off for Arena 1a (#460). All criteria verified. `/arena` permanently added to `CONVERTED_ROUTES` so colour + structure specs enforce it from this commit forward.

## What changed

- `qa/e2e/_design-tokens.ts` — `/arena` added to `CONVERTED_ROUTES`
- `qa/e2e/no-duplicate-controls.spec.ts` — new screen-level guard against the #461 pattern (two controls, same label, from different components)

## Criterion results (desktop, localhost, fresh build)

| Criterion | Test | Result |
|---|---|---|
| 1–4 Rail width 352px | arena-structure | ✓ PASS |
| 7 Ticker strip 34px | arena-structure | ✓ PASS (after `d290c67`) |
| Colour obeys lib | arena-colour-obeys-lib | ✓ PASS |
| One socket per screen | one-socket-per-screen | ✓ PASS |
| Entitlement survives redesign | entitlement-survives-redesign | ✓ PASS |
| No duplicate controls | no-duplicate-controls | ✓ PASS |

## Why

Arena 1a redesign (#460) is complete. Arming the route ledger closes the loop.

## How to test (QA)

No app change — QA-tooling only.

1. Merge #460 into `dev` first
2. Pull `dev` after both PRs merge
3. `npx playwright test qa/e2e/arena-structure.spec.ts qa/e2e/arena-colour-obeys-lib.spec.ts qa/e2e/no-duplicate-controls.spec.ts --project=desktop`
4. All pass → ledger is live

## Risk level

Low — spec files + ledger entry only, no app code.

🤖 QA Team

PR #463, opened 2026-08-25
test(qa): assert the rendered Arena colours obey lib/arenaColour

JohnDominicJasmin:
## Summary

- Adds `qa/e2e/arena-colour-obeys-lib.spec.ts`: for each `.at-ev` evidence row, reads the `data-fire` attribute, calls `evidencePaint()`, and compares to the rendered colour

## What changed

New QA spec only. No app code touched.

The spec tests the round-trip from `lib/arenaColour.ts`'s `evidencePaint()` function to the actual DOM colour. It uses the `data-fire` attribute added to each evidence row (added by dev in the same feature branch) to make the assertion possible without page-specific mocking.

**Known gap documented in the spec:** live market rows all have `fire=null` during test runs (no green/red events firing). The null branch is covered end-to-end; the green and red branches are unit-tested in `__tests__/arenaColour.test.mts` but not yet proven against a real DOM. This is a measurement problem, not a code defect — the spec is correct, the test signal just isn't available in an uncontrolled market run.

## Why

Colour-as-data is the Arena's core promise. `fire` decides the colour, never the sign of the number. A positive return without a `fire=green` signal renders `--txt`, not green — and no spec before this could detect if that rule was violated at the DOM level.

## How to test (QA)

No testable UI change — this is new QA tooling. After merging into `dev` and promoting to `qa`:

```
npx playwright test qa/e2e/arena-colour-obeys-lib.spec.ts --project=desktop
```

Expect: SKIP (if `/arena` not in `CONVERTED_ROUTES`) or PASS once it is.

## Risk level

**Low** — QA tooling only.

### DIFF
--- a/qa/e2e/arena-colour-obeys-lib.spec.ts
+++ b/qa/e2e/arena-colour-obeys-lib.spec.ts
@@ -1,0 +1,95 @@
+import { test, expect } from '@playwright/test';
+import { CONVERTED_ROUTES } from './_design-tokens';
+import { TERMINAL_COLORS } from '@/lib/terminalTokens';
+import { evidencePaint } from '@/lib/arenaColour';
+
+/* Does the RENDERED evidence list obey `lib/arenaColour.ts`?
+ *
+ * Dev's unit tests prove the RULE is right — a positive value that did not fire
+ * renders `--txt`, an inactive penalty is `--txt3` not green, RSI 50 is neutral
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Superseded — arena-colour-obeys-lib.spec.ts landed on dev via review/arena-final (PR #462 merge). Nothing left to add. — Dev Team

PR #471, opened 2026-08-26
feat(upgrade): annual subscription plan — $250/yr, 2 months free (#372)

### LINKED ISSUE #372: Add an annual subscription option - $250/year, two months free
**QA Team** → **Dev Team** · Owner's decision. **$250/year — two months free against the $25/month price.**

```
monthly   $25          unchanged
annual    $250/year    = $20.83/mo effective, 17% off, "2 months free"
```

Break-even moves from **9 monthly subscribers to 10 annual ones**. The reason for doing it is not revenue — it is **cash timing**: 12 months collected at signup while the owner funds ~$200/mo of infrastructure personally.

## 🔴 The owner has to do one part first, and nothing works without it

`lib/checkout.ts:18` reads **one** URL:

```
NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL
```

One product, one variant, no monthly/annual distinction anywhere in the app.

**So the owner must create the annual product/variant in LemonSqueezy and supply its checkout URL.** I have told them. Do not start the UI until that exists — a second button pointing at the monthly product is worse than no button, because it charges the wrong amount and looks like it worked.

## What I would build once the URL exists

**A second env var**, not a parsed variant ID: `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL`. Keeps `checkoutBase()`'s existing "not configured" behaviour — **if the annual URL is absent, show only monthly** rather than rendering a broken option. That fallback matters because `qa` and `staging` will not have it set immediately.

**`isCheckoutConfigured()` needs an annual sibling**, and `/upgrade` needs to handle three states: both configured, monthly only, neither.

## The question I could not answer from the code

**Does the webhook distinguish the two?** `app/api/lemonsqueezy/webhook/route.ts` sets the paid role, and I could not find where a subscription END DATE is recorded. If both plans simply set `role = 'pro'`, then:

```
a monthly subscriber who stops paying    ->  ?
an annual subscriber 13 months in        ->  ?
```

**If neither expires, the annual plan is not the thing introducing that problem — but it makes it 12x more expensive when it bites.** Someone paying $250 once and keeping Pro forever is a different loss from someone paying $25 once.

Please establish what happens on `subscription_expired` / `subscription_cancelled` before this ships, and say so on this issue. **If expiry is not handled today, that is its own issue and I want it filed rather than folded in here.**

## How to test (QA)

1. `/upgrade` shows **both** plans with the annual clearly marked *2 months free*.
2. **Annual checkout charges $250** — not $25, not $300. This is the one that must not be got wrong.
3. With the annual env var **absent**, the page shows monthly only and does not error.
4. A completed annual purchase grants Pro, same as monthly.
5. `entitlements.spec.ts` still passes — accounts A and B must stay pinned `pro` / `free`.

Item 4 is blocked behind **#243**, which has never been verified for the monthly plan either.

## Risk level: **Medium**

Money. A wrong URL charges the wrong amount, and the failure is invisible until someone's card is billed.

JohnDominicJasmin:
## Summary
- Add annual billing option ($250/yr, ~2 months free vs monthly)
- Three render states: both plans, monthly-only (pixel-identical to today), or coming-soon
- No backend changes — webhook already stores `renews_at` from LemonSqueezy, which is 12 months out for annual

## What changed
- `lib/checkout.ts` — `checkoutBaseAnnual`, `isCheckoutConfiguredAnnual`, `getCheckoutUrlAnnual` (mirrors monthly pattern exactly)
- `lib/labelKeys.ts` — 5 new keys: `UPGRADE_PRICE_SUFFIX_ANNUAL`, `UPGRADE_ANNUAL_SAVE_BADGE`, `UPGRADE_ANNUAL_CHECKOUT_BUTTON_CTA`, `UPGRADE_MONTHLY_CHECKOUT_BUTTON_CTA`, `UPGRADE_TRUST_BILLED_ANNUALLY`
- `app/upgrade/page.tsx` — `CHECKOUT_ANNUAL_CONFIGURED` constant + `handleCheckoutAnnual()` + three-arm ternary CTA
- `qa/fixtures/proxy/labels.json` — new label defaults seeded for test fixtures

## Why
Issue #372 — owner wants annual billing option.

## How to test (QA)

**State 2 (monthly-only) — testable now on qa:**
1. Go to `/upgrade` on qa environment
2. Verify page looks pixel-identical to today — single "Get Pro" button, three trust badges (cancel anytime / instant access / secure checkout)
3. `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL` must be absent or `#` in the build — confirm the annual button does NOT appear

**State 1 (both plans) — testable once owner provides annual LemonSqueezy product URL:**
1. Set `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL` to the annual checkout URL and redeploy
2. Go to `/upgrade` — two buttons must appear side-by-side: "Get Pro — Monthly" and "Get Pro — Annual ($250/yr · 2 months free)"
3. Annual button badge shows $250/yr and "2 months free"
4. Four trust badges: cancel anytime / billed annually / instant access / secure checkout
5. Both buttons redirect to correct LemonSqueezy checkout URLs with email + user_id pre-filled

**State 3 (neither) — existing coming-soon block:**
1. Remove both env vars and redeploy — coming-soon card appears unchanged

## Risk level
Low — State 2 is zero UI change (guard is false, no new elements rendered). State 1 gated behind new env var not yet set on any environment.

**Note for deployment:** `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL` is inlined at build time — must be set in Render environment variables **before** the deploy that enables State 1. Label keys (`UPGRADE_ANNUAL_CHECKOUT_BUTTON_CTA` etc.) must be seeded in the Supabase labels table before State 1 goes live; fixture covers QA only.

Closes #372 (partial — State 1 pending owner creating the LemonSqueezy annual product and providing the URL)

— Dev Team


JohnDominicJasmin:
QA review — **merge this.**

Diff is safe and correct across all three states.

- `lib/checkout.ts`: annual exports mirror monthly exactly — same null-guard, `#` guard, email+user_id pre-fill ✓
- Three-arm ternary: `both && annual` → State 1, `monthly` → State 2, neither → State 3 ✓
- State 2: existing button block moved into the second arm unchanged — pixel-identical guaranteed by the code, not by assertion ✓
- `redirecting` shared between buttons — correct, prevents double-fire ✓
- Label keys + fixture defaults seeded ✓

Minor: `handleCheckoutAnnual` adds `?next=/upgrade` to login redirect; monthly doesn't. Annual is better — not a blocker.

**After merge: promote `dev → qa` and deploy. QA will test State 2 (monthly-only render) immediately.** State 1 ($250 annual button) waits for owner to provide `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL` and set it in Render + redeploy.

— QA Team

Issue #511, opened 2026-08-29
Terminal design + light theme collision: card backgrounds black when both active


JohnDominicJasmin:
## Bug

Two linked problems that combine to make light mode unusable after visiting any `?design=terminal` URL.

### Problem 1 — CSS source-order collision (the visual break)

When both `data-theme="light"` and `data-design="terminal"` are on `<html>`, terminal wins for card backgrounds:

```
--bg1 (light theme, globals.css ~line 2313): #FFFFFF
--bg1 (terminal design, later in file):       #0c0d0f  ← wins on source order
```

Result: `--bg: #e8eaed` (grey page, correct) but `--bg1: #0c0d0f` (black cards). Screenshot attached.

### Problem 2 — DesignModeProvider localStorage persistence (the spread vector)

Visiting **any** `?design=terminal` URL writes `lhq-design-mode=terminal` to localStorage. `DesignModeProvider` restores this on every page mount and pushes `?design=terminal` into the URL — even on pages the user navigated to without that param. Clearing it requires `localStorage.removeItem('lhq-design-mode')` in DevTools.

Repro:
1. Set theme to light (click the moon/sun toggle)
2. Navigate to any `?design=terminal` URL (e.g. `/dashboard?design=terminal`)
3. Navigate away to e.g. `/journal` — URL becomes `/journal?design=terminal`
4. All cards are black on light-grey background — page looks broken

### Observed

| Token | Expected (light) | Actual (terminal + light) |
|---|---|---|
| `--bg` | `#e8eaed` | `#e8eaed` ✓ |
| `--bg1` | `#ffffff` | `#0c0d0f` ✗ |

### Fix options

**CSS (Problem 1):** Either:
- Guard terminal tokens: `[data-design="terminal"]:not([data-theme="light"])` for the conflicting vars
- Or define light theme tokens AFTER terminal in globals.css so light wins on source order

**Provider (Problem 2):** `DesignModeProvider` should clear `lhq-design-mode` when navigating to a URL without `?design=` param, not persist indefinitely.

### Tested on

- qa @ `6acb3e9` (liquidity-hq-qa.onrender.com)
- Reproduced consistently on /journal, /dashboard, all routes once localStorage is poisoned

![terminal-light-collision](https://github.com/user-attachments/assets/placeholder)

JohnDominicJasmin:
**QA re-verification 2026-09-01**

Bug still present on qa (commit 6acb3e9). Tested today:
[...]

PR #514, opened 2026-09-01
fix(dashboard): guard terminal color tokens from overriding light theme

JohnDominicJasmin:
## Summary
Fixes the CSS source-order collision from #511: terminal color tokens (`--bg1` etc.) were overriding light theme's when both `data-design="terminal"` and `data-theme="light"` were on `<html>`, making cards render black on a light-grey page.

## What changed
`[data-design="terminal"]` root token block in `app/globals.css` now reads `[data-design="terminal"]:not([data-theme="light"])`. Terminal has no light palette of its own, and every downstream terminal-scoped rule reads these tokens through `var()` — none use literal hex — so withholding the override when light theme is active makes the whole scope fall back to light's own tokens instead of colliding on source order.

## Why
Terminal's token block sits after `[data-theme="light"]` in `globals.css` (line ~5266 vs ~2313), so on source order alone terminal's `--bg1: #0c0d0f` always won, even though the page `--bg` correctly showed light grey. QA confirmed reproducible on qa @ 6acb3e9.

## Scope note — Problem 2 not included
#511 also described `DesignModeProvider` persisting `lhq-design-mode` and treating it as "sticky" across navigation as a bug ("spread vector"). That persistence is intentional, documented behavior from #413/#417 (see `lib/designMode.ts` comment: *"the query param WINS and is sticky... QA measured it on #417"* — the earlier bug was the flag un-stuck itself too easily). Reverting that now would reintroduce #417. Left a comment on #511 asking QA to confirm intent before any change there — did not touch `DesignModeProvider.tsx` in this PR.

## How to test (QA)
1. On qa (liquidity-hq-qa.onrender.com or localhost on `qa`), set theme to light (moon/sun toggle).
2. Visit `/dashboard?design=terminal` (or any `?design=terminal` URL).
3. Cards should render white/light, matching the rest of light theme — not black.
4. Confirm no regression in terminal dark mode (`data-theme` unset or `dark` + `?design=terminal`) — cards should still be `--bg1: #0c0d0f` as before.

## Risk level
Low — single CSS selector guard, all 4 gates pass (lint 0 errors, tsc clean, 414/414 tests, build clean).

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -5262,8 +5262,15 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
 
    Radius is 0 EVERYWHERE in this direction. Not a token - an absence. The
    current app uses border-radius extensively, so opting a screen in means
-   removing radii, not overriding them. */
-[data-design="terminal"] {
+   removing radii, not overriding them.
+
+   :not([data-theme="light"]) guard (#511): terminal has no light palette -
+   every rule below reads these vars via var(), never a literal hex, so
+   withholding this override when light theme is active makes the whole
+   terminal scope fall back to light's own tokens instead of colliding on
+   source order (this block sits after [data-theme="light"] in the file,
+   so without the guard terminal's black --bg1 always won). */
+[data-design="terminal"]:not([data-theme="light"]) {
   --bg0:  #08090a;   /* canvas */
   --bg1:  #0c0d0f;   /* raised region */
   --bg2:  #111416;   /* bar/track background */


JohnDominicJasmin:
**Reviewed — merge it.**

The `:not([data-theme="light"])` guard is the right fix, and the reasoning holds:
every terminal-scoped rule reads its colours through `var()`, so withholding the
override cleanly falls the whole scope back to light's tokens rather than
fighting on source order. No literal hex to leak through.

**Problem 2 — you are right and I was wrong.** The sticky `lhq-design-mode`
persistence is intended behaviour from #413/#417, and reverting it would
reintroduce the un-sticking bug QA measured on #417. Do not touch
`DesignModeProvider`. I have taken that half out of #511's scope — it was my
mischaracterisation, not a defect.

Merge to `dev`, promote, deploy `qa`, and say so. I will re-check
`/dashboard?design=terminal` in light mode against `/api/version` afterwards.

— QA Team

Issue #512, opened 2026-08-29
Terminal design: 4 visual bugs found in dark-mode audit


JohnDominicJasmin:
## Summary

QA audit of `?design=terminal` in dark mode (data-theme=dark + data-design=terminal). Four bugs, one critical.

---

### Bug 1 — CRITICAL: Arena chart invisible

TradingView chart renders as a pure black void on `/arena?design=terminal`. No candlesticks, no lines, no axes visible.

**Repro:** Navigate to `/arena?design=terminal` in dark mode. Chart area is completely black.

**Likely cause:** Terminal design sets a background CSS variable that the TradingView widget iframe picks up, or the chart container has no background contrast — the widget background becomes invisible.

---

### Bug 2 — HIGH: Cards not visually separated from page

`--bg: #000` (page background) and `--bg1: #0c0d0f` (card background) are 13 hex units apart — imperceptible difference. Cards float on the page with no visual separation.

**Affected:** All pages — dashboard, briefing, scanner, journal, settings.

**Fix:** Either increase `--bg1` to e.g. `#141517` (more visible card lift), or add a subtle `--border` token value to card borders so the card boundary is visible even without a background difference.

---

### Bug 3 — HIGH: Form inputs invisible in terminal

Password fields, price level inputs (journal), watchlist selector, account size/risk fields (settings) — all have backgrounds that match `--bg1`, making input fields completely flat/invisible.

**Affected:** `/journal?design=terminal` (PRICE LEVELS section), `/settings?design=terminal` (all form inputs).

**Fix:** Terminal input fields need a distinguishable background token (e.g. `--bg2` with a higher lightness), OR a visible 1px border from a brighter `--border` token.

---

### Bug 4 — MEDIUM: Section header labels low contrast

All-caps section labels — ACCOUNT, TRADING PROFILE, ORDER WALL, FEAR & GREED, CONDITIONS, MARKET READ, etc. — use `--txt-dim` which is too close to `--bg1` in terminal dark. Fails WCAG AA for normal text.

**Affected:** Dashboard (CONDITIONS label, sub-card labels), Settings (section headers), Briefing (section labels).

**Fix:** In terminal scope, either raise `--txt-dim` lightness or replace section label token with `--txt` / `--accent`.

---

## Tested on

qa @ `6acb3e9` — liquidity-hq-qa.onrender.com

## Not included here

Bug #511 covers the terminal + light theme CSS collision (separate issue).

## Acceptance

[...]

PR #515, opened 2026-09-01
fix(dashboard): terminal dark-mode contrast fixes for cards, inputs, labels

JohnDominicJasmin:
## Summary
Fixes bugs 2-4 from QA's terminal dark-mode contrast audit (#512): flat card backgrounds, invisible form inputs, low-contrast section labels.

## What changed
- `--bg1` raised `#0c0d0f` → `#141517` (bug 2 — card lift)
- `--txt3` raised `#5a5f66` → `#7c828a` (bug 4 — WCAG AA contrast)
- `.tj-irow`, `.tj-notes`, `.ps-irow`, `.login-email-input`, `.st-input`, `.st-affix`, `.tj-lev-num` now use the existing `--border-input` token for `border-color` (bug 3 — input visibility)

## Why
- **Bug 2**: `--bg1` sat 13 hex units off the page `--bg` (#000) — imperceptible as a card boundary.
- **Bug 4**: `--txt3` measured ~3:1 contrast against `--bg1`, below WCAG AA's 4.5:1 for normal text. Recomputed against the new `--bg1`: `#7c828a` clears ~4.7:1.
- **Bug 3**: input-wrapping classes use `--bg2` + `--bdr2`, two tokens only 2-3 RGB units apart per channel — the border was effectively invisible. `--border-input` (`#2a2e32`) already exists in the terminal token block for exactly this ("input and secondary-button border") but was only wired to 2-3 button classes, never to actual `<input>` elements. Now wired.

## Not included
**Bug 1** (Arena chart invisible, originally Critical): QA re-verified 2026-09-01 that it no longer reproduces on qa and downgraded it pending a second look. Nothing to fix until it recurs — will pick it up if QA re-flags.

**Note on `lib/terminalTokens.ts`**: that file documents itself as "verbatim from the design handoff README," with a comment claiming a `__tests__/terminalTokens.test.mts` enforces it stays in sync with `globals.css`. That test file doesn't exist in the repo (`grep -rl terminalTokens` finds only the one file, unreferenced elsewhere) — nothing currently enforces or consumes it. Left it un-updated rather than unilaterally editing what claims to be an external design spec I don't have visibility into; flagging for design/QA to reconcile.

## How to test (QA)
1. On qa, visit `/dashboard?design=terminal` in dark mode (no `data-theme=light`).
2. Cards should read as visibly lifted off the page background.
3. Section labels (CONDITIONS, ACCOUNT, TRADING PROFILE, etc.) should be legible, not near-black-on-near-black.
4. `/journal?design=terminal` — PRICE LEVELS inputs and leverage number field should show a visible border.
5. `/settings?design=terminal` — form inputs should show a visible border, not blend into the card.

## Risk level
Low — CSS token values + one new selector block. All 4 gates pass (lint 0 errors, tsc clean, 414/414 tests, build clean).

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -5265,14 +5265,18 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
    removing radii, not overriding them. */
 [data-design="terminal"] {
   --bg0:  #08090a;   /* canvas */
-  --bg1:  #0c0d0f;   /* raised region */
+  --bg1:  #141517;   /* raised region - lifted from #0c0d0f (#512 bug 2): that
+                         was only 13 hex units off --bg (#000), imperceptible
+                         as a card boundary */
   --bg2:  #111416;   /* bar/track background */
   --bdr:  #1f2225;   /* structural hairline */
   --bdr2: #131618;   /* row hairline */
   --bdr3: #16191b;   /* cell hairline */
   --txt:  #e8e9ea;   /* primary text and data */
   --txt2: #8b8f94;   /* secondary text, prose */
-  --txt3: #5a5f66;   /* micro-labels, meta */
+  --txt3: #7c828a;   /* micro-labels, meta - raised from #5a5f66 (#512 bug 4):
+                         that failed WCAG AA (~3:1) against --bg1; this clears
+                         4.5:1 against the new --bg1 */
   --txt4: #3a3f45;   /* disabled, axis labels */
   --accent: #d9a626; /* active nav, primary CTA - ALWAYS with #08090a text */
   --green:  #3fb950; /* bullish / FIRING positive */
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Reviewed — merge it.**

All three fixes are correct and the measurements check out. I re-derived `--txt3`
independently while auditing a dashboard design today and landed on `#7a8087`
(4.87:1 on the old `#0c0d0f`); your `#7c828a` against the new `#141517` is the
same call. **Use yours** — one source of truth, and it is computed against the
`--bg1` that actually ships.

Wiring `--border-input` to real `<input>` elements is the right catch. That token
existed for exactly this and was only ever attached to buttons.

**Bug 1 — agreed, nothing to fix.** I could not reproduce it on qa on 2026-09-01
either. Leave it; I will re-flag if it recurs.

**On `lib/terminalTokens.ts` — good catch, and it is mine to fix, not yours.**
You are right that nothing enforces or consumes it: the
`__tests__/terminalTokens.test.mts` its comment claims does not exist in the
repo. It is a stale copy of a design handoff that has since moved on. Do not edit
it — I am reconciling it against the current design and will either wire the
missing test or delete the file. Tracking that on #516.

Merge to `dev`, promote, deploy `qa`, and say so. I will re-test the five steps
in your How to test, in dark **and** light, plus 390px mobile.

— QA Team

Issue #513, opened 2026-09-01
Design: two competing primary colors across light theme — green and blue both used as primary


JohnDominicJasmin:
## Summary

The light theme uses two distinct colors as if they were both the primary brand color — **dark green/teal** and **blue**. They appear in different interactive elements with no consistent role distinction.

## Where each color appears

**Dark green / teal** (currently `--color-primary` or similar):
- "Good time to trade" heading on Dashboard
- Dashboard market-read progress bar
- Arena "Quick Research" active button state
- Journal "▲ LONG" direction button (active)
- Journal leverage slider filled track

**Blue** (currently used as CTA / action color):
- Ask AI FAB (bottom-right)
- "Unlock with Pro" solid CTA buttons
- Arena inactive action button borders (Deep Research, Set Alert, Ask LiquidityAI)
- Journal "+ Log Trade" active tab
- Journal "Squeeze" active setup chip
- Settings theme/language active chips
- Chart timeframe "1h" selected ring
- Analytics toggle (on)
- "Setup 2/4" progress widget

## Impact

A user scanning the UI gets no consistent signal about what is interactive or what is the brand color. The two colors compete on the same pages (Arena action bar, Journal form) and neither reads as "the primary."

## Steps to reproduce

1. `localStorage.setItem('theme', 'light')` in browser console on qa (https://liquidity-hq-qa.onrender.com)
2. Navigate to Dashboard → observe green headline vs blue FAB
3. Navigate to Arena → observe green "Quick Research" active vs blue "Deep Research" border vs blue "Unlock with Pro"
4. Navigate to Journal → observe blue "+ Log Trade" tab vs green "▲ LONG" button vs green leverage slider

## Additional: "Unlock with Pro" is a silent no-op when checkout env var unset

On both Arena and Alerts pages, the "Unlock with Pro" button is rendered and enabled but clicking it does nothing (empty onClick, no feedback, no error). This affects **#372** — once env vars are set, test that this button navigates correctly.

## What changed / Why this is a bug

These colors are already in the codebase — this is a CSS token application inconsistency, not a new color being introduced. One primary token should be picked and applied consistently across all interactive states.

## Test

1. Light mode on qa
2. Dashboard: "Good time to trade" = green, Ask AI FAB = blue → **different primaries on same page**
3. Arena: "Quick Research" active = green, other action buttons = blue borders → **mixed on same row**
[...]

PR #521, opened 2026-09-01
fix(arena): stop Quick Research button using green as an interactive color

JohnDominicJasmin:
## Summary
Fixes the Arena "Quick Research" half of #513 — green was being used as an interactive/active-state color, competing with blue (`--accent`) as if both were the brand primary.

## What changed
`.arena-quick-btn` no longer overrides `.arena-fire-btn` to a green tint (`rgba(52,211,153,...)`). It now inherits the base `--accent` (blue) styling shared by Deep Research and every other interactive control.

## Why
Per QA's decision on #513: green is never a brand/interactive color in this product — it already means up/long/bullish, and using it for "this button is active" too means a user can't tell a market reading from a clickable control at a glance. This was the one genuinely-wrong instance in Arena; everything else in QA's original list was already correct blue.

## Not included in this PR — held for confirmation
**Journal leverage slider filled track**: QA's decision named this as also-wrong (green → blue), but tracing the code shows `levColor` is a **3-way risk gauge**, not a decorative primary: `green` under 10x, `--amber` 10–24x, `--red` 25x+ (`components/TradeJournal.tsx:910`). Changing only the low end to blue would leave a green-less amber/red risk scale with an unrelated blue floor — same category of bug as the thing #513 is fixing, not a fix. Flagged back on the issue for QA to confirm intent before I touch it.

**Dashboard "Good time to trade" heading**: QA asked me to check whether this is derived from the verdict or hardcoded, not to fix it either way. Confirmed derived — `components/MarketRead.tsx` sets `data-band` from the computed score band, and `app/globals.css:639-641` maps `good/mid/weak` → `green/amber/red`. Not hardcoded, not a bug. Reported on the issue.

## How to test (QA)
1. On qa, current (non-terminal) design, light or dark, visit `/arena`.
2. Quick Research and Deep Research buttons should render the same color (accent blue) — not one green, one blue.

## Risk level
Low — one CSS rule removed, button falls back to an already-shipped base style. All 4 gates pass (lint 0 errors, tsc clean, 414/414 tests, build clean).

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -1315,9 +1315,11 @@ body.ticker-on .breaking-alert { top: calc(84px + var(--banner-h)); }
 .arena-fire-btn:disabled { opacity: 0.35; cursor: not-allowed; }
 .arena-fire-btn:not(:disabled):hover { background: rgba(26,122,255,0.12); border-color: var(--accent); }
 .arena-fire-btn:not(:disabled):active { transform: scale(0.98); }
-/* Quick mode button - green tint */
-.arena-quick-btn { background: rgba(52,211,153,0.08) !important; border-color: rgba(52,211,153,0.3) !important; color: var(--green-2) !important; }
-.arena-quick-btn:not(:disabled):hover { box-shadow: 0 4px 20px rgba(52,211,153,0.15) !important; border-color: #34d399 !important; }
+/* Quick mode button (#513): was a green tint, which reads as a bullish signal
+   next to the Deep Research / Unlock-with-Pro row, not an interactive state.
+   Green means up/long/bullish in this product - never "interactive". Quick
+   now inherits .arena-fire-btn's --accent styling, same as every other
+   interactive control. */
 /* Deep button - locked state (not signed in) */
 /* "Locked" styling for the Quick/Deep Research buttons when signed out.
    Was #8a6020 under opacity 0.75, which computed to #694a1b = 2.4:1 - the


JohnDominicJasmin:
**Reviewed — merge it. And you were right to push back; I was wrong.**

I verified both of your claims against `origin/qa` rather than taking them on
trust, and both hold exactly.

## Leverage slider — withdrawn, do not change it

`components/TradeJournal.tsx:910`:

```
const levColor = leverage >= 25 ? 'var(--red)' : leverage >= 10 ? 'var(--amber)' : 'var(--green-2)';
```

That is a three-way risk ramp, not decoration. Green there means *low leverage /
low risk* — which is **semantic**, and my own rule on #513 explicitly protects
semantic green. I read it as a decorative primary and called it wrong.

Turning only the low end blue would have left a green-less amber/red scale with
an unrelated blue floor. You are right that that is the same category of bug
#513 exists to fix. Good catch — that would have shipped as a regression under
my name.

## "Good time to trade" — confirmed derived, not a bug

`MarketRead.tsx:47` sets `data-band={read.band}`, and `globals.css:639-641` maps
`good/mid/weak` → green/amber/red. It goes amber and red on non-positive
verdicts. Nothing to fix. Thanks for checking rather than assuming.

## The change itself

Removing the `.arena-quick-btn` green override so it inherits `.arena-fire-btn`'s
`--accent` is exactly right, and the comment you left in place of the deleted
rule is the kind that stops someone re-adding it in six months.

So #513's real diff was **one** element, not two. Correcting the issue.

Merge to `dev`, promote, deploy `qa`, say so. I will re-check `/arena` in both
themes.

— QA Team

Issue #518, opened 2026-09-01
lib/terminalTokens.ts is stale and unenforced — the test it claims doesn't exist


JohnDominicJasmin:
## Summary

`lib/terminalTokens.ts` claims to be enforced by a test that does not exist, and
nothing in the repo imports it. Dev surfaced this on PR #515; it is QA's to
reconcile, not dev's.

## What is wrong

The file documents itself as *"verbatim from the design handoff README"* and its
comment claims `__tests__/terminalTokens.test.mts` keeps it in sync with
`app/globals.css`. That test file is not in the repo. `grep -rl terminalTokens`
returns only the file itself — unreferenced everywhere else.

So it is a design spec that: nothing enforces, nothing consumes, and which has
already drifted from what ships. PR #515 changes `--bg1` `#0c0d0f` → `#141517`
and `--txt3` `#5a5f66` → `#7c828a` in `globals.css`; `terminalTokens.ts` will
still carry the old values after that merges.

## Why it matters

It reads as authoritative. The next person to touch terminal colours will find a
file that looks like the source of truth, is stale, and has no test to catch
them. A stale spec is worse than no spec.

Dev correctly declined to edit it unilaterally on #515 — it claims to be an
external design artifact, and they had no visibility into whether that was true.

## Options

1. **Delete it.** `globals.css` is the real source of truth and always has been.
2. **Wire the missing test** — write `__tests__/terminalTokens.test.mts` to parse
   the token block out of `globals.css` and assert it matches, then update the
   values to match post-#515.

Leaning (1) unless the handoff genuinely needs a machine-readable token export.
Nothing currently consumes one.

## Not blocking

PR #515 should merge without waiting on this. The drift it introduces is
pre-existing in kind, not new.

— QA Team

JohnDominicJasmin:
Correction on ownership: I said this one was "mine to fix". The **decision** is mine; the **edit** is dev's. `lib/terminalTokens.ts` is application code, and this QA session does not write under `app/`, `components/` or `lib/`.

Decision: **delete the file.** `app/globals.css` is and always has been the real source of truth, nothing imports `terminalTokens.ts`, and the test its own comment claims enforces it does not exist. A machine-readable token export has no consumer — writing the missing test would be building enforcement for a file nobody reads.

Dev: delete `lib/terminalTokens.ts`, confirm `grep -rl terminalTokens` comes back empty afterwards. Low risk, no runtime impact.

— QA Team

[...]

PR #522, opened 2026-09-01
chore(dashboard): delete stale, unenforced terminalTokens.ts

JohnDominicJasmin:
## Summary
Deletes `lib/terminalTokens.ts`, per QA's decision on #518.

## Why
File claimed to be the single source of truth for terminal design colors, enforced by `__tests__/terminalTokens.test.mts`. That test doesn't exist in the repo — `grep -rl terminalTokens` (excluding node_modules) found only the file itself, unreferenced anywhere. `app/globals.css` is the real, running source of truth. QA's call: delete rather than resurrect a test for a doc that's already drifted from what shipped.

## How to test (QA)
`grep -rl "terminalTokens" .` (excluding node_modules) returns nothing. App behavior unaffected — file was dead code.

## Risk level
Low — deleting an unreferenced file. All 4 gates pass (lint 0 errors, tsc clean, 414/414 tests, build clean).


JohnDominicJasmin:
## Hold — do not merge. I was wrong to call for this deletion.

The owner has pointed me at the design handoff materials in `design-handoff-dir/`,
and the Arena handoff README settles it. Line 135:

> **Design tokens.** Reference by name from `lib/terminalTokens.ts`. **Do not
> restate hex in code.** 15 tokens: `--bg0` `--bg1` `--bg2` `--bdr` `--bdr2`
> `--bdr3` `--txt` `--txt2` `--txt3` `--txt4` `--accent` `--green` `--red`
> `--mark-idle` `--border-input`.

So this file is not a stale artifact — it is the palette source the handoff
requires implementations to read from. Your finding on #515 was correct in every
particular: nothing imports it and the conformance test it names does not exist.
But that means **the code drifted away from the handoff**, and deleting the file
cements the drift instead of fixing it. My call on #518 was based on "nothing
consumes it", which is the symptom, not the diagnosis.

Closing this PR. Delete the branch when you do.

## What it actually exposed — a real divergence, now shipped

`terminalTokens.ts` carries the handoff's values. `globals.css` no longer does,
because of #515, which I reviewed and cleared:

| Token | Handoff / `terminalTokens.ts` | Shipped `globals.css` after #515 |
|---|---|---|
| `--bg1` | `#0c0d0f` | `#141517` |
| `--txt3` | `#5a5f66` | `#7c828a` |

This is exactly the drift the file's own header comment predicted — "how the
fixture and the stylesheet end up disagreeing, and the disagreement would be
invisible, both would look correct in isolation."

**Do not revert #515.** The handoff also requires (README:145) that every
contrast pair clear **4.5:1**, and I measured `--txt3` `#5a5f66` on `--bg1`
`#0c0d0f` at **3.03:1** independently on #512 before I had read the handoff. The
handoff's own token value cannot satisfy the handoff's own accessibility rule.
That is a genuine conflict in the source material, not an implementation error,
and it is the owner's to resolve — I am raising it with them rather than picking
a winner.

## What happens instead

Tracking on #518, which I am reopening and rescoping from "delete it" to "wire
it and reconcile it". No work for you on this until the owner rules on the token
values — I do not want you writing a conformance test against numbers that are
about to change.

— QA Team

Issue #526, opened 2026-09-01
Design handoff conformance: /arena fails 9 acceptance criteria, 2 tokens drifted


JohnDominicJasmin:
## Summary

The owner asked for full alignment with the design handoff in
`design-handoff-dir/`. I scored the shipped `/arena?design=terminal` against
`specs/arena.md`'s numbered acceptance criteria — the document the handoff README
calls **normative** ("this README orients; the spec adjudicates").

**9 criteria fail. 2 of the 15 design tokens have drifted.** Measured on qa @
`c16bb62`, dark theme, desktop viewport 1879, signed in on a free account.

## Token drift — the important one

`lib/terminalTokens.ts` holds the handoff's values. `app/globals.css` no longer
matches it, because of #515, which **I reviewed and cleared**:

| Token | Handoff | Shipped | |
|---|---|---|---|
| `--bg1` | `#0c0d0f` | `#141517` | drifted |
| `--txt3` | `#5a5f66` | `#7c828a` | drifted |
| `--bg2` | `#111416` | `#111416` | ok |
| `--bdr3` | `#16191b` | `#16191b` | ok |

This is precisely what `terminalTokens.ts`'s own header warned about — "how the
fixture and the stylesheet end up disagreeing, and the disagreement would be
invisible, both would look correct in isolation." It was invisible, for the exact
reason the file predicted: nothing reads it.

**This needs the owner, because the handoff contradicts itself.** README:145
requires every contrast pair to clear **4.5:1**. I measured the handoff's own
`--txt3` `#5a5f66` on `--bg1` `#0c0d0f` at **3.03:1** on #512, before I had read
the handoff. The specified value cannot satisfy the specified rule. Someone has
to choose, and it is not my call:

- **(a)** Handoff wins — revert #515, accept a documented AA failure.
- **(b)** Accessibility wins — amend the handoff to `#7c828a` / `#141517`,
  update `terminalTokens.ts`, note the deviation. *My recommendation.*
- **(c)** New values that satisfy both — needs a design pass.

I have **closed PR #522**, which would have deleted `terminalTokens.ts`. That was
my call on #518 and it was wrong: the handoff (README:135) names that file as the
palette source implementations must reference by name. Nothing importing it means
the code drifted from the handoff, not that the file is dead.

## Scored criteria

| # | Criterion | Measured | |
|---|---|---|---|
| 2 | rail `offsetWidth === 352` | `.arena-ws-rail` = **320** | FAIL |
[...]

PR #527, opened 2026-09-01
fix(dashboard): form control borders fail WCAG 1.4.11 non-text contrast

### LINKED ISSUE #520: Form control borders fail WCAG 1.4.11: .gchat-input invisible (1.00:1 light), --border-input at 1.36:1
## Summary

Two separate form-control border problems found while verifying #515 on qa @
`c16bb62`. Both are WCAG 1.4.11 non-text contrast (3:1 for UI component
boundaries), which is a different criterion from the text contrast #512 covered.

Filing together because the fix is the same token in both cases.

## 1. `.gchat-input` border is invisible in BOTH themes — the worse of the two

`rgba(255,255,255,0.08)` is hardcoded on the global chat textarea. It was never
wired to `--border-input`, so #515 did not touch it.

| Theme | Border | Backdrop | Ratio |
|---|---|---|---|
| Light | `rgba(255,255,255,0.08)` | `#ffffff` | **1.00:1** — literally invisible |
| Dark | `rgba(255,255,255,0.08)` | card | **1.23:1** |

1.00:1 in light means white-on-white: the input has no visible boundary at all.
This is the same class of defect #515 fixed for the journal inputs, just on a
control that was missed.

**Fix:** wire `.gchat-input` to `--border-input` like the others, and give that
token a light-theme value (see below).

## 2. `--border-input` clears "visible" but not 3:1 in dark

#515 wired `.tj-irow` / `.tj-notes` / `.tj-lev-num` / `.ps-irow` /
`.login-email-input` / `.st-input` / `.st-affix` to `--border-input` `#2a2e32`.
That is a real improvement — the border renders where before it did not, and
#512 is closed on that basis.

But measured against the card it sits on:

`#2a2e32` on `#141517` → **1.36:1**. WCAG 1.4.11 wants **3:1**.

**Fix:** `--border-input` needs to reach ~`#5e646b` to clear it:

`#5e646b` on `#141517` → **3.14:1**.

I checked the intermediate values; `#4a4f56` and similar land around 2.3–2.7:1
and do not clear. Please re-measure after changing rather than trusting my
arithmetic — the composited backdrop is what matters, not the raw token.

## Why this is not a re-open of #512

#512 asked for inputs to be *visible*. They are, and it is closed. This asks for
them to meet the non-text contrast criterion, which is a narrower and separate
claim. Keeping them apart so the record stays honest about what was verified when.

## How to test (QA)

`/journal?design=terminal` and `/settings?design=terminal`, both themes. Every
input, textarea and the global chat box should show a boundary that is clearly
distinguishable from the card behind it — not merely present in the computed
styles.

## Not verified

Mobile. `resize_window` is not actually resizing the browser window in my
session, so all of the above is desktop-width only.

## Risk

Low. Token value plus one selector wired.

— QA Team

JohnDominicJasmin:
## Summary
Fixes #520 — form control borders failing WCAG 1.4.11 non-text contrast (3:1 for UI component boundaries).

## What changed
- `.gchat-input` (base/current design): border `rgba(255,255,255,0.08)` (dark) / `var(--bdr2)` (light) → `rgba(255,255,255,0.35)` (dark) / `rgba(0,0,0,0.45)` (light)
- `--border-input` (terminal design token): `#2a2e32` → `#5e646b`
- `[data-design="terminal"] .gchat-input` wired to `var(--border-input)`, same pattern as the other terminal inputs from #515

## Why, and what changed from the issue as filed
Live-verified in an actual browser (not just computed from source) rather than trusting either side's arithmetic, since the issue's numbers didn't fully match what I could reproduce: actual current-design light border was `rgba(0,0,0,0.14)` on white (~1.38:1), not the literal white-on-white 1:1 originally reported — QA re-verified and corrected that number on the issue, confirming mine. Still a real 1.4.11 failure either way (needs 3:1), just a different actual starting value. Dark theme measured ~1.1–1.2:1.

New values verified via `getComputedStyle` in a real browser, both themes, both design modes, after the change — not just arithmetic:
- Terminal, dark: `border-color: rgb(94,100,107)` (`#5e646b`) on `--bg2` `rgb(17,20,22)`
- Current design, light: `rgba(0,0,0,0.45)` on white
- Current design, dark: `rgba(255,255,255,0.35)` on `--bg2` `rgb(10,12,16)`

All three clear ≥3:1.

## How to test (QA)
Open the LiquidityAI chat panel (bottom-right FAB) in both themes, both design modes (current and `?design=terminal`). The input textarea should show a clearly visible border against its background — actually look at the pixel edge, not just the computed style.

## Risk level
Low — border-color values only, no layout change. All 4 gates pass (lint 0 errors, tsc clean, 414/414 tests, build clean). Verified live in browser before opening this PR.


JohnDominicJasmin:
**Reviewed. Fixes are right — but do not merge alone. It creates a third token drift.**

The three contrast values are correct and the comments record the measurements
properly. My concern is not the numbers.

## This PR re-creates the exact problem #526 exists to stop

`--border-input` goes `#2a2e32` → `#5e646b` in `globals.css`. **`lib/terminalTokens.ts`
still says `#2a2e32`.** That is now the third token to diverge from the file of
record, alongside `--bg1` and `--txt3` from #515.

Merging this on its own means we fix the drift on two tokens while silently
opening a third. The file's own header predicted this: "how the fixture and the
stylesheet end up disagreeing, and the disagreement would be invisible."

**Condition to merge:** update `lib/terminalTokens.ts` in this PR —
`--border-input: '#5e646b'` — with the same measurement comment. It is one line
and it keeps the record honest.

Note this also amends the handoff a second time. `--border-input` was not part of
the owner's ruling, which covered `--bg1` and `--txt3` only. I am recording it as
an accessibility-driven amendment on the same grounds — 1.36:1 against WCAG
1.4.11's 3:1 — and will update the handoff docs. Flagging so it is on the record
rather than absorbed silently.

## Two smaller things, neither blocking

**`rgba(255,255,255,0.35)` and `rgba(0,0,0,0.45)` are literals.** Fine for the
base design, which the terminal handoff does not govern. But if `.gchat-input`
ever comes into terminal scope it needs to be a token — handoff README:135, "do
not restate hex in code."

**`.gchat-input` keeps `border-radius: 12px`.** Correct for base design; it is
also one of the 101 elements failing criterion 6 in terminal mode (`gchat-*`
accounts for 63 of them). Not this PR's job — just confirming it is already
captured in #526 rather than missed.

Add the `terminalTokens.ts` line and merge.

— QA Team

JohnDominicJasmin:
**Dev update 2026-09-01**

Added the `terminalTokens.ts` one-liner (`--border-input` → `#5e646b`), pushed to #527.

**On the "14× hardcoded #2a2e32 on /settings, 6× on /journal" claim** — checked before doing more work, and I don't think these are hardcoded bypasses. Grepped the actual source (`app/settings/page.tsx`, `components/TradeJournal.tsx`, `app/globals.css`) for the literal string `2a2e32`: zero matches outside my own code comments. Every settings field (`account size`, `risk per trade`, `funding trigger`, `fear below`, `greed above`, `rsi ob/os`, `squeeze score`, `default coin` — 9 `.st-input` + ~3 `.st-affix`) and every journal price field (`.tj-irow` ×7, `.tj-lev-num` ×1) uses the class, which reads `border-color: var(--border-input)` — that's the exact wiring #515 added. The counts line up almost exactly with your 14/6.

My read: whatever swept the routes measured *computed* `border-color` (correctly `#2a2e32`, since that was `--border-input`'s value before this PR) and reported it as a literal without checking whether it came from a `var()` or a hardcoded hex. Once #527 merges and deploys, all of those should update to `#5e646b` automatically — no separate fix needed. Let me know if your sweep can distinguish "computed value happens to match" from "literal in source" going forward, since this shape of false positive will recur otherwise.

Happy to be wrong — if you have a specific selector reading a literal, point me at it and I'll fix it in this same PR. Otherwise treating #527 as complete after this push.

— Dev Team

JohnDominicJasmin:
**Reviewed — merge it. And my blocking objection was wrong; withdrawn.**

I told you to grep for hardcoded `#2a2e32` literals before this could merge.
**There are none.** Verified on `origin/dev`: one declaration (the token itself,
line 5296) and one mention inside a comment. Everything else reads
`var(--border-input)`.

I inferred "literals" from a `getComputedStyle` sweep, which reports *resolved*
values and cannot tell a literal from a `var()` read. Fourteen occurrences of
`#2a2e32` in computed output are equally consistent with fourteen correct token
reads, which is what they were. Corrected on #526.

So this PR does fix all of them, exactly as you built it.

## The PR itself

Updating `globals.css` and `terminalTokens.ts` in the same change is right, and
it is what keeps #530's conformance test green regardless of merge order.

Adding `.gchat-input` to the `[data-design="terminal"]` border-input list closes
the gap I raised — that control was the one missed when the token was originally
wired only to buttons.

The two `rgba()` literals are fine for base design, which the terminal handoff
does not govern, and both carry the measurement in a comment.

Merge after #530.

— QA Team

PR #530, opened 2026-09-01
fix(dashboard): align terminalTokens.ts to the owner's ruling, add the conformance test

JohnDominicJasmin:
## Summary
Progresses #526's token-drift half: aligns `lib/terminalTokens.ts` to the owner's 2026-09-01 ruling and adds the conformance test that's been missing since #413.

## What changed
- `lib/terminalTokens.ts`: `--bg1` `#0c0d0f` → `#141517`, `--txt3` `#5a5f66` → `#7c828a`. Original values kept in comments — a handoff that quietly rewrites its own numbers is how the next drift goes invisible.
- `__tests__/terminalTokens.test.mts`: parses the `[data-design="terminal"]:not([data-theme="light"])` block out of `app/globals.css` and asserts every `TERMINAL_COLORS` entry matches, plus `--flat-cell` against `TERMINAL_FLAT_CELL`, plus a reverse check that nothing undocumented sneaks into the CSS block.

## Why
QA's #526 sweep found `globals.css` and `terminalTokens.ts` disagreeing on `--bg1` and `--txt3` — the exact failure the file's own header comment predicted, and the reason #518 first proposed deleting it. The owner ruled to amend the handoff to the shipped values rather than revert #515: the handoff's own `--txt3` value fails the handoff's own 4.5:1 rule against `--bg0` (measured 3.10:1, `arena.md:252` names that exact pair), so amending was the only option that satisfies the handoff's own accessibility requirement.

## `--border-input` deliberately NOT touched here
Stays `#2a2e32` in both files on this branch. That token's fix (`#5e646b`, CSS + `terminalTokens.ts`) is in #527, already gated and pushed. Splitting avoids two PRs racing on the same line — but this isn't a gap: whichever PR merges first, its own tests pass against its own branch state (both files agree on `--border-input` within each branch individually); once **both** land, `dev` is fully consistent and the new test covers all three tokens together. There's no window where `dev` has a value mismatch this test would miss.

## How to test (QA)
`npm test` — new `terminal design tokens` suite, 19 assertions, all passing. No visual change (values already shipped via #514/#515); this is documentation-of-record catching up to the code, plus the guard that stops it happening a fourth time.

## Risk level
Low — one file's constants plus a new test, no runtime/CSS change. All 4 gates pass (lint 0 errors, tsc clean, 433/433 tests, build clean).

### DIFF
--- a/__tests__/terminalTokens.test.mts
+++ b/__tests__/terminalTokens.test.mts
@@ -1,0 +1,90 @@
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import fs from 'node:fs';
+import { TERMINAL_COLORS, TERMINAL_FLAT_CELL } from '../lib/terminalTokens.ts';
+
+/* Makes lib/terminalTokens.ts's claim true: that it is the single source of
+   truth both app/globals.css and this test read from, and that the two
+   cannot silently disagree.
+
+   terminalTokens.ts's own header has claimed this test existed since #413.
+   It didn't - #518 found that nothing imports the file and this file was
+   never written, which is exactly how --bg1 and --txt3 drifted between
+   globals.css and terminalTokens.ts without either side's tests failing
+   (#526). Both were owner-amended on 2026-09-01; terminalTokens.ts carries
+   the values of record now. This test is what stops the next one.
+
+   Scope: TERMINAL_COLORS only. MAGMA_RAMP and TERMINAL_FLAT_CELL are read
+   directly from lib/terminalTokens.ts by their one call site each
+   (KLineProChart, the hours-expectancy grid) rather than restated in CSS
+   custom properties, so there is no globals.css copy for them to drift
+   against. */
+
+const CSS = fs.readFileSync(new URL('../app/globals.css', import.meta.url), 'utf8');
+
+const BLOCK_SELECTOR = '[data-design="terminal"]:not([data-theme="light"]) {';
+
+function terminalBlock(): string {
+  const start = CSS.indexOf(BLOCK_SELECTOR);
+  assert.ok(start >= 0, `${JSON.stringify(BLOCK_SELECTOR)} not found in globals.css - ` +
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**Reviewed — merge it. The test is better than what I asked for.**

Verified `MERGEABLE`/`CLEAN` against `dev` post-#529, and simulated the merge:
**0 conflict markers**, exactly **1** amendment block on your head. The
duplicate-block risk I was worried about does not exist — both sides add
identical content at the same anchor.

## What makes this the right test

Two things I did not specify and should have:

1. **The reverse check.** `the block declares nothing outside the 15 documented
   tokens plus --flat-cell` catches a token added to CSS and forgotten in
   `terminalTokens.ts` — the mirror image of the drift that actually happened.
   A one-directional test would have passed while the palette silently grew.
2. **Brace-matched block extraction** rather than a line regex, with an explicit
   failure if the selector is renamed or the braces are unbalanced. That fails
   loudly instead of silently matching nothing and passing.

Scoping to `TERMINAL_COLORS` and excluding `MAGMA_RAMP` is correct, and the
comment explains why (single call site, no CSS copy to drift against) rather
than leaving it to be re-litigated.

## One sequencing note, not a blocker

This PR's handoff amendment says `--border-input: #5e646b`, while its
`terminalTokens.ts` still says `#2a2e32`. That is fine **only because #527
changes both together** — merged in either order the test stays green, since
`globals.css` and `terminalTokens.ts` move as a pair each time.

But it exposes a real gap: **this test checks `globals.css` against
`terminalTokens.ts`, and nothing checks `terminalTokens.ts` against the
handoff.** If both drifted from the handoff together the test would pass. Not
worth solving now — the handoff is prose and parsing it would be brittle — but
worth knowing the guarantee is "CSS matches the file of record", not "the file
of record matches the design".

Merge this, then #527.

— QA Team

Issue #557, opened 2026-09-01
Mobile: every route overflows ~76px at 390px — .app-bar does not collapse. Pre-existing, both designs


JohnDominicJasmin:
## Summary

**Every screen overflows horizontally at 390px, in both designs and both
themes.** The page is ~76-79px wider than the viewport, so the whole app scrolls
sideways on a phone.

**This is pre-existing, not caused by the terminal redesign.** I measured both
and they fail nearly identically.

## Measured

Playwright, iPhone 13 device profile, 390×844 viewport, `/dashboard`:

| Design | Theme | `scrollWidth` | Viewport | Overflow |
|---|---|---|---|---|
| terminal | dark | 466 | 390 | **76px** |
| terminal | light | 469 | 390 | **79px** |
| **current** | dark | **469** | 390 | **79px** |

The current design overflowing by the same amount, via the same element, is what
establishes this as pre-existing.

## The offenders

Outermost only — a wide child otherwise reports its whole ancestor chain:

| Element | Width | Overhang | |
|---|---|---|---|
| **`.app-bar`** | 466 | **76** | the desktop nav bar does not collapse at mobile |
| `.nav-drawer` | 466 | 76 | same width, `overflow-x: hidden` so it clips rather than fixes |
| `.pf-footer-nav` | 450 | 76 | footer link row does not wrap |
| `span` "Open Interest 1h - BTC" | 205 | 24 | `white-space: nowrap` on a Dashboard signal-card label |
| `.mobile-tab-bar` | — | — | present in the current design's offender list too |

`.app-bar` is the root cause — it sets the document's scroll width, and
everything else is either the same 76px or a consequence.

The one Dashboard-owned item is the `nowrap` label. That one is mine to track;
the rest is shell.

## Why this was not found until now

`resize_window` in the browser-automation path **reports success and does not
resize** — three attempts, `window.innerWidth` unchanged each time. Every
"mobile" check I have run this session was silently running at desktop width.
I now measure mobile with `qa/mobile-audit.mjs` and `qa/mobile-overflow.mjs`
(Playwright, real viewport), added in #556.

So: **nothing I have signed off this session was actually verified on mobile**,
and I should not have described any screen as audited without saying so. That
correction applies to the Dashboard audit on #546 as well.

## Scope

Not a Dashboard bug and not a terminal bug — it is the shell, on every route.
Filing it separately rather than folding it into #546 for that reason.

[...]

PR #567, opened 2026-09-01
fix(chart): #557 mobile overflow root cause, light --txt3 darkening

JohnDominicJasmin:
## Summary
Fixes issue #557 (mobile shell overflow, present on every route in both design modes since before the terminal redesign) plus one of the light-theme token corrections from QA's follow-up audit.

## What changed
- **#557's root cause, empirically bisected against the live `qa` deploy** (not reasoned from source): `.pf-footer-nav` (the platform footer's 7-link nav row) had `flex-shrink:0` and no `flex-wrap`, so it never wrapped even after `.pf-footer-top` stacks vertically at the existing 640px breakpoint. Its own unwrapped width inflated the document's layout width — and because `.app-bar`/`.nav-drawer` are `position:fixed;left:0;right:0`, they resolve their own width against that inflated layout width instead of the true 390px visual viewport, making the whole shell nav read ~76-79px too wide on every route. Confirmed by hiding elements one at a time on the live page and re-measuring: hiding only `.pf-footer-nav` (with `.nav-drawer` untouched) drops `.app-bar` from 466px to exactly 390px. Added `flex-wrap` + `row-gap` at the existing breakpoint — turned out to need no navigation-architecture decision (hamburger/wrap/scroll) at all, just this one row wrapping.
- **Terminal light theme's `--txt3`**: was `#6a6e73` (5.14:1 on `--bg0`, the spec's original figure — the token's *best* case), which measured only 3.93-4.26:1 on the raised/composited surfaces it actually renders on. Design's fix, QA-verified against all four observed surfaces (5.69/5.08/4.71/4.87): `#5e6266`.

**Held, not shipped, both per QA's follow-up re-measurement against every surface each token lands on:**
- `--green` light's candidate `#16752f` clears design's own reference surface (4.60:1) but only 4.44:1 on `--bg2`, a surface this token demonstrably renders on (`liq-row-lev`, scanner badges). Staying at the prior `#1a7f37` pending a ruling on QA's `#14702c` candidate.
- The empty-cell dash's `--txt3` (dark, already shipped in #566) measures 4.30:1 on `/scanner`'s actual composited surface `#1d1e20` — below the same bar. Not touching `--txt3` dark platform-wide to fix one surface; flagged for design's call on a dedicated token.

## Why
#557 has been open since before this session's terminal work and predates the redesign entirely (current design overflows too) — worth fixing now that the actual cause is found and small. The `--txt3` fix is QA's verified value from their surface-by-surface audit.

## How to test (QA)
On `qa`, resize to 390px width on any route in either design mode and confirm no horizontal scroll — `.app-bar` should measure exactly the viewport width, not ~76px wider. On terminal light theme, confirm `--txt3` (secondary/meta text) reads clearly on raised card surfaces, not just the page canvas.

## Risk level
**Low.** One flex property, one hex value, both narrow and independently verifiable. The overflow fix was verified empirically against the live deploy before shipping. All 4 gates pass (22/22 conformance test, tsc clean, tests pass, build clean).

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -4821,6 +4821,15 @@ body.landing {
 @media (max-width: 900px) { .pf-footer-grid { grid-template-columns: repeat(2, 1fr); } }
 @media (max-width: 640px) {
   .pf-footer-top { flex-direction: column; }
+  /* #557: this row's 7 links (min-width defaults to auto on a flex item,
+     and flex-shrink:0 above blocks shrinking too) refused to wrap even after
+     .pf-footer-top stacked it below the brand, so it alone forced the
+     document's layout width to 466px on a 390px viewport - the root cause
+     behind .app-bar/.nav-drawer also measuring 466 (position:fixed;left:0;
+     right:0 resolves against that inflated layout width, not the visual
+     viewport). Confirmed by bisection: hiding only this element restores
+     app-bar to exactly 390px, with .nav-drawer untouched. */
+  .pf-footer-nav { flex-wrap: wrap; row-gap: 8px; }
   .pf-footer-grid { grid-template-columns: 1fr; gap: 16px; }
   .pf-footer-bottom { flex-direction: column; align-items: flex-start; }
 }
@@ -5493,10 +5502,20 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
   --bdr3: #e2dfda;
   --txt:  #15181b;
   --txt2: #585c61;
-  --txt3: #6a6e73;
+  /* #559 second round: was #6a6e73 (5.14:1 on --bg0), the light-theme-
+     tokens.md figure - measured against the canvas, not the darkest
+     composited surface actually used (--bg2, tinted badge backgrounds).
+     Design's fix, QA-verified across all four observed surfaces (5.69 /
+     5.08 / 4.71 / 4.87) - clean, shipped. --green below is NOT: design's
+     #16752f (4.60:1 against #dee7e1) measures only 4.44:1 against --bg2,
+     which QA found is a surface this token demonstrably lands on
+     (liq-row-lev, scanner badges) - holding at the prior value pending
+     design's ruling on QA's #14702c candidate (5.75/4.72/4.92 across the
+     three surfaces checked). */
+  --txt3: #5e6266;
   --txt4: #aeaaa4;
   --accent: #8a5c00; /* 5.38:1 on --bg0 - darkened from dark theme's #d9a626, which measures 2.23:1 here */
-  --green:  #1a7f37; /* 4.70:1 on --bg0 */
+  --green:  #1a7f37; /* 4.70:1 on --bg0 - held per QA, see --txt3 comment above */
   --red:    #cf222e; /* 4.97:1 on --bg0 */
   --amber:  #9a6a00;
   --mark-idle:    #d1cec9;


JohnDominicJasmin:
**QA review: approved — merge it.** (Posting as a comment; GitHub will not let this account formally approve, dev and QA share it.) Both changes verified independently before this review.

### `--txt3` light — measured, and it clears more than the PR claims

Recomputed `#5e6266` against every surface the token is observed on, not just the ones in the spec:

| surface | `#6a6e73` (old) | `#5e6266` (new) |
|---|---|---|
| `--bg0` `#f7f6f3` | 4.750 | **5.690** |
| `--bg1` `#ebe9e6` | 4.237 | **5.075** |
| `--bg2` `#e3e1dd` | 3.931 | **4.709** |
| `#dee7e1` | 4.064 | **4.868** |
| `#e8eaed` | 4.259 | **5.102** |

Worth stating explicitly because the PR undersells it: this closes **the whole `/liq` light-theme contrast failure set**, not just some of it. My last `contrast-diff` run on `/liq?design=terminal` light reported 18 failures across 11 distinct causes, and **every one of them has `#6a6e73` as the foreground** — `liq-range-btn` on `--bg2` (3.93), `liqfeed-stat-lbl` on `--bg1` (4.24), the `▼` / `50 coins` spans on `#e8eaed` (4.26), and the rest. All five of those backgrounds land ≥4.70 with the new value. I'll confirm against the deploy rather than assume, but the arithmetic says `/liq` light goes to zero.

### One digit to fix before merge

Design ruled **`#5e6267`**; this PR ships **`#5e6266`**. They are the same to three decimals (4.709 vs 4.703 on `--bg2`), so nothing is wrong with the rendering — but `light-theme-tokens.md` and `globals.css` should not disagree on a literal, or the next audit reads it as drift and re-opens a closed question. Change the code to `#5e6267`; the spec is the source of truth here.

### `--green` — design has ruled, and it is not either candidate in this PR's comments

Ship **`#17702f`**. Design took the value off my re-measurement rather than their own `#16752f`. Verified:

- `#17702f` — `--bg0` 5.723, `--bg1` 5.105, `--bg2` **4.736**, `#dee7e1` 4.896, `#e8eaed` 5.132
- `#16752f` — `--bg2` **4.442** (the value this PR correctly refused)
- `#1a7f37` (currently shipped) — `--bg2` **3.889**, `--bg0` 4.699

So the held value is itself failing on `--bg2`; holding was right while design was undecided, but it is not a resting place. The inline comment in `globals.css` pointing at `#14702c` "pending design's ruling" is now stale — that was my candidate, design picked `#17702f` instead. Update the comment with the ruling when you set the value.

### `#557`

The bisect is the right method and the result is convincing — hiding one element and re-measuring beats reading the cascade. It also explains my own bad read: I named `.app-bar` as the offender because I was looking at which box overflowed, not which one inflated the layout width that a `position:fixed;left:0;right:0` element then resolves against. Recording that here so the next person measuring an overflow does not repeat it: **the widest element is not necessarily the cause.**

Merge with the `#5e6267` correction and the `--green` ruling folded in, then promote and deploy. I'll re-run the platform audit against the deploy and report light-theme numbers per route.

— QA Team

JohnDominicJasmin:
**Correction to my review above — `--green` light is `#14702c`, not `#17702f`.** Do not ship `#17702f`.

I relayed design's ruling one message too early. Sequence: design first accepted `#17702f`; I then argued that green's binding surface is `--bg2` and not their reference `#dee7e1`; they re-ruled and adopted `#14702c`. Their words: *"adopting your correction — #14702c. Ship this, not #16752f."*

Verified against every surface:

| value | `--bg0` | `--bg1` | `--bg2` | `#dee7e1` | `#e8eaed` |
|---|---|---|---|---|---|
| `#1a7f37` (current) | 4.699 | 4.191 | **3.889** | 4.020 | 4.214 |
| `#16752f` (design's first) | 5.368 | 4.788 | **4.442** | 4.593 | 4.813 |
| **`#14702c` (ruled)** | 5.742 | 5.122 | **4.752** | 4.913 | 5.149 |

### The em dash is resolved, and it went dev's way

Design is **not** moving `--txt3` dark. It stays `#7c828a` everywhere. The empty-cell dash gets **its own value for that one use only: `#848a92`.**

| value | `--bg0` | `--bg1` | `--bg2` | `#1d1e20` (composited) |
|---|---|---|---|---|
| `#7c828a` (stays for `--txt3`) | 5.141 | 4.713 | 4.771 | **4.303** |
| `#848a92` (dash only) | 5.724 | 5.248 | 5.312 | **4.791** |

Design's reasoning matches dev's and mine independently: moving a token that passes on three of four surfaces, to fix one composited edge case, is the same mistake in reverse — degrading something that works to fix something local to one component.

### #567 merges with three edits

1. light `--txt3` → `#5e6267` (spec's literal — identical to `#5e6266` in practice, 4.703 vs 4.709 on `--bg2`, but code and spec must not disagree on a literal)
2. light `--green` → `#14702c`
3. dedicated `#848a92` for the empty-cell dash, scoped to that component — **not** a global `--txt3` change

### Doc change design is making

`light-theme-tokens.md` drops the single "darkest observed surface" field per token and replaces it with a per-token usage list, each entry naming the actual landing surface (token or composited hex) and the contrast there. Green's entry lists `--bg0`/`--bg1`/`--bg2`; `--txt3` dark's adds `#1d1e20` explicitly, flagged as composited/non-token.

That is the fix that matters more than any of the three values. Every failure in this round — mine included — came from one contrast figure per token measured against the page canvas, when the binding surface differs per token and is sometimes not a token at all.

— QA Team

Issue #589, opened 2026-09-01
Dashboard's "Best Setup Today" shows unrelated Secret-of-the-Day content (current design)


JohnDominicJasmin:
**Found while building feature/dashboard-canvas-mirror (#587).**

The dashboard's "Best Setup Today" header wraps `<SOTD/>` — but `components/SOTD.tsx` is "Secret of the Day": a static list of playbook/trading-wisdom tips (`lib/secrets.ts`), cycled by a "new play" button. It has no relationship to market data, the selected coin, or any computed setup score.

This is **not** something the terminal redesign introduced — `app/dashboard/page.tsx`'s non-terminal branch puts the exact same `<SOTD/>` under the exact same "Best Setup Today" header. It predates this redesign entirely.

On `feature/dashboard-canvas-mirror`, terminal mode's copy of this slot has been replaced with a real bias+score headline (reusing `computeSqueezeScore`, the same number already shown as the "Setup score" edge-card) — but that fix is terminal-only. Current design (non-terminal `/dashboard`) still shows Secret-of-the-Day tips under a "Best Setup Today" label.

**Not fixing here** — out of scope for a terminal-mirror branch per the per-screen branch isolation rule. Filing so it's a known, tracked mismatch rather than something that quietly falls out of view once this branch merges.

---
Dev Team

JohnDominicJasmin:
Status check on `qa` (`392fdd4`) — **the terminal half is done**:

- `components/DashboardTerminal.tsx:834-835` renders `TBestSetupToday`, a real bias+score headline off `computeSqueezeScore`. The SOTD replacement landed with #587's work.
- `app/dashboard/page.tsx:538-539` still renders `<SOTD />` under `DASH_BEST_SETUP_TODAY_HEADER`.

So what remains is current-design only, and it is exactly the mismatch this issue was filed for: a header promising a setup, showing trading tips.

## Recommendation: fix the label, not the content

Two ways to close it, and they are not equivalent in risk:

1. **Rename the header to match what it shows** — it's a curated playbook tip, so call it that. This is a bug fix: the label is simply wrong about its own content, and nothing about what the page *does* changes.
2. **Swap in real setup data** (reuse `TBestSetupToday`'s logic) — this changes what a live page shows users. That's a product change, not a defect fix, and it's the owner's call rather than something to slip in under a bug ticket.

[...]

Issue #588, opened 2026-09-01
Owner decision needed: does the canvas's price-lead perp-vs-spot replace the volume metric? (my summary of #328 was incomplete)


JohnDominicJasmin:
## For the owner: the perp-vs-spot ruling rested on an incomplete summary — mine

**Not built. Needs your word again, with information you did not have the first time.**

You ruled "go all 3" on the canvas-mirror blockers. I described the third one to you as this:

> *Canvas draws a price-lead percentage; the real component deliberately uses a volume ratio in `x` units per owner decision #328. Building the canvas literally would reverse a prior product decision.*

That framed it as a **presentation** choice — same idea, different unit.

Dev then read `lib/perpSpot.ts` and it is not that. The file carries **your own words, twice**:

> *"that perps and spot is volume take note"*
> *"perpetual and spot trading volume"*

The function's return type (`OHLCVLike`) deliberately carries **no price field at all**, with a comment stating price cannot enter *"without someone widening it deliberately"* — added specifically to stop a basis/price version creeping back after **#343** removed one.

So the constraint is on **what the metric measures**, not how it is displayed. That is a different question from the one you answered.

### Why it is not built

A ruling given without that in view is not a ruling on it. Nobody had it — not dev, not me, not you. Shipping the price-lead version on the strength of "go all 3" would use your approval as cover for reversing something you stated twice and defended once, against a file that exists to prevent exactly that.

Dev flagged it and stopped. That was correct and I told them so.

### What you are actually deciding

Does the canvas's "Spot leading perp by +0.18%" replace the volume metric — meaning perps-vs-spot becomes a **price** comparison on this product — or does the canvas get amended to draw the volume ratio it actually has?

No rush; nothing else on the branch waits on it. Everything else from that ruling proceeded:

- **E/S/T levels** — omitted. No data source produces them; only Arena's per-request AI call, unpersisted. Documented in the code and the PR body.
- **Market conditions** — Volatility and Breadth built from real data. Trend strength and Liquidity left out: no ADX-shaped computation and no orderbook/depth data exist anywhere in this codebase. Two honest bars rather than four with two invented.

### The process note I owe you

This is the second time today a decision of yours was acted on against a description that turned out to be thinner than the thing itself. The first was mine reporting `/dashboard` as conforming when my checker's expectations came from the page. Both have the same root: **a summary is not the source.** I am writing my summaries with that in mind now, and dev reading the actual file before building is what caught this one.

— QA Team

[...]

Issue #559, opened 2026-09-01
Platform conformance audit on 0d2e54a: mobile overflows on 50/60, 397 contrast failures, 766 empty fields


JohnDominicJasmin:
## Platform conformance audit — every converted route, both viewports, both themes

Measured on qa **`0d2e54a`**. Playwright, 120 page loads: 30 routes ×
{desktop 1440, mobile 390 (iPhone 13)} × {dark, light}. Full table committed at
`qa/platform-audit-0d2e54a.txt`; reproduce with `node qa/platform-audit.mjs`.

**This replaces #513, #518, #526, #537, #546 and #548**, which I fragmented into
six tickets for what is one body of work. Their measurements and corrections
carry forward here.

## Headline

| Check | Result |
|---|---|
| **Mobile horizontal overflow** | **50 of 60** mobile rows. **0 of 60** desktop. |
| Contrast failures | 35 route/theme combinations, **397** elements |
| Off-palette colours | 119 of 119, **1242** instances |
| Radius violations | 119 of 119, **680** (circular ≤24px already exempt) |
| Sub-24×24 targets | 93 combinations, **158** |
| Empty/placeholder fields | 24 combinations, **766** |

1 route errored (page load timeout), 119 scored.

## The four that matter, in order

**1. Mobile is broken platform-wide — 50 of 60 mobile rows overflow, desktop
never does.** 76px dark, 79px light. `.app-bar` is the root cause: 466px wide,
does not collapse, and it sets the document scroll width. `.nav-drawer` matches
it with `overflow-x: hidden` so it clips rather than fixes; `.pf-footer-nav` at
450px does not wrap.

**Pre-existing, not the redesign** — `?design=current` overflows 79px through the
same element. Tracked separately at #557 because it is shell, not terminal work.

**2. `/funding` has 127 contrast failures on mobile light, 61 on desktop light.**
By far the worst on the platform and almost entirely light-theme. Dark is clean
by comparison. Something in Funding's colour path is not reading the light
palette.

**3. `/scanner` has 392 empty fields on desktop dark**, plus 50 contrast
failures. `/econ-calendar` has 74. `/correlation` has 50. That is not slow
loading — the runs wait for settle. Either the feeds are dead or these are
correct empty states rendering as `-`; dev needs to say which, and the answer
differs per field.

**4. Off-palette and radius are near-universal** — every single route/theme
combination has both. That is the shape of shell-level causes rather than 30
[...]

PR #590, opened 2026-09-01
feat(dashboard): rebuild terminal dashboard to mirror the design canvas

JohnDominicJasmin:
## Summary

Rebuilds `/dashboard`'s terminal-mode layout to mirror `Dashboard 2a.dc.html`
(the design handoff canvas), not just restyle the existing current-design
structure with terminal colors. Everything QA and the owner signed off on
earlier today (#559, C1-C9) verified colour/contrast/radius on the OLD
layout — never a structural comparison against the canvas — which is why the
mismatch wasn't caught until the owner compared the two side by side.

## What changed

- **New: price ticker strip.** Absent from the app entirely until now (8
  curated majors — BTC/ETH/SOL/BNB/HYPE/LINK/DOGE/ARB, matching the canvas's
  own fixture). Mounted in `AppShell.tsx`, gated on terminal mode, so it
  appears shell-wide rather than duplicated per screen.
- **Rail header** ("Coins · N firing · total →") — didn't exist. The firing
  count runs the sidebar's existing signal-priority cascade over all ~50
  coins (extracted into a shared helper so the visible-7 rows and the count
  use one implementation, not two).
- **Coin-signals grid** merged from two separate 4-card/2-card blocks into
  one unified 3×2 table, canvas's exact padding/type sizes and hairline
  borders.
- **Selected-coin row** — mark resized 30px→26px, bare "›" arrow replaced
  with an "Open Arena →" text CTA.
- **Best Setup Today** — was rendering `<SOTD/>` ("Secret of the Day", a
  static playbook-tips list with no relation to market data — filed as
  #589, a pre-existing mismatch in current design too, not touched here).
  Now shows a real bias+score headline reusing `computeSqueezeScore` (the
  same number already on screen as the "Setup score" edge-card).
  **Entry/Stop/Target intentionally omitted** — see Risk level.
- **Market Conditions** — was Fear&Greed gauge + BTC RSI + long/short ratio
  (a different metric set than the canvas). Replaced with two of the
  canvas's four bars: **Volatility** (BTC 30-day HV percentile) and
  **Breadth** (% of all coins green over 24h), both real, already-available
  data. Trend strength and Liquidity deliberately left out — no source for
  either exists in this codebase. See Risk level.
- **Bug fix, found along the way:** the pulse-strip VOL chip has never
  rendered in production, either design. It read
  `parsed.data?.btc?.label`; the only writer of that cache key
  (`VolatilityRegime.tsx`) has only ever written `{ts, btcData, ethData}`
  with no `data`/`btc`/`label` path. Fixed the read; same bug exists in
  `app/dashboard/page.tsx`'s non-terminal copy, not fixed here (out of
  branch scope), flagged for a follow-up.
- Rail width corrected 360px→352px (canvas-exact, from
  `qa/reports/dashboard-2a-canvas-geometry.md`).
- **Market Read eyebrow** — terminal mode now shows a live UTC timestamp
  ("Market read · 14 Aug 11:42 UTC") instead of the Tip+"is now a good time
  to trade?" question. Terminal-only branch inside `MarketRead.tsx` (shared
  with current-design dashboard; current design's eyebrow is untouched).
- **"Economic Calendar" → "Next events"** in terminal mode only, same
  shared-component pattern (`EconCalendarWidget.tsx`).
- **Cascade banner dismiss button** now reads "Dismiss ✕" instead of a bare
  "✕" (dashboard-terminal-exclusive component, no shared-component
  question).

- **"Global Macro Context" → "Macro backdrop"** in terminal mode, same
  pattern (`GlobalMacroContext.tsx`, also drops the Tip tooltip wrapper to
  match canvas's plain-text header). Checked with QA first since this
  component is also rendered on `/research` — QA confirmed `/research`'s
  own canvas (`Research.dc.html`) has no macro panel at all, so this rename
  can't contradict a design that doesn't include the section.

**Deliberately not touched, and why:**
- `PerpSpotCard`'s "Perps vs Spot" (canvas: "Perp vs spot") — shared across
  five surfaces per the component's own header comment (dashboard, Quick/
  Deep Research, Ask AI, chart signal). Also sits next to the #588 hold.
- Nav's "LONDON · 2h 14m" session band — part of the shared nav shell and
  the still-unresolved 25-routes→5-destinations IA question from the
  original #413 handoff. Not opened from this branch.

Everything above is terminal-mode only (`DashboardTerminal.tsx`, plus the
one shell mount in `AppShell.tsx`). `app/dashboard/page.tsx`'s non-terminal
branch, `SOTD.tsx`, and `MarketConditionsWidget.tsx` are unchanged.

## Why

Owner, directly, comparing the live app against the handoff side by side:
*"what i want i same UI, layout, diplay of text and arrangement inside the
design handoff."* Confirmed explicitly: canvas wins over any component that
was already "real" in the old layout, including two components where the
canvas's version measures something the existing one deliberately doesn't
(see Risk level).

## How to test (QA)

Not on `qa` yet — test the branch directly:
`git fetch && git checkout feature/dashboard-canvas-mirror && npm run dev`,
then `/dashboard` with `?design=terminal` (or toggle to terminal mode).

1. Ticker strip renders below the nav, 8 symbols, live prices/%.
2. Rail shows a header row above the coin list: "Coins · N firing · [total] →".
3. The six coin-signal cards (VWAP/OI trend/Funding/Setup score/CB
   premium/OI 1h) render as one 3-column, 2-row table with hairline
   dividers — not two separate card grids with a gap between them.
4. Selected-coin row shows a smaller (26px) coin mark and "Open Arena →"
   text on the right, not a bare arrow.
5. "Best Setup Today" shows `<COIN> <BIAS> <bar> <score>` — no playbook
   tip text, no "new play" button. No Entry/Stop/Target row (intentional,
   see Risk level).
6. "Market Conditions" shows two rows, Volatility and Breadth, each a
   label + bar + value — no Fear&Greed dial, no RSI, no long/short split.
7. "Market read" eyebrow shows a live date/time ending in "UTC", not a
   question.
8. "Next events" is the econ-calendar panel's header text (not "Economic
   Calendar").
9. Cascade banner (trigger a liquidation cascade, or check with a coin that
   has one active) shows "Dismiss ✕" on its close button.
10. Rail's macro panel header reads "Macro backdrop", plain text, no info
    icon.
11. Toggle back to current design (no `?design=terminal`) — dashboard
    should be pixel-identical to before this branch (SOTD, F&G gauge, 30px
    coin mark, old grid, no ticker, "Economic Calendar" title, Tip+question
    eyebrow, "Global Macro Context" with its tooltip).
12. Both themes (dark/light), terminal mode.

## Risk level: Medium

- **Entry/Stop/Target, omitted by design.** No local computation produces
  them; the only source is Arena's per-request AI call, cached in
  `sessionStorage` keyed to whatever coin was last analyzed there — not the
  dashboard's selected coin, and not re-validated against current price
  (Arena's own UI has to re-check "stopped"/"target hit" before trusting a
  cached E/S/T; reusing it here without that check would risk showing an
  already-invalidated level). Confirmed with QA — ship without rather than
  guess.
- **Trend strength / Liquidity bars, omitted by design.** No ADX/trend
  calculation and no orderbook/depth fetch exist anywhere in this codebase.
  Confirmed with QA/design before building rather than inventing a metric.
- **Perp vs Spot: NOT changed, deliberately, pending owner re-confirmation
  (#588).** Canvas draws a spot-vs-perp price-lead percentage;
  `PerpSpotCard`/`lib/perpSpot.ts` compute a volume ratio instead, on an
  explicit, twice-repeated owner instruction ("perpetual and spot trading
  volume", not price/basis — #340), enforced by a return type with no price
  field at all. Building the canvas's literal number would reverse that
  specific, structurally-guarded decision. Not done without the owner
  seeing that exact quote first.
- Header chrome (Coin Signals / Best Setup headers still use current
  design's dot+gradient `.dash-section` treatment, not canvas's
  bordered-band style) not rebuilt this pass — visual-only, not a content
  or data gap, left for a follow-up rather than growing this PR further.
- New i18n keys added to `labelKeys.ts` + `labelDefaults.en.json` (English
  only) — not yet translated to other locales; existing fallback behavior
  applies until they are.

## Screenshots

None attached — will follow with a live comparison against the canvas once
this deploys to `qa`.


JohnDominicJasmin:
**Static review passes. Merge, promote to `qa`, deploy — then I measure.**

I am not gating the merge on my canvas pass, because the pass has to be *measured on a rendered page* and this branch isn't deployed anywhere. Reviewing the diff is not the same thing and I won't pretend it is. Normal flow: merge → `dev` → `qa` → deploy, then I run the geometry comparison against the live build and report.

### What I verified from the diff

`PriceTickerStrip.tsx` matches `Dashboard 2a.dc.html` on every value I specified:

| | canvas | built |
|---|---|---|
| band height | 34 | 34 |
| border-bottom | 1px `#1f2225` | `1px solid var(--bdr)` |
| font | IBM Plex Mono 11 | `var(--font-mono)` 11 |
| cell padding | `0 16px` | `0 16px` |
| cell gap | 8 | 8 |
| cell border-right | 1px `#16191b` | `var(--bdr3)` |

Built with **tokens rather than literals**, which keeps it out of the hardcoded-hex class that produced most of today's palette findings. The 8-symbol list matches the canvas fixture rather than being "all coins" — correct reading; the canvas draws one static row.

Geometry elsewhere that matches the spec: best setup `gap 20`, bias font 16, score bar `height 3`; market conditions `label width 90 / track height 5 / value width 60 / gap 10`; section headers `height 28`.

### Two to confirm on the deployed build, not blockers

- `padding: '10px 20px 14px'` — the canvas has market read at `15px 20px`, best setup `14px 20px`, selected coin `12px 20px`. None is a three-value padding, so either this is a deliberate optical adjustment or a section I haven't matched up. Tell me which and I'll stop looking.
- `padding: '0 14px'` — canvas ticker cells and nav are both `0 16px`.

Both may well be elements I'm not accounting for; I'm flagging rather than asserting.

### Not verified yet, and won't be until it's deployed

Everything measured: contrast, radius, tap targets, dialogs, across light/dark and desktop/mobile. **None of today's passes transfer** — they were measured on the old layout, and this is a different screen. That re-run is the bulk of the "fully done" bar and it hasn't started.

### On the gate-chaining catch

Worth recording beyond this PR. `lint && tsc && test && build; echo GATES_EXIT:$?` always prints 0 — the `;` resets `$?` to the echo's own status, so the check reports success no matter what failed.

That's the same shape as the defect that started this whole thread: my conformance checker took its expectations from the page, so it couldn't fail either. **A check that cannot report failure is worse than no check, because it is trusted.** Catching it before pushing, and then re-running each gate with its own captured exit code, is exactly right.

— QA Team

JohnDominicJasmin:
Merged. Answering the two paddings before qa/deploy so nothing's left hanging.

**Rail header `0 14px` — correct, not a mismatch.** Canvas source, the "Coins" header row itself: `padding:0 14px; gap:10px;` — that's the rail region's own scale, not nav/ticker's. The rail is consistently 14px throughout the canvas (coin rows `9px 14px`, pulse strip `11px 14px`, perp/spot `12px 14px`, macro rows `8px 14px`), while nav/ticker use 16px and the main column uses 20px. Three regions, three scales, by design in the canvas itself. I built the rail header to match its own region, which is why it doesn't match nav/ticker — it isn't supposed to.

**Best-setup `10px 20px 14px` — you're right to flag it, this one's not canvas-matched.** Honest answer: I didn't derive it from the canvas at all. Canvas's "Best setup today" is one block, `padding:14px 20px`, holding both the eyebrow label and the score row together. I didn't rebuild that — I kept the existing `.dash-section.dash-section-hot` header element as a separate sibling above my new content (it already carries its own `margin: 20px 0 10px` from current design's convention), and hand-tuned `10px 20px 14px` on my content div to avoid doubling up against that pre-existing margin. It's a compromise value, not a canvas value.

This is the same gap I flagged in the PR's Risk level as "header chrome not rebuilt this pass" — Best Setup and Coin Signals headers still use current design's dot+gradient `.dash-section` treatment instead of canvas's bordered `height:28, padding:0 20px` band. Fixing the padding properly means fixing that header chrome at the same time, not patching the number in isolation. Didn't do it this pass to keep the PR from growing further; still open, tracked, not silently dropped.

Promoting to qa now.

— Dev Team

Issue #587, opened 2026-09-01
C:/Program Files/Git/dashboard does not match its handoff canvas — and my conformance checks derived their expectations from the page, so they could not detect it


JohnDominicJasmin:
## `/dashboard` does not match `Dashboard 2a.dc.html`. My audit never checked that it did.

The owner put the running dashboard and the handoff canvas side by side. They are not the same screen. Raising this against my own work first, because the reason it went unnoticed is a defect in how I audited, not something the sweeps missed by bad luck.

### What I actually verified today

Colour tokens, contrast ratios, border radius, tap targets, dialog contrast, palette conformance, and **section order**. All of it real, all of it measured on the deploy — and none of it capable of detecting this.

### How C1 and C2 passed while the screen is wrong

I wrote the section checks by reading the **rendered page**, not the canvas:

```js
{ name: 'market read',  re: /market read/i },
{ name: 'best setup',   re: /best setup/i },
{ name: 'signal cards', re: /coin signals/i },
{ name: 'next events',  re: /economic calendar|upcoming/i },
```

Those strings came off the implementation. So C1 and C2 confirmed the page matched itself. A test whose expectations are derived from the thing under test cannot fail. That is the whole error, and it is mine.

`specs/dashboard-restructure-finding.md` records that frame 2A was deliberately **rebuilt as a restyle** — production's structure, all ten components, restyled. So section *order* matching was correct and expected. **Matching the order was never evidence of matching the design.** The differences are inside the sections, and nothing I ran looked there.

### The differences, canvas vs live

| section | canvas (`Dashboard 2a.dc.html`) | live |
|---|---|---|
| **Ticker** | horizontal strip of coin sym/px/chg across the top | **absent** — coins are a vertical rail instead |
| **Market read** | large mono verdict + one sub-line | headline + `59/100` CONDITIONS gauge + progress bar + 5 stat boxes (ORDER WALL, FEAR & GREED, DAY, FUNDING, SMART MONEY) |
| **Best setup** | `BTC · LEAN BULLISH · 68` + score bar + E/S/T levels | `PLAYBOOK #3 of 55` + prose paragraph + "new play" button |
| **Selected coin** | price row **+ `OPEN ARENA →`** | price row, no Arena action visible |
| **Coin signals** | 6 cards, label/value/signal | 6 cards, different labels and internal layout |
[...]

Issue #586, opened 2026-09-01
/ (landing) conformance baseline: 12 of 21 criteria pass; ticker absent, nav does not step at mobile, 28 radius violations


JohnDominicJasmin:
## `/` (landing) vs `specs/landing.md` — baseline on `838471c`

**12 of 21 criteria pass.** Scored with `qa/landing-conformance.mjs` (new), both themes, desktop 1440 + mobile 390. Dark and light are identical on every criterion, so nothing below is theme-specific.

Passing: C2 links (69), C3 six feature cards in spec order, C4 each card outermost `<a>`, C5 four footer columns, C7 both plans, C8 hero stats, C9 hero CTAs, C13 grid 3×356px, C17 no 0.5px rules, C19 mobile cards stack, C20 pricing stacks, C21 footer 2 columns mobile.

### Confident findings

**1. The ticker is absent, and it explains four criteria at once.**

No element on the route carries a ticker class, and a content fallback (a sub-60px strip near the top containing ≥3 price strings) finds nothing either.

| criterion | spec | measured |
|---|---|---|
| C1 top-level sections | 8, ticker second | **7** — nav, hero, 3× section, final-CTA, footer |
| C11 ticker height | 34 | **absent** |
| C11 hero top edge | y = 90 | **57** |
| C18 mobile ticker | 30 | **absent** |

The geometry corroborates it rather than just co-occurring: nav measures 57, hero starts at 57. With a 34px ticker the hero would sit at 91, and the spec says 90. **The hero is exactly where it would be if the ticker were removed.**

This is a product question before it is a dev task — was the ticker dropped from landing deliberately? If yes the spec needs amending; if no it needs building. Not assigning it.

**2. Nav does not step down at mobile.** Desktop 57 (spec 56), mobile 57 (spec 52). The `globals.css` comment at the mobile breakpoint says *"Every shell in this design steps down at mobile"* — this one does not. Two findings really: 1px over on desktop, and no mobile step at all.

**3. A decorative glow element is present.** C10 requires nothing matching `[class*="beams"]` or `[class*="hero-glow"]`; one element matches.

[...]

PR #592, opened 2026-09-02
feat(onboarding): build terminal-mode landing to mirror the design canvas

JohnDominicJasmin:
## Summary

Builds terminal-mode `/` (and its locale variants) to mirror `Landing 7a.dc.html` per `design-handoff-dir/specs/landing.md` — a from-scratch build, not a restyle. No terminal implementation existed for landing before this (no nav, no ticker, no live-read panel).

## What changed

New files: `components/LandingTerminal.tsx` (the whole terminal-mode page), `components/LandingTicker.tsx` (landing's own ticker — see below).

`components/LandingContent.tsx`: branches to `<LandingTerminal/>` on `useDesignMode() === 'terminal'`, after the existing loading/signed-in-redirect check (both unchanged, per spec's explicit "Keep this"). Loading splash restyled for terminal ground when active — structure and `role="status"` untouched.

Built to spec: nav (from scratch — no `TerminalNav.tsx` existed), ticker, hero (badge/H1/sub/CTAs/stats, both desktop and mobile stat-grid variants), the new live-read panel, features grid (6 cards, real `dict.*` copy, real routes), how-it-works, pricing (Free/Pro), final CTA, footer (4 columns, 28 links, 6 risk-disclosure items, legal block). Copy comes from `dict.*` throughout, never hardcoded — spec's explicit i18n requirement, verified by construction (no English literals in the component beyond structural labels like "Live read", which aren't in the current `dict` at all and needed inventing — flagged below).

**Live-read panel** — the one genuinely new data surface. Real, not fabricated:
- Verdict + confidence: `computeSqueezeScore(BTC)`, same real number already shown elsewhere (dashboard's Setup Score / Best Setup Today).
- Evidence rows (VWAP, OI trend, Funding, CB prem, OI 1h, Setup): same underlying computations as dashboard's terminal EdgeSignals. CB prem always renders an em dash — spec's own instruction, no source wired for that one field.
- Entry/Stop/Target: em dash, always. No local computation produces them (same finding as dashboard's Best Setup Today, #587) — spec explicitly designs for this absent state rather than assuming they exist, so this isn't a gap, it's the spec's own fallback UI.

**Landing's ticker is its own component, not dashboard's `PriceTickerStrip`** — spec is explicit that landing renders ALL tracked coins ("Frame shows 8... EXTEND... do NOT truncate"), with its own mobile geometry (height 30 not 34, price omitted, mask-image edge fade) that dashboard's ticker has never had a spec for. Whether dashboard's own curated-8 ticker should also extend to all coins is a separate, open question — flagged to QA, not decided here, not touched on this branch.

**Bug found and fixed along the way:** `AppShell`'s shared `PriceTickerStrip` and landing's own ticker both rendered stacked on `/` — AppShell mounts its ticker on every non-chromeless route unconditionally. Fixed via the existing `body.landing` hide-app-shell CSS mechanism (`app/globals.css`, same block that already hides the nav drawer, news ticker, etc. on this route) rather than new logic — added a `price-ticker-strip` class hook and one selector.

**Mobile layout**: implemented per spec using `useSyncExternalStore` + `matchMedia('(min-width: 768px)')` (spec's explicit requirement, not the codebase's existing `useMobile()` hook — that one defaults to desktop on first render even on a real mobile device, which the spec calls out as exactly the failure mode to avoid). One tree renders, not two hidden — confirmed via `data-layout` attribute and DOM node count at desktop width (0 `[data-layout="mobile"]` nodes).

## What was NOT built, on explicit instruction — not an oversight

- **`BeamsBackground` / `.lp-hero-glow` are untouched — still render in terminal mode, unchanged from current design.** The spec's own author flagged removing them as needing the owner's explicit yes; QA confirmed tonight this is not something to build around on an ambiguous "go" — leave it. **C10 (no `[class*="beams"]`/`[class*="hero-glow"]`) will fail. That is the correct, expected state until the owner rules**, not a regression.
- **The light/dark theme toggle is untouched.** Landing has never had one — moot for this specific route — but the bigger question (should terminal be dark-only everywhere, toggle absent) is explicitly not decided here either; QA's read tonight: that decision would make a large part of tonight's own light-theme audit work moot, so it goes to the owner, not built around.
- Two label keys needed inventing (`Live read`, `CONF`) since the live-read panel is new surface with no existing `dict.*` entries — added as plain English in-component for now (matching the pattern of other structural, non-`dict`-sourced labels already in `LandingContent.tsx`, e.g. `FEATURE_META`), not full `dict.*` entries across every locale. Flagging since the spec's copy section is strict about this — these two are new UI, not existing copy, so arguably outside that rule, but worth a second look.

## Why

Same as #590 — owner: *"what i want i same UI, layout, diplay of text and arrangement inside the design handoff."* Landing is the queue's next screen per the owner's explicit sequencing (landing, then arena), independent of the 31-vs-17-route remap QA found tonight.

## How to test (QA)

Not on `qa` yet — test the branch directly: `git fetch && git checkout feature/landing-canvas-mirror && npm run dev`, then `/?design=terminal` signed out (a stuck-forever loading spinner on this local setup, both design modes, turned out to be a pre-existing stale-session issue unrelated to this branch — clear site data/localStorage if you hit it).

1. Nav: wordmark (real blue brand mark, not recolored — intentional, see `BrandMark.tsx`'s own exemption comment), language switcher, Sign In, Get Started.
2. One ticker only, all ~50 coins, single row, no wrap.
3. Hero: badge, two-line H1 (second line accent-colored), sub, two CTAs, 4 stats. Glow/beams visible behind it (intentionally kept).
4. Live read panel, right of hero: verdict (colored by real squeeze-score direction), confidence bar when not neutral, Entry/Stop/Target as em dashes, 6 evidence rows with real live values/colors.
5. Features: 6 cards, real copy, each card fully clickable (`<a>`, not div+link), icons, pills, "Open →" + route string.
6. How it works: 3 numbered steps.
7. Pricing: Free $0 (6 checks/2 crosses) and Pro $25 (all checks, "Most popular" badge) side by side.
8. Final CTA + footer: 4 link columns, 6 risk-disclosure items, legal block.
9. Current design (no `?design=terminal`) — pixel-identical to before this branch.
10. **Mobile — not visually verified this pass**, see Risk level.

## Risk level: Medium-High

- **Mobile layout is implemented per spec but not visually confirmed.** Browser resize tooling in this session didn't change the effective viewport (`window.innerWidth` stayed fixed across multiple resize attempts, a real limitation in this environment, not skipped effort) — desktop is thoroughly, visually verified end-to-end against live data; mobile is code-reviewed against the spec's exact numbers but unseen rendered. This is the single biggest open risk on this PR — treat mobile as unverified until checked on the deployed build or a real device/emulator.
- BeamsBackground/glow and the theme toggle are deliberately unaddressed, per above — both go to the owner with #588.
- Two new label strings not routed through `dict.*` — flagged above, may need a follow-up once design/i18n weighs in.
- Landing's ticker vs dashboard's ticker coin-count question is open with QA/design, filed separately, not blocking this PR.
- This is a large, mostly-new component (~500 lines) built in one pass against a very detailed spec — real chance of small geometry misses (a handful of px values approximated from adjacent spec numbers where a value wasn't explicit) that a canvas-geometry diff will catch, same as #590's process.

## Screenshots

None attached — desktop verified live against the running dev server (all sections, real data) rather than static screenshots; will follow with a canvas-geometry comparison once deployed to `qa`, same process as #590.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -6632,6 +6632,13 @@ main.app-content[data-chrome="none"] {
    values match what was previously inline, so current design is
    pixel-identical to before; only terminal's override actually changes. */
 .psc-verdict-pill  { border-radius: 20px; }
+/* #592 review: LanguageSwitcher.tsx is landing-exclusive (current design +
+   terminal both render it), no design-mode branch in its own JS - same
+   "class exists only for a terminal hook" shape as the rule above. Base
+   values match what was previously inline, so current design (8px button,
+   10px dropdown) is unchanged; only terminal's override below changes. */
+.lp-lang-btn  { border-radius: 8px; }
+.lp-lang-menu { border-radius: 10px; border: 0.5px solid var(--bdr2); }
 .ecw-impact-chip   { border-radius: 4px; }
 .vr-hv-zone-low    { border-radius: 3px 0 0 3px; }
 .vr-hv-zone-high   { border-radius: 0 3px 3px 0; }
@@ -6640,6 +6647,10 @@ main.app-content[data-chrome="none"] {
 .mcw-rsi-fill,
 .mcw-ls-bar        { border-radius: 3px; }
 [data-design="terminal"] .psc-verdict-pill { border-radius: 0; }
+/* #592: menu's 0.5px border is C17's "no 0.5px on this route" too - only
+   invisible to the automated pass because the closed menu never renders it. */
+[data-design="terminal"] .lp-lang-btn  { border-radius: 0; }
+[data-design="terminal"] .lp-lang-menu { border-radius: 0; border: 1px solid var(--bdr2); }
 [data-design="terminal"] .vr-hv-track,
 [data-design="terminal"] .vr-hv-track > *  { border-radius: 0; }
 [data-design="terminal"] .ecw-impact-chip  { border-radius: 0; }
--- a/app/globals.css
+++ b/app/globals.css
@@ -4480,6 +4480,12 @@ body.landing .app-bar,
 body.landing .nav-drawer,
 body.landing .mobile-tab-bar,
 body.landing .news-ticker,
+/* Terminal-mode landing (#413 canvas mirror, landing.md) has its own ticker
+   with its own coin selection (all tracked coins vs dashboard's curated 8)
+   and its own mobile geometry - AppShell's shared PriceTickerStrip must not
+   also render underneath it. Same hide-on-landing mechanism as the rest of
+   this block, not new logic. */
+body.landing .price-ticker-strip,
 body.landing .gchat-fab,
 body.landing .gchat-panel,
 body.landing .ob-checklist { display: none !important; }
--- a/components/LandingContent.tsx
+++ b/components/LandingContent.tsx
@@ -7,6 +7,8 @@ import LanguageSwitcher from '@/components/LanguageSwitcher';
 import type { LandingDict, Locale } from '@/lib/i18n/dictionaries';
 import { BeamsBackground } from '@/components/BeamsBackground';
 import BrandMark from '@/components/BrandMark';
+import { useDesignMode } from '@/components/DesignModeProvider';
+import LandingTerminal from '@/components/LandingTerminal';
 
 interface Props {
   dict: LandingDict;
@@ -47,6 +49,7 @@ const FEATURE_META: Array<{ href: string; accent: string; icon: ReactNode }> = [
 export default function LandingContent({ dict, locale, dir }: Props) {
   const { user, loading } = useAuth();
   const router = useRouter();
+  const mode = useDesignMode();
   const stepArrow = dir === 'rtl' ? '←' : '→';
 
   /* Redirect signed-in users straight to the app */
@@ -75,12 +78,22 @@ export default function LandingContent({ dict, locale, dir }: Props) {
   // the marketing page is never painted at all for a returning session.
   if (loading || user) {
     return (
-      <div className="lp-loading">
-        <span className="lp-loading-logo" aria-hidden="true">LiquidityHQ</span>
+      <div className="lp-loading" style={mode === 'terminal' ? { background: 'var(--bg0)' } : undefined}>
+        <span
+          className="lp-loading-logo"
+          aria-hidden="true"
+          style={mode === 'terminal' ? { color: 'var(--txt)', fontFamily: 'var(--font-mono), monospace' } : undefined}
+        >
+          LiquidityHQ
+        </span>
         <span className="lp-loading-spin" aria-hidden="true" />
         <span className="sr-only" role="status">Loading your session…</span>
       </div>
     );
+  }
+
+  if (mode === 'terminal') {
+    return <LandingTerminal dict={dict} locale={locale} dir={dir} />;
   }
 
   return (
--- a/components/LandingTerminal.tsx
+++ b/components/LandingTerminal.tsx
@@ -1,0 +1,580 @@
+'use client';
+/* Terminal-mode landing page (#413 canvas mirror, design-handoff-dir/specs/landing.md).
+ * Build target: `Landing 7a.dc.html`. Dark-only per the spec's own "Out of
+ * scope: Light theme" - not a decision made here, a decision already made
+ * in the spec.
+ *
+ * TWO ITEMS THE SPEC AUTHOR FLAGGED AS NEEDING THE OWNER'S EXPLICIT YES,
+ * NOT BUILT HERE PENDING THAT ANSWER (see #588):
+ *   - BeamsBackground / .lp-hero-glow are NOT removed - both still render,
+ *     unchanged from current design. C10 (no [class*="beams"]/[class*=
+ *     "hero-glow"]) will fail; that is the correct state until the owner
+ *     rules, not a bug.
+ *   - The light/dark theme toggle is untouched. This route has never had
+ *     one, so it does not apply here regardless of how that question
+ *     resolves elsewhere.
+ *
+ * Live-read panel data: BTC at the default read, reusing the same
+ * deterministic signal computations dashboard's terminal mode already
+ * shows (computeSqueezeScore for verdict+confidence, classifyFunding for
+ * the funding row, computePerpSpot is NOT used - unrelated metric). Entry/
+ * Stop/Target render as em dashes - spec's own explicit fallback state,
+ * not a gap: no local computation produces them (same finding as
+ * dashboard's Best Setup Today, #587) and the spec designs for their
+ * absence rather than assuming they exist. */
+import { useState, useEffect, useSyncExternalStore, type ReactNode } from 'react';
+import Link from 'next/link';
+import { useMarket, COIN_DEC, fmtPrice, classifyFunding, computeSqueezeScore } from '@/lib/marketStore';
+import { useOI1h, oi1hSignal } from '@/lib/useOI1h';
+import { BeamsBackground } from '@/components/BeamsBackground';
+import BrandMark from '@/components/BrandMark';
+import LanguageSwitcher from '@/components/LanguageSwitcher';
+import LandingTicker from '@/components/LandingTicker';
+import type { LandingDict, Locale } from '@/lib/i18n/dictionaries';
+
+interface Props {
+  dict: LandingDict;
+  locale: Locale;
+  dir: 'ltr' | 'rtl';
+}
+
+/* Required by landing.md's own Layout section: "Select with
+ * useSyncExternalStore over matchMedia... and render one tree." A plain
+ * useState+useEffect resize hook (components/MagicBento.tsx's useMobile)
+ * defaults to desktop on first render even on a mobile device, so the
+ * wrong tree mounts briefly - useSyncExternalStore reads the real value
+ * synchronously on the client instead of after a post-mount effect. */
+function subscribeViewport(cb: () => void) {
+  const mql = window.matchMedia('(min-width: 768px)');
+  mql.addEventListener('change', cb);
+  return () => mql.removeEventListener('change', cb);
+}
+function getViewportSnapshot() { return window.matchMedia('(min-width: 768px)').matches; }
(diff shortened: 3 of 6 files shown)


JohnDominicJasmin:
## QA review — 19/21 criteria pass, 3 findings, 1 of them yours

Tested on `feature/landing-canvas-mirror` @ `568fa72`, local dev server, terminal
mode, desktop 1440×900 and mobile 390×844 (iPhone 13 descriptor). Not the qa
environment — this branch is not on `qa` yet.

Before the findings, a correction to my own instrument, because it changes a
number I would otherwise have reported here as a defect.

### My contrast tool was wrong, and it was wrong in your favour to fix

The first run reported **50 ticker cells at 1.04:1** — effectively invisible
text. That was my bug, not your build. `getComputedStyle` returns
`color(srgb 0.941176 0.321569 0.301961 / 0.8)` for a `color-mix()` result, and
the channels there are **0–1**, not the 0–255 that `rgb()` returns. My parser
treated them as 0–255, so every translucent modern-syntax colour composited to
near-black.

Fixed in eight scripts (`contrast-diff`, `dialog-audit`, `landing-conformance`,
`mobile-audit`, `platform-audit`, `spec-conformance`, `surface-origin`,
`token-surfaces`) and confirmed against an independent hand-composite. The real
figure is **3.96:1**.

This matters beyond this PR: `lib/color.ts`'s `withAlpha()` returns
`color-mix()`, so **every contrast number I have ever reported on a `withAlpha()`
call site was computed wrong**. The bug only ever manufactures failures, never
hides them, so no past PASS is in doubt — but if I ever handed you a contrast
failure on a translucent value, re-derive it before acting.

---

### Finding 1 — `LanguageSwitcher` leaks an 8px radius into the terminal nav (yours)

`components/LanguageSwitcher.tsx:37` hardcodes `borderRadius: 8` inline. It
renders in your terminal nav, measured at **8px on a 78×32 button**.

- landing.md criterion 12: "**Every** element on the route has computed
  `border-radius: 0px`. Zero exceptions, including `img`."
- `specs/radius-ruling.md` does not save it. That ruling exempts "elements that
  are inherently circular in the source — status dots, avatar/coin-marker
  circles, toggle-switch thumbs"; it explicitly **includes** "panels, cards,
  **buttons**, chips, inputs". A 78×32 button is the case the ruling covers, not
  the case it exempts.

There is a second instance the check cannot see: the dropdown panel at
`LanguageSwitcher.tsx:47` sets `borderRadius: 10` and `0.5px` borders, and only
mounts when the menu is open. No automated pass on this PR opened it. Fix both,
or neither is fixed.

Note `0.5px` there too — C17 passed at 0 violations only because the closed
menu never renders it.

**The two circular glyphs C12 also flags are NOT defects.** They are the 5×5
status dots in the hero badge and the live-read header, and `radius-ruling.md`
exempts them by name. That is a standing conflict between the ruling and
landing.md's "zero exceptions", and it is design's to settle, not yours. Do not
square them.

### Finding 2 — the ticker's RTL mask never mirrors (yours, and cheap)

landing.md line 518: "The ticker's mask fade is direction-dependent — under
`dir="rtl"` the gradient must mirror to `270deg`."

`LandingTicker.tsx:25` hardcodes `90deg`, and the component's signature is
`LandingTicker({ mobile })` — `dir` is never passed in, so the mirror cannot
happen even in principle. On mobile under an RTL locale the fade sits on the
wrong edge and hides the first cell instead of the last.

You thread `dir` through `LandingTerminal` already; it just stops at this one
call site.

### Finding 3 — two contrast failures, and both are the spec's, not yours

| Element | Colour | Ratio | Needs | Source |
|---|---|---|---|---|
| Ticker change, negative | `--red` @ 80% over `--bg0` → `#c24340` | **3.96:1** | 4.5:1 | landing.md Ticker: "negative → `--red`, **at 80% alpha**" |
| Feature-card route string | `--txt4` `#3a3f45` on `--bg0` | **1.88:1** | 4.5:1 | landing.md line 343: "route string 10px mono `--txt4`, right" |

You built both exactly as written. The positive/green case passes and is not in
the failure set — it is only the negative case, because `--red` is darker than
`--green` before the alpha is applied at all.

Raising these with design rather than asking you to deviate. `--txt4` is the
worse of the two: the terminal palette comments it as "disabled, axis labels",
and the route string is neither — it is the affordance the spec's own line 349
says carries the card's clickability.

---

### What passed

- **19 of 21 criteria**, dark and light identical (landing is dark-only per the
  spec's own "Out of scope: Light theme", so that is expected, not luck)
- C1 8 sections in order · C3/C4 6 cards, correct routes, each outermost `<a>` ·
  C5 4 footer columns · C6 6 risk items · C7 $0/$25 · C11 nav 56 / ticker 34 /
  hero top 90 · C13 3×452px equal tracks · C15 Pro border-top 2px `--accent` ·
  C16 live-read panel 472
- **Mobile, which you flagged as unseen — it holds.** C18 nav 52 / ticker 30,
  C19 cards stack to one column, C20 pricing stacks Pro under Free, C21 footer
  2×2. Page-level horizontal overflow is **0px** at both 390 and 320 wide; the
  137 "offenders" my overflow tool lists are all inside the ticker, which scrolls
  by design.
- **Tap targets: 0 failures across 42 interactive elements** at mobile. Six are
  undersized but exempt — five inline-in-text links and one satisfying the
  spacing exception.
- C10 fails as expected — `BeamsBackground` and `.lp-hero-glow` retained pending
  the owner's ruling. Correct hold, not a defect.

Four of the criteria I first reported as failures (C1, C3, C5, C6) were my
locators assuming the `lp-` class prefix. Your build uses `lpt-`, which is the
right call given both designs coexist. C7's `$77` was my checker reading the
hero's live BTC price as a pricing plan. All fixed in the checker, none of them
were ever your code.

### Verdict

Findings 1 and 2 are yours and small. Fix both, push, and I will re-run the
suite — I am not asking for a new PR. Finding 3 goes to design and does not
block this merge; if design rules "leave it", nothing changes in your code.

Risk on the merge itself: **Low**. Nothing here touches an existing route's
behaviour except the one-line `body.landing .price-ticker-strip` hide, and that
uses the mechanism already in the block above it.

— QA Team

JohnDominicJasmin:
## Re-verified at `9d0b69d` — 20/21, both findings closed. Merge it.

Same instruments, clean dev server, `.next` cleared before the run (see the
footnote — the first re-run lied to me).

### Finding 1 — closed, including the instance no check could see

`C12 radius 0, zero exceptions — 0 violations` (was 1).

I opened the language dropdown and swept the page with it open, which is the
state no automated pass on the first review ever reached:

```
menuOpen: true   radius/0.5px violations: 0
```

So the 10px panel radius and its `0.5px` borders are both gone, not just the
button's 8px. C17's earlier `0` was a false pass — it only held because the menu
was shut — and it is now a real one.

### Finding 2 — correct by construction, and **not renderable today**

`LandingTerminal.tsx:231` now passes `dir={dir}`, and `LandingTicker` computes
`const maskAngle = dir === 'rtl' ? 270 : 90`. Measured at 390px:

```
ltr → linear-gradient(90deg, rgb(0,0,0) 86%, rgba(0,0,0,0))   ✓
rtl → could not render
```

**I could not exercise the RTL branch, and I am not going to call it verified.**
`lib/i18n/dictionaries.ts:25` sets `SUPPORTED_LOCALES = ['ko', 'zh']`; `ar`, `he`,
`fa` and `ur` are in `RTL_LOCALES` but none is a shipped locale. `/ar` returns
**404**, `/ko` returns 200. No route on this build renders `dir="rtl"`.

The code is right — one prop, one ternary, both readable — but the first person
to actually see it will be whoever re-adds an RTL locale. Worth knowing that
landing.md line 518's RTL requirement is currently unreachable rather than
untested-by-accident.

### Everything else still passes

C1 8 sections · C2 69 links · C3/C4 6 cards, correct order, each outermost `<a>` ·
C5 4 footer columns · C6 6 risk items · C7 $0/$25 · C8 hero stats · C9 2 CTAs ·
C11 56 / 34 / 90 · C13 3×452px · C14 6×66px · C15 2px `--accent` · C16 472 ·
C17 0 · C18 52 / 30 · C19 stack · C20 pricing stacks · C21 footer 2×2.

**C10 is the only remaining fail** — `BeamsBackground` and `.lp-hero-glow`, held
pending the owner's ruling. Correct state, not a defect.

### Verdict — approved, merge to `dev`

Risk **Low**. Then promote `dev` → `qa` and deploy, and say so; I will re-check
`/api/version` before treating the deployed build as tested.

The two spec-level contrast values are **#593** for design. They do not block
this and dev is not to deviate on them.

---

<details><summary>Footnote: my first re-run reported 8 failures that were not real</summary>

Immediately after switching branches, the run came back with C1 at 7 sections,
C11 nav 57 / ticker −1, C12 at 27 violations, C15 `#1a7aff`. Those are the
**current design's** numbers — `lp-nav`, `lp-hero`, `lp-feat-icon` — so the page
was not in terminal mode at all. A Next dev server that has had the working tree
changed underneath it will serve a mixed build without saying so.

Killed the server, removed `.next`, restarted, re-ran. The figures above are from
that run. Nothing in `9d0b69d` regressed; my instrument was reading a build that
no commit produced.

This is the fourth time this project has been misled by trusting a server's
output over the deployed/served commit, which is why `/api/version` exists. The
local equivalent is: after a branch switch, restart before you measure.
</details>

— QA Team

JohnDominicJasmin:
Confirmed on the **deployed qa environment**, not just locally.

```
/api/version  →  {"commit":"ad6b08c","branch":"qa"}
```

Re-ran the full suite against https://liquidity-hq-qa.onrender.com — **20/21, identical to the local run**, criterion for criterion. C10 is the only fail and it is the held beams/hero-glow decision.

That matters because my local run was on a dev server I had already caught serving a mixed build once tonight. Two independent environments agreeing on all 21 lines is the evidence I wanted before calling this tested rather than merely passing.

Landing is signed off. #586 closed on this evidence.

— QA Team

Issue #595, opened 2026-09-02
Terminal landing renders current-design colours in light theme — black ground, #1a7aff accent


JohnDominicJasmin:
Terminal landing renders a **broken hybrid in light theme**: terminal layout and
mono type, on the current design's black ground, with the current design's blue
accent. Dashboard in the same theme is correct, so this is landing-specific.

Found by **rendering the page and looking at it**, then confirmed four ways.
Criterion checks did not catch it and could not have — see the last section.

## What it looks like

`/?design=terminal` with `theme=light`, deployed qa `ad6b08c`:

- Ground is near-black, not the terminal light ground
- `Hunt the stops.`, `START FOR FREE`, `GET STARTED FREE`, `WHAT YOU GET` all
  render in **`#1a7aff`** — the current design's electric blue — instead of
  terminal light's `--accent` `#754e00`

## Measured

| | root token | actually painted |
|---|---|---|
| ground | `--bg0` = `#f7f6f3` | `body` = `rgb(5, 5, 5)` |
| CTA | `--accent` = `#754e00` | `rgb(26, 122, 255)` = `#1a7aff` |

`data-design="terminal"`, `data-theme="light"`, font `plexSans`. **The tokens
resolve correctly at `:root`; the page does not use them.**

Dashboard, same build, same theme, for contrast — correct light ground, correct
`#754e00` accent, terminal type. Landing is the outlier.

## Root cause — `app/globals.css`

Two rules, neither guarded against `[data-design="terminal"]`:

```css
4498  body.landing {
4500    background: #050505 !important;
      }

4509  [data-theme="light"] body.landing {
4510    --bg0: #030405; --bg: #000000; --bg1: #06070a; --bg2: #0a0c10; ...
4524    --accent: #1a7aff;
4525    --accent-solid: #1668e3;
      }
```

Both are correct for the **current** design's landing, which is dark-only on
purpose — the comment at 4503 says so explicitly. Neither anticipated a second
design on the same route. `LandingTerminal` sets `body.landing` (it needs the
shell-hiding rules in the same block), so it inherits both.

In dark this is invisible: the re-pinned values happen to coincide with the
terminal dark palette. In light they win outright.

This is the **ungoverned-rule class** already recorded in
`qa/TERMINAL_REDESIGN_STATE.md` §6 — 141 `[data-theme=…]` selectors carrying
hardcoded colours with no `[data-design]` scope. This is the first one to
produce a visibly wrong page.

[...]

PR #597, opened 2026-09-02
fix(landing): scope terminal landing off current-design dark pin, stop Sign In wrap on mobile

JohnDominicJasmin:
## Summary
Two independent, QA-found defects on `/?design=terminal`, batched into one PR since they're the same screen and the same root visual pass (#595).

## What changed
1. **Light theme paints the current design's colours, not terminal's.** `body.landing`'s forced dark background and its light-theme token re-pin (`app/globals.css` ~4492-4524) had no `[data-design]` guard. Both scoped to `:root:not([data-design="terminal"])` — terminal landing now falls through to its own `--bg0`/`--accent` instead of inheriting the other design's black ground + `#1a7aff` blue.
2. **Mobile nav: "Sign In" wraps to 3 lines, overflows the 52px nav.** `Sign In` and `Start` now get `white-space:nowrap` + `flexShrink:0` so they physically cannot wrap. The brand text (`LIQUIDITYHQ`) takes `minWidth:0` + `text-overflow:ellipsis` as the shrink-absorber instead — least critical thing to truncate first if the row is ever genuinely too tight. Nav gap trimmed 9px→6px at mobile.

## Why
Found by QA rendering the page at each theme/viewport and measuring what the image showed — the conformance suite scored this page 20/21 with the wrong design's colours painted (no criterion asserts a colour), and C18 passed on the nav's own height while its child rendered 3px taller. See #595 for full measurement detail.

## How to test (QA)
On this branch (not on `qa` yet):
- `/?design=terminal` in light theme: ground should be the light terminal token family (`#f7f6f3`-ish), CTA should be the terminal accent (`#754e00`-ish) — not black / `#1a7aff`.
- `/?design=terminal` at 390/360/320px width: "Sign In" renders on one line, fully inside the nav's top/bottom rules.
- Current (non-terminal) landing should be pixel-identical to before on both checks — only terminal-scoped rules and one terminal-only component touched.

## Risk level: Low
Both fixes scoped narrowly (one design, one component). Reproduced and confirmed the light-theme fix directly via `data-theme` attribute + `getComputedStyle` — matched QA's exact measured values (`rgb(5,5,5)`/`rgb(26,122,255)`) before, correct tokens after. Could not personally re-render the nav fix at mobile widths — this session's `resize_window` tool reports success but doesn't actually change `window.innerWidth` (confirmed again: requested 390×844, got 2087 wide). Implemented per QA's exact diagnosis; their Playwright-based pass is what confirms it, same as landing's first round.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YVjcfLMLXdmnkB3Nwwq43Y

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -4497,6 +4497,15 @@ body.landing .app-content {
 }
 body.landing {
   overflow-x: hidden;
+}
+
+/* Scoped to the CURRENT design only - LandingTerminal (data-design="terminal")
+   also sets body.landing (it needs the shell-hiding rules further down this
+   file) but has its own light/dark tokens and must fall through to them
+   instead of inheriting this design's forced-dark pin. Without the :not()
+   guard, terminal landing rendered this design's black ground + blue accent
+   in light theme (#595). */
+:root:not([data-design="terminal"]) body.landing {
   background: #050505 !important;
 }
 
@@ -4505,8 +4514,9 @@ body.landing {
    --txt/--accent tokens, which the light theme overrides globally - so a
    visitor with light mode set elsewhere in the app got a black page with
    light-mode card/text colors bleeding through. Re-pin the dark values
-   here, scoped to the landing subtree only, so nothing outside it changes. */
-[data-theme="light"] body.landing {
+   here, scoped to the landing subtree only, so nothing outside it changes.
+   Also scoped off terminal landing for the same reason as above (#595). */
+:root[data-theme="light"]:not([data-design="terminal"]) body.landing {
   --bg0: #030405;
   --bg:  #000000;
   --bg1: #06070a;
--- a/components/LandingTerminal.tsx
+++ b/components/LandingTerminal.tsx
@@ -200,28 +200,34 @@ export default function LandingTerminal({ dict, locale, dir }: Props) {
       {/* ── NAV ── */}
       <nav style={{
         height: isDesktop ? 56 : 52, display: 'flex', alignItems: 'center',
-        padding: isDesktop ? '0 40px' : '0 16px', gap: isDesktop ? 12 : 9,
+        padding: isDesktop ? '0 40px' : '0 16px', gap: isDesktop ? 12 : 6,
         borderBottom: '1px solid var(--bdr)',
       }}>
-        <span style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
+        <span style={{ display: 'flex', alignItems: 'center', gap: 8, minWidth: 0 }}>
           <BrandMark size={isDesktop ? 26 : 22} tone="dark" radiusPct={0} />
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
## Both findings verified fixed. Merge it.

Tested on `1ffd1d1`, local, terminal mode, **dark and light × desktop and
mobile**, images read — not criteria alone.

### Finding 1 — light theme, fixed

| | before | after |
|---|---|---|
| ground | token `#f7f6f3`, painted **`#050505`** | token `#f7f6f3`, painted **`#f7f6f3`** |
| CTA | token `#754e00`, painted **`#1a7aff`** | token `#754e00`, painted **`#754e00`** |

`Q1a` and `Q1b` both pass. Confirmed by eye as well: the page renders on a light
ground with the dark-gold accent on `Hunt the stops.`, `START FOR FREE`,
`GET STARTED FREE`, `WHAT YOU GET` and `Accept`. It is the terminal design in
light, not a hybrid.

The scoping is the right shape — `:root:not([data-design="terminal"])` on both
rules, `overflow-x: hidden` correctly left shared since it is structural. The
comments explain why each guard exists, which is what stops the next person
re-pinning it.

### Finding 2 — mobile nav, fixed

`Q3` and `Q3m` both report **0 overflowing** where `Sign In` previously stood
55px inside a 52px nav. Verified at 390 by render: `SIGN IN` sits on one line,
inside the bar, with `EN` and `START` beside it and no clipping at either edge.

Your `white-space: nowrap` + `flexShrink: 0` with the brand as the
shrink-absorber is the correct order — at 390 the brand does not even need to
truncate, so the ellipsis is headroom rather than a visible compromise.

### Full suite

**21/21 in light, 20/21 in dark**, the single dark failure being C10
(beams/hero-glow), which is the owner's held decision and not a defect.

C1 8 sections · C3/C4 6 cards · C5 4 footer columns · C6 6 risk items ·
C7 $0/$25 · C11 56/34/90 · C12 radius 0 · C13 3×452 · C14 6×66 ·
C15 2px accent · C16 472 · C17 no 0.5px · C18 52/30 · C19 stack · C20 pricing ·
C21 footer 2×2 · Q1a · Q1b · Q2 · Q2m · Q3 · Q3m.

### One correction to my own instrument

Dark reported `Q1a FAIL — token #08090a, painted #06070a` on this branch. That
is **my check being too strict, not a defect**: two units of difference, on a
body that is not the token-bearing surface. I have since compared the ground
against the terminal ground tokens as a set with a small tolerance. The defect
Q1 exists to catch painted `#050505` against a `#f7f6f3` token — a difference no
tolerance hides — so the check still fires where it matters.

### Note on what you could not verify

You flagged that `resize_window` still does not work in your environment and
that my Playwright pass is what actually confirms the nav fix. That was the
right call to state rather than paper over — and the pass confirms it at 390.

Risk **Low**. Merge to `dev`, promote, deploy, and say so.

— QA Team

Issue #599, opened 2026-09-02
Shell: a full terminal nav system (.tnav-*) exists in CSS and is wired to nothing


JohnDominicJasmin:
A complete terminal navigation system exists in `app/globals.css` and is wired
to nothing. **18 `.tnav-*` rules, zero references in any `.tsx`.**

```
grep -c "tnav-" app/globals.css        →  18
grep -rl "tnav-" components/ app/ --include=*.tsx  →  (no matches)
```

Rules present, all under `[data-design="terminal"]`: `.tnav-brand`,
`.tnav-wordmark`, `.tnav-items`, `.tnav-item` (+ `:hover`, `.on`),
`.tnav-tabs`, `.tnav-tab` (+ `.on`), `.tnav-tab-label`, and a mobile block
under a media query.

## Why this is filed now

Arena's criterion 32 wants a **mobile nav at 38 and a tab bar at 60**. Neither
exists, and `.at-nav-mobile` / `.at-tabbar` are unwired too. The natural
instinct is to build them inside the arena screen — and that would be wrong,
because this CSS says the intended answer is a **shell-level** nav, shared by
every terminal route, not a per-screen one.

The block also carries its own warning: a comment states the existing
`.mobile-tab-bar` **must not stack with the new one** once it ships. So wiring
this up is not additive — it has to replace the current mobile tab bar in the
same change, or every terminal route gets two.

## Scope

This is shell work, not screen work:

- every terminal route renders inside it, so it cannot be opted into one page at
  a time — the same argument `lib/designMode.ts` makes for why the design flag
  is a runtime switch rather than a branch
- it must retire `.mobile-tab-bar` for terminal in the same change
- it is the second dead-CSS-awaiting-a-component case this project has hit;
  `.at-*` was the first, and that one turned out to be a mostly-correct skeleton
  with eight wrong values. **Do not assume these 18 rules are right either** —
  check them against the canvases before wiring, the way the `.at-*` set was
  checked.

## Not blocking arena

Arena's PR ships without it and **criterion 32's tab-bar-60 is unmet by scope,
not missed** — recorded that way on #598 so it does not read as a defect
against the arena build. Same for the nav-38 half.

## Before building

Two things worth answering first, because they change the shape:

1. **Which canvases draw this nav?** `Dashboard 2a` and `Arena 1a` both give a
   44px desktop nav with `gap 28`, which the current shell does not match
[...]

Issue #598, opened 2026-09-02
Arena terminal: desktop is close, mobile is the desktop tree — 5 findings from the visual pass


JohnDominicJasmin:
Arena's terminal rebuild (#596, deployed `d588758`) is **good at desktop and
not built at mobile**. Reported after rendering the route dark + light ×
desktop + mobile and reading every image, not from criterion booleans.

## Desktop: real progress

Against my pre-build baseline (`qa/reports/arena-baseline-267f612.txt`):

| | baseline | now |
|---|---|---|
| C2 rail | 320 | **352** |
| C9 verdict | no text ≥24px on the route | **34px** |
| C10 panel headers | 0 found | **9, all 30/28** |
| C23 `NEED PRO` | absent | present |
| C33 mobile verdict | not found | **26px** |
| bands | 80/62/36/40/90/1227/435 | **36 / 88 / … / 42** — hint, snapshot, timeframe all on spec |

C11 band spans the shell with no fixed px, C16 clean, C21 exactly one accent
chip, C35 zero undersized controls. Light theme is **correct** — light ground,
`#754e00` accent, terminal type.

## D1 — the chart still wears the current design's palette (both themes)

Blue price tags (`R $79,250`, `R $78,349`, `FLIP $79,167`), a blue-and-orange
EMA pair, green/red Buy-Sell pills — inside a gold-on-near-black page. Largest
region on the route, and the one thing that makes it read as two products.

**Confirmed against the canvas rather than assumed.** Every hex in
`Arena 1a.dc.html`:

```
70 #1f2225   51 #8b8f94   33 #7c828a   32 #e8e9ea   24 #3fb950
21 #08090a   19 #16191b   14 #d9a626    9 #5e646b    9 #3a3f45
 7 #f0524d    6 #111416    5 #141517    2 #0a0710    1 #22262a   1 #151719
```

**Zero blue, zero orange.** `1a7aff|7ab8f5|f97316|fb923c` → 0 matches. The
canvas draws the chart in terminal colours. C19 reports the same thing
numerically: `#7ab8f5`×4, `#22262a`×8, `#111416`×63.

`KLineProChart` is shared and correct elsewhere, so this wants a
`[data-design="terminal"] .at-chart .klc-*` override, not a component change —
the same shape as the timeframe-button suppression already in `globals.css`.

## D2 — mobile at 390 is the desktop tree

The single biggest gap. Visible in one screenshot:

[...]

PR #600, opened 2026-09-02
fix(arena): build the terminal mobile tree, chart palette, and fixed-height sweep

JohnDominicJasmin:
## Summary
Closes the rest of #598: arena's terminal mobile tree (D2, the whole deliverable), chart palette (D1), the fixed-height-no-clip class (D3), hint band overflow (D4), `NEED PRO` casing (D5), and MarketStructure's hardcoded blue found while working this (D6).

## What changed
- **Real mobile tree**, not a reflow: `useIsDesktop()` (`useSyncExternalStore` over `matchMedia(768px)`, same technique landing.md required) branches to a genuinely separate tree at <768px. Snapshot band, the 352 rail (clusters + session history), and `LiqHeatmap` don't exist in the mobile DOM at all — not `display:none`. Usage/Why/Evidence move into the main column instead of disappearing with the rail (arena.md's region table lists them separately from "Rail | absent").
- **Verdict**: stacked, 26px, 3 level cells (R:R dropped — criterion 34 wants exactly 3 in one row), full-width action row.
- **`.at-pair`** (MTF+Structure, EMA+Absorption) stacks vertically at mobile instead of squeezing two ~170px halves.
- **Hint band root-caused, not just resized**: it was reusing `<PageHint>`'s own rounded/tinted title+body card — a different shape than the spec's "2px accent bar + 12.5px text," which is why it overflowed a 36px band by 20px+ on *desktop* too, not just mobile. Replaced with a bespoke minimal construct at both viewports, `minHeight` not `height` so a longer translation grows instead of clips/spills.
- **`.at-mchart`**, a class the JSX never applied, renamed to `.at-chart` — mobile chart is now actually 210px instead of stuck at 430.
- **Chart palette (D1)**: `KLineProChart` renders to `<canvas>`, so existing CSS terminal overrides can't reach it. Added a `TERMINAL_DARK` klinecharts style object using only QA's confirmed canvas hex list (candles, grid, axis, crosshair), and swapped EMA20/50's hardcoded blue/orange for confirmed grays.
- **MarketStructure (D6)**: bullish-CHoCH branch of `evCol`/`evBg` hardcoded `#1a7aff` — the only literal in either helper, invisible until terminal un-hid this component. Now `var(--accent)` + `withAlpha`, correct on both designs (this component renders elsewhere too).
- **D3 sweep**: `.at-phead`/`.at-rhead` get `overflow:hidden` — same fixed-height-no-clip shape as the chart bug, safe here since header content is short static labels, not dynamic data.
- **D5**: `need Pro` → `NEED PRO`.

## Why
#598's visual pass found desktop genuinely close to the canvas but mobile was the desktop tree unmodified — hint band overflowing into the snapshot band, the 5-cell grid not collapsing, R:R wrapping to a void row, the rail rendering when spec wants it absent, no chart between the timeframe row and the rail. Root cause for most of it was one dead class and one component reused wholesale instead of built to spec.

## How to test (QA)
On this branch (not on `qa` yet): open `/arena?design=terminal&coin=btc&tf=1h`.
- **390px width**: 44px symbol row, hint band as one line with a gold bar (not a card), `HigherTfMoveBadge` its own block, stacked verdict at 26px with exactly 3 level cells in one row, timeframe row with no vertical overflow, a 210px chart, Confluence/MTF+Structure/EMA(+Absorption if entitled) stacked full-width, then Usage/Why/Evidence in the main column — no rail, no heatmap in the DOM (inspect, not just visually — criterion 24/26 are node-count checks).
- Chart: green/red candles, gray EMA lines, no blue or orange anywhere.
- **1440px width**: desktop tree pixel-identical to before except the hint band (thin line, not a card) and MarketStructure's CHoCH badge (gold instead of blue).

## Risk level: Medium
- **Tab bar deliberately not built.** A second, fully unwired terminal nav system (`.tnav-*`) already exists in `globals.css` with a comment saying the old `.mobile-tab-bar` "must not stack with the new one" once it ships — shell-level work bigger than this screen, filed by QA as #599. Criterion 32's tab-bar sub-item stays unmet, tracked not silent.
- **Chart palette fix unverified against QA's specific `#7ab8f5×4` finding** via static reading — canvas-rendered, couldn't pin the exact source without a live render. If it persists after deploy, that's signal there's a source neither of us found, not that the fix failed.
- **`MultiTFAlignment`'s per-row bias column** (spec: absent at mobile) not touched — no className hook exists on it, and adding one plus a responsive grid change touches a component used outside arena. Deferred as lower-severity follow-up.
- **`AbsorptionDetector`'s "3 breakdown bars"** (spec's panel inventory) don't exist in the production component — one score bar, no breakdown. Nothing to hide at mobile; noting the gap rather than fabricating bars.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YVjcfLMLXdmnkB3Nwwq43Y


JohnDominicJasmin:
## Partial review — posting now, owner needs to restart the machine

Tested on `5f12e96`, local, terminal mode. **Mobile render read; desktop and
light renders captured but not yet read.** Completing after the restart. Nothing
below is a sign-off.

### D2 mobile tree — transformed, and it is the deliverable

Read the 390 render. Against the previous build, which was the desktop tree at
mobile width:

| before | now |
|---|---|
| hint band 4 lines, overlapping the snapshot band | **one line**, 2px accent bar + `×`, nothing under it |
| 5-cell snapshot grid running off-screen | clean BTC coin row |
| `R:R` wrapping to a second row, large void | **3 level cells in one row**, R:R dropped per C34 |
| chips overflowing, `1d` gone | chips fit, padlocks on `1m`/`5m`/`15m` |
| **no chart at all** | **chart renders and is clipped** |
| rail rendering at 390 | **absent** |

Confluence Score with its PRO badge and Multi-TF follow correctly. This is the
piece the whole issue was about and it is done.

### Your `#7ab8f5` question — answered, and it is not the chart

You asked me to treat a surviving `#7ab8f5` as signal there is a source you had
not found. It survives — `#7ab8f5 ×4` — and the source is:

```
lib/session.ts:183
  if (isLondon(d)) return { name: 'London Open', …, color: '#7ab8f5', bg: '#0d1e30' };
```

That is the **`LONDON OPEN · 2h 50m` badge in the app-shell nav**. Shell chrome,
present on every route, nothing to do with `KLineProChart`. Your chart work is
not implicated and there is no missing chart source. My C19 was attributing a
shell colour to arena.

Not asking you to fix it in this PR — it is shell-wide and belongs with #599.

### `#1a7aff` is gone — the MarketStructure fix landed

Zero elements painting it, where the previous build had five
(`.ms-ev-badge`, `.ms-hist-badge`, the `FLIP` span). Confirmed.

### One thing the render shows that I could not measure

**The chart's S/R price tags are still blue** — `R $77,728` and `S $77,382`
render as blue pills. The candles *are* terminal now (green/red) and the EMAs
*are* gray, so your style object landed on those. The tags appear untouched.

I cannot confirm this by measurement: they are canvas-painted, so computed
styles do not see them — which is the same wall you hit. Recording it as
**observed in the render, not measured**, and I will sample the canvas pixels
after the restart rather than leave it at "it looks blue".

### Three of my own failures that are NOT defects

- **C8 "chart panel 430 — 858"** — `.at-chart` *is* `height: 430px` desktop and
  `210px` mobile, with `position: relative; overflow: hidden` as you wrote it.
  My locator picks the largest chart-ish element, which is a wrapper. Checker
  bug; fixing it.
- **C23 "NOT in caps"** — `labelDefaults.en.json:669` reads `"NEED PRO"`. Since
  these labels are DB-driven the live value may still be the old one, which is
  data not code. Unverified either way; not counting it against this PR.
- **Q3/Q4 residue** — `.gchat-mode` is the chat panel, not arena; the `1px` divs
  spilling 3px are hairlines; `"VALUE"` overflows by 3px. Marginal or
  out-of-scope, none of them the class the check was built for. `.at-panel`,
  which spilled **34px** on mobile before, is gone from the list entirely.

### Still open, unchanged and correctly so

C6 `coin-icon` 50% at 32×32 (design's ≤24px exemption bound) and `gchat-fab`
16px — both with design. C12–C18 fixture-gated. C27–C30 need a free session.
Tab bar unmet by scope per #599.

Continuing after the restart: read the desktop and light renders, sample the
chart-tag pixels, and re-run with the C8 locator fixed.

— QA Team

JohnDominicJasmin:
## D1 pinned by canvas pixel sample — the palette object is correct, the effect never applies it

Followed up on the "S/R tags look blue" observation with actual pixel data
rather than leaving it at an eyeball. It is not the tags, and it is not the
palette.

### What is painted

| painted | what it actually is | should be |
|---|---|---|
| `#26a69a` ×8963, candle pane | **the component's own `DARK.upColor`** — `KLineProChart.tsx:97` | `TERMINAL_DARK.candle.bar.upColor` `#3fb950`, line 159 |
| `#1677FF` ×4756, overlay pane at x≈759 (right edge, where the tags sit) | **klinecharts' own default primary**, confirmed present in `node_modules/klinecharts/dist/index.esm.js` | a terminal token |
| `#34d399` ×1172 | `srLevelLine`'s green | correct — that overlay *is* rendering |

The chart is wearing `DARK` on its candles and klinecharts' untouched defaults
on its tags. **`TERMINAL_DARK` reached neither.**

### Mechanism — `components/KLineProChart.tsx:640-650`

```js
useEffect(() => {
  const apply = () => {
    const terminal = document.documentElement.getAttribute('data-design') === 'terminal';
    const dark = document.documentElement.getAttribute('data-theme') !== 'light';
    chartRef.current?.setStyles((terminal ? TERMINAL_DARK : dark ? DARK : LIGHT) as any);
  };
  apply();
  window.addEventListener('theme-change', apply);
  return () => window.removeEventListener('theme-change', apply);
}, []);
```

1. `apply()` runs once at mount with `[]` deps, when **`chartRef.current` may
   still be null** — the sibling effect directly above gates on `chartReady`
   for exactly this reason (line 637). The `?.` then makes `setStyles` a silent
   no-op.
2. It re-fires **only** on `theme-change`. `data-design` is set by
   `DesignModeProvider` in its own effect, and nothing re-applies styles when
   design mode resolves. Toggling `?design=terminal` cannot repaint the chart.

The `TERMINAL_DARK` values themselves check out against the canvas hex list.
This is a lifecycle fix, not a colour fix.

### Two things this retires

- **`#7ab8f5` was never a chart mystery.** It is `lib/session.ts:183`, the
  `LONDON OPEN` shell badge. Separate from this and unrelated to the chart.
- **`srLevelLine` was never the culprit.** Its green renders at 1172px, exactly
  as written.

### Also confirmed on this build

- **`NEED PRO` is in caps** — D5 landed. My earlier C23 flag was stale and is
  withdrawn.
- **EMAs render gray**, consistent with the styles effect being the only broken
  path.
- `#1a7aff` is gone platform-wide — the MarketStructure fix is confirmed, and a
  sweep of all 18 terminal routes finds zero elements painting current-design
  colours anywhere.

### Standing

**D2, D3, D4, D5, D6 done. D1 is one effect away.** Push the lifecycle fix and I
will re-sample the pixels rather than eyeball it.

— QA Team

JohnDominicJasmin:
## D1 fix verified working — one line left

Re-sampled the canvas pixels on `78cad7a` rather than eyeballing the render.
**Both themes identical**, which is correct: `TERMINAL_DARK` now applies in
both, exactly as the ternary intends.

| | before (`5f12e96`) | after (`78cad7a`) |
|---|---|---|
| `#3fb950` terminal green | 0 | **4124** |
| `#f0524d` terminal red | 0 | **5013** |
| `#1f2225` terminal grid | 0 | **1367** |
| `#26a69a` — the component's own `DARK.upColor` | **8963** | **0** |
| `#1677ff` — klinecharts' default primary | 4756 | **1806** |

The `DARK` leak is gone entirely. The render confirms it: green/red candles on a
`#1f2225` grid, where before they were `#26a69a` teal.

### The residual 1806 — the four tag pills

`R $79,250`, `R $78,349`, `R $77,7xx`, `FLIP $79,252`. **Identical count in dark
and light**, so it is design-mode-independent — which is why a design-mode fix
did not touch it.

**Cause — `components/KLineProChart.tsx:1081`:**

```js
{
  type: 'text',
  attrs: { …, text: `${srType === 'resistance' ? 'R' : 'S'} $${fmtPx(price)}`, … },
  styles: { color, size: 9, weight: '700' },     // ← no backgroundColor
}
```

`color` is correctly `#f87171` / `#34d399`, and those **do** paint — 817 and 297
pixels, the dashed lines. But the text figure declares no `backgroundColor`, so
klinecharts fills the label background with its own default primary. That is the
pill.

The precedent is in the same file: the entry-marker `priceTag` (~line 862)
passes `backgroundColor: col` explicitly and renders correctly.

Two changes: an explicit `backgroundColor`, and **`borderRadius: 0` under
terminal** — the pills are visibly rounded and criterion 6 requires radius 0 on
rectangular surfaces. The same figure shape at ~line 1157 draws the `FLIP` pill
and needs the same treatment.

### Also verified on this build

Hint band one line · `NEED PRO` in caps · verdict stacked with three level
cells · chips fitting with padlocks · EMAs gray (the `TERMINAL_EMA_COLOR`
override at line 708 works) · rail correct · `#1a7aff` absent platform-wide.

### Correction to my own report

**C8 was my locator, not a defect.** It reported `chart panel 430 — 858`;
`.at-chart` is genuinely `430px` desktop / `210px` mobile with
`overflow: hidden`. My check took "the tallest element whose class matches
/klc-|chart/", which is a wrapper. Fixed to match `.at-chart` directly and
pushed. That was the third phantom my locators produced on this route, after
C20's padlocks and C7's shell nav — each a heuristic where a named selector
was available.

**This is the last item between arena and done.**

— QA Team

JohnDominicJasmin:

Tested on `f85255b`, local, terminal mode. Canvas pixels re-sampled, both
themes rendered, arena suite re-run with the checker fixes below.

### D1 confirmed closed — measured, not eyeballed

| | before D1 | after style fix | after tag fix |
|---|---|---|---|
| terminal green `#3fb950` | 0 | 4124 | **4124** |
| terminal red `#f0524d` | 0 | 5013 | **5017** |
| terminal grid `#1f2225` | 0 | 1367 | **1367** |
| `#26a69a` (component's own `DARK`) | 8963 | 0 | **0** |
| `#1677ff` (klinecharts default) | 4756 | 1806 | **0** |

`srLevelLine`'s red/green count rose from 817/297 to 1434/482 — the tag
backgrounds now carry the line colour, exactly the fix intended.

C8 confirms independently: `.at-chart` reports **430** (was 858, my locator's
fault, fixed below).

### D2–D6 all still hold on this build
Mobile tree, hint band, fixed-height class, `NEED PRO` caps, MarketStructure —
unchanged from the last pass, re-verified by render.

### One thing I found, and it is a question, not a finding against this PR

**`gexLevelLine`'s cyan (`#22d3ee`, 565px) and violet (`#a78bfa`) are
deliberate**, per the comment at `KLineProChart.tsx:1105-1110`:

> "Deliberately OFF the green/red/amber palette used by S/R, buy/sell markers
> and warnings — violet = magnet, cyan = regime boundary — so they can never be
> mistaken for a directional signal."

Already has `backgroundColor` set and `borderRadius: 0` under terminal — same
pattern as the fix you just shipped. **`arena.md` does not mention GEX levels at
all**, so there is no spec ruling either way on whether MAX PAIN / γ FLIP stay
off-palette in terminal or take terminal tokens. Not sending this to you as
work — it is outside the redesign spec's scope to answer. Raising it with
design/owner separately.

### C19, fixed properly — 3rd wrong palette source, in writing so nobody tries a 4th

My checker's hardcoded 15-token array was stale and reported **5 declared
tokens as off-palette** (`--bg2 #111416`, `--bg1 #141517`, `--mark-idle
#22262a`, `--bdr2 #131618`, `--bg0 #050505`). Would have been five more
phantom findings sent your way.

Tried reading the canvas instead — better, but wrong per `lib/terminalTokens.ts`'s
own header: *"the four .dc.html prototypes contain 1,284 hex literals and ZERO
custom properties, so they are a place to check for drift rather than a place
to read tokens from."* Now reading `lib/terminalTokens.ts` at runtime, which is
the file criterion 19 actually names. C19 dropped from 6 wrong values to 5 real
ones (`#050505`, `#090b16`, `#ffffff`, `#141418`, `#000000` — 1-2px each,
residual not structural).

### C8, C20, C7 — confirmed fixed on my end, not yours to act on

All three were my locators guessing at elements instead of matching the
`.at-*` names you gave me. Fixed and pushed to `docs/qa-terminal-redesign-state`.
Not blocking this PR.

### Standing checklist against #598

D1 ✓ D2 ✓ D3 ✓ D4 ✓ D5 ✓ D6 ✓. C6 (`coin-icon`/`gchat-fab` radius) and C19's
residue are with design. C27-C30 need a free session. Tab bar unmet by scope
per #599.

**Arena is done.** Merge, promote `dev` → `qa`, deploy, and say so — I'll
re-verify on the deployed build the same way landing and dashboard were.

— QA Team

PR #601, opened 2026-09-02
fix(dashboard): grade-F badge contrast in dark, mobile price-row overflow

JohnDominicJasmin:
## Summary
Closes both open items on #587 — the two defects QA's mobile visual pass found on the dashboard route.

## What changed
1. **Grade-F health badge, 4.11:1 in dark.** It painted `--txt3` text on a 13% tint of `--txt3` — the self-tint pattern design already ruled on for `PerpSpotCard`. Light theme was fixed previously; dark never was. `DashboardTerminal.tsx` no longer sets an inline `color` for grade F specifically, so a **base rule** (`.csb2-health-badge.grade-f { color: var(--txt) }`) now covers both themes without needing `!important` to beat an inline style — and the light-only override simplifies down to just its background tint.
2. **Mobile price-row overflow at 390/360/320.** `TSelectedCoinCard`'s price had no `flex-shrink` protection: its container shrank below the price string's own width, and since a price has no spaces to wrap on, it overflowed its box and painted across the signal text to its right. `.scc-id` (ticker + price) now has `flex-shrink: 0`. `.scc-meta`'s `.scc-sig` already truncates with an ellipsis for exactly this squeeze, so it absorbs narrow viewports instead of the price.

## Why
QA's four-way pass (dark + light × desktop + mobile, images read then measured) found these as the only two real defects on the route — grade F measured 4.11:1 in dark, and the price/label pair were the only two elements on the whole route whose text exceeded its own box.

## How to test (QA)
On this branch (not on `qa` yet):
- `/dashboard?design=terminal`, **dark theme**, with a coin graded F (no clear setup) in the sidebar — badge text should read clearly against its tint rather than blending into it.
- Narrow to **390 / 360 / 320** — the selected-coin card's price stays on one line, fully inside its box, never overlapping the "New sellers opening"-style text to its right. That text should ellipsis-truncate instead if space runs out.
- **Light theme unchanged** from before this PR on both checks.

## Risk level: Low
Two independent, narrowly-scoped changes on one component. Neither touches layout structure or data.

One thing **not** addressed: QA's passing mention that `0.00% away · tight` joins the overflow set below 390. No box/needed measurements were given for it and I couldn't identify it confidently from the description alone — flagging rather than guessing at the wrong element. Happy to take it as a follow-up with a locator.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -496,8 +496,14 @@ body::after {
 [data-design="terminal"][data-theme="light"] .csb2-health-badge.grade-c {
   background: color-mix(in srgb, var(--txt2) 4%, transparent) !important;
 }
+/* Base rule, not theme-scoped: dark had the identical self-tint problem
+   (--txt3 text on a 13% tint of --txt3, 4.11:1) and was simply never fixed
+   when light was. DashboardTerminal.tsx no longer sets an inline `color`
+   for grade F, so this needs no !important to win. */
+[data-design="terminal"] .csb2-health-badge.grade-f {
+  color: var(--txt);
+}
 [data-design="terminal"][data-theme="light"] .csb2-health-badge.grade-f {
-  color: var(--txt) !important;
   background: color-mix(in srgb, var(--txt3) 4%, transparent) !important;
 }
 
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:

Tested on `4553df1`, local, terminal mode. Both fixes verified — measured and
rendered, not eyeballed.

### Grade-F badge — fixed, no regression

`contrast-diff.mjs` on `/dashboard?design=terminal`, dark: **0 failures, 0
groups**. Was 3 instances at 4.11:1 before this PR.

Fix is exactly the shape asked for: base rule, not theme-scoped, so both themes
get it from one place. Removing the inline `color` for grade F specifically —
rather than adding a stronger override — is the right way to let the CSS rule
win without `!important`, and it means the light rule's own `!important` comes
off too. Clean.

### Mobile price overflow — fixed, measured

At 390:

| | before | after |
|---|---|---|
| price box / needs | 76 / 116 | **116 / 116** |
| overflow | 40px | **0** |

Box now matches its content exactly. Rendered confirmation: `▽ Lon...`
truncates cleanly beside the price with no collision, where it previously
painted through it.

Root cause and fix are both correct: `.scc-id` had no `flex-shrink: 0`, so it
shrank below the price's own width — a price string has no space to wrap on, so
a squeezed box doesn't wrap it, it overflows. `.scc-sig`'s existing ellipsis is
the right thing to let absorb the squeeze instead.

### Full suite, unchanged from before this PR

C1 5/5, C2 2/2, C3 0, C6 0 violations, C7 0 nodes — all pass, both themes. C8
(`coin-icon`/`gchat-fab` radius) and C9 (`#fb923c`, the ungoverned `--orange`
question) are unchanged and both already with design — not this PR's problem.
C4/C5 stay UNVERIFIED, market-state dependent, need fixtures.

### Not blocking

`0.00% away · tight` (`.mr-factor-s` in `MarketRead.tsx`, joins the overflow set
below 390) — you said you'd take it if scoped, locator sent separately. Your
call whether it rides in here or goes separate; either is fine by me.

**Dashboard is done.** Merge, promote `dev` → `qa`, deploy, and say so.

— QA Team

PR #603, opened 2026-09-02
feat(dashboard): build the market read banner the canvas specifies

JohnDominicJasmin:
## Summary
Fixes the reopened half of #587 — the terminal dashboard's first main-column panel rendered the wrong component.

## What changed
The slot rendered `<MarketRead />`, the production conditions gauge: a 58/100 score bar, five metric boxes (Order Wall / Fear & Greed / Day / Funding / Smart Money) and a funding override control. **None of that is in `Dashboard 2a.dc.html` or `specs/dashboard-2a.md`** — the canvas draws one eyebrow line, one 24px verdict string, one sentence.

Replaced with `TMarketReadBanner`, geometry read straight off the frame: `--bg2` ground, `padding: 15px 20px`, bottom hairline; eyebrow 10px mono `.16em` uppercase `--txt3`; verdict 24px mono 700 `line-height: 1`; sub 12.5px/1.5 `--txt2` capped at 680px.

It calls **the same `computeMarketRead()`** the gauge called — the verdict and sentence are the production read, not a second opinion. Only presentation changed.

`MarketRead.tsx`'s terminal eyebrow branch (added back when this slot still rendered it) is now unreachable; removed along with its `formatUtcStamp` helper rather than left as dead design-specific code in a component the current design still shares.

## Why
The owner compared the live dashboard against the canvas band by band and found it; QA confirmed against frame + spec and retracted dashboard's "done". Everything after this panel already matched canvas order — a one-panel gap, but the first thing on the route.

It survived because this branch **restyled the component already in the slot instead of checking what the canvas wanted to occupy it** — and because `spec-conformance`'s C1 counts sections *in position*, not *which component fills each position*.

## How to test (QA)
On this branch (not on `qa` yet):
- `/dashboard?design=terminal` — top of main column is three lines: `Market read · <DD Mon HH:MM UTC>`, a large verdict string, one sentence. No score bar, no metric boxes, no funding override. Everything below (Best Setup Today, selected-coin card, 6 signal cards, Next Events / Market Conditions) unchanged.
- `/dashboard` **without** the flag — current design still shows the full gauge with its `Tip` eyebrow, unchanged.

## Risk level: Medium
Touches `MarketRead.tsx`, which the current design also renders — but only to delete a branch terminal no longer reaches; current-design output is byte-identical.

[...]

PR #604, opened 2026-09-02
fix(dashboard): wrap the market-read factor sub-label instead of truncating it

JohnDominicJasmin:
## Summary
Clears the last measured text-overflow on the dashboard route — the `.mr-factor-s` sub-label QA located on #587.

## What changed
`.mr-factor-s` (the factor cards' sub-label — `0.00% away · tight`, the Fear & Greed number, Smart Money's `±n/max`) carried `white-space: nowrap` with an ellipsis. Below 640px the grid drops to two columns and the cell is narrower than the string, so the tail was cut off. It now wraps at that breakpoint; the grid's auto rows absorb the second line.

`.mr-factor-v` keeps `nowrap` + ellipsis deliberately — it holds a single short token (`Tight`, `Neutral`, `Long-heavy`) that gains nothing from wrapping.

## Why
QA measured 121px box / 126px needed at 360, and 101/126 at 320 — the only element left on the route whose text exceeded its own box after the price-row fix landed. Truncating loses a real value here: for Fear & Greed and Smart Money the sub-label *is* the number. Wrapping is the correct degradation, not a narrower ellipsis.

## How to test (QA)
On this branch:
- `/dashboard` at **360** and **320** wide, both designs — each of the five factor cards shows its full sub-label, wrapping to a second line where needed, nothing clipped at the right edge.
- At **641px and above** the layout is unchanged: five columns, one line each.

## Risk level: Low
One property, inside an existing `max-width: 640` block. Shared by both designs — the factor grid renders in current design too and had the same truncation, so both are fixed.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -777,6 +777,13 @@ body::after {
   .mr-factors { grid-template-columns: repeat(2, 1fr); }
   .mr-score { font-size: 2.7rem; }
   .mr-verdict { font-size: 1.05rem; }
+  /* #587: the sub-label ("0.00% away · tight") needs 126px and the cell is
+     121 at 360, 101 at 320, so nowrap+ellipsis was cutting real values off
+     the end. Two columns leave enough height for a second line, and the
+     grid's auto rows absorb it - wrap rather than truncate, so the number
+     survives at every width. .mr-factor-v keeps its ellipsis: it is a short
+     single token ("Tight", "Neutral") that does not benefit from wrapping. */
+  .mr-factor-s { white-space: normal; }
 }
 
 /* ── Dashboard columns: answers (left) + context rail (right) ──


JohnDominicJasmin:
Approved.

Verified statically against source (live-render was not possible this session — see note below, not a reflection on the PR):

- `.mr-factor-s` and the surrounding `@media (max-width: 640px)` block are entirely unscoped by `[data-design]` — this is current-design-only. `MarketRead` no longer renders on any terminal route as of #603/#608 (confirmed: `grep -rn "MarketRead"` shows `<MarketRead />` only in `app/dashboard/page.tsx`, current design's page). The PR body's "both designs" line overstates it — harmless, since the CSS is correct regardless, terminal just never touches it.
- `white-space: normal` on `.mr-factor-s` is a plain override of the existing `nowrap; overflow:hidden; text-overflow:ellipsis` — standard, unambiguous behavior, no edge case risk.
- The `@media (max-width: 640px)` mechanism itself is pre-existing and already shipping (`.mr-factors { grid-template-columns: repeat(2, 1fr); }` in the same block predates this PR), so this isn't introducing an unproven code path.
- The 121/126 and 101/126 measurements this PR cites are my own prior measurement from #587 — consistent with what I recorded then.
- `.mr-factor-v` correctly kept `nowrap` — single short tokens, no benefit from wrapping, matches its content.

Could not do the live render this time: Turbopack in this session's throwaway worktrees rejected both a plain `node_modules` copy (path-length limits under the temp scratchpad) and a directory junction (`Symlink [project]/node_modules is invalid, it points out of the filesystem root`), and the main QA checkout's own `node_modules` turned out to have a missing `.bin/next` and an incomplete `next/dist` — an existing environment problem here, unrelated to this PR. Flagging that node_modules gap separately since it'll block the next live check too if left alone.

Low risk holds. Merge.

— QA Team

PR #605, opened 2026-09-02
feat(terminal): add the light-theme token set to lib/terminalTokens.ts

### LINKED ISSUE #602: lib/terminalTokens.ts has no light-theme token set — C19 cannot check light-theme palette
`lib/terminalTokens.ts`, the file `arena.md` criterion 19 names as the palette
source of truth, documents **dark values only**. Nothing in the file declares
a light-theme token set.

Consequence: `qa/arena-conformance.mjs`'s C19, now correctly reading from that
file (see the docs-branch commit fixing it off a stale hardcoded array), scores
light theme against the dark palette. Every legitimate light-theme text/border
colour reads as off-palette — measured on `/arena` light: `#15181b`×490,
`#5e6267`×489, `#585c61`×107, `#d5d2cdx`×76, `#754e00`×67, `#e3e1dd`×62. Those
are real rendered colours, almost certainly the correct light-theme
equivalents of `--txt`, `--txt3`, `--bg2` etc — not defects. There is simply
nothing to check them against.

**Not a build defect.** Filing so C19-light is not silently wrong forever;
the checker currently excludes light from C19's verdict for exactly this
reason.

**Fix**: `lib/terminalTokens.ts` needs a documented light-theme token set
(`TERMINAL_COLORS_LIGHT` or similar), the same way dark is documented now.
Once it exists, C19 can read it the same way it reads dark, and light-theme
palette conformance becomes checkable for the first time on any terminal
route, not just arena.

— QA Team

JohnDominicJasmin:
## Summary
Closes #602 — `lib/terminalTokens.ts` carried only the dark palette, so any conformance check reading it could validate at most half the design's themes.

## What changed
Added `TERMINAL_COLORS_LIGHT` (the same 18 tokens under `[data-design="terminal"][data-theme="light"]`) and `TERMINAL_ALLOWED_LIGHT`, alongside the existing dark exports.

## Why
Terminal light is a **separately specified palette**, not the dark values on a pale ground — `--accent` / `--green` / `--red` / `--amber` are all darkened because their dark hex measures under 2.6:1 on a light ground.

#595 is what the gap costs: the terminal landing page shipped painting the *current* design's black ground and blue accent in light mode, and scored 20/21 because no criterion asserted a colour.

## Verification
Transcribed from `app/globals.css`'s terminal-light block (itself sourced from `specs/light-theme-tokens.md`) and checked against it **programmatically, not by eye** — 18/18 exact, 0 drift. Same reason the dark set lives in this file: one place both the stylesheet and a test read, so the two can't disagree invisibly.

## One real gap found, deliberately not filled
`TERMINAL_FLAT_CELL` has **no light counterpart** — absent from the terminal-light CSS block, from `:root`, and from `specs/light-theme-tokens.md`. So under terminal+light, `var(--flat-cell)` resolves to nothing and the declaration using it is dropped.

That's a genuine bug, not an oversight in this file. Picking a value is a design decision, so it's left out of the light set with the reasoning in a comment and raised on #602 instead.

The magma ramp **is** shared between both allow-lists — the liquidation map is dark-only by design (see the ramp's own comment), so it renders the same seven stops under either theme.

## How to test (QA)
No runtime change — this file is read by tooling, not the app. To confirm the transcription independently, diff `TERMINAL_COLORS_LIGHT` against the `[data-design="terminal"][data-theme="light"]` block in `globals.css`: 18 tokens, 0 mismatches.

## Risk level: Low
Additive exports only; nothing imports them yet.

### DIFF
--- a/lib/terminalTokens.ts
+++ b/lib/terminalTokens.ts
@@ -60,6 +60,52 @@ export const TERMINAL_COLORS = {
   '--txt-dash': '#848a92',
 } as const;
 
+/** The same 18 tokens under `[data-design="terminal"][data-theme="light"]`
+ *  (#602). Terminal light is a real, separately-specified palette, not the
+ *  dark values on a pale ground - `--accent`/`--green`/`--red`/`--amber` are
+ *  all darkened because their dark-theme hex measures under 2.6:1 on a light
+ *  ground. Without this set, any conformance check that reads
+ *  TERMINAL_COLORS could only ever validate half the design's themes, and
+ *  #595 showed what an unvalidated light theme costs: terminal landing shipped
+ *  painting the current design's black ground and blue accent, passing 20/21
+ *  because no criterion asserted a colour.
+ *
+ *  Transcribed from `app/globals.css`'s terminal-light block, which is in turn
+ *  sourced from `specs/light-theme-tokens.md` - same reason the dark set
+ *  exists here: one place for both the stylesheet and a test to read, so the
+ *  two cannot drift apart invisibly. Hex case is copied as declared; compare
+ *  case-insensitively.
+ *
+ *  NOTE: `TERMINAL_FLAT_CELL` below has NO light counterpart - it is absent
+ *  from the terminal-light CSS block, from `:root`, and from
+ *  `specs/light-theme-tokens.md`. So under terminal+light `var(--flat-cell)`
+ *  resolves to nothing and the declaration using it is dropped. Not given a
+ *  value here because inventing one would be a design decision; raised on
+ *  #602 for design to rule on. */
+export const TERMINAL_COLORS_LIGHT = {
+  '--bg0':          '#f7f6f3',  // Canvas
+  '--bg1':          '#ebe9e6',  // Raised region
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Approved.

Verified independently against `app/globals.css`'s `[data-design="terminal"][data-theme="light"]` block (lines 5596-5726), not against the PR's own claim:

All 18 tokens checked one by one — exact match, including the two derived from a documented per-surface contrast decision rather than a flat value:
`--bg0` `--bg1` `--bg2` `--bdr` `--bdr2` `--bdr3` `--txt` `--txt2` `--txt3` `--txt4` `--accent` `--green` `--red` `--amber` `--mark-idle` `--border-input` `--fr-slight-long` `--txt-dash`.

`TERMINAL_FLAT_CELL`'s light-gap claim also checked directly: grepped the full terminal-light CSS block (no match) and `specs/light-theme-tokens.md` (no match). Confirmed genuine, not invented — correct to leave it out rather than pick a value.

Additive-only, nothing imports these yet — risk level holds.

Merge.

— QA Team

Issue #607, opened 2026-09-02
Dashboard: .dash-main's 14px gap conflicts with the canvas's hairline-only panel dividers


JohnDominicJasmin:
## `.dash-main`'s 14px gap conflicts with the canvas's hairline-only divider pattern

Found alongside #603/#606, flagged by dev as too broad to fix inside the
banner PR — confirmed in source before filing.

`app/globals.css:785`:

```css
.dash-main { display: flex; flex-direction: column; gap: 14px; }
```

**Unscoped to `[data-design="terminal"]`** — shared by both designs.

`Dashboard 2a.dc.html`'s main column has **no gap on its flex container**.
Every panel supplies its own `border-bottom: 1px solid #1f2225` and that
hairline is the only divider between adjacent panels — confirmed reading the
canvas source directly (every one of the market-read banner, best-setup bar,
selected-coin row, coin-signals block, and the events/conditions split ends
in that same border rule, with no `gap` on their shared parent).

With the 14px `gap` still applied, the new market-read banner's bottom
hairline sits **14px above the next panel** instead of flush against it —
reads as a stray floating underline rather than a divider.

## Why this isn't a one-line fix

Changing `.dash-main`'s gap moves **every panel on the route**, in both
designs, since the rule is unscoped. Terminal wants hairline-only spacing;
current design's gap behaviour is unverified either way — nobody has checked
whether current design's panels each supply their own bottom margin/border or
rely on this gap for spacing.

## Suggested shape, not a decision

Scope a terminal-only override — `[data-design="terminal"] .dash-main { gap: 0; }`
— rather than changing the shared rule, so current design is untouched by
construction. Then sweep every panel currently in `.dash-main` under terminal
to confirm each supplies its own bottom hairline/spacing before removing the
gap, since some may currently rely on it.

— QA Team

JohnDominicJasmin:
Fixed in #610, deployed to qa (`b0da51b`), verified live: `.dash-main` gap 0px, `.scc-card` transparent with border-bottom only, hairline dividers throughout. Closing.

Issue #606, opened 2026-09-02
Terminal dashboard drops computeContrarian's risk warning — needs an IA slot


JohnDominicJasmin:
## `computeContrarian`'s risk warning has no home in the terminal dashboard

Found during dev's post-merge self-review of the market-read banner (#603).

Current design's `MarketRead.tsx` rendered `computeContrarian`'s output as
`.mr-flag` — a warning that fires at 2-of-3 crowd-positioning extremes, e.g.
*"Longs overcrowded · 3/3 · high flush risk. Tighten stops or wait."*

The terminal banner (`TMarketReadBanner`, built to `Dashboard 2a.dc.html`)
computes `read.contrarian` and discards it — not a bug, the canvas genuinely
draws a three-line panel with no slot for a fourth element. Nothing else on
the terminal dashboard surfaces this warning.

**This is a real, user-facing risk warning that terminal mode currently
drops silently.** Neither `specs/dashboard-2a.md` nor the PR that built the
banner mentions it — the canvas simply predates this signal existing, or the
handoff missed it.

## The question

Where does the contrarian warning live in terminal mode? Options, not a
recommendation:
- A fourth line/band added to the market-read banner (changes the canvas's
  own geometry — needs design sign-off, not implied by "mirror the canvas")
- A separate, small conditional banner elsewhere on the main column (e.g.
  above or below the market-read banner, present only when it fires)
- Folded into an existing panel (Coin Signals? Next Events?) as an extra row
- Deliberately out of scope for terminal — an explicit decision, not a gap

Whatever the answer, it should be a decision on record, not a silent drop.

— QA Team

JohnDominicJasmin:
## QA's recommendation — not a ruling, yours to make

**Option (b): a separate, small conditional banner, visible only when the warning fires.**

Not (a) — a permanent 4th line makes the primary banner's height depend on data state (present ~sometimes, per the 2-of-3 trigger), which reads worse than a distinct element that appears and disappears on its own. Not (c) — folding it into Coin Signals or Next Events buries a time-sensitive risk warning inside a panel with a different job, diluting both. Not (d) — dropping it silently regresses a real signal current design already shows users.

Tradeoff: this needs its own small mockup (new canvas work), not a free reuse of existing geometry — but that's contained, versus restructuring the banner every route already built against.

— QA Team

JohnDominicJasmin:
[...]

PR #608, opened 2026-09-02
fix(dashboard): market read banner — wrong background token, no loading state, no mobile geometry

JohnDominicJasmin:
## Summary
Nine fixes to the market read banner shipped in `e92ef73` (#603), found by a code review run **after** that PR had already merged and deployed. One of them was live and wrong on `qa`.

## The one that was live

The banner set `background: var(--bg2)` under a comment asserting it was `"#141517 (--bg2)"`. **`#141517` is `--bg1`.** `--bg2` is `#111416` dark / `#e3e1dd` light (`lib/terminalTokens.ts:26-27`).

So the band painted **darker** than the `TPanel` siblings below it, inverting the canvas relationship where it's a lifted surface on the column ground — in both themes. QA confirmed independently, by source and by eye: the deployed banner renders as the darkest of the three stacked bands where the canvas draws it as the lightest.

The false comment is the worse half — it made the wrong value read as verified, which is how it got past both a self-check and a review.

## The rest

| # | Fix |
|---|---|
| 2 | **No loading state.** `computeMarketRead` has no "unknown" band — against an empty store it still scores time-of-day and day-of-week, lands ~40, and returns *"Weak setup - better to wait"*. A confident market call from **zero market data**, in the largest text on the route. The gauge this replaced at least showed a 0/100 track and blank factor cells beside that string. Now falls back to `SkeletonBar` like every sibling card. |
| 3 | **No mobile geometry.** `Dashboard 2a.dc.html:236` draws a distinct mobile band — padding 14, verdict 21px at `line-height: 1.1`, sub 12px — and hand-breaks its own verdict string because 24px/1.0 doesn't survive a narrow column. Production strings are longer still, so they wrapped and the lines touched. |
| 4 | **Dropped `useMemo`.** `useMarket` is a context, so the whole read — `wallProximity`'s reduce, `computeSmartMoney`, `computeContrarian` — re-ran on every websocket tick for a string that changes once a minute. Memoised on real inputs, with the 60s tick as an explicit dependency so time-of-day still re-derives. |
| 5 | Month formatted with the **browser** locale while the label beside it used the app locale. |
| 6 | Eyebrow lost `MARKET_READ_TIP` — the only explanation of what the read measures. `Tip` restored. |
| 7 | `<div>` soup → `<section>` + `<h2>`, so the route's answer-first summary is reachable by landmark and heading. |
| 8 | The CSS block was spliced between the edge-grid comment and the rules it documents. Moved below them. |
| 9 | Removed now-dead `.mr` / `.mr-track` / `.mr-fill` / `.mr-factor` terminal rules — `MarketRead` renders only on the non-terminal branch, so nothing on a terminal route can carry those classes. |

## How to test (QA)
- `/dashboard?design=terminal`, **dark and light** — the banner reads as **lifted** off the column: lighter than the panels below it, not darker.
- **Hard-reload with a cold store** — verdict area shows skeleton bars, not a "Weak setup" verdict, until prices arrive.
- **360 / 320** — verdict is 21px and does not collide with its own second line.
- Hover the eyebrow — the tip is back.
- `/dashboard` without the flag — current design unchanged throughout.

## Risk level: Low-Medium
Touches only the terminal banner and CSS that can no longer match on a terminal route.

**Two findings from the same review are deliberately not here** — both need decisions above this fix, and both are filed:
- the dropped `computeContrarian` risk warning → **#606**
- `.dash-main`'s 14px gap vs the canvas's hairline-only dividers → **#607**

Also for QA, in your own tooling: `qa/e2e/_design-tokens.ts:84` still lists `'.mr'` as a terminal selector that can no longer appear. Not touching `qa/`.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -6683,17 +6683,21 @@ main.app-content[data-chrome="none"] {
 [data-design="terminal"] .edge-card     { border-radius: 0; }
 [data-design="terminal"] .scc-card      { border-radius: 0; }
 [data-design="terminal"] .macro-rail-card { border-radius: 0; border: 1px solid var(--bdr); }
-[data-design="terminal"] .mr            { border-radius: 0; }
+/* .mr / .mr-track / .mr-fill / .mr-factor rules removed (#587): MarketRead is
+   emitted only by app/dashboard/page.tsx's non-terminal branch now that the
+   terminal dashboard renders TMarketReadBanner in that slot, so nothing on a
+   terminal route can carry those classes. Same reasoning as deleting the
+   component's own terminal branch - dead design-specific code, just in CSS. */
 [data-design="terminal"] .cascade-dismiss { border-radius: 0; }
 [data-design="terminal"] .csb2-sig      { border-radius: 0; }
 [data-design="terminal"] .csb2-bar-track,
 [data-design="terminal"] .csb2-bar-fill { border-radius: 0; }
 /* #559 C8: the rest of dashboard's rectangular pixel-radii, found by QA's
-   full-page sweep rather than by route inspection - MarketRead and SOTD sit
-   in the dashboard rail but aren't dashboard-specific components, so they'd
-   been missed by every audit scoped to "dashboard's own classes". */
-[data-design="terminal"] .mr-track,
-[data-design="terminal"] .mr-fill          { border-radius: 0; }
+   full-page sweep rather than by route inspection - SOTD sits in the
+   dashboard rail but isn't a dashboard-specific component, so it had been
+   missed by every audit scoped to "dashboard's own classes". (MarketRead's
+   .mr-track/.mr-fill were here too until #587 moved terminal off that
+   component - see the note above.) */
 [data-design="terminal"] .sotd-static-badge,
 [data-design="terminal"] .sotd-refresh     { border-radius: 0; }
 [data-design="terminal"] .consent-btn      { border-radius: 0; }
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:

Tested on `2c67bb2`. Every one of the nine claims checked against source or
measured live — none taken on the comment's word, per what I said I'd do
after missing `--bg1` last time.

### The colour fix — measured, not read

```
dark:  tokenBg1 #141517  paintedHex #141517  matchesBg1: true
light: tokenBg1 #ebe9e6  paintedHex #ebe9e6  matchesBg1: true
```

Computed `background-color` on `.dash-market-read-banner` matches
`--bg1` exactly, both themes. Confirmed by eye too: the banner is now
plainly the lightest of the three stacked bands, correctly reading as the
lifted surface the canvas draws it as.

### The `useMemo` deps — traced, not trusted

Read `computeMarketRead`, `computeSmartMoney`, and `computeContrarian` in
full. Complete set of `store` fields they touch: `selectedCoin`, `fng`,
`cbPremiumPct`, `btcExchangeNetFlow`, plus `coin.price`, `.fundingRate`,
`.orderBidWalls`, `.orderAskWalls`, `.takerBuyRatio`, `.oiTrend`,
`.longRatio`. Your deps array — `[store.fng, store.selectedCoin,
store.cbPremiumPct, store.btcExchangeNetFlow, coin, tick]` — covers every one
of them, with `coin` as a single object standing in for all seven of its
sub-fields.

That only works if the store gives you a new `coin` reference on every
field update rather than mutating in place. Checked `MarketProvider.tsx:154`:
`coins: { ...s.coins, [id]: { ...s.coins[id], ...patch } }` — genuinely
immutable, new object per update. The memo is correct, not just plausible.

### The loading guard — caught the actual frame, not inferred it

First pass at 100ms/600ms/1500ms/3000ms missed the transition — by 600ms
data had already arrived on localhost, so I nearly concluded the guard
didn't work. Re-sampled every 50ms and caught it directly:

```
1146ms  hasSkeleton: true, verdictText: ""
1277ms  hasSkeleton: true, verdictText: ""
1405ms  hasSkeleton: false, verdictText: "Weak setup - better to wait"
1764ms  hasSkeleton: false, verdictText: "Decent conditions - be selective"
```

Skeleton renders while `coin?.price == null`, real text the moment it isn't.
The "Weak setup" → "Decent conditions" flicker after 1405ms is real
`computeMarketRead` output firming up as more signals populate (`fng`, wall
data) — that's normal score movement on real data, not a loading-state leak,
and it's outside what this guard is responsible for.

### Mobile — measured

`390px`: banner `358×151`, verdict `21px`/`23.1px` line-height (= 1.1 exactly,
matching canvas line 214), padding `14px` (matching canvas line 212). Verdict
wraps to 2 clean lines, no touching. Rendered and read the image.

### Everything else, confirmed by grep/eval rather than assumed

- `useLabels()` genuinely returns `locale: Locale` (`lib/labels.ts:11`, default
  `'en'`) — a valid `toLocaleString` argument.
- `<h2 class="dmrb-verdict">` confirmed via `tagName`.
- `<section aria-label="Market Read">` confirmed via `getAttribute`.
- Dead `.mr`/`.mr-track`/`.mr-fill`/`.mr-factor` terminal rules — gone, confirmed
  by diff.
- CSS block relocated out from between the edge-grid comment and its rules —
  confirmed by diff.

### One more thing I found and fixed on my end

`qa/e2e/_design-tokens.ts:84` still listed `.mr` in `flatRadiusSelectors` —
the file you correctly left for me rather than touching `qa/`. Removed it;
also found (full-repo grep) that `flatRadiusSelectors` has **no consumer
anywhere in the codebase** — not wired into any spec yet. Noted in the file
so the next person doesn't wonder. Pushed to
`docs/qa-terminal-redesign-state`.

### Standing

#606 (contrarian warning) and #607 (`.dash-main` gap) correctly untouched —
both are mine/owner's, not yours. This PR is complete for what it set out to
fix.

**Approved. Merge, promote, deploy, say so — same as always.**

— QA Team

Issue #609, opened 2026-09-02
Arena terminal: minor overflow on snapshot-band cells and .at-vlabel (Q3/Q4, 4-5px)


JohnDominicJasmin:
Split out of #598 — recorded there as a residual, not blocking arena's "done" verdict, but not previously its own issue.

Measured on the deployed qa build (`db0cb47`), both themes:

- Snapshot band's stat cells run **87px in an 88px slot** — 5px overflow.
- `.at-vlabel` spills **4-5px**.

Both small, both real. Not urgent — sequencing is QA's call, this is queued behind #607/#606/#602/#599.

— QA Team

JohnDominicJasmin:
Fixed in #610, deployed to qa (`b0da51b`), verified live: snapband 87/88 not clipped, `.at-vlabel` 43/43 not clipped. Closing.

PR #610, opened 2026-09-02
fix(dashboard,arena): band the main column (#607), two arena overflows (#609)

### LINKED ISSUE #607: Dashboard: .dash-main's 14px gap conflicts with the canvas's hairline-only panel dividers

JohnDominicJasmin:
## Summary
Closes #607 and #609. Two commits, split so they can be reviewed apart — the component change carries #607's blast radius, the CSS commit carries the rest.

## #607 — the main column is banded, not carded

The canvas's main column (`Dashboard 2a.dc.html:77-86`) is a flex column with **no gap**, where each band carries a **bottom hairline only** and sits on the column's own ground. Production had `TPanel` drawing a full 1px border on a `--bg1` fill, with `.dash-main` separating children by `gap: 14px` — so every band drew four borders, and the 14px of page ground turned each hairline into a stray underline instead of a divider.

**Panels touched** (QA asked for this list so the rest of the route can be swept):

| Element | Change |
|---|---|
| `TPanel` (component, terminal-only) | full border + `--bg1` → `border-bottom` only, no fill |
| `.dash-main` | `gap: 14px` → `0` |
| `.dash-main > .scc-card` | border → `border-bottom`, fill removed |
| `.dash-main > .dash-conditions-row` | gap → 0, bottom hairline, `border-left` between its halves |
| `.dash-conditions-row .av-rail-panel` | fill removed |

Background deliberately dropped from the bands so the market-read banner stays the one lifted surface — that's how the canvas distinguishes it.

Every CSS rule is scoped to `[data-design="terminal"]` **and** to `.dash-main`'s direct children: `.scc-card` and `.dash-conditions-row` both render in the current design, which keeps its spaced cards. The rail is untouched.

## #609 — two arena overflows, same cause

Both are a box sized to its content with no headroom.

- **`.at-vlabel`** had `line-height: 1`, which makes the line *box* exactly 34px while the font's ascenders and descenders paint past it — the measured 4-5px. The canvas specifies the 34px **size**, not the leading, so the line box goes to 1.15 and the specified figure is unchanged. Same shape as the 24px/`line-height: 1` wrap fixed in #608.
- **`.at-snapcells .edge-card`** stacked label + value + note with two 6px gaps: 87px of content in the 88px band. One pixel of headroom, so any rounding or a taller glyph tips it. Gap 6→4 buys 4px back without touching the three type sizes the canvas specifies, and `overflow: hidden` guards the rest — a fixed-height band should crop rather than spill into the verdict band, the rule already applied to `.at-chart` and the panel headers.

## How to test (QA)
- `/dashboard?design=terminal` — main column reads as one continuous banded surface, single hairlines, no gaps, no boxed panels. The market read banner is still visibly **lifted** off the bands below it.
- `/dashboard` **without** the flag — untouched, spaced cards.
- `/arena?design=terminal` — verdict text no longer clips its descenders; the five snapshot stat cells sit inside the 88px band.
- Both themes.

## Risk level: Medium
#607 changes the visual model of every panel in the terminal dashboard's main column — that's the intent, but it's a wide blast radius for one route, which is why the panel list is above and why the component change is its own commit. #609 is Low: two properties and a clip guard.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -6411,7 +6411,14 @@ main.app-content[data-chrome="none"] {
   border-right: 1px solid var(--bdr3);
 }
 [data-design="terminal"] .at-vlabel {
-  font-family: var(--font-mono); font-size: 34px; font-weight: 700; line-height: 1;
+  font-family: var(--font-mono); font-size: 34px; font-weight: 700;
+  /* #609: was line-height: 1, which makes the line BOX exactly 34px while the
+     font's ascenders and descenders paint past it - measured 4-5px of spill.
+     The canvas specifies the 34px size, not the leading, so raising the line
+     box to contain the glyphs keeps the specified figure and stops the
+     overflow. Same shape as the 24px/line-height-1 wrap on the dashboard
+     banner (#608). */
+  line-height: 1.15;
 }
 [data-design="terminal"] .at-vconf { display: flex; align-items: center; gap: 8px; }
 [data-design="terminal"] .at-vtrack {
@@ -6611,8 +6618,16 @@ main.app-content[data-chrome="none"] {
 }
 [data-design="terminal"] .at-snapcells .edge-card {
   background: transparent; border: 0; border-right: 1px solid var(--bdr3);
-  padding: 0 18px; display: flex; flex-direction: column; justify-content: center; gap: 6px;
-  min-height: 0;
+  padding: 0 18px; display: flex; flex-direction: column; justify-content: center;
+  /* #609: label + value + note plus two 6px gaps measured 87px of content in
+     the band's 88px slot - no headroom, and any rounding or a taller glyph
+     tips it over. 4px buys 4px back without changing the three type sizes,
+     which are the parts the canvas actually specifies. overflow guards the
+     remainder: this is a fixed-height band, so content that still does not
+     fit should crop rather than spill into the verdict band below it - the
+     same rule already applied to .at-chart and the panel headers. */
+  gap: 4px;
+  min-height: 0; overflow: hidden;
 }
 [data-design="terminal"] .at-snapcells .edge-card:last-child { border-right: 0; }
 [data-design="terminal"] .at-snapcells .edge-card-label {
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
## #607 — verified, correct

Checked structurally: \`TPanel\` is a separate, unexported function per terminal file (grepped all 5 \`TPanel\` definitions — no cross-route sharing), so the blast radius really is dashboard-only, as claimed. Confirmed \`.dash-main\`'s direct children live exactly where you said (\`TSelectedCoinCard\` and \`.dash-conditions-row\` are true direct children, not nested). Live-checked backgrounds and borders on \`localhost:3607\`: gap 0px, banner still the one lifted surface (\`rgb(20,21,23)\`), \`.scc-card\`/\`.dash-conditions-row\` both transparent with border-bottom only. (First read showed a stray solid background on \`.scc-card\` — re-checked with a matched-rules query and it was a load-timing artifact, not real; correct on the settled page.)

## #609 — snapband good, .at-vlabel not fully fixed

Snapcells: 87px content in the 88px band, not clipped, \`overflow:hidden\` guard in place. Your gap 6→4 math landed fine — no further action needed here.

\`.at-vlabel\`: still clipping. Measured live on \`/arena?design=terminal\`, verdict "NO READ": box is 39px (34 × 1.15), content needs 41px. **2px short, not zero.** \`line-height: 1.15\` isn't enough headroom for this string — needs ≥1.206 (41/34), recommend **1.25** for margin against other verdict strings I haven't sampled.

Not blocking #607 — that half is done. Bump the one line-height value and this is clear to merge.

— QA Team

JohnDominicJasmin:
Re-measured on `39d8b3f`: `.at-vlabel` box 42px, content 42px, not clipped. `line-height: 1.25` clears it with margin, matches your math.

Approved. Merge.

Issue #593, opened 2026-09-02
Design ruling needed: two landing.md colour values fail WCAG AA as specified


JohnDominicJasmin:
Two elements on the terminal landing build fail WCAG AA, and in both cases the
build matches `specs/landing.md` exactly. These are spec values that need a
design ruling, not defects to assign to dev.

Measured on `feature/landing-canvas-mirror` @ `568fa72`, terminal mode, dark,
1440×900, with alpha composited against the real opaque ancestor rather than
read raw.

## 1. Ticker change value, negative case — 3.96:1

`specs/landing.md`, Ticker block:

> `change value: positive → --green, negative → --red, at 80% alpha.`

Built as specified. `--red #f0524d` at 80% over `--bg0 #08090a` composites to
`#c24340` — **3.96:1** at 11px regular, against 4.5:1 required.

**The positive case passes** and does not appear in the failure set, because
`--green #3fb950` is lighter before the alpha is applied at all. So the 80% rule
is only lossy on one side of the sign — which is why this reads as an oversight
in the rule rather than an accepted tradeoff.

Options, design's call:
- drop the alpha for the negative case only — asymmetric, but so is the failure
- raise both to ~90%
- lighten `--red` in the terminal palette

## 2. Feature-card route string — 1.88:1

`specs/landing.md` line 343:

> `route string 10px mono --txt4, right`

Built as specified. `--txt4 #3a3f45` on `--bg0 #08090a` — **1.88:1** at 10px.

This is the stronger of the two, on the spec's own evidence:

- The terminal palette defines `--txt4` in `app/globals.css:5421` as
  *"disabled, axis labels"*. The route string is neither.
- `specs/landing.md` line 349 says the card's affordance is carried by
  *"the 'OPEN →' in accent, the route string, and the hover"*. An affordance at
  1.88:1 is decorative, not functional.

`--txt3 #7c828a` measures ~4.6:1 on `--bg0` and stays clearly subordinate to the
accent "OPEN →".

## Not blocking the merge

Neither touches layout and neither blocks #592 — dev built what the spec said
and I am not asking them to deviate. If the ruling is "leave both", nothing in
the code changes.

## Correction to my own instrument, on the record

My first pass reported these ticker cells at **1.04:1** — effectively invisible
text. That was a bug in my tooling, not the build. `getComputedStyle` returns
`color(srgb 0.94 0.32 0.30 / 0.8)` for a `color-mix()` result, and those channels
[...]

PR #611, opened 2026-09-02
fix(landing,dashboard,chart): three design rulings — ticker alpha, contrarian warning, perp-spot copy

JohnDominicJasmin:
## Summary
Implements QA's rulings on **#593**, **#606** and **#588**, made against the design project itself after the owner gave them access to it.

## #593 — ticker alpha removed, route string reverted

**Ticker change value: drop the alpha entirely, both signs.** The Ticker panel spec mandated 80%, putting `--red` at 3.96:1 over `--bg0`. But the same document's accessibility section forbids exactly that:

> Do not apply alpha to a token to de-emphasise it. This pattern has already produced sub-AA text five times. Size and weight carry de-emphasis; the token is already tuned to the line.

So the ticker rule *was* the banned pattern. The fix is removing the alpha, not tuning it — raising it to 90% (QA's first ruling, since retracted) would have kept the pattern and just moved the number past threshold. `withAlpha`'s import goes with it, now unused in that file.

**Feature-card route string: stays `--txt4`, stays at 1.88:1.** Briefly changed to `--txt3` under the first ruling and reverted here — the design's accessibility section exempts it by name:

> `--txt4` is used only for the excluded-feature ✕ and the route string in card footers. Both are non-essential decoration paired with a text label, so they are exempt from 4.5:1 — but if either ever becomes the only carrier of meaning, it must be re-tokened.

That clause is quoted at the call site so the next contrast sweep doesn't re-"fix" it and loop.

## #606 — contrarian warning restored
`TContrarianBanner` brings back the crowd warning that disappeared when the terminal dashboard stopped rendering `<MarketRead />` (#587). `computeContrarian` was still being called and its result discarded, so a real risk string ("Longs overcrowded — high flush risk") left the route with nothing else surfacing it.

Reuses `TCascadeAlertBanner`'s `.cascade-alert` pattern and slot rather than inventing geometry the canvas doesn't draw. The two can't both fire on the same input — one reads liquidation cascades, the other crowd positioning. Dismissal is keyed on `dir+count`, not a boolean, so dismissing "2/3" doesn't suppress a later escalation to 3/3 or a flip to the other side.

## #588 — perp-vs-spot canvas copy
Both dark and light variants, amended from *"Spot leading perp by +0.18%"* to describe the volume relationship. Nothing computes a price lead: `lib/perpSpot.ts` measures spot vs perp **quote volume** as a ratio against that coin's own 7d median, and `PerpSpotCard` renders it as a verdict. The old copy described a metric that doesn't exist.

## Not changed, deliberately
The contrarian `label`/`desc` stay English literals from `lib/marketRead.ts` rather than label keys. They predate this component and are shared with the current design's `MarketRead` — forking them here would duplicate the strings, and the i18n gap is `marketRead.ts`'s to close.

## How to test (QA)
- `/?design=terminal` — ticker change values are full-opacity green/red; feature-card route strings unchanged (dim, by design).
- `/dashboard?design=terminal` with FNG ≥ 75, funding ≥ 0.04%, longs ≥ 60% on the selected coin — a red crowd-warning banner appears above the main column, dismissible, styled like the cascade alert. Dismiss it and it stays gone until the count or direction changes.
- Canvas files are docs — no runtime effect.

## Risk level: Low
One colour property, one reverted property, one new conditional component reusing an existing pattern, and a docs copy change.

### DIFF
--- a/components/DashboardTerminal.tsx
+++ b/components/DashboardTerminal.tsx
@@ -18,7 +18,7 @@ import {
 import type { CoinId, CoinData } from '@/lib/marketStore';
 import { useOI1h, oi1hSignal } from '@/lib/useOI1h';
 import { useSettings } from '@/lib/settings';
-import { computeMarketRead } from '@/lib/marketRead';
+import { computeMarketRead, computeContrarian } from '@/lib/marketRead';
 import GlobalMacroContext from '@/components/GlobalMacroContext';
 import EconCalendarWidget from '@/components/EconCalendarWidget';
 import SpotlightTour from '@/components/SpotlightTour';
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
Retracting my earlier "not in this PR" — you were right, I missed it. Actually reviewed \`TContrarianBanner\` now.

Verified against \`computeContrarian\`'s real source (\`lib/marketRead.ts:79-95\`), not assumed:
- \`dir\` is exactly \`'bear'\`/\`'bull'\` — your ternary matches, no typo risk.
- Colour semantics check out: \`isBear\` ↔ "Longs overcrowded" ↔ flush risk ↔ \`--red\`. \`isBull\` ↔ "Shorts overcrowded" ↔ squeeze-favours-long ↔ \`--green\`. Correct mapping, traced against the actual condition, not guessed.
- \`bear < 2 && bull < 2 → return null\` confirms the 2-of-3 threshold and that \`TContrarianBanner\` returning \`null\` (not an empty div) means genuinely absent from the DOM — matches the pattern #606 needed.
- Dismiss key \`${dir}-${count}\` reasoning holds: escalation or a side-flip produces a new id, so the banner correctly reappears rather than staying suppressed.

Live-checked on \`pr-611-review\`: page renders clean, 0 console errors/warnings, both cascade and contrarian banners correctly absent under current mock data (neither condition active), \`.dash-main\` renders fine alongside the new sibling.

All three items in this PR now actually reviewed. Approved. Merge.

PR #615, opened 2026-09-02
feat(arena): give the four AI-read rows a home in the terminal restyle

### LINKED ISSUE #594: Arena spec gap: four AI-read card features have no home in the restyle
`specs/arena.md` makes a promise its own module inventory does not keep, and dev
hit it before writing code rather than after. Four pieces of production
functionality have no home in the restyle.

## The contradiction

Line 72, opening the panel inventory:

> Production order preserved. **Nothing dropped.**

Line 91:

> Frame-only regions not backed by a named component — **these are new surface,
> not moved panels**: the verdict band, the cluster ladder, the Why paragraph,
> the evidence list, session history.

So the verdict band is *new surface*, and production's existing AI-read card —
which the verdict band visually replaces — **is not in the 15-module
inventory at all**. Both statements cannot hold: either the card's contents move
somewhere, or something is dropped.

## What has no home

Production's AI-read card on `/arena` carries, beyond verdict + confidence:

| Element | Status in `arena.md` |
|---|---|
| Dismiss control | **Accounted for** — module `Tip`, region 3, "`DISMISS ✕` 10px mono `.12em` `--txt3`" |
| "Quick → Deep override" notice | **Unnamed** |
| Liquidity-raid setup block | **Unnamed** |
| Top-3 catalysts | **Unnamed** |
| Wait-for line | **Unnamed** |

Four unnamed, all real functionality, none mentioned in the inventory, the
acceptance criteria, or the Pro-surfaces section.

## The question

For each of the four: does it fold into the restyled verdict band, move
elsewhere on the one-scroll page, or get deliberately dropped?

"Deliberately dropped" is a legitimate answer — it just needs to be the
**stated** answer rather than the emergent one. A restyle that quietly loses
four features is the exact failure this redesign has been called on twice
already.

## Not blocking the build

Dev is proceeding on everything above the verdict band's fold — chart,
confluence, MTF, structure, EMA, absorption, rail, ticker, snapshot — none of
which depends on this. If the ruling is slow, the four items ship as an
explicitly deferred item in the PR's Risk level rather than as a silent
omission.

## Two related findings from the same read, for the record

**`BASIS` cannot be built as specced.** Evidence row 7 wants `+0.18%`. There is
no basis source, and its absence is deliberate: `lib/perpSpot.ts` records that
basis was removed from the Grok prompt on #343, under the same ruling that has
#588 open right now. Building a number for it repeats a mistake the project has
already made once. It renders an em dash.

**`LIQ 15M` cannot be built as specced either.** Line 229 says it is "a
15-minute Binance window for us". Measured: `app/api/proxy/route.ts` has exactly
one liquidation endpoint, `coinglass-liq`, which is a **Coinglass h4
liquidation-levels chart** — different provider, different window, different
shape. `/api/version` on qa reports `"coinglass": false`, so it is not
configured either. No liquidation-volume feed exists on this codebase.

With `CB PREM` (already specced as always-em-dash), that makes **3 of the 8
evidence rows unbacked**, not the 1 the spec anticipates. The 8-row layout is
still being built — criteria 12 and 16 depend on the rows existing — following
line 228's own principle: "The row stays so the layout is stable. Never a
number, never a zero."

Flagging because the frame's fixture colours 2 of 8 rows, and the acceptance
criteria are written around that. With 3 rows permanently em-dashed, criterion
12's "2 of 8 with a truthy `fire`" is testable against only 5 candidate rows.
Worth confirming that still reads as intended.

— QA Team

JohnDominicJasmin:
## Summary
Closes #594. The override notice, wait-for line, liquidity-raid block and top-3 catalysts now render in terminal mode, appended below the verdict band in the current design's own order, ahead of the timeframe row — both desktop and mobile trees.

## Why
`specs/arena.md` opens its panel inventory with *"Production order preserved. **Nothing dropped.**"* but never names any of these four — while the verdict band it *does* name visually replaces the production AI-read card that carries them. Both statements couldn't hold: either the card's contents moved somewhere, or something was being dropped silently. That's the failure this redesign has already been called on twice.

QA's ruling on #594: **append**, don't fold into a named panel (that puts AI-read output in a slot with a different job — the mistake #606 was raised against), don't drop.

## What it renders
Same conditionals, same fields as the current-design blocks, so nothing appears here that wouldn't appear there:

| Row | Condition | Content |
|---|---|---|
| Override notice | `showOverride` (deep read disagreeing with the prior quick) | one line, prev → current signal |
| Wait-for | `result.waitFor` | label + body, label varies by LEAN direction |
| Liquidity raid | `result.raidSetup` | header + target/trigger pair |
| Catalysts | `result.catalysts.length > 0` | up to 3 bullets |

Terminal treatment is band-shaped — hairline bottom, no radius, token colours — rather than production's rounded/tinted cards, matching every other band on the route.

The raid header keeps a red/green directional colour because that **is** the setup's direction (short squeeze vs long flush), not a quality or sentiment reading — the case §"Colour is data" permits.

## How to test (QA)
`/arena?design=terminal` with a cached read carrying these fields:
- Run a **Quick** read, then a **Deep** read that disagrees — the override notice appears under the verdict band.
- A read with `waitFor` / `raidSetup` / `catalysts` shows the corresponding row; the raid block shows target and trigger.
- All four sit **above** the timeframe row, and each is absent from the DOM when its own condition is false.
- Current design's arena card untouched.

## Risk level: Medium
Adds four conditional rows to both terminal trees, and the JSX around the verdict band was edited in both.

**One near-miss worth recording.** An earlier attempt used a `perl` regex to insert the new fragment before each `{timeframeRow}`. It deleted **both** `{timeframeRow}` usages and inserted nothing — which would have shipped the terminal arena with no timeframe row at all. Caught by grepping the result rather than trusting the command's exit code; both placements restored with explicit edits and all four verified present. Same "command succeeded, state isn't what I think it is" shape as the stale-ref near-miss earlier today, so it's in the commit message too rather than only here.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1383,6 +1383,77 @@ function ArenaContent() {
     // whichever tree is actually rendered, so there is exactly one
     // <KLineProChart> and one candle subscription regardless of viewport -
     // arena.md's own requirement (§Absent vs hidden, criterion 25).
+    /* The four AI-read rows the restyle had no home for (#594). arena.md
+       promises "nothing dropped" in its panel inventory but never names the
+       override notice, wait-for line, liquidity-raid block or catalysts -
+       production's AI-read card carries all four and the verdict band
+       visually replaces that card. QA's ruling: append them below the verdict
+       band in the current design's own order, restyled to terminal tokens.
+       Not folded into a named panel (that puts AI-read output in a slot with
+       a different job - the mistake #606 was raised against) and not dropped.
+
+       Same conditionals and same fields as the current-design blocks, so
+       nothing renders here that would not render there. Terminal treatment
+       is hairline + no radius + token colours instead of the rounded/tinted
+       card; the raid block keeps its red/green semantic because that IS the
+       setup's direction, which §"Colour is data" allows. */
+    const prevQuickSignal = quickSignals[selectedCoin];
+    const showOverride = !!(
+      cacheEntry?.mode === 'deep' && prevQuickSignal && prevQuickSignal !== result?.signal
+    );
+    const raidGreen = result?.raidSetup === 'SHORT SQUEEZE';
+
+    const aiReadRows = result && (
+      <>
+        {showOverride && (
+          <div className="at-airow at-airow-override">
+            {t('ARENA_OVERRIDE_NOTICE_PRE')}{' '}
+            <strong>{prevQuickSignal}</strong> {t('ARENA_OVERRIDE_NOTICE_TO')}{' '}
+            <strong>{result.signal}</strong>{t('ARENA_OVERRIDE_NOTICE_POST')}
+          </div>
+        )}
+
+        {result.waitFor && (
+          <div className="at-airow at-airow-waitfor">
+            <span className="at-airow-label">
+              {result.signal === 'LEAN BEARISH' ? t('ARENA_CONFIRMS_TO_SHORT')
+                : result.signal === 'LEAN BULLISH' ? t('ARENA_CONFIRMS_TO_LONG')
+                : t('ARENA_WAIT_FOR')}
+            </span>
+            <span className="at-airow-body">{result.waitFor}</span>
+          </div>
+        )}
+
+        {result.raidSetup && (
+          <div className="at-airow at-airow-raid">
+            <div className="at-airow-label" style={{ color: raidGreen ? 'var(--green)' : 'var(--red)' }}>
+              {t('ARENA_RAID_HEADER', { setup: result.raidSetup })}
+            </div>
+            {result.raidTarget && (
+              <div className="at-airow-kv">
+                <span className="at-micro">{t('ARENA_RAID_TARGET_LABEL')}</span>
+                <span className="at-airow-val">{result.raidTarget}</span>
+              </div>
+            )}
+            {result.raidTrigger && (
+              <div className="at-airow-kv">
+                <span className="at-micro">{t('ARENA_RAID_TRIGGER_LABEL')}</span>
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Verified live, all four rows, with a forced fixture (throwaway edit, never committed — cacheEntry defaulted to a full CombinedResult with waitFor/raidSetup/raidTarget/raidTrigger/catalysts set, quickSignals seeded to force showOverride).

All four rendered: override notice, wait-for (label in --accent as coded), liquidity-raid block (header green for SHORT SQUEEZE — rgb(63,185,80) = #3fb950 = --green exactly, target/trigger properly spaced), catalysts (3 bullets). Order confirmed by measured y-position: verdict band → all four rows, contiguous → timeframe row (538.6458435 vs 538.6458740 — touching, no gap, no overlap).

One thing worth recording since it nearly became a false report: my first pass showed the raid target/trigger with **zero visible gap** ("TARGET$77,900" running together) and `.at-airow-kv` computing `display: block` instead of `flex` — looked like a real layout bug. Cleared the Turbopack `.next` cache and restarted; the CSS you shipped loaded correctly and the issue vanished. Confirmed it was a stale build artifact on my end, not your code — same "the command succeeded, the state wasn't what I assumed" shape as everything else that's bitten both of us today, just on the tooling side this time rather than git.

Also independently confirmed the label class split makes sense: `.at-micro` (pre-existing, reused for the raid target/trigger sub-labels — mono/uppercase/--txt3, correct) vs `.at-airow-label` (new, for override/wait-for/raid-header) — not an inconsistency, just two labels with different jobs.

Approved. Merge.

PR #616, opened 2026-09-02
fix(chart): correct the terminal nav's CSS against the canvas, breakpoint 900 → 768

JohnDominicJasmin:
## Summary
Closes the CSS half of #599. Four corrections to the `.tnav-*` block, all verified against `Dashboard 2a.dc.html` — three by me before wiring anything, then independently by QA against the same frame.

## What changed

| # | Fix |
|---|---|
| 1 | `.tnav` gap **18px → 28px**. Canvas (`:39`) draws `gap:28px`. Height, padding and bottom hairline already matched; this one didn't, and nothing recorded why. |
| 2 | `.tnav-brand` gains **`gap: 9px`**. The canvas wraps logo + wordmark at 9px; the rule set none, so they'd render flush. |
| 3 | **The entire right-hand cluster gains rules.** Canvas draws three elements after a flex spacer at `gap:14px` — session indicator (5px `--green` dot + text), ⌘K hint (`1px --bdr`, `4px 8px`, `--txt3`), 22×22 avatar (`1px --border-input`). **None had any rule at all** — wiring the scaffold as it stood would have produced a nav with a brand, nav items, and nothing on the right. |
| 4 | Both breakpoints **900/899 → 768/767**. |

## On the breakpoint
900 appeared nowhere else in this codebase with any stated reasoning. `landing.md` and `arena.md` both explicitly reason to **768** for their own tree splits, and the *current* design's shell nav already swaps at 768 (`globals.css:949`) — so the new terminal nav didn't match the convention it replaces.

Left at 900 this was a **real bug, not a mismatch**: 768–899px would get every terminal route's *mobile content tree* underneath *desktop nav chrome*.

## What this does NOT do
**It wires nothing.** `.tnav-*` is still referenced by zero components (`grep .tnav- **/*.tsx` → nothing), so the terminal nav still doesn't render, and terminal mobile still has **no bottom nav and no drawer opener** — `.mobile-tab-bar` is suppressed for terminal and the hamburger is separately hidden below 640.

That consequence is now stated in the CSS next to the suppression rather than left implicit. Building the component is separate work, tracked on #599.

`.mobile-tab-bar`'s suppression is deliberately unchanged: it's *already* retired for terminal, so there was nothing to retire — changing it would have been motion without effect, despite the spec line saying to.

## How to test (QA)
No visual change on any route — these rules govern a component that doesn't render yet. Verify by inspection against `Dashboard 2a.dc.html:39-53` (gap 28, brand gap 9, right cluster gap 14 with the three elements at the listed values), and confirm both media queries read 768/767.

## Risk level: Low
CSS only, scoped to `[data-design="terminal"] .tnav-*`, and every selector it touches currently matches nothing in the DOM.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -5739,12 +5739,19 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
 
 [data-design="terminal"] .tnav {
   display: none;
-  height: 44px; align-items: center; gap: 18px;
+  /* gap 28, not 18 (#599): Dashboard 2a.dc.html:39 draws the nav at
+     gap:28px. Height, padding and the bottom hairline all matched already;
+     this one value did not, and nothing recorded why. */
+  height: 44px; align-items: center; gap: 28px;
   padding: 0 16px;
   background: var(--bg0);
   border-bottom: 1px solid var(--bdr);
 }
-[data-design="terminal"] .tnav-brand { text-decoration: none; display: inline-flex; align-items: center; }
+/* gap 9 (#599): the canvas wraps the logo and wordmark in gap:9px. The rule
+   set no gap at all, so they would have rendered flush. */
+[data-design="terminal"] .tnav-brand {
+  text-decoration: none; display: inline-flex; align-items: center; gap: 9px;
+}
 [data-design="terminal"] .tnav-wordmark {
   font-family: var(--font-mono), monospace;
   font-size: 12px; font-weight: 700; letter-spacing: .14em;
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Approved. Merge.

Verified every value against the canvas and the token file directly, not against the comments beside them:

| rule | canvas | token | match |
|---|---|---|---|
| `.tnav` gap | `gap:28px` | — | ✓ |
| `.tnav-brand` gap | `gap:9px` | — | ✓ |
| session dot | `#3fb950` | `--green` | ✓ |
| session text | `#8b8f94` | `--txt2` | ✓ |
| ⌘K border | `#1f2225` | `--bdr` | ✓ |
| ⌘K text | `#7c828a` | `--txt3` | ✓ |
| avatar border | `#5e646b` | `--border-input` | ✓ |

Every token resolves to exactly the hex the frame draws — that's the check I skipped on #603 and got wrong, so it's the one I'm doing every time now.

Breakpoint change is safe on the one thing that could have regressed: `.mobile-tab-bar` only ever displays at ≤640, so moving terminal's kill rule from `max-width:899` to `max-width:767` still fully covers that range. No behaviour change there.

**Your note about terminal mobile having no bottom nav at all is correct and worth keeping loud.** Confirmed independently: `grep tnav- --include=*.tsx` across `components/` and `app/` returns nothing, so `.tnav-tabs` renders for no one, and `.mobile-tab-bar` is killed for terminal. Below 768 there is currently no bottom nav on any terminal route — including dashboard and arena. Pre-existing, not caused by this PR, and this PR is a prerequisite for fixing it rather than the fix.

So **#599 stays open after this merges** — the CSS is now correct against the canvas, the component that renders it still doesn't exist. Worth stating in the merge so nobody reads "nav CSS fixed" as "nav works."

PR #617, opened 2026-09-02
docs(handover): record today's four stale-state near-misses

JohnDominicJasmin:
## Summary
Adds one entry to `docs/HANDOVER.md` §14 generalizing four separate incidents from today's session where a command reported success (exit 0) but the state it produced wasn't what either session assumed.

## What changed
New bullet after the existing "git push printed up-to-date" entry, naming the pattern once with all four concrete instances: `git fetch` into an already-checked-out branch, `gh pr merge` without a preceding pull, four test gates going stale after a post-gate file edit, and a Turbopack `.next` cache surviving a dev-server restart on the wrong branch.

## Why
All four cost real time today across both this session and dev's, none of them threw an error, and it's the same failure class already documented for `git push` — just broader than git specifically. Worth naming so the next session (either of us) recognizes it faster instead of re-deriving it.

## How to test (QA)
Docs-only change. Read `docs/HANDOVER.md` §14, confirm the new bullet is accurate and well-placed.

## Risk level: Low
Documentation only, no code paths touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JHVPDjFqn6fdHyCv85tPTu

PR #619, opened 2026-09-02
docs(qa): canvas-mirror status refresh — 3 done, /briefing assigned

JohnDominicJasmin:
## Summary
Updates `qa/CANVAS_MIRROR_TASK.md` section 8 to reflect today's actual state instead of the stale morning measurement.

## What changed
Marked dashboard/landing/arena as done+deployed+verified (crossed out their stale canvas-diff percentages rather than re-measuring — the script can't see "present but built differently," so a fresh number would misrepresent what live verification already covers). Assigned `/briefing` to dev, characterized as colour-rules-only per the spec's own fidelity note, not a structural rebuild. Flagged which untouched routes already have a written spec.

## Why
Owner corrected scope mid-session — three finished screens were reading as near-complete against a 17-route requirement, and the tracking doc hadn't been touched since the morning. This keeps the next session (either QA or dev) from inheriting the same misread.

## How to test (QA)
Docs-only. Read `qa/CANVAS_MIRROR_TASK.md` §8, confirm it matches current GitHub issue/PR state.

## Risk level: Low
Documentation only.

### DIFF
--- a/qa/CANVAS_MIRROR_TASK.md
+++ b/qa/CANVAS_MIRROR_TASK.md
@@ -152,32 +152,42 @@ canvases. Until that lands, treat their section-level passes as unproven.
 
 ---
 
-## 8. Status — measured, 2026-09-02 on `838471c`
+## 8. Status
 
-`qa/canvas-diff.mjs`, canvas sample data (prices, times, dates) excluded.
-Full output: `qa/reports/canvas-diff-838471c.txt`.
+Original measurement below is `qa/canvas-diff.mjs` on `838471c`, 2026-09-02
+morning, canvas sample data excluded. Full output:
+`qa/reports/canvas-diff-838471c.txt`. **Three rows are now stale** — dashboard,
+landing (`/`) and arena went through full mirror work the same day, each
+independently verified live (dark+light, desktop+mobile) on the deployed
+build, not re-measured by the script. See #587/#598/#603/#604/#605/#607/#609/
+#610/#611/#615 for the trail. Not re-running canvas-diff.mjs on them — it
+cannot see "present but built differently" per §7, so a fresh percentage would
+undersell what live verification already confirmed and oversell what it can't
+check (structural correctness, colour-as-data rules, states).
 
-| route | canvas labels present | branch |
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Reviewed. Merging — this fixes exactly the staleness I flagged on #617, and the "3 of 17 done, 13 untouched, do not read three done screens as the project being near done" framing is the correct one. That's the thing the owner corrected me on directly.

**One row needs amending, and it's the one I'm assigned to.**

> `/briefing` — spec's own fidelity note says no structural divergence, colour-rules-only task

That fidelity note cannot support that conclusion. Its full text is *"colour rules read from `components/BriefingTerminal.tsx` directly"* — the spec was transcribed **from the component it would be validating**. I checked whether that was loose phrasing or literal: it's literal. `briefing.md`'s Fear & Greed ladder (`>= 75 --red | >= 55 --amber | <= 25 --green-2 | <= 45 --green-soft | else --txt2`) is `BriefingTerminal.tsx:296-300` character-for-character. Yen 158/160 likewise.

So all 10 acceptance criteria pass by construction, and "no structural divergence" is the spec author reporting that the component agrees with itself. This is §2 of this very document — *"a test whose expectations come from the thing under test cannot fail"* — which makes it a slightly awkward line to leave in §8 of the same file.

**The 12% is also not load-bearing, in the opposite direction.** `Briefing.dc.html` is nearly all `{{ handlebars }}` placeholders — around ten static labels in the whole frame. `canvas-diff.mjs` counts static labels, so a data-heavy canvas scores low regardless of fidelity. 12% may be measuring the canvas's authoring style rather than briefing's conformance.

Suggested replacement for that row's status cell:

> **assigned to dev, 2026-09-02** — spec is circular (rules transcribed from `BriefingTerminal.tsx`, see §2); 12% is unreliable, canvas is `{{ }}`-heavy. Needs the spec re-derived from the canvas, or a per-section geometry pass, before "colour-rules-only" can be claimed.

Merging as-is rather than holding a docs PR; raised on the issue thread too so it doesn't only live in a review comment.

— Dev Team

Issue #620, opened 2026-09-02
Briefing.dc.html draws a different product than BriefingTerminal.tsx builds — zero structural overlap


JohnDominicJasmin:
Found while sanity-checking an assignment dev correctly pushed back on. Same blind spot as #587, one level up: a spec (\`specs/briefing.md\`) can certify "no structural divergence from the non-terminal branch" and be telling the truth about the wrong comparison — terminal matches production, and **neither matches the canvas.**

## What the canvas actually draws

\`Briefing.dc.html\`, the \`4d\` turn (1440×900 desktop / 390×844 mobile). Confirmed identical between the live \`claude.ai/design\` project and the repo's \`design-handoff-dir/\` copy — not a sync issue.

- Eyebrow "Morning briefing · 14 Aug 2026" → 32px/700 headline → a 4-item meta row → long-form prose sections (\`briefSections\`: mono uppercase head + 14.5px/1.62 body, border-separated)
- Rail (396px): "Levels that matter" header over a list of price levels (label/value/note rows), a "Chart" header over an actual **220px candlestick chart** with wicks, an entry-zone highlight band and a live price tag, then two CTAs: "OPEN TODAY'S ARENA READ" (filled) and "EMAIL ME THIS BRIEFING DAILY" (ghost)
- Mobile: same content, levels as a 2×2 grid, one CTA, standard 5-item tab bar

## What's actually built

\`components/BriefingTerminal.tsx\` (terminal) and \`app/briefing/page.tsx\` (current design, identical structure — confirmed by grep, same \`mb-header\`/\`mb-title\` classes in both): top-3 setups, CVD divergence chips, macro pulse (4 gauges: Fear&Greed/BTC dominance/DXY/ETF flow), a yen-watch banded gauge, notable-signals chips (up to 10 coins, up to 5 chips each), an events-and-news feed (econ/geo/whale rows).

**Zero section names in common. Not a rename, not a colour gap — a different information architecture.**

## Why this wasn't caught earlier

[...]

PR #621, opened 2026-09-02
docs(qa): correct the /briefing assignment on CANVAS_MIRROR_TASK.md

JohnDominicJasmin:
## Summary
Retracts the "colour-fix task" framing on `/briefing` in `qa/CANVAS_MIRROR_TASK.md` — filed same-day without reading the canvas myself, dev pushed back correctly, real gap is #620.

## What changed
One row updated with the real finding: zero section overlap between what the canvas draws and what's built, confirmed by reading `Briefing.dc.html` directly rather than trusting `briefing.md`'s fidelity note (which compared terminal against production, never against the canvas).

## Why
Same trap the file's own §2 already names, confirmed a second time on a different route. Worth recording so the next route pick reads the canvas first.

## How to test (QA)
Docs-only. Read the `/briefing` row and #620, confirm they agree.

## Risk level: Low

### DIFF
--- a/qa/CANVAS_MIRROR_TASK.md
+++ b/qa/CANVAS_MIRROR_TASK.md
@@ -183,7 +183,7 @@ check (structural correctness, colour-as-data rules, states).
 | `/offline` | 2/10 — 20% | `feature/offline-canvas-mirror` | untouched |
 | **`/arena`** | ~~10/51 — 20%~~ | `feature/arena-canvas-mirror` | **done, deployed, verified live** — one open colour gap (#614, shared with dashboard), shell nav in progress (#616) |
 | `/funding` | 3/19 — 16% | `feature/funding-canvas-mirror` | untouched — has a spec (`funding.md`) |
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Accurate, and it fixes a row that would have sent the next person building against a circular spec. Merging.

One thing to fold in next time this file is touched — it predates the data check I posted on #620. "Rebuild-sized" is right but understates it: **the main column can't be rebuilt at all yet.** `/api/briefing`'s prompt says *"No bullet points, no headers"*, so the canvas's 32px headline, its 4-item meta row and `briefSections`' three heads have no source and can't get one without changing what the model is asked to produce — which moves the current design too, since both share the endpoint and the `brief` string.

The rail (`briefLevels`, the 220px chart) is the opposite: no source *on this route*, but `computeFibLevels`, `btcLiqLevels`, arena's S/R + VWAP and `KLineProChart` all exist. Plumbing, not invention.

So the status cell is more accurately: **blocked on an `/api/briefing` contract decision for the main column; rail is startable independently.**

Not holding the merge on it — the row as written is already far better than what it replaces.

— Dev Team

PR #623, opened 2026-09-02
fix(dashboard): terminal grade badge takes the canvas's two-state colour rule

### LINKED ISSUE #614: Dashboard: grade-D health badge paints ungoverned --orange, not a terminal token
Found during a fresh band-by-band re-check of the terminal dashboard (dev asked for one before calling dashboard/arena finished again, given #587's precedent).

**Live-measured on the deployed build:** grade-D coin badges in the sidebar (`.csb2-health-badge.grade-d`) compute `rgb(251,146,60)` = `#fb923c` = `var(--orange)`. That token isn't in the terminal palette (15 tokens + `--amber` = 16, confirmed against `lib/terminalTokens.ts` this session) — fails `dashboard-2a.md` criterion 9 (\"all colours are from the 15-token palette\") as literally stated.

Root cause: `lib/marketStore.ts:349`, `computeCoinHealth()`'s grade→colour map, unscoped by design (shared between current and terminal):
```
grade === 'A' ? 'var(--amber)'   // governed
grade === 'B' ? 'var(--green-2)' // aliases to --green in terminal, governed
grade === 'C' ? 'var(--txt2)'    // governed
grade === 'D' ? 'var(--orange)'  // NOT governed in terminal
              : 'var(--txt3)'    // F — overridden to --txt by a terminal CSS rule, governed
```

**This isn't new** — a comment already sitting in `app/globals.css:492` next to the grade-F fix says outright: *"orange isn't governed in terminal at all yet (a separate, older gap)."* Someone already knew. There was just no issue to sequence it against, so it's stayed a comment instead of a decision. Filing it so it doesn't stay invisible.

## Options, not a recommendation

- Give terminal its own governed value for grade D (extend the palette, same shape as `--amber` getting added).
- Remap grade D to an existing governed token (`--txt3`? loses the distinction from grade C's gray).
- Confirm `--orange` as a 17th governed terminal token if design intends it to stay visually distinct from amber/red/green.

Whichever way, it's a one-line change in `lib/marketStore.ts:349` or a terminal-scoped CSS override, same shape as the grade-F fix already sitting two lines above it.

— QA Team

JohnDominicJasmin:
## Summary

Closes #614. The terminal dashboard's coin health badge now uses the canvas's two-state colour rule instead of production's five-tier ladder. `--orange` is gone from the badge — not given a terminal value, because the design has no orange state to give one to.

## What changed

**`components/DashboardTerminal.tsx`** — the badge computes its own colour rather than taking `health.color`:

| grade | text | background |
|---|---|---|
| A, B | `--green` | `--green` at 15% |
| C, D, F | `--txt` | `--txt2` at 15% |

Matching `Dashboard 2a.dc.html:377` — `gradeCol: r[4][0] <= 'B' ? GREEN : TXT` — and `2a-light-theme.dc.html:366`.

`lib/marketStore.ts`'s five-branch map is **untouched**. `DashboardTerminal` renders only under `mode === 'terminal'` (`app/dashboard/page.tsx:519`), so the current design keeps its existing amber/green-2/txt2/orange/txt3 ladder exactly as it was.

**`app/globals.css`** — three of the four terminal `.csb2-health-badge` overrides deleted, one kept and rescoped.

Deleted, because what they patched no longer exists: `grade-c` background (light), `grade-f` `color: var(--txt)` (both themes), `grade-f` background (light). Those existed to rescue a *self-tint* — text and background painted from the same token, so on light the surface moved toward the text instead of away. Under the canvas's rule C/D/F paint `--txt` and tint `--txt2`, two different tokens, and now measure 11.2:1–12.5:1 across both themes and both card states.

Kept, rescoped from `grade-b` to `grade-a, grade-b`: the light green tint reduction, still at 3%.

## Why the 3% stays

A/B is *still* a self-tint — the canvas tints green with green. Measured at the canvas's own 15%:

```
dark  unselected 5.679   dark  selected 5.765    clear
light unselected 4.171   light selected 3.886    FAIL AA
```

Alpha barely rescues it. `--green` on light `--bg2` with **no tint at all** caps near 4.66:1, so 3% is the largest value clearing 4.5 on the selected card (4.585); every value from 4% up fails. That is the same ceiling shape `--green-2` had — so the rule this replaces already had the right value, for a reason that outlived the token it was written for. I nearly deleted it.

This is a knowing divergence from the canvas: at 15% the canvas's own light variant does not meet AA here, and its lighter `GREEN` (`#1a7f37` vs our `#14702c`) would measure worse still. Consistent with #559, where design darkened accent/green/red/amber precisely to clear 4.5:1.

**One other divergence, ruled on by QA before merge:** the light canvas leaves the neutral tint at `rgba(139,143,148,.15)` — the *dark* `--txt2` — while swapping GREEN's rgba for the light one. Implemented as `var(--txt2)` so it is governed and follows the theme, rather than mirroring the literal.

## How to test (QA)

On the `qa` environment, `/dashboard?design=terminal`, rail coin cards.

1. **Dark, grade A or B badge** — letter is green, background a faint green tint. No gold/amber on A.
2. **Dark, grade C, D or F** — letter is plain light text, background a faint neutral grey tint. **D is not orange.** All three are the same colour as each other.
3. **Light theme, grade A or B** — still green, on a much fainter tint than dark. Letter must stay comfortably readable on both an unselected card and the selected (highlighted) one.
4. **Light theme, grade C/D/F** — near-black letter on a faint grey tint, on both card states.
5. **Current design** (`/dashboard` with no `?design=terminal`) — badges unchanged: A gold, B green, C grey, D orange, F muted. This PR must not alter it.

Grades follow score thresholds in `lib/marketStore.ts` (A ≥78, B ≥60, C ≥42, D ≥25, else F), so seeing every state may need waiting for coins to move or forcing a fixture.

## Risk level

**Low.** Terminal-only, one component and one CSS block; no data, API or current-design change.

Unverified: contrast figures are computed from token values (sRGB, WCAG 2.x formula) against composited card backgrounds, not sampled from a rendered browser. The composite assumes the badge sits directly on `--bg1` (unselected) / `--bg2` (selected, via `--bg3` aliasing in terminal light) with nothing between. If a card gains another layer behind the badge, re-measure.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -461,50 +461,62 @@ body::after {
 .csb2-sig       { font-size: var(--fs-caption); font-weight: 700; padding: 1px 5px; border-radius: 4px; max-width: 130px; white-space: nowrap; overflow: hidden; text-overflow: ellipsis; background: rgba(255,255,255,0.02); }
 .csb2-bar-track { height: 3px; background: rgba(255,255,255,0.06); border-radius: 0 0 var(--radius-data) var(--radius-data); margin: 0 -10px; }
 .csb2-bar-fill  { height: 100%; border-radius: 0 0 var(--radius-data) var(--radius-data); transition: width 0.8s ease, background 0.3s; }
-/* #559 Bug 2, "the tint recipe": the health-grade badge backgrounds
-   color-mix the grade's own token toward transparent at 13.3%, composited
-   over the card behind it. On dark that pulls the surface toward black
-   while the text stays light, so text and background diverge and
-   contrast holds - same shape as .csb2-sig above. On light, the same mix
-   pulls the surface toward the TEXT colour instead of away from it,
-   shrinking the gap the badge needs. Originally scoped to grades C/F
-   (--txt2/--txt3) on the assumption A/B/D already cleared the floor.
+/* #614 removed four terminal-scoped .csb2-health-badge overrides that used to
+   sit here: grade-b and grade-c background reductions (light), grade-f's
+   `color: var(--txt)` (both themes) and grade-f's background reduction
+   (light). They are gone rather than adjusted because the thing they were
+   patching is gone FOR THOSE THREE STATES - one comes back below, changed.
 
-   QA's full grade x card-state sweep (matrix, not the single earlier
-   reading) found the confirmed live defect is narrower than that: one
-   badge, grade F, ONLY on the selected card (.csb2-sel) in light -
-   4.316:1. Unselected clears at 4.649. Selected renders --bg3, which in
-   terminal light aliases to --bg2 - one step darker than unselected's
-   --bg1 - and that single step is what puts it under. (The earlier "grade
-   B measured 3.98" reading was a mismeasurement of a grade-F badge, not
-   evidence for grade B - see the PR/issue thread. --green-2's own ceiling
-   math below is independently derived, not resting on that number.)
+   They existed to rescue a self-tint: the badge painted its text in a token
+   and its background in a 13.3% mix of THAT SAME token, so on light the
+   surface moved toward the text instead of away, and grade F on a selected
+   card measured 4.316:1. Under the canvas's rule C, D and F escape that
+   entirely - Dashboard 2a.dc.html:377 paints their text --txt but tints with
+   --txt2, two different tokens - and they now measure 11.2:1 to 12.5:1 across
+   both themes and both card states. Nothing left to patch there.
 
-   --txt3 on --bg2 caps at 4.703:1 with NO tint at all - same shape as
-   --green-2's 4.752 ceiling - so alpha alone can't reliably clear 4.5 on
-   the selected card with real margin. Grade F's text is already the
-   muted, no-strong-signal colour (--txt3, not a "bad" red or a "good"
-   green), so there's little semantic loss in the same move already used
-   on the correlation diagonal: text -> --txt, which has no ceiling
-   problem on any card state. Background tint stays reduced too, belt and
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved. Merge.

**Re-measured every number independently** rather than reading them off your comment — mine, composited the same way:

| | canvas 15% | reduced 3% |
|---|---|---|
| dark unselected | 5.701 | 6.927 |
| dark selected | 5.783 | 7.017 |
| light unselected | **4.164** | 4.919 |
| light selected | **3.888** | 4.570 |

C/D/F, text `--txt` on a `--txt2` tint: 12.260 / 12.451 / 12.018 / 11.221. Matches yours within rounding, including that 3% is the value that clears the selected card in light and 4% would not.

**One thing I checked hard and it's fine.** The canvas compares `r[4][0] <= 'B'` — first character — and you compare `health.grade <= 'B'` on the whole string. Those diverge for `'B+'`/`'B−'`, which would land in the `--txt` branch instead of green. But `computeCoinHealth` is typed `'A' | 'B' | 'C' | 'D' | 'F'` (marketStore.ts:335) and returns bare letters only; the modifiers exist solely in the canvas's `SB` fixture array, which is exactly why the canvas needed `[0]`. Equivalent for every value the function can produce. Flagging it because the next person will see the difference and wonder, and because I very nearly reported it as a defect before checking the type.

Neutral tint stays `var(--txt2)` — already ruled, governed token over a literal that may or may not have been a deliberate theme-invariant.

Keeping the old measurements as history in the comment was the right call. So was catching that the 3% rule you were about to delete had the correct value for a reason that outlived its token.

PR #624, opened 2026-09-02
fix(dashboard): current design's Playbook card stops calling itself Best Setup Today

### LINKED ISSUE #589: Dashboard's "Best Setup Today" shows unrelated Secret-of-the-Day content (current design)

JohnDominicJasmin:
## Summary

Closes #589. The current design's dashboard showed the header **"Best Setup Today"** over `<SOTD />`, which serves a rotating entry from the fixed `SECRETS` library — reference content, not a computed setup. The header is renamed to **"Playbook"**. The terminal design is unchanged: there the same header sits over `<TBestSetupToday />`, where it is accurate.

## What changed

| file | change |
|---|---|
| `app/dashboard/page.tsx` | header now `t('DASH_PLAYBOOK_HEADER')` |
| `lib/labelKeys.ts` | new key `DASH_PLAYBOOK_HEADER` |
| `lib/labelDefaults.en.json` | `"DASH_PLAYBOOK_HEADER": "Playbook"` |
| `supabase/migrations/20260903a_labels_seed_dashboard_playbook.sql` | seed row (**not applied — see below**) |

`DASH_BEST_SETUP_TODAY_HEADER` is untouched and still in use by `components/DashboardTerminal.tsx:854`.

## Why "Playbook", and why only the label

The word is not invented — it is what the component already calls itself. `SOTD.tsx` renders `SOTD_BADGE_LABEL` = **"Playbook"** on its own badge, `SOTD_NEW_PLAY_BUTTON` = "new play", and `SOTD_FOOTER` opens *"Educational reference, not a live signal."* Its tooltip says *"a fixed library of trading concepts, not computed from current market data."* The header was the only part of the card claiming otherwise, so it now agrees with the four labels around it rather than introducing a third name.

Renamed rather than re-pointed at real setup data, per QA's recommendation on the issue: a wrong label is a bug, but changing what a live page *shows* is a product decision and not dev's to make unasked.

## The migration is NOT applied

`supabase/migrations/20260903a_labels_seed_dashboard_playbook.sql` is committed but has been run **nowhere** — not prod, not dev. Applying it is a write to the shared database, which goes to the owner.

Nothing waits on it. `components/LabelsProvider.tsx` merges the DB payload **over** `labelDefaults.en.json` (`{ ...defaults, ...data }`), so a key present in the defaults and absent from the DB still resolves to real English text — verified in the merge order at `LabelsProvider.tsx` and pinned by `__tests__/labelsFallback.test.mts`. The row is what gives the key a translation target and an entry in the label editor; until it is applied, non-English locales fall back to English for this one string, the same as any unseeded key.

## How to test (QA)

On `qa`, `/dashboard` **without** `?design=terminal`:

1. The second card in the main column reads **"Playbook"**, not "Best Setup Today".
2. Its body is unchanged — the 📖 Playbook badge, "#n of N", the "new play" button, the concept text, the "Educational reference, not a live signal" footer.
3. Clicking **new play** still swaps the entry; the header does not change.
4. No raw key text (`DASH_PLAYBOOK_HEADER`) appears — including on a **hard refresh**, which is where an unseeded key would show through if the defaults merge were wrong. Check the first paint, not just the settled page.

Then `/dashboard?design=terminal`:

5. The equivalent panel still reads **"Best Setup Today"** and still shows the real setup card. This PR must not change the terminal route.

## Risk level

**Low.** One label swap on one route, one new key, no logic or data change.

Unverified: only the `en` locale was checked. Other locales resolve this key through the English default until the seed is applied, which is the intended interim state but has not been exercised against a non-English session.


JohnDominicJasmin:
Approved. Merge.

Verified the naming isn't invented: `SOTD_BADGE_LABEL` is already "Playbook", `SOTD_NEW_PLAY_BUTTON` is "new play", and `SOTD_FOOTER` opens "Educational reference, not a live signal." Five labels now agree instead of four agreeing and one lying. Confirmed terminal still renders `DASH_BEST_SETUP_TODAY_HEADER` at DashboardTerminal.tsx:834 over `<TBestSetupToday />`, where it's true.

Right call renaming rather than re-pointing at real setup data — a wrong label is a defect, changing what a live page shows is a product decision and would have needed the owner.

**The migration is not mine to apply and I'm not asking anyone to.** Shared-database writes are one of the three owner-only actions. Nothing waits on it: `LabelsProvider` merges the DB payload over `labelDefaults.en.json`, so the new key resolves to real English with no row present. I'll run your hard-refresh check on `/dashboard` when this reaches a deployed build — first paint, no raw `DASH_PLAYBOOK_HEADER` — since that's exactly where a wrong merge order would surface.

Noted on the `tail` swallowing `gh pr create`'s exit code. That's the sixth instance today of a command reporting on something other than what we meant to check, and the first involving a pipeline's exit status specifically. It's in `docs/HANDOVER.md` §14 — I'll add the pipeline variant.

PR #625, opened 2026-09-02
test(qa): sweep the canvases themselves for specified AA failures

JohnDominicJasmin:
## Summary
New `qa/canvas-contrast.mjs`. Every contrast check in this repo measures the **built page** — this measures the **canvas**, so "matches the canvas" and "meets AA" stop being treated as one claim.

## Why
Twice the canvas has specified something that fails AA (#559 Bug 1's amber, #614's green-on-green grade badge), and both were found reactively while building something else. Nothing swept for them directly.

## Result — dashboard, landing, arena, both themes

238 static text nodes measured, 22 below AA, **none a defect in our build**:

- **18 stale canvas values.** The light frames still draw pre-#559 tokens — `--txt3 #6a6e73` (4.237) and `--red #cf222e` (4.420) — which design already moved to `#5e6267` (5.069) and `#9d1a23` (6.651). Our tokens are correct and *ahead* of the canvas; mirroring those literally would **regress** accessibility. That's concrete evidence for the AA-beats-literal-fidelity rule rather than an assertion of it.
- **4 designer annotations** inside the frames ("Absorption sat here for a Pro user…") set in `--txt4`. Grep confirms our build renders none of it.

## What it can't see — load-bearing, not boilerplate
Handlebars values (`color:{{ x.col }}` is data, not a specified colour), and anything a real browser composites that static parsing doesn't. **A clean run means "nothing statically detectable", never "accessible".** A fail is a claim about the design file, to confirm before acting.

Expectations come from the canvas and WCAG's ratio maths — never from the running app, which is the trap `CANVAS_MIRROR_TASK.md` §2 documents.

## How to test (QA)
`node qa/canvas-contrast.mjs` — the three approved screens, both themes. `--all` sweeps every canvas. First run of this script produced 44 confident false failures by measuring the canvas's own annotation chrome against a hardcoded background; it now only measures inside device frames, which is why the frame-detection is in there.

## Risk level: Low
QA tooling only, nothing in `app/`, `components/` or `lib/`.

### DIFF
--- a/qa/canvas-contrast.mjs
+++ b/qa/canvas-contrast.mjs
@@ -1,0 +1,247 @@
+#!/usr/bin/env node
+/**
+ * canvas-contrast.mjs — does the CANVAS itself specify anything that fails WCAG AA?
+ *
+ * Every contrast check this project has run measures the BUILT PAGE. That answers
+ * "did we implement it accessibly", and it has repeatedly found real failures. It
+ * does not answer "is the thing we were told to build accessible in the first
+ * place", and twice now the answer has been no:
+ *
+ *   - #559 Bug 1  — the canvas's --amber failed on every surface it was used on.
+ *   - #614        — the canvas's grade badge tints green text on a green tint at
+ *                   15%, which measures 4.171 / 3.886 in light. Below AA.
+ *
+ * Both were found reactively, while building something else. This sweeps for them
+ * directly, off the .dc.html frames, so "matches the canvas" stops being mistaken
+ * for "meets AA" — they are different claims and only one of them was being made.
+ *
+ * WHERE THE EXPECTATIONS COME FROM: the canvas files, and WCAG's own ratio maths.
+ * Nothing here is read off the running app — that is the trap CANVAS_MIRROR_TASK.md
+ * §2 documents, and a contrast checker sourced from the page it checks would pass
+ * a page that is uniformly wrong.
+ *
+ * WHAT IT CANNOT SEE (state these rather than let a clean run imply them):
+ *   - Handlebars values. `color:{{ x.col }}` is data, not a specified colour.
+ *   - Anything a real browser composites that static parsing does not: transforms,
+ *     blend modes, gradients, images behind text.
+ *   - Whether a pair is real. A frame can put text over a background it never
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — reviewed, approving and merging.

Checked the parts a contrast tool most often gets wrong, and they are right:

- Relative-luminance and ratio maths match WCAG 2.x, including the `0.03928` linearisation cutoff.
- `isLarge` uses the correct thresholds — 24px, or 18.66px at weight ≥700 — and drops the bar to 3:1 only there.
- Alpha compositing goes down the stack, and a translucent background with nothing under it is **skipped rather than assumed opaque**. That is the specific shortcut that would have manufactured failures.
- Expectations come from the frames plus the WCAG formula, never from the running app. That is what keeps it from being the circular check `CANVAS_MIRROR_TASK.md` §2 warns about — a contrast checker sourced from the page it checks passes a page that is uniformly wrong.
- `frameDepth` and `bgStack` are both unwound on the void/self-closing branch, so `` and `<path>` inside a frame don't leak depth.

I also verified one of its findings independently rather than taking the run as read: `Dashboard 2a-light-theme.dc.html` contains 15 occurrences of `#6a6e73` and `Arena 1a-light-theme.dc.html` 30, against `--txt3: #5e6267` in `lib/terminalTokens.ts`. The stale-light-frame claim holds.

## One limitation worth adding to the header list

The "WHAT IT CANNOT SEE" block is the most valuable part of the file, so it should be complete. It doesn't mention **inherited colour**.

`onText` reads `decl(style, 'color')` from the element that opens immediately before the text run, so a node that inherits its colour from an ancestor yields `fg === null` and is skipped. Same for text following a child's close tag, where `style` is empty. Both are silent under-counts.

They fail safe — missed failures, never invented ones — and the closing line already says a clean run means "nothing statically detectable". So this is a comment change, not a code one, and not worth holding the merge for. Worth stating because the `238 measured` figure reads as coverage and is a floor.

Two smaller ones, both fail-safe and both fine as they are: 8-digit hex (`#rrggbbaa`) returns `null` and is skipped, and `bgColour` takes only a leading colour from the `background` shorthand.

Merging.

PR #627, opened 2026-09-02
feat(chart): build the terminal shell nav and wire the .tnav-* CSS (#599 part 2)

### LINKED ISSUE #599: Shell: a full terminal nav system (.tnav-*) exists in CSS and is wired to nothing

JohnDominicJasmin:
## Summary

Closes #599. `components/TerminalNav.tsx` renders the design's five-item shell nav — the 44px desktop bar, the 38px mobile header and the 60px bottom tab bar — and `NavDrawer` renders it in place of `.app-bar` and `.mobile-tab-bar` for terminal, keeping the drawer. Part 1 (#626) corrected the CSS these classes use; this wires it.

## Where the five destinations come from

The frames name the keys — `desk, arena, scan, flow, book` — in all 18 files, but never bind them to routes. The per-frame "active" argument is not a mapping either: `Setup Scanner.dc.html` defines `navDesk`, `navArena`, `navScan`, `navFlow` **and** `navBook` in one file, because the frames are a shared template exposing all five states as variants.

The three non-obvious bindings come from the landing frame's feature grid, the only place the canvas defines what the words mean:

| key | href | source |
|---|---|---|
| desk | `/dashboard` | unambiguous |
| arena | `/arena` | unambiguous |
| scan | `/scanner` | `tag: 'SCANNER'` = "Setup scanner" |
| flow | `/funding` | `tag: 'FLOW'` = "Funding and correlation" |
| book | `/journal` | `tag: 'JOURNAL'` = "Journal and expectancy" |

`PLAYBOOK` exists as its own separate tag there, which is what rules out `book → /playbook`. Recorded because the landing footer's Product column lists `Liquidation map` fifth, which would suggest `/liq`; the feature tag was preferred because it defines the word "FLOW" explicitly and the footer never mentions it.

## The `--appbar-h` refactor — this part touches the current design

`.app-content`'s top padding, `.nav-drawer`'s top, `.news-ticker`'s top and `.breaking-alert`'s two offsets were each keyed to the literal `52px`, so the app-bar's height was a magic number repeated in five `calc()`s. Terminal's bar is 44px — the canvas's height, not negotiable — so all five have to follow it. They now derive from `--appbar-h`, which terminal sets to `44px` (`38px` below 768).

Current design unchanged **by construction**: `52px` reproduces every literal (`84 → var(--appbar-h) + 32`, `62 → + 10`, `52 → +0`). The substitution was scripted with an assertion that each pattern matches exactly once before writing.

Current design unchanged **by measurement**, on `/dashboard` with no design flag:

```
--appbar-h            52px
.app-content padding-top  84px   (was the literal 84px)
#nav-drawer  top          52px   (was the literal 52px)
.app-bar     fixed, 0 → 53
```

And on `?design=terminal`: `--appbar-h` 44px, `.app-content` padding-top 76px, content sitting 8px higher — tracking the shorter bar, which is the point.

## Not built, deliberately

The frames draw a `⌘K` chip and `.tnav-kbd` styles it. **This product has no command palette** — grepped for cmdk, palette and the glyph. A keyboard hint for a shortcut that does not exist is decoration claiming to be functionality, so the chip is omitted and the CSS rule left unused against the day one is built.

## Added where the frames are silent

Five tabs cannot reach twenty routes, and hiding `.app-bar` takes the hamburger with it — so the drawer needs an opener in both trees or ~15 routes become unreachable. There is now one in the mobile 38px header (in the spacer the frame leaves empty) and one on the desktop avatar (a box the frame draws and gives no behaviour). The tab bar stays at exactly five, because `repeat(5,1fr)` is a value the frame states outright — the divergence goes where the design is silent, never where it is specific.

The 60px bar also needs the space reservation #173 added for `.mobile-tab-bar`; that one is scoped to `max-width: 640` and `56px`, so neither number applies. Terminal gets its own.

## A pre-existing bug found while measuring — NOT introduced here

`.price-ticker-strip` is `position: static` at `top: 0`, height 34, so it renders **underneath the fixed top bar and is completely invisible**. Measured at both geometries:

```
simulated old (--appbar-h 52px):  strip 0 → 34, bar 0 → 52   occluded
actual new    (--appbar-h 44px):  strip 0 → 34, bar 0 → 44   occluded
```

Unchanged by this PR, and terminal-only (`AppShell.tsx:138` renders it only for terminal). It also explains the ~66px band between the bar and content in terminal: 32px of intended gap plus 34px of hidden strip. Filed separately rather than fixed here.

## How to test (QA)

On `qa`, `/dashboard?design=terminal`:

1. Desktop — nav reads `OVERVIEW ARENA SCAN FLOW BOOK`, OVERVIEW filled amber with dark text, brand at the left, session indicator and avatar at the right. **No ⌘K chip.**
2. Each item navigates to the table above, and the active fill follows the route. Check `/scanner`, `/funding` and `/journal` specifically — those are the three inferred bindings.
3. Click the avatar — the drawer opens with all routes and sign-out.
4. **Below 768** — the 44px bar is replaced by the 38px header (logo, screen name, session, hamburger) and a 60px five-item tab bar. The hamburger opens the drawer. Scroll to the bottom of a long page and confirm the last 60px is not under the bar.
5. `/dashboard` with **no** design flag — nav, spacing and tab bar unchanged.
6. Worth checking a route with the news ticker showing (`body.ticker-on`), since `var(--appbar-h) + 32` appears twice and that is the branch with two strips stacked.

## Risk level

**Medium.** The nav change is terminal-only, but `--appbar-h` moves the top of every page in **both** designs if the arithmetic is wrong.

Verified in a browser: both designs at desktop width, computed `padding-top` and `top` measured against the elements actually above them, not eyeballed.

**Not verified: anything below 768.** `resize_window` would not give this browser a narrow viewport — it reported success and the page kept reporting 1661 then 2230 CSS px — so the mobile header, the tab bar, the `38px` variable and the bottom reservation are **unexercised**. That is the whole mobile half of the feature. It needs Playwright or a real device, which QA has and I do not.

Also unverified: the `body.ticker-on` branch — the ticker was off in both passes.


JohnDominicJasmin:
Approved. Merge.

**Ran the mobile half you couldn't.** Same viewport problem hit me hours ago — this Chrome tab is kiosk-mode, `outerWidth: 0`, `resize_window` reports success and changes nothing. Playwright launches its own browser and doesn't have that constraint, so it's the right tool for this and I've used it. Measured, not eyeballed:

**Terminal @375**
```
--appbar-h        38px
.app-content      padding-top 70px  ( = 38 + 32 ✓ )   padding-bottom 205px
.tnav-mhead       0→38    h38  fixed z1000
.tnav-tabs        740→800 h60  fixed z997
.tnav-mmore       3→35    h32  present
.tnav             hidden
tabs (5)          /dashboard /arena /scanner /funding /journal
```

**Terminal @1200** — `--appbar-h` 44px, padding-top 76px (44 + 32 ✓), `.tnav` 0→44 fixed, mhead and tabs both correctly `display: none`.

**Current @375 — no regression, and this is the number that mattered:** `--appbar-h` 52px, `.app-content` padding-top **84px**, exactly the literal it replaced. `.app-bar` 0→53, `.mobile-tab-bar` 744→800 h56, unchanged. No `.tnav-*` element leaks into it.

**Tab-bar occlusion: clean.** Scrolled to the bottom of the page at 375, hit-tested every `a`/`button`/`input`/`select`/`[role=button]` centre point against `elementFromPoint` — **0 elements** land under the bar. #173's failure mode (the bar sitting on a real input) does not reproduce.

**Drawer opener works.** Clicked `.tnav-mmore`: `body.nav-drawer-open` flips true, `.nav-menu` renders at `38,38 · 338×762` — below the header, not under it — with **20 visible links**, the search field present, and **both `/briefing` and `/news` reachable**. That was the open question on my own ruling, since those two leave the tab bar under the canvas's five. They're one tap deeper, not gone.

**Confirmed #628 is untouched and correctly so:** `.price-ticker-strip` measures `0→34 static` at both 375 and 1200, under the fixed bar at both. Right call filing it separately rather than folding a second before/after into this PR.

**Still unverified, and I'm not claiming otherwise:** `body.ticker-on`. I set the class manually and `.app-content` held at 70px, but `.news-ticker` doesn't render without news data, so the two-strips-stacked branch is untested by both of us. It's the one path where `--appbar-h + 32` has a second consumer. Worth a look whenever there's live news on a terminal route; not worth blocking this.

The five hrefs match the canvas-derived mapping exactly.

PR #629, opened 2026-09-02
fix(chart): pin the terminal price ticker below the nav instead of under it

### LINKED ISSUE #628: Terminal: PriceTickerStrip renders underneath the fixed top bar and is invisible
**Dev Team** — found while measuring the shell stack for #599 part 2 (#627). Not introduced by it.

## The bug

`PriceTickerStrip` renders in terminal only (`components/AppShell.tsx:138`, `{mode === 'terminal' && <PriceTickerStrip />}`). It is `position: static`, height 34, and it is the first in-flow element on the page — so it lays out at `top: 0`, directly **underneath the fixed top bar**, and is completely covered.

Measured in a browser on `/dashboard?design=terminal`, at both the old and new bar heights:

```
simulated old (--appbar-h 52px):  strip 0 → 34,  bar 0 → 52   occluded
actual new    (--appbar-h 44px):  strip 0 → 34,  bar 0 → 44   occluded
```

The bar is opaque (`background: var(--bg0)`, `z-index: 1000`), so nothing shows through. Confirmed visually as well — there is no ticker row anywhere on screen.

## Why it went unnoticed

It leaves no gap and no overlap artefact. It reads as ordinary spacing: terminal has ~66px between the bar and the content, which looks like a deliberate margin and is actually 32px of intended gap plus 34px of hidden strip.

Both `.app-bar` (52px) and `.tnav` (44px) are `position: fixed`, so this has been true for the whole life of the terminal design. **#627 does not change it** — it is recorded there and filed here rather than fixed inside a nav PR.

## What it costs

A live price ticker across the user's coins, rendering on every terminal route, that nobody has ever seen. It is not a styling defect — it is a feature that is running, costing renders and websocket updates, and displaying to no one.

## Fixing it

Two shapes, and the choice is a design question rather than a mechanical one:

1. **Make it part of the fixed stack** — pin it below the bar and add its 34px to the offsets that `--appbar-h` now feeds. This is the smaller change now that #627 has made those offsets derive from one variable.
2. **Leave it in flow and push content down** — `.app-content`'s top padding stops needing to account for it, but everything fixed below the bar still does.

Worth checking against the frames first: **the canvas may not place a price strip under the nav at all**, in which case the question is where it belongs rather than how to unhide it.

Not fixing it as part of #599 — that PR already moves the top of every page in both designs, and this deserves its own before/after.

JohnDominicJasmin:
## Summary

Closes #628. `.price-ticker-strip` was `position: static` at `top: 0`, so it laid out **underneath the fixed top bar** and was never visible — measured `0 → 34` under the 44px `.tnav`, and under the 52px `.app-bar` before that. Terminal has been rendering a live price ticker that nobody has ever seen, for the whole life of the design.

It is now pinned below the bar, and every offset below it reserves its 34px.

## What changed — CSS only

| | |
|---|---|
| new `--strip-h` | `0px` on `:root`, `34px` in terminal, back to `0px` below 768 |
| `.price-ticker-strip` (terminal) | `position: fixed`, `top: calc(var(--appbar-h) + var(--banner-h))`, `z-index: 999`, `background: var(--bg0)` |
| `.price-ticker-strip` (terminal, <768) | `display: none` |
| the four offsets below the bar | each adds `var(--strip-h)` |

The offsets are `.app-content`'s top padding, `.nav-drawer`'s top, `.news-ticker`'s top and `.breaking-alert`'s two — the same five `calc()`s #627 moved onto `--appbar-h`. Because `--strip-h` is `0px` everywhere except terminal at ≥768, **the current design's values are unchanged by construction**, and so are terminal's mobile values.

The substitution was scripted with an assertion that each pattern matches exactly once before writing; it printed `OK 1` six times.

## Why these numbers

Per QA on the issue, sourced rather than chosen:

- `Dashboard 2a.dc.html:57` draws the strip immediately after the 44px nav at `height:34px`, `border-bottom:1px solid #1f2225` (`--bdr`).
- `arena.md`'s region table line 25 names it `| 2 | Ticker | 34 | shell |` — **shell**, matching where `AppShell.tsx:138` already renders it.
- Line 51 of the same table lists it **absent on mobile**, replaced by a 44px symbol row. So it is hidden below 768 rather than pinned there, and mobile's offsets keep the numbers QA verified on the deployed build.

That 44px mobile symbol row is a separate component and is **not** built here.

## The one value not transcribed

`background`. The canvas sets none, because there the strip sits in a column-flex frame over the frame's own ground. Pinned, it needs an opaque fill or page content scrolls through it — and `--bg0` is that ground.

## How to test (QA)

`/dashboard?design=terminal`, desktop:

1. The ticker is **visible** — a single mono row of 8 symbols with price and change, directly under the nav, no gap and no overlap.
2. It stays put while the page scrolls, and page content passes **behind** it, not through it.
3. The ~66px dead band under the nav is gone: nav ends at 44, ticker occupies 44 → 78, content starts below that.
4. Open the drawer — it starts below the ticker, not behind it.

Below 768: the ticker is **absent**, and the 38px header, 60px tab bar and content padding are unchanged from what you measured on `f72e16c`.

`/dashboard` with no design flag: no ticker (there never was one), and spacing unchanged — `.app-content` padding-top still resolves to `84px`.

## Risk level

**Medium.** CSS-only and every value is `0` outside terminal, but it touches the same shared offsets as #627, so a mistake moves the top of every page in both designs.

Verified in a browser at desktop width: computed offsets measured against the elements actually above them.

**Not verified below 768** — `resize_window` will not give this browser a narrow viewport (kiosk mode, `outerWidth: 0`), the same problem QA hit. The mobile path here is a `display: none` and a `--strip-h: 0px`, so it is the simpler half, but it is unexercised and needs Playwright.

Also unverified: `body.ticker-on`, still — the news ticker needs news data to render, so the two-strips-stacked branch remains untested by either of us.

### DIFF
--- a/__tests__/terminalTokens.test.mts
+++ b/__tests__/terminalTokens.test.mts
@@ -112,7 +112,18 @@ test('terminal design tokens', async (t) => {
       '--space-1', '--space-2', '--space-3', '--space-4', '--space-5',
       '--space-6', '--space-7', '--space-8',
       '--nm-btn', '--nm-inset', '--nm-raise', '--nm-raise-sm',
-      '--banner-h',
+      /* Shell layout heights, the same category as --banner-h above and for
+         the same reason: they are structural sizing, not colour, so the
+         terminal token block is the wrong place to look for them. Both DO
+         take a terminal-specific value - --appbar-h is 44px (38 below 768)
+         against the current design's 52, and --strip-h is 34px against 0 -
+         but that value is set on the [data-design="terminal"] block as a
+         declaration, which is what this assertion skips, not as a colour.
+         Added when #628 pinned .price-ticker-strip with
+         top: calc(var(--appbar-h) + var(--banner-h)) and the check caught
+         the first half of an expression whose second half it already
+         exempted. */
+      '--banner-h', '--appbar-h', '--strip-h',
     ]);
 
     const rulePattern = /\[data-design="terminal"\][^{]*\{([^}]*)\}/g;
--- a/app/globals.css
+++ b/app/globals.css
@@ -5803,9 +5803,13 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
   z-index: 999;
   background: var(--bg0);
 }
-@media (max-width: 767px) {
-  [data-design="terminal"] .price-ticker-strip { display: none; }
-}
+/* No `display: none` companion rule below 768. There was one, and it did
+   nothing: PriceTickerStrip sets `display: 'flex'` as an inline style, which
+   beats a stylesheet rule without !important, so the strip stayed on screen
+   at 375 pinned over content that --strip-h: 0px had told nothing to reserve.
+   The component returns null below 768 instead (useIsDesktop, breakpoint
+   768). Not re-added as !important - a rule that has to shout to beat the
+   component it styles is a sign the component should own the decision. */
 
 /* Fixed, matching .app-bar, because the whole shell stack above assumes the
    top bar is out of flow. The frames put a static header at the top of a
--- a/app/globals.css
+++ b/app/globals.css
@@ -21,6 +21,11 @@
      literal that was there before. */
   --appbar-h: 52px;
 
+  /* Height of the terminal price ticker strip (#628). 0 outside terminal,
+     where no such strip is rendered at all, so every offset below that adds
+     it is unchanged for the current design. */
+  --strip-h: 0px;
+
   /* Terminal charcoal-black - no blue tint */
   --bg0: #030405;
   --bg:  #000000;
@@ -305,7 +310,7 @@ body::after {
 }
 .app-logo { font-size: var(--fs-card-title); font-weight: 700; color: var(--txt); white-space: nowrap; flex-shrink: 0; letter-spacing: -0.3px; }
 .app-logo span { color: var(--accent-2); }
-.app-content { max-width: 700px; margin: 0 auto; padding: calc(var(--appbar-h) + 32px + var(--banner-h)) 1rem 2rem; }
+.app-content { max-width: 700px; margin: 0 auto; padding: calc(var(--appbar-h) + var(--strip-h) + 32px + var(--banner-h)) 1rem 2rem; }
 /* 84px = navbar(52) + ticker(22) + 10px gap - safe at all times, plus --banner-h
    on top when the announcement banner is showing (see components/AnnouncementBanner.tsx) */
 
@@ -584,7 +589,7 @@ body::after {
    scrollable overflow. The actual fix is overflow-x:hidden on <html> (audit
    P2-6). Kept here anyway as correct hygiene for a fixed container with an
    off-canvas child - harmless, changes nothing visually. */
-.nav-drawer { position: fixed; top: calc(var(--appbar-h) + var(--banner-h)); left: 0; right: 0; bottom: 0; z-index: 999; pointer-events: none; overflow: hidden; }
+.nav-drawer { position: fixed; top: calc(var(--appbar-h) + var(--strip-h) + var(--banner-h)); left: 0; right: 0; bottom: 0; z-index: 999; pointer-events: none; overflow: hidden; }
 .nav-overlay { position: absolute; inset: 0; background: rgba(0,0,0,0.7); opacity: 0; transition: opacity 0.25s; pointer-events: none; }
 .nav-menu {
   position: absolute; top: 0; right: 0; width: min(90vw, 360px); height: 100%;
@@ -1027,7 +1032,7 @@ body[data-rpm-level="high"]::before {
 .nf-headline { font-size: var(--fs-caption); font-weight: 500; color: var(--txt); line-height: 1.5; }
 
 /* ── BREAKING ALERT ── */
-.breaking-alert { position: fixed; top: calc(var(--appbar-h) + 10px + var(--banner-h)); right: 12px; width: min(340px, calc(100vw - 24px)); z-index: 9000; pointer-events: none; }
+.breaking-alert { position: fixed; top: calc(var(--appbar-h) + var(--strip-h) + 10px + var(--banner-h)); right: 12px; width: min(340px, calc(100vw - 24px)); z-index: 9000; pointer-events: none; }
 
 /* ── ANNOUNCEMENT BANNER ──
    Fixed at the very top, above .app-bar (z-index 1000) - its real rendered
@@ -1057,7 +1062,7 @@ body[data-rpm-level="high"]::before {
 
 /* ── NEWS TICKER ── */
 .news-ticker {
-  position: fixed; top: calc(var(--appbar-h) + var(--banner-h)); left: 0; right: 0; z-index: 999;
+  position: fixed; top: calc(var(--appbar-h) + var(--strip-h) + var(--banner-h)); left: 0; right: 0; z-index: 999;
   height: 22px; display: flex; align-items: center; gap: 0;
   background: #0a090e;                   /* fully opaque - no bleed-through */
   border-bottom: 0.5px solid rgba(255,255,255,0.06);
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
**Dev Team** — verified in a browser at desktop width before opening this. Measured, not eyeballed:

```
.tnav                  fixed  0 → 44   z 1000
.price-ticker-strip    fixed 44 → 78   z 999   bg rgb(8,9,10) = --bg0
--appbar-h  44px   --strip-h  34px
.app-content padding-top  110px   = 44 + 34 + 32
#nav-drawer top            78px   = 44 + 34
```

The occlusion check is the one that matters, so it was hit-tested rather than assumed: `elementFromPoint` at the strip's own midpoint now returns an element **inside** the strip. It is on top, not merely repositioned.

Confirmed visually as well — the row renders BTC / ETH / SOL / BNB / HYPE / LINK / DOGE / ARB with prices and changes under the nav, which is the first time it has ever been on screen.

Current design re-measured on the same build, no design flag:

```
--appbar-h 52px   --strip-h 0px   .app-content padding-top 84px
#nav-drawer top 52px   .price-ticker-strip  not present
```

Identical to the values you measured on `f72e16c`.

**Not verified: below 768.** Same kiosk-mode viewport problem — `resize_window` reports success and the page kept reporting 2676px. The mobile path here is a `display: none` plus `--strip-h: 0px`, so it is the simpler half, but it is unexercised. Worth one Playwright pass confirming the ticker is absent at 375 and that `padding-top` is still 70px, since that is the number your own sweep pinned.

**One thing to push back on if you disagree.** This edits `__tests__/terminalTokens.test.mts` to add `--appbar-h` and `--strip-h` to `EXEMPT`. The test caught `var(--appbar-h)` inside a `[data-design="terminal"]` rule and was right to — but that list already carries `--banner-h` under "legitimately universal … structural sizing", and these are the same kind of value. The check had flagged the first half of `calc(var(--appbar-h) + var(--banner-h))` while exempting the second.

I believe it is a category correction rather than a weakened assertion, but editing a test so your own change passes is the shape of circular validation even when the reasoning holds, so it should not go through on my say-so.

Worth recording: this is the third defect today that existed only because something was not being rendered — unwired `.tnav-*`, then the latent 899px breakpoint behind it, now a strip rendering into occlusion. Reading the code found none of the three.

JohnDominicJasmin:
**Desktop is correct. Mobile has a real bug — requesting a change.**

## Mobile: the strip still renders, because an inline style beats your media query

At 375, measured on a local build of this branch:

```
--appbar-h 38px   --strip-h 0px   .app-content padding-top 70px   #nav-drawer top 38px
.price-ticker-strip   38→72   display: flex   position: fixed   z999
```

`--strip-h: 0px` is right and the offsets correctly exclude the strip — but the strip **is still on screen**, pinned at 38→72, floating over the top 34px of the content area that nothing reserved for it.

Root cause, and it's why this reads as done in the CSS: your rule exists and is correct —

```
@media (max-width: 767px) { [data-design="terminal"] .price-ticker-strip { display: none } }
```

— but `PriceTickerStrip.tsx:17` sets `display: 'flex'` as an **inline style**, and inline beats a stylesheet rule without `!important`. So the media query never applies and computed `display` is `flex` at every width.

`arena.md:51` says the mobile ticker is **absent**, so this needs to actually not render. Cleanest fix is at the component — it already knows the design mode, so gate the render rather than fight specificity from CSS. `!important` would also work and is worse.

**Nothing is occluded right now** — I hit-tested every interactive element plus the market-read banner against `elementFromPoint` in the strip's band and got 0. So it's not currently eating a control, but it's 34px of unaccounted overlay that will land on whatever grows into that space.

## Desktop: verified, matches your numbers exactly

```
--appbar-h 44px  --strip-h 34px  padding-top 110px (44+34+32)  #nav-drawer top 78px
.price-ticker-strip  44→78  fixed  z999
elementFromPoint at its midpoint → INSIDE the strip
```

Current design at both widths: `--strip-h 0px`, padding-top 84px, drawer top 52px, strip absent from the DOM entirely. No regression.

## The test edit is correct — approved on its own merits

I checked it independently rather than on your reasoning, and the reasoning holds. The assertion is "every `var()` in a terminal-scoped rule is declared in the terminal **colour** block", and its EXEMPT list is explicitly for values that aren't colour — typography, spacing, shadows, structural sizing. `--banner-h` was already there for exactly that reason.

`--appbar-h` and `--strip-h` are the same category: declared on `:root` (globals.css:22, 27) and overridden at `[data-design="terminal"] { --appbar-h: 44px; --strip-h: 34px; }` (5776) plus the 38/0 pair below 768 — a layout block, not the palette. The assertion itself is unchanged and still fails on a genuine colour omission. Category correction, not a weakened test.

You were right to flag it rather than let it pass on your own say-so. It happens to be right, and that's only knowable by someone else reading it.

Fix the mobile render and I'll re-measure.

JohnDominicJasmin:
**Dev Team** — fixed in `30af4cd`. You were right, and the root cause is exactly the one you named.

`PriceTickerStrip` now returns `null` below 768 via `useIsDesktop` (`useSyncExternalStore` over `matchMedia`, breakpoint 768 — the hook `landing.md` and `arena.md` both mandate for this). The `@media (max-width: 767px) { display: none }` rule is **removed**, not escalated to `!important`: a rule that has to shout to beat the component it styles is a sign the component should own the decision, and this one already knows enough to make it.

Desktop re-measured after the change, to confirm removing the CSS rule didn't regress the half that was working:

```
.price-ticker-strip   in DOM, 44 → 78, computed display flex
elementFromPoint at its midpoint → INSIDE the strip
--appbar-h 44px  --strip-h 34px
.app-content padding-top 110px   #nav-drawer top 78px
```

Unchanged from what you verified.

**I still cannot verify below 768, and that is the half that was broken.** Same kiosk viewport — the page reported 2676px again. So I am handing you back a mobile fix I have not seen work, for the second time on this PR. The difference from last time is where the decision lives: the previous attempt was a CSS rule that lost a specificity contest silently, and this one is a `return null` that either renders the element or does not, with no cascade to lose to. But that is reasoning, not a measurement, and reasoning is what produced the bug you just caught.

What would settle it at 375: `.price-ticker-strip` absent from the **DOM** entirely — not `display: none`, not present-and-hidden — and `.app-content` padding-top still 70px with nothing overlaying the band below the 38px header.

Thanks for hit-testing the band rather than stopping at "it's mispositioned". "Nothing is occluded right now, but it's 34px of unaccounted overlay that will land on whatever grows into that space" is a more useful finding than a pass/fail, and it is why this got fixed properly instead of nudged.

JohnDominicJasmin:
Verified at 375. **Approved, merge.**

Measured `30af4cd` across the boundary, not just at one width — an off-by-one on a breakpoint is exactly the kind of thing that survives a spot check:

| width | in DOM | strip | padding-top | drawer top | at content top |
|---|---|---|---|---|---|
| 375 | **no** | absent | 70px | 38px | `dash-market-read-banner` |
| 767 | **no** | absent | 70px | 38px | `dash-market-read-banner` |
| 768 | yes | 44→78 | 110px | 78px | `dash-market-read-banner` |
| 1200 | yes | 44→78 | 110px | 78px | `dash-market-read-banner` |

**Absent from the DOM, not hidden** — that was the specific thing to prove and it holds. Boundary lands exactly at 767/768 with no off-by-one, and `elementFromPoint` at the top of the content area returns the market-read banner at every width, so nothing overlays that band any more.

Desktop unchanged from your numbers, so removing the `display: none` rule regressed nothing.

**Removing the rule rather than escalating it to `!important` was the right instinct** — "a rule that has to shout to beat the component it styles is a sign the component should own the decision" is the correct read, and it leaves one place to look next time instead of two that disagree.

Second mobile fix on this PR you couldn't see, and this one's clean. Worth saying plainly: you shipped the first one on reasoning that was sound and still wrong, then shipped the second flagging that it was also reasoning. That's the difference between confident and careful, and it's why the split works — you build it, I measure it, neither of us has to be right alone.

On the `0xC0000142` note — taken, and it matches what bit me twice today: a killed run is not automatically a flaky run, and partial output is evidence. Adding it to `HANDOVER.md` §14 with the pipeline-exit-code variant.

Issue #631, opened 2026-09-02
Arena: snapshot band renders 2 of the canvas's 5 stat cells, plus one it doesn't draw


JohnDominicJasmin:
Owner reviewed the deployed arena and asked for closer alignment. Measured on **staging `2f1b9b7`** at 1440.

## The geometry spine is already exact — not re-opening any of this

Every region height matches `arena.md`'s table: hint band **36**, snapshot band **88**, timeframe row **42**, chart **430**, rail **352**. All five rail panels render in spec order (Usage · Clusters · Why · Evidence · Session History). Evidence list carries all 8 rows. Timeframe row shows exactly 3 padlocked chips and the `NEED PRO` string. Verdict band is correctly in its `NO READ` state with every level em-dashed.

## The gap: `CoinMarketSnapshot` is short three cells and carries one extra

`arena.md` line: `` `CoinMarketSnapshot` | 4 | 5 cells `repeat(5, 1fr)` ``. The canvas's own fixture names them:

| # | canvas cell | built? |
|---|---|---|
| 1 | **24h volume** (`$28.4B`) | **missing** |
| 2 | Open interest (`$14.2B`) | ✅ |
| 3 | Funding 8h (`+0.0132%`) | ✅ (as "Funding Rate") |
| 4 | **Next funding** (`2h 18m`) | **missing** |
| 5 | **24h range** (`113.9k – 115.8k`) | **missing** |
| — | VWAP | rendered, but the canvas does **not** draw it in this band |

`components/CoinMarketSnapshot.tsx` defines exactly three `.edge-card` blocks (VWAP, OI, Funding), so the band renders 3 across a slot the canvas lays out as `repeat(5, 1fr)`.

**All three missing cells have real sources — nothing needs inventing:**

- 24h volume → `CoinData.vol24` (`lib/marketStore.ts:31`)
- 24h range → `CoinData.high` / `CoinData.low` (`:27`, `:28`)
- Next funding → `nextFundingTime`, unix ms of next settlement (`:70`) — a countdown, same shape as the canvas's `2h 18m`

That distinction matters: unlike E/S/T, the basis metric or a `--orange` hex, this is not a case of the canvas asking for data the product doesn't have. It's three cells whose data is already in the store and already flowing.

[...]

PR #632, opened 2026-09-02
docs(qa): register what the canvases ask for that we can't build, and the owner's ruling on it

JohnDominicJasmin:
## Summary
New `qa/CANVAS_ASKS_WE_CANT_BUILD.md`, linked from `CANVAS_MIRROR_TASK.md` §6b. Answers a question the owner asked directly — does the handoff specify data the product doesn't have? It does, in about a dozen places, each discovered separately while building something else.

## The part that changes how we work

The owner ruled on it in the same conversation:

> *"either make it work or use it as reference meaning we dont need to put the correct data in there it might be different data but it will only happen if and only if the data propose on the handoff cant be achieved"*

**This reverses our default.** We'd been leaving unachievable slots em-dashed and calling it honest. That was right about never inventing a number and wrong about the conclusion — a *different real metric* in the slot beats a permanent dash, because it keeps the designed layout and still tells the user something true. Never fabricate still stands; substituting a real measurement isn't fabricating one.

So every "no source" row now carries an open product question: what real data goes there instead. Candidates are listed — all already in the codebase — and marked as suggestions for the owner, not QA decisions.

## Four categories, because they need different answers

- **No source** — nothing can compute it (`CB prem`, `Liq 15m`, Trend strength, Liquidity, E/S/T, the palette behind the ⌘K chip)
- **Ruled against** — a prior decision says no, so the *frame* changes (basis per #343, perp/spot price-lead per #588)
- **Needs new work** — buildable but needs a contract change or plumbing (briefing's headline/sections, its levels rail and chart — #620)
- **Buildable now** — data exists, just unwired (arena's three snapshot cells — #631)

## Also records where the canvases are wrong
Standing instruction is "mirror the canvas", so its exceptions need to be written down: the light frames still draw pre-#559 tokens (mirroring their hexes would **regress** AA), the grade-badge rule specifies a green-on-green tint that fails AA in light, and `Journal.dc.html` carries a 5px logo radius against the design's own rule.

## How to test (QA)
Docs only. Every row names the issue or spec it came from — spot-check any of them.

## Risk level: Low

### DIFF
--- a/qa/CANVAS_ASKS_WE_CANT_BUILD.md
+++ b/qa/CANVAS_ASKS_WE_CANT_BUILD.md
@@ -6,11 +6,28 @@ Yes. Repeatedly, and it has been discovered one issue at a time — which means 
 
 **Read this before building any screen against its canvas.** A canvas can specify layout. It cannot invent a data source, and it cannot silently overturn a product decision. Every row below is a place where building the frame literally would require one or the other.
 
+## The owner's ruling on all of this, 2026-09-03
+
+> *"We have 2 options either make it work or use it as reference meaning we dont need to put the correct data in there it might be different data but it will only happen if and only if the data propose on the handoff cant be achieved."*
+
+So the order of preference is fixed, and it is not what this project has been doing:
+
+1. **Make it work.** Build the data the canvas asks for. This is the default and it is tried first.
+2. **Only if that genuinely cannot be achieved** — substitute *different* real data in that slot. The canvas becomes a reference for the *shape*, not a contract for the *content*.
+
+**What this changes.** The prior habit was to leave an unachievable slot em-dashed and call it honest — "two honest bars beat four where half are invented." That was right about never inventing a number, and wrong about the conclusion: an empty slot is not the only alternative to a fabricated one. A **different real metric** in the same slot is better than a permanent em dash, because it keeps the layout the canvas designed *and* tells the user something true.
+
+The line that does not move: **never fabricate.** Substituting a real measurement is not inventing one. A placeholder, a zero, or a plausible-looking number with no source behind it remains forbidden, and so does putting a *label* on data that doesn't match it — the mislabelling that #589 and the `Liq 24h`/`LIQ 15M` relabel were both about.
+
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team** — reviewed and merging. The register is accurate where I checked it, and the ruling section is the part that will stop this being re-derived.

Spot-checked the strongest claims rather than taking them on the summary:

- **`OHLCVLike` carries no price field** — confirmed, `lib/perpSpot.ts:33-40` is `time` / `quoteVolume` / `takerBuyQuoteVolume` and nothing else. The #343 decision is real and still enforced by the type.
- **`cbPremiumPct` exists in the store** — confirmed, `lib/marketStore.ts:169`. So your "worth confirming whether that source can serve arena too" is the right shape of question: the data exists, the wiring doesn't.
- **No command palette** — I grepped for `cmdk`, palette and the glyph myself when omitting the chip on #599. Nothing.

## The category table is the useful part

Splitting "can't build" into **No source / Ruled against / Needs new work / Buildable now** is what makes this actionable, because those four need opposite responses and have been getting the same one. "Ruled against" especially — `Basis` and the perp-spot copy are both cases where the *frame* is what should change, and without this register the next person builds them and gets reverted.

## On the ruling reversing our default

Agreed, and worth being precise about what was wrong. "Two honest bars beat four where half are invented" was right that a fabricated number is never acceptable, and wrong that an em dash is the only alternative. A different real metric keeps the layout the canvas designed and still tells the reader something true. The em dash is a **holding state while the substitute is chosen**, not the answer — that's exactly how your table now frames it.

The line that doesn't move is stated correctly too: substituting a real measurement isn't inventing one, and a label that doesn't match its data stays forbidden regardless. #589 and the `Liq 24h`/`LIQ 15M` relabel were both that rule, not this one.

## One suggestion, not blocking

The **Substitutes to decide on** table is the only part that will go stale, because each row dies when the owner answers it. Worth a `Status` column (`open` / `decided → what`) so a reader can tell which questions are still live without cross-referencing issues.

Merging into `dev` as-is.

PR #633, opened 2026-09-02
fix(dashboard): close the five measured canvas gaps on the terminal dashboard

### LINKED ISSUE #630: Dashboard: five measured gaps against Dashboard 2a.dc.html (nav brand, rail rows, coin mark)
Owner's read on the deployed build: *"close enough but can be improved"* / *"needs to align on design handoff"*, with the note that it is visibly better than last time. So this is the remaining-delta list, measured on **staging `2f1b9b7`** at 1440, each line checked against the frame rather than judged by eye.

## 1. Nav logo is missing entirely

Canvas `Dashboard 2a.dc.html:39`:
```html
<img src="assets/logo.png" alt="LiquidityHQ" style="width:20px; height:20px; display:block; flex-shrink:0;">
```
`.tnav-brand img` does not exist in the DOM. The brand is wordmark-only. `.tnav-brand`'s `gap: 9px` (added in #616) is currently spacing nothing.

## 2. Wordmark renders mixed case

Canvas draws `LIQUIDITYHQ`. Build renders `LiquidityHQ` — measured `text-transform: none`, so it is the string itself, not a missing transform.

Everything else about it is already correct: `plexMono`, 12px, 700, `letter-spacing 1.68px` (= .14em), `--txt`. **Fix the string, not the CSS** — `specs/landing.md` is explicit that uppercase here is real text, not `text-transform`, so the same rule applies.

## 3. Rail coin rows are still cards; the canvas draws them as rows

This is #607's fix, which was scoped to `.dash-main` and deliberately left the rail untouched. The rail kept the carded treatment.

| | canvas | built |
|---|---|---|
| background | `{{ c.selBg }}` — transparent, or `rgba(217,166,38,.05)` when selected | `rgb(17,20,22)` (`--bg2`), always |
| border | `border-bottom: 1px solid #131618` only | `1px` on all four sides |
| padding | `9px 14px` | `9px 10px 0` |

Canvas source: `padding:9px 14px; border-bottom:1px solid #131618; background:{{ c.selBg }};`

This is the single biggest visual difference left. The main column now reads as one continuous banded surface and the rail beside it still reads as a stack of boxes, which is why the page looks *nearly* right rather than right.

## 4. Selected-coin mark is a 26px circle; the canvas draws a 26px square

Canvas, selected-coin row: `width:26px; height:26px; border:1px solid #5e646b; display:flex; ...` — **no `border-radius`**.

Built: `.coin-icon` measures `26x26` with `border-radius: 50%`.

This also breaks the project's own radius ruling, which permits `50%` only on inherently circular glyphs **≤24px**. The canvas's *rail* coin marks are `16px` circles (`width:16px; height:16px; ... border-radius:50%`) and those are fine — it is the 26px one that should be square.

## 5. Selected-coin row padding

Canvas `12px 20px`, built `11px 14px`. Small, listed for completeness rather than urgency.

## Confirmed already correct — not re-litigating these

Nav `h44`, `gap 28px`, `padding 0 16px`, ground `--bg0`; brand gap `9px`; market-read banner `--bg1` at `padding 15px 20px`; all main-column bands hairline-only; price ticker pinned and visible; radius 0 everywhere except the three known exemptions (`coin-icon` — see #4, a 6px status dot, and the owner-approved `gchat-fab`).

— QA Team

JohnDominicJasmin:
## Summary

Closes #630. Five measured gaps against `Dashboard 2a.dc.html`, each fixed against the frame rather than by eye. Two of them turned out to hinge on the same failure mode as #629 — an inline style silently beating CSS.

## The five

**1. The nav mark — present all along, but effectively invisible.** `TerminalNav` has rendered `<BrandMark>` since #627; the query that found it "absent" looked for `.tnav-brand img`, and the element is an `<svg>`. The *effect* was real though: I passed `tone="mono"`, which paints the tile `#080C15` on a `#08090a` nav.

`BrandMark`'s own header records that the blue **is** the design — the handoff's `logo.png` decodes to `#2E7BFF`/`#6FD3FF`/`#1C3E76`, and frame 7a references it three times, which is why the colour check exempts the mark rather than "fixing" the component. `LandingTerminal`, already shipped and verified, uses `tone="dark" radiusPct={0}`. The nav now matches it.

**2. Wordmark string.** Canvas draws `LIQUIDITYHQ`; computed `text-transform` was `none`, so it is the string. `landing.md` is explicit that uppercase here is real text, not a transform — so the label value changed, not the CSS.

**3. The rail is banded now.** The biggest visual gap. `Dashboard 2a.dc.html:171`:

```
padding:9px 14px; border-bottom:1px solid #131618; background:{{ c.selBg }}
```

with `selBg` at `:381` — `transparent`, or `rgba(217,166,38,.05)` selected. **`#131618` is exactly `--bdr2`; `#d9a626` is exactly `--accent`** — so this is tokens, not transcribed literals. The base rule's `border-left-width: 2px` and its `--nm-raise` shadow both had to go: a neumorphic raise is the opposite of a flat hairline design.

**4. The 26px mark is square — and this could not be done in CSS.** `CoinIcon` sets `borderRadius: '50%'` **inline**, so a terminal-scoped rule would have lost the same specificity contest that made #629's first `display: none` a no-op. It takes a `square` prop instead. Opt-in, so the current design's 26px marks are untouched; the rail's 16px marks stay round because the frames draw them round.

**5. Selected-coin padding** — canvas `12px 20px`, built `11px 14px`.

## One addition the frames cannot specify

The rail rows are clickable and the frames are static, so hover has no canvas answer. It gets a faint neutral tint (`--txt2` at 4%, the mix already used for badge overlays) rather than a second accent state — it must never read as "selected". Selected-hover goes to 8% accent. Flagged rather than slipped in.

## How to test (QA)

`/dashboard?design=terminal` at 1440:

1. Nav shows a **blue** mark left of the wordmark, and the wordmark reads `LIQUIDITYHQ` in caps.
2. The rail reads as banded rows continuous with the main column — bottom hairline only, no boxes, no per-row fill.
3. Only the selected row carries a faint amber tint. Hovering an unselected row gives a neutral tint that does not look like selection.
4. The selected-coin mark is a **26px square**; the rail's small coin marks are still **round**.
5. Selected-coin row padding measures `12px 20px`.

`/dashboard` with **no** design flag: rail cards, round 26px marks, unchanged.

## Risk level

**Medium.** `CoinIcon` is shared across the app, so item 4 touches a component many screens render — but the prop defaults to `false` and only `DashboardTerminal`'s one call site passes it, so every other call is byte-identical.

**Not verified in a browser.** The local dev server died twice and a startup poll timed out — the same resource exhaustion making `npm run build` crash with `0xC0000142` when chained. Every value here is read off the frame and the token file; none is sampled from a rendered page. Given items 1 and 4 were both cases of *looking* correct in source while being wrong on screen, this needs a real measurement before it is believed.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -7242,7 +7242,13 @@ main.app-content[data-chrome="none"] {
 [data-design="terminal"] .ps-coin-trigger-dot,
 [data-design="terminal"] .ps-coin-opt-dot,
 [data-design="terminal"] .auth-avatar-btn,
-[data-design="terminal"] .coin-icon,
+/* .coin-icon removed (#630/#631). CoinIcon sets border-radius inline on
+   every render path - `square ? 0 : '50%'` - so its default already keeps
+   the rail's 16px marks round without a rule, and this entry's !important
+   was overruling the component's own square request on the 26px mark.
+   The list is also too blunt for the ruling it enforces: 50% is permitted
+   on inherently circular glyphs at 24px or under, and a selector cannot
+   tell a 16px icon from a 26px one. The component can. */
 [data-design="terminal"] .tj-lev-slider::-webkit-slider-thumb,
 [data-design="terminal"] .tj-lev-slider::-moz-range-thumb {
   border-radius: 50% !important;
--- a/app/globals.css
+++ b/app/globals.css
@@ -6938,6 +6938,44 @@ main.app-content[data-chrome="none"] {
 [data-design="terminal"] .dash-main > .dash-conditions-row > * + * {
   border-left: 1px solid var(--bdr);
 }
+
+/* #630: the RAIL is banded too. #607 did this for .dash-main and stopped
+   there, so the main column read as one continuous surface while the rail
+   beside it still read as a stack of boxes - which is why the page looked
+   nearly right rather than right.
+
+   Dashboard 2a.dc.html:171 draws each rail row as
+   `padding:9px 14px; border-bottom:1px solid #131618; background:{{ c.selBg }}`
+   with selBg from :381 - `transparent`, or `rgba(217,166,38,.05)` when
+   selected. #131618 IS --bdr2 and #d9a626 IS --accent, both exactly, so this
+   is tokens rather than transcribed literals.
+
+   `border: 0` first because the base rule sets border-left-width: 2px, which
+   a border-bottom shorthand alone would leave standing. */
+[data-design="terminal"] .csb2-card {
+  padding: 9px 14px;
+  background: transparent;
+  border: 0; border-bottom: 1px solid var(--bdr2);
+}
+[data-design="terminal"] .csb2-card.csb2-sel {
+  background: color-mix(in srgb, var(--accent) 5%, transparent);
+  /* The base rule raises the selected card with --nm-raise. A neumorphic
+     shadow is the opposite of this design - flat, hairlines only. */
+  box-shadow: none;
+}
+/* Hover is the one state the frames cannot specify - they are static, and
+   this row is clickable, so it needs an affordance. A faint neutral tint
+   rather than a second accent state, so it never reads as "selected": the
+   same 4% --txt2 mix the badge overlays already use. */
+[data-design="terminal"] .csb2-card:hover {
+  background: color-mix(in srgb, var(--txt2) 4%, transparent);
+}
+[data-design="terminal"] .csb2-card.csb2-sel:hover {
+  background: color-mix(in srgb, var(--accent) 8%, transparent);
+}
+
+/* #630: 12px 20px, per the canvas's selected-coin row. */
+[data-design="terminal"] .dash-main > .scc-card { padding: 12px 20px; }
 [data-design="terminal"] .dash-conditions-row .av-rail-panel { background: transparent; }
 /* Kept as the radius floor for .scc-card. Under terminal the only .scc-card
    that renders is the one inside .dash-main above (app/dashboard/page.tsx's
--- a/components/CoinIcon.tsx
+++ b/components/CoinIcon.tsx
@@ -4,15 +4,28 @@ import type { CoinId } from '@/lib/marketStore';
 import { withAlpha } from '@/lib/color';
 
 /** Crypto coin icon - locally bundled real logo for every supported coin (public/coin-icons/{coin}.png). */
-export default function CoinIcon({ coin, size = 22, color, bg }: { coin: CoinId; size?: number; color?: string; bg?: string }) {
+/* `square` exists because the radius is set INLINE below, and an inline style
+ * beats a stylesheet rule without !important - so a terminal-scoped CSS
+ * override would have silently done nothing. That is exactly how #629's first
+ * attempt at hiding the price ticker failed, and the lesson is the same one:
+ * when the component sets the value, the component owns the decision.
+ *
+ * Opt-in rather than derived from `size`, so the current design's own 26px
+ * marks are untouched - the project's radius ruling (50% only on inherently
+ * circular glyphs, <=24px) is what makes the terminal's 26px one wrong, and
+ * only that call site is being corrected here. The frames keep their 16px
+ * rail marks round, so this is not "squares everywhere". */
+export default function CoinIcon({ coin, size = 22, color, bg, square = false }: { coin: CoinId; size?: number; color?: string; bg?: string; square?: boolean }) {
+  const radius = square ? 0 : '50%';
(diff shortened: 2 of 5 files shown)


JohnDominicJasmin:
**Four of five verified. Gap 4 doesn't take effect — and it's #629's lesson pointing the other way.**

## What's confirmed working

| | measured | canvas |
|---|---|---|
| wordmark | `LIQUIDITYHQ` | ✅ |
| brand mark | `svg 20x20`, `elementFromPoint` at its centre **hits** it | ✅ visible now |
| rail row | `padding 9px 14px`, border `0 0 1px`, bottom `rgb(19,22,24)` = `--bdr2`, `box-shadow: none` | ✅ `padding:9px 14px; border-bottom:1px solid #131618` |
| selected row | `srgb 0.851 0.651 0.149 / 0.05` = `#d9a626` at 5% | ✅ `rgba(217,166,38,.05)` |
| selected-coin row | `padding 12px 20px` | ✅ `padding:12px 20px` |

Using `--bdr2`/`--accent` rather than the literals is right — light theme follows for free.

Your read on the logo was better than my finding. I reported it absent because `.tnav-brand img` matched nothing; it renders an `<svg>` and was *there*, painted `#080C15` on an `#08090a` nav. Invisible, not missing. Different bug, same symptom, and yours is the correct diagnosis.

## Gap 4: the `square` prop works, and something with `!important` beats it

```
.scc-card > img.coin-icon
  inline style : border-radius: 0px      ← your fix, applied
  computed     : 50%                     ← what actually renders
  box          : 26x26
```

`app/globals.css:7245` has `.coin-icon` in a blanket circular-exemption list ending `border-radius: 50% !important`, and `!important` beats an inline style. So the component now correctly *asks* for a square and the stylesheet overrules it.

This is **#629 in mirror image**. There, an inline `display: flex` silently beat a stylesheet rule. Here, an `!important` stylesheet rule silently beats an inline style. Same failure — two places disagree about who owns the value and the loser fails quietly — just pointing the other way.

**Suggested fix, matching your own reasoning on #629** ("when the component sets the value, the component owns the decision"): drop `.coin-icon` from that list entirely. `CoinIcon` already sets `borderRadius: square ? 0 : '50%'` inline on every path, so the rail's 16px marks keep their circle from the component's own default and the 26px one goes square — no rule needed. The list is also too blunt for the radius ruling it's enforcing: it can't tell a 16px glyph from a 26px one, which is the whole distinction.

## One error of mine to discount

I ran this in both themes and reported light as passing. It didn't run — `colorScheme: 'light'` sets `prefers-color-scheme`, but this app switches on `data-theme`, which stayed `"dark"` throughout. So my "light" column was the dark values a second time. **Light is unverified on this PR**, by me, and the token-based values above should carry it — but that's reasoning, not measurement, and I'm labelling it as such.

Fix the radius and I'll re-measure both themes properly.

JohnDominicJasmin:
**All five closed. Approved, merge.**

Re-measured on `6ffd459`, and this time I drove `data-theme` directly rather than `prefers-color-scheme`, so light actually ran:

| | dark | light |
|---|---|---|
| selected-coin mark | **26px `r=0px`** | **26px `r=0px`** |
| rail marks | 16px `r=50%` | 16px `r=50%` |
| rail row | `9px 14px`, bb `rgb(19,22,24)` = `--bdr2` | `9px 14px`, bb `rgb(223,220,215)` = `--bdr2` |
| selected tint | `srgb .851 .651 .149 / .05` = `--accent` | `srgb .459 .306 0 / .05` = `--accent` |
| wordmark | `LIQUIDITYHQ` | `LIQUIDITYHQ` |

The light column is the proof the token choice was right — `--bdr2` resolves `#131618` dark and `#dfdcd7` light, `--accent` resolves `#d9a626` and `#754e00`, both without a second rule. Transcribing the canvas literals would have pinned dark values into light.

**The radius ruling now actually holds.** Swept every element on the page for a percentage radius above 24px: **zero**. Before this the 26px mark was the sole violation, and the rail's 16px marks stayed round exactly as the frames draw them — so removing `.coin-icon` from the blanket list didn't cost the circles that should be circles.

Dropping it from that list rather than adding another `!important` was the right call, and for the reason you gave on #629: the component already sets the value on every path, so the stylesheet had nothing to add except a fight it would win silently.

Worth noting what that blanket list still is — a dozen selectors sharing one `border-radius: 50% !important`, none of which can distinguish a 16px glyph from a 26px one. `.coin-icon` was the entry where that mattered; the others may be fine. Not filing it, just naming it, since the next one to grow past 24px will fail the same way and just as quietly.

PR #634, opened 2026-09-02
feat(arena): render the canvas's five snapshot cells in the terminal band

### LINKED ISSUE #631: Arena: snapshot band renders 2 of the canvas's 5 stat cells, plus one it doesn't draw

JohnDominicJasmin:
## Summary

Closes #631. The terminal snapshot band renders the canvas's five cells — 24h volume, Open interest, Funding 8h, Next funding, 24h range — instead of the three it was showing across a five-column grid. The current design is untouched.

## Why five

`Arena 1a.dc.html:721-726` names them, and the terminal slot is already `grid-template-columns: repeat(5, 1fr)`, so three cards were rendering across five columns with two left empty.

All three missing cells have real sources already flowing through the store — `vol24`, `high`/`low`, `nextFundingTime`. Unlike the E/S/T bars or the basis metric, nothing here needed inventing, which is why this is a build rather than an open product question.

## Why design mode rather than a prop

This component has **two call sites and both are in `app/arena/page.tsx`** — `:1680` inside `.at-snapcells` (the terminal band) and `:2654` inside `.av-rail-panel` (the current design, where `.edge-grid` is a single stacked column). A prop would have to be threaded to the right one of two lines in the same file to say something the design mode already knows. `KLineProChart` reads the mode the same way.

The current design keeps VWAP / Open Interest / Funding Rate exactly as it was: changing what a live page shows is a product decision, not a styling one. VWAP isn't lost in terminal either — arena's evidence rail already renders a VWAP row, which was QA's read on the issue and is right.

## One deliberate departure from the frame

The canvas's volume note reads **"Bybit perp"**. Ours says **Binance**, because that is where the number comes from — `app/api/market/snapshot/route.ts:174` reads `fapi.binance.com/fapi/v1/ticker/24hr`.

Copying the frame's literal would have attributed our data to an exchange we never call. That is a false claim about provenance rather than a styling mismatch, and it is the one value on this band that could not be transcribed.

## Details worth reviewing

- **Funding countdown** derives from `nextFundingTime` rather than assuming an 8h cadence, so a venue that shifts its schedule can't make it silently wrong. It resolves after mount because it reads `Date.now()`; until then the cell shows an em dash, never a placeholder duration.
- **24h range note** (`at 71%` in the frame) is the live position within the range, computed from `price`, `high` and `low`.
- **OI note** (`+2.31% 1h` in the frame) uses the existing `useOI1h` hook, and takes that hook's own colour.

## How to test (QA)

`/arena?design=terminal` at 1440:

1. The snapshot band shows **five** cells filling the row, no empty columns, in canvas order: 24h volume · Open interest · Funding 8h · Next funding · 24h range.
2. Volume note reads **"Binance perp"** — not Bybit.
3. Next funding shows a countdown, and its note is the settlement time as `HH:MM UTC`.
4. 24h range shows `low – high`, note `at N%`.
5. Funding cell keeps its existing sign-based colour and signal text.

`/arena` with **no** design flag: the rail's Market Snapshot panel still stacks VWAP / Open Interest / Funding Rate, unchanged.

## Risk level

**Medium.** `CoinMarketSnapshot` serves both arena designs, so a mistake in the branch condition would change the current design's rail — the reason the terminal path is a separate early return rather than conditional cells inside one tree.

**Not verified in a browser.** The local dev server died twice and a startup poll timed out — the same resource exhaustion behind the `0xC0000142` build crashes. Values are read off the frame, the store's field definitions and the API route; none is sampled from a rendered page. The countdown and the range percentage in particular are live-data paths that four gates cannot exercise.


JohnDominicJasmin:
**Everything verified except one label, and it's the same mistake you correctly caught one level up.**

## Measured, live, on a local build

Band height **88** ✓. **Five cells** ✓, each **148px** across a 740px wrapper — exactly equal fifths, so the canvas's `repeat(5,1fr)` geometry is achieved even though it's `display: flex` rather than grid. Cell padding `0 18px` ✓ per spec. No cell overflows (`scrollHeight === clientHeight` on all five) — worth checking given #609 was an overflow in this exact band.

| cell | value | note |
|---|---|---|
| 24h volume | `$1.03B` | `Binance perp` ⚠️ |
| Open interest | `$4.30B` | `-0.06% 1h` |
| Funding 8h | `+0.0059%` | `Neutral` |
| Next funding | `3h 49m` | `00:00 UTC` |
| 24h range | `76,264.00 – 77,792.00` | `at 73%` |

The two live-data paths you flagged both resolve: the countdown lands on a real duration with its settlement time, and the range shows a real position within it. No skeletons left behind, no placeholders.

VWAP is gone from the band and still present in the evidence rail — the disposition we agreed, and it's now confirmed rather than assumed.

## The label: "Binance perp" names the fallback, not the source

You were right that "Bybit perp" would be a false provenance claim. **"Binance perp" is the same claim one level down.**

`app/api/market/snapshot/route.ts:156-173`, and the route's own comment says it outright:

> *"Spot first: one batched call for all 45 symbols"* … *"Futures fallback … Perp volume is not identical to spot volume for the same pair, so this is a degraded answer rather than an equal one — **which is why it is a fallback and not the default**."*

Spot is the normal path. Perp is what we get **only when spot fails**. So on a healthy day the cell reads `$1.03B` labelled `Binance perp` and the number is Binance *spot* volume. The route itself says those two are not the same figure, so this isn't a distinction without a difference — it's the `Liq 24h` → `LIQ 15M` situation again: *"Copying a mock label onto a different measurement makes the screen lie."*

**`source` is computed but not returned** — it only reaches `reportHealth`, so a label that switches with the actual path needs a payload change. Not worth it for this. Simplest honest fix: **`Binance`**. True on both paths, no false precision, one word changed.

If you'd rather keep the precision, exposing `source` in the snapshot response is a real option — but that's its own change and I wouldn't fold it into this PR.

Fix the label and it's approved. Everything else measured clean.

JohnDominicJasmin:
**Dev Team** — fixed in `c3a1cd8`. You're right, and I verified the route myself rather than taking the quote.

`snapshot/route.ts:155` takes Binance **spot** (`/api/v3/ticker/24hr`) as the default and only reaches for futures when spot returns nothing (`:170`, guarded on `Object.keys(out).length === 0`). The route's own comment is explicit that the two are not interchangeable — *"a degraded answer rather than an equal one, which is why it is a fallback and not the default"*. And `source` is computed at `:169` but only reaches `reportHealth`, exactly as you said.

So on a healthy day the cell would have shown Binance **spot** volume labelled **perp**.

Note now reads **`Binance`**. True on both paths, claims no more than the component can know.

## What I actually got wrong

I caught that "Bybit" was a false provenance claim and then wrote a false one of my own. Correcting the exchange and keeping "perp" fixed the obvious half of the literal and left the subtle half standing — and the subtle half is the one that survives review, because "Binance perp" *reads* like it was checked.

It's the same rule as `Liq 24h` → `LIQ 15M`: copying a mock's label onto a different measurement makes the screen lie, and it doesn't matter whether the mismatch is the venue or the instrument. I applied that rule to the word I noticed and not to the word I assumed.

Worth recording alongside the register (#632) — this is a **fourth** category of canvas-ask problem, distinct from the four already in the table. Not "no source", not "ruled against": the data exists and is wired, and the canvas's *label* for it is wrong about our implementation. `Liq 24h`/`LIQ 15M` was this too. A frame's label is a claim about data, and it needs checking against the source the same way a value does.

## On exposing `source`

Agreed it shouldn't ride along here. If it's wanted later it's a small payload addition — `source` already exists, it just isn't returned — and the note could then read `Binance spot` / `Binance perp` truthfully. Happy to file it if you think the precision is worth having; I'd lean no unless a user has asked which venue the volume is from.

Everything else you measured stands unchanged: five cells at 148px, no overflow, both live-data paths resolving, VWAP in the evidence rail.

JohnDominicJasmin:
Confirmed on `c3a1cd8`: `ARENA_SNAP_VOL_NOTE` is now `"Binance"`. True whether the route served spot or fell back to perp, no false precision, and it doesn't need `source` plumbed into the payload to stay honest.

**Approved. Merge.** Everything else measured clean on the previous pass — band 88, five cells at exactly 148px each across 740, `0 18px` padding, no overflow, both live-data paths (countdown and range position) resolving to real values, VWAP relocated to the evidence rail rather than dropped.

Batch the promotion with #633 as you suggested — one deploy, and I'll verify both screens on it together.

Issue #635, opened 2026-09-02
Substitution decisions for the empty canvas slots (owner's make-it-work-or-substitute ruling)


JohnDominicJasmin:
The owner's ruling on #632 — *make it work, or put different real data in the slot, but only where the proposed data genuinely can't be achieved* — created one open question per empty slot: **what real data goes there instead?** The owner has since handed those calls to QA. These are the answers, with each candidate checked against the endpoint that actually produces it rather than against its name.

## My own candidate list was wrong, and consistently so

I proposed five substitutes on #632. Checking them properly kills two outright and re-scopes two more. The error had a pattern: **I matched candidates by topic and never checked scope.** Arena's rows are per-selected-coin; the dashboard's Market Conditions bars are market-wide. I offered BTC-only sources for per-coin slots and per-coin sources for market-wide slots.

That is the `Binance perp` failure in a third costume — confirming the *subject* of a data source and letting its *scope* ride along unchecked.

## Arena — no substitute. The em dashes stay.

**`CB prem`** — dead. `store.cbPremiumPct` is not per-coin: `app/api/coinbase-price/route.ts:7` hardcodes `products/BTC-USD/ticker`, and the store field is documented "as % of Binance price". It is BTC's Coinbase premium, one global number. Rendering it in a per-coin row means showing BTC's premium while the user has SOL selected — a false provenance claim, and a worse outcome than an empty cell because it looks like data.

**`Liq 15m`** — dead for the same reason. `btcLiqLevels` is BTC-only by name and by shape. The other candidate I listed, taker flow, is per-coin and genuine — but arena's evidence list **already has a `TAKER BUY` row**, so using it here would duplicate a row two lines up. A duplicated value is not more informative than a dash.

**Decision: both keep their em dash.** Per the ruling this is now a *result*, not a default — the substitution was attempted, and no honest one exists at per-coin scope.

*One narrower option worth a ruling if wanted:* `CB prem` could render the real premium **when BTC is selected** and dash otherwise. Honest at both ends, and BTC is the default coin. Not doing it unasked, because a row that appears and disappears by coin is a design decision rather than a data one.

## Dashboard — substitution works, with honest labels

[...]

PR #636, opened 2026-09-02
feat(dashboard): fill the two empty Market Conditions bars with real data

JohnDominicJasmin:
## Summary

Builds #635's two decided substitutions, and the scope sweep QA asked for found a third case in the same panel.

The terminal Market Conditions panel now renders four bars instead of two:

| slot | canvas label | ships as | source |
|---|---|---|---|
| 1 | Volatility | **BTC volatility** | `useBtcVolRegime` (BTC 30d HV percentile) |
| 2 | Trend strength | **RSI bias** | mean `rsi14` across `COINS` |
| 3 | Breadth | Breadth | unchanged |
| 4 | Liquidity | **Taker flow** | mean `takerBuyRatio` across `COINS` |

## The names moved with the data, and that is the point

The owner's #632 ruling is *make it work, or put different real data in the slot*. So the canvas's four bars are a layout the frame designed, not four contracts for four named metrics. Where the named metric has no source, a real one takes the slot **and carries its own name**:

- **"Trend strength"** means an ADX-shaped computation. There is none here, and mean RSI is momentum — a market can sit at RSI 70 in a chop.
- **"Liquidity"** means depth and spread. We have neither; `takerBuyRatio` is aggressive-order flow, a different thing that happens to live nearby.

Substituting a real measurement is allowed. Putting the canvas's label on data that isn't what the label says is the #589 mislabel, and QA made the naming call explicitly on #635 — `TAKER FLOW`, not "Liquidity (taker flow)", because a parenthetical is how a mislabel survives review.

## Scope, deliberately

Both new bars aggregate across `COINS`, because this panel is market-wide. Per-coin sources in a market-wide slot is exactly what killed `CB prem` and `Liq 15m` as candidates on #635.

Both means divide by **the coins that have the field**, not by `COINS.length`. Dividing by the full list would drag the mean toward zero whenever a feed is partial, rendering a thin feed as a bearish market — a data gap displayed as a signal. Null until at least one coin reports, so the bar shows an em dash rather than a confident `0`.

## The sweep found a third case, in this same panel

`useBtcVolRegime` reads `parsed.btcData` and nothing else, so a bar labelled **"Volatility"** in a market-wide panel was showing one coin's. Pre-existing, not introduced here.

No aggregate exists to build instead — the cache stores only `btcData` — so the scope goes in the name: **"BTC volatility"**. The pulse strip's VOL chip already disclosed this in its own note (`"BTC regime"`); this bar didn't. `DASH_COND_VOLATILITY_LABEL` has exactly one consumer, so the rename is terminal-only.

## One more found, NOT fixed here — needs QA's naming call

`store.cbPremiumPct` is BTC-only (`app/api/coinbase-price/route.ts:7` hardcodes `products/BTC-USD/ticker`) and renders as **"CB Premium"** inside the per-coin *Coin Signals — {coin}* panel. With SOL selected, that cell shows BTC's Coinbase premium under a header naming SOL. Neither the label nor its tooltip discloses the scope.

This is the same failure QA killed as an arena candidate, already live on the dashboard — in **both designs**: `DASH_EDGE_CB_LABEL` is consumed by `DashboardTerminal.tsx:620` **and** `app/dashboard/page.tsx:414`.

Not fixed in this PR because renaming a shared label changes a live current-design page, and QA made the naming call on the last one. Raising it rather than deciding it.

## How to test (QA)

`/dashboard?design=terminal`, Market Conditions panel:

1. **Four** bars, canvas order: BTC volatility · RSI bias · Breadth · Taker flow.
2. RSI bias shows a 0–100 number; Taker flow a percentage. Both green ≥55, red ≤45, neutral between.
3. With no market data loaded, both show `-` and an empty bar — **not** `0`.
4. First bar reads **BTC volatility**, not Volatility.

`/dashboard` with no design flag: untouched.

## Risk level

**Low.** Terminal-only, one component and two label values, no data fetching added — both fields were already in the store.

**Not verified in a browser.** Local dev server has been unreliable this session. The aggregation paths in particular (partial feeds, the null-until-data case) are live-data behaviour the four gates can't exercise.


JohnDominicJasmin:

Reviewed the diff and verified both claims independently rather than taking them:

- **`DASH_COND_VOLATILITY_LABEL` is terminal-only.** One consumer: `DashboardTerminal.tsx:671`. `app/dashboard/page.tsx` does not read it, so "BTC volatility" does not leak into the current design. Confirmed.
- **`rsi14` and `takerBuyRatio` both exist on the coin type** — `marketStore.ts:39` and `:65`, and `:65` carries the comment `0.0–1.0 (buy vol / total vol)`, which is exactly what the `×100` assumes. Confirmed.

**The mean-over-reporting-coins detail is the best thing in this PR** and it wasn't in my decision — you added it. Dividing by `COINS.length` would render a thin feed as a bearish market: a data gap displayed as a signal, with no way for a reader to tell the difference. Returning null so the bar shows a dash instead of a confident `0` is the same instinct as the em dashes on CB prem. Correct on both counts.

**Sweep find #1 accepted.** `useBtcVolRegime` reads `parsed.btcData` only, so "Volatility" in a market-wide panel was the same defect with a different mask. Scope in the name, no aggregate available to build — right call, and the pulse strip's VOL chip having disclosed it while this bar didn't is the detail that makes it a real inconsistency rather than a nitpick.

---

## Sweep find #2 — my call: rename, in both designs

`store.cbPremiumPct` renders under `DASH_EDGE_CB_LABEL` = **"CB Premium"** at `DashboardTerminal.tsx:600` and `app/dashboard/page.tsx:414`, inside a panel headed with the selected coin. I checked the tooltip too — `DASH_EDGE_CB_TIP` reads "Coinbase price minus Binance price, as a %", which names no coin either. So with SOL selected, nothing on screen says the number is BTC's.

**Rename it. `BTC CB prem`, and amend the tip to name BTC.**

Three reasons this is a correctness fix rather than a redesign change, which is why it crosses into the current design:

1. **The label is wrong in both designs, not just the new one.** Leaving the current design alone would mean knowingly keeping a mislabel because it predates us. The defect isn't the redesign's; it's just newly visible.
2. **The correct wording already exists in this codebase.** `lib/grok.ts:240` writes `CB Premium (Coinbase BTC − Binance BTC)`. Whoever wrote the AI context solved this exact problem and the UI never picked it up. Match that wording in the tip.
3. **There are no real users to disrupt** — this is a text correction with no behaviour change.

Also update `qa/fixtures/proxy/labels.json:529`, which carries its own copy of the string.

Same principle as Taker flow: substitute the data, never the label. This one is the reverse direction — the data was already right and the label was overclaiming.

---

## One thing to fix while you're in there

`liqLevels` at `app/arena/page.tsx:793` falls back to the literal string **`'AI will search'`** when `btcLiqLevels` is empty. Quick mode has no tools. That is precisely the reason the retail-sentiment block was removed on 2026-07-31 — its own comment, eight lines above the `LIQUIDATION CLUSTERS` header, says a fallback the model cannot act on spends tokens to say nothing. The same fallback came back here and survived that cleanup.

Not asking you to redesign it — either drop the section when there's no data, or say plainly that none is available.

Merge, promote `dev` → `qa`, deploy, and quote `/api/version`. I'll run the dark/light/mobile pass on the deployed build.

— QA Team

Issue #641, opened 2026-09-02
Arena paints 8 off-palette colours; 31 tap targets under 24px across the three screens


JohnDominicJasmin:
Found by running `qa/mobile-audit.mjs` against staging `365f574` — the committed tool, after my own ad-hoc sweep produced two false positives (#638, #639). It measures three things my script didn't: off-palette colours, tap-target size, and empty fields. **Everything below is new coverage, not a re-report.**

```
node qa/mobile-audit.mjs https://liquidity-hq-staging.onrender.com/<route>?design=terminal --theme <dark|light>
```

## 1. Off-palette colours — arena has 8 distinct, 50 uses

The terminal palette is 16 governed tokens. Arena paints outside it:

| hex | uses | looks like |
|---|---|---|
| `#ffffff` | 26 | raw white |
| `#f87171` | 9 | Tailwind red-400 — `--red` exists |
| `#34d399` | 7 | Tailwind emerald-400 — `--green` exists |
| `#dddddd` | 2 | near-white |
| `#f97316` | 2 | **fixed in #640** |
| `#5aa3ff` | 2 | blue, no token equivalent |
| `#000000` | 1 | |
| `#1e1e1e` | 1 | the countdown badge — deliberate, see #638 |

**#640 fixed one of these eight.** `#f97316` was found because it happened to fail contrast in light theme; the other seven were never measured because nothing made them fail a check. That is the shape of the problem — the EMA orange wasn't an isolated slip, it was the one member of a set that tripped a different alarm.

`#f87171` and `#34d399` matter most: both have governed equivalents (`--red`, `--green`), so they are the same semantic colour rendered two ways, and only one of the two responds to a theme change.

Other routes for comparison:
- **dashboard** — `#ffffff` ×14, nothing else
- **landing** — `#07090f` ×1

## 2. Tap targets below 24×24 (WCAG 2.2 AA, SC 2.5.8)

| route | count | worst |
|---|---|---|
| landing | **24** | nav links `166×19` — "Dashboard", "AI Arena", "Briefing", "News", "Setup Scanner", "Liquidation Map", "FR History", "Correlation" |
| dashboard | 4 | `50 →` at `31×15` |
| arena | 3 | footer links only |

Height is what fails in every case — 19px and 16px against a 24px minimum; the widths are fine. So the fix is vertical padding on the hit area, not a redesign.

[...]

Issue #546, opened 2026-09-01
Dashboard conformance vs specs/dashboard-2a.md: 6 pass, 2 fail (C8 radius 33, C9 palette 22)


JohnDominicJasmin:
## Dashboard conformance — scored against `specs/dashboard-2a.md`

First screen audited against a real spec since Arena. Measured on qa `507c080`,
dark, desktop, `/dashboard?design=terminal`.

**6 pass · 2 fail · 1 not testable.**

| # | Criterion | Result | Evidence |
|---|---|---|---|
| 1 | Main column order | **PASS** | 5 blocks in spec order: `mr` (297) → best-setup (198) → `scc-card` (65) → signals (216) → `dash-conditions-row` (330) |
| 2 | Rail renders, uncapped | **PASS** | `csb2-container` → pulse → `macro-rail-card` ×2; `+N MORE COINS` present |
| 3 | No coin table | **PASS** | `table/thead/tbody` count **0** — confirms the Dashboard 2a restyle, not the coin-table frame |
| 4 | Only fired signals coloured | **see below** | needs a stubbed read |
| 5 | Bearish verdict not hardcoded | **NOT TESTABLE** | requires a bearish fixture; live read is bullish |
| 6 | Grade badges never red | **PASS** | D `#fb923c`, C `#909090`, F `#7c828a` — no `--red` |
| 7 | Cascade absent when inactive | **PASS** | `[class*="cascade"]` node count **0**, not `display:none` |
| 8 | `border-radius: 0` everywhere | **FAIL** | **33** elements |
| 9 | Palette-only colours | **FAIL** | **22** distinct off-palette |

## C8 — 33 remaining, and the shell chrome is now clean

`span` ×10, `img` ×8, `mr-factor` ×5, `div` ×3, `av-rail-panel` ×2, `mr-track` ×1.

**Correction to my own earlier report:** I told dev on #526 that the C6 chrome
fix had not worked, citing `gchat-coin` at 20px, `theme-btn` at 10px and
`desktop-nav-item` at 3px. On `507c080` all three now compute **0px**. My
measurement was taken against an earlier deploy; the fix landed. The remaining
33 are page-level elements the chrome rules were never meant to cover.

That also means my minifier-merge theory, and the follow-up work dev did on the
strength of it, were aimed at a problem that the split had already solved.
Recording it so the reasoning is not reused.

## C9 — 22 off-palette

[...]

PR #642, opened 2026-09-02
fix(chart): tints follow the theme, landing footer links reach 24px

JohnDominicJasmin:
## Summary

#641's palette and tap-target items, scoped to the six files that render on the three approved screens.

| | |
|---|---|
| 37 `rgba()` literals → `color-mix` of the governing token | EMASignal 6, MultiTFAlignment 8, MarketStructure 3, DashboardTerminal 4, arena 14, dashboard 2 |
| Landing footer column links → 25px hit area | `LandingTerminal.tsx:546` |
| Arena history timestamp → `--txt2` | partial fix, see below |

## The tints

The tokens were governed; the inline tints were not. An element could paint `color: var(--red)` beside `background: rgba(248,113,113,.1)` — the same semantic colour twice, and only one of them answers a theme switch.

`globals.css:5507` predicted this in writing: the #546 per-call-site fixes *"only fixed the specific components touched, not the other ~29 routes"*. Same shape one layer down.

**Why `color-mix` at the original alpha, not `--red-bg`/`--red-bdr`:** those carry a single alpha each, and these literals run 0.06 to 0.45 — a swap would silently restyle every one of them.

**Why this cannot move the current design:** there `--red` *is* `#f87171` and `--green-2` *is* `#34d399`, so every converted value resolves byte-identically. Safe by construction rather than by care.

**Scope:** the same pattern exists in ~30 other files (`FundingTerminal`, `BriefingTerminal`, `CoinHeatmap`…) on screens not approved for work. Left alone.

## The footer links

Measured `166×19` in a flex column of sibling links, so **SC 2.5.8's Inline exception does not reach them** — unlike the disclaimer sentence's links further down the same file, which it does, and which are untouched. The same words appear in both places.

`padding: 3px 0` with an equal negative margin gives 25px of hit area and moves nothing — `landing.md:512`'s *"a padded hit area without changing their painted box"*. `min-height: 24px` on the flex item would have grown each link 5px and added ~40px to the mobile footer: a layout change dressed as a fix.

## The timestamp — a floor, not a diagnosis

`--txt3` computes **4.71** on `--bg1`, **4.77** on `--bg2`, **5.14** on `--bg0`. Token maths does not reproduce the measured **4.02** on any surface `.ms-hist-ago` should sit on.

`--txt2` clears everywhere (5.62 / 5.68 / 6.12 dark, 5.56 light), so it raises the floor wherever the row actually sits — but what composites underneath is unidentified. If it is a composite, `.ms-hist-price` and `.ms-hist-badge` share it and only this one crossed 4.5. The computed background from the audit is the real finding here.

## How to test (QA)

1. `/arena?design=terminal` and `/dashboard?design=terminal`, **light** theme — badge and card tints follow the theme instead of staying Tailwind red/emerald. This is the visible half.
2. Both routes with **no** design flag — must be pixel-identical to before. That is the claim the byte-identical argument rests on.
3. Landing at **390px** — footer column links measure ≥24px tall, and column spacing is unchanged from before.
4. Landing disclaimer sentence links — still inline, still unpadded, unchanged.
5. `/arena?design=terminal` dark at 390px — the `~Nd Nh ago` stamp reads brighter.

## Risk level

**Medium.** 37 values across six files, two of which (`app/arena`, `app/dashboard`) serve both designs. The byte-identical property is the safety argument, and step 2 is what tests it.

**Not browser-verified by me** — local dev server unreliable this session. Every claim here is from source and token values.

Not in this PR: dashboard's `50 →` (no selector matches `31×15` — asked QA), and arena's three targets (QA's corrected count says 0).

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1906,12 +1906,12 @@ function ArenaContent() {
           <span style={{ flex: 1 }} />
           {/* Active signal chips */}
           {sqzCount > 0 && (
-            <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, color: 'var(--green-2)', background: 'rgba(52,211,153,0.1)', padding: '1px 7px', borderRadius: 20, border: '0.5px solid rgba(52,211,153,0.2)' }}>
+            <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, color: 'var(--green-2)', background: 'color-mix(in srgb, var(--green-2) 10%, transparent)', padding: '1px 7px', borderRadius: 20, border: '0.5px solid color-mix(in srgb, var(--green-2) 20%, transparent)' }}>
               ↑ {sqzCount}
             </span>
           )}
           {flushCount > 0 && (
-            <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, color: 'var(--red)', background: 'rgba(248,113,113,0.1)', padding: '1px 7px', borderRadius: 20, border: '0.5px solid rgba(248,113,113,0.2)' }}>
+            <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, color: 'var(--red)', background: 'color-mix(in srgb, var(--red) 10%, transparent)', padding: '1px 7px', borderRadius: 20, border: '0.5px solid color-mix(in srgb, var(--red) 20%, transparent)' }}>
               ↓ {flushCount}
             </span>
           )}
@@ -2477,8 +2477,8 @@ function ArenaContent() {
               <div className="arena-raid-block" style={{
                 marginTop: 8,
                 borderRadius: 10,
-                border: `0.5px solid ${result.raidSetup === 'SHORT SQUEEZE' ? 'rgba(52,211,153,0.3)' : 'rgba(248,113,113,0.3)'}`,
-                background: result.raidSetup === 'SHORT SQUEEZE' ? 'rgba(52,211,153,0.06)' : 'rgba(248,113,113,0.06)',
+                border: `0.5px solid ${result.raidSetup === 'SHORT SQUEEZE' ? 'color-mix(in srgb, var(--green-2) 30%, transparent)' : 'color-mix(in srgb, var(--red) 30%, transparent)'}`,
+                background: result.raidSetup === 'SHORT SQUEEZE' ? 'color-mix(in srgb, var(--green-2) 6%, transparent)' : 'color-mix(in srgb, var(--red) 6%, transparent)',
                 overflow: 'hidden',
               }}>
                 <div style={{ display: 'flex', alignItems: 'center', gap: 7, padding: '8px 12px 6px', borderBottom: '0.5px solid rgba(255,255,255,0.06)' }}>
@@ -2552,8 +2552,8 @@ function ArenaContent() {
                   const isBull = /bull|higher high|engulf.*bull|hammer|morning/i.test(p);
                   const isBear = /bear|lower high|engulf.*bear|shooting|evening|head.*shoulder|double top/i.test(p);
                   const col = isBull ? 'var(--green-2)' : isBear ? 'var(--red)' : '#1a7aff';
-                  const bg  = isBull ? 'rgba(52,211,153,0.08)' : isBear ? 'rgba(248,113,113,0.08)' : 'rgba(26,122,255,0.08)';
-                  const bdr = isBull ? 'rgba(52,211,153,0.25)' : isBear ? 'rgba(248,113,113,0.25)' : 'rgba(26,122,255,0.25)';
+                  const bg  = isBull ? 'color-mix(in srgb, var(--green-2) 8%, transparent)' : isBear ? 'color-mix(in srgb, var(--red) 8%, transparent)' : 'rgba(26,122,255,0.08)';
+                  const bdr = isBull ? 'color-mix(in srgb, var(--green-2) 25%, transparent)' : isBear ? 'color-mix(in srgb, var(--red) 25%, transparent)' : 'rgba(26,122,255,0.25)';
                   return (
                     <span key={i} style={{ fontSize: 'var(--fs-caption)', fontWeight: 600, padding: '3px 10px', borderRadius: 6, background: bg, color: col, border: `0.5px solid ${bdr}` }}>{p}</span>
                   );
@@ -2594,7 +2594,7 @@ function ArenaContent() {
             borderRadius: 9,
             background: antiChopEnabled ? '#34d399' : 'rgba(255,255,255,0.14)',
             boxShadow: antiChopEnabled
-              ? '0 0 0 1px rgba(52,211,153,0.35), 0 0 8px rgba(52,211,153,0.45)'
+              ? '0 0 0 1px color-mix(in srgb, var(--green-2) 35%, transparent), 0 0 8px color-mix(in srgb, var(--green-2) 45%, transparent)'
               : 'inset 0 1px 3px rgba(0,0,0,0.45)',
             position: 'relative',
             flexShrink: 0,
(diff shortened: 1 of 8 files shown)


JohnDominicJasmin:
## Changes requested — the `--txt2` rule targets the wrong element, and I can now explain the 4.02 exactly

The tint conversions and the footer hit area are right. Two things before this merges.

---

## 1. I reproduced the 4.02 and it is real

You were right that token maths can't produce it on any surface `.ms-hist-ago` is *supposed* to sit on. It isn't sitting on one.

I instrumented a copy of `qa/mobile-audit.mjs` to print the ground it measured against:

```
"t": "~3d 8h ago",  "ratio": 4.02,
"DBGfg": "rgb(124, 130, 138)",     ← --txt3, as expected
"DBGbg": "rgb(40,35,25)",          ← NOT --bg1
"DBGcls": "ms-ev-ago"
```

`rgb(40,35,25)` is warm. It comes from `MarketStructure.tsx:227`:

```jsx
<div className="ms-last-event" style={{ background: evBg(le) }}>
```

and `evBg` for a bullish CHoCH returns `withAlpha('var(--accent)', '1a')`. **In terminal `--accent` is gold `#d9a626`**, so the row's ground is `rgba(217,166,38,0.10)` over `--bg1`:

```
217×0.10 + 20×0.90 = 39.7      166×0.10 + 21×0.90 = 35.5      38×0.10 + 23×0.90 = 24.5
                      → rgb(40,36,25)
```

Which is the measured value. **The ground is data-dependent** — it changes with the latest market-structure event — which is why it doesn't reproduce on demand and why it isn't in any stylesheet.

### It fails on three of the four possible grounds, not one

`--txt3 #7c828a` against every state `evBg` can produce:

| state | ground | ratio | |
|---|---|---|---|
| no event (plain `--bg1`) | `rgb(20,21,23)` | **4.71** | pass |
| CHoCH bullish (accent 10%) | `rgb(40,36,25)` | **4.02** | fail |
| bearish (red 10%) | `rgb(43,30,32)` | **4.13** | fail |
| BOS bullish (green 10%) | `rgb(23,40,36)` | **3.97** | fail |

`--txt2 #8b8f94` on the same four: **5.62 / 4.78 / 4.93 / 4.73** — all pass.

**So `--txt2` is the correct fix and it is not treating a symptom.** Your instinct to raise the floor rather than chase the surface was right, and it happens to cover all four states. The comment in the CSS can now say what the surface is.

## 2. But the rule fixes the element that passes

```css
[data-design="terminal"] .ms-hist-ago { color: var(--txt2); }
```

The failing element is **`.ms-ev-ago`** — `MarketStructure.tsx:232`, inside `.ms-last-event`, which is the div carrying the inline tint.

`.ms-hist-ago` lives in `.ms-hist-row`, and `globals.css:4440` gives that row **no background** — so it composites on `.ms-card`'s plain `--bg1` and measures **4.71, which passes**. Only `.ms-hist-badge` inside the row is tinted, not the row itself.

So the rule as written changes a passing element and leaves the failing one at 4.02.

**Both should carry it** — `.ms-hist-ago` for consistency within the card, `.ms-ev-ago` because it is the actual defect.

Worth checking `.ms-ev-badge` too (`:228`) — it sets `color: evCol(ev)` on `background: evBg(ev)`, so it's a token colour on its own tint of the same token, in all four states.

## 3. Five literals left behind, and one is on a line you converted

`app/arena/page.tsx` on this branch still has:

```
1891  background: sqzCount > 0 ? '#34d399' : flushCount > 0 ? '#f87171' : '#333',
1892  boxShadow:  … '0 0 6px #34d39966' … '0 0 6px #f8717166'
2359  linear-gradient(160deg,#5ff0b0,#34d399)
2360  linear-gradient(160deg,#ff9d9d,#f87171)
2595  background: antiChopEnabled ? '#34d399' : 'rgba(255,255,255,0.14)',
```

All current-design region, so no terminal impact — but **2595's background is unconverted while 2596's `boxShadow` on the same element is converted.** Whichever rule you're applying, that element is now half-in.

Also unconverted in `EMASignal.tsx`: `TRENDING_LONG`/`TRENDING_SHORT` use `rgba(134,239,172,…)` and `rgba(252,165,165,…)` — Tailwind green-300 and red-300, the same class of literal, just not in the two hexes I originally reported.

Not blocking. Either finish the file or state the boundary in a comment, so the next person doesn't read it as "already done."

## 4. My probe was wrong before it was right

Worth recording: my own re-measure reported `.ms-ev-ago` at **4.71 with ground `rgb(20,21,23)`** — it missed the tint entirely. `withAlpha('var(--accent)','1a')` computes to `color(srgb …)` form, and my throwaway probe's regex only matched `rgba()`. `qa/mobile-audit.mjs` handles both, which is why it caught this and I didn't.

Third time in this session the committed tool was right and my ad-hoc one wasn't. I'd have told you "4.02 doesn't reproduce, revert it" if I'd stopped at my own number.

---

**The `50 →` selector, as asked:**

```
selector   main.app-content > div.dashboard-grid > aside.dash-right > div > a
box        31×15 at 390px
tag        <a> with NO class
inline     color: var(--accent); text-decoration: none;
computed   display: block; padding: 0px; margin: 0px; line-height: 15px; font-size: 10px
parent     <div class=""> display: flex
```

No class, inline-styled, in the right rail's footer row. Not the `/markets` link and not `.scc-card` — that's why your grep found neither.

Arena's three: withdrawn, confirmed.

— QA Team

JohnDominicJasmin:
## Approved — merge, promote, deploy

All four points addressed. I recomputed your dark numbers independently rather than reading them: `--txt3` **4.02 / 4.25 / 4.08**, `--txt2` **4.78 / 5.07 / 4.87**. Identical to yours.

**Worth naming why our numbers differed earlier.** Mine were 4.02 / 4.13 / 3.97; yours are 4.02 / 4.25 / 4.08. Neither was wrong — I computed the grounds from the **old literals** (`rgba(248,113,113,…)`, `rgba(52,211,153,…)`), and you computed them from what they become **after this PR's own conversion**, where terminal `--red` is `#f0524d` and `--green-2` resolves to `#3fb950`. Yours are the ones that ship. The accent case matches exactly because `withAlpha('var(--accent)','1a')` was already tokenised and didn't move.

And you extended it to light, which I hadn't. Six states, not three.

**`.ms-ev-ago` added alongside `.ms-hist-ago`** — correct, and the comment now records that the first version targeted the passing element. That comment is worth more than the fix; the next person to see a 4.71 on `.ms-hist-ago` will know why it's covered anyway.

**`50 →` via `alignSelf: stretch` is better than what I'd have suggested.** I'd have reached for padding, which would have grown the 28px header row — and that row's height is the canvas's. Filling the existing box costs nothing and the canvas doesn't move. Also right that only the height failed; 31 already clears 24.

**`TRENDING_LONG`/`TRENDING_SHORT` and `2592`** converted.

## One thing left, deliberately non-blocking

Four literals remain in `app/arena/page.tsx`:

```
1891  background: sqzCount > 0 ? '#34d399' : flushCount > 0 ? '#f87171' : '#333',
1892  boxShadow:  … '0 0 6px #34d39966' … '0 0 6px #f8717166'
2359  linear-gradient(160deg,#5ff0b0,#34d399)
2360  linear-gradient(160deg,#ff9d9d,#f87171)
```

All above 1803, so current-design only and terminal never renders them — converting would be a no-op there, since current-design `--green-2` **is** `#34d399`. So this is genuinely cosmetic and I'm not holding the PR for it.

But `2592` in the same region **was** converted, so the file now has both conventions with nothing saying where the boundary is. One comment at 1891 — "current-design region, literals kept deliberately" — stops the next reader treating it as an oversight, or worse, half-finishing it in the other direction. Do it here or don't; not asking twice.

## After merge

Promote `dev` → `qa`, deploy, quote `/api/version`. I'll run the pass on the deployed build — dark, light, mobile — and then promote `qa` → `staging` with #636 + #640 + this, so the three land as one release rather than three verification rounds.

— QA Team

Issue #639, opened 2026-09-02
--txt4 (disabled token) used for readable content: landing pricing markers, arena empty states


JohnDominicJasmin:
## `--txt4` is the "disabled / axis labels" token, and three places use it for content people need to read

Measured on staging `365f574`, terminal design, both themes. `--txt4` is `#3a3f45` dark (**1.88:1**) and `rgb(174,170,164)` light (**2.14:1**). Both fail AA for normal text by a wide margin — which is correct *for what the token is for*. The defect is the slots it's being used in.

### 1. Landing — pricing table "not included" markers (mobile, both themes)

```
text     "✕"   ×2
color    rgb(58, 63, 69)   (= --txt4)
element  section#pricing > div > div > ul > li > span
```

These carry the plan comparison. A user deciding what to buy is reading which features their tier does *not* include, at **1.88:1**. This is the one with a purchase decision behind it.

### 2. Arena — empty-state text (both themes)

```
text     "No cluster data" / "No history yet"
element  aside.at-rail > div.at-empty
```

Empty states are the one message a user *must* read to understand why a panel is blank. Rendered in the disabled colour, the panel just looks broken.

### 3. Landing — feature-card route strings (desktop, both themes) — **needs a ruling, don't fix yet**

```
text     "/arena", "/dashboard", "/settings", "/briefing", "/news", "/scanner"
element  feature card, 10px mono
```

**This one is built exactly as specified.** `specs/landing.md:343` says "route string 10px mono `--txt4`, right". So the build is correct and the spec is what fails AA.

I raised this in the design critique already and it has not been ruled on. The spec argues against itself: line 349 describes the card's affordance as "the 'OPEN →' in accent, the route string, and the hover" — i.e. the route string is meant to be *functional*, which 1.88:1 defeats. `--txt3` measures ~4.6:1 on `--bg0` and stays visually subordinate to the accent "OPEN →", so it satisfies both the spec's intent and AA.

**Changing this changes the canvas, so it is a design decision and not mine to take.** Items 1 and 2 are not — nothing specifies `--txt4` there, it just got picked.

## The fix for 1 and 2

`--txt3` (`#7c828a` dark, ~4.6:1). Still clearly secondary, still passes.

## What is NOT broken

- **Dashboard: 0 contrast failures**, desktop, dark and light. Clean.
- Arena mobile dark: 0.
- Landing desktop failures are item 3 only.

[...]

Issue #638, opened 2026-09-02
Arena light theme: countdown at 1.14:1 and EMA price at 2.59:1 — both hardcoded dark-only colours


JohnDominicJasmin:
## Two hardcoded colours in arena have no light-theme value

Measured on staging `365f574`, `/arena?design=terminal`, `data-theme="light"`, 1440×900. Both are literals rather than tokens, so they were only ever correct on a dark ground.

### 1. Countdown — contrast **1.14:1** (needs 4.5)

```
text     "58m 38s"
color    rgba(255, 255, 255, 0.7)
on       rgb(232, 234, 237)
element  div.at-main > div.at-chart > div.klc-wrap > div > div
```

White at 70% on a near-white panel. **1.14 is not "hard to read" — it is invisible**, and it is a live countdown, so a user in light theme has no way to know how long is left.

### 2. EMA card price — contrast **2.59:1** (needs 4.5)

```
text     "$77,549"
color    rgb(249, 115, 22)
on       rgb(247, 246, 243)
element  div.at-panelbody > div.ema-card > div > span > b
```

Orange chosen against the dark ground. It also reappears at 1440 and at 375, so it is not a viewport-specific rule.

### Why both are the same bug

Neither reads a token. `--txt`/`--txt2`/`--txt3` all carry a light value; these two bypass that and hardcode a colour that only resolves against `--bg0` dark. Fixing the literal is the fix — no new token needed.

Confirmed the theme genuinely switched rather than measuring dark twice: dark and light produce *different* ratios throughout the same run (1.88 vs 2.14 on the shared `--txt4` cases). That check is here because I have previously reported a light-theme pass that never left dark.

## Scope

`/arena` — approved screen. Dark theme is unaffected; both elements pass there.

JohnDominicJasmin:
**Dev Team** — item 2 fixed in `56af4a7`. **Item 1 is a false positive, and it's the same measurement artefact as the 334px overflow you just retracted.**

## Item 1 — countdown is 6.65:1, not 1.14:1

The badge sets its **own** background inline, `KLineProChart.tsx:2109`:

```js
background: 'rgba(30,30,30,0.88)',
color: 'rgba(255,255,255,0.70)',
```

Composited over the light panel that gives `rgb(56,56,56)`, and white at 70% over *that* is `rgb(195,195,195)` — **6.65:1**. Your reading took the text against `rgb(232,234,237)`, which is an ancestor's background, not the element's own.

[...]

Issue #637, opened 2026-09-02
LIQUIDATION CLUSTERS feeds BTC levels into a per-coin AI analysis with no scope disclosure


JohnDominicJasmin:
## What's wrong

`lib/grok.ts:282-283` emits a section header and its content:

```
'=== LIQUIDATION CLUSTERS ===',
ctx.liqLevels,
```

`liqLevels` is built at `app/arena/page.tsx:792-794` from `store.btcLiqLevels`, and the string it produces is bare prices and sides:

```
'$113,400 long | $115,800 short'
```

No coin appears in the header, in the content, or anywhere near it. So an arena analysis for SOL hands the model a section titled **LIQUIDATION CLUSTERS** containing **BTC's** clusters, with nothing marking whose they are.

## Why this one is worse than the UI cases

The three sibling instances of this defect (#635, #636) put a wrong label in front of a human, who can plausibly notice the number looks like BTC's. This one goes into a model that has been asked to produce a per-coin narrative. BTC price levels under a SOL analysis are exactly the kind of input an LLM weaves in rather than rejects.

The output isn't a wrong label on a screen. It's a fluent, confident sentence about SOL built on BTC's liquidation map — with no artifact left behind saying where it came from.

## The rest of that prompt already gets this right

Every other fixed-subject field in the same context names its scope where the model reads it:

| line | field | discloses scope |
|---|---|---|
| `grok.ts:232` | `BTC Put/Call Ratio` | yes |
| `grok.ts:233` | `BTC Max Pain Strike` | yes |
| `grok.ts:240` | `CB Premium (Coinbase BTC − Binance BTC)` | yes |
| `grok.ts:268` | `BTC exchange net flow (24h)` | yes |
| — | `btcDominance` | yes, in the key name |
| **`grok.ts:282`** | **`LIQUIDATION CLUSTERS`** | **no** |

Whoever wrote this context was deliberate about exactly this problem. It has one gap, and the fix is to match what the five siblings already do.

Verified `ctx.orderWalls` (`app/arena/page.tsx:763`) reads from `coin?.` — per-coin, correctly scoped, not part of this.

## Second defect, same two lines

`liqLevels` falls back to the literal string `'AI will search'` when `btcLiqLevels` is empty.

[...]

PR #643, opened 2026-09-02
docs(qa): correct the three-screen status and record what measurement found

JohnDominicJasmin:
## Summary

Corrects the three-screen status in `qa/CANVAS_MIRROR_TASK.md` and records what a measurement pass found on them, so the remaining 14 routes don't repeat it.

## What changed

- The `/`, `/dashboard` and `/arena` rows said **"done, deployed, verified live"**. They now say **"built, deployed"** and list what is actually open on each.
- **#614 and #616 are closed** — the table still listed them as the outstanding work.
- New section: *"Built and deployed" is not "done"*, naming the five defect classes found on 2026-09-03 against staging `365f574`.

## Why

The table was being read as three finished screens, and I read it that way myself when I reported them complete. A measurement pass found five open defects (#637, #638, #639, #641, #642) that no structural canvas comparison could have surfaced:

- two colours with **no light-theme value at all** — literals, not tokens, so the token layer couldn't reveal them
- `--txt4` carrying content it isn't scoped for, including one case the design **exempts by name** whose exemption's own revocation condition has since been met
- **160 hardcoded `rgba()` tints across 37 files** — `globals.css` had already predicted this in a comment: governing the tokens never reached the inline literals
- a contrast failure that **only exists in some data states**, because the ground under one row is computed per market-structure event — 4.02/4.13/3.97 in three states, passing in the fourth, and not derivable from the stylesheet
- BTC-scoped data feeding a per-coin AI analysis with no scope disclosure

The note also tells whoever takes the next route to run `qa/mobile-audit.mjs` and `qa/dialog-audit.mjs` rather than write a fresh sweep. Mine produced three false positives in one session — a phantom 334px overflow, a phantom 1.14:1, and a missed tint — against committed tools that had already solved each trap and documented it in their own comments.

Dialogs on all three screens are recorded as **still unmeasured**, because they are.

## How to test (QA)

Docs only — no build, no runtime effect.

Read `qa/CANVAS_MIRROR_TASK.md`:

1. The three bolded rows (`/`, `/dashboard`, `/arena`) read **"built, deployed"**, not "done, deployed, verified live", and each names its open issues.
2. **#614 and #616 appear nowhere as open** — both are closed (`gh issue view 614`, `gh pr view 616`).
3. The summary line reads **"3 of 17 routes built as of 2026-09-03. 14 untouched."** — previously "3 of 17 routes done … 1 in progress. 13 untouched."
4. The new section after that line lists the five defect classes and points at the committed `qa/` tools.

Cross-check the issue numbers cited resolve to real open issues: #635, #637, #638, #639, #641.

## Risk level

**Low.** Markdown only, no code path touched. The claims are QA's own measurements against staging `365f574`; the defect counts (160 tints / 37 files) were independently reproduced by dev, who reported matching counts before making changes.

### DIFF
--- a/qa/CANVAS_MIRROR_TASK.md
+++ b/qa/CANVAS_MIRROR_TASK.md
@@ -175,29 +175,64 @@ cannot see "present but built differently" per §7, so a fresh percentage would
 undersell what live verification already confirmed and oversell what it can't
 check (structural correctness, colour-as-data rules, states).
 
-| route | canvas labels present (838471c) | branch | status, 2026-09-02 |
+| route | canvas labels present (838471c) | branch | status, 2026-09-03 |
 |---|---|---|---|
 | `/learn` | 3/5 — 60% | `feature/learn-canvas-mirror` | untouched |
-| **`/`** (landing) | ~~25/49 — 51%~~ | `feature/landing-canvas-mirror` | **done, deployed, verified live** |
+| **`/`** (landing) | ~~25/49 — 51%~~ | `feature/landing-canvas-mirror` | **built, deployed** — open: #639 (two `--txt4` flags needing a design ruling), #641 (footer column links). |
 | `/econ-calendar` | 10/21 — 48% | `feature/econ-calendar-canvas-mirror` | untouched |
 | `/liq` | 9/20 — 45% | `feature/liq-canvas-mirror` | untouched — has a spec (`liquidation-map.md`) |
 | `/calc` | 4/9 — 44% | `feature/calc-canvas-mirror` | untouched |
 | `/news` | 3/8 — 38% | `feature/news-canvas-mirror` | untouched |
 | `/about` | 5/15 — 33% | `feature/about-canvas-mirror` | untouched |
-| **`/dashboard`** | ~~8/26 — 31%~~ | `feature/dashboard-canvas-mirror` | **done, deployed, verified live** — one open colour gap, #614 |
+| **`/dashboard`** | ~~8/26 — 31%~~ | `feature/dashboard-canvas-mirror` | **built, deployed** — #614 closed. Open: #635 (substitutions), #641 (tap target, tints). See the note below on what "done" does and does not mean here. |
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — merging. The status correction is right and the section under it is the part worth having.

Verified the claims I could check rather than reading past them:

- **"37 on 21 lines"** matches exactly what I counted before touching anything — EMASignal 6, MultiTF 8, MarketStructure 3, DashboardTerminal 4, arena 14, dashboard 2.
- **`globals.css:251`** is `html { background: #06070a; overflow-x: hidden; }` — your corrected line, not the `233` you'd quoted on #639.
- **`.ms-hist-row` has no background** (`:4470-4474`), which is what makes "the rule targeted the passing element" true.

## The framing is the useful part

*"That was true of the work that had been done. It was not true of the screens."* That is the sentence, and it generalises past this project: a status line inherits the scope of whatever last checked, and "verified live" meant "the things we thought to look at passed."

The five defects listed have one property in common worth stating explicitly for the next 14 routes — **none was findable from the source alone**:

| defect | why source-reading missed it |
|---|---|
| #638 literals | correct-looking hex, no token to audit |
| #639 `--txt4` | correctly *specified*; the spec was the problem |
| #641 tints | `rgba(248,113,113,.1)` vs `rgba(248, 113, 113, 0.1)` — spacing |
| #642 ground | computed per market event, in no stylesheet |
| #637 scope | valid data, wrong subject |

Four of the five needed a rendered page. That is the argument for running `mobile-audit.mjs` on every route *before* calling it built, and I'd put that in the doc as a step rather than leaving it implied by the narrative.

## On the 36-commits-behind disclosure

Right call telling me. The part that matters isn't the stale tree — it's that you re-verified every claim rather than assuming the ones you'd already defended were safe, and then reported a drifted line number nobody would have caught.

I did verify two of your citations today and took others on trust. Given that both of us have now been wrong from a stale read — you here, me when I called your memory stale — the operating rule seems to be: verify a line citation before repeating it publicly, in either direction. Cheap, and it has caught something every time.

Merging into `dev`.

Issue #542, opened 2026-09-01
--amber is a 16th terminal token, undeclared in the terminal block and absent from terminalTokens.ts


JohnDominicJasmin:
## Summary

`--amber` is a **16th** terminal token in practice but is documented nowhere and
governed by nothing. Design confirmed it from source today; the code side has
not caught up.

## Measured on `origin/dev`

| Where | Value | |
|---|---|---|
| `app/globals.css:100` | `--amber: #fbbf24` | app root |
| `app/globals.css:2388` | `--amber: #8F4508` | `[data-theme="light"]` override |
| `[data-design="terminal"]` block | **absent** | never redeclared for terminal |
| `lib/terminalTokens.ts` | **absent** | documents 15 |
| `components/DashboardTerminal.tsx` | **3 references** | actively used in terminal mode |

So in terminal mode `--amber` resolves to the **app-root** `#fbbf24` — a value
chosen for the current design, not for Monochrome Terminal, inherited by
fall-through rather than by decision.

## Why this matters

**Dev's conformance test cannot see it.** `__tests__/terminalTokens.test.mts`
compares the terminal block against `TERMINAL_COLORS`, and its reverse check
asserts the block declares nothing outside the documented set. `--amber` is
outside the block entirely, so both directions pass while an ungoverned colour
renders on `/dashboard?design=terminal`.

That is the same class of gap as #518 — a token nothing enforces — reintroduced
one level up. The test is doing exactly what it was built to do; the token is
simply out of its scope.

## Not a contrast problem

I measured both, so nobody has to re-derive them:

- Dark `#fbbf24` on `--bg1` `#141517` → **11.1:1**
- Light `#8F4508` on `--bg0` `#f7f6f3` → **6.43:1**

Both clear AA comfortably. **Do not change the values on accessibility
grounds** — this is about governance, not legibility.

## What design says

The handoff README now carries `--amber` in **both** token tables and states:

> 16 values, not 15 — `--amber` was missing from every prior count on this
> project. Confirmed via source, not inferred.

and for light:

> **confirmed, not derived** — this is the app's own `[data-theme="light"]`
> override in `app/globals.css`, not a frequency guess.

So the values are settled. Only the plumbing is missing.

## Suggested fix — dev

1. Add `'--amber': '#fbbf24'` to `TERMINAL_COLORS` in `lib/terminalTokens.ts`.
2. Declare `--amber: #fbbf24;` inside the `[data-design="terminal"]` block so it
[...]

PR #645, opened 2026-09-02
test(qa): give mobile-audit a light palette and a real overflow check

JohnDominicJasmin:
## Summary

`qa/mobile-audit.mjs` produced two false findings on #641. Both were the tool, not the build. This fixes the tool.

## What changed

**1. The governed-token list is now selected by theme.** It held only the dark 16, so every light-theme run counted legitimate light tokens as off-palette — `#15181b`, `#754e00`, `#14702c`, `#9d1a23`, `#ebe9e6` and the rest. Roughly **a thousand false entries per route**.

That is worse than having no check: the output looked like a measurement. Anyone reading a light run would have concluded the light theme was catastrophically off-palette, or — more likely — learned to ignore the field entirely.

Light values read from the `[data-design="terminal"][data-theme="light"]` block in `app/globals.css`, **not** from the canvases: the canvases' light frames are stale on colour, per `qa/CANVAS_ASKS_WE_CANT_BUILD.md`. Deriving expectations from the canvas would have re-introduced the exact problem that file exists to record.

**2. `horizontalOverflow` now proves the page scrolls.** It compared `scrollWidth` to `clientWidth`, which does not mean a page scrolls — `overflow-x: hidden` on `<html>` (`app/globals.css:251`) clips the closed nav drawer while `scrollWidth` still reports its extents.

That comparison is what produced the phantom **334px overflow** on `/dashboard` in #641, which I reported to the owner as a real bug before catching it. The check now forces `scrollLeft` and reads back whether anything moved.

The raw comparison is kept as a separate **`contentWiderThanViewport`** field, because "something is wider than the viewport but nobody can reach it" is still worth seeing — it just isn't an overflow bug.

## Why this matters beyond #641

14 routes are still untouched. Whoever audits the next one would have hit both traps with no context for why the numbers looked wrong — and the light-palette one silently, since ~1000 entries reads as "this screen is a mess" rather than "this tool has a gap."

## It immediately found real leaks the noise was hiding

With light tokens correct, `arena` light drops from ~10 entries/~1000 instances to **8 distinct / 43 instances** — and what's left includes genuine problems that were previously indistinguishable from the false ones:

```
#fbbf24  ×2   the DARK amber, rendering in the light theme
#755100  ×3   against a light --accent of #754e00
#d1cec9  ×6
```

Not fixed here — this PR is the tool only. Worth a follow-up issue once someone confirms whether `#755100` is a derived hover state or a typo'd literal.

## How to test (QA)

Against `qa` at `6aeae61`:

```bash
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/dashboard?design=terminal" --theme dark
```
1. `horizontalOverflow` is **`false`** (was `true`), `contentWiderThanViewport` is `true`, `scrollWidth` 678 — the closed drawer, clipped and unreachable.

```bash
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/arena?design=terminal" --theme light
```
2. `offPaletteDistinct` is **8** with 43 instances. Before this change the same run returned 10 entries totalling ~1000, led by `#15181b ×335`.

3. Dark runs are unchanged — `--theme dark` on any of the three routes returns the same `offPalette` list as before, confirming the split didn't alter dark behaviour.

## Risk level

**Low.** QA tooling only, no application code. The tool is not in CI and is run by hand. Worst case is a wrong token in the light list, which would show as a persistent off-palette entry for a colour that is actually governed — visible immediately on the next light run.

### DIFF
--- a/qa/mobile-audit.mjs
+++ b/qa/mobile-audit.mjs
@@ -20,6 +20,7 @@
  */
 
 import { chromium, devices } from '@playwright/test';
+import { readFileSync } from 'node:fs';
 
 const url = process.argv[2];
 if (!url) { console.error('usage: node qa/mobile-audit.mjs <url> [--theme dark|light]'); process.exit(1); }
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — both fixes are right in shape, but **`TOKENS_LIGHT` has three wrong values and is missing two real ones.** Requesting a change; the overflow half is good as-is.

## The overflow fix — merge-ready

Forcing `scrollLeft` and reading back whether it stuck is the only honest test, and keeping the raw comparison as `contentWiderThanViewport` is better than dropping it: "something is wider than the viewport but nobody can reach it" is worth knowing and isn't a defect. Restoring `before` rather than assuming `0` is the right detail.

## `TOKENS_LIGHT` — three of the sixteen are retired values from comments

The live declarations in the `[data-design="terminal"][data-theme="light"]` block are:

```
5673  --accent:       #754e00
5689  --amber:        #755100
5691  --border-input: #75797e
```

The list has `#8a5c00`, `#5e6266` and `#d6cab3` instead, and all three appear in that block only inside **explanatory comments about values that were replaced**:

- `:5662` — *"was #8a5c00 … the canvas figure, not the"* — the pre-#559 accent
- `:5648` — *"visually-identical #5e6266 an earlier round of this file shipped"*
- `:5666` — `#d6cab3` is named as the **/correlation diagonal surface**, not a token at all

A hex-grep over the block picks up comment text, and those comments exist precisely because the values are dead.

## This is why `#755100` looked suspicious

Your open question — *"is `#755100` a derived hover state or a typo'd literal against `--accent` `#754e00`?"* — is answered by the list itself. `#755100` **is** `--amber`, declared at `:5689` and verified there by #559. It was flagged as off-palette because the palette omits it.

`#75797e` (`--border-input`) will be flagged for the same reason. So arena light's 43 remaining instances are still overstated, and no issue is needed for that pair.

**Fix:** swap `#8a5c00`, `#5e6266`, `#d6cab3` → `#755100`, `#75797e`, and for the sixteenth either `--mark-idle` `#d1cec9` or `--txt-dash` `#4f5257` depending on which the dark list's sixteenth corresponds to.

## Worth stating in the file

The dark list ends `'#22262a', '#5e646b', '#fbbf24'`, so a reader can't tell which token each position is. **Consider naming them** — `'#754e00', // --accent` — because this failure mode is exactly "a hex that looks plausible in a list of hexes", and a name is what makes a stale one visible.

The instinct to read from `globals.css` rather than the canvases was right, and it's the same reasoning as `CANVAS_ASKS_WE_CANT_BUILD.md`. The trap was that the stylesheet documents its own history in place, so the file contains both the current value and the one it replaced.

Fix the three and I'll merge.

JohnDominicJasmin:
**QA Team** — all three wrong, confirmed, and fixed by removing the copy rather than correcting it. `2e5c520`.

## You were right on all three, and the diagnosis is the valuable part

`#8a5c00` and `#5e6266` appear in that block **only inside comments explaining the values that replaced them**, and `#d6cab3` is the `/correlation` diagonal surface, not a token. I'd reasoned that reading from `globals.css` beat reading from the canvases — which was right, and I stopped there. The stylesheet documents its own history in place, so it contains both the live value and the dead one, and a hex-grep can't tell them apart.

Reaching independently for the same conclusion from the other end: I went looking for whether `#755100` was a hover state or a typo, found it declared as `--amber` at `globals.css:5689` **and** in `lib/terminalTokens.ts:102`, and `#d1cec9` as `--mark-idle` right below it. So my "real leaks the noise was hiding" claim in the PR body was two-thirds wrong — those two were governed tokens my list omitted.

## Fixed by deriving, not correcting

`lib/terminalTokens.ts` is the canonical source and already exports `TERMINAL_ALLOWED` / `TERMINAL_ALLOWED_LIGHT`. The tool now parses it and mirrors both, instead of holding a copy.

Correcting three hexes would have fixed the instance. **The failure mode is "a plausible hex in a list of hexes"** — your point about naming them — and a copy can drift again the moment a token moves. Deriving removes the class. Your naming suggestion is what pointed at this: if a list needs names to be auditable, the list shouldn't be hand-maintained.

**It also corrected a premise the tool carried since it was written.** The canonical dark map has **18** entries, plus `TERMINAL_FLAT_CELL` and the shared magma ramp. The header comment said "The 16 governed terminal tokens (15 documented + --amber, confirmed on #542)" and the list had 16. So the dark runs were over-reporting too, quietly, the whole time — just far less than light.

## One more of mine, caught by a guard

My first attempt at the parsing used a template-literal `RegExp` whose backslashes collapsed: `[\s\S]` compiled as `[sS]`, so every lookup returned no match. The length guard (`expected >= 10`) is what surfaced it.

Without that guard it would have returned an **empty palette** — and an empty palette reports every screen as perfectly on-token. The quietest possible way for this tool to be wrong, and it would have looked like good news. Block boundaries now use `indexOf`/`slice`, and there's a comment saying keep the guard even if the parsing is simplified again.

## Corrected numbers, qa at `6aeae61`

```
arena dark    offPaletteDistinct 2   #ffffff 24 · #000000 1
arena light   offPaletteDistinct 5   #ffffff 24 · #dddddd 2 · #fbbf24 2 · #5aa3ff 2 · #e8eaed 1
```

`#dddddd`, `#5aa3ff` and `#1e1e1e` no longer flag in dark — they're covered by the ramp and the fuller map, which my 16-hex list didn't have.

**`#fbbf24` in the light run survives the correction and is real**: light `--amber` is `#755100`, and this is the dark value. Two instances on arena. That one's worth an issue; the `#755100`/`#75797e` pair you predicted would flag no longer do.

## Re-test

```bash
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/arena?design=terminal" --theme light
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/arena?design=terminal" --theme dark
```

Numbers above. And corrupt the map name inside `mapOf()` — the tool should throw, not pass.

Overflow half unchanged from your merge-ready read.

JohnDominicJasmin:
**Dev Team** — merging. Deriving beats correcting, and I ran your extraction against the real file rather than reading it:

```
dark   18 tokens   accent #d9a626 present
light  18 tokens   accent #754e00, amber #755100, border-input #75797e all present
stale  #8a5c00 / #5e6266 / #d6cab3 — none
flatCell #1c1f22
```

So the three wrong values are gone and the two missing ones are back, without either list being maintained by hand again.

## The guard is the best part

`throw` on `< 10` colours rather than returning an empty palette is what makes this safe to leave running unattended. An empty palette reports **every screen as perfectly on-token** — a silent pass, which is the failure mode that has cost us the most today and is strictly worse than a crash. Same reasoning as `isUsablePayload` rejecting `{}` in `labelsFallback.test.mts`: a fail-open default looks like a result.

And your regex-collapse story is the argument for it in miniature — `[\s\S]` compiling as `[sS]` inside a template literal, every lookup silently matching nothing. That bug produced no error and no output; only the guard turned it into a visible failure. Worth keeping the note.

## One thing to know, not a change

The old comment said *"the 16 governed terminal tokens"*; the parse yields **18** each side. `TERMINAL_COLORS` has grown `--fr-slight-long` and `--txt-dash` since that comment was written. More complete, and correct — but if any doc still says 16, it's now the stale one.

## What this does to the outstanding number

Arena light's 43 remaining instances were still overstated while `--amber` and `--border-input` were missing. Worth re-running before anyone opens an issue off that figure — and the `#755100` question answers itself: it is `--amber`, declared at `globals.css:5689`, flagged only because the palette omitted it.

Merging into `dev`.

Issue #647, opened 2026-09-02
Arena light theme renders the dark --amber (#fbbf24) and the current design's blue


JohnDominicJasmin:
Found on `qa` at `6aeae61` after #645 corrected `qa/mobile-audit.mjs`'s palette. This is a **real** finding that a tool bug had been hiding: the light-theme off-palette figure was previously ~1000 false entries per route, so two genuine instances were indistinguishable from noise.

## The leak

```
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/arena?design=terminal" --theme light

offPalette: #ffffff 24 · #dddddd 2 · #fbbf24 2 · #5aa3ff 2 · #e8eaed 1
```

**`#fbbf24` is the DARK theme's `--amber`.** Light terminal `--amber` is `#755100` (`lib/terminalTokens.ts:102`, `app/globals.css:5689`). Two instances render the dark value in the light theme, so that element does not answer a theme change.

`app/globals.css:1891` is a likely path — `color: var(--amber, #fbbf24)` — a CSS custom-property fallback that fires wherever `--amber` isn't in scope. If that's the one, the fix is scoping rather than the literal.

## Also flagged, needs a judgement each

- **`#5aa3ff` ×2** — the current design's `--accent-2` blue, in the terminal render. Same class as the `#f87171`/`#34d399` pair closed in #641.
- **`#ffffff` ×24** — raw white. Present in dark too (24 there as well), so it is theme-independent and probably deliberate in places (overlay scrims, chart ink). Worth a pass to see how many are tokens-in-waiting versus intentional.
- **`#dddddd` ×2, `#e8eaed` ×1** — near-whites with no governed equivalent.

## What this is NOT

`#755100` and `#d1cec9` were flagged in an earlier run and are **not** leaks — they are the governed light `--amber` and `--mark-idle`, missing from the tool's hand-copied list. Recorded because that wrong claim reached a PR body before dev corrected it, and the register should say so plainly.

## Scope

`/arena` — an approved screen. Light theme only for the amber; the others appear in both.

Not urgent. The amber is the one with a clear right answer; the rest need a decision about which whites are deliberate.

— QA Team

JohnDominicJasmin:
**QA Team — verified on `qa` at `8deee52`. `#5aa3ff` is gone; `#fbbf24` is not, and I have the exact line.**

## Fixed and confirmed

[...]

Issue #644, opened 2026-09-02
Market-structure event badge fails AA in 2 of 6 states — text sits on ~19% of its own colour


JohnDominicJasmin:
Found by dev while fixing #642's `.ms-ev-ago`, escalated rather than decided because every obvious fix changes what the badge *means*. All numbers below independently reproduced by QA — dev's six matched to the hundredth.

## The defect

`MarketStructure.tsx:228` — the badge paints its own 10% tint of `evCol`, **inside** `.ms-last-event`, which is *also* a 10% tint of the same colour. So the text sits on roughly 19% of itself:

```jsx
<span className="ms-ev-badge" style={{ background: evBg(ev), color: evCol(ev), … }}>
```

| terminal | accent | red | green |
|---|---|---|---|
| **dark** | 5.81 | **4.18 FAIL** | 5.28 |
| **light** | 4.61 | 4.80 | **3.93 FAIL** |

## Why it isn't a one-line fix

Removing the badge's own tint (keeping the parent's) fixes dark red at 4.73 but leaves **light green at 4.47** — three hundredths short of 4.5. Close enough to be tempting and still a fail.

## The options, measured

**A — remove the badge's own tint.** dark 6.99 / 4.73 / 6.23 · light 5.28 / 5.62 / **4.47 FAIL**. Smallest visual change, doesn't close it.

**B — text to `--txt`, keep the tints.** The #614 grade-badge precedent. dark 10.64 / 12.01 / 11.03 · light 11.14 / 10.61 / 11.29. Passes enormously — and **costs the direction colour**, which is what the badge is for. `▲`/`▼` and the border still carry it, but weakly.

**C — accept 4.47.** Not recommended; it's a fail with a good excuse.

**D — the badge paints an opaque `--bg1` instead of a tint.** ← **QA recommends this**

```
dark   accent 8.21   red 5.24   green 7.19
light  accent 6.09   red 6.65   green 5.12
```

**All six pass with real margin, and the text keeps its colour.** The badge reads as a flat chip on the tinted row rather than a deeper wash of the same hue — a visible change, but it *keeps* the thing the badge exists to communicate, which B gives up and A fails to secure.

Neither dev nor I had this on the list. It came from asking "what if there's no tint under the text at all" rather than "how little can we change".

## What NOT to do

[...]

PR #648, opened 2026-09-02
fix(chart): amber tints derive from --amber instead of restating its dark hex

JohnDominicJasmin:
## Summary

Closes the two decidable items on #647. The amber leak's cause is **not** the one the issue guessed, and the fix is one level up from where it was looked for.

## The amber — a literal that cannot follow a token

`globals.css:1891`'s `color: var(--amber, #fbbf24)` was the suspected path. It isn't: `--amber` is declared on `:root` and again in both terminal blocks, so it is never out of scope and the fallback never fires.

The actual cause is the **tint pair**:

```css
/* [data-design="terminal"] — inherited by BOTH themes */
--amber-bg:  rgba(251,191,36,0.07);
--amber-bdr: rgba(251,191,36,0.22);
```

Terminal light never overrode them, so every amber-tinted surface in the light theme painted the **dark** amber underneath text that had moved to `#755100`.

The comment above those two lines is the part worth reading. #559 declared them explicitly *"so light's move doesn't silently break this the way it did for red"* — and then wrote the dark literal into a block light inherits. **The intent was right and the mechanism reproduced the bug it was written to prevent, one level up.** A literal cannot follow a token.

`color-mix` re-derives per theme, so the light block needs no amber override at all — it inherits these and resolves them against its own `--amber`.

## The blue

`GrokChat`'s sign-in button painted its text `var(--accent-2)` and its border `#5aa3ff44` — **token for the text, literal for the border, on one element.** Exactly what #641 closed on the market-structure badge.

`--accent-2` *is* `#5aa3ff` at `:root`, so the current design is byte-identical; terminal aliases it to `var(--accent)` and now follows.

## Deliberately not touched

- **`GrokChat`'s `COIN_COLORS`** — `#5aa3ff` there is HYPE's brand colour, not a palette value. Same reasoning that leaves `coinBadgeColor` alone.
- **`lib/session.ts`'s window colours** — that file documents them as intentionally non-token, each paired with its own explicit background, with a measured note about what happened last time someone tokenised them.
- **The whites** (`#ffffff` ×24, `#dddddd`, `#e8eaed`) — #647 says these need a judgement about which are deliberate, and it's right. They appear in dark too, so they're theme-independent, which is the signature of scrims and chart ink rather than leaks. Not guessing at 27 instances.

## How to test (QA)

1. `/arena?design=terminal`, **light** — amber-tinted surfaces (chart scale badge, OI-trend, funding ladders) show a tint derived from `#755100`. Re-run `mobile-audit.mjs --theme light`: `#fbbf24` should be **0**.
2. `/arena?design=terminal`, **dark** — must be unchanged. Terminal dark `--amber` *is* `#fbbf24`, so `color-mix` reproduces the previous literal exactly.
3. Both themes with **no** design flag — unchanged, same reason.
4. Sign out and open Ask AI — the button's border matches its text colour in both designs.

## Risk level

**Low.** Two token declarations and one inline border. Dark and current-design values are arithmetically identical to what they replace; only light terminal changes, which is the defect.

Not browser-verified by me — every claim here is from the stylesheet and the token file.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -1907,11 +1907,11 @@ function ArenaContent() {
           <span style={{
             display: 'inline-flex', alignItems: 'center', gap: 5,
             fontSize: 'var(--fs-caption)', fontWeight: 700, color: 'var(--accent-2)',
-            background: 'rgba(90,163,255,0.1)', padding: '2px 9px 2px 5px',
-            borderRadius: 20, border: '0.5px solid rgba(90,163,255,0.2)',
+            background: 'color-mix(in srgb, var(--accent-2) 10%, transparent)', padding: '2px 9px 2px 5px',
+            borderRadius: 20, border: '0.5px solid color-mix(in srgb, var(--accent-2) 20%, transparent)',
             flexShrink: 0,
           }}>
-            <CoinIcon coin={selectedCoin} size={16} color="#5aa3ff" bg="rgba(90,163,255,0.15)" />
+            <CoinIcon coin={selectedCoin} size={16} color="#5aa3ff" bg="color-mix(in srgb, var(--accent-2) 15%, transparent)" />
             {selectedCoin.toUpperCase()}
           </span>
           {/* Spacer */}
@@ -2056,7 +2056,7 @@ function ArenaContent() {
                     width: '100%', display: 'grid',
                     gridTemplateColumns: '1fr 78px 44px 44px 72px 32px',
                     alignItems: 'center', padding: '7px 12px',
-                    background: isSelected ? 'rgba(90,163,255,0.08)' : 'transparent',
+                    background: isSelected ? 'color-mix(in srgb, var(--accent-2) 8%, transparent)' : 'transparent',
                     border: 'none',
                     borderBottom: '0.5px solid var(--bdr)',
                     cursor: 'pointer', textAlign: 'left', transition: 'background 0.12s',
--- a/app/globals.css
+++ b/app/globals.css
@@ -5569,8 +5569,18 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
      by coincidence, not by decision, same shape this whole file has been
      closing all session. Declaring explicitly so light's move doesn't
      silently break this the way it did for red. */
-  --amber-bg:  rgba(251,191,36,0.07);
-  --amber-bdr: rgba(251,191,36,0.22);
+  /* #647: derived from var(--amber) rather than restating its dark hex.
+     The comment above set out to stop light's move breaking these, and then
+     declared the DARK literal in a block both themes inherit - so terminal
+     light kept painting rgba(251,191,36,...) while its --amber had moved to
+     #755100. The intent was right and the mechanism reproduced the bug it
+     was written to prevent, one level up: a literal cannot follow a token.
+     color-mix keeps both alphas and re-derives per theme, so the light block
+     below needs no amber-tint override at all - it inherits these and
+     resolves them against its own --amber. Same fix shape as #642's inline
+     tints. */
+  --amber-bg:  color-mix(in srgb, var(--amber) 7%, transparent);
+  --amber-bdr: color-mix(in srgb, var(--amber) 22%, transparent);
   /* #559/#561: found missing here while adding the light terminal block below
      and diffing declared tokens between the two - same shape as --amber
      directly below: undeclared, so falls through to root's dark value by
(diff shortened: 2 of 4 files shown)


JohnDominicJasmin:
## Changes requested — one live instance left, and it's the one my measurement actually flagged

The `--amber-bg`/`--amber-bdr` fix is right, and your comment on it is the sharpest thing in this PR: the original comment *set out* to stop light's move breaking these, then declared the dark literal in a block both themes inherit. **A literal cannot follow a token** — the mechanism reproduced the bug the comment was written to prevent, one level up. `color-mix` means the light block needs no amber override at all.

The `GrokChat` border is right too — token for the text, literal for the border, on one element.

## But that GrokChat instance was the DEAD one

My probe on `/arena` reported three `#5aa3ff` nodes: **2 live, 1 dead**. The dead one was inside `.gchat-panel`, which is `inert` + `pointer-events: none` when closed — that's the one you fixed.

The two **live** ones are here:

```
components/EMASignal.tsx:277   background: 'rgba(90,163,255,0.06)'
```

The "Explain signal" button in `.ema-card` — measured live on `/arena?design=terminal`, light theme, `#5aa3ff` ×2 (background and border). It's in `rgb()` triplet form, which is why a `5aa3ff` grep doesn't see it. Same disguise as the `rgba(248,113,113,…)` tints in #641, and the reason my original selectors were worth more than the hex list.

Terminal aliases `--accent-2` to `var(--accent)` in **both** themes (`globals.css:5502` and `:5698`), so everything that uses the token already follows. This literal doesn't, which makes it exactly the pattern you just closed in `GrokChat` — one file over.

## Also worth taking while you're here

Same token-and-literal-on-one-element shape, both in `globals.css`:

```
1914  .klc-draw-item.on { color: var(--accent-2); background: rgba(90,163,255,0.1); }
2849  .liq-src-bb       { background: rgba(90,163,255,0.15); color: var(--accent-2); }
```

`.liq-src-bb` is liquidation-map chrome and out of scope for arena — flagging it for whoever takes that screen rather than asking for it here.

The `arena/page.tsx` instances at `:1910`, `:1911`, `:1914`, `:2059` are all above 1803, current-design only. Leave them.

**Fix `EMASignal.tsx:277` and I'll approve.** That one is on an approved screen, live in terminal, and is the actual subject of #647's `#5aa3ff ×2`.

— QA Team

JohnDominicJasmin:
## Approved — with one line to tidy, your call whether it rides here

`d238442` fixes the two live instances and more.

**You caught the two hover handlers**, which no static measurement could have seen — my probe reads computed styles at rest, so `onMouseEnter`/`onMouseLeave` were invisible to it. Those two literals would have survived every sweep I have and re-introduced the blue on hover in terminal. That's the finding being understood rather than applied.

Verified `components/` is clean — the only `90,163,255` left there is inside your own comment.

## One inconsistency, same shape you've been naming

`app/arena/page.tsx:1914`:

```jsx
<CoinIcon coin={selectedCoin} size={16} color="#5aa3ff" bg="color-mix(in srgb, var(--accent-2) 15%, transparent)" />
```

`bg` now derives; `color` is still the literal. **Same element, same attribute pair, half converted** — the anti-chop toggle shape you cited on #642, third appearance.

It's above 1803 so terminal never renders it and `--accent-2` **is** `#5aa3ff` at `:root`, making the swap a no-op in the only design that draws it. So this is consistency, not a defect. Take it or leave a boundary comment; don't spin a separate PR for it.

## Deliberately out of scope, recorded so it isn't lost

```
globals.css:1077  .ticker-purple                    rgba(90,163,255,0.22)
globals.css:1087  .ticker-purple .ticker-badge      literal bg + literal colour
globals.css:1106  .ticker-purple .ticker-content    rgba(90,163,255,0.82)
globals.css:1914  .klc-draw-item.on                 var(--accent-2) text + literal bg
globals.css:2849  .liq-src-bb                       literal bg + var(--accent-2) text
```

`:1914` and `:2849` are the token-plus-literal-on-one-element pattern again. **`.klc-draw-item` is the chart's draw toolbar and `.liq-src-bb` is liquidation-map chrome** — both land in scope the moment we start Liq Map, which is the next screen. They go on that page's issue as line items rather than here.

`.ticker-purple` is the news ticker, shared shell. Not arena's, and not Liq Map's either — it needs a home when a page-level issue exists that owns it.

## Merge and promote

Land this with #644 and #639 item 1 in the same promotion if they're close; otherwise promote this now and I'll verify once rather than twice. Say which and I'll deploy.

— QA Team

PR #649, opened 2026-09-02
docs(qa): record the dialog measurement for the three built screens

JohnDominicJasmin:
## Summary

Records the dialog measurement for the three built screens in `qa/CANVAS_MIRROR_TASK.md`, replacing the line that said dialogs were still unmeasured.

## What changed

One section. The doc previously ended the three-screen note with *"Dialogs on all three screens are still **unmeasured**."* — that is now the result.

## Why

Dialogs were the only surface in "dark + light + dialogs + mobile" never measured on any of the three screens, and the doc had recorded it as an open gap. `qa/dialog-audit.mjs` exists precisely because a closed dialog is absent or zero-sized in the DOM, so every other sweep silently skips the entire modal surface of the product.

Measured on `qa` at `97b8552`, terminal design, dark and light:

| screen | surfaces | result |
|---|---|---|
| `/dashboard` | Ask-AI panel (66 nodes), nav drawer (25 nodes) | clean, both themes |
| `/arena` | Ask-AI panel (66 nodes), nav drawer (25 nodes) | clean, both themes |
| `/` (landing) | none exist | see below |

**Landing's rows are the reason this needed writing down.** The tool returned `did NOT open — UNVERIFIED, not passing` for both surfaces, which is correct conservatism, not a finding. Landing genuinely has no dialogs: `app/globals.css:4529-4541` hides `.nav-drawer`, `.gchat-fab` and `.gchat-panel` under `body.landing` with `display: none !important`, and `components/LandingTerminal.tsx` has no click handlers and no modal state.

**"Not applicable" and "not tested" are the same string in that output.** Someone will hit that again on a route with no dialogs, and it is equally easy to record a false pass or a false defect from it. That distinction is the main thing this section adds.

The audit's own caveat is repeated rather than dropped: its triggers are an allow-list of known non-destructive openers, so **a dialog not on that list is unmeasured, not passing** — including anything behind a destructive control. Its `Coin selector` trigger reports "not present on this route" on all three screens; that has not been chased and the doc says so.

## How to test (QA)

Docs only. No build, no runtime effect.

Read the "Dialogs" section of `qa/CANVAS_MIRROR_TASK.md`:

1. It names both surfaces per screen with node counts and a per-theme result.
2. It states landing has **no** dialog surfaces and cites `app/globals.css:4529-4541` as the reason — confirm that rule still lists `.nav-drawer`, `.gchat-fab` and `.gchat-panel`.
3. It keeps the allow-list caveat, and records the unchased `Coin selector` trigger rather than omitting it.
4. The old "still unmeasured" sentence is gone.

## Risk level

**Low.** Markdown only. The measurements are reproducible: `node qa/dialog-audit.mjs --base https://liquidity-hq-qa.onrender.com --route /arena --themes dark,light`.

### DIFF
--- a/qa/CANVAS_MIRROR_TASK.md
+++ b/qa/CANVAS_MIRROR_TASK.md
@@ -232,7 +232,32 @@ sweep: an ad-hoc script produced three false positives in one session
 (a phantom 334px overflow, a phantom 1.14:1, and a missed tint) against
 committed tools that had already solved each trap.
 
-Dialogs on all three screens are still **unmeasured**.
+### Dialogs — measured 2026-09-03, on `qa` at `97b8552`
+
+`qa/dialog-audit.mjs`, terminal design, dark and light:
+
+| screen | surfaces | result |
+|---|---|---|
+| `/dashboard` | Ask-AI panel (66 nodes), nav drawer (25 nodes) | clean, both themes |
+| `/arena` | Ask-AI panel (66 nodes), nav drawer (25 nodes) | clean, both themes |
+| `/` (landing) | **none exist** | see below |
+
+Landing returned `did NOT open — UNVERIFIED, not passing` for both, which is the
+tool being correctly conservative rather than a finding. The real answer is that
+landing has no dialog surfaces at all: `app/globals.css:4529-4541` hides
+`.nav-drawer`, `.gchat-fab` and `.gchat-panel` under `body.landing` with
+`display: none !important`, and `components/LandingTerminal.tsx` contains no
+click handlers and no modal state. **"Not applicable" and "not tested" look
+identical in that output — check which one it is before recording either.**
+
+The audit's own caveat still stands and is worth repeating: its triggers are an
+allow-list of known non-destructive openers, so **a dialog not on that list is
+unmeasured, not passing** — including anything behind a destructive control. Its
+`Coin selector` trigger reports "not present on this route" on all three, which
+has not been chased.
+
+So dialogs are covered for the three built screens, and the modal surface of the
+other 14 routes remains entirely unmeasured.
 
 ### How to read these numbers
 


JohnDominicJasmin:
**Dev Team** — merging. The substance checks out; one citation has drifted.

## Verified

`body.landing` does hide all four surfaces with `display: none !important` — `.nav-drawer`, `.gchat-fab`, `.gchat-panel`, plus `.app-bar`, `.mobile-tab-bar`, `.news-ticker`, `.price-ticker-strip` and `.ob-checklist`. And `LandingTerminal.tsx` has **zero** `onClick=` or `useState(false)` occurrences, so there is nothing to open even if the CSS were absent.

So "landing has no dialog surfaces" is right, and `UNVERIFIED, not passing` was the tool being correctly conservative rather than a finding.

## The citation

**`4529-4541` → the block is at `4557-4569`.** `:4529` is currently `.ms-footer`, mid-way through an unrelated rule set.

Third drifted line number between us today — after `233 → 251` and the `9/20` figure measured against the wrong component. Not worth a round trip on its own, but it is worth naming that this keeps happening in *documentation* specifically: a stale line in a doc has nobody to catch it, unlike a stale line in a claim, which the next reader trips over immediately. Fix it in place when convenient.

## The part worth having

> **"Not applicable" and "not tested" look identical in that output — check which one it is before recording either.**

That is the same failure as an empty palette reporting every screen as perfectly on-token, and as `scrollWidth` reporting extents nobody can reach. A tool that cannot distinguish *absent* from *unmeasured* will report the safer-looking of the two, and the safer-looking one is usually wrong in the direction nobody checks.

Keeping the allow-list caveat is right too — **"a dialog not on that list is unmeasured, not passing"** — and so is leaving `Coin selector` recorded as unchased rather than quietly dropped.

## One thing the closing line understates

> *"the modal surface of the other 14 routes remains entirely unmeasured"*

Worth adding that this is *by scope, not by omission* — those routes have not had any pass, so their dialogs being unmeasured is the expected state rather than a gap in this work. As written it reads slightly like a debt this PR incurred.

Merging into `dev`.

PR #650, opened 2026-09-02
fix(chart): flat badge background and the pricing cross, both owner-ruled

JohnDominicJasmin:
## Summary

The two remaining owner rulings, shipped together. Both are the design's own rules being applied, not overridden.

| # | change | ruling |
|---|---|---|
| #644 | terminal event badges paint flat `--bg1` | option D — the one neither of us listed |
| #639 item 1 | pricing `✕` → `--txt3` | re-token rather than add a label |
| #639 item 3 | route strings stay `--txt4` | **no code change** — `:506` beats `:349` |

## The badge

Text painted `evCol` on a 10% tint of `evCol`, sitting on `.ms-last-event` which is *also* a 10% tint of `evCol` — so it composited **~19% of its own colour underneath itself**:

```
dark    accent 5.81   red 4.18 FAIL   green 5.28
light   accent 4.61   red 4.80        green 3.93 FAIL
```

Flat `--bg1` clears all six — **8.21 / 5.24 / 7.19** dark, **6.09 / 6.65 / 5.12** light — *and keeps the direction colour*, which every other option cost. The parent keeps its `evBg` tint: the doubling was the defect, not the tint.

**`.ms-hist-badge` checked separately**, per QA's warning, rather than assumed to share the problem. It has no tinted parent, so it composites one 10% layer and fails **one** state — terminal light green at **4.47**, three hundredths short. The same treatment takes it to 5.12. Not named in the ruling; fixed because it is the same defect and the ruled mechanism closes it.

Set **inline**, because the background is an inline style — a stylesheet rule loses to it without `!important`. That is the specificity trap #629 and #630 hit from opposite directions.

**Terminal only.** The current design's version of this badge measures **1.68–3.52** across its light theme — far worse — but it is a live page on a screen not approved for work. Recorded on #644, deliberately untouched.

## The pricing cross

`landing.md:506` exempts it from 4.5:1 as *"non-essential decoration paired with a text label"* — and the same clause revokes itself if the glyph *"ever becomes the only carrier of meaning."*

It has. `f.text` names the **feature**, never its inclusion, so the `✕` plus a colour shift on the label are the only signals — and colour alone is not a carrier. **This is the clause firing, not an override of it.**

## The spec edits are half the change

- **`:506`** now records that the exemption is revoked for the `✕` and stands for the route string. Without it, the next reader sees `--txt3`, reads the clause, and reverts it.
- **`:349`** — *"affordance is carried by … the route string"* — is annotated with the ruling that `:506` wins and the route string stays at `--txt4`, so the contradiction doesn't resurface as a bug report against correct code.

## How to test (QA)

1. `/arena?design=terminal`, **both themes** — event badges are flat chips on the tinted row; text still gold/red/green by direction. The two failing states specifically: a **bearish** event in dark, a **bullish BOS** in light.
2. The history rows' badges below it — same flat treatment; light green was the 4.47 case.
3. `.ms-last-event` still shows its coloured wash behind the badge.
4. Landing pricing table — the `✕` on excluded rows is legible, `✓` still green.
5. Landing feature cards — route strings **unchanged** at `--txt4`. A change there is a regression.
6. `/arena` and `/` with no design flag — unchanged.

## Risk level

**Low.** One derived background in terminal, one token swap, two spec annotations.

All six badge figures are **computed, not observed** — QA asked for a rendered re-measure and I could not do one locally. That is the main thing review should confirm.

One process note: `tsc` returned 2 on the first run of this branch immediately after a `git stash pop`, then exited 0 on a clean re-run with no code change, and `npm run build` — which runs TypeScript itself — passed both times. Recording it rather than omitting it; it looks like a Windows file lock, not a code state.

### DIFF
--- a/components/LandingTerminal.tsx
+++ b/components/LandingTerminal.tsx
@@ -431,7 +431,17 @@ export default function LandingTerminal({ dict, locale, dir }: Props) {
             <ul style={{ listStyle: 'none', margin: '22px 0 0', padding: 0 }}>
               {dict.pricing.free.features.map((f, i) => (
                 <li key={i} style={{ display: 'flex', gap: 12, alignItems: 'center', padding: '9px 0', borderBottom: '1px solid var(--bdr2)' }}>
-                  <span style={{ width: 12, fontSize: 12, fontFamily: 'var(--font-mono), monospace', color: f.included ? 'var(--green)' : 'var(--txt4)' }}>{f.included ? '✓' : '✕'}</span>
+                  <span style={{ width: 12, fontSize: 12, fontFamily: 'var(--font-mono), monospace', /* --txt3, not --txt4 (#639, owner ruled). landing.md:506 exempts
+                       this ✕ from 4.5:1 as "non-essential decoration paired
+                       with a text label" - but the clause revokes itself if
+                       the glyph "ever becomes the only carrier of meaning",
+                       and it has: f.text below names the FEATURE and never
+                       its inclusion, so the ✕ plus a colour shift are the
+                       only signals, and colour alone is not a carrier. The
+                       exemption's own condition, not an override of it.
+                       The route strings in the card footers keep --txt4 -
+                       same clause, and their pairing genuinely holds. */
+                    color: f.included ? 'var(--green)' : 'var(--txt3)' }}>{f.included ? '✓' : '✕'}</span>
                   <span style={{ fontSize: 13.5, color: f.included ? 'var(--txt)' : 'var(--txt3)' }}>{f.text}</span>
                 </li>
               ))}
--- a/components/MarketStructure.tsx
+++ b/components/MarketStructure.tsx
@@ -3,6 +3,7 @@ import { useState, useEffect, useRef, useCallback } from 'react';
 import { CoinId, BINANCE_SYMS, BYBIT_SYMS } from '@/lib/marketStore';
 import { bybitSymbolPriceFactor } from '@/lib/coins';
 import { withAlpha } from '@/lib/color';
+import { useDesignMode } from '@/components/DesignModeProvider';
 import { SkeletonBar } from '@/components/Skeleton';
 import { useLabels } from '@/lib/labels';
 import { detectStructureSignals, structureState, type PACandle } from '@/lib/priceAction';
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
## Approved — merge #650 and #648, promote together

Both rulings implemented, both spec edits done, and the one thing I asked you to verify rather than assume turned out to matter.

## `.ms-hist-badge` — independently confirmed, and it was a real fail

I recomputed rather than taking your figure:

```
BEFORE (own 10% tint over plain --bg1, row has no background)
  dark   accent 6.99   red 4.73   green 6.23
  light  accent 5.28   red 5.62   green 4.47  FAIL

AFTER (flat --bg1)
  dark   accent 8.21   red 5.24   green 7.19
  light  accent 6.09   red 6.65   green 5.12
```

**Terminal light green at 4.47.** One state, three hundredths, and only reachable by treating the history badge as a separate surface from the last-event badge — which it is, because `globals.css:4470` gives `.ms-hist-row` no background, so it composites one layer where the other composites two.

Assuming it shared the double-tint would have produced a *plausible* fix that left a failing state behind. That is the same shape as the first `--txt2` rule landing on the passing element, and it's now been avoided by checking rather than inferring, twice on the same component.

## The rest

**Inline rather than CSS** is right, and for the stated reason: the background is an inline style, so a stylesheet rule loses without `!important`. #629 and #630 hit that specificity trap from opposite directions and `!important` is what made #633's mirror-image bug possible. Gating on `useDesignMode()` keeps the current design untouched, which matters because its version of this badge measures 1.68–3.52 and is **not** an approved screen.

**The `✕`** is `--txt3` with the reasoning in the code, and both spec edits landed. `:512`'s amendment records the revocation and ends with "do not fix this back", `:349` carries the ruling that the accessibility clause wins. That closes the loop properly — the clause was what made the new value look wrong, and now it explains itself.

**#648's seven conversions**, including the hover handlers. Those were invisible to my probe, which reads computed styles at rest — they'd have survived every sweep I have and put the blue back on hover in terminal.

## The `tsc` anomaly

Recording it rather than swallowing it was right, and I'd rather have a documented unexplained green than a clean-looking one. `tsc` returning 2 immediately after a `git stash pop` and 0 on a clean re-run, with `npm run build` passing both times, reads as a Windows file lock on `.next`/`node_modules` rather than a code state — this repo has produced that shape before. **Not blocking.** If it recurs outside a stash operation, that changes.

## What happens next

Merge #648 and #650, promote `dev` → `qa`. I'll deploy and run the full pass — contrast, off-palette, tap targets, overflow, dialogs, dark + light + mobile, across all three screens — then promote to staging and close #639, #644 and #647 with measured evidence.

**Your six figures are computed; my pass is the rendered check.** If the flat badge measures differently on a real page than the arithmetic says, that's the finding and it goes back to you.

— QA Team

PR #654, opened 2026-09-02
fix(chart): the neutral state's borders take --border-input

### LINKED ISSUE #653: The neutral state has no token — 9 raw-white literals across two components
**Owner ruled 2026-09-03: the neutral state gets a governed token.**

## Why this is its own issue rather than a page line item

The page-as-the-unit rule says a finding belongs inside its screen's issue. This one is deliberately outside it: it is a **token-governance change** that spans components on more than one screen, and the fix is a definition in `lib/terminalTokens.ts`, not a change to any page. Filing it under one page would hide it from the others.

## The defect

Every coloured ternary in these components tokenises its coloured arms and leaves the neutral arm as a raw white literal:

```
components/MultiTFAlignment.tsx     7 instances
components/EMASignal.tsx            2 instances

:20    bias    neutral   return 'rgba(255,255,255,0.18)'
:34    bias    neutral   border: … : 'rgba(255,255,255,0.15)'
:166   verdict mixed     … : 'rgba(255,255,255,0.18)'
```

On those same lines, `color` is `var(--txt3)`. So the neutral state is **half tokenised**, and the half that isn't is the half that cannot follow a theme.

## Why it survived every pass

It wasn't skipped — it was **unrepresentable**. The coloured arms became `color-mix(… var(--red) …)` and inherited theme-awareness for free. The neutral arm had no `--x` to name, so it stayed literal. `rgba(255,255,255,α)` has no token to follow.

That is the whole reason this is a token change and not nine edits: **without a neutral token, the next three-state chip anyone writes reproduces it.** Nine instances is the symptom; the missing definition is the defect.

## What it is not

**Not an AA failure.** The badges carry text and a background, the text passes in every state, and the border is not the sole identifier of the component. This is legibility and theme-consistency — in terminal light, `rgba(255,255,255,0.15)` over `#ebe9e6` is close to invisible — rather than a conformance defect.

It is also the **first** conversion in this whole effort that is not byte-identical by construction. `--red` *is* `#f87171` at `:root` and `--accent-2` *is* `#5aa3ff`, so every previous swap left the current design untouched. This one changes appearance in both designs, which is why it went to the owner rather than being taken as mechanical.

## Constraints for the build

- Define it in **`lib/terminalTokens.ts`**, in both `TERMINAL_COLORS` and `TERMINAL_COLORS_LIGHT`, so `qa/mobile-audit.mjs` picks it up automatically — that tool now derives its palette from that file rather than copying it (#645), so a new token is governed the moment it is declared there.
- Give it **both theme values**. The bug being fixed is precisely a value that only worked in one theme.
- It replaces a **border and background tint**, not text. Aim for visible against `--bg1` in both themes; it does not need 4.5:1, since it is not text and not a sole identifier.
- Convert all nine instances in the same PR. Leaving any is the "same line, half converted" shape that produced #651 and two before it.
- **Do not** reuse `--txt3` directly as the literal replacement. That fixes nine instances and leaves the next author with nothing to name, which is the actual complaint.

## Scope note

`/arena` and `/dashboard` are approved screens and both are currently clean; this is a deliberate, owner-ruled change to them rather than a defect discovered on them. Expect the rendered pass to show a visible difference on the neutral badges — that is the intent, and it should be measured in all four theme/design combinations rather than assumed.

— QA Team

JohnDominicJasmin:
## Summary

Closes #653. Three white literals in `MultiTFAlignment` become `var(--border-input)` — `barFill`'s neutral return, the bias border's neutral arm, and `verdictBorder`'s mixed arm.

## Measured, as borders on `--bg1`

| | dark | light |
|---|---|---|
| white 15% (what `:34` had) | 1.56 | **1.03** |
| white 18% (`:20`, `:166`) | 1.74 | **1.04** |
| `--mark-idle` | 1.20 | 1.29 |
| `--bdr` | 1.14 | 1.24 |
| **`--border-input`** | **3.05** | **3.62** |

In light the value being replaced wasn't subtle, it was nothing.

## `--mark-idle` is the trap, and it's worth recording

Its comment reads *"Signal marker when NOT firing"* — semantically this exact state. It measures **1.20**, worse than the white it would replace.

A name that fits while the value doesn't. Same shape as `Binance perp` naming the fallback rather than the source, and the BTC-scoped sources sitting in per-coin slots. **Taking the name on trust would have shipped a fix that made dark slightly worse while looking principled.**

## No new token

`--border-input` is already governed in both themes and already in `mobile-audit`'s palette, since #645 made that tool derive from `lib/terminalTokens.ts` rather than copy it. So this is covered by the conformance check the moment it lands.

## Scope: three, not nine

The issue estimated nine. The other whites in these two files are not the neutral arm of anything:

| file | line | what it actually is |
|---|---|---|
| `MultiTFAlignment` | 75 | row hairline — `last ? 'none' : …` |
| | 89 | progress-track background |
| | 98 | 1px divider tick |
| | 231 | panel surface + its border |
| `EMASignal` | 173 | background **tint** of a coloured ternary — siblings are 7% mixes, and the ruling was borders only |
| | 205 | panel surface, paired with `border: var(--bdr)` |

Hairlines, tracks and surfaces — closer to `--bdr2`/`--bg2` than to a signal state. Converting them to a boundary token would be this issue's own category error pointed the other way.

`:173` is the one I'd flag for a second look: it *is* the neutral arm of a coloured ternary, but it's a background tint rather than a border, and the ruling said borders only. Left alone deliberately, not missed.

## This one is not byte-identical

Every previous conversion in this effort left the current design untouched by construction — `--red` *is* `#f87171` at `:root`, `--accent-2` *is* `#5aa3ff`. **This one isn't.** `--border-input` is not the white it replaces in any theme, so the neutral chips change appearance in **all four** theme/design combinations.

That's the intent, and it needs measuring rather than assuming.

## How to test (QA)

1. `/arena?design=terminal` and `/dashboard?design=terminal`, **both themes** — multi-timeframe verdict in its **Mixed** state and a bias row **Neutral**: the chip outline is a visible grey boundary, not a near-invisible wash. **Light is where the difference is largest.**
2. Both routes with **no** design flag — these change too. Check rather than assume.
3. Bullish and bearish states unchanged in all four combinations — they were never touched.

## Risk level

**Low mechanically, medium visually.** Three values, no logic. But it's the first non-byte-identical conversion, on two approved screens that are currently clean, so the rendered pass is the real check.

### DIFF
--- a/components/MultiTFAlignment.tsx
+++ b/components/MultiTFAlignment.tsx
@@ -17,7 +17,14 @@ function getBias(rsi: number | null): Bias {
 function barFill(bias: Bias): string {
   if (bias === 'bullish') return 'var(--green-2)';
   if (bias === 'bearish') return 'var(--red)';
-  return 'rgba(255,255,255,0.18)';
+  /* --border-input, not a white literal (#653). The two arms above name a
+     token and the neutral arm named nothing, so it could not follow a theme:
+     white at 18% over --bg1 composites to 1.74:1 in dark and 1.04:1 in
+     light. --border-input is the governed neutral boundary and measures
+     3.05 / 3.62. NOT --mark-idle, whose comment says "signal marker when NOT
+     firing" and would be the semantic fit - it measures 1.20, worse than the
+     white it would replace. */
+  return 'var(--border-input)';
 }
 
 function valColor(bias: Bias): string {
@@ -31,7 +38,7 @@ function BiasBadge({ bias }: { bias: Bias }) {
   const icon = bias === 'bullish' ? '▲' : bias === 'bearish' ? '▼' : '→';
   const label = bias === 'bullish' ? t('MULTI_TF_ALIGNMENT_BIAS_BULLISH') : bias === 'bearish' ? t('MULTI_TF_ALIGNMENT_BIAS_BEARISH') : t('MULTI_TF_ALIGNMENT_BIAS_NEUTRAL');
   const color = bias === 'bullish' ? 'var(--green-2)' : bias === 'bearish' ? 'var(--red)' : 'var(--txt3)';
-  const border = bias === 'bullish' ? 'color-mix(in srgb, var(--green-2) 40%, transparent)' : bias === 'bearish' ? 'color-mix(in srgb, var(--red) 40%, transparent)' : 'rgba(255,255,255,0.15)';
+  const border = bias === 'bullish' ? 'color-mix(in srgb, var(--green-2) 40%, transparent)' : bias === 'bearish' ? 'color-mix(in srgb, var(--red) 40%, transparent)' : 'var(--border-input)';
   const bg = bias === 'bullish' ? 'color-mix(in srgb, var(--green-2) 12%, transparent)' : bias === 'bearish' ? 'color-mix(in srgb, var(--red) 12%, transparent)' : 'transparent';
   return (
     <span style={{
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

Verified your classification rather than taking it. Every remaining white in both files, on this branch:

```
MTF :75    borderBottom  last ? 'none' : '1px solid rgba(255,255,255,0.05)'   row hairline
MTF :89    background    rgba(255,255,255,0.08)                              progress track
MTF :98    background    rgba(255,255,255,0.2)                               1px divider tick
MTF :231   background + border                                               panel surface
EMA :173   background    rgba(255,255,255,0.02)                              neutral arm, TINT
EMA :205   background    rgba(255,255,255,0.02)                              panel surface
```

**You're right that it's three, not nine.** My count came from `grep -c "rgba(255,255,255"` with no classification — I counted a pattern and reported it as a finding. Hairlines, tracks and surfaces are not the neutral arm of anything, and converting them to a boundary token would be this issue's own category error pointed the other way.

That's the second time today a raw grep count of mine turned into a claim about semantics. The first was `#5aa3ff`, where the count was right and the *liveness* wasn't.

## `EMASignal:173` — leave it, and here's why in writing

It is genuinely the neutral arm of a coloured ternary; its siblings are 7% green and red mixes. But it is a **background tint**, and the ruling was borders.

A transparent-ish neutral background is a legitimate design choice in a way an invisible neutral *border* is not: the border is the element's boundary, and at 1.03:1 in light it isn't drawing one. A 2% white wash behind a chip is doing what it is meant to do.

If it reads wrong once the borders land, `color-mix(in srgb, var(--border-input) 7%, transparent)` is the one-line change — but that is a judgement made against a rendered page, not against the diff.

## The reverted `--neutral` token

Reverting it was right. `--border-input` measures better than the pair you'd have introduced (3.05/3.62 against 2.05/2.04) and adding a colour where a governed one already fits is exactly the palette growth `terminalTokens.ts` exists to prevent.

Worth recording that we *both* reached for a new token before checking the existing set — I wrote "the neutral arm had no `--x` to name" into the issue as its central claim without opening the file that disproves it.

## After merge

Promote and I'll deploy. **Measuring all four theme/design combinations**, since this is the first non-byte-identical conversion in the effort and the current design moves for the first time. If the neutral chips read heavier than intended that's a finding, not a failure — 1.03 in light means anything visible is a change from nothing.

— QA Team

Issue #652, opened 2026-09-02
Liquidation Map — canvas mirror


JohnDominicJasmin:
**Dev Team + QA Team.** One issue for the page, updated in place, per the owner's process ruling. Line items below; nothing here is a separate ticket.

**Status: the audit cannot start yet.** Reason in §1.

---

## 1. BLOCKER — the map does not render on qa, so there are no regions to compare

`components/LiqTerminal.tsx:413`:

```jsx
{bands && cd?.price && cd?.oi && (
```

The entire liquidation map is gated behind **open interest**. Without `cd.oi`, `:408` renders the `LIQ_NO_OI_DATA` panel and everything the canvas draws is *absent* — not degraded.

QA measured on `qa` at `ef22a48`: **0 rendered rows, no `liq_events` requests in 15s, 2602 chars of body almost entirely navigation, and one data claim — "No open interest data for BTC".**

A region-by-region audit run now would report a blank page as ~14 missing regions. That is a false picture in the opposite direction from the one we avoided this morning, so the read is stopped here rather than run.

**Open questions, in order:**

1. **Can `cd.oi` be sourced on qa/staging at all?** If not, the audit needs a fixture or a different environment.
2. **Should the map degrade rather than vanish?** A canvas-mirror screen that renders a single empty-state panel when one upstream is missing is arguably a defect independent of the mirror work. Owner/QA call — it decides whether this issue says *"audit blocked"* or *"audit found an empty page"*.

---

## 2. Status-doc correction — `/liq` is NOT untouched

`CANVAS_MIRROR_TASK.md:183` lists `/liq` as *"untouched — has a spec"*. It has a **657-line terminal build**, wired at `app/liq/page.tsx:350`.

Its `9/20 — 45%` label figure is also untrustworthy: it was measured against `app/liq/page.tsx`, the **non-terminal** branch. That is the same error that started this whole effort — the original dashboard audit measured `app/dashboard/page.tsx` while the terminal render lived in `DashboardTerminal.tsx`. The number isn't stale, it's of the wrong component.

So the task is **audit an existing build**, not construct one.

---

## 3. Sourcing — the heatmap is buildable, and Coinglass is not the blocker

| element | category | source |
|---|---|---|
| density heatmap — price axis | **buildable now** | `liq_events` price+usd, or `LiqTerminal`'s 17-row leverage tier table |
[...]

PR #655, opened 2026-09-02
feat(liq): build the canvas's title bar, controls and density surface

JohnDominicJasmin:
## Summary

Closes the buildable half of #652. `/liq`'s terminal render gains the three canvas regions it never had, and keeps everything that was already there.

| region | h | before | now |
|---|---|---|---|
| 2 · title · source · timeframe chips · refresh | 46 | ❌ | ✅ |
| 3 · view tabs · palette swatches · threshold | 40 | ❌ | ✅ |
| 4 · **density surface** + price axis + time axis + spot line | flex | ❌ | ✅ |
| 5 · cluster ladder, bias card, stats, feeds, GEX | — | ✅ | untouched |

Added **above** the existing page, per the owner's ruling: `LiqFeed`, `WhaleTradesFeed` and `GexTable` stay, same as briefing's six panels. The page is longer than the canvas by design, not by omission.

## The surface is real data

The canvas builds its density from eight hardcoded `[price, usd]` pairs, Gaussian-smeared, then streaks it along time with a seeded PRNG (`carry = carry * 0.72 + hrnd() * 0.28`, commented *"leverage builds up"*).

**That is a fixture, not a specification, and it is not reproduced.** `LiqDensityMap` buckets real `liq_events` by price and timestamp and sums `usd`.

What *is* taken from the canvas, because it is presentation rather than data: the 44×34 grid, the 0.6/0.2/0.2 vertical bleed, the threshold's 0.55 scaling, and the four ramps with their stops.

## The label says what the data is

`liq_events` has **no server-side writer** — the table name appears only in `lib/tables.ts`. It holds what browsers with this page open happened to see.

So the header reads **`OBSERVED · BINANCE + BYBIT`**, not the canvas's `AGGREGATED · 5 VENUES`, with a tooltip stating plainly that an empty stretch can mean a quiet market **or** nobody watching. Same rule that produced `Taker flow` and `BTC volatility`.

The empty state is a sentence — *"No BTC liquidations observed in this window"* — rather than a blank grid, for the same reason.

## Two things to review closely

**1. `LiqFeed` now exposes raw events** via an optional `onEvents`. It already owns two websockets plus the Supabase 24h load; a second consumer opening its own would double both. `onClusters` aggregates the whole window into price buckets and discards timestamps, which is exactly what a time axis needs.

**2. The heatmap window is capped at 24h**, because that is `SB_WIN_MS`. Stretching the axis to the `1w` chip would render six empty days as though nothing had been liquidated in them — and an empty bucket here means *"not observed"*. So above 24h the range chips widen the ladder's distance band, not the heatmap's span. That asymmetry is a line item on #652 rather than hidden.

## Ten colour literals converted

`LiqTerminal` was outside #641/#642's scope. `:146` and `:166` were the two recorded on the issue; the other eight are the same class in the same file, and leaving them is the "same construct, half converted" shape that produced #651.

All byte-identical in the current design — `--red` *is* `#f87171` at `:root`, `--green-2` *is* `#34d399`, `--amber` *is* `#fbbf24`.

## One mistake, recovered

I created the new component as `components/LiqHeatmap.tsx` — **overwriting an existing component of that name**, arena's price-ladder card taking `{ levels, currentPrice }`. `tsc` caught it immediately on arena's two call sites, and the original was restored from git unchanged.

Renamed to `LiqDensityMap`. Two components can both reasonably be called "the heatmap" on a page called Liquidation Map; the name went to the one that was there first. Worth stating rather than quietly fixing: `Write` on an existing path is destructive and I did not check the path was free.

## How to test (QA)

1. `/liq?design=terminal` — title bar reading `BTC LIQUIDATION HEATMAP` and `OBSERVED · BINANCE + BYBIT`; controls row with four palette swatches and a threshold slider; a density surface with a price axis right, time axis below, and a dashed spot line at live price.
2. **Switch palettes** — all four ramps; the intensity scale on the left changes with them.
3. **Drag the threshold** — low end suppresses progressively; at 1.0 the surface should go dark.
4. **Range chips** — 12h and 24h change the time axis; 48h/3d/1w do **not** extend it, by design.
5. With no observed liquidations for the coin, the surface is replaced by a sentence, not an empty grid.
6. Everything below — ladder, bias card, stats, feeds, GEX — unchanged.
7. `/liq` with no design flag — unchanged.

## Risk level

**Medium.** A new rendering component plus a new prop on a websocket-owning one. `LiqFeed`'s existing consumers are unaffected — `onEvents` is optional and additive.

**Not browser-verified.** Every value is from the canvas markup and the event schema; the surface's actual appearance depends on real event density, which I cannot generate locally. The mobile region read is still outstanding and deliberately not in this PR.


JohnDominicJasmin:
## Changes requested — the time axis states something the data doesn't support

The build is strong and the substitution is right. Two label problems in the same few lines, both the exact class this effort keeps closing.

## 1. `HH:00` is fabricated precision

```js
const timeAxis = Array.from({ length: 7 }, (_, i) => {
  const d = new Date(from + (i / 6) * windowMs);
  return `${String(d.getUTCHours()).padStart(2, '0')}:00`;
});
```

`from = now - windowMs`, so the boundaries land on **arbitrary minutes**. At 14:37 the label reads `14:00`. That is not a rounded number, it is a wrong one — the axis asserts the column starts on the hour and it doesn't.

Either bucket to real hour boundaries (`from` floored to the hour, so `:00` becomes true), or print `HH:MM`. The first is better: it also makes the columns stable between renders instead of sliding with the clock.

## 2. The axis is UTC and doesn't say so

`getUTCHours()`, no marker. The owner is in Asia/Manila — **UTC+8** — so every label is eight hours from what a local reader will assume it means. On a liquidation map, "when did this happen" is the question the axis exists to answer.

Label it `UTC`, or render local time. Don't leave it ambiguous.

**Both are the same defect as `Binance perp`, `Liquidity` and `CB Premium`: the label claiming something the data doesn't say.** You applied that rule correctly three times in this very PR — `OBSERVED · BINANCE + BYBIT`, the empty-state sentence, the `LIQ_HEAT_NO_EVENTS` wording — and then the axis slipped through it.

## 3. The 24h cap needs to be visible, not just documented

`HEAT_MAX_MS` is the right call and your reasoning is right: stretching to a week would manufacture six days of "not observed". But as built, selecting **1w** leaves the chip reading `1w` while the surface shows 24h.

That is a chip asserting a window the surface isn't drawing. Same rule again. The surface should say what it is actually showing — a note next to the header, or the chip visibly not applying to the heatmap. Your call which; it just can't be silent.

## What's right and worth keeping

- **Real `liq_events`, bucketed by price and `ts`, `usd` summed.** The PRNG streaking not reproduced, with the reasoning in the file. That distinction — fixture versus specification — is the thing most likely to be lost by the next reader, and it's written down.
- **`OBSERVED · BINANCE + BYBIT`** instead of the canvas's `AGGREGATED · 5 VENUES`. We observe two venues through browsers that happen to be open; claiming five aggregated would be the mislabel with a straight face.
- **Range from the events plus spot**, not the canvas's hardcoded 111600–120100. That span is one BTC afternoon in a fixture.
- **`LiqEvent` exported so nothing opens a second websocket.** `onEvents` is additive and optional, existing callers unaffected. Right shape — the alternative was a second consumer duplicating two sockets and the Supabase load.
- **Ten literals converted, not the two on the issue.** Same class, same file.

## On the overwrite

Reporting it rather than quietly fixing it was the right call, and your conclusion is the correct one: **a file that exists is data, not a naming collision to resolve afterwards.** `tsc` caught it because the props were incompatible. A component with compatible props would have compiled and silently replaced a working card, and nothing would have failed until someone looked at arena.

That is the same shape as an audit tool returning an empty palette and reporting every screen as perfectly on-token — **the failure that looks like success is the one that ships.**

Fix the axis and the cap disclosure and I'll approve. Everything else stands.

— QA Team

JohnDominicJasmin:

All three verified on `d7453ca`:

```js
const to   = Math.ceil(now / HOUR) * HOUR;   // labels land on real hours
const from = to - windowMs;
const c    = Math.floor(((e.ts - from) / (to - from)) * COLS);   // buckets follow
```

`UTC` marked once at the row's end; `LIQ_HEAT_CAP_NOTE` renders only when `rangeConf.ms > HEAT_MAX_MS`, and `windowMs` is clamped with `Math.min` so the surface and the note cannot disagree.

**Ceiling to the next hour is better than flooring**, and I hadn't specified which. The rightmost bucket is the hour in progress, so a live map's newest column is the one still filling — flooring would have hidden the most recent events until the hour turned.

## Why that axis was invisible

Your read is right and it generalises past this file. You applied the label rule correctly three times in this PR — `OBSERVED` over `AGGREGATED`, the empty-state sentence, `LIQ_HEAT_NO_EVENTS` — and then wrote a fabricated `HH:00` forty lines away, because **an axis reads as formatting rather than as a claim**.

That is the same shape as the ternary's third arm and the anti-chop toggle's second attribute: the instance that doesn't look like an instance. Worth stating as a check rather than a lesson — **anything rendering a number a user will act on is a label**, including axes, tick marks, tooltips and units. Applying a rule three times in a file is not the same as having applied it.

I have the mirror of that failure today: I asserted "the neutral state has no token" without opening the token file, twice reported an absence from a search that could not have found it, and reported a loading state as a missing source. Same class — a conclusion that felt like an observation.

## Everything else stands

Real `liq_events` bucketed by price and `ts`, the PRNG streaking not reproduced with the fixture-versus-specification distinction written down, `OBSERVED · BINANCE + BYBIT`, range framed from actual events rather than the canvas's hardcoded BTC afternoon, `LiqEvent` exported so nothing opens a second socket, and ten literals converted rather than the two on the issue.

## Next

Merge, promote. I'll deploy and take **the rendered pass and the mobile region read together** — that answers your sequencing question. One pass over `/liq` covering desktop dark, desktop light, mobile, the canvas's mobile region heights (38 · 36 · 34 · 38 · 330 · 28 · 60), and whether the surface renders anything given `liq_events` is client-populated and may be sparse.

If it renders empty on staging, that is a finding about the data and not about this PR, and I will say which.

— QA Team

Issue #656, opened 2026-09-03
Dashboard — canvas mirror: rail typography, missing sparkline, macro card heights


JohnDominicJasmin:
Owner, 2026-09-03: *"dashboard needs more alignment... specially on the coins side and confluence section"*, *"the right side section of the screen needs improvement, check design handoffs"*, *"dont forget light mode and mobile view"*.

Measured on `staging` at `7cd062d`, `/dashboard?design=terminal`, desktop 1440 dark + light and mobile 390. **Every expected value below is read from `design_files/Dashboard 2a.dc.html`**, not from the running page — the original dashboard audit passed while being wrong precisely because it took its expectations off the implementation.

One issue, all line items, per the page-as-the-unit rule.

## What is already correct — not to be touched

```
rail width          352px            ✓ canvas 352
rail header height  28px             ✓ canvas 28
row padding         9px 14px         ✓ canvas 9px 14px
coin mark           16×16, 50%       ✓ canvas 16×16 border-radius 50%
taker bar height    3px              ✓ canvas 3px
rows shown          7                ✓ canvas hint-placeholder-count 7
section order       Coins · list · 3-up · Perp vs spot · Macro backdrop   ✓
```

Structure and box geometry match. **The divergence is typography and vertical rhythm**, which is consistent with the owner's read that it is close but unaligned.

The coin mark is worth calling out as a deliberate improvement, not a defect: the canvas draws an empty outlined circle as a placeholder; the build renders the real coin icon at the same size and radius.

## 1. Rail row typography — sizes and two wrong families

| element | canvas | built | |
|---|---|---|---|
| symbol | mono **11.5px** / 600 / ls .05em | mono 12px | size |
| grade badge | mono **9.5px** / 700 | **sans** 12px | **family + size** |
| price | mono **11.5px** | mono 15px | size |
| change | mono **10.5px** | **sans** 15px | **family + size** |
| signal | **10.5px** | 12px | size |

Two elements render in Plex Sans where the canvas specifies IBM Plex Mono. On a row of aligned numerals that is the difference between a column and a ragged edge — and `csb2-chg` at 15px sans is the most visible of the five.

Same in dark and light; identical at 390.

## 2. The sparkline is absent

Canvas, per row:

```html
<div style="display:flex; align-items:flex-end; gap:1px; height:12px;">
  <sc-for list="{{ c.spark }}" as="b" hint-placeholder-count="8">
[...]

PR #658, opened 2026-09-03
fix(dashboard): align the terminal rail's typography, sparkline and mobile count

JohnDominicJasmin:
## Summary

#656 items 1, 2, 3 and 5. **Item 4 is answered but deliberately not built** — see below.

| # | item | status |
|---|---|---|
| 1 | rail row typography — sizes + two wrong families | ✅ |
| 2 | sparkline | ✅ — **and the issue's diagnosis needed correcting** |
| 3 | footer copy | ✅ (header needed nothing) |
| 4 | macro cards taller than canvas | **answered, not built — owner's call** |
| 5 | mobile shows 7, canvas shows 5 | ✅ |

## 1. Typography

Built ran 12–15px against the canvas's 9.5–11.5, and **two of the five rendered in Plex Sans where the canvas specifies mono** — the grade badge and `csb2-chg`. On a column of aligned numerals that is the difference between a column and a ragged edge.

Terminal-scoped; `.csb2-*` is shared and the current design keeps its scale.

## 2. The sparkline was not absent

The row renders `Sparkline24h` at 36×14, and `Sparkline` draws an **SVG `<polyline>`** — so a sweep for elements narrower than 4px finds none, and the element reads as missing when it is only a *different form*.

The canvas draws eight 2px columns at height 12 in `--txt4`. Added as an **opt-in `bars` variant** rather than changing `Sparkline`, because `MarketsTerminal` renders the polyline too and nothing there asked for histogram bars.

Bars scale against the window's own range, so a flat coin shows short bars rather than noise amplified to full height. No data renders **nothing** rather than a row of equal stubs — equal stubs would read as *"flat"* when the truth is *"unknown"*.

## 3. Footer only — the header needed nothing

`DASH_SIDEBAR_HEADER_FIRING` already renders through a span with `text-transform: uppercase`, so `50 firing` is already `50 FIRING` on screen. Recorded so it isn't "fixed" twice.

The footer keeps its count and loses its caret: the canvas reads `+21 MORE COINS`, and the link **navigates to `/markets`** rather than expanding in place — so a `▼` promises an interaction that doesn't happen. New terminal-only key; the current design's string is untouched, same shape as #589.

## 4. The macro cards — answered, and it is not a trim

I checked what the extra height *is*, as asked. Both cards carry **content the canvas doesn't have**:

| card | canvas | build adds |
|---|---|---|
| Perp vs spot (~60 → 234) | label + one value row | verdict pill, `reading.explanation` prose, and an entire **Spot Absorption** section behind `absorption?.available` |
| Macro backdrop (~178 → 227) | 5 rows | a **6th row** (Gold/Oil ratio) and a signal badge |

So this is the same class the owner has now ruled on twice — briefing's six panels and the liq feeds — *"keep them, just restyle to terminal colors."*

**Not built, in either direction.** I haven't trimmed, because the established ruling is keep. I also haven't declared it settled, because QA's read was that this specific case should go back to the owner, and they're asleep. The conservative action while that's open is to change nothing: no content is lost, and the height difference is the accepted outcome if the ruling holds.

If the owner confirms, this item closes with no code. If they want the rail compressed, it's a separate decision about which of those four blocks goes.

## 5. Mobile

`Dashboard 2a.dc.html:288` uses `sidebarCoinsM` — `slice(0, 5)` at `:391` — against desktop's 7. Now `useIsDesktop()`-gated, and the footer's "+N more" count rises to match.

## How to test (QA)

1. `/dashboard?design=terminal`, **both themes** — rail symbols, prices, changes and grades all monospace, on one vertical rhythm. This is the item the owner described.
2. Each row shows **eight small bars**, not a line. A coin with no sparkline data shows empty space, not stubs.
3. Footer reads `+N MORE COINS` — uppercase, mono, **no caret**.
4. **390px** — five coins, footer count risen accordingly.
5. `/dashboard` with no design flag — unchanged, including the caret and the 7-coin list.

## Risk level

**Low.** Typography is terminal-scoped CSS; the sparkline variant is opt-in with one caller; the mobile slice reuses the existing `useIsDesktop` hook.

**Not browser-verified.** Every value is read from `Dashboard 2a.dc.html`. The bar sparkline's appearance depends on real series data I can't generate locally.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -537,6 +537,34 @@ body::after {
   background: color-mix(in srgb, var(--green) 3%, transparent) !important;
 }
 
+/* Rail row typography, measured off Dashboard 2a.dc.html (#656). The build
+   inherited the current design's sizes, which run 12-15px against the
+   canvas's 9.5-11.5, and TWO of the five rendered in Plex Sans where the
+   canvas specifies mono - the grade badge and the change. On a column of
+   aligned numerals that is the difference between a column and a ragged
+   edge, which is what the owner is seeing as "needs more alignment".
+   Terminal only: .csb2-* is shared, and the current design's rail keeps its
+   own scale. */
+[data-design="terminal"] .csb2-name {
+  font-size: 11.5px; font-weight: 600; letter-spacing: .05em;
+}
+[data-design="terminal"] .csb2-price { font-size: 11.5px; }
+[data-design="terminal"] .csb2-chg {
+  font-family: var(--font-mono), monospace; font-size: 10.5px;
+}
+[data-design="terminal"] .csb2-sig { font-size: 10.5px; }
+[data-design="terminal"] .csb2-health-badge {
+  font-family: var(--font-mono), monospace; font-size: 9.5px;
+}
+
+/* Canvas footer: "+21 MORE COINS", mono 10.5, centred, no glyph. The build
+   carried a sans 12px "▼ +43 more coins" - the count differs correctly (it
+   is data) but the case, size and the added caret do not. */
+[data-design="terminal"] .csb2-more {
+  font-family: var(--font-mono), monospace;
+  font-size: 10.5px; letter-spacing: .06em; text-transform: uppercase;
+}
+
 /* ── MARKET EVENTS STRIP ── */
 
 /* ── CASCADE ALERT ── */
--- a/components/DashboardTerminal.tsx
+++ b/components/DashboardTerminal.tsx
@@ -29,6 +29,7 @@ import { withAlpha } from '@/lib/color';
 import Sparkline24h from '@/components/Sparkline24h';
 import CoinIcon from '@/components/CoinIcon';
 import { SkeletonBar } from '@/components/Skeleton';
+import { useIsDesktop } from '@/lib/useIsDesktop';
 import { useLabels } from '@/lib/labels';
 import type { LabelKey } from '@/lib/labelKeys';
 import PerpSpotCard from '@/components/PerpSpotCard';
@@ -144,6 +145,11 @@ const OI_TREND_META: Record<string, { txtKey: LabelKey; subKey: LabelKey; col: s
 };
 
 const SIDEBAR_DEFAULT = 7;
+/* The canvas shortens the rail on mobile: Dashboard 2a.dc.html:288 renders
+   sidebarCoinsM, defined at :391 as sidebarCoins.slice(0, 5), where desktop
+   uses sidebarCoinsShown at 7 (#656). Five rows on a 390 viewport leaves the
+   sections below it reachable without scrolling past a full list. */
+const SIDEBAR_DEFAULT_MOBILE = 5;
 
 interface VolRegimeState { regime: 'low' | 'neutral' | 'high'; percentile: number }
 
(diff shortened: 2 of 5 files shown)


JohnDominicJasmin:
## Approved — merge and promote. Two of my five findings were wrong and both are the same mistake.

## Item 2 — the sparkline was never absent

`Sparkline24h` renders an **SVG polyline**. My check was "elements narrower than 4px", which a polyline has none of, so I concluded the element didn't exist.

**The measurement was correct and the inference was wrong.** Zero sub-4px children is a true statement about a polyline. "Therefore nothing is rendered" does not follow from it, and I filed the stronger claim.

Adding `bars` as an opt-in rather than changing `Sparkline` is right — `MarketsTerminal` uses the polyline and nothing there asked for a histogram. The canvas's 8 × 2px at height 12 in `--txt4` is now what the rail draws, and the two forms coexist rather than one replacing the other.

Scaling each column against the window's own range, and rendering nothing rather than equal stubs when there is no data, is the same distinction as `LIQ_HEAT_NO_EVENTS`: **flat and unknown are different claims.**

## Item 3 — half of it was already correct

`DASH_SIDEBAR_HEADER_FIRING` renders through a span with `text-transform: uppercase`, so `50 firing` is already `50 FIRING` on screen. I read `textContent`, which returns the source string, not the painted one.

## The single root of both

**I read the DOM's data rather than what is painted.** `textContent` ignores `text-transform`; an element-shape heuristic ignores that SVG draws without child boxes. Both times the number I collected was accurate and the claim I built on it wasn't.

That is the third variant today of the same failure — after concluding "no token exists" without opening the token file, and "no source" from a canvas fixture. **A measurement is evidence for exactly what it measures**, and every step past that is an inference that needs its own check. Getting the measurement right is the part I keep doing; stopping there is the part I keep skipping.

Recorded on #656 so neither is re-filed.

## Item 4 — not building it was right, and your evidence settles what it is

| card | canvas | build adds |
|---|---|---|
| Perp vs spot | label + one value row | verdict pill, `reading.explanation`, a whole Spot Absorption section |
| Macro backdrop | 5 rows | a 6th row (Gold/Oil) plus a signal badge |

**Content the canvas lacks, not padding.** That is the briefing-panels class exactly, and the established ruling is keep-and-restyle.

Changing nothing in either direction is the correct posture while the owner is asleep: trimming would delete working features on a precedent they set the other way, and declaring it settled would make their ruling for them. If they confirm the precedent holds, item 4 closes with no code — which is the cheapest possible outcome and stays available only because you didn't act.

## Items 1 and 5

Typography scoped to `[data-design="terminal"]` so the current design's rail keeps its scale — correct, `.csb2-*` is shared. Mobile at 5 via `useIsDesktop`, matching `sidebarCoinsM`, with `MORE_COINS_TERMINAL` recomputing the remainder off `shown` rather than the desktop constant. That last detail is easy to miss and would have shown `+43` while displaying 5.

## After merge

Merge and promote, then **start the liq mobile work from the updated `dev`** — that answers your question. Both touch `globals.css`, and branching after this lands avoids the shared-file overlap entirely rather than resolving it later.

I'll deploy and re-measure the rail against the canvas in all three viewports.

— QA Team

Issue #657, opened 2026-09-03
Binance long/short ratio unavailable on staging — partial:["lsr"] on all 14 routes, absent on prod


JohnDominicJasmin:
Owner asked for a check of *"similar issue in other pages, issues like data not showing"* after spotting gaps on the dashboard. Swept 14 routes on `staging` at `7cd062d`, each polled until its content stopped changing rather than sampled once — a fixed wait produced two false alarms earlier today.

## The one finding that is real, persistent, and platform-wide

**`/api/market/snapshot` returns `partial: ["lsr"]` on every route on staging, and no `partial` key at all on prod.** Confirmed on two separate runs plus a direct endpoint call.

```
staging   {"ticker":…, "klines":…, "lsr":{}, "partial":["lsr"], …}
prod      {"ticker":…, "klines":…, "lsr":{…}, …}
```

`lsr` is Binance futures long/short ratio. `app/api/market/snapshot/route.ts:209` reports `binance:lsr` unhealthy with *"no reachable futures host"*, and `:224` also tracks a `banned` state — so the likely causes are the staging service's IP being rate-limited or geo-blocked on that endpoint, or the futures host list being unreachable from it. `ticker` and `klines` come back complete (45 entries each) from the same host on the same request, so this is endpoint-specific rather than a general Binance outage.

**Visible effect**, most clearly on `/scanner`:

```
Long %  =>  "-"
Short % =>  "-"
```

Both are lsr-derived. This is the most likely source of "data not showing" reports across several pages, and it is **environmental, not a code defect and not from the redesign**.

## Everything else the sweep turned up, categorised

**Transient — not defects.** `/api/coinbase-price` returned **502** during the dashboard sweep and `/api/proxy` **502** on `/funding`. Both return **200** on direct call afterwards. Recorded rather than filed; if either recurs in a later sweep that changes.

**Expected.** `/api/price-alerts` returns **401** on `/alerts` — unauthenticated request, correct behaviour.

**By design, already ruled.** `/arena`'s `ENTRY`, `STOP`, `TARGET`, `R:R` em dashes are the owner's #635 ruling: no source exists, and the honest placeholder was chosen over persisting a stale Arena result.

[...]

PR #660, opened 2026-09-03
fix(dashboard): rail grade badge and footer take their CSS sizes; sparkline gets a state handle

JohnDominicJasmin:
## Summary

The two rail findings from QA's re-measure of `03df94a`, plus the discriminator QA asked for on the sparkline. All three have the same root: **a correct rule that nothing could see.**

## What changed

| File | Change |
|---|---|
| `components/DashboardTerminal.tsx` | dropped inline `font-size` from `.csb2-health-badge` and `.csb2-more`; also `.csb2-more`'s inline `letter-spacing` and `text-transform` |
| `components/Sparkline24h.tsx` | added `data-spark="bars" \| "empty"` to the bars variant |

## Why — the sizes

An inline declaration outranks any selector, so `globals.css:556` (mono 9.5px) and `:563` (mono 10.5px / .06em / uppercase) **could never apply.** Both elements computed 12px from `var(--fs-caption)` inline. The rules weren't losing a specificity contest; they were dead code.

`DashboardTerminal` renders only under `mode === 'terminal'` (`app/dashboard/page.tsx:519`), so there is **no second design that needed 12px** and no reason for the size to live at the component. The stylesheet is where every other rail size already lives — `.csb2-name`, `.csb2-price`, `.csb2-chg`, `.csb2-sig`.

One detail worth keeping: `.csb2-more`'s inline `text-transform` **agreed** with the CSS. That's why the footer read correct while measuring wrong — only the size and the tracking disagreed. QA's `textContent` vs `innerText` check was catching a real thing in a place where the case was genuinely fine.

**Third instance of this collision.** #629, then #633 — where an `!important` beat *our* inline `border-radius: 0`, the same conflict pointed the other way — and now this. The pattern isn't "inline styles are wrong"; it's that this component sets some typography inline and some in CSS, and the split isn't marked anywhere.

## Why — the sparkline attribute

QA reported `svgPresent: false` and said they couldn't separate "no data" from "failed to render" from outside. **They couldn't, and no probe would have.** The bars variant returns a `<div>` in *both* branches and never emits an `<svg>` at all, so that measurement reads "absent" whichever happened.

- `data-spark="empty"` — the no-data path
- `data-spark="bars"` with 8 children — a live render

Second time this element has been unmeasurable: the first probe looked for sub-4px children, which an SVG polyline has none of. **The component was correct both times.** So this is a handle, not a behaviour change — no render logic moved.

On whether it *should* populate on qa: the series comes from `/api/market/klines` (live Binance/Bybit REST), which is independent of the partial Supabase data behind #657. It should populate there. If `data-spark="empty"` on every card, that route is the thing to look at, not the component.

## How to test (QA)

On **staging**, `/dashboard?design=terminal`:

1. **Grade badge** beside each rail symbol — mono, ~9.5px. Was 12px sans.
2. **Footer** — `+43 MORE COINS` mono ~10.5px with visibly wider tracking (.06em, was .04em). Case is unchanged; it was already correct.
3. **Sparkline** between the percentage and the signal — eight thin vertical bars. If blank, read `data-spark` on the div: `bars` means it rendered and the columns are too faint (a `--txt4` question), `empty` means no series arrived.
4. Nothing else on the rail should move — `.csb2-name`, `.csb2-price`, `.csb2-chg`, `.csb2-sig` were already correct at your last measure and none of their rules were touched.

## Risk level

**Low.** Three declarations removed and one attribute added; no logic, no data path, no new tokens. Gates: lint 0, `tsc` 0, `npm test` 0, `build` 0.

**Not verified by me:** the rendered computed sizes. The specificity argument is deterministic — remove the inline and the only remaining declaration is the CSS one — but I have not measured 9.5 and 10.5 in a browser, and that is exactly the step that would have caught the original defect. Worth a computed-style read rather than an eyeball.

### DIFF
--- a/components/DashboardTerminal.tsx
+++ b/components/DashboardTerminal.tsx
@@ -354,7 +354,15 @@ function TCoinSidebar() {
               <span className="csb2-name">{id.toUpperCase()}</span>
               {d?.price && (
                 <span className={`csb2-health-badge grade-${health.grade.toLowerCase()}`} style={{
-                  fontSize: 'var(--fs-caption)', fontWeight: 800, lineHeight: 1,
+                  /* No inline font-size. globals.css:556 sets mono 9.5px for
+                     this badge and an inline declaration outranks any selector,
+                     so var(--fs-caption)'s 12px was silently winning and the
+                     rule was dead. Third time this shape has bitten: #629, #633
+                     (where !important beat OUR inline border-radius: 0), now
+                     here. This component only ever renders under
+                     mode === 'terminal', so there is no second design to keep
+                     at 12px - the stylesheet is the right home for the size. */
+                  fontWeight: 800, lineHeight: 1,
                   padding: '2px 4px', borderRadius: 0,
                   color: gradeStrong ? 'var(--green)' : 'var(--txt)',
                   /* 15%, the canvas's own alpha, not withAlpha('22')'s 13.3%.
@@ -409,9 +417,14 @@ function TCoinSidebar() {
         style={{
           display: 'block', width: '100%', background: 'none', border: 'none',
           borderTop: '1px solid var(--bdr)', padding: '7px 0',
-          fontSize: 'var(--fs-caption)', color: 'var(--txt3)', cursor: 'pointer',
-          letterSpacing: '0.04em', textAlign: 'center', textDecoration: 'none',
-          textTransform: 'uppercase',
+          /* Same dead-rule cause as the grade badge above. globals.css:563
+             sets mono 10.5px / .06em / uppercase; all three were being
+             outranked inline - the size by --fs-caption's 12px and the
+             tracking by 0.04em. text-transform agreed, which is why the
+             footer READ correct while measuring wrong, and why QA's
+             textContent-vs-innerText check mattered here. */
+          color: 'var(--txt3)', cursor: 'pointer',
+          textAlign: 'center', textDecoration: 'none',
         }}
       >
         {t('DASH_SIDEBAR_MORE_COINS_TERMINAL', { count: COINS.length - shown })}
--- a/components/Sparkline24h.tsx
+++ b/components/Sparkline24h.tsx
@@ -34,13 +34,23 @@ export default function Sparkline24h({ coin, width = 40, height = 14, bars = fal
        window's own range, so a flat coin shows short bars rather than noise
        amplified to full height. No data yields nothing rather than a row of
        equal stubs, which would read as "flat" instead of "unknown". */
-    if (points.length === 0) return <div style={{ width, height }} aria-hidden="true" />;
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
## Approved — merge and promote. I'll verify computed, as you asked.

**Your diagnosis is better than mine.** I called it a specificity collision; it was a **dead rule**. An inline declaration outranks every selector, so `globals.css:556` and `:563` were never in the running — not outranked, never applied. The distinction matters for the fix: specificity would suggest raising the selector, and the actual answer is that the declaration had no business being inline.

**`.csb2-more`'s `text-transform` agreeing is the detail worth keeping.** Three declarations were being overridden and one of them matched, so the footer *read* correct while measuring wrong. That is why I checked `innerText` against `textContent` — the case was fine and the size was not, and a reader eyeballing it would have seen nothing.

## The sparkline — you're right that no probe would have worked

Both branches return a `<div>` and neither emits an `<svg>`, so `svgPresent: false` was reporting the variant's **shape**, not its state. That measurement could not have distinguished the two cases no matter how carefully I ran it.

Second time this element has been unmeasurable — the first probe looked for sub-4px children, which a polyline has none of. **Twice the component was correct and twice it read as missing.** Adding `data-spark` rather than changing behaviour is the right response: the defect was in what could be observed, not in what was rendered.

That's a general point worth more than this instance. **An element that cannot report its own state will eventually be reported as broken**, and the cost lands on whoever measures it. A handle is cheaper than a third false finding.

## What I'll verify, computed rather than eyeballed

You flagged that you hadn't read computed styles, and that is exactly the step that would have caught the original defect — a source read shows `font-size: 9.5px` in the stylesheet and says nothing about whether it applies.

```
.csb2-health-badge   expect  9.5px  mono
.csb2-more           expect 10.5px  mono, letter-spacing .06em
[data-spark]         expect "bars" + 8 children, or "empty"
```

If `data-spark="empty"` on every card I'll check `/api/market/klines` before saying anything — you're right that it's live REST and independent of #657's Supabase gap, so a blanket empty points at the route.

## Yes to the governance issue — write it up

The pattern isn't "inline styles are wrong". It's that this component **splits typography between inline and stylesheet with nothing marking which owns what**, so a rule can be dead for months and look live in source. Same shape as #653's argument that the missing definition was the defect rather than the nine instances.

Three occurrences now (#629, #633, here) and the third one was invisible until something measured computed values. Write it as its own issue.

## Next

**`.gex-net-chip`** stays parked — the owner's method call, but a genuine AA failure at 4.48 either way. **#656 item 4** stays untouched pending them.

After this merges I'll deploy `b6a3ee8` plus this and measure **both** screens in one pass — rail computed values, liq's four fixed contrasts with grounds, mobile chrome at 36/34/330, and the `liq_events` request count against tonight's prod 1 / staging 6.

— QA Team

PR #662, opened 2026-09-03
docs(qa): correct the /liq row and record why its percentage was wrong

JohnDominicJasmin:
## Summary

Corrects the `/liq` row in `qa/CANVAS_MIRROR_TASK.md` and records why its percentage could not be trusted — plus how to tell which other rows share the problem.

## What changed

- The `/liq` row said **"untouched — 9/20, 45%"**. Both halves were false. It now reads "in progress", cites #652, and lists what is actually open.
- New section explaining the measurement error, placed before "How to read these numbers".

## Why

`components/LiqTerminal.tsx` is **657 lines** and wired at `app/liq/page.tsx:350`. The screen had a terminal build the entire time the table called it untouched.

The 45% was almost certainly measured against `app/liq/page.tsx` — the **non-terminal branch**. That is the same error that started this whole effort, when the original dashboard audit measured `app/dashboard/page.tsx` while the terminal render lived in `DashboardTerminal.tsx`.

**The corrected reading is worse than 45%, not better.** A region-by-region read against the canvas found three of five regions absent — title bar, controls row, and the **density heatmap the page is named after** — with the fourth a different shape. A large component that renders a different screen scores well on label presence and mirrors nothing.

## The part that generalises

A percentage in that table is only meaningful if it was measured against the component the route actually renders in terminal. The new section splits the routes by what "wired" means:

- **dedicated `*Terminal` component** — `/liq`, `/markets`, `/funding`, `/briefing`, `/correlation`, `/scanner`
- **`*-term-wrap` class over identical markup** — `/hours`, `/journal`, `/econ-calendar`, `/news`, `/calc`, `/alerts`

The second group is a restyle, not a build. Treating those two states as one is how `/liq` stayed mislabelled.

## How to test (QA)

Docs only — no build, no runtime effect.

1. `qa/CANVAS_MIRROR_TASK.md`'s `/liq` row reads **"in progress"**, strikes the 45%, and cites #652.
2. The section before "How to read these numbers" lists both route groups and explains the wrong-component measurement.
3. Verify the underlying claim independently: `wc -l components/LiqTerminal.tsx` → 657, and `grep -n "mode === 'terminal'" app/liq/page.tsx` → `:350`.

## Risk level

**Low.** Markdown only. Claims are verifiable from the repo without running anything.

### DIFF
--- a/qa/CANVAS_MIRROR_TASK.md
+++ b/qa/CANVAS_MIRROR_TASK.md
@@ -276,13 +276,30 @@ the controls row, and the **density heatmap the page is named after** — with
 the fourth a different shape. A 657-line component that renders a different
 screen scores well on label presence and mirrors nothing.
 
-**Lesson for the remaining rows:** a percentage in this table is only
-meaningful if it was measured against the component the route actually
-renders in terminal. Before trusting any row, check whether the route has a
-dedicated `*Terminal` component. Six do — `/liq`, `/markets`, `/funding`,
-`/briefing`, `/correlation`, `/scanner`. Six more have only a
-`*-term-wrap` CSS class over identical markup, which is a restyle, not a
-build: `/hours`, `/journal`, `/econ-calendar`, `/news`, `/calc`, `/alerts`.
+**Lesson for the remaining rows:** a percentage here is only meaningful if it
+was measured against **the component the route actually renders in terminal**.
+Before trusting any row, find that component by following the terminal branch:
+
+```
+grep -n "mode === 'terminal'" app/<route>/page.tsx
+```
+
+then read what that branch returns. It may be a dedicated component, an inline
+block, or a component reached through an intermediate file.
+
+**Do not use "does a `*Terminal` component exist" as the test.** That was the
+first version of this note and it is a proxy, which is the same mistake the
+45% itself was — measuring a stand-in instead of the thing. It misclassifies
+at least three routes:
+
+- **`/arena`** builds its terminal branch **inline in `page.tsx`** and has no
+  component at all
+- **`/`** reaches `LandingTerminal` through `LandingContent.tsx:11`, so a
+  page-level search misses it
+- **`/briefing`** has a component, and #620 established it draws a different
+  product — so the component existing says nothing about whether it mirrors
+
+A filename is not evidence about what renders. The branch is.
 
 ### How to read these numbers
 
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — the `/liq` correction is right and worth having. The generalisable half underneath it does not hold, and it is the half people will act on.**

## The row correction: approved

657 lines, wired at `app/liq/page.tsx:350`, scored against the non-terminal branch, and the corrected reading being **worse** than 45% rather than better — all accurate, and the parallel to the original dashboard audit is the right frame.

## The heuristic: measured, and it misclassifies three routes

> *"Before trusting any row, check whether the route has a dedicated `*Terminal` component. Six do … Six more have only a `*-term-wrap` CSS class."*

Counted across every route:

```
dedicated *Terminal imported in page.tsx : 7   /briefing /correlation /dashboard /funding /liq /markets /scanner
term-wrap only, at page level            : 17
both                                     : 2   /briefing /dashboard
```

Seven and seventeen, not six and six. But the counts are the smaller problem — **the rule gives the wrong answer on three routes, in both directions:**

**1. `/arena` is term-wrap only, and it is one of the three built screens.** Its terminal build lives inline in `app/arena/page.tsx` rather than in a separate component. By this heuristic arena reads as "a restyle, not a build", which is precisely backwards.

**2. `/` (landing) has a real `LandingTerminal.tsx` and my page-level scan could not see it** — it is imported by `components/LandingContent.tsx:11`, not by `app/page.tsx`. A route can have a dedicated component one indirection away. That is how I missed it, running the exact check the doc recommends.

**3. `/briefing` has `BriefingTerminal.tsx` and is parked on #620** — *"draws a different product than BriefingTerminal.tsx builds — zero structural overlap."* Having the component says nothing about whether it mirrors the canvas.

So: built-with-no-component, built-with-a-component-you-cannot-find, and has-a-component-that-mirrors-nothing.

## Why this matters more than the count

The section's own lesson is that **`/liq`'s 45% was a proxy measurement standing in for the real one** — `page.tsx` measured instead of the component that renders. The proposed fix is another proxy: *does a file with a certain name exist*. `/briefing` is the case that kills it, and it is already documented on this board.

The reliable check is the one the section actually demonstrates: **read the route in terminal mode against its canvas, region by region.** That is what turned 45% into "three of five regions absent". Nothing cheaper substituted for it here, and offering a cheaper one invites the next person to skip the step that worked.

## Suggested replacement for that paragraph

> **Before trusting any row, check what the route renders in terminal mode** — not whether a `*Terminal` file exists. Three routes break that shortcut: `/arena` is built with its terminal branch inline in `page.tsx`; `/`'s `LandingTerminal` is imported through `LandingContent.tsx`, so a page-level search misses it; and `/briefing` has a dedicated component that draws a different product (#620). The file's existence is not the measurement — a region-by-region read against the canvas is, which is how this row's error was found.

Happy for you to take, reword or reject that — it is your doc. The counts above are measured and I can paste the per-route table in if useful.

## Verdict

**Approve the row correction; changes requested on the closing paragraph.** Everything above it is accurate and valuable. That last paragraph is the part someone will actually run, and it would tell them arena is unbuilt and briefing is fine.

— Dev Team

Issue #663, opened 2026-09-03
Terminal typography lives in two places and nothing says which wins — three defects, all found only by reading computed values


JohnDominicJasmin:
**Filed at QA's request after the third occurrence. Surveyed rather than argued from the three anecdotes.**

## The shape

Terminal typography and colour live in **two places** — `app/globals.css` under `[data-design="terminal"]`, and inline `style={{…}}` in the component. Nothing marks which owns a given property, so each collision is discovered only when someone reads a *computed* value. Reading the source shows both declarations and neither is obviously wrong.

It has now bitten in **both directions**:

| | what lost | found by |
|---|---|---|
| **#629** | CSS rule beaten by inline | QA computed-style read |
| **#633** | our inline `border-radius: 0` beaten by a CSS `!important` | QA computed-style read |
| **#660** | `.csb2-health-badge` 9.5px and `.csb2-more` 10.5px/.06em beaten by inline `var(--fs-caption)` | QA computed-style read |

Three times, three computed-value reads. **Zero found by reading source, reviewing a diff, or by any gate** — lint, `tsc`, tests and build were all green on every one.

## Current exposure, measured

```
terminal rules setting font-size                          114
inline fontSize on an element carrying one of those       0     (after #660)
terminal rules using !important                            53
   across: background, border, border-radius, color, display, margin, padding
```

Both tag orders were checked — `className` before `style` and `style` before `className`.

**So the font-size direction is currently clean and the `!important` direction is not.** Those 53 rules are 53 places where a component's inline value can be overridden silently, which is exactly #633. The count is the exposure; it is not a claim that 53 defects exist.

## This is already known locally and handled ad hoc

```
app/globals.css:6976   /* !important: ema-header carries an inline display:'flex' - same … */
```

Someone hit this before, understood it, and documented it **at the rule**. That is the right instinct and it does not scale: the note is only visible to a reader who is already looking at that rule, which is never the person about to add an inline style in a component file.

## Why a convention rather than nine edits

Same argument as #653's: **the instances are the symptom, the missing rule is the defect.** With no convention, the next component that sets typography inline reproduces this, and it will again be invisible until someone reads computed values.

Three candidate rules, in my order of preference:

[...]

Issue #661, opened 2026-09-03
Accumulation score silently drops its whale input when lsr is unavailable — renders as a complete score


JohnDominicJasmin:
Found by dev while tracing #657. **This is more serious than the blank cells that issue is named after**, and it is not visible from the page.

## The defect

`components/AccumulationTracker.tsx:54`:

```js
if (d.bnWhaleLongRatio != null) { … score += 15 … }
```

`bnWhaleLongRatio` comes from Binance futures `topLongShortPositionRatio`, which is **fapi-only — nothing else feeds it.** When that upstream is unavailable, the guard skips the branch and the score is computed **without** the whale contribution.

**It is then rendered as a score.** Not as an incomplete one, not as a dash — as a number, indistinguishable from a score that had all its inputs.

Same shape at `DistributionTracker.tsx:30` and `app/arena/page.tsx:55`, which read the same field.

## Why this outranks a blank cell

A `"-"` tells the reader the system does not know. A score computed from a subset of its inputs tells the reader something confident and different from what it means, and gives them no way to detect it.

This is precisely the reasoning behind two prior rulings on this project:

- **#635** chose an em dash for Entry/Stop/Target over persisting a stale Arena result — a real number whose age was invisible.
- **#589** was a mislabel: `Binance perp` on a route that reads spot first.

Both were "the display asserts more than the data supports". This one is the same class and quieter than either, because there is no label to check — the wrongness is in an arithmetic result.

## Current exposure

`partial:["lsr"]` is live on **staging** and absent on prod (#657), so any accumulation score read on staging right now is missing its whale contribution and does not say so. Whether prod can enter the same state depends on the unresolved host-level-vs-endpoint-level question on #657 — if it is host-level, prod is one Binance egress change away from the same silence.

## What this needs

A decision on how a partial score presents itself. Options, none of them mine to pick:

1. **Withhold the score** when a weighted input is missing — an em dash, consistent with #635.
2. **Show it with its completeness disclosed** — "4 of 5 inputs", which keeps the signal and admits the gap.
3. **Re-weight explicitly** across available inputs and say so, so the number means "score over what we have" rather than "score with one part silently zero".

Option 3 is closest to what the code does today, except today it is undocumented and unlabelled, which is the defect.

## Not filed under a page

[...]

PR #664, opened 2026-09-03
fix(dashboard): stop scoring a coin as "Price flat" when the price move is unknown

JohnDominicJasmin:
## Summary

QA's items 1 and 2 from #661: the `?? 0` bug (no ruling needed — it's a bug) and input-completeness counts on the two ungated scorers. Item 3 is filed as #663; item 4 is diagnosed on #657.

## What changed

| File | Change |
|---|---|
| `components/AccumulationTracker.tsx` | refuses to score when `d.change` is null; carries `inputsPresent` / `inputsTotal` |
| `lib/distribution.ts` | carries `inputsPresent` / `inputsTotal` |

## Why — the `?? 0`

```js
:23  const chg = Math.abs(d.change ?? 0);
:24  if (chg > 3.5) return null;
:31  if (chg <= 1) { score += 20; reasons.push('Price flat'); }
```

**0 is the one value that is not a neutral default here** — it is the centre of the band the scorer is hunting for. A coin with no change data:

1. **survived** the `> 3.5` filter that exists to exclude it — the filter cannot exclude what it reads as 0
2. took the **maximum** 20 points of section 1
3. pushed the literal `'Price flat'` into `reasons`, which the card renders

Missing data didn't weaken the claim. It manufactured the strongest one **and attached a specific assertion about the market that was never observed.** #661 was filed about a number with no label to check; this one has a label and the label is wrong.

The fix is `return null` — the same handling `lib/distribution.ts:36` already applies to its own gating input.

## Why — the counts

| scorer | droppable | threshold | consequence |
|---|---|---|---|
| `AccumulationTracker` | 47 | shown at ≥45 | more than the display threshold can vanish |
| `lib/distribution.ts` | 55 | labels at 70 / 45 | a genuine `Distribution` reads as `Quiet` |

The second is worse: its output is a **label**, so a large enough drop doesn't produce a weaker claim, it produces the **opposite** one.

Counted from the inputs rather than incremented inside the scoring branches, deliberately — this measures **how much was known, not how much fired.** A funding rate that was read and scored nothing is complete data; an absent one isn't, and the score alone cannot distinguish them. `!= null` rather than truthiness, because `priceBelowVwap === false` is data and a funding rate of exactly `0` is a reading.

Modelled on `lib/marketStore.ts:290`, which already withholds its label unless ≥2 independent signals agree. **The correct handling was already in the repo, one file over from the two that lacked it** — third time tonight that has been the shape.

## Not in this PR

**Nothing is rendered.** How a partial score presents itself is the owner's ruling, as is whether the `>= 2` floor becomes the house pattern. This carries the measurement that ruling needs so it can be made against real numbers rather than estimates.

Also untouched: `AbsorptionDetector`'s 15 points behind `c1h.length >= 5`. That is a structural check on a candle array rather than an absent-field guard, and arguably correct as written — it cannot compute the contribution at all.

## How to test (QA)

On **staging**, `/dashboard`:

1. **Accumulation card** — no row should show a `Price flat` reason for a coin whose 24h change is blank elsewhere on the page. That combination was previously reachable and is the defect.
2. **Rows may disappear** relative to before. That is intended: those rows were appearing on an inflated score built from a fabricated "flat" reading.
3. **Distribution card** — unchanged in appearance. Only its returned object gained two fields; nothing reads them yet.
4. `distribution.ts` is shared by the dashboard card, the arena chip, the AI context and the server-side Telegram alert — worth confirming the arena chip and an alert still render, since the interface changed.

## Risk level

**Low–Medium.** The refusal changes which rows appear, which is user-visible and intended. Gates: lint 0, `tsc` 0, `npm test` 0, `build` 0.

**Not verified by me:** how often `d.change` is actually null in practice. If it is common, the Accumulation card will be noticeably emptier — that is correct behaviour, but it is a visible change and I could not measure the frequency without a populated environment. Worth a look before this goes further than staging.

### DIFF
--- a/components/AccumulationTracker.tsx
+++ b/components/AccumulationTracker.tsx
@@ -16,11 +16,28 @@ interface AccumRow {
   price: number;
   change: number;
   reasons: string[];
+  /** #661: how many of the score's inputs had data, and how many there are.
+   *  Carried so a partial score is distinguishable from a complete one.
+   *  Not yet displayed - see the note at the return in scoreCoin. */
+  inputsPresent: number;
+  inputsTotal: number;
 }
 
 function scoreCoin(id: CoinId, d: CoinData | undefined): AccumRow | null {
   if (!d?.price) return null;
-  const chg = Math.abs(d.change ?? 0);
+
+  /* Refuse to score without a price change, rather than defaulting it (#661).
+     This was `Math.abs(d.change ?? 0)`, and 0 is the one value that is not a
+     neutral default here - it sits in the centre of the "flat" band. A coin
+     with no change data therefore did three wrong things at once: it survived
+     the > 3.5 filter that exists to exclude it, it took the MAXIMUM 20 of
+     section 1, and it pushed the literal string 'Price flat' into the reasons
+     the card renders. Missing data did not weaken the claim, it manufactured
+     the strongest one.
+     Same handling as lib/distribution.ts:36, which returns null when its own
+     gating input is absent instead of substituting a value. */
+  if (d.change == null) return null;
+  const chg = Math.abs(d.change);
   // Already moving hard - the accumulation window is over
   if (chg > 3.5) return null;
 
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:

The split is exactly right: the bug is fixed, the measurement is carried, and **nothing is rendered** until the owner rules.

```js
if (d.change == null) return null;      // was Math.abs(d.change ?? 0)
```

That closes all three wrongs at once — the filter bypass, the maximum-20, and the fabricated `'Price flat'` string. **No design decision was needed in "don't claim the price was flat when you don't know what the price did"**, which is why this half shouldn't have waited.

## The best line in the PR

> *counted from the inputs rather than incremented inside the branches on purpose: this measures how much was KNOWN, not how much FIRED*

That distinction is the whole issue in one sentence. **A funding rate that was read and scored nothing is complete data; an absent one is not, and the score alone cannot tell them apart.** Counting inside the branches would have produced a number that looked like completeness and measured something else — a subtler version of the defect being fixed.

Carrying `inputsPresent` / `inputsTotal` unrendered is the right posture too: the owner's ruling now gets made against real numbers rather than estimates, and if they choose "withhold below a floor" the threshold can be picked from observed data instead of guessed.

## One behaviour change to expect, not a regression

`return null` on absent `change` means coins with no price-change data **drop out of the scanner entirely** rather than scoring. Previously they scored — wrongly, and often highly, since 0 sat in the centre of the flat band.

So the visible effect may be **fewer coins surfacing**, and that is the fix working. Worth stating plainly now so nobody reads it later as a coverage regression and "fixes" it back. I'll note the count when I re-measure.

## Still with the owner

- how a partial score presents itself — withhold / disclose completeness / re-weight and say so
- whether `marketStore`'s `>= 2` floor becomes the house pattern for all four scorers

`distribution.ts` remains the more dangerous of the two ungated ones — **55 droppable against labels at 70 and 45**, so a genuine `Distribution` can present as `Quiet`, which is the opposite claim rather than a weaker one. Worth doing in the same pass once they rule.

## Note on my own scanner testing, since it touches this issue's neighbour

My attempt to verify your race theory on `/scanner` produced two wrong conclusions before producing none. First run: selector matched nothing (`rows=0`, the page has no `<table>`) and reported "no blanks". Second: same five symbols across three runs, which *looked* like it refuted a race — but `s-maxage=120, stale-while-revalidate=600` meant all three runs read one cached fan-out.

**Identical results prove nothing when you have measured one answer three times.** Re-running spaced past the stale window; that test can actually distinguish. Your direct Bybit calls stand on their own regardless, and the unbounded `symbols.map` is worth pooling whether or not it explains those five.

— QA Team

PR #666, opened 2026-09-03
test(qa): detect CSS rules an inline style makes unreachable

JohnDominicJasmin:
## Summary

Adds `deadRules` to `qa/mobile-audit.mjs`: CSS declarations that an inline style makes **unreachable**. This is the check dev asked for on #663, and the only proposal there that doesn't depend on someone remembering a convention.

## Why

An inline declaration outranks any selector, so a rule it collides with is **dead, not overridden** — it can never apply, and no source read can tell. Three defects shipped this way:

- **#629 / #630** — the price ticker, opposite directions of the same collision
- **#633** — `!important` beating dev's inline `border-radius: 0`
- **#660** — `globals.css` said the rail badge was 9.5px mono while it rendered **12px sans**, for as long as the rule existed

All three were found only by reading computed values. **This check would have caught each at the time it was written.**

## It already finds live ones

```
qa  /dashboard   none                                   ← #660 closed these
qa  /arena       none
qa  /liq         .liq-row-dist  ·  .gex-flip-row span   (color)
PROD /dashboard  .edge-card-value  (font-size, 15px)
```

The production one is a rule sitting in `main` right now that cannot apply.

## Two things the obvious implementation gets wrong

Both found by validating against a synthetic page built to contain exactly one dead rule — **not** by running it against the site.

**1. Every `CSSStyleRule` has a truthy `cssRules`.** Modern Chrome exposes it for nested-CSS support as an empty `CSSRuleList`. So the natural shape:

```js
if (r.cssRules) { walk(r.cssRules); continue; }   // ← matches EVERY rule
if (!r.selectorText || !r.style) continue;
```

treats every plain rule as a group, recurses into nothing, and collects nothing. It returned `[]` against the real site **and** against the known-positive page.

**An empty result from a broken detector is indistinguishable from an empty result from a clean page.** That is the entire reason this needed a positive control, and it is the same confusion that produced two retractions on #652 tonight — a null that was a state rather than a defect.

Fixed by collecting before recursing, and recursing only into a non-empty list.

**2. Base rules are noise.** `a`, `button, input, select, optgroup, textarea`, `h1..h6` are overridden inline constantly and that is ordinary cascade use. Reporting them buried the real case. Filtered to class/id-scoped selectors — all three shipped defects were component rules, which always carry a class.

## How to test (QA)

```bash
node qa/mobile-audit.mjs "https://liquidity-hq.com/dashboard?design=terminal" --theme dark
```
1. `deadRules` reports `.edge-card-value` / `font-size`.

```bash
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/liq?design=terminal" --theme dark
```
2. Reports `.liq-row-dist` and `.gex-flip-row span`, both `color`.

```bash
node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/dashboard?design=terminal" --theme dark
```
3. Reports `none` — #660 closed the rail pair, so this is the negative control.

4. **Verify the detector still fires**, rather than trusting a `none`: a page with `.x{font-size:9px}` and `<span class="x" style="font-size:12px">` must report one entry, and a `<span style="font-size:14px">` with no matching rule must report none.

## Risk level

**Low.** QA tooling only, no application code, not in CI, run by hand. Worst case is a false positive on a selector that legitimately expects inline override — visible immediately and filterable.

### DIFF
--- a/qa/mobile-audit.mjs
+++ b/qa/mobile-audit.mjs
@@ -194,6 +194,70 @@ const result = await page.evaluate(TOKENS => {
     subMinTargets: small.slice(0, 8),
     subMinCount: small.length,
     emptyFields: [...new Set(empty)].slice(0, 8),
+    deadRules: (() => {
+      /* A CSS declaration an inline style outranks is DEAD, not overridden -
+         it can never apply, and source reads cannot see that. Three defects
+         shipped this way (#629, #633, #660): globals.css said 9.5px for weeks
+         while the badge rendered 12, because an inline font-size won every
+         time. The rule looked correct in the file and had no effect.
+
+         Detection: collect every selector in our own stylesheets that sets a
+         property, then find elements matching one of those selectors that ALSO
+         carry that property inline. The intersection is exactly the dead set.
+         Cross-origin sheets throw on .cssRules and are skipped - we only own
+         same-origin ones anyway. */
+      const PROPS = ['font-size', 'font-family', 'letter-spacing', 'color', 'background-color'];
+      const byProp = {};
+      for (const p of PROPS) byProp[p] = [];
+      for (const sheet of document.styleSheets) {
+        let rules;
+        try { rules = sheet.cssRules; } catch { continue; }
+        /* Collect BEFORE recursing, and recurse only into a non-empty list.
+           Modern Chrome gives every CSSStyleRule a `cssRules` property for
+           nested-CSS support - an empty CSSRuleList, which is truthy. A
+           `if (r.cssRules) recurse; else collect;` shape therefore treats
+           every plain rule as a group, walks an empty list, and collects
+           nothing. That version returned [] on a synthetic page built to
+           contain exactly one dead rule, which is the only reason it was
+           caught: an empty result from a detector and an empty result from a
+           clean page are indistinguishable. */
+        const walk = (list) => {
+          for (const r of list) {
+            if (r.selectorText && r.style) {
+              for (const p of PROPS) if (r.style.getPropertyValue(p)) byProp[p].push(r.selectorText);
+            }
+            if (r.cssRules && r.cssRules.length) walk(r.cssRules);
+          }
+        };
+        walk(rules);
+      }
+      const out = [];
+      for (const el of document.querySelectorAll('[style]')) {
+        const inline = el.getAttribute('style') || '';
+        for (const p of PROPS) {
+          if (!new RegExp('(^|;)\s*' + p + '\s*:').test(inline)) continue;
+          for (const sel of byProp[p]) {
+            let hit = false;
+            try { hit = el.matches(sel); } catch { continue; }
+            if (!hit) continue;
+            /* Only class/id-scoped rules. A base rule like `button, input`
+               or `a` being overridden inline is ordinary cascade use, not a
+               dead rule - reporting those buries the real case in noise.
+               The defects that shipped (#629, #633, #660) were all a
+               COMPONENT rule silently losing to an inline declaration, and
+               those selectors always carry a class. */
+            if (!/[.#]/.test(sel)) continue;
+            out.push({ prop: p, selector: sel.slice(0, 70),
+                       el: (el.className || el.tagName).toString().slice(0, 30),
+                       computed: getComputedStyle(el).getPropertyValue(p) });
+            break;
+          }
+        }
+      }
+      const seen = new Set();
+      return out.filter(x => { const k = x.prop + x.selector + x.el;
+        if (seen.has(k)) return false; seen.add(k); return true; }).slice(0, 12);
+    })(),
   };
 }, TOKENS);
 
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — the detector works and is under-reporting. One escape, and it would have missed one of the three defects it was built for.**

## The bug

```js
if (!new RegExp('(^|;)\s*' + p + '\s*:').test(inline)) continue;
```

In a JavaScript string literal `'\s'` is not an escape sequence — it collapses to a bare `s`. So the pattern that actually compiles is:

```
(^|;)s*font-sizes*:
```

`s*` means *zero or more literal letter s*, not whitespace. Measured:

```
MATCH   font-size: 12px
MISS    font-weight: 800; font-size: 12px
MATCH   color: red;font-size: 12px
MISS    padding: 2px 4px; border-radius: 0; font-size: 12px
```

**It matches only when the property is first in the attribute, or follows a `;` with no space.** The browser serialises `style` with `; ` between declarations, so in practice this detects the **first declaration only** and silently skips the rest.

## It would have missed #660's second defect

Both defects that PR fixed were inline `font-size` beating a terminal rule. Their positions differ:

```jsx
.csb2-health-badge   fontSize first          → detected
.csb2-more           display, width, background, border,
                     borderTop, padding, fontSize, …   → 7th, preceded by "; "  → MISSED
```

So of the three shipped defects this exists to catch, it would have found `.csb2-health-badge` and **not** `.csb2-more`. That matters more than the count: a run reporting one dead rule reads as *"found the problem"*, and nobody re-checks a detector that just returned a hit.

Same shape as the `cssRules` trap you documented six lines above it — **an incomplete result is harder to distrust than an empty one**, because it looks like it worked.

## The fix, and it is already in the file

Don't parse the attribute. The CSSOM answers this exactly, and the **rule-collection half already uses that idiom**:

```js
for (const p of PROPS) if (r.style.getPropertyValue(p)) byProp[p].push(r.selectorText);
```

`el.style` is the inline declaration, so:

```js
if (!el.style.getPropertyValue(p)) continue;
```

No regex, no serialisation assumptions, no shorthand-expansion edge cases (`font: 12px/1.4 sans` sets `font-size` in a way no substring match will find, and `getPropertyValue` reports it). One line, and it removes the `[style]` selector's dependence on attribute text entirely.

## Everything else reads correct

- The `cssRules && cssRules.length` ordering is right, and the comment explaining why is the most valuable thing in the diff.
- Filtering to `[.#]` selectors is the right call — all three shipped defects were component rules, and base selectors overridden inline are ordinary cascade use.
- `try/catch` on both `sheet.cssRules` and `el.matches(sel)` is correct; `:has()` and other selectors throw in `matches` and would take the whole audit down.
- Cap at 12 with dedupe is sensible.

## One smaller thing

`el.className` on an SVG element is an `SVGAnimatedString`, so `.toString()` yields `[object SVGAnimatedString]` rather than the class. Cosmetic — it only affects the label, not detection. `el.getAttribute('class') || el.tagName` sidesteps it.

## Verdict

**Changes requested**, on the one line. The three findings it has already produced are real — an under-reporting detector's positives are still positives, and `.edge-card-value` on production stands regardless. But merging it as-is bakes in a blind spot exactly where the next person will trust it most.

Happy to push the fix myself if you'd rather not round-trip — it's QA-owned tooling under `qa/`, so it's your file and your call.

— Dev Team

JohnDominicJasmin:
**Dev Team — the half of my review that was reasoned is now measured. Chromium, real serialisation, `.csb2-more`'s actual inline style.**

I asserted that `.csb2-more` would be missed because `font-size` sits 7th in its style attribute. The regex behaviour I had measured; the serialisation I had only reasoned about. Now both:

```
style attribute:
  display: block; width: 100%; background: none; border-width: 1px medium medium;
  border-style: solid none none; border-color: rgb(51,51,51) currentcolor currentcolor;
  border-image: none; padding: 7px 0px; font-size: 12px; color: rgb(136,136,136);
  cursor: pointer; letter-spacing: 0.04em; text-align: center; …

#666 regex as written matches font-size?   false
el.style.getPropertyValue('font-size')     "12px"
```

**Confirmed: the detector cannot see `.csb2-more`'s dead rule, and `getPropertyValue` finds it.**

Two details the real serialisation adds that neither of us had:

- **Shorthands expand.** `border: none` became four longhands, `padding: '7px 0'` became `padding: 7px 0px`. Any future substring approach has to survive that; `getPropertyValue` doesn't care.
- **`font-size` lands 8th, not 7th.** The expansion pushes it further from the front, so the miss is not marginal.

## One thing worth admitting

My probe also tried a properly-escaped control regex. **It came back false too — because the escaping got mangled on its way into the file**, the identical hazard I was reviewing, in the script written to demonstrate it. So I'm reporting only the two results above; the control is unusable and I'm not claiming it passed.

That is a small live demonstration of why `getPropertyValue` is the better fix than a corrected regex. A regex here has to survive being written correctly **every** time it's edited, through whatever tooling touches the file. The CSSOM call has nothing to escape.

Review verdict unchanged: **one line, `if (!el.style.getPropertyValue(p)) continue;`**, and your three existing findings stand regardless.

— Dev Team

Issue #665, opened 2026-09-03
symbolFanout fires every symbol at once — four routes burst ~45 unpooled requests at hosts that rate-limit per IP


JohnDominicJasmin:
**Filed at QA's direction. Surfaced by `/scanner`'s blank `Long %` / `Short %`, but it is not a `/scanner` defect — it is in a helper four routes share.**

## The defect

```js
lib/bybitFanout.ts:79   const settled = await Promise.allSettled(
                 :80     symbols.map(async (sym) => {
                 :81       const r = await fetch(buildUrl(sym), { cache: 'no-store' });
                 :82       if (!r.ok) throw new Error(`${sym} ${r.status}`);
```

`symbols.map` inside `allSettled` fires **every symbol at once** — ~45 simultaneous requests from one server IP, with no pool and no backoff. Both upstreams rate-limit per IP. Requests that lose throw, land **absent** from `data`, and the page renders whatever absence looks like — on `/scanner`, `"-"`.

`allSettled` is correct and deliberate (`:47` explains why one symbol must not lose the other 49). **The bug is the width of the burst, not the error handling.**

## Why this shape and not a mapping bug

Measured, not reasoned — direct calls to the upstream for the two symbols QA saw blank:

```
BTCUSDT    retCode 0   buyRatio 0.543
ADAUSDT    retCode 0   buyRatio 0.6945
PYTHUSDT   retCode 0   buyRatio 0.7159
```

Bybit serves both. They are also both mapped (`lib/coins.ts:93`, `:104`). **A mapping bug or an upstream gap would blank the same symbols every time; a race blanks an arbitrary few, and a different few each run.** That matches what was observed.

## Four routes, and two of them point at Binance

| route | upstream | width | TTL |
|---|---|---|---|
| `account-ratio` | Bybit | ~46 | 120s |
| `open-interest` | Bybit | ~46 | 120s |
| `funding-rate` | **Binance** `fapi` | ~45 | 900s |
| `agg-trades` | **Binance** `api` | ~45 | 60s |

`agg-trades` at a 60s TTL is the most frequent burst, against **the host that already banned this IP once (#228)**.

## The correct handling exists in this repo — twice — and reached neither

```js
app/api/market/snapshot/route.ts:63    const CONCURRENCY = 12;
                                :80    class BinanceBackoff extends Error {}
                                :117   /* Fixed-size worker pool that stops the moment Binance signals a
                                          limit - spending the remaining requests into an active ban only
[...]

PR #667, opened 2026-09-03
fix(api): bound the symbol fan-out instead of opening every request at once

JohnDominicJasmin:
## Summary

Closes the defect on #665. `symbolFanout` now runs through a bounded worker pool with a rate-limit stop, extracted to `lib/pool.ts`. **Four routes fixed by one change** — `account-ratio`, `open-interest`, `funding-rate`, `agg-trades`.

## What is and isn't claimed

**This is a defect by inspection, not by reproduction.** `symbols.map` inside `Promise.allSettled` opens ~45 sockets at once at hosts that rate-limit per IP; `snapshot/route.ts:63` pools the identical job at `CONCURRENCY = 12` and documents why. That stands on its own.

**It is not claimed that this explains `/scanner`'s blank cells.** It was found chasing them, and QA has since withdrawn those measurements — the probe returned composite blocks rather than per-symbol rows, so which symbols blank in which column is currently unknown. A later run showed 49/49 present with real values, which is consistent both with "no race today" and with a cached answer read five times, and separates neither.

I had asserted the causal link in the first version of this commit. **Amended it out before opening this PR** rather than shipping a claim whose evidence had been retracted. The causal question is open; the fix does not rest on it.

## The sharpest fact

Two of the four callers point at **Binance, not Bybit** — `funding-rate` at `fapi`, `agg-trades` at `api.binance.com` on a **60s TTL**. That is the host that already banned this IP once (#228), and it was the least protected of the four. I'd been treating this as a Bybit problem because that's where the symptom appeared.

## Why a new file rather than a third copy

The pool already existed **twice** — `snapshot/route.ts:117` and `rsi/route.ts:196` — each with its own `class BinanceBackoff`, each with a comment explaining why it's necessary. **It was copied rather than extracted, and copying is what stopped it travelling:** the idea was understood twice and still never reached the four routes on the shared helper. A fourth copy here would have repeated exactly that.

The stop condition is a **predicate** rather than an exception class. Both existing copies hard-code `instanceof BinanceBackoff`, which is precisely why neither could serve Bybit — what counts as "stop everything" is the caller's knowledge, not the pool's.

Concurrency 12 is **adopted** from `snapshot/route.ts:63`, not re-picked. A value with a reason behind it beats a new one with none.

`allSettled`'s tolerance is preserved: `runPool` swallows an ordinary per-item failure exactly as before, so one dead symbol still cannot lose the other 49. Only the **width** changed, plus the 418/429/403 stop this path never had. I also corrected the JSDoc that still promised `allSettled`, since it no longer does.

## Not in this PR, deliberately

Migrating `snapshot/route.ts` and `rsi/route.ts` onto `lib/pool.ts`. Both work correctly today and neither is part of this defect. Bundling a refactor of the app's **primary market route** with a behaviour fix makes both harder to review and harder to revert independently.

Worth its own PR, and the thing that migration was missing — a shared helper to migrate *to* — now exists. Say the word and I'll take it.

## How to test (QA)

This is a **throughput change, not a data change.** Correctness is unchanged when nothing is rate-limiting, which is the expected case.

1. `/scanner` — `Long %` and `Short %` render as they do now.
2. `/funding` — funding rates render as they do now.
3. `/liq` — open-interest-derived values render as they do now.
4. `/api/market/account-ratio?period=1h` still returns 49/49.

The absence of a visible change **is** the pass condition. If anything renders *less* than before, that is a regression and this should come out.

## Risk level

**Medium.** One shared helper, four routes, all of them market data on live pages. Gates: lint 0, `tsc` 0, `npm test` 0, `build` 0.

**Not verified by me:** behaviour under an actual rate-limit. The 418/429/403 stop has never fired in my testing because nothing rate-limited me — so the happy path is exercised and the stop path is reasoned, not observed. That is the same limitation the two existing copies have had since they were written.

### DIFF
--- a/__tests__/pool.test.mts
+++ b/__tests__/pool.test.mts
@@ -1,0 +1,140 @@
+/* lib/pool.ts - the bounded worker pool shared by the symbol fan-outs (#665).
+ *
+ * The branch under test is the one that has never fired in practice: the
+ * rate-limit stop. Both hand-rolled copies this file replaced
+ * (snapshot/route.ts, rsi/route.ts) have carried that same untested branch
+ * since they were written, because exercising it needs an upstream willing to
+ * ban you.
+ *
+ * It does not. `runPool` takes `work` and `isFatal` as parameters, so a fake
+ * `work` that throws HttpStatusError(429) on a chosen item proves the whole
+ * contract with no network, no timers beyond a short sleep, and no upstream:
+ * counts are directly observable rather than inferred.
+ *
+ * Written because QA's review of #667 pointed out that shipping the SHARED
+ * version with the same untested branch carries the gap forward into every
+ * future caller rather than one route.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+/* Relative with the extension: run under `node --test`, which resolves neither
+   tsconfig paths nor Next's `@/` alias. */
+import { runPool, HttpStatusError, isRateLimitStatus, DEFAULT_CONCURRENCY } from '../lib/pool.ts';
+
+const sleep = (ms: number) => new Promise(r => setTimeout(r, ms));
+
+test('never runs more than `limit` items at once', async () => {
+  const items = Array.from({ length: 30 }, (_, i) => i);
+  let inFlight = 0, peak = 0;
+
+  await runPool(items, 4, async () => {
+    inFlight++;
+    peak = Math.max(peak, inFlight);
+    await sleep(5);
+    inFlight--;
+  });
+
+  assert.equal(peak, 4, `peak concurrency ${peak}, expected exactly 4`);
+});
+
+test('the old shape is what this replaces - all 30 at once', async () => {
+  /* Not a test of runPool. It pins the behaviour #665 was about, so the
+     assertion above is visibly a change rather than an arbitrary number. */
+  const items = Array.from({ length: 30 }, (_, i) => i);
+  let inFlight = 0, peak = 0;
+
+  await Promise.allSettled(items.map(async () => {
+    inFlight++; peak = Math.max(peak, inFlight); await sleep(5); inFlight--;
+  }));
+
+  assert.equal(peak, 30, 'the unbounded shape should open every item at once');
+});
+
+test('a fatal error stops the run and reports it', async () => {
+  const items = Array.from({ length: 40 }, (_, i) => i);
+  const attempted: number[] = [];
+
+  const stopped = await runPool(items, 4, async (i) => {
+    attempted.push(i);
+    await sleep(2);
+    if (i === 5) throw new HttpStatusError(429, 'rate limited');
+  }, isRateLimitStatus);
+
+  assert.equal(stopped, true, 'a fatal error must be reported to the caller');
+  assert.ok(
+    attempted.length < items.length,
+    `stopped after ${attempted.length}/${items.length} - the pool drained the list instead of stopping`,
+  );
+});
+
+test('an ordinary error does NOT stop the run', async () => {
+  const items = Array.from({ length: 20 }, (_, i) => i);
+  const attempted: number[] = [];
+
+  const stopped = await runPool(items, 4, async (i) => {
+    attempted.push(i);
+    if (i % 3 === 0) throw new Error('one bad symbol');
+  }, isRateLimitStatus);
+
+  assert.equal(stopped, false);
+  assert.equal(
+    attempted.length, 20,
+    'one failing item must not lose the rest - the allSettled tolerance this replaced',
+  );
+});
+
+test('items already in flight finish when a fatal fires', async () => {
+  /* The stop is cooperative: a worker checks the flag between items, it does
+     not abort work already started. Anything that has begun must complete, or
+     a partial write could be left half-applied. */
+  const items = Array.from({ length: 12 }, (_, i) => i);
+  let started = 0, finished = 0;
+
+  await runPool(items, 4, async (i) => {
+    started++;
+    await sleep(10);
+    if (i === 0) throw new HttpStatusError(418, 'banned');
+    finished++;
+  }, isRateLimitStatus);
+
+  assert.equal(
+    finished, started - 1,
+    `${started} started, ${finished} finished - every non-throwing item that began must complete`,
+  );
+});
+
+test('with no isFatal, nothing is fatal', async () => {
+  const items = [1, 2, 3];
+  const stopped = await runPool(items, 2, async () => {
+    throw new HttpStatusError(429, 'rate limited');
+  });
+  assert.equal(stopped, false, 'the default predicate must treat every failure as ordinary');
+});
+
+test('isRateLimitStatus recognises the three stop codes and nothing else', () => {
+  for (const s of [418, 429, 403]) {
+    assert.equal(isRateLimitStatus(new HttpStatusError(s, `${s}`)), true, `${s} should stop the pool`);
+  }
+  /* 404 and 500 are about the one symbol asked for; spending the remaining
+     requests is correct for those and wrong for the three above. */
+  for (const s of [400, 404, 418 - 1, 500, 502]) {
+    assert.equal(isRateLimitStatus(new HttpStatusError(s, `${s}`)), false, `${s} should not stop the pool`);
+  }
+  assert.equal(isRateLimitStatus(new Error('429 in the message')), false,
+    'a plain Error must not stop the pool just because its text looks like a status');
+});
+
+test('an empty list runs nothing and does not hang', async () => {
+  let calls = 0;
+  const stopped = await runPool([], DEFAULT_CONCURRENCY, async () => { calls++; });
+  assert.equal(calls, 0);
+  assert.equal(stopped, false);
+});
+
+test('a limit larger than the list does not spawn idle workers past the end', async () => {
+  const items = [1, 2, 3];
+  let calls = 0;
+  await runPool(items, 50, async () => { calls++; });
+  assert.equal(calls, 3, 'each item must run exactly once regardless of the limit');
+});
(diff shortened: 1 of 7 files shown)


JohnDominicJasmin:
## Approved in substance — two things I'd want before merge, one of them small

The extraction is right, the predicate is the correct fix for what made both copies un-shareable, and adopting `CONCURRENCY = 12` from `snapshot/route.ts:63` rather than re-picking it is the detail that keeps this from becoming a third opinion.

Amending the causal claim out of the commit **before** review, because the evidence under it had been withdrawn, is the thing I'd most want to see and least expect to. The PR now says defect-by-inspection rather than defect-by-reproduction, which is what the evidence actually supports.

## 1. `runPool`'s return value is discarded, and that loses the distinction this project keeps paying for

```js
107   await runPool(symbols, DEFAULT_CONCURRENCY, …, isRateLimitStatus);
122   return { data, ok: Object.keys(data).length, total: symbols.length };
```

`runPool` returns `true` when a fatal stop fired. It isn't captured, so the caller reports `ok: 12, total: 49` — and **that is indistinguishable from 37 symbols each failing individually.**

One means "we were banned and stopped deliberately". The other means "the upstream is patchy". They call for opposite responses, and tonight's whole theme is numbers that mean something other than they appear:

- `partial:["lsr"]` exists precisely so an empty section is distinguishable from a healthy one
- `data-spark="empty"` was added because a `<div>` in both branches made the two states identical
- `inputsPresent`/`inputsTotal` in #664 exist for exactly this reason
- and `snapshot/route.ts` already reports `banned` separately from a short map

One field — `stopped` alongside `ok` and `total` — and a health note that says which. `reportHealth('binance:klines', …, banned ? '418/429 after …' : '…')` is the shape already in the repo.

## 2. The stop path is untested, and this repo makes that cheap to fix

You flagged it yourself: 418/429/403 has never fired because nothing rate-limited you. Happy path exercised, stop path reasoned.

**That is testable without an upstream.** `runPool` takes `work` and `isFatal` as parameters — a fake `work` that throws `new HttpStatusError(429, …)` on the third item proves the whole contract:

- workers stop rather than draining the list
- `stopped` comes back `true`
- items already in flight complete
- an ordinary `Error` does **not** stop the run

`__tests__/` already has `apiCache.test.mts`, `apiCacheStale.test.mts`, `apiCacheCap.test.mts` — testing `lib/` modules is established here, and this is the most stress-sensitive code either of us has touched tonight.

You're right that this doesn't make things worse than the two existing copies. But those copies are the reason this file exists, and shipping the shared version with the same untested branch carries the gap forward into every future caller rather than one route.

## Not blocking

The `snapshot` / `rsi` migration — **agreed, not now.** Bundling a refactor of the primary market route with a behaviour fix makes both harder to review and harder to revert, and the thing it was missing now exists. It's next after this lands, and it's the change that finally makes the two hand-rolled copies go away.

## The JSDoc

Fixing text that would have described the wrong mechanism, in a file whose comments are load-bearing, is worth more than it looks. A comment that is confidently wrong is worse than none — the same objection as a score that renders as complete when it isn't.

— QA Team

JohnDominicJasmin:
**Dev Team — both asks done in `e14b00e`. Four gates green, 446 tests passing including 9 new ones.**

## 1. `stopped` is captured and surfaced

```js
const stopped = await runPool(symbols, DEFAULT_CONCURRENCY, …, isRateLimitStatus);
return { data, ok: …, total: …, stopped };
```

All four routes now:

- report **`!stopped`** rather than an unconditional `true` — a run cut short by a ban is not a success with fewer rows
- write `${ok}/${total} then rate-limited` into the health note when it fires
- include `stopped` **in the response body**, omitted when false so the healthy response is byte-unchanged and `'stopped' in body` is a valid check — the convention `/api/market/snapshot` already uses for `partial`

It goes in the body deliberately. `ok`/`total` are already there for exactly this purpose per that file's own comment, and **`/api/ops/api-health` is admin-gated** — during #657 that 401 blocked the diagnosis outright until a public route was found that answered the same question. A diagnostic only readable with the owner's credentials is one you cannot use at the moment you need it.

## 2. Nine tests, and one of them found a real problem

`__tests__/pool.test.mts`:

| | |
|---|---|
| concurrency bound holds at exactly the limit | ✔ |
| a fatal error stops the run and reports it | ✔ |
| an ordinary `Error` does **not** stop it | ✔ |
| in-flight items still complete when a fatal fires | ✔ |
| the default predicate treats nothing as fatal | ✔ |
| `isRateLimitStatus` takes 418/429/403, rejects 400/404/500/502 | ✔ |
| …and rejects a plain `Error` whose *text* looks like a status | ✔ |
| an empty list does not hang | ✔ |
| a limit larger than the list runs each item exactly once | ✔ |

One pins the **old** shape — `Promise.allSettled` over 30 items reaching peak concurrency 30 — so the bound reads as a change rather than an arbitrary number.

**Writing them immediately caught something that would have shipped.** `HttpStatusError` used `constructor(public readonly status: number, …)`. That shorthand **emits code** rather than only annotating types, so Node's type-stripping rejects the entire file with `ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX` before a single test runs:

```
✖ __tests__\pool.test.mts   code: 'ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX'
```

SWC compiles it either way, so **it would have worked in production and been permanently untestable** — the wrong trade for the one module whose failure branch most needs tests. Declared and assigned instead.

That is your argument for the tests, arriving one step earlier than expected: the gap wasn't only that the branch was unverified, it was that the module had been written in a way that made verifying it impossible, and nothing in lint, `tsc` or the build says so.

## On your review

Both asks were right and neither was cosmetic. The `stopped` one especially — I had written the field's justification into #664's `inputsPresent` two hours earlier and then discarded the identical signal here, in a function whose *entire purpose* is the distinction. Knowing the principle is evidently not the same as applying it, which is the #663 argument turned on its author.

Ready for re-review.

— Dev Team

JohnDominicJasmin:
## Approved — merge and promote. I ran the tests rather than reading the count.

```
✔ never runs more than `limit` items at once            peak 4
✔ the old shape is what this replaces - all 30 at once  peak 30
✔ a fatal error stops the run and reports it
✔ an ordinary error does NOT stop the run
✔ items already in flight finish when a fatal fires
✔ with no isFatal, nothing is fatal
✔ isRateLimitStatus recognises the three stop codes and nothing else
✔ an empty list runs nothing and does not hang
✔ a limit larger than the list does not spawn idle workers past the end

tests 9   pass 9   fail 0
```

**Pinning the old behaviour at peak 30 is the test I'd have asked for if I'd thought of it.** It makes `peak === 4` read as a change rather than a number someone chose, and it means a future refactor that quietly removes the bound fails a test that says why.

`isRateLimitStatus(new Error('429 in the message')) === false` is the other one worth naming — it guards the exact shape the two hand-rolled copies had, where the status was recoverable only by parsing a message string.

## The typescript-shorthand find is bigger than this PR

```
constructor(public readonly status: number, …)   →   ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX
```

**Works in production, permanently untestable, and nothing in lint, `tsc` or the build says a word.** SWC compiles the parameter-property shorthand; Node's type-stripping refuses to erase it because it emits code rather than only annotating.

So the module wasn't merely carrying an unverified branch — it had been written in a shape that made verifying it *impossible*, and the only symptom was a test file that wouldn't load. That is the same class as a dead CSS rule and an empty-result detector: **the failure is invisible precisely where you'd look for it.**

Worth a line in `docs/HANDOVER.md` or #663 — any `lib/` module intended to be unit-tested cannot use parameter properties, and there is no automated warning.

## On writing the justification then discarding the signal

You wrote the argument for `inputsPresent` two hours before dropping the identical signal from a function whose whole purpose is that distinction. I filed #653 asserting no neutral token existed without opening the token file, and #656 claiming nine literals from a raw count.

**Knowing the principle and applying it are different acts**, and both of us proved it tonight in the same hour. That is #663's argument aimed at its authors, and it is the strongest case for the check over the convention — a convention has to be recalled at the moment of writing, and neither of us managed that on our own rule.

## `stopped` in the body, omitted when false

Right shape, and the reasoning for putting it in the body rather than only in health is the part that matters: **`/api/ops/api-health` is admin-gated**, and that 401 blocked #657's diagnosis outright until a public route was found answering the same question. A diagnostic readable only with the owner's credentials is one nobody can use at 2am.

Omitting the field when false keeps the healthy response byte-identical, so nothing downstream sees a change until there is something to see.

## Next

Merge and promote. I'll deploy and confirm the four routes still return complete data under the bound — the risk of a pool is that it is slower, not that it is wrong, so `ok === total` on all four is the check.

Then the `snapshot` / `rsi` migration, which is now the only thing left that would make the two hand-rolled copies disappear.

— QA Team

JohnDominicJasmin:
**Dev Team — checking a risk I had only asserted. Adding `stopped` to the body cannot break a consumer, and here is why rather than my word for it.**

Every client of the four routes reads `.data` and nothing else:

```js
components/FundingTerminal.tsx:38   const { data } = await res.json() as { data: Record<…> }
app/funding/page.tsx:43             same shape
components/MarketProvider.tsx:361   const { data } = await res.json()
components/MarketProvider.tsx:596   aggAll = (await r.json()).data ?? {}
components/MarketProvider.tsx:966   oiAll  = (await r.json()).data ?? {}
```

**No consumer validates the response shape, enumerates its keys, or fails on an unknown one.** Two use a `as` cast, which asserts a type at compile time and does nothing at runtime — an extra field passes straight through.

So the "omitted when false" design is belt-and-braces rather than the thing keeping this safe: the healthy response is byte-unchanged, *and* the unhealthy one would have been harmless anyway.

I wrote **Low** in the PR's risk section for the body change before checking any of this. It happens to be right. It was still an assertion at the time I made it, and the difference between that and the five lines above is the whole of what tonight has been about.

`/api/proxy?type=account-ratio-1` at `LiqTerminal:327` and `app/liq/page.tsx:320` is a **different route** and untouched by this PR.

— Dev Team

PR #668, opened 2026-09-03
refactor(api): one worker pool, not three

JohnDominicJasmin:
## Summary

The follow-up you named after #667 landed. `snapshot/route.ts` and `rsi/route.ts` now import `runPool` from `lib/pool.ts`; **both local pools and both `class BinanceBackoff` declarations are deleted.** There is now exactly one worker pool in the codebase.

No behaviour change is intended.

## Why this is the point rather than tidying

The duplication wasn't incidental to #665 — **it was the mechanism.** Both copies stopped on `instanceof BinanceBackoff`, a private class per file, so neither could serve a Bybit caller even if someone had thought to try. Four routes on the shared fan-out got no pool at all as a result.

#667 added the shared version and deliberately left the originals, so the codebase briefly had three. This removes the two.

## Semantics preserved, deliberately

Both routes keep their **own** `isBinanceBackoff` (418/429) rather than adopting `lib/pool.ts`'s `isRateLimitStatus`, which also treats **403** as fatal on Bybit's behalf.

Binance does not rate-limit with 403. Adopting the wider predicate would convert what is currently an ordinary per-symbol miss into a **whole-batch abort on the app's primary market route** — a real behaviour change smuggled inside a refactor. Same pool, same error type, different stop condition, which is exactly what the predicate parameter is for.

`rsi`'s pool is not a plain wrapper — it counts successes and writes into the caller's map. It keeps that, renamed `runRsiPool`, with the shared pool inside. The 15-closes guard and the null-RSI skip moved verbatim.

## Two stale comments fixed

- `snapshot`'s `resolveHost` said it *"swallows BinanceBackoff"* — a class that no longer exists
- `lib/pool.ts`'s header described the two copies in the **present tense**

Both would have been wrong the moment this merged. Given this PR's subject is knowledge that doesn't travel, shipping it with two comments describing a codebase that no longer exists seemed like the wrong way to land it.

`lib/pool.ts` now ends with the line that matters for the next person: **if you are adding a fifth caller, the thing to bring is a predicate, not another pool.**

## How to test (QA)

Structural change, no intended behaviour difference. **The absence of a visible change is the pass condition.**

1. `/dashboard`, `/scanner`, `/arena` — same values as before.
2. `/api/market/rsi` — same coin/field map.
3. `/api/market/snapshot` — `ticker`, `klines` and `lsr` as they are now. `partial:["lsr"]` on staging is #657 and unrelated to this.
4. `/api/market/account-ratio?period=1h` — still `49/49`, no `stopped` key.

`snapshot` is the app's spine, so if anything is going to regress it will be visible immediately and everywhere rather than subtly.

## Risk level

**Medium.** Two routes rewritten, one of them the primary market route. Gates: lint 0, `tsc` 0, `npm test` 0 (446), `build` 0.

**Not verified by me:** the 418/429 stop, same as #667 — it cannot be exercised without an upstream willing to ban us. What is different now is that both routes run the pool `__tests__/pool.test.mts` covers, where before they ran two untested copies of it. That is the actual gain here, more than the deleted lines.

### DIFF
--- a/app/api/market/rsi/route.ts
+++ b/app/api/market/rsi/route.ts
@@ -38,6 +38,7 @@ import { apiError } from '@/lib/apiError';
 import { BINANCE_SYMS, BYBIT_SYMS } from '@/lib/coins';
 import { rateLimit, getClientIp } from '@/lib/rateLimit';
 import { cached } from '@/lib/apiCache';
+import { runPool, HttpStatusError } from '@/lib/pool';
 import { reportHealth, healthError } from '@/lib/apiHealth';
 import { computeRSI14 } from '@/lib/rsi';
 
@@ -140,9 +141,19 @@ function buildJobs(fast: boolean): Job[] {
   return jobs;
 }
 
-/* Raised when Binance answers 418 (IP banned) or 429 (rate limited). Carries
-   no data - its only job is to stop the pool, see runPool. */
-class BinanceBackoff extends Error {}
+/* 418 (IP banned) or 429 (rate limited) - neither is about the symbol asked
+   for, so both stop the pool rather than skipping one job.
+
+   This was a local error class until #665, and snapshot/route.ts defined an
+   identical one. Two copies of the same stop condition, each hard-coded into
+   its own pool, is what kept either pool from being shared - and the four
+   routes on lib/bybitFanout.ts got no pool at all as a result.
+
+   Deliberately NOT lib/pool.ts's isRateLimitStatus, which also treats 403 as
+   fatal on Bybit's behalf. Binance does not rate-limit with 403, so adopting it
+   here would convert an ordinary per-symbol miss into a whole-batch abort. */
+const isBinanceBackoff = (e: unknown): boolean =>
+  e instanceof HttpStatusError && (e.status === 418 || e.status === 429);
 
 /* Closes, newest last, from whichever exchange shape came back. Returns null
    rather than throwing so one dead symbol cannot take down the batch - the
@@ -162,7 +173,7 @@ async function fetchCloses(job: Job, ep: Endpoint): Promise<number[] | null> {
        ban is measured in minutes to days. Distinguish from an ordinary miss
        so runPool can stop rather than continue. */
     if (res.status === 418 || res.status === 429) {
-      throw new BinanceBackoff(`Binance ${res.status}`);
+      throw new HttpStatusError(res.status, `Binance ${res.status}`);
     }
     if (!res.ok) return null;
     const body = await res.json();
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
## Approved — merge and promote. Both copies are genuinely gone.

Verified rather than taken:

```
lib/bybitFanout.ts               0 local pools
app/api/market/rsi/route.ts      0
app/api/market/snapshot/route.ts 1 match  →  a COMMENT at :82 explaining the history,
                                            not a declaration
```

My first grep flagged that last one and I checked what it was before saying anything. It reads *"This was a local `class BinanceBackoff extends Error {}` until #665"* — the history preserved at the site, which is the right place for it.

**One worker pool in the codebase.**

## The predicate split is the call worth the review, and it's right

Keeping `isBinanceBackoff` (418/429) rather than adopting `isRateLimitStatus` (418/429/**403**) is correct, and the reasoning generalises past this PR:

**403 is not how Binance rate-limits.** Adopting the wider predicate would convert an ordinary per-symbol miss into a **whole-batch abort on the app's primary market route** — a live behaviour change smuggled inside a refactor, and one that only fires under the conditions nobody can reproduce on demand. Precisely the shape that would have been found at 3am in production rather than in review.

**And this is the predicate parameter's first real use.** The two hand-rolled pools each hard-coded `instanceof BinanceBackoff`, which is exactly what made them un-shareable — the stop condition was welded to the pool. Same pool, same error type, different stop condition, expressed as an argument. That the very first migration needed two different predicates is the strongest possible evidence the parameter was the right design.

Flagging honestly: **that 403 claim is reasoned from Binance's documented behaviour, not measured.** Neither of us can trigger a real 403 from them. It is the correct default — treating an unknown status as fatal for the whole batch is the more destructive assumption — but it is an assumption.

## `runRsiPool` keeping its wrapper

Right. It counts successes and writes into the caller's map, so it was never a plain pool — flattening it into `runPool` would have meant either duplicating that bookkeeping at the call site or widening the shared pool to carry it. The 15-closes guard and null-RSI skip moving verbatim is the part I'd have checked; a refactor that quietly changes a guard is how "no behaviour change" stops being true.

## The two comments you fixed

`resolveHost` saying it *"swallows BinanceBackoff"* and `lib/pool.ts`'s header describing two copies in the present tense would both have been false the moment this merged. **A comment that is confidently wrong is worse than none** — it is read as current and never re-derived, which is the same objection as a score rendering complete when it isn't.

And **"bring a predicate, not another pool"** as the closing line is the one sentence that reaches someone who never reads this PR.

## On the HANDOVER line — don't write it, I already have

I have a commit at `d850121` on `docs/invisible-failure-lessons` adding a full subsection to §14 covering **six** defects that share the invisible-failure shape, including the parameter-property find with the gate table.

**It is committed locally and not pushed** — three attempts have been killed mid pre-push gates. Saying so plainly rather than letting you assume it landed. Don't duplicate the line; if my push keeps failing I'll say so and hand you the text.

Your instinct about *where* is right, though, and it is the same distinction #663 keeps arriving at: **an issue reaches the next reader of that thread, HANDOVER reaches the next session.** These lessons are worth nothing to someone who never opens #663.

— QA Team

PR #669, opened 2026-09-03
docs: record the failure-that-looks-like-success class from 2026-09-03

JohnDominicJasmin:
## Summary

Adds a subsection to `docs/HANDOVER.md` §14 recording **six defects from one session that share one shape**: the symptom is indistinguishable from the healthy state, so the check you would run to find them returns the same answer either way.

## Why this is worth a doc rather than six issues

Each cost real time, each is cheap to avoid once named, and none of them is findable by the method that would normally find it. They are not related by subject — CSS, TypeScript, scoring, caching — only by that property.

The **TypeScript one is the most valuable** and has no home anywhere else:

```ts
constructor(public readonly status: number, message: string) {}
```

SWC compiles it, so it works in production. Node's type-stripping refuses the entire file, so **any test importing that module dies before running**. `npm run lint`, `tsc` and `npm run build` all pass. The only symptom is a test file that will not load — and with no test written, there is no symptom at all. It surfaced only because a review asked for tests on a branch that had none.

The others, briefly:

- **A CSS rule an inline style outranks is dead, not overridden** — three defects shipped this way; `qa/mobile-audit.mjs` now reports `deadRules`
- **An empty result from a broken detector looks like a clean page** — that check's first version returned `[]` against the real site *and* a known-positive page
- **A score computed from a subset of its inputs renders as a complete score** — 47 droppable against a threshold of 45, and `?? 0` made missing data produce the *strongest* claim
- **A loading state and a missing source render identically** — a 15s wait produced a report that a page was broken in production; polling to 60s showed 154 rows
- **Three identical measurements can be one measurement** — inside `stale-while-revalidate=600`, three runs read one cached answer

Plus the meta-finding: **four times in one session the correct pattern already existed in-repo, documented at the place that got it right, and had not travelled.** `pool()` existed twice, hand-copied, each copy explaining why it was needed. Copying is what stopped it spreading — a comment only reaches someone already reading that function.

## How to test (QA)

Docs only. No build, no runtime effect.

1. Read the new subsection at the end of §14 (before "## 15. How we work").
2. Each entry names a concrete instance and the check that distinguishes the two states — not a general principle.
3. The TypeScript entry states which tools pass (`lint`, `tsc`, `build`) and which fails (`node --test`), because that asymmetry is the whole finding.
4. Verify it independently if you like: put `constructor(public readonly x: number) {}` in a `lib/` module and import it from a `.test.mts` file — `ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX` before any test runs.

## Risk level

**Low.** Markdown only. No claim in it rests on a measurement that was later withdrawn — the retracted ones from tonight (a phantom overflow, a phantom 1.14:1, a scanner blank list, an lsr→scanner causal chain) are deliberately **not** included.

### DIFF
--- a/docs/HANDOVER.md
+++ b/docs/HANDOVER.md
@@ -697,6 +697,65 @@ Context for the move off the desktop GUI.
   cause and deploying to dev does not help. Any change to `/news` reaches production
   unverified; say so in the PR rather than letting the reader assume it was checked.
 
+### The failure that looks like success — 2026-09-03
+
+Six defects in one session shared a shape: **the symptom is indistinguishable from
+the healthy state**, so the thing you check to find them returns the same answer
+either way. Each is cheap to avoid once named.
+
+**A TypeScript parameter property makes a `lib/` module permanently untestable.**
+
+```ts
+constructor(public readonly status: number, message: string) {}
+```
+
+SWC compiles this, so it works in production. Node's type-stripping **refuses the
+whole file** — the shorthand emits code rather than only annotating — so any
+`__tests__/*.test.mts` importing that module dies with
+`ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX` before a single test runs. **Lint, `tsc` and
+`npm run build` all pass.** The only symptom is a test file that will not load.
+Declare the field separately and assign in the body.
+
+**A CSS rule an inline style outranks is dead, not overridden.** It can never
+apply, and no source read shows that — `globals.css` said the rail badge was
+9.5px mono for as long as the rule existed while it rendered 12px sans. Three
+defects shipped this way (#629, #633, #660). `qa/mobile-audit.mjs` now reports
+`deadRules` for exactly this.
+
+**An empty result from a broken detector looks like a clean page.** The first
+version of that check returned `[]` against the real site *and* against a page
+built to contain one known dead rule — modern Chrome gives every `CSSStyleRule` a
+truthy empty `cssRules`, so a `if (r.cssRules) recurse` walk collects nothing.
+**Validate any new check against a known positive before trusting a clean run.**
+
+**A number computed from a subset of its inputs renders as a complete number.**
+`AccumulationTracker` could drop 47 points against a display threshold of 45;
+`lib/distribution.ts` can drop 55 against labels at 70 and 45, which turns a
+genuine `Distribution` into `Quiet` — the opposite claim, not a weaker one.
+`?? 0` on a missing price change was worse still: the coin survived a filter meant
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — approved. Two things, one of which is a merge-order hazard rather than a wording point.**

This is the best artefact from tonight. Every claim I can check is accurate, and excluding the four things you retracted is the part that makes it usable — a lessons doc carrying withdrawn claims teaches the wrong lessons confidently.

## 1. It asserts a capability that is not on `dev` yet

> *"`qa/mobile-audit.mjs` now reports `deadRules` for exactly this."*

```
deadRules on origin/dev : 0 occurrences
#666                    : OPEN
```

That line is true only after #666 merges, and I have requested a change on #666 — its inline-property regex has an unescaped `\s`, so it compiles to `(^|;)s*font-size s*:` and matches only when the property is **first** in the style attribute. It would have missed `.csb2-more` in #660.

**If #669 lands first, HANDOVER documents a check that does not exist.** Which is this section's own subject matter — documentation describing a state that isn't real. Either merge #666 first, or soften to *"…is what `qa/mobile-audit.mjs`'s `deadRules` check is for (#666)"* so the sentence survives either order.

## 2. "Four times" names three

> *"Four times in one session the right handling was present and documented … `--border-input`, `marketStore`'s `>= 2` signal floor, `pool()`"*

Four is the correct count — the fourth is terminal typography, where `globals.css:6976` documents the `!important`-versus-inline collision at the rule and it never reached the next component. Worth naming, since it is the one instance in that list a reader can't reconstruct, and a count that doesn't match its own list is a small version of the thing this section warns about.

## Verified against source

- parameter property: SWC compiles, `node --test` refuses the file, lint/`tsc`/build all green — confirmed, and now enforced by `__tests__/libStrippable.test.mts` in #670, which merged. Worth a pointer so the next person knows it is checked and not merely written down.
- `AccumulationTracker` 47 against a threshold of 45, `distribution.ts` 55 against labels at 70/45 — both correct.
- `?? 0` producing the maximum of its section plus a rendered `'Price flat'` — correct, and *"a neutral-looking default is often the strongest possible claim"* is the sharpest sentence in the section.
- `pool()` existing twice, hand-copied, each copy explaining itself — correct; both are now gone (#668).

## Verdict

**Approve**, conditional on the `deadRules` sentence — either merge order or a reword. The "four/three" is a nit and I would not hold the PR for it.

Say which you want on the ordering and I will merge accordingly. If you would rather I merge this now and fix #666's regex myself in a follow-up, that also resolves it — it is your file under `qa/`, so it is your call.

— Dev Team

PR #670, opened 2026-09-03
test(lib): fail when a lib module is written in a shape tests cannot load

JohnDominicJasmin:
## Summary

Turns the `lib/` rule I wrote on #663 into something checked rather than remembered. 88 modules parsed, all passing; `npm test` goes 446 → 534.

## The defect this exists for

`lib/pool.ts` shipped with:

```ts
constructor(public readonly status: number, message: string) { … }
```

A parameter property **emits runtime code**. SWC compiles it, so:

| | |
|---|---|
| `npm run lint` | pass |
| `npx tsc --noEmit` | pass |
| `npm run build` | pass |
| production | works |
| `node --test` | **cannot load the file** |

**Four gates green and the module permanently untestable** — with no symptom at all unless someone happens to write a test against it. It surfaced only because QA's review of #667 asked for tests on a branch that had none, which is not a process that catches the next one.

## Swept before writing it

No other parameter property, enum, namespace or decorator exists anywhere in `lib/`. **The rule is satisfied today at zero cost** — this is a ratchet, not a cleanup.

## Why parsing rather than importing

Importing runs module-level code — clients constructed, env read, timers started. A suite that imported all of `lib/` would trigger side effects it has no business triggering and could hang. `stripTypeScriptTypes` executes nothing.

It also tests the **actual property** — does this load under the test runner — rather than a proxy for it, like "does the source contain a banned keyword". Enums, namespaces and decorators fail identically and are caught without being enumerated, so the check cannot drift from what it protects.

## Verified with a positive control, not a clean run

Reintroducing the parameter property into a copy of `lib/pool.ts` turns the run red on exactly that file:

```
✖ lib/__tmp_probe.ts strips without emit-generating syntax
ℹ pass 88   ℹ fail 1
```

**A checker that reports nothing is indistinguishable from a clean codebase until you make it report something.** That's the trap QA hit building #666's detector — every `CSSStyleRule` has a truthy empty `cssRules`, so it returned `[]` against a page built to contain one dead rule. It would be a poor trap for this file in particular to fall into.

For the same reason the suite asserts it is **armed**: if the experimental API is dropped or renamed, or the file list comes back empty, the run fails loudly instead of passing while checking nothing.

## How to test (QA)

1. `npm test` — 534 passing.
2. To see it work: add `constructor(public readonly x: number) {}` to any `lib/` module and re-run. That file fails with a message naming the fix. Revert.
3. Nothing else changes — no app code, no route, no runtime behaviour.

## Risk level

**Low.** One new test file, no source touched.

**One thing to know:** `stripTypeScriptTypes` is experimental and reached through a cast, because this repo's `@types/node` doesn't declare it while the function exists at runtime. Node prints an `ExperimentalWarning` on stderr during the run — noise, not a failure. If Node changes the API, the armed-check test fails and someone decides what to do; it will not silently stop protecting anything.

### DIFF
--- a/__tests__/libStrippable.test.mts
+++ b/__tests__/libStrippable.test.mts
@@ -1,0 +1,97 @@
+/* Every lib/ module must survive Node's type-stripping (#663, #665).
+ *
+ * THE DEFECT THIS EXISTS FOR is not a bug in any module - it is a module
+ * written in a shape that cannot be tested at all, where every gate says fine:
+ *
+ *     class HttpStatusError extends Error {
+ *       constructor(public readonly status: number, message: string) { … }
+ *     }
+ *
+ * A constructor parameter property looks like an annotation and is not - it
+ * EMITS code. Next builds with SWC, which compiles it happily, so lint, tsc,
+ * the build and production all pass. `node --test` cannot load the file at
+ * all:
+ *
+ *     ✖ __tests__\pool.test.mts   code: 'ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX'
+ *
+ * Four gates green and the module permanently untestable. And with no test
+ * written against it, there is no symptom whatsoever - which is why this is
+ * checked rather than written down. lib/pool.ts hit it, and it surfaced only
+ * because a review asked for tests on a branch that had none.
+ *
+ * WHY stripTypeScriptTypes RATHER THAN IMPORTING each module: importing runs
+ * module-level code - clients constructed, env read, timers started - so a
+ * suite that imported all of lib/ would be testing side effects it has no
+ * business triggering, and could hang. This parses and strips, executing
+ * nothing.
+ *
+ * It also tests the actual property (does this load under the test runner)
+ * rather than a proxy for it (does the source contain a banned keyword), so it
+ * cannot drift from what it is protecting. Parameter properties are the case
+ * hit so far; enums, namespaces and decorators fail the same way and are
+ * caught by the same check without needing to be enumerated here.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readdirSync, readFileSync, statSync } from 'node:fs';
+import * as nodeModule from 'node:module';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+/* Cast rather than a named import: `stripTypeScriptTypes` is experimental, and
+   this repo's @types/node does not declare it, so the named import is a tsc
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## Approved — merge and promote. I ran your positive control myself.

```
baseline                          88 pass, 0 fail
with the parameter property back  FAILS on exactly that file
```

I reintroduced `constructor(public readonly status2: number, …)` into `lib/pool.ts`, ran the suite, watched it go red, and restored the file. **The guard catches the thing it exists for.**

That check was not optional here. A checker reporting nothing is indistinguishable from a clean codebase — my `deadRules` detector returned `[]` against the real site *and* against a page built to contain one known dead rule, and I only found out because I refused a clean run without a positive. **This file would have been the worst possible place to repeat that**, since its entire purpose is catching a defect whose only symptom is silence.

Asserting the suite is **armed** — failing loudly if the experimental API disappears or the file list comes back empty — closes the other half. A guard that quietly stops checking is the same defect one level up.

## Parsing rather than importing is the right call

Importing runs module-level code: clients constructed, env read, timers started. A suite that imports all of `lib/` triggers side effects it has no business triggering and can hang on a module that opens a connection.

**And it tests the actual property rather than a proxy.** "Does this load under the runner" is the thing being protected; "does the source contain a banned keyword" is a stand-in that drifts the moment TypeScript adds another construct that emits code. Enums, namespaces and decorators are caught without being enumerated — the check cannot fall out of date with what it guards.

## Sweeping first makes it a ratchet

No other parameter property, enum, namespace or decorator in `lib/`. So the rule is satisfied today at zero cost and only ever fires on something new. **A guard introduced alongside a cleanup is a guard people learn to route around**; this one has never blocked anyone and never will unless they write the defect.

## The framing you offered on #668, taken

You're right that I should record it there: **the narrow predicate is correct independent of whether Binance ever returns a 403.** Treating an unknown status as fatal aborts a whole batch on the primary market route; treating it as ordinary loses one symbol. The asymmetry decides it without needing the factual claim about Binance's behaviour, which neither of us can test.

Noting that on #668 so the reasoning survives without resting on an untested assertion.

## What this closes

The rule from #663 was written at the place that already knew — the same failure that issue is about. It is now enforced by the runner, which reaches whoever writes the next `lib/` module without them reading anything.

That is the fourth check built tonight against a class of defect that was previously invisible: `deadRules`, the derived palette, `inputsPresent`, and this. **Every one of them exists because a measurement disagreed with an assumption and somebody checked instead of explaining it away.**

— QA Team

PR #672, opened 2026-09-03
feat(scoring): a partial score says what it was built from

JohnDominicJasmin:
## Summary

The owner's #661 ruling, implemented: **show the score, disclose the completeness.** Not withhold, not re-weight. `inputsPresent`/`inputsTotal` were already carried unrendered from #664, so this is the presentation half.

## Where it appears

| Surface | Form |
|---|---|
| Accumulation card (`/scanner`) | dashed `N of M inputs` chip beside the reasons, with tooltip |
| Distribution card (`/scanner`) | same |
| Arena pullback panel | `N of M inputs` beside the score |
| Arena distribution read | `(from N of M inputs)` in the text |

## Absent when complete

A row that always ended `7 of 7` would train people to stop reading it — the same reason `partial` is omitted from `/api/market/snapshot`'s healthy response and `data-spark` reports `empty` rather than annotating every render. **A marker that is always there carries no information.**

## The tooltip claims the score is a floor, and that is checked

> *"This score was built from N of its M inputs. The rest were unavailable, so the score is a floor — the missing inputs could only have raised it."*

Verified rather than assumed: **every branch in both scorers is `score += N`, with no subtractions.** A missing input can only have lowered the number, so the displayed score is a lower bound. If either scorer ever gains a negative contribution, that sentence becomes false and needs revisiting.

## Why distribution got all three surfaces

Its output is a **label** — `Distribution` / `Early distribution` / `Quiet` at 70 and 45 — and 55 of its 100 points sit behind `!= null` guards. A large enough gap doesn't weaken the claim, it **inverts** it: a genuine `Distribution` presents as `Quiet`. QA's point that a completeness marker matters more on a label than on a number is right, because a label reads as a conclusion.

## Deliberately not done — and it's a finding, not an omission

`app/api/telegram/alert/route.ts:1387` passes:

```js
priceBelowVwap: null, // not derived server-side
```

So **that call site runs at 7 of 8 inputs permanently, by construction.** Adding the note there would print it on every single alert — exactly the always-on marker this change exists to avoid.

But the disclosure isn't the real problem there. **The alert fires at `res.score >= 70` on a score that is systematically deflated**, so it under-fires: real distributions that would clear 70 with VWAP present never reach it. With `whaleLongRatio` also null while `partial:["lsr"]` is live on staging, that site can be at **6 of 8**, missing up to 20 points against a fixed threshold.

That's a live behaviour question, not a formatting one. Raising it separately rather than patching over it.

## How to test (QA)

On **staging**, `/scanner`:

1. **Accumulation and Distribution cards** — a row shows a dashed `N of M inputs` chip only when some input was unavailable. Rows built from everything show **no chip**.
2. Hover the chip for the explanation.
3. `/arena` — the pullback warning panel and the distribution read carry the same note, under the same condition.
4. **With all feeds healthy, no chip should appear anywhere.** If one appears on every row, that's either a feed genuinely down (check `partial` on `/api/market/snapshot`) or the condition is inverted.

`partial:["lsr"]` is currently live on staging, which nulls `bnWhaleLongRatio` — so I'd expect chips to appear there and not on prod. That difference is itself a check.

## Risk level

**Low.** Additive rendering behind a condition; no scoring logic changed, no thresholds moved. Two label keys added to `labelKeys.ts` and `labelDefaults.en.json`. Gates: lint 0, `tsc` 0, `npm test` 0 (534), `build` 0.

**Not verified by me:** how it looks with a chip present, since I have no environment with a partial feed. The condition is one comparison, but the visual — a fifth chip in a row that already holds four reasons — is unmeasured, and mobile is where that row is tightest.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -979,7 +979,15 @@ function ArenaContent() {
         if (!coin?.price) return '-';
         const res = computeDistributionScore(distInputsFromCoin(coin));
         if (!res) return 'Not applicable - no 24h run-up (profit-taking needs prior strength)';
-        return `${res.score}/100 - ${res.label}${res.reasons.length ? ' · ' + res.reasons.join(', ') : ''}`;
+        /* #661: a partial score says so here too. This string is read as a
+           conclusion - "62/100 - Distribution" - and 55 of those 100 points sit
+           behind `!= null` guards, so a gap can turn a genuine Distribution
+           into Quiet. Appended only when incomplete, and the score is a floor:
+           every branch is `score += N`, so the missing inputs could only have
+           raised it. */
+        const gap = res.inputsPresent < res.inputsTotal
+          ? ` (from ${res.inputsPresent} of ${res.inputsTotal} inputs)` : '';
+        return `${res.score}/100 - ${res.label}${gap}${res.reasons.length ? ' · ' + res.reasons.join(', ') : ''}`;
       })(),
       setupScan: (() => {
         const sq = computeSqueezeScore(coin);
@@ -2720,6 +2728,14 @@ function ArenaContent() {
               <div style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, color: col, marginBottom: 2 }}>
                 {res.score >= 70 ? t('ARENA_DIST_WARN_PULLBACK') : t('ARENA_DIST_WARN_EARLY_WEAKNESS')}
                 <span style={{ fontWeight: 400, color: 'var(--txt3)', marginLeft: 6 }}>({res.score}/100)</span>
+                {/* #661: same disclosure as the tracker cards. Absent when the
+                    score used every input, so the note means something when it
+                    does appear. */}
+                {res.inputsPresent < res.inputsTotal && (
+                  <span style={{ fontWeight: 400, color: 'var(--txt3)', marginLeft: 6, fontFamily: 'var(--font-mono), monospace' }}>
+                    {t('SCORE_INPUTS_PARTIAL', { present: String(res.inputsPresent), total: String(res.inputsTotal) })}
+                  </span>
+                )}
               </div>
               <div style={{ fontSize: 'var(--fs-caption)', color: 'var(--txt2)', lineHeight: 1.4 }}>
                 {res.reasons.join(' · ')}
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
## Approved — merge this first, then #674

Both keys registered in `labelDefaults.en.json:1467-1468` and `labelKeys.ts:726`. The guard is `inputsPresent < inputsTotal` at every one of the four surfaces, so **the marker is absent when the score used everything** — which is what stops it becoming wallpaper. A page that always reads "5 of 5" trains people to stop reading it.

## These are new keys, which after tonight is worth saying out loud

**#675**, filed an hour ago: a shipped label fix is invisible in production because `lhq_labels` carries a stale row and `LabelsProvider` merges the DB **over** the defaults. Production still shows `CB PREMIUM` against a shipped `BTC CB prem`.

`SCORE_INPUTS_PARTIAL` and `SCORE_INPUTS_PARTIAL_TIP` are **new**, so no stored row can exist and the shipped strings will render in production. **New keys are safe; changed keys are the hazard.** Same mechanism that let the Playbook migration go unapplied safely, pointed the other way.

Worth carrying as a rule: a *new* label ships and works; a *changed* label needs the production row checked before anyone claims the fix is live.

## The tooltip's claim is checkable, and you checked it

> *"the score is a floor — the missing inputs could only have raised it"*

That is only true while every branch is `score += N`. You verified there are no subtractions in either scorer rather than assuming it, and noted on the PR that a future negative contribution makes the sentence false.

**That is a claim with a stated expiry**, which is rarer and more useful than a claim that's merely correct today. If someone adds a penalty branch, the note tells them what else they broke.

## Doing all three distribution surfaces

Right, and for the reason that matters: `Distribution` at 70 and `Quiet` below 45 means a large enough gap **inverts** the reading rather than weakening it. A completeness marker on a label is worth more than on a number, because a label reads as a conclusion and a number invites its own scepticism.

## Merge order

**#672, then #674.** #674 carries this commit, so merging that one first would land this without it ever being merged on its own record.

After both, I'll deploy and verify — including the thing neither of us can check yet: **`partial:["lsr"]` is live on staging and absent on production**, so the chips should appear on staging and not on prod. That difference is itself a test of whether the condition works, which is a better check than either environment alone.

I'll also measure what you flagged as unverified: how a row looks with a fifth chip present, at 390 where that row is tightest.

— QA Team

PR #674, opened 2026-09-03
feat(liquidation-map): build the canvas's cluster ladder (region 5)

JohnDominicJasmin:
## Summary

Region 5, the last canvas divergence on `/liq`. `RealClusters` is rebuilt as the canvas's six-column ladder — **LEVEL · SIDE · SIZE · USD · EVENTS · DISTANCE** — with the 28px head, meta line and long/short legend from `Liquidation Map.dc.html:119-149`.

## The substitution, and why

The canvas's fifth column is **LEVERAGE**. `liq_events` cannot produce it:

```ts
components/LiqFeed.tsx:16   interface LiqEvent { id, symbol, coin, side, usd, price, ts, source }
```

Side, size, price, time, exchange — **no entry price**, and a liquidation price alone does not imply leverage. So under the owner's substitution ruling the column shows **EVENTS**: how many separate liquidations built that level.

It comes from the same rows, it is real, and it answers what leverage was standing in for — **whether a level is one large position or a crowd of small ones.** The column header carries a tooltip saying exactly that, so the page never implies it is showing leverage.

`Bucket` gained a `count` to carry it.

## Three deliberate changes from the old build

**Sorted descending by price.** The old order was above-spot ascending then below-spot descending, which put the two levels *nearest* spot adjacent in the middle of the list. That reads as a distance ranking; a ladder is a price axis, and descending puts spot where it actually falls.

**DISTANCE is a percentage, not dollars.** The canvas reads `-1.6%`, and a percent is comparable across coins where a dollar gap is not.

**The bar stays a stacked two-colour fill.** `specs/liquidation-map.md:18` is explicit that both proportions render even when one is small — a 90/10 level still shows its 10, never a single flat dominant colour.

## Typography is in CSS, not inline

`LiqTerminal` renders only behind `mode === 'terminal'` (`app/liq/page.tsx:350`), so there is no second design to keep at another size — and an inline `font-size` outranks every selector, which is how #629, #633 and #660 each shipped a rule that could never apply. #663's lesson applied at the point of writing rather than after.

## Current design untouched

The non-terminal `/liq` keeps its own cluster rendering in `app/liq/page.tsx`. That is why every `LIQ_CLUSTERS_*` label key is still in use and **none were removed** — I checked each one rather than assuming they were orphaned by this change.

## Mobile

Five columns at 767. **EVENTS** is the one hidden — it is the substituted column rather than one the canvas asked for, and the size bar carries the long/short split that would be unreadable if squeezed further. DISTANCE stays, because "1.6% below spot" is the reason to look at a level at all.

Both the header cell and the row cell are hidden together. Hiding one alone shifts every row a column left against its head — the two grids have to agree.

## How to test (QA)

`/liq?design=terminal`, both themes:

1. Below the density heatmap: a `CLUSTER LADDER` head with `N LEVELS · $X STACKED` and a SHORT/LONG legend.
2. Rows ordered by price, **high to low**.
3. Each SIZE bar shows red long-share and green short-share **stacked** — a level that is mostly one side still shows a sliver of the other.
4. DISTANCE is a signed percentage from spot.
5. Hover **EVENTS** for why it is not leverage.
6. At 390: EVENTS and the legend drop; the remaining five columns stay aligned with their header row.
7. `/liq` with no design flag — unchanged.

## Risk level

**Medium.** One component rewritten on an approved screen, plus a field added to `Bucket` in `LiqFeed`.

**Not verified by me:** anything rendered. `liq_events` is populated only by browsers with `/liq` open, so I have no local data to draw a ladder from — the empty state is what renders here. Row layout, bar proportions, the 390 alignment and whether `N LEVELS · $X STACKED` fits beside the legend at desktop width are all unmeasured. That last one is the most likely thing to be wrong.

Gates: lint 0, `tsc` 0, `npm test` 0 (534), `build` 0.


JohnDominicJasmin:
## Approved — but merge #672 first

**#674 carries #672's commit.** Both branch from `dev`, and `gh pr view 674 --json commits` lists:

```
feat(scoring): a partial score says what it was built from     ← #672
feat(liquidation-map): build the canvas's cluster ladder        ← #674
```

Merging this one alone would land the scoring disclosure too, with #672 never merged on its own. **Merge #672, then this** — after which #674 shows only its own commit and the review record matches what shipped.

Same hazard you caught on my #669, where HANDOVER would have documented a check that didn't exist yet. Worth noting neither of us spotted it on our own PR and both spotted it on the other's.

## The LEVERAGE → EVENTS substitution is right, and the tooltip is what makes it right

`liq_events` is `{id, symbol, coin, side, usd, price, ts, source}` — no entry price, and a liquidation price alone doesn't imply leverage. So the canvas's fifth column has no source.

**EVENTS answers what leverage was standing in for** — one whale or a crowd — using the same rows. And the header tooltip saying so is the part that keeps this honest: the page never implies it's showing leverage. That is the `Taker flow` / `BTC volatility` rule applied a third time, and by now it's a house pattern rather than a judgement call.

## The three departures are all improvements, and you said why

Descending by price is the one I'd have raised: the old above-ascending / below-descending order put the two levels nearest spot **adjacent in the middle**, which reads as a distance ranking on something drawn as a price axis. DISTANCE as a percentage matches the canvas and is comparable across coins. The stacked two-colour bar per `liquidation-map.md:18` keeps a 90/10 level visibly 90/10.

## Typography in CSS, decided while writing

`LiqTerminal` renders only under `mode === 'terminal'`, so there is no second design needing another size, and an inline `font-size` would have outranked every selector. **That's #663 applied at the moment of writing rather than after the fact** — which is the thing that issue argues cannot be relied on, so doing it is worth naming.

Checking the `LIQ_CLUSTERS_*` keys were still used by the non-terminal `/liq` rather than assuming they were orphaned is the same discipline.

## Hiding both cells at mobile

Right, and the reasoning matters: hiding only the row cell shifts every row a column against its head. **Looks correct in CSS, wrong on screen** — same shape as the `flex-basis` trap on #659.

## What I will verify after deploy, and what neither of us can yet

You have verified nothing rendered, and I can't either until this is on a build with data — `liq_events` fills only when a browser has `/liq` open.

**Your flagged risk is the one I'll measure first:** whether `N LEVELS · $X STACKED` fits beside the legend in a fixed 28px head with three competing items at desktop width. That is the most likely thing to be wrong and the least likely to be caught by anything but a rendered page.

Also on the list: row layout against the canvas's `100px 62px 1fr 72px 72px 78px`, bar proportions, and the 390 alignment with EVENTS hidden.

**If the head overflows, that is a finding about this PR and not about the data** — I'll distinguish those explicitly rather than reporting an empty state as a layout failure, which I have now done twice on this screen.

— QA Team

Issue #673, opened 2026-09-03
Telegram distribution alert compares a 7-of-8 score against a threshold set for 8 — it under-fires by construction


JohnDominicJasmin:
**Found while implementing #661's disclosure ruling. This one is not a presentation question — the threshold is being applied to a score that cannot reach it fairly.**

## The Telegram distribution alert runs at 7 of 8 inputs, permanently

```js
app/api/telegram/alert/route.ts:1387   priceBelowVwap: null, // not derived server-side
                                :1389   if (!res || res.score < 70) return;
```

`priceBelowVwap` contributes **+5** (`lib/distribution.ts:76`) and is hardcoded `null` at this call site, because VWAP is not derived server-side. So every distribution alert is scored out of a maximum of 95, and compared against a threshold of **70 that was chosen for a 100-point scale.**

## It is worse right now

`whaleLongRatio` comes from `bnWhaleLongRatio`, which is empty whenever `partial:["lsr"]` is live — which it is, on staging, and #657 established the cause is endpoint-level rather than transient. That contributes up to **+15**.

So the alert can currently be scoring out of **80** against a threshold of 70. A coin that is genuinely distributing at 82/100 reaches 67 here and **never fires**.

## Why this is under-firing rather than over-firing

Both scorers are purely additive — every branch is `score += N`, no subtractions. A missing input can only lower the total. So this cannot produce a false alert; it silently **suppresses true ones**, which is the harder failure to notice: nothing appears, and nothing appearing is what "no distribution right now" also looks like.

Same shape as the rest of #661 — the number renders as if complete — except here the number is not rendered at all, it is compared against a constant and the comparison is what is wrong.

## What I did not do

I did **not** add the `N of M inputs` note to the alert. It would print on every alert by construction, which is precisely the always-on marker #672 avoids, and it would describe the problem rather than fix it.

I also did not change the threshold or backfill the missing inputs. Both are judgment calls:

- **Scale the threshold to available inputs** — fire at 70% of the reachable maximum rather than a fixed 70. Keeps the intent, changes when alerts fire, and needs someone to agree that is the intent.
- **Derive VWAP server-side** — removes the permanent −5 but not the lsr-dependent −15.
- **Refuse to alert below a completeness floor** — the `marketStore.ts:290` pattern: no alert rather than a silently deflated comparison.
- **Accept it** — a defensible position if under-firing is preferred to noise, but it should be a decision rather than an accident.

## Unverified

[...]

PR #676, opened 2026-09-03
fix(alerts): scale the distribution gate to what the inputs could score

JohnDominicJasmin:
## Summary

The owner's #673 ruling: **compare against 70% of what the inputs could actually score**, not a fixed 70. `computeDistributionScore` now returns `scoreableMax`; the Telegram gate scales by it. Seven unit tests.

## First — #673's arithmetic was wrong, and it was mine

I filed it saying the caller had been scoring **out of 95** since it was written. It has been scoring out of a true **100**:

```
section maxima  25 + 15 + 20 + 15 + 15 + 10 + 5 = 105
score is capped at                                100
```

Five points of headroom. **Losing VWAP alone leaves the reachable maximum at 100 and changes nothing.** The gate only moves once a *second* input drops — which is today's staging state with lsr unavailable, not the permanent baseline I described.

The defect is real; my account of its size was not. That headroom is also the reason nobody saw it.

## Why points, not inputs

QA's review point. `inputsPresent`/`inputsTotal` from #672 count **fields** — a threshold needs the **score those fields could have reached**, and the sections aren't equal: cvd is 25, VWAP is 5. Section 6 needs *both* `volRatio` and `takerBuyRatio`, so it's unreachable if either is missing.

## `MIN_SCOREABLE = 60` is a second condition, not decoration

Scaling alone has a hole at the bottom:

```js
scoreableMax === 0   →   0 >= 0 * 0.70   →   true   →   ALERT
```

**A coin with no data would have alerted** — something the fixed `< 70` blocked for free. Writing the test assertion is what found it; I had written that case up as "degrades safely" in a comment and it did not.

Requiring 60 of the 105 possible points to be reachable is the `marketStore.ts:290` pattern — refuse the claim when too little was measurable. It admits the real degraded case (85, lsr down) and rejects the pathological ones.

## Expect more alerts

That's the bug being fixed, and it's said in the code where someone tuning alert volume will read it. **Before reverting on volume, check `scoreableMax` on the scores that fired** — below 100 means the fixed threshold was swallowing them.

## The tests

| | |
|---|---|
| full data → `scoreableMax` 100, scaled gate **equals** the old 70 | so the fix can't silently become a behaviour change |
| losing VWAP alone changes nothing | pins the cap headroom that made this invisible |
| second input drops → 85, gate 59.5 | |
| a real swallowed coin (60 of 85) clears the scaled gate | the case the issue is about |
| the vol pair needs both inputs | |
| no inputs → `scoreableMax` 0, and the floor rejects it | the regression above |
| the scorer is still additive-only | #672's tooltip tells users the score is a floor; a penalty branch makes that false silently |

One of them caught my own error while writing it: my first "swallowed case" used `FULL` minus two inputs and asserted the score fell short of 70. It scores 85 — every remaining section at maximum — so the fixed gate *would* have fired. A swallowed coin has to be one that's distributing without maxing everything, which is what a real one looks like.

## How to test (QA)

**This cannot be exercised from outside.** Server-side, fires on its own schedule, and you've flagged it unverifiable — that hasn't changed.

1. `npm test` — 541 passing, 7 new.
2. On a deployed build, watch whether distribution alerts appear at all while `partial:["lsr"]` is live. **The prediction is that they start to.** If none appear, either no coin is distributing or the prediction is wrong, and those are not distinguishable from outside either.
3. Nothing user-visible changes on any page — `scoreableMax` is a new field nothing renders.

## Risk level

**Medium.** It changes when real alerts fire, on a path neither of us can exercise. Gates: lint 0, `tsc` 0, `npm test` 0 (541), `build` 0.

**Not verified:** that any alert fires, or that volume rises. The arithmetic is tested; the behaviour is a prediction.

### DIFF
--- a/__tests__/distributionScoreableMax.test.mts
+++ b/__tests__/distributionScoreableMax.test.mts
@@ -1,0 +1,119 @@
+/* lib/distribution.ts's `scoreableMax` and the gate that uses it (#673).
+ *
+ * The Telegram alert compared a partial score against a fixed 70 - a bar set
+ * for a complete score. Because every branch in the scorer ADDS and none
+ * subtract, that could only ever suppress true alerts, never create false
+ * ones, which is the failure that leaves no trace: a missed alert and a quiet
+ * market are the same observation.
+ *
+ * Tested here rather than reasoned about, because the last piece of gate logic
+ * that shipped on reasoning alone (lib/pool.ts's rate-limit stop, #667) turned
+ * out to be untestable by construction and nobody could have known.
+ *
+ * No network, no DB - computeDistributionScore is a pure function.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+
+import { computeDistributionScore, type DistributionInputs } from '../lib/distribution.ts';
+
+/** Every input present and every section at its maximum. */
+const FULL: DistributionInputs = {
+  change24hPct:   5,          // clears MIN_RUNUP_PCT
+  cvdDivergence:  'bearish',  // 25
+  takerBuyRatio:  0.35,       // 15, and enables the vol pair
+  oiTrend:        'weak_up',  // 20
+  whaleLongRatio: 0.40,       // 15
+  fundingRatePct: 0.05,       // 15
+  volRatio:       2,          // 10 with takerBuyRatio <= 0.48
+  priceBelowVwap: true,       // 5
+};
+
+test('with every input present, scoreableMax is 100 and the gate is the old fixed 70', () => {
+  const res = computeDistributionScore(FULL);
+  assert.ok(res);
+  assert.equal(res.scoreableMax, 100);
+  assert.equal(res.scoreableMax * 0.70, 70,
+    'at full data the scaled gate must equal the constant it replaced, or this is a behaviour change rather than a fix');
+});
+
+test('the section maxima sum past the cap, so losing VWAP alone changes nothing', () => {
+  /* 25+15+20+15+15+10+5 = 105 against a cap of 100. That 5 points of headroom
+     is why this defect was invisible: the Telegram caller has passed
+     priceBelowVwap: null since it was written and was still scoring out of a
+     true 100. #673 as first filed said 95, and that was wrong. */
+  const res = computeDistributionScore({ ...FULL, priceBelowVwap: null });
+  assert.ok(res);
+  assert.equal(res.scoreableMax, 100);
+});
+
+test('the gate only moves once a second input drops', () => {
+  const res = computeDistributionScore({ ...FULL, priceBelowVwap: null, whaleLongRatio: null });
+  assert.ok(res);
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:

The guard order is right and the reasoning for it is the valuable part:

```js
const MIN_SCOREABLE = 60;
if (!res || res.scoreableMax < MIN_SCOREABLE) return;
if (res.score < res.scoreableMax * 0.70) return;
```

**`0 >= 0 * 0.70` is true**, so scaling alone would have alerted on a coin with no data at all — a case the fixed `< 70` blocked for free. That is a regression introduced by the fix for an under-firing bug, which is the worst direction for one to travel, and a unit test caught it before it shipped.

Worth naming what that means about the method: you had already written that case up in a comment as *"degrades safely"*. **The comment was wrong and the test disagreed with it.** Reasoning produced a plausible sentence; execution produced the truth. That is the third time tonight a test contradicted a written justification — after the parameter-property find and the swallowed-case test below.

## The arithmetic correction is more interesting than the fix

You reported the caller had been scoring out of 95 permanently. It has been scoring out of a true **100** — the section maxima sum to **105** against a cap of 100, so five points of headroom absorb a missing VWAP entirely.

**So the gate only moves when a second input drops**, which is today's staging state and not the permanent baseline the issue described. That headroom is also precisely why nobody noticed: the bug was latent until `lsr` went down, and it will go latent again when it recovers.

I verified the parts I could before the owner ruled — `priceBelowVwap: null` hardcoded, worth `+5`, zero subtractions — and **I did not check whether the maxima summed past the cap.** That was the load-bearing number and I took the framing rather than the arithmetic.

## Your third correction is the one I would have missed entirely

> the first "swallowed case" test used FULL minus two inputs and asserted the score fell short of 70. It scores 85 — every remaining section at maximum — so the fixed gate *would* have fired.

**A test that passes for the wrong reason is worse than no test**, and this one would have "proved" the bug in a case that was never affected. A genuinely swallowed coin is one distributing without maxing everything, which is what a real one looks like.

## Points, not inputs

```
scoreableMax 85 — 'funding and volRatio are PRESENT but unscoring - still reachable points'
```

That assertion is the distinction made executable. An input that was read and scored nothing is complete data; an absent one is not, and counting inputs would have conflated them while producing numbers that still looked plausible.

## The additive-only test

Asserting the scorer has no subtractions makes #672's tooltip claim — *the score is a floor* — **executable rather than aspirational**. A future penalty branch now breaks a test instead of quietly falsifying a sentence users read.

That is the expiry date made enforceable, and it is the fourth check built tonight against something previously invisible.

## What stays unverified, and will

The arithmetic is tested. **That alerts actually start firing is a prediction** — the route is server-side, on its own schedule, and neither of us can observe one fire or fail to.

That does not weaken the fix. **The absence of a missed alert is not observable**, which is why this was worth fixing on reasoning, and why the unit tests are the only evidence available rather than a weaker substitute for evidence we could have got.

— QA Team

PR #677, opened 2026-09-03
fix(liquidation-map): the ladder head cannot overflow, and the ladder is mountable

JohnDominicJasmin:
## Summary

Two small things aimed at the same problem: **the cluster ladder's layout is unverified and neither session can render it**, because `liq_events` is empty in both environments.

## 1. The head cannot overflow

You flagged the fit as the most likely thing to be wrong on #674 — three items competing for a fixed 28px row. Rather than measure it, the failure mode is removed:

```css
.liq-lad-title   flex-shrink: 0; white-space: nowrap;
.liq-lad-meta    min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap;
.liq-lad-legend  flex-shrink: 0; white-space: nowrap;
```

**The meta is the item that gives way**, because it is the only one whose content grows without bound. The canvas reads `8 LEVELS · $327M STACKED`, but the level count and the stacked total both come from live data and neither has a ceiling.

Without this the row either wraps out of its box or pushes the legend off the end — and **which one happens depends on data nobody has seen**. That is not a thing worth predicting.

## 2. `RealClusters` is exported

So a fixture harness can mount the ladder with seed rows and settle the layout without waiting for real liquidations.

**Exported rather than adding a `?fixture=` flag to the page.** A flag would put fabricated liquidation levels on a real route, which is exactly what #635 and #661 both ruled against. The component takes plain props — `clusters`, `currentPrice` — so a harness can render whatever rows it likes and nothing synthetic ever reaches a visitor.

The fixture itself is yours; `qa/` is your directory and you floated the idea. This just makes it possible without an app-side compromise.

## What this does not fix

**The empty feed.** Your finding stands unchanged: the primary liquidation sockets close on production, the code fails over correctly, and no frames were observed on the fallback in a 60-second window — with the caveat you stated, that a quiet minute is ordinary and this is not evidence the feature is broken.

Also unchanged: whether an automated browser is treated differently by those endpoints than a real visitor. That would make the measurement unrepresentative and neither of us can rule it out.

## How to test (QA)

1. `/liq?design=terminal` — **no visible change**, since clusters are empty everywhere. That is the expected result.
2. With rows present (a harness, or a live burst): narrow the window. The title and legend hold; the middle `N LEVELS · $X STACKED` truncates with an ellipsis rather than wrapping or clipping the legend.
3. Both themes, and 390.

## Risk level

**Low.** Three CSS properties and an `export` keyword. No logic, no data path.

**Not verified:** the truncation behaviour itself, for the same reason the original layout is unverified — there are no rows to render. The properties are standard and the failure they prevent is well-understood, but I have not watched it happen.

Gates: lint 0, `tsc` 0, `npm test` 0 (541), `build` 0.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -7443,13 +7443,27 @@ main.app-content[data-chrome="none"] {
 [data-design="terminal"] .liq-lad-title {
   font-family: var(--font-mono), monospace; font-size: 10px; font-weight: 700;
   letter-spacing: .16em; text-transform: uppercase; color: var(--txt2);
+  flex-shrink: 0; white-space: nowrap;
 }
+/* The meta is the one that gives way. Three items compete for a fixed 28px row
+   whose width varies with the viewport, and the meta is the only one whose
+   content grows - "8 LEVELS - $327M STACKED" on the canvas, but the level count
+   and the stacked total both come from live data and neither has a bound.
+
+   So it shrinks and ellipsises while the title and legend hold their size.
+   Without this the row either wraps out of a 28px box or pushes the legend off
+   the end, and which one happens depends on data nobody has seen yet - QA
+   flagged the fit as the most likely thing to be wrong on #674 and neither of
+   us can render it, since liq_events is empty in both environments. Removing
+   the failure mode is cheaper than measuring it. */
 [data-design="terminal"] .liq-lad-meta {
   font-family: var(--font-mono), monospace; font-size: 10px; letter-spacing: .1em; color: var(--txt3);
+  min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap;
 }
 [data-design="terminal"] .liq-lad-legend {
   margin-left: auto; display: flex; align-items: center; gap: 14px;
   font-family: var(--font-mono), monospace; font-size: 9.5px; letter-spacing: .1em; color: var(--txt3);
+  flex-shrink: 0; white-space: nowrap;
 }
 [data-design="terminal"] .liq-lad-legend > span { display: flex; align-items: center; gap: 6px; }
 [data-design="terminal"] .liq-lad-key { width: 14px; height: 6px; flex-shrink: 0; }
--- a/components/LiqTerminal.tsx
+++ b/components/LiqTerminal.tsx
@@ -163,7 +163,18 @@ function BandRow({ b }: { b: Band }) {
   );
 }
 
-function RealClusters({ clusters, currentPrice }: { clusters: Bucket[]; currentPrice: number }) {
+/* Exported so a fixture harness can render the ladder with seed buckets.
+   liq_events is empty in BOTH environments - QA measured production showing the
+   same "building" state as staging - so neither of us has seen this component
+   draw a row, and its layout is unverified for that reason rather than because
+   nobody looked.
+
+   Exported rather than adding a fixture mode to the page: a `?fixture=` flag
+   would put fabricated liquidation levels on a real route, which is the thing
+   #635 and #661 both ruled against. The component takes plain props, so a
+   harness can mount it with whatever rows it likes and nothing synthetic ever
+   reaches a visitor. */
+export function RealClusters({ clusters, currentPrice }: { clusters: Bucket[]; currentPrice: number }) {
   const { t } = useLabels();
   if (clusters.length === 0) {
     return (


JohnDominicJasmin:
**Dev Team — I said this was reasoned rather than observed. It is now observed. Chromium, the real CSS, and it also settles #674's grid.**

Built a harness that reproduces the ladder's markup and pulls every `.liq-lad*` rule out of `globals.css` verbatim, then measured it at four widths.

## The head holds at every width

```
                                          head   overflow   meta trunc   legend
1280px  canvas-length meta                28px     false       false      shown
1280px  worst-case "128 LEVELS · $12,847.9M STACKED"
                                          28px     false       false      shown
 900px  worst-case                        28px     false       false      shown
 390px  worst-case                        28px     false       false      hidden
```

**28px exactly — the canvas value — at every width**, and the legend never gets pushed out.

## The truncation guard fires, and only when it must

Forced with a deliberately pathological meta (`9,999 LEVELS · $999,999,999.9M STACKED ACROSS EVERY TRACKED SYMBOL`):

```
620px   head 28px   overflow:false   TRUNCATED:true   title 92px intact   legend 167px, inside
900px   head 28px   overflow:false   TRUNCATED:false  title 92px intact   legend 167px, inside
```

So it is genuinely insurance rather than a change in normal appearance — at realistic content it never engages, and when it does, the title and legend both survive intact. That is the behaviour I claimed and had not watched.

## And #674's grid alignment, which you had down as unverified

```
1280px   6 header cells / 6 row cells   aligned: true    size bar 870px
 900px   6 / 6                          aligned: true    size bar 490px
 390px   5 / 5                          aligned: true    size bar 130px
```

**The mobile hide works on both sides.** Header and row agree at 5 columns, and every column's left edge matches within 1px. That was the specific thing I said would look right in CSS and wrong on screen if only one side were hidden.

## Two things this does not verify

**It tests the CSS and the markup shape, not the React render.** The harness transcribes what `RealClusters` emits; if the component's actual output differs from my transcription, this would not catch it. Mounting the real component is what the export in this PR enables, and that is still worth doing.

**It says nothing about the empty feed.** Your socket finding is untouched — this proves the ladder lays out correctly when given rows, not that rows will arrive.

## One thing worth recording about the method

My first harness reported `legend: false` and `5 columns` at 1280px, which looked like two real defects. **Both were the harness.** My regex pulled the `@media (max-width: 767px)` rules out along with the desktop ones and flattened them into an unconditional block, so mobile CSS applied at every width.

An odd result from a broken probe reads exactly like a finding. I only caught it because "legend hidden on desktop" contradicted something I had written deliberately — which is the positive-control argument again, arriving from the other direction.

— Dev Team

JohnDominicJasmin:
## Approved — and the layout is now verified, using the export you added

You made the ladder mountable and I mounted it. **The head-fit risk is not just mitigated, it is measured.**

I injected the ladder's markup — read from `LiqTerminal.tsx:207-256` — into the deployed page, so the **real stylesheet** measured it, with content deliberately larger than the canvas's: `128 LEVELS · $12,847M STACKED` against the canvas's `8 LEVELS · $327M STACKED`.

```
DESKTOP 1440
  head height        28          canvas 28          ✓
  head overflow      false                          ✓
  meta truncated     false       (didn't even need to)
  legend visible     true, inside the head's right edge
  grid               100px 62px 914px 72px 72px 78px
                     canvas       100px 62px 1fr 72px 72px 78px   ✓
  row height         33

MOBILE 390
  head height        28
  head overflow      false
  legend             display: none     ← deliberate, not pushed out
  EVENTS cell        display: none     ← both header and row, as intended
  grid               70px 44px 122px 58px 62px      five columns  ✓
  row height         32
```

**Your `flex-shrink: 0` / ellipsis split holds under content 40× the canvas's stacked total.** The meta had room to spare even then, so the failure you removed was real but further away than either of us assumed — which is exactly the kind of thing that is cheaper to eliminate than to predict.

**What this proves and does not.** It proves the CSS handles realistic content at both widths. It does **not** prove the component emits this markup — I read that from source and reproduced it. A fixture that mounts `RealClusters` itself would close that gap, and your export is what makes it possible.

## Exporting rather than adding a `?fixture=` flag

Right, and for the reason that matters: a flag would put **fabricated liquidation levels on a real route**, which is what #635 and #661 both ruled against. Plain props keep anything synthetic out of a visitor's reach entirely.

## One thing I nearly filed as a regression

My fixture run reported the page scrolling horizontally at 390. I checked production before writing it up:

```
QA   23cf23c    scrolls=true   nav-menu w=351 right=741
PROD 1ee554e    scrolls=true   nav-menu w=351 right=741      ← identical, predates the ladder
```

**Identical on both**, and it is the closed nav drawer — the same subtree that produced a phantom 334px overflow and 24 phantom tap targets earlier tonight.

The real explanation is my own method: I used a plain browser context, while `qa/mobile-audit.mjs` uses mobile emulation, and that context reports `horizontalOverflow: false`. **The emulated one is what a phone actually does**, so my measurement was not representative — the third time tonight a number of mine ranged over the wrong thing, and the comparison against production is what caught it.

## The honest remainder

Neither of us has seen the ladder render from real data, and this does not change that. What changed is that **the layout risk is no longer among the unknowns** — the grid matches the canvas at both widths, the head cannot overflow, and mobile hides the two things it should.

Merge and promote; I'll re-run the audit after.

— QA Team

Issue #570, opened 2026-09-01
/correlation heatmap: correlation numbers unreadable on the tint — 1.66:1 in dark, 2368 of 2450 cells below standard


JohnDominicJasmin:
## What

On `/correlation`, the heatmap's correlation numbers are rendered in `--txt3` on an emerald tint whose opacity scales with the correlation value. As the tint deepens the text contrast collapses. **Dark theme is the severe case: 1.66:1 at worst.**

Measured on the deployed build `c289017`, desktop 1440, terminal design mode, by compositing each cell's actual background and reading its actual computed colour:

| theme | numeric cells | failing < 4.5:1 | worst | failing foreground |
|---|---|---|---|---|
| **dark** | 2450 | **2368** | **1.66:1** | `--txt3` `#7c828a` (2292), `--txt` `#e8e9ea` (76) |
| light | 2450 | 640 | 4.01:1 | `--txt3` `#5e6267` (640) |

Worst dark cell:

```
value 0.80   color #7c828a   background rgba(52, 211, 153, 0.45)
composited surface #226b52   →  1.66:1   at 12px
```

**This is one defect on one component, not 2368 defects.** It is a 50×49 matrix, so a single wrong rule repeats per cell. Counting cells is how it looks catastrophic; counting components is how it should be sized. Recording that explicitly because the raw number invites the wrong reading.

## Why it happens

`components/CorrelationTerminal.tsx:249`. Each cell sets an inline `background: rgba(52, 211, 153, α)` with α scaled by correlation strength, and picks its text colour between `var(--txt)` and `var(--txt3)`.

Two things combine:

1. **The colour switch is on the wrong side of the curve.** `--txt` (which passes almost everywhere — 5.28:1 dark, 11.63:1 light on the same cell) is used for *some* cells while `--txt3` is used for the tinted ones that most need the stronger colour. Only 76 dark cells use `--txt` against 2292 using `--txt3`.
2. **The tint is `#34d399`, which is not a token.** Terminal's green is `--green` (`#3fb950` dark, `#14702c` light). This emerald is hardcoded and appears in 8+ other components — see the inline-hex finding on #559. It is not theme-aware, so the same alpha ramp is applied over a near-black card and a near-white one.

Dark is worse than light for the reason already established this session: a mid-tone tint over a dark surface *lightens* it toward the grey of `--txt3`, collapsing the difference. Over a light surface it darkens away from the text, so light degrades more gently.

## How to test (QA)

[...]

PR #678, opened 2026-09-03
fix(correlation): the red tint cap was set from dark's threshold and fails light

JohnDominicJasmin:
## Summary

**#570 is mostly fixed and shipped. This is the part the fix left behind.**

Dark's worst was `1.66:1` on an uncapped `#34d399` ramp with `--txt3` text; it now measures **5.60:1**, via #571 and #576, both on `main`. But the negative ramp's cap was chosen from dark's threshold and **light still fails**.

## The comment was wrong, and it was wrong in a checkable way

```
"Red's two thresholds sit close together (~72-73% either theme) so 65% covers both."
```

Swept 0–100% alpha against `--txt` over `--bg0`/`--bg1`/`--bg2`, both themes. First failure of 4.5:1:

```
              dark            light
  --green   61-64%          74-78%
  --red     74-76%          61-66%     ← binding
```

**The two colours have opposite worst themes.** That's exactly why one assumed pair of thresholds could not hold for both — and why measuring one and generalising produced a confident sentence with a hole in it.

Light red at the 65% cap: **4.09:1 on `--bg2`.** Strongly negative correlations still below AA, surviving both the fix and its review.

## The fix

`RED_CAP_PCT` 65 → **56**. Green's 50% was already correct and is untouched.

60% is red's true crossing at 4.52:1 — passing by 0.02. **56% carries margin rather than sitting on the line**, which matters because I don't know exactly which card surface the grid composites over, so the worst of three grounds is what has to pass.

## Why a new module

**A comment cannot be run, and this is the second time that mattered on this issue.** The ramp is now `lib/correlationRamp.ts` — pure, no React imports — so `node --test` can load it and the caps are verified by compositing against `lib/terminalTokens.ts` rather than asserted in prose.

Four tests:

| | |
|---|---|
| every reachable cell, both themes, three grounds, clears 4.5:1 | 1,206 states swept |
| **the old 65% cap fails** | positive control |
| `r = ±1` produces exactly the caps | catches a cap that stops being reachable |
| the ramp is monotonic within each branch | the tint *is* the reading |

The positive control is there because a passing contrast sweep proves only that it ran. Without a known failure to detect, a clean result and a broken check are the same observation — which has now bitten this project in four forms.

## One test caught me, not the code

My monotonicity assertion crossed `r = 0`, where the green ramp's 4% floor sits below the red ramp's 6%. Comparing across colours is meaningless; **the test was wrong, not the ramp.**

It did surface a cosmetic quirk worth recording: `r` exactly `0` takes the `r > 0 ? … : …` false branch, so a **zero correlation renders a faint red tint** rather than a neutral one. Pre-existing, 6% alpha, deliberately left alone.

## How to test (QA)

`/correlation?design=terminal` in **light theme**:

1. Cells in the strongly negative band (`r` below about −0.8) are red-tinted — their numbers should be comfortably readable.
2. Compare with dark, already fixed by #571/#576.
3. **Nothing else should change.** The green ramp and every cell above `r = −0.8` are untouched.

## Risk level

**Low.** One constant, plus a pure-function extraction with no behaviour change. Gates: lint 0, `tsc` 0, `npm test` 0 (546), `build` 0.

**Not verified by me:** the rendered page. The composite arithmetic is tested against the real tokens, but I have not seen a light-theme cell at `r = −1`. Worth one look, and worth checking which surface the grid actually sits on — I tested all three plausible grounds rather than establishing it.

### DIFF
--- a/__tests__/correlationRamp.test.mts
+++ b/__tests__/correlationRamp.test.mts
@@ -41,7 +41,13 @@ const THEMES = [
    is what has to pass. */
 const GROUNDS = ['--bg0', '--bg1', '--bg2'];
 
-test('every reachable cell clears 4.5:1 in both themes, on any card ground', () => {
+test('every TINTED cell clears 4.5:1 in both themes, on any card ground', () => {
+  /* Tinted only, and the name says so. QA's review of #678 caught that this
+     read "every reachable cell" while sweeping only cells that carry a tint and
+     take --txt. The null cell is a different pair - --txt3 on
+     rgba(255,255,255,0.03) - and is covered by the test below rather than by
+     this one. An overstated coverage claim is exactly what this file exists to
+     prevent, so the name had to narrow. */
   const failures: string[] = [];
   for (const { name, T } of THEMES) {
     for (const ground of GROUNDS) {
@@ -89,4 +95,24 @@ test('the ramp is monotonic away from zero', () => {
     assert.ok(tintPct(i / 100) >= tintPct((i - 1) / 100), `positive ramp dipped at r=${i / 100}`);
     assert.ok(tintPct(-i / 100) >= tintPct(-(i - 1) / 100), `negative ramp dipped at r=${-i / 100}`);
   }
+});
+
+test('the null cell is --txt3, not --txt, and its margin is thin (#679)', () => {
+  /* NOT an assertion that it passes - it does not, on two of three dark
+     grounds. Recorded here so the numbers live next to the ones above rather
+     than in a comment, and so this file stops implying it covered them.
+
+         dark   --bg0 4.90  --bg1 4.40  --bg2 4.46
+         light  --bg0 5.70  --bg1 5.10  --bg2 4.74
+
+     This is #679: --txt3 clears AA by about 0.1 on --bg0 and spends that margin
+     on any raised surface. QA measured the same 4.46 on /liq's "current price"
+     label - different screen, live browser, different method, same token. It is
+     a palette question, not a correlation one, and spot-fixing it here would
+     make the token look healthier than it is. */
+  const T = TERMINAL_COLORS as Record<string, string>;
+  const nullCell = over('#ffffff', T['--bg1'], 3);
+  const c = ratio(hex(T['--txt3']), nullCell);
+  assert.ok(c < 4.5, `expected the known #679 shortfall on --bg1, measured ${c.toFixed(2)}`);
+  assert.ok(c > 4.0, 'if this drops below 4.0 the token moved and #679 needs re-measuring');
 });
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
## Approved — I re-derived your table from the tokens and got your numbers exactly

I did not take the sweep on trust, because the whole point of this PR is that the *previous* fix's numbers were taken on trust. Recomputed independently from `lib/terminalTokens.ts`, compositing `color-mix(in srgb, TINT p%, transparent)` over each ground and measuring `--txt`:

```
                          --bg0   --bg1   --bg2      first alpha that fails 4.5:1
  light  --red  #9d1a23    66%     63%     61%       ← binding
  light  --green #14702c   78%     76%     74%
  dark   --red  #f0524d    76%     74%     75%
  dark   --green #3fb950   64%     61%     61%
```

**Identical to your table, including that the two colours have opposite worst themes.** And the consequence lands where you said:

```
old cap 65%  light red on --bg2   4.09      ← the residual, below AA
new cap 56%  light red on --bg2   4.90
new cap 50%  dark green on --bg2  5.65
```

Red keeps 5 points of margin under its binding 61%, green keeps 11. **A comment could not have been checked this way; the module can, which is the actual fix here.**

The positive control is what makes the sweep mean anything — asserting the old 65% cap *fails* proves the harness can see the defect. That is the missing half of three of tonight's broken probes, and it is now in the repo as a pattern.

## One gap the tests do not cover — and it is the same token as `/liq`

`cellColor` at `:115` returns **`--txt3`** for a null cell, on `rgba(255,255,255,0.03)` over the card. The sweep only ever measures `--txt`, so this state is outside it. Composited:

```
                      --bg0   --bg1   --bg2
  dark   --txt3       4.90    4.40    4.46     ← fails on two of three
  light  --txt3       5.70    5.10    4.74
```

**Dark null cells fail on `--bg1` and `--bg2` and pass on `--bg0`** — so whether this is live depends on which surface the grid actually sits on, and your own comment says that is not pinned in CSS. **Conditional, and I am not claiming it is currently broken.**

What makes it worth acting on is that **this is the third independent sighting of the same root cause tonight.** `/liq`'s `← current price` measures 4.46 dark — `--txt3` on a raised card, the identical number to `--bg2` above. `--txt3` sits at ~4.6:1 against `--bg0`, so it passes on the page ground and fails on anything above it. Two screens, two probes, one token.

I would **not** widen this PR to fix that. But the sweep currently reads as "every reachable cell clears 4.5:1", and it means "every *tinted* cell". Either add `--txt3` on the null ground to the sweep, or narrow the test name and say why the null cell is excluded — a test that overstates its coverage is the thing this PR exists to prevent.

## Checked and clear

- **`tintPct` can never see null or NaN.** `cellBg:100` guards `r == null` first, and `pearson` returns `null` rather than NaN — `n < 5` and `denom < 1e-10` both exit early. No invalid `color-mix` percentage is reachable.
- **`--txt` is right.** The component sets cell text to `--txt` unconditionally, so the test's premise matches source rather than matching the test.
- **Caps are exactly reachable** — `tintPct(±1)` returns each cap, so the curve cannot quietly stop hitting them.

Leaving the `r === 0` faint-red quirk alone is correct: 6% alpha, pre-existing, and fixing it inside a contrast PR would blur what the PR proves.

**Merge it.**

— QA Team

Issue #679, opened 2026-09-03
Two composited surfaces land --txt3 below 4.5 — /liq's current-price span and /correlation's null cell


JohnDominicJasmin:
`--txt3` passes AA on the page ground and fails on every surface raised above it. It is not a broken token — it is a token with roughly a tenth of a point of margin, and that margin is spent by any card.

Three independent sightings tonight, two screens, two different measurement methods, all landing on the same number.

## The measurements

**`/liq`, deployed `qa` `4a326eb`, browser-composited** — `← current price`, a bare `<span>` with no class inside `div.gex-strike`:

```
dark    4.46 : 1     need 4.5
light   4.32 : 1     need 4.5
color   rgb(124, 130, 138)  = #7c828a  = --txt3
```

**`/correlation`, computed from `lib/terminalTokens.ts`** — the null-cell text (`CorrelationTerminal.tsx:115` returns `var(--txt3)`) on `rgba(255,255,255,0.03)` over the card:

```
             --bg0   --bg1   --bg2
  dark       4.90    4.40    4.46
  light      5.70    5.10    4.74
```

**`--bg2` dark is 4.46 in both places.** Same token, same shape, arrived at from a live browser on one screen and from the token source on another.

## Why this is a token problem and not two element problems

```
--txt3 dark   #7c828a  on --bg0 #08090a   ≈ 4.6 : 1
```

**It clears AA by about 0.1.** Every card, panel, chip and raised surface in the terminal design sits above `--bg0`, so every one of them consumes more than that margin. The token is fine exactly where the palette was checked and fails everywhere the palette was not.

Light behaves better — `#5e6267` holds 4.74 even on `--bg2` — so **dark is the failing side**, which is the opposite of the `/liq` green finding and worth not conflating with it.

## What a fix has to decide

This wants a rule, not three patches:

1. **Give `--txt3` headroom** — darken it in the light palette / lighten it in dark until it clears 4.5 on `--bg2`. Governed, one place, fixes every site at once. Costs the visual distinction between `--txt3` and `--txt2` some of its separation.
2. **Rule that card-mounted secondary text uses `--txt2`**, leaving `--txt3` for text on `--bg0` only. Keeps the palette, needs every call site audited, and the rule has to be written down somewhere it will be read.
3. **Something else** — but not a per-element override, which is how this became invisible in the first place.

[...]

PR #681, opened 2026-09-03
fix(dashboard): three dead terminal typography rules, and a check that keeps them dead

JohnDominicJasmin:
## Summary

#663's enforceable half. A check that fails when an inline style sets a typography property a terminal CSS rule also sets — and the **three live collisions it found on first run.**

## It found the class QA's runtime detector found on production

`.edge-card-value`, in three files. QA's `deadRules` check flagged the same class on **production**, from a browser, by reading computed values.

**Two methods — one static, one live — agreeing on the same class.** Neither found it by reading source, which is the whole argument of #663.

## What the collisions actually cost, measured

| sites | inline | terminal rule | effect |
|---|---|---|---|
| 8 | `var(--fs-data)` = **15px** | `.dash-main .edge-card-value` = **15px** | dead but **visually neutral** — which is why it survived |
| 2 | `var(--fs-label)` = **13px** | `.at-snapcells .edge-card-value` = **14px** | a real 1px divergence, invisible in source |

The first group is the interesting one: the rule could never apply and *nothing looked wrong*. That is the failure mode this issue is about.

## The current design is byte-identical

`globals.css:2380` already sets `font-size: var(--fs-data)` on the base class, so removing an inline copy of it changes nothing. `.is-label` sets exactly the 13px the inline value did.

**Only terminal changes, and only to what its own rules already asked for.** `.is-label` deliberately carries *lower* specificity than the terminal `.at-snapcells` rule, so the current design keeps 13px on the two word-valued cells while terminal gets its 14px — both intents survive.

## The test

A **ratchet** — it now finds zero. It carries a positive control that plants a collision on a real class taken from the parsed rule map and asserts it is caught in **both attribute orders**, because a passing sweep otherwise proves only that it executed.

Scope is typography on classes a terminal rule already styles. Colour and layout are legitimately inline throughout this codebase, and flagging them would bury the real case — the mistake #666's first version made in the other direction.

**Not covered, and stated in the file rather than implied:** the reverse collision, a CSS `!important` beating a component's inline value (#633's shape). **53 terminal rules use `!important` across 7 properties.** Live exposure, no detector.

## A self-inflicted bug worth recording

My scripted edit over-matched and added `.is-label` to two cells that carried `var(--fs-data)`, not `var(--fs-label)` — silently shrinking them 15px → 13px.

**The tests passed either way.** It was caught by auditing the diff for which token each removal actually carried. A check that verifies the end state does not verify that you changed the thing you meant to.

## Also in this branch

A correction to the `--txt3` comment #678 added. It repeated QA's #679 framing — *"clears AA by about 0.1 on `--bg0`"* — which QA has since retracted and I re-derived independently: `--txt3` passes on every bare palette ground (dark 5.14 / 4.71 / 4.77). Only the **composited** 3% white overlay fails, and only in dark. The assertion was always right; its explanation was not.

## How to test (QA)

`/dashboard` and `/dashboard?design=terminal`, both themes:

1. Edge-card values render as they do now.
2. **In terminal**, the two open-interest cells showing words rather than numbers move 13px → 14px, matching the rule that was always meant to apply.
3. Current design: no change anywhere.
4. `npm test` → 550.

## Risk level

**Low–Medium.** Ten inline declarations removed across three files, one CSS class added. Byte-identical in the current design by construction; terminal moves 1px on two cells.

**Not verified by me:** the rendered pages. The specificity argument is deterministic and the 15px/15px equivalence is arithmetic, but I have not seen either design after the change.

### DIFF
--- a/__tests__/correlationRamp.test.mts
+++ b/__tests__/correlationRamp.test.mts
@@ -97,40 +97,33 @@ test('the ramp is monotonic away from zero', () => {
   }
 });
 
-test('the null cell is --txt3, not --txt, and its margin is thin (#679)', () => {
-  /* NOT an assertion that it passes - it does not, on two of three dark
-     grounds. Recorded here so the numbers live next to the ones above rather
-     than in a comment, and so this file stops implying it covered them.
+test('the null cell clears 4.5:1 too, on --txt-dash (#679)', () => {
+  /* The null cell paints rgba(255,255,255,0.03) over the card. That overlay
+     lightens the ground toward the text, and --txt3 landed at 4.40 / 4.46 in
+     dark - so it takes --txt-dash instead, the value #559 created for exactly
+     this composited shape.
 
-         dark   --bg0 4.90  --bg1 4.40  --bg2 4.46
-         light  --bg0 5.70  --bg1 5.10  --bg2 4.74
+     --txt3 itself is NOT the problem and did not move: it passes on every bare
+     palette ground in both themes (dark 5.14 / 4.71 / 4.77). An earlier version
+     of this comment said it "clears by about 0.1 on --bg0 and spends that
+     margin on any raised surface" - both halves wrong, and the table above is
+     why. Fixing the token would have degraded three passing surfaces to rescue
+     one composited one, which qa/TERMINAL_REDESIGN_STATE.md already rejected. */
+  for (const { name, T } of THEMES) {
+    for (const ground of GROUNDS) {
+      const overlay = name === 'dark' ? '#ffffff' : '#000000';
+      const alpha   = name === 'dark' ? 3 : 4;
+      const c = ratio(hex(T['--txt-dash']), over(overlay, T[ground], alpha));
+      assert.ok(c >= 4.5, `${name} null cell on ${ground}: ${c.toFixed(2)}`);
+    }
+  }
+});
 
-     This is #679, and its cause is COMPOSITING, not the token. --txt3 passes on
-     every bare palette ground in both themes:
-
-         dark   --bg0 5.14  --bg1 4.71  --bg2 4.77
-         light  --bg0 5.68  --bg1 5.07  --bg2 4.70
-
-     What fails is --txt3 over a 3% WHITE OVERLAY on a card - a surface that is
-     not a palette member at all. The overlay lightens the ground toward --txt3
-     and eats the margin, in dark only.
-
-     An earlier version of this comment said --txt3 "clears AA by about 0.1 on
-     --bg0 and spends that margin on any raised surface". Both halves are wrong,
-     and the table above is why: bg0 is 5.14, and the raised surfaces pass. That
-     framing came from generalising two composited samples to the whole palette;
-     QA retracted it and I verified the retraction rather than swapping one
-     unchecked claim for another.
-
-     It matters because the wrong framing pointed at changing the token, which
-     qa/TERMINAL_REDESIGN_STATE.md already rejected: moving a value that passes
-     on three of four surfaces to fix one composited case degrades what works.
-     The precedent is the empty-cell dash taking its own #848a92 - a local
-     value for a local ground. QA measured the same 4.46 on /liq's current-price
-     label, which is the second instance of that same composited shape. */
+test('--txt3 would still fail there, so the swap is doing the work', () => {
+  /* Positive control for the fix itself. If --txt3 ever passes on this ground,
+     the reason for --txt-dash is gone and this should be revisited rather than
+     left as cargo. */
   const T = TERMINAL_COLORS as Record<string, string>;
-  const nullCell = over('#ffffff', T['--bg1'], 3);
-  const c = ratio(hex(T['--txt3']), nullCell);
-  assert.ok(c < 4.5, `expected the known #679 shortfall on --bg1, measured ${c.toFixed(2)}`);
-  assert.ok(c > 4.0, 'if this drops below 4.0 the token moved and #679 needs re-measuring');
+  const c = ratio(hex(T['--txt3']), over('#ffffff', T['--bg1'], 3));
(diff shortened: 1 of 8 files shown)


JohnDominicJasmin:
## Approved — and I found a false-negative path using your own harness

Ran the test as committed: **3 pass**, including the planted-collision control and the `RULES.size > 20` arming check. Three independent guards against the empty-sweep failure — the rule-map guard, the file-walk guard (`files.length > 50`), and the positive control in both attribute orders. That is the strongest version of this check anyone on this project has written.

## The gap: the style-object capture stops at the first `}`

`collisions()` grabs the inline object with `[^}]{0,600}`, so **anything containing a `}` before the offending property truncates the capture and the collision is never seen.**

I planted five shapes, each in its own file so the `[...new Set(hits)]` dedup could not merge them — my first attempt put all three in one file and got "1 dead rule", which is equally consistent with catching one or catching all three:

```
DETECTED   ZzPlain      style={{ fontSize: 12 }}
DETECTED   ZzRev        style={{ fontSize: 12 }} ... className    (reversed order)
DETECTED   ZzTernary    style={{ color: up ? 'a' : 'b', fontSize: 12 }}

MISSED     ZzTpl        style={{ transform: `translate(${x}px)`, fontSize: 12 }}
MISSED     ZzNested     style={{ ...(up ? { a: 1 } : { b: 2 }), fontSize: 12 }}
```

**A template literal with `${…}` and a spread of a nested object both close the capture early.** The ternary survives because it contains no braces at all — which is why the shapes that fail are exactly the ones that look most like real code in this repo.

So the sweep's current "zero" means **zero among collisions whose inline object has no braces before the property.** That is still a real ratchet and still worth merging. It just is not the whole set, and the file should say so rather than let a future reader take zero as none.

I would not fix this by growing the regex. A brace-counting scan from `style={{` — walk forward, track depth, stop at depth 0 — is a few lines and has no failure mode of this kind.

## Verified, not taken

- **Test run as committed:** 3/3 pass.
- **Planted collision, per-shape isolation:** the table above, with the dedup trap named because I nearly reported "1 of 3" off a merged count.
- **My own replication of your regex failed on every case including the plain one**, while your control passed. That meant my harness was broken, not your pattern — so I discarded it and used yours. Recording that because the wrong conclusion was one step away.

## The two groups in your breakdown

```
8 sites   inline var(--fs-data) 15px   vs rule 15px    dead, visually neutral
2 sites   inline var(--fs-label) 13px  vs rule 14px    a real 1px divergence
```

**The 8 neutral ones are the important half.** They are why this survived three separate defects: the rule could never apply and nothing looked wrong, so no one had a symptom to chase. A check is the only thing that finds those.

Your `.is-label` over-match — scripted edit silently shrinking two cells 15px → 13px, with tests passing either way — is the same lesson one level up: **verifying the end state does not verify you changed the thing you meant to.** Auditing the diff for which token each removal carried is the right control.

## Merge it

Add a line to the header naming the brace-truncation limit, or open a follow-up — your call, and neither blocks. The `!important`-beats-inline direction being stated in the file rather than implied to be handled is exactly right.

— QA Team

JohnDominicJasmin:
## Re-reviewed — the diff grew well past what I approved, and it holds. Merge.

You asked me to look rather than take your word, so I ran it and planted against it.

```
node --test __tests__/terminalTypographyOwnership.test.mts    5 pass, 0 fail
```

**Then I planted the two shapes that used to slip through, plus two that must NOT fire**, each in its own file so the dedup could not merge them:

```
FIRED       ZzTpl        style={{ transform: `translate(${x}px)`, fontSize: 12 }}
FIRED       ZzNested     style={{ ...(up ? { a: 1 } : { b: 2 }), fontSize: 12 }}

not fired   ZzOther      <div className="csb2-chg">  +  <em style={{ fontSize: 12 }}>
not fired   ZzCompound   <span className="on" style={{ fontWeight: 700 }}>
```

**Both directions.** The brace-depth walk catches what the regex hid, and neither guard over-fires: a style on a *different element* is ignored, and an element carrying only `.on` is not accused of the `.tnav-item.on` rule. That is the compound guard you asked me to see, working on a case built to break it.

## The two bugs my finding exposed are the bigger result

```js
new RegExp(`…\b${cls}\b`)        // template literal → \b is BACKSPACE
new RegExp('(^|;|\s)' + prop)    // single-quoted    → \s is 's'
```

**The sweep was reporting a clean codebase because it was comparing against nothing.** I found a capture that truncated; you found the matcher had never matched. Mine was the smaller half.

And it is the same bug I shipped tonight in the dead-rule detector — `'\s'` in a JS string collapsing to a literal `s`, which made that check under-report until you caught it. **Third instance in this codebase of a backslash dying inside a quoted string**, all three in detectors whose failure mode is silence.

> Both were caught by the positive control, which is the entire argument for having one.

That is the sentence I would put in `docs/HANDOVER.md`. A sweep that matches nothing and a clean codebase are the same observation, and only a control that is *known* to fail can tell them apart.

## The 31 → 0

Right to chase it rather than report it. Both false-positive sources were real:

- **Compound flattening** — crediting `.on` alone for `.tnav-item.on` accused every `ps-preset` toggle, 8 of the 31. A check that cries wolf gets switched off, which is worse than not having it.
- **className matched within 600 characters** — a class on a different element counted. My `ZzOther` case is exactly this and it now stays quiet.

**And the zero is verified, not assumed** — reinstating one real inline `fontSize` turns the run red on exactly that file, reverting returns it green. That is the reinstate-and-revert control, and it is the difference between "we found nothing" and "we would have found something."

## Known limits, and they are stated in the file rather than left to be found

- a brace inside a string literal inside the style object counts, which can only make a capture **too long** — a false positive someone looks at, never a silent miss
- the `!important`-beats-inline direction is uncovered: **53 terminal rules, 7 properties, no detector**

The second is a live exposure and I would rather it stay written down on #663 than get quietly folded in here.

**Merge it.** Then #679's two elements are already on this branch, so `/liq`'s marker and `/correlation`'s null cell land with it — I will re-measure both after the promotion and deploy.

— QA Team

PR #682, opened 2026-09-03
test(qa): derive platform-audit's palette from source, and stop a failed run reading as clean

JohnDominicJasmin:
## Summary

`qa/platform-audit.mjs` hard-coded both terminal palettes and both had drifted from `lib/terminalTokens.ts`. The light one had drifted far enough to **invert the check**. Same PR also stops an all-errored run from printing a clean-looking table.

## What changed

**1. The palette is derived, not copied.** The two arrays are replaced by a parse of `lib/terminalTokens.ts`, the same approach and the same loud guards `qa/mobile-audit.mjs` already uses.

**2. The summary cannot read as clean when nothing was measured.** `bad` — which held the *good* rows — is renamed `measured`, the run prints `measured N of M`, errored loads are listed rather than silently excluded, and the process exits non-zero.

No behaviour change to any measurement itself.

## Why

**The light list was scoring the wrong way round:**

```
still allowed, no longer tokens    #6a6e73  #8a5c00  #1a7f37  #cf222e  #9a6a00
real tokens, flagged as violations --txt3 #5e6267   --accent #754e00
                                   --green #14702c  --red #9d1a23
                                   --amber #755100  --fr-slight-long #7c5e2e
                                   --txt-dash #4f5257
```

The first two stale values are the ones design **replaced because they failed AA** — `--txt3` light `#6a6e73` measured 3.93:1 on `--bg2`, `--green` light `#1a7f37` measured 3.89:1, both recorded in `qa/TERMINAL_REDESIGN_STATE.md`. So the audit passed the values that fail and flagged the values that fixed them. **Acting on its output meant reverting a correctness fix.** The dark list had drifted less — two tokens missing, no stale extras — by the same mechanism.

`mobile-audit.mjs` was fixed this way on #641 and this file was never brought along. Its comment already says why: *a copied list drifts, and a stale hex is invisible among live ones.*

**The summary change is the same defect one level up.** Every count was computed over non-errored rows, so a run where nothing loaded printed six zeros under an `errors 4` line and read as a pass. That happened here: a shell mangled `--routes /dashboard` into a Windows path, all four loads errored, and the output was:

```
errors 4
overflowing            0 of 0
with contrast failures 0   (total 0)
...
```

The variable holding the successful rows was named `bad`, which is how it survived being read.

## How to test (QA)

On any deployed environment:

1. `node qa/platform-audit.mjs --base https://liquidity-hq-qa.onrender.com --routes /liq --themes dark,light --viewports desktop`
   **Expect 1 contrast failure in dark and 4 in light.** These match `qa/mobile-audit.mjs` on the same build — two tools, different code paths, same answer.
2. `node qa/platform-audit.mjs --base https://example.invalid --routes /x --viewports desktop --themes dark`
   **Expect `NOTHING WAS MEASURED`, the errored load listed, and exit code 1** — not a table of zeros. Check with `echo $?`.
3. The header comment names the five stale values and the seven missing ones, so the diff can be checked against `lib/terminalTokens.ts` directly.

On Windows/Git Bash, prefix with `MSYS_NO_PATHCONV=1` or the leading `/` in `--routes` is rewritten to a filesystem path — that is what produced the failure in point 2 for real.

## Risk level

**Low.** QA tooling only, no `app/`, `components/` or `lib/` files touched. It cannot affect the product.

The one thing to know: **off-palette counts will change on the next full run**, and light-theme counts should drop substantially. That is the fix working, not a regression — the previous numbers were measured against a list containing five non-tokens.

Verified in both directions: the derived palettes contain the current tokens and none of the five stale ones, and the all-errored run exits 1.

### DIFF
--- a/qa/platform-audit.mjs
+++ b/qa/platform-audit.mjs
@@ -14,6 +14,7 @@
  * time triples the wall clock for no isolation benefit here.
  */
 
+import { readFileSync } from 'node:fs';
 import { chromium, devices } from '@playwright/test';
 
 const arg = (name, dflt) => {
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
## Two more tools had the same defect, and a controlled measurement of what it cost

I swept the rest of `qa/` rather than fixing the one I tripped over. Two of the eighteen scripts still carried literal palettes, and both had drifted.

```
qa/token-surfaces.mjs    light --accent   #8a5c00  → source #754e00
                         light --red      #cf222e  → source #9d1a23
                         light --amber    #9a6a00  → source #755100
                         light --txt-dash #5e6267  → source #4f5257
                         missing --fr-slight-long (both themes), --txt-dash (dark)

qa/spec-conformance.mjs  light list current
                         dark missing --fr-slight-long, --txt-dash
```

**`token-surfaces` is the worse of the two despite having fewer wrong values**, because of a rule its own comment already records:

> An unrecognised foreground is SKIPPED, not counted — so a stale palette here silently omits exactly the tokens that were most recently changed.

A stale entry there does not produce a wrong number. It produces a **missing** one, and what goes missing is the set a run is usually checking. That comment was written after it happened once.

`qa/audit-handoff.mjs` also matches a hex grep and is **correctly left alone** — it is an old-value → new-value remap table, not a palette check.

## What the drift actually cost, measured

Same build (`qa` `33f5eee`), same three routes, same theme, desktop — **only the tool changed**:

```
                         OLD tool    NEW tool
off-palette instances       23          11      ← 12 were false
contrast failures            4           4
radius violations            3           3
sub-24px targets            11          11
```

**Just over half the light-theme off-palette instances were the tool's own stale list.** Every other metric is identical, which is the control that says I changed the palette and nothing else.

Empty fields read 17 then 15 across the two runs. **That is live-data variance between page loads, not the tool** — nothing in this change touches the empties check, and I am naming it rather than letting it look like an effect.

## Verified, not assumed

- `qa/token-surfaces.mjs` run end to end against `/liq` dark: prints a real per-token surface table and ends `dark: 1 failing surface across 1 token, out of 35 observed`. Not a throw, and not zero observed.
- A wrong parse throws by design in all three scripts rather than yielding an empty palette and reporting a clean sweep.
- Full 120-load run on `qa` `33f5eee` with the new code: `measured 120 of 120   errors 0`, so the new summary line reports honestly on a healthy run too.

## Note for whoever compares against older reports

**Off-palette counts before and after this PR are not comparable.** The earlier numbers were measured against a list containing five values that are no longer tokens while missing seven that are. Any `token-surfaces-*.md` in circulation predates this and under-reports the tokens it was most likely commissioned to check.

Same "How to test" as above, plus the `token-surfaces` step in the second commit body.

— QA Team

Issue #684, opened 2026-09-03
Empty fields: /econ-calendar 72 reproducible, /dashboard's six named, /scanner unstable (15-482)


JohnDominicJasmin:
Owner, 2026-09-03: *"some data not showing in dashboard pls check"* and *"check similar issue in other pages, issues like data not showing"*.

Measured on `qa` `33f5eee` with `qa/platform-audit.mjs`, 30 routes × desktop 1440 + mobile 390 × dark + light. **Answering the second half properly required naming the fields, not counting them** — the audit reported `/dashboard empty 7` and `/econ-calendar empty 72`, neither of which can be acted on. It now prints labels (QA PR pending), and those are what this issue is built from.

## The dashboard, named

```
/dashboard  desktop dark
  HYPE                    => "-"
  FUNDING RATE · BTC      => "-"
  OPEN INTEREST 1H · BTC  => "-"
  BTC volatility          => "-"
  BTC.D                   => "-"
  ALT                     => "-"
```

**This is the owner's report, reproduced and identified.** It is a small number of fields — 3 to 7 depending on the run — not a broken page.

## Every route with empty fields

Desktop 1440, both themes, `qa` `33f5eee`:

| route | dark | light | reads as |
|---|---|---|---|
| `/econ-calendar` | **72** | **72** | every event's actual/forecast column |
| `/scanner` | 15 | 15 | per-coin RSI / funding / long % cells |
| `/arena` | 10 | 10 | entry/stop/target/R:R and CVD |
| `/dashboard` | 3 | 3 | the six named above, varying by run |
| `/funding` | 2 | 2 | one or two symbols |
| `/briefing` | 1 | 1 | single field |
| `/correlation` | 0 | 50 | **see below — timing, not theme** |
| `/` | 16 | 4 | **see below — timing, not theme** |

Every other route measured zero. Mobile is broadly the same except **`/scanner` rises to 66** and `/econ-calendar` sits at 53 — a real viewport difference worth its own look.

## Two rows that are NOT findings

`/correlation` (0 vs 50) and `/` (16 vs 4) are **the same page measured twice at different moments.** A page's empty-field count cannot depend on the theme — the data path is identical — so the split is async data arriving before one load and after the other.

[...]

PR #685, opened 2026-09-03
test(qa): name the empty fields, and prove overflow instead of measuring width

JohnDominicJasmin:
## Summary

`qa/platform-audit.mjs` counted empty fields without recording which ones. This adds the labels. Follow-up to #682 — the same commit was written during that PR's review and did not land before it merged.

## What changed

Up to 12 distinct empty-field labels per page load, printed under the route's row. **The count itself is unchanged.**

```
desktop dark  /dashboard   ovf 0  off 1  rad 1  contrast 0  <24px 2  empty  7
              empty: HYPE => "-" | FUNDING RATE · BTC => "-" | OPEN INTEREST 1H · BTC => "-"
                     BTC volatility => "-" | BTC.D => "-" | ALT => "-"
```

## Why

The audit reported `/dashboard empty 7` and `/econ-calendar empty 72`. **Neither can be acted on** — *which* field is blank is the entire question, and a bare number sends the reader back to a browser to find out what the tool already knew and discarded.

This came up answering the owner's *"some data not showing in dashboard pls check"*. With counts alone I could confirm something was empty and nothing more. With labels the same run is a bug report — **#684 is written from this output** and would not exist otherwise.

The label is the nearest non-empty text within three ancestors, deduped, capped at 12. The cap matters: `/econ-calendar` has 72 and needs two examples, not 72 lines.

## How to test (QA)

```
MSYS_NO_PATHCONV=1 node qa/platform-audit.mjs --base https://liquidity-hq-qa.onrender.com \
  --routes /dashboard,/econ-calendar --viewports desktop --themes dark
```

1. `/dashboard` shows a non-zero `empty` count and an indented `empty:` line naming fields — expect `HYPE`, `BTC.D`, `ALT` and the two BTC metrics among them.
2. `/econ-calendar` shows `empty 72` and **at most 12** labels, not 72.
3. A route with `empty 0` — `/markets`, `/journal` — prints **no** `empty:` line at all.

On Git Bash the `MSYS_NO_PATHCONV=1` prefix is required or the leading `/` in `--routes` becomes a Windows path.

## Risk level

**Low.** QA tooling only — nothing under `app/`, `components/` or `lib/`. Output-only change; the counts, the contrast/overflow/palette checks and the exit codes are untouched.

Worth knowing: the labels come from `innerText` of ancestors, so a field whose label is an icon rather than text falls back to the class name or tag. That is a legibility limit, not a miscount — the count was already correct and still is.

### DIFF
--- a/qa/platform-audit.mjs
+++ b/qa/platform-audit.mjs
@@ -118,7 +118,7 @@ const PAGE_EVAL = ({ tokens, radiusExempt }) => {
      it attributes shell-wide colours to whichever route is being measured. */
   const isChrome = el => !!el.closest('.nav-menu, .gchat-panel, .app-bar, .nav-drawer, .pf-footer, .mobile-tab-bar');
 
-  const off = {}, radius = {}; let contrastFails = 0, subMin = 0, empties = 0, scanned = 0;
+  const off = {}, radius = {}, emptyLabels = []; let contrastFails = 0, subMin = 0, empties = 0, scanned = 0;
   const worstContrast = [];
   const BAD = /^(—|-|–|N\/A|--|NaN|null|undefined|CANNOT MEASURE)$/;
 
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## Second commit: the `ovf` column has been repeating a finding I retracted

Same file, same class of defect — a number that reads as a measurement and is not one.

`platform-audit` computed overflow as `scrollWidth - clientWidth`. **That is not overflow, it is width.** `html { overflow-x: hidden }` clips the closed nav drawer while `scrollWidth` still reports its extents, so the subtraction flags pages nobody can scroll.

The 120-load run on `qa` `33f5eee` flagged **4 of 60** mobile loads — `/dashboard` and `/liq`, both themes. Checked each against `qa/mobile-audit.mjs` on the same build:

```
/dashboard   horizontalOverflow=false   contentWiderThanViewport=true   678 vs 390
/liq         horizontalOverflow=false   contentWiderThanViewport=true   477 vs 390
```

**None of the four scrolls. The honest count is 0 of 60.**

This is the phantom 334px `/dashboard` overflow from #641 — the one I reported and retracted. `mobile-audit` was taught to prove scrollability then; **this file was never brought along, so it has been reproducing a retracted finding on every run since.**

## The fix keeps both numbers

`ovf` now counts only proven scrolling — try to scroll, check the offset sticks — and appends the unreachable extent in brackets so nothing is hidden:

```
mobile dark /dashboard   ovf 0(w288)  …  empty 5
mobile dark /liq         ovf 0(w87)   …  empty 0
                         overflowing 0 of 2
```

A page that genuinely scrolls sideways reports a non-zero `ovf` with no bracket.

## Consequence worth stating

**#559's headline "mobile overflows on 50/60" was measured with the unfixed comparison**, so an unknown share of those 50 were the same clipped drawer. I have said so on that issue rather than quietly reporting an improvement to a number that was never sound. I am not re-measuring `0d2e54a` to find out how many — the useful fact is that on a check which proves scrolling, nothing scrolls sideways on mobile today.

## How to test (QA), added to the steps above

```
MSYS_NO_PATHCONV=1 node qa/platform-audit.mjs --base https://liquidity-hq-qa.onrender.com \
  --routes /dashboard,/liq --viewports mobile --themes dark
```

Both rows read `ovf 0(wNNN)`, and the summary says `overflowing 0 of 2`.

Risk unchanged — **Low**, QA tooling only, no app code.

— QA Team

PR #687, opened 2026-09-03
docs: three more silent-detector failures from 2026-09-03

JohnDominicJasmin:
## Summary

Four more entries in `docs/HANDOVER.md` §14, "The failure that looks like success" — all from 2026-09-03, all sharing that section's shape: **the symptom is indistinguishable from the healthy state.**

## What changed

`docs/HANDOVER.md` only. Four paragraphs appended to §14, before §15.

## Why

**A backslash dies inside a quoted string, and the detector goes silent.** Three instances in one session, across both sessions' work:

```js
if (!/[.#]/.test(sel)) …            // fine - a regex literal
new RegExp('(^|;|\s)' + prop)       // '\s' is just the letter s
new RegExp(`…\b${cls}\b`)              // \b in a template literal is BACKSPACE
```

The first made `mobile-audit`'s dead-rule check under-report. The other two made the typography ownership sweep **match nothing at all** — and therefore report a clean codebase. All three are checks whose failure mode is finding nothing.

**A measurement over the wrong field gives a confident wrong answer.** A probe read `e.date || e.time || e.datetime` when the field is `isoDate`. Everything fell back to epoch and read as *past*, producing "19 past, 19 missing actual" — a plausible defect pointing at the **opposite** conclusion from the truth. Caught only because a printed sample row showed a September 4 date under a "past" heading. Same shape as `rsi symbols=2`, which counted the two top-level keys instead of the 46 inside `rsi`.

**A hex grep reads prose.** Sweeping `qa/` for stale palettes, a grep matched `token-surfaces.mjs` — inside a *comment* documenting the previous stale-palette incident. Files here deliberately record their own history in place, which makes grep-for-a-value structurally unreliable on them.

**Two tools measuring the same thing must be reconciled, or the weaker one keeps publishing.** `mobile-audit` proves overflow by scrolling; `platform-audit` subtracted widths. The second reproduced a phantom overflow QA had already retracted, on every run for weeks (#686).

## How to test (QA)

Read `docs/HANDOVER.md` §14. The four entries follow the "the correct pattern already existing in-repo does not mean it travelled" paragraph and precede §15.

**The one thing to check rather than skim:** the code block must show the **buggy** single-backslash forms — `'\s'` and `` `\b` ``. If it renders `'\s'`, that is the *corrected* form and the entry's point has been inverted. I made exactly that mistake writing it and caught it on re-read.

## Risk level

**Low.** Documentation only, one file, one section. No code, no tooling, no config.

### DIFF
--- a/docs/HANDOVER.md
+++ b/docs/HANDOVER.md
@@ -760,6 +760,53 @@ each copy explaining why it was needed. Copying is what stopped it spreading: a
 comment only reaches someone already reading that function. Prefer a check over a
 convention, because a convention has to be recalled at the moment of writing.
 
+**A backslash dies inside a quoted string, and the detector goes silent.** Three
+instances in one session, all in checks whose failure mode is *finding nothing*:
+
+```js
+if (!/[.#]/.test(sel)) …            // fine - a regex literal
+new RegExp('(^|;|\s)' + prop)       // '\s' is just the letter s
+new RegExp(`…\b${cls}\b`)              // \b in a template literal is BACKSPACE
+```
+
+The first was `qa/mobile-audit.mjs`'s dead-rule detector under-reporting; the
+other two were `__tests__/terminalTypographyOwnership`'s sweep **matching
+nothing at all** and therefore declaring the codebase clean. A regex literal is
+safe; the moment a pattern is built from a string — because it interpolates a
+variable — every backslash needs doubling, and nothing warns you. **Build
+patterns from plain strings with explicit `\\`, never from template
+literals**, and assume any new string-built matcher is broken until a positive
+control says otherwise.
+
+**A measurement over the wrong field gives a confident wrong answer, not an
+obviously broken one.** Investigating `/econ-calendar`'s 72 blank cells, a probe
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — checked the thing you asked me to check rather than skimming it, and the escapes are correct.**

The code block shows the **buggy** forms, which is what makes the entry readable:

```
new RegExp('(^|;|\s)' + prop)       single backslash - the bug
new RegExp(`…\b${cls}\b`)            single backslash - the bug
```

Your generator did not double them here. Had it, the entry would have shown *correct* code under a heading explaining a defect — which reads fine and teaches the opposite of the lesson. Worth having flagged.

One addition I would make if you touch it again: the third instance is the one that proves the point hardest. `'\s'` collapsing to `s` meant **barely any CSS rules were recognised as setting a property at all**, so the rule map came back near-empty and the sweep declared the codebase clean by comparing against nothing. Not under-reporting — reporting on an empty set.

Merging. Docs only.

— Dev Team

PR #688, opened 2026-09-03
fix(liquidation-map): green foreground gets its own value on composited tints

JohnDominicJasmin:
## Summary

The last three `/liq` failures. A governed `--green-fg` token, taken by the three elements that paint green text on a tinted ground.

## Measured

| element | before | ground |
|---|---|---|
| `.gex-net-chip` | **4.05** | its own 12% green tint |
| `.gex-value` | **4.36** | |
| `.liq-current-chg` | **4.38** | the current-price bar's accent wash |

`#14702c` on a 12% self-tint over `--bg1` is **4.35**; on the bar's 9% accent wash over `--bg2`, **4.19**. `#0f5a22` measures **5.87** and **5.66**, and 6.92 on bare `--bg1`.

## `--green` does not move

It passes everywhere else it is used, so changing it would shift **every green in the light theme** to fix three elements on one screen — the trade `qa/TERMINAL_REDESIGN_STATE.md` already rejected for `--txt3`.

This is the `--txt-dash` precedent again: a dedicated value for a composited ground. **Third time today that shape has been the answer.**

`--green-fg` is `--green-2` in every scope except terminal light, so the current design and terminal dark are **byte-identical**.

## The colour stays inline, deliberately

It is chosen from data — green when positive, `--red` when negative — which is the case #681's ownership check explicitly scopes out. Moving a data-driven ternary into a stylesheet means two rules and a class toggle to express one conditional.

## The token test earned itself

Declared in `lib/terminalTokens.ts` for both themes so `mobile-audit` governs it automatically (#645), and in `globals.css`'s terminal blocks to match.

`__tests__/contrastTokens` **failed** when I had added it to the token file with only an alias in the stylesheet — exactly the drift it exists to stop, on its first real exercise since #645. I would not have caught that by reading.

## How to test (QA)

`/liq?design=terminal` in **light**: the GEX net chip, per-strike values and the current-price change % are readable green rather than washed out. Dark unchanged; no design flag unchanged.

## Risk level

**Low.** One new token, three call sites. Gates: lint 0, `tsc` 0, `npm test` 0 (554), `build` 0.

**Not verified by me:** the rendered page. Composite arithmetic is measured against the real tokens; I have not seen a light-theme chip.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -121,6 +121,9 @@
      Worth collapsing into --green deliberately at some point. Two greens for
      one meaning is the reason this got missed in the first place. */
   --green-2:   #34d399;
+  /* #652: green foreground on a composited tint. Aliases --green-2 everywhere
+     except terminal light, where the composited grounds pull it under 4.5. */
+  --green-fg:  var(--green-2);
   /* Weaker-signal tier (e.g. "slightly high funding" vs "very high") - a
      paler version of --green/--red so the two tiers stay visually distinct. */
   --green-soft: #86efac;
@@ -5622,6 +5625,9 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
      - terminal's --txt3 is the identical #7c828a literal, so the same
      composited-surface failure applies here too. */
   --txt-dash: #848a92;
+  /* #652: matches lib/terminalTokens.ts. Dark's green passes on every ground
+     it meets, so this is --green; terminal light takes its own darker value. */
+  --green-fg: #3fb950;
   /* #546 C9: derived tints for --green/--red, same pattern as the already-
      existing --accent-bg/--accent-bdr. Undeclared here, .chg-up/.chg-dn's
      background/border (app/globals.css:786-787) fell through to the current
@@ -5794,6 +5800,7 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
   --bg3: var(--bg2);
   --bg4: var(--bg1);
   --green-2:    var(--green);
+  --green-fg:   #0f5a22;
   --green-soft: var(--green);
   --red-soft:   var(--red);
   --txt-dim:    var(--txt2);
--- a/components/GexTable.tsx
+++ b/components/GexTable.tsx
@@ -19,7 +19,7 @@ export default function GexTable() {
   const gexLoaded = btcNetGex !== null && btcGexLevels.length > 0;
   const isLongGamma = (btcNetGex ?? 0) >= 0;
 
-  const gexCol     = isLongGamma ? 'var(--green-2)' : 'var(--red)';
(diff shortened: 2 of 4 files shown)


JohnDominicJasmin:
## Approved — I computed the new value against six grounds, not the two you quoted

Merge it. This is the right shape: a **foreground-only** token for a local ground, exactly the `--txt-dash` precedent, with the tints and borders still on `--green-2` so nothing visual moves except the text that was failing.

Composited `#0f5a22` against every ground these three elements can land on:

```
                                       old     new
.gex-value      flat card --bg2        4.75    6.42   PASS
.gex-value      flat card --bg1        5.12    6.92   PASS
.gex-net-chip   green 12% on --bg2     4.05    5.47   PASS   ← was the worst
.gex-row-atm    white 3% on --bg2      4.79    6.47   PASS
.liq-current-chg accent 6% on --bg2    4.38    5.91   PASS
.liq-current-chg accent 6% on --bg1    4.70    6.35   PASS
```

**Every ground clears, and the worst case moves from 4.05 to 5.47** — margin rather than a value sitting on the line.

**One small divergence worth naming rather than smoothing over.** Your comment says `5.87 / 5.66 on the two tints`; I get **5.91 / 5.47**. The accent tint agrees to within rounding, the green tint does not. Most likely we composited over different card grounds — I used `--bg2`, and `--bg1` gives me 5.66 for that chip, which is exactly your number. So we are probably both right about different surfaces, and since the *lower* of the two still clears by nearly a point it changes nothing. Flagging it because "close enough" is how a wrong ground assumption survives, and this screen has already produced one unattributable ground tonight.

## The token is governed, which is the part that makes it stick

```
lib/terminalTokens.ts   TERMINAL_COLORS        '--green-fg': '#3fb950'
                        TERMINAL_COLORS_LIGHT  '--green-fg': '#0f5a22'
```

So `qa/platform-audit.mjs` and `qa/mobile-audit.mjs` will both accept it — they derive their palettes from that file now (#682), so a new token is recognised without anyone remembering to add it anywhere. **A hard-coded palette would have reported this fix as an off-palette violation**, which is the failure mode I spent tonight removing from three scripts.

Dark aliasing to `--green` rather than taking its own literal is right too: dark's green passes on every ground it meets, so a separate value there would be two things to keep in step for no benefit.

## Coverage

Three failures measured, three call sites changed — `gexCol`, `vcol`, `liq-current-chg`. **The counts match**, and `gexBg` / `gexBorder` correctly keep `--green-2` since a tint is not a foreground.

I will re-measure `/liq` in both themes on the deployed build after the promotion. If this holds, `/liq` closes: cluster-ladder layout verified, `← current price` fixed, swatch hit area fixed, greens fixed, empty feed documented as architecture rather than a defect.

— QA Team

Issue #692, opened 2026-09-03
/econ-calendar fades past events to 0.45 opacity — ~1.9:1 in both themes, latent until an event passes


JohnDominicJasmin:
`app/econ-calendar/page.tsx:227` fades a whole event row when it is past and has no actual value:

```jsx
opacity: isPast && !e.actual ? 0.45 : 1
```

At 45% opacity `--txt3` composites to roughly **1.9:1** in both themes — far worse than any failure currently on the page, and worse than anything measured on the platform tonight.

```
--txt3 at 45% opacity     dark 1.93     light 1.87
```

## It is not firing right now, and that is the problem

All 19 events on the calendar are in the future, so `isPast` is false and every row renders at full opacity. **All 23 of the screen's current contrast failures measured `opacity: 1`** — this rule contributed none of them.

**So the defect is latent and self-triggering.** No code change is needed to make it appear: the moment an event's scheduled time passes without an `actual` value arriving from the feed, that row drops to ~1.9:1. Given #684's other half — `previous` and `estimate` missing on 17 of 19 events, so the ingest is already incomplete — rows reaching that state is likely rather than hypothetical.

It will also fail in **both themes at once**, unlike the current failures which are dark-only. Anyone watching the contrast count will see it jump with no deploy to blame it on.

## Why it is filed separately from #684

#684 is about fields rendering `-`. This is about a row that renders its values and makes them unreadable. Different defect, different fix, and folding it in would let "fixed the 23" close an issue that still contains this.

Credit: found by dev while investigating #684's contrast half, and correctly reported as *not* the cause rather than being made to fit the count.

## Suggested direction, not a prescription

The intent — de-emphasise a stale row — is legitimate. Opacity on the whole row is the wrong instrument for it, because it scales the text's contrast against its own background rather than restating the colour. The `--txt-dash` / `--green-fg` treatment used three times today is the shape that has worked: **give the de-emphasised state its own token value rather than multiplying the live one.**

Not assigning. `/econ-calendar`'s current 23 failures come first — 20 impact badges and 4 overlay cells, both named on #684 with measurements.

— QA Team

JohnDominicJasmin:
## Closing — the rule is gone, and there was no value it could have been given

Removed in #694, merged and deployed on `qa` `2ef168c`.

## Why removal rather than a smaller number

[...]

PR #693, opened 2026-09-03
fix(econ-calendar): the impact badge and the next-event row, all 23 dark failures

JohnDominicJasmin:
## Summary

All 23, from your per-element output rather than a guess.

```
20x  impact badge    --red on its own 14% wash        4.04 worst
 4x  next-event row  --txt3 on a 2.5% white overlay   4.46
```

## Tokenising fixes most of the badge

The tint literal was `rgba(248,113,113,0.14)` — **the current design's `--red`**. Terminal was painting a lighter ground than its own palette gives, which is why your measured **4.04** is worse than the **4.49** the tokenised tint produces.

`--red` *is* `#f87171` at `:root`, so the current design is byte-identical. Only terminal moves, and only to its own colour. Amber and the grey follow for the same reason.

4.49 still misses `--bg1` by **0.01**, so HIGH's foreground takes **`--red-fg`** — the `--green-fg` counterpart from #688. `#ff8a85` dark measures 6.88 worst-of-three, `#7d0f18` light 6.52.

## The four muted cells could not be reached by CSS

They set colour inline, which outranks every selector (#663) — a rule written for them would have been **dead on arrival**, in a file that already had that problem.

But an inline value that *reads* a custom property still inherits. So the row publishes `--ec-muted` and moves all four without a class toggle per cell. Same collision #681 catches, solved without splitting inline and CSS.

## Not in this PR

**The opacity fade.** All 23 measured `opacity: 1`, so it is not part of these numbers — and #692 is the right home for it. Folding it in would let "fixed the 23" close an issue that still contains a worse defect.

## How to test (QA)

`/econ-calendar?design=terminal` in **dark**: red HIGH badges readable against their wash; the highlighted next-event row's muted text (`in 28h 21m`, country code, `-` placeholders) legible. **Light unchanged** — it had zero failures. No design flag byte-identical.

## Risk level

**Low–Medium.** Three tint literals tokenised, one new token, one inherited property. Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0.

**Not verified by me:** the rendered page. Arithmetic is against the real tokens; I have not seen a badge.

### DIFF
--- a/app/econ-calendar/page.tsx
+++ b/app/econ-calendar/page.tsx
@@ -68,9 +68,26 @@ function calcDelta(actual?: string, estimate?: string): { text: string; positive
    painted the whole calendar in the LOW style. See lib/classify.ts.
    LOW's colour also failed AA on its own (#6b7280 = 3.77:1). */
 const IMPACT_CFG: Record<EconImpact, { color: string; bg: string; border: string; accent: string }> = {
-  HIGH:   { color: 'var(--red)', bg: 'rgba(248,113,113,0.14)', border: 'rgba(248,113,113,0.35)', accent: 'var(--red)' },
-  MEDIUM: { color: 'var(--amber)', bg: 'rgba(251,191,36,0.12)',  border: 'rgba(251,191,36,0.3)',   accent: 'var(--amber)' },
-  LOW:    { color: 'var(--txt-dim)', bg: 'rgba(107,114,128,0.10)', border: 'rgba(107,114,128,0.25)', accent: 'rgba(107,114,128,0.4)' },
+  /* #684: tints derive from the tokens instead of rgba() literals, and HIGH's
+     foreground takes --red-fg.
+
+     20 of this screen's 23 dark contrast failures were this one badge - --red
+     on a wash of itself, worst 4.04, the same arithmetic as .gex-net-chip's
+     4.05 on /liq that #688 just answered.
+
+     TWO SEPARATE CAUSES, and tokenising fixes most of the first. The literal
+     was rgba(248,113,113,...) - the CURRENT design's --red - so terminal was
+     painting a lighter ground than its own palette would, which is why the
+     measured 4.04 was worse than the 4.49 the tokenised tint gives. --red IS
+     #f87171 at :root, so the current design is byte-identical; only terminal
+     moves, and only to its own colour. Amber and the grey follow the same
+     rule for the same reason.
+
+     4.49 still misses on --bg1, so HIGH's foreground takes --red-fg - the
+     --green-fg counterpart from #688, aliased to --red wherever it passes. */
+  HIGH:   { color: 'var(--red-fg)', bg: 'color-mix(in srgb, var(--red) 14%, transparent)', border: 'color-mix(in srgb, var(--red) 35%, transparent)', accent: 'var(--red)' },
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
## Approved — every number reproduces, and you found a cause I missed

Merge it. Recomputed independently from `lib/terminalTokens.ts`:

```
dark   HIGH badge, --red on its own 14% wash    --bg1 4.49   --bg2 4.57
       with --red-fg #ff8a85                    --bg1 6.88   --bg2 7.00
light  with --red-fg #7d0f18                    --bg1 7.00   --bg2 6.52
```

**`4.49 over --bg1 — short by 0.01` and `6.52 worst of three grounds` both match my figures exactly.**

**The cause I missed is the better half of this PR.** I reported the badge as `--red` on its own tint and left it there. You found the tint was `rgba(248,113,113,…)` — **the current design's `--red`, not terminal's** — so terminal was painting a ground from a palette it does not use, and the measured 4.04 was worse than the 4.49 its own colour would give. Tokenising fixes most of it; `--red-fg` closes the remaining 0.01.

That is the same class as the palette drift in #682: a literal from one palette living inside another, invisible to everyone reading either.

## The `--ec-muted` custom property is the right answer to a real constraint

An inline `color` outranks every selector, so a CSS rule could not reach those four cells — #663's exact finding. **But an inline value that *reads* a custom property still inherits**, so the row moves all four without a class toggle per cell. That threads the needle rather than working around it.

## One side effect to state, not to fix

Swapping `rgba(255,255,255,0.025)` for `color-mix(in srgb, var(--txt) 2.5%, transparent)` makes the overlay **theme-aware**, which is right — but it means light's overlay now *darkens* a light card where the white one lightened it:

```
--txt3 on the isNext row      OLD overlay      NEW overlay
  dark  --bg1                 4.45 FAIL        4.48 FAIL
  dark  --bg2                 4.52             4.54
  light --bg1                 5.09             4.83
  light --bg2                 4.74             4.48  FAIL
```

**In light on `--bg2`, `--txt3` crosses from passing to failing on that row.**

**This PR is still correct**, because every cell on the isNext row now reads `--ec-muted` → `--txt-dash`, which measures 5.73 there. I checked all five call sites (`:266`, `:271`, `:288`, `:297`, `:305`) and none uses `--txt3` directly. Nothing fails today.

What changes is the **hazard**: that row used to be a dark-only trap and is now an all-theme one, currently masked. Anyone adding a `--txt3` cell to it later gets a light failure as well as a dark one, and the dark-only pattern in every other finding today would lead them to expect otherwise. Worth a line in the comment at `:248` rather than a code change.

`--amber`, `--red` and `--green-fg` all clear on that row in both themes (worst 4.98), so the other foregrounds are fine.

## `/correlation`, since #679 covers both

Measured on `staging` `080b7a5`, 2600 cells, both themes:

```
dark   0 fails   worst 4.71   colours: --txt3 (100 headers), --txt (2500 values)
light  0 fails   worst 5.07   colours: --txt3 (100 headers), --txt (2500 values)
```

**Clean — but `--txt-dash` does not appear at all**, because every pair currently has a correlation and no null cell renders. So your `/correlation` half of #679 is **correct in source and not observable today**. I am recording that rather than claiming I verified it, the same way the cluster ladder's data path is recorded as unverified.

— QA Team

PR #694, opened 2026-09-03
fix(econ-calendar): drop the 0.45 stale fade - opacity cannot carry this

JohnDominicJasmin:
## Summary

#692. `opacity: isPast && !e.actual ? 0.45 : 1` put every cell in a stale row at **~1.9:1 in both themes** — the worst contrast on the screen, on rows that render real values.

## Opacity cannot be tuned out of it

Minimum opacity each token needs to hold 4.5:1, worst of `--bg1`/`--bg2`:

| token | dark | light |
|---|---|---|
| `--txt` | 50% | 62% |
| `--txt2` | 87% | 94% |
| `--txt3` | **97%** | **98%** |
| `--txt-dash` | 91% | 87% |

**Only `--txt` survives a fade anyone would notice.** Every muted token needs ~90%+, which is not a fade. So the mechanism is wrong, not the value — there is no number to pick here.

## What replaces it

Staleness reads from `--ec-muted`: a stale row's muted text takes `--txt-dash`, which passes on every ground on this screen, and the `-` placeholders already say the data never arrived.

**That is a weaker visual signal than a 45% fade, and it is deliberate.** #635's ruling was that an honest placeholder beats a confident wrong one, and an unreadable row is worse than a quiet one.

**If design wants staleness louder, a background tint or a left border moves no text contrast at all** — noted in the comment so the next person has the option without rediscovering why opacity is not it.

## Latent, not live

All 23 of #684's failures measured `opacity: 1`, so this never appeared in those numbers. It fires when an event passes without an actual — and with `previous`/`estimate` already missing on 17 of 19, that is likely rather than hypothetical.

## How to test (QA)

Hard to reach on demand: it needs a past event with no actual value. `/econ-calendar?design=terminal`, both themes — **no row should render faded**. If one does, this did not apply.

The observable change today is none, which is the expected result while every event is still in the future.

## Risk level

**Low.** One property removed, one conditional widened. Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0.

**Not verified by me:** the stale state itself, for the same reason it was latent — I cannot make an event pass.

### DIFF
--- a/app/econ-calendar/page.tsx
+++ b/app/econ-calendar/page.tsx
@@ -269,8 +269,34 @@ export default function EconCalendarPage() {
                          outranks every selector (#663) - but an inline value that READS a
                          custom property still inherits, so the row can move all four at
                          once without a class toggle per cell. */
-                      ['--ec-muted' as string]: isNext ? 'var(--txt-dash)' : 'var(--txt3)',
-                      opacity: isPast && !e.actual ? 0.45 : 1,
+                      /* #692: NO opacity fade. It was 0.45, which put every
+                         cell in a stale row at ~1.9:1 in BOTH themes - the
+                         worst contrast on the screen, on rows that render real
+                         values.
+
+                         Opacity cannot be tuned out of this. Minimum opacity
+                         each token needs to hold 4.5:1, worst of --bg1/--bg2:
+
+                             --txt   50% dark  62% light
+                             --txt2  87%       94%
+                             --txt3  97%       98%
+                             --txt-dash 91%    87%
+
+                         Only --txt survives a fade anyone would notice; every
+                         muted token needs ~90%+, which is not a fade. So the
+                         mechanism is wrong rather than the value.
+
+                         Staleness now reads from --ec-muted instead: the row's
+                         muted text takes --txt-dash, which passes on every
+                         ground here, and the "-" placeholders already say the
+                         data never arrived. That is a WEAKER visual signal than
+                         a 45% fade and it is deliberate - #635's ruling was that
+                         an honest placeholder beats a confident wrong one, and
+                         an unreadable row is worse than a quiet one. If design
+                         wants the staleness louder, a background tint or a left
+                         border moves no text contrast at all. */
+                      ['--ec-muted' as string]: isNext || (isPast && !e.actual)
+                        ? 'var(--txt-dash)' : 'var(--txt3)',
                       minWidth: 680,
                     }}
                   >


JohnDominicJasmin:
## Approved — I rebuilt the table and it says the same thing: there was no value to pick

Merge it. Minimum opacity each token needs to hold 4.5:1 against its own ground, computed independently from `lib/terminalTokens.ts`:

```
             yours            mine
--txt        50 / 62         50 / 61
--txt2       87 / 94         87 / 91
--txt3       97 / 98         97 / 95
--txt-dash   91 / 87         91 / 85
```

Dark agrees exactly. Light differs by 1–3 points, which is a **ground assumption**, not a disagreement — I used `--bg1` and the light figures move that much between `--bg1` and `--bg2`. **It changes nothing: every muted token still needs 85%+ and the conclusion is identical.** Naming it because that is the third time today our numbers diverged on an unstated ground, and each time it has been worth pinning rather than waving through.

At the shipped 0.45 I measure `--txt3` **1.93 dark / 1.87 light** and `--txt-dash` 2.04 / 2.03 — matching what this issue was filed on.

## The finding is the mechanism, not the number

**Only `--txt` survives a fade anyone would perceive.** Everything muted needs ~90%+, and a 90% opacity is not a fade — it is a rounding error with a performance cost. So there was never a safe value to tune 0.45 down to, which is exactly the kind of thing that looks like a tuning problem until someone computes the floor.

**Removing it beats fixing it, and reverting to the white literal would have been worse** — that reintroduces the palette-crossing bug #693 exists to fix. Recording the hazard in the comment rather than reverting was right on both counts.

## The replacement signal being weaker is the correct trade

`--ec-muted` → `--txt-dash` plus the `-` placeholders that already say the value never arrived. That is quieter than a fade, and #635's ruling covers it: **an honest placeholder beats a confident wrong one**, and an unreadable row is worse than a quiet one.

Noting the alternative in the comment — a background tint or a left border moves no text contrast — is the part that stops this being rediscovered. Someone will ask for a louder stale state, and the answer to "why not opacity" needs to be in the file rather than in this thread.

## Both of your caveats are right and I will not contradict them by measuring

**You cannot make an event pass, so the stale state has not been seen.** And the observable change today is **none** — all 19 events are in the future, `isPast` is false, and this code path does not execute. A run showing "nothing moved" is the expected result, not evidence the fix works.

So this is verified by **computation and code reading**, not by observation, and I am recording it that way. That is the third item on the board in that category, alongside `/correlation`'s null cell and the cluster ladder's data path. Worth keeping visible: three fixes are correct-by-reading and unobservable in this environment, and they will only ever be exercised by real usage.

**Merge, and I'll deploy and confirm the no-change is a no-change** — that at least rules out a regression from the removal.

— QA Team

PR #700, opened 2026-09-03
docs(qa): record the -fg token pattern, and the two cases it does not solve

JohnDominicJasmin:
## Summary

`qa/TERMINAL_REDESIGN_STATE.md` already carried the ruling that kept `--txt3` dark at `#7c828a` — reading it is what stopped me filing a recommendation that had already been considered and rejected. **It did not carry what replaced that recommendation**, so the next reader finds the rejection without the alternative.

## What changed

Section 8 gains: the three composited-ground foreground tokens shipped 2026-09-03, the recurring failure they answer, **two cases the pattern cannot answer**, and the opacity floor table.

```
token         dark               light      for
--txt-dash    #848a92            #4f5257    muted text on a translucent overlay
--green-fg    #3fb950 (= base)   #0f5a22    positive value on a green tint
--red-fg      #ff8a85            #7d0f18    negative value on a red tint
```

## Why

**The recurring failure is a signal colour on a wash of itself** — `/liq`'s chip at 4.05, `/econ-calendar`'s badge at 4.04, `/scanner` at 3.51–4.23 across four tint strengths. Each `-fg` token aliases its base wherever the base passes, so only the composited case moves. **Same reasoning as the `--txt3` ruling above it**: fix what is local, do not degrade what works.

**Both limits are recorded because both were found the hard way:**

- **A data-driven tint can outrun any foreground.** `/scanner`'s green tint puts `--green-fg` at 3.73 — stronger than the wash it was sized against — and a new value would be beaten by the next stronger mover. `/correlation`'s answer (text to `--txt`, tint carries the meaning) is the one that cannot be reopened, and it measures 2450 cells at zero on production.
- **A theme-aware overlay is not automatically safer than a literal.** Tokenising one correctly stopped a dark-palette literal painting in light *and* took `--txt3` on that row from 4.74 to 4.48. Nothing failed, but the row went from a dark-only hazard to an all-theme one.

**The opacity table is there so nobody tries to tune a fade again.** Only `--txt` survives one; everything muted needs 85%+.

## How to test (QA)

Read `qa/TERMINAL_REDESIGN_STATE.md` section 8.

1. New material follows the *"arithmetic, not yet a measurement"* paragraph and precedes section 9.
2. The three-token table, both limits and the opacity floors are present.
3. **The existing `--txt3` dark ruling above it is unchanged** — this adds to that decision, it does not revisit it.

## Risk level

**Low.** QA documentation only. No code, no tooling, no config.

### DIFF
--- a/qa/TERMINAL_REDESIGN_STATE.md
+++ b/qa/TERMINAL_REDESIGN_STATE.md
@@ -549,6 +549,35 @@ because every cell on that row had already moved to `--txt-dash`, but the row
 went from a dark-only hazard to an all-theme one. **A tokenisation can move a
 number in the direction nobody was watching.**
 
+### A literal from the other palette is invisible from both sides
+
+**Three instances on 2026-09-03**, and they share one signature: **a composited
+ground that corresponds to no token in the palette being audited.**
+
+| where | the literal | what it actually was |
+|---|---|---|
+| `/econ-calendar` impact badge | `rgba(248,113,113,…)` | current design's `--red` |
+| `/scanner` heatmap tiles | `rgba(52,211,153,…)`, `rgba(248,113,113,…)` | current design's `--green-2`, `--red` |
+| `qa/platform-audit.mjs` | five hard-coded hexes | superseded light values |
+
+The failure is symmetric and that is what makes it hard to see. **Reading the
+terminal stylesheet, the value is not there. Reading the current design's, it is
+correct.** Only a composited measurement shows it, and then only as a ground
+that matches nothing — on `/scanner` the measured grounds were `rgb(30,78,62)`
+and `rgb(77,44,46)`, neither of which is any tint of terminal's `--green` or
+`--red`. **QA recorded both as observed values and moved on; the mismatch was
+the finding.**
+
+It also skews the contrast in a way that hides its own size: terminal was
+painting a *lighter* ground than its own palette would, so `/econ-calendar`'s
+badge measured 4.04 where its own colour gives 4.49 — the literal made the
+defect look worse than the token error it was, and tokenising alone closed most
+of the gap.
+
+**The check that catches it is compositing, not reading.** `color-mix(in srgb,
+var(--token) N%, transparent)` cannot cross palettes; an `rgba()` literal always
+can, and no source read on either side will say so.
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## Second commit: the cross-palette literal, which happened a third time while this PR was open

Added a fourth entry to the same section. **Three instances on 2026-09-03, diagnosed independently each time**, sharing one signature.

| where | the literal | what it actually was |
|---|---|---|
| `/econ-calendar` impact badge | `rgba(248,113,113,…)` | current design's `--red` |
| `/scanner` heatmap tiles | `rgba(52,211,153,…)`, `rgba(248,113,113,…)` | current design's `--green-2`, `--red` |
| `qa/platform-audit.mjs` | five hard-coded hexes | superseded light values |

## The signature, and why it is worth writing down

**A composited ground that corresponds to no token in the palette being audited.**

On `/scanner` I measured `rgb(30,78,62)` and `rgb(77,44,46)`. Neither is any tint of terminal's `--green` or `--red`. **I recorded both as observed values and moved on — that mismatch was the finding**, and it took dev reading the component to name it.

The failure is symmetric, which is what makes it hard: **reading the terminal stylesheet the value is absent; reading the current design's it is correct.** Nobody is looking at both at once.

It also skews the number in a way that hides its own size. Terminal was painting a *lighter* ground than its own palette gives, so `/econ-calendar`'s badge measured **4.04** where its own colour gives **4.49** — the literal made the defect look worse than the token error it was, and tokenising alone closed most of the gap. A fix aimed at the measured 4.04 would have overshot.

## What catches it

**Compositing, not reading.** `color-mix(in srgb, var(--token) N%, transparent)` cannot cross palettes; an `rgba()` literal always can, and no source read on either side will say so.

That is now three of the four entries in this section that were only findable by measuring a composited surface — the `-fg` pattern, the theme-aware overlay, and this.

## Unchanged

Same file, same section, docs only. The existing `--txt3` dark ruling is still untouched, and the first commit's content is unmodified.

**Test step, added to the ones above:** the new entry sits between *"a theme-aware overlay is not automatically safer"* and *"opacity is not a fade tool here"*, and carries the three-instance table.

— QA Team

PR #701, opened 2026-09-03
fix(scanner): the coin-heatmap tile's ramp crossed two palettes

JohnDominicJasmin:
## Summary

All 26 of `/scanner`'s dark failures are `CoinHeatmap`'s tiles. **Two causes, and the first is the one neither of us had seen.**

## The tints were the current design's palette

`rgba(52,211,153,…)` and `rgba(248,113,113,…)` — the **current design's** `--green-2` and `--red` — while the text resolves through terminal's tokens. **Terminal was painting the tile in one palette and its number in another.**

That is why your measured grounds (`rgb(30,78,62)`, `rgb(77,44,46)`) did not correspond to anything in terminal's own palette. Third instance of this shape today, after #693's impact badge and #682's audit palettes.

## Tokenising alone does not fix it

The ramp **lightens the text as the tile saturates** — `#86efac` then `#fee2e2` — so severity rides on the tile while the number stays readable, 8.6:1 to 9.6:1 in the current design.

**In terminal, `--green-soft` and `--red-soft` alias flat to `--green`/`--red`**, so that lightening never happens. The strongest tiles put a mid-tone signal colour on a saturated wash of itself. The mechanism the current design relies on is simply not there.

## So terminal takes your option 3

Measured after, terminal dark, **ground: the tile's own tint over `--bg1`**:

```
green 30%   8.74      red 38%   8.71
green 22%  10.37      red 25%  10.93
green 13%  12.36      red 15%  12.72
```

against the 3.51–4.23 that failed. It cannot be reopened by a stronger tint, which matters because the tint strength is data-driven.

**Direction survives the colour** — the tile tint and the `+`/`-` sign both still carry it, which is also what keeps this clear of WCAG 1.4.1.

## Current design byte-identical

`--heat-fg` is unset there, so each span falls back to its own ramp colour.

A fallback rather than a class per bucket because the text colour is data-driven — a stylesheet version needs one rule per ramp step. `var(--heat-fg, <ramp>)` lets terminal override all three spans with one declaration. Same shape as `--ec-muted` on #693.

## How to test (QA)

`/scanner?design=terminal` in **dark**: tile symbol, price and percentage are white on the coloured tile rather than a deeper shade of the tile's own colour. **Strongest movers are the ones to check** — they had the worst contrast. Light unchanged; no design flag byte-identical.

## Risk level

**Low–Medium.** Nine tint literals tokenised, one CSS hook. Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0.

**Not verified by me:** the rendered tiles. Arithmetic is against the real tokens; I have not seen one.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -7477,6 +7477,28 @@ main.app-content[data-chrome="none"] {
    different ground and was not measured. */
 [data-design="terminal"] .liq-current-oi { color: var(--txt2); }
 
+/* #698: the coin-heatmap tile's text in terminal.
+   The tile's colour ramp was built for the CURRENT design - text lightens as
+   the tile saturates (#86efac, #fee2e2), so severity is carried by the tile
+   while the number stays readable, 8.6:1 to 9.6:1 across the ramp. In terminal
+   --green-soft and --red-soft alias flat to --green/--red, so that lightening
+   does not happen and the strongest tiles put a mid-tone signal colour on a
+   saturated tint of itself: 3.51 worst, 26 failures on /scanner (#698).
+
+   Terminal takes --txt and lets the tint carry direction, which is what
+   /correlation already shipped on #570 - 2368 failing cells fixed exactly that
+   way, now measuring 2450 with zero failures on production. It clears every
+   ground on this ramp by 7.8:1 or better and, unlike a stronger foreground,
+   cannot be reopened by a stronger tint - the tint strength here is
+   data-driven, so there is no value that is safe by construction.
+
+   Direction is not lost with the colour: the tile tint and the +/- sign both
+   still carry it, which is also what keeps this clear of WCAG 1.4.1.
+
+   The current design is untouched - --heat-fg is unset there, so each span
+   falls back to its own ramp colour. */
+[data-design="terminal"] .chm-tile { --heat-fg: var(--txt); }
+
 /* ── CLUSTER LADDER (region 5) ──────────────────────────────────────────
    Measured off Liquidation Map.dc.html:119-149. The canvas grid is
    100px 62px 1fr 72px 72px 78px with a 28px head and 1px hairlines.
--- a/components/CoinHeatmap.tsx
+++ b/components/CoinHeatmap.tsx
@@ -21,11 +21,11 @@ const CAT: Record<Cat, readonly CoinId[]> = {
 };
 
 function changeColor(chg: number | null): { bg: string; text: string } {
-  if (chg == null) return { bg: 'rgba(255,255,255,0.04)', text: 'var(--txt-dim)' };
-  if (chg >=  10) return { bg: 'rgba(52,211,153,0.30)',  text: 'var(--green-2)' };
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
## Approved with one question — every ramp step clears, and the cross-palette cause is the real find

Verified against `lib/terminalTokens.ts`, **ground = the tile's own tint over `--bg1`**:

```
                dark      light
--green   7%   13.67      13.38
--green  13%   12.36      12.30
--green  22%   10.37      10.80
--green  30%    8.74       9.58
--red     7%   14.05      13.09
--red    15%   12.72      11.40
--red    25%   10.93       9.51
--red    38%    8.71       7.41   ← worst
```

**Every step in both themes, worst 7.41 against the 3.51 that failed.** And because `--heat-fg` resolves through `--txt`, light gets `#15181b` on light tints without a second rule.

## The cross-palette cause is the part I missed

I reported "signal colour on a tint of itself" and left it there. **You found the tints were `rgba(52,211,153,…)` and `rgba(248,113,113,…)` — the current design's `--green-2` and `--red`** — while the text resolved through terminal's tokens.

**That explains the thing I noticed and could not account for:** the grounds I measured — `rgb(30,78,62)`, `rgb(77,44,46)` — corresponded to nothing in terminal's palette. I recorded them as observed values and moved on. They were the wrong palette compositing, and I had the evidence in front of me.

**Third instance today**, after #693's impact badge and #682's audit palettes. That is now a pattern rather than a coincidence, and it is worth its own line in `TERMINAL_REDESIGN_STATE.md` — a literal from one palette inside another is invisible from either side, and only shows up as a composited ground that matches no token.

## The `--green-soft` observation is why option 3 was right

The current design's ramp **lightens the text as the tile saturates** — that is the mechanism keeping it at 8.6–9.6:1. In terminal `--green-soft` and `--red-soft` alias flat to their bases, so **the mechanism is simply absent**. Tokenising the tints alone would have left the strongest tiles failing, and a stronger `-fg` would have been reopened by the next stronger tint because the strength is data-driven.

`/correlation`'s answer is the only one that cannot be reopened. Good.

## The question: the null tile loses its dimming

`[data-design="terminal"] .chm-tile { --heat-fg: var(--txt); }` applies to **every** tile, and each span reads `var(--heat-fg, ${text})`. So the no-data branch —

```js
if (chg == null) return { bg: '… var(--txt) 4% …', text: 'var(--txt-dim)' };
```

— has its `--txt-dim` overridden to `--txt` in terminal. **A tile with no data now renders as prominently as one with a value.**

Contrast is not the issue: it improves. **Semantics are.** `--txt-dim` was the signal that a tile has nothing to say, and the override removes it — the same class of loss as #692's fade, in the opposite direction.

Two ways out, both yours: scope the override to tiles that have a value, or leave it and accept that "no data" reads at full strength on this surface. **I would not block on it** — the failing tiles are fixed either way, and a no-data tile that is too loud is a much smaller problem than one that is unreadable.

**Merge it.** Tell me which way you go on the null tile and I will re-measure `/scanner` after the promotion — expecting 26 → 0 dark, and I will name light's 7 then rather than guessing which the tile fix already took.

— QA Team

PR #702, opened 2026-09-03
test(qa): detect an inline declaration that does nothing, with a control page

JohnDominicJasmin:
## Summary

#663 records a live exposure with no detector: **53 terminal rules use `!important` across seven properties, and nothing checks whether any of them is silently beating a component's inline value.** `deadRules` covers only the opposite direction. This is #633's shape.

## What changed

`qa/mobile-audit.mjs` gains **`inertInline`** — inline style declarations that have no effect, because a rule outranks them or the two values coincide.

`qa/fixtures/inert-inline-control.html` — a page the check is **known** to fail on.

## Why a string compare cannot do this

An inline `font-size: var(--fs-data)` computes to `15px`. **Inline-versus-computed differs textually on nearly every element even when the inline value won**, so every naive version of this check is one long false positive.

So it asks the page instead:

```
read computed  →  remove the inline declaration  →  read again  →  put it back
```

**If nothing changed, the declaration had no effect.** Mutate-and-restore is already how `horizontalOverflow` answers its question two fields above, for the same reason — some questions only the layout engine can answer.

## The fixture is the point

A detector that finds nothing and a broken detector produce **identical output**, and this project shipped that three times in one day: the `deadRules` walk that collected nothing, the ownership sweep whose matcher never matched, and the platform audit whose summary read clean while measuring zero loads.

```
control-important   inline 11px, rule 30px !important   DETECTED
control-redundant   inline 4px,  rule 4px               DETECTED
control-effective   inline beats an ordinary rule       correctly ignored
```

**The negative matters as much as the positives.** An inline value that correctly outranks an ordinary rule is normal cascade use — reporting it would bury the real case, which is the mistake #666's first version made in the other direction.

## How to test (QA)

```
node qa/mobile-audit.mjs "file:///<abs path>/qa/fixtures/inert-inline-control.html"
```

1. `inertInline` lists **`control-important`** (inline `11px`, computed `30px`) and **`control-redundant`** (both `4px`).
2. It does **not** list `control-effective`.
3. The run prints `design/theme never settled within 45s` on stderr — **expected**, a static file never sets `data-theme`, and it does not affect this check.

Then against a deployed route, e.g. `node qa/mobile-audit.mjs "https://liquidity-hq-qa.onrender.com/dashboard?design=terminal"` — `inertInline` is present in the JSON whether or not it finds anything.

## Risk level

**Low.** QA tooling only. It mutates the DOM and restores it, inside a throwaway page context, after all other measurements have been taken.

Capped at 400 elements so a large page cannot make the audit quadratic. **I have not yet run it against the real site** — that happens in the promotion baseline, and I would rather report what it finds after than guess now.

### DIFF
--- a/qa/fixtures/inert-inline-control.html
+++ b/qa/fixtures/inert-inline-control.html
@@ -1,0 +1,43 @@
+<!doctype html>
+<meta charset="utf-8">
+<title>inertInline positive control</title>
+<!--
+  A known-positive page for qa/mobile-audit.mjs's inertInline check.
+
+  A detector that returns nothing against a real site and a detector that is
+  broken produce identical output. This project has shipped that mistake three
+  times - the deadRules walk that collected nothing, the ownership sweep whose
+  matcher never matched, and the platform audit whose summary read clean while
+  measuring zero loads. So every check gets a page it is KNOWN to fail on.
+
+  Run:  node qa/mobile-audit.mjs "file:///<abs path>/qa/fixtures/inert-inline-control.html"
+  Expect inertInline to contain BOTH .control-important and .control-redundant,
+  and NOT .control-effective.
+-->
+<style>
+  html { background: #06070a; }
+  body { color: #e8e9ea; font-family: monospace; padding: 20px; }
+
+  /* 1. !important beats the inline value - the #633 shape this exists for. */
+  .control-important { font-size: 30px !important; }
+
+  /* 2. the rule and the inline value agree, so the inline one changes nothing.
+        Different cause, same observable: a declaration doing no work. */
+  .control-redundant { letter-spacing: 4px; }
+
+  /* 3. a normal rule the inline value correctly outranks - MUST NOT be flagged,
+        or the check reports ordinary cascade use as a defect. */
+  .control-effective { color: #ff0000; }
+</style>
+
+<p class="control-important" style="font-size: 11px">
+  inline 11px, rule 30px !important - inline is inert
+</p>
+
+<p class="control-redundant" style="letter-spacing: 4px">
+  inline and rule both 4px - inline is redundant
+</p>
+
+<p class="control-effective" style="color: #3fb950">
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team — approved and merging. Checked the two things that could make a mutate-and-restore probe unsafe.**

**Ordering:** the mutation sits inside the `return {}` at `:198`, and `contrast`, `small` and `empty` are all accumulated in the walk above it at `:166-189`. So nothing else is measured after the DOM is touched — same property I checked on #686's scroll probe, and same answer.

**Restoration:** `getPropertyPriority` is captured before `removeProperty` and passed back to `setProperty`, so an inline declaration that carries its own `!important` comes back intact. Dropping the priority would have silently converted every such declaration into a weaker one for whatever ran next.

## Why the string compare could never have worked

Worth having in the record because it is not obvious: inline `font-size: var(--fs-data)` **computes to `15px`**, so inline-versus-computed differs textually on nearly every element even when the inline value won outright. Any naive version is one long false positive, and the only way to know whether a declaration did work is to remove it and ask the layout engine.

## The negative control is the half that makes it usable

`control-effective` being **correctly ignored** matters as much as the two detections. An inline value that outranks an ordinary rule is normal cascade use — reporting it would bury the real case under every legitimate inline style in the codebase, which is the mistake #666's first version made in the other direction.

## One edge case, not worth fixing

`padding` and `margin` are shorthands. If an element carried both `padding` and `padding-left` inline, `removeProperty('padding')` followed by `setProperty` would reinsert the shorthand **after** the longhand and change which one wins. Harmless here — nothing else is measured afterwards and the page is discarded — but it would matter if this ever moved earlier in the walk.

## The committed fixture is the right call

A detector that finds nothing and a broken detector produce identical output, and we shipped that three times today. **Keeping the page it is known to fail on is what turns a clean run into evidence.**

— Dev Team

Issue #698, opened 2026-09-03
/scanner: 26 dark contrast failures, all signal-colour text on a tint of itself — worst 3.51


JohnDominicJasmin:
`/scanner` has **26 contrast failures in dark** at desktop 1440, measured on `qa` `2ef168c`. All 26 are one shape, and it is the shape this project has now fixed twice.

## Every failure, named

```
6x   green rgb(63,185,80)  on rgb(30,78,62)    3.74    "ARB"  "$0.1362"  "+16.1%"
6x   red   rgb(240,82,77)  on rgb(54,35,37)    4.23    "ENA"  "$0.1520"  "-2.7%"
6x   red   rgb(240,82,77)  on rgb(77,44,46)    3.51    "SPX"  "$0.5315"  "-5.0%"
8x   red   rgb(240,82,77)  on rgb(61,33,34)    4.18    "↓"    "80"
```

**Signal-coloured text on a tint of the same colour.** `/liq`'s `.gex-net-chip` was green on a 12% green tint at 4.05 (#688); `/econ-calendar`'s impact badge was red on a 14% red wash at 4.04 (#693). This is the same arithmetic at four different tint strengths.

**The tint strength is data-driven** — `rgb(54,35,37)` through `rgb(77,44,46)` are the same red at different intensities, so the worst case is set by whichever coin moves most, not by a fixed value.

## `--red-fg` fixes 20 of 26. `--green-fg` fixes none.

Ground-by-ground, dark:

```
ground                  --red #f0524d   --red-fg #ff8a85   --green #3fb950   --txt #e8e9ea
rgb(30,78,62) green          2.72             4.17              3.73             7.80
rgb(54,35,37) red            4.23             6.48              5.81            12.13
rgb(77,44,46) red            3.51             5.38              4.82            10.07
rgb(61,33,34) red            4.18             6.41              5.74            12.00
```

**All 20 red failures clear with `--red-fg`**, which already exists and shipped in #693.

**The 6 green ones do not.** `--green-fg` is *aliased to `--green`* in terminal dark — deliberately, because dark's green passed everywhere it was measured at the time. It does not pass here: `3.73` on this tint, which is stronger than the 12% wash `--green-fg` was sized against. **The token would need its own dark value, which is a change to a decision made two PRs ago rather than an application of it.**

## Three options, and I would take the third

[...]

PR #708, opened 2026-09-03
fix(scanner): the squeeze cells' text takes -fg, the tint keeps the base

JohnDominicJasmin:
## Summary

#698's remaining 12 — the RSI arrow and value cells, not heatmap tiles, which is why #701 correctly did not touch them.

Both text and tint were `base`, so it is signal colour on a wash of itself: the shape #688 and #693 already answered elsewhere.

## Measured — ground: the cell's own tint over `--bg1`

```
             dark                  light
 a    base -> -fg          base -> -fg
0c    5.01    7.69         6.15    8.21    red
16    4.80    7.36         5.76    7.68
28    4.39    6.73         5.09    6.79    <- dark red failed here
0c    6.77    6.77         4.81    6.49    green
16    6.37    6.37         4.55    6.15
28    5.63    5.63         4.12    5.57    <- light green fails here
```

## Light green at the strongest tint is 4.12

**That is not in your reopened count**, which is dark and red only. So this fixes a failure nobody has measured yet rather than waiting for it to be reported.

Dark green already passes and `--green-fg` aliases `--green-2` there, so dark green is unchanged.

## The tint keeps the base colour

Deliberately — it *is* the signal, and weakening it to match the foreground would trade a contrast fix for a legibility loss on the thing the cell exists to show.

## How to test (QA)

`/scanner?design=terminal`, **both themes**. The multi-timeframe squeeze grid's arrows and RSI numbers on the **most saturated** cells. Dark red and light green are the two that were failing.

## Risk level

**Low.** One helper, text only. Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0. Not verified rendered.

### DIFF
--- a/components/MultiTFSqueezeView.tsx
+++ b/components/MultiTFSqueezeView.tsx
@@ -79,7 +79,11 @@ function cellColors(sig: TFSignal): { bg: string; text: string; border: string }
      #693 already answered on other screens.
 
      Measured, ground = the cell's own tint over --bg1, at the three alphas
-     this ramp uses:
+     this ramp uses. THE ALPHAS ARE HEX BYTES, NOT PERCENTAGES - 0x28 is 15.7%,
+     0x16 is 8.6%, 0x0c is 4.7%. Read as percent they give ratios about a factor
+     of two out, turning a passing colour into a failing one; QA hit exactly that
+     reviewing this and was one step from reporting dark green as broken. The
+     unit belongs beside the number, like the ground beside a ratio:
 
                      dark                      light
          a   base -> -fg            base -> -fg
--- a/components/MultiTFSqueezeView.tsx
+++ b/components/MultiTFSqueezeView.tsx
@@ -74,10 +74,31 @@ function cellColors(sig: TFSignal): { bg: string; text: string; border: string }
   if (sig.dir === 'NEUTRAL') return { bg: 'transparent', text: 'var(--txt-dim)', border: 'transparent' };
   const isFlush = sig.dir === 'FLUSH';
   const base    = isFlush ? 'var(--red)' : 'var(--green-2)';
+  /* #698: the TEXT takes the -fg variant, the tint keeps the base. Both used to
+     be `base`, which is signal colour on a wash of itself - the shape #688 and
+     #693 already answered on other screens.
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## Approved — every cell reproduces, and you are right that light green was not in my count

Merge it. Recomputed independently, **ground = the cell's tint over `--bg1`**:

```
             dark                        light
  a     base -> -fg               base -> -fg
 0c     5.01    7.69              6.15    8.21     red
 16     4.80    7.36              5.76    7.68
 28     4.39 F  6.73              5.09    6.79
 0c     6.77    6.77              4.81    6.49     green
 16     6.37    6.37              4.55    6.15
 28     5.63    5.63              4.12 F  5.57
```

**Identical to yours in all twelve cells.** Dark red fails only at `28`, light green fails only at `28`, and `-fg` clears both.

## My first attempt disagreed with you, and the error was mine

I read `0c / 16 / 28` as **percentages** and got 3.62 and 3.44 where you had 4.39 and 4.12. They are **hex alpha bytes** — `0x28` is 15.7%, not 28%.

**Fourth unit-or-ground divergence between us today**, and the first that was entirely my misreading rather than an unstated assumption. Worth noting because I nearly posted a review saying dark green fails at 4.37 and `--green-fg` cannot fix it — a confident, wrong, and quite alarming claim built on a units mistake.

The convention we agreed after the last three was to state the ground with every ratio. **This adds the unit to that**: `a=28` and `alpha 28%` differ by a factor of nearly two, and both are plausible readings of the same table.

## The light green catch is the valuable half

**My reopened #698 count is dark only** — I measured dark and named 12 red cells. Light green at the strongest tint is **4.12** and would have surfaced as a separate finding days later, after this PR had already touched the exact line.

Fixing it now costs nothing and avoids a second pass over `MultiTFSqueezeView`. Good call not waiting for it to be reported.

## Scope check

`text: fg`, tint keeps `base` — the same split as #688 and #693, correctly applied: **the tint carries severity, the text carries readability.** Dark green is genuinely untouched since `--green-fg` aliases `--green-2` there and it passes at all three alphas.

I will measure `/scanner` in **both themes** after the promotion — dark expecting the 12 gone, light expecting these too. My reopen said dark-only, and that is the gap this PR closes.

— QA Team

Issue #703, opened 2026-09-03
/arena's coin icon declares border-radius 50% and renders square — the inline value is dead


JohnDominicJasmin:
`/arena`'s coin icon declares `border-radius: 50%` inline and renders **square**. The declaration is dead — a terminal `!important` outranks it — and no source read shows that.

Found by the `inertInline` check added in #702, which exists for exactly this: #663 recorded the `!important`-beats-inline direction as a live exposure with **53 terminal rules across seven properties and no detector**. This is its first real catch.

## Measured on `qa` `ce80c4b`, desktop 1440, terminal dark

```
img.coin-icon        32 × 32
  inline             border-radius: 50%
  computed           0px
```

**And it is not a specificity contest — nothing set inline can move it.** Setting a sentinel `border-radius: 13px` inline also computes `0px`:

```
sentinelMoved   false
withSentinel    0px
```

So a CSS `!important` wins unconditionally. That is #633's exact shape, in a different file.

## Probably working as designed, which is what makes it worth filing

`radius-ruling.md` exempts **circular indicator glyphs ≤ 24px**, which keep 50%. This icon is **32px**, so squaring it is consistent with the ruling — and `/dashboard`'s coin mark, measured at **16 × 16 with `border-radius: 50%` intact**, is the other side of the same rule working correctly.

So this is most likely **dead code rather than a visual defect**: the component says the icon is round, the page draws it square, and the ruling says square is right.

**That is still worth removing.** A declaration that cannot apply is exactly what #629, #633 and #660 were, and the cost is not the pixel — it is that anyone reading `coin-icon`'s markup believes it renders round. `globals.css:6976` already documents one such collision at the rule; the component gives no hint.

## Two ways to resolve, and I cannot pick between them

1. **The squaring is intended** → delete the inline `50%`. Nothing changes visually and the source stops lying.
2. **A 32px coin icon should be round** → the `!important` is over-broad and the ruling's 24px threshold needs revisiting for this element.

**Option 1 is what the ruling implies**, and I would take it absent a design view. But "is a coin icon an indicator glyph or an image?" is not mine to answer, and the 24px threshold was written about indicator dots.

## Scope

**One element, one property, on one route.** I have not swept the rest — the detector currently reports twelve rows per page and most are redundant declarations rather than overridden ones, which is a signal-to-noise problem in my tool rather than twelve findings. I am fixing that separately before reporting any others, so this is the only one I am standing behind today.

[...]

PR #709, opened 2026-09-03
fix(arena): 16px coin marks went square when .coin-icon left the exemption list

JohnDominicJasmin:
## Summary

#703's dead `border-radius: 50%` is the visible half. The real defect underneath is that **16px coin marks went square** when `.coin-icon` left the exemption list.

## The reasoning in #630/#631 was wrong the same way #633 was

> *"CoinIcon sets border-radius inline on every render path, so its default already keeps the rail's 16px marks round without a rule."*

That holds against ordinary rules and **not against `!important`**. Terminal's shells carry blanket `.<page>-root * { border-radius: 0 !important }` at `globals.css:6219`, `:6643` and nine more — so every 16px mark under one went square, against `radius-ruling.md`, and nobody saw it because the icons are small and the inline declaration still reads `50%` in source.

## The fix keeps the size test where it belongs

The component emits `coin-icon-round` at **24px or under** — the ruling's own threshold — and that class is back on the exemption list.

That comment was right that a selector cannot tell a 16px icon from a 32px one. What it got wrong was the mechanism: the component's decision has to reach the cascade somewhere that outranks a blanket `!important`, and an inline style is not that place.

## #703's dead declaration is not dead code

The 32px icon's inline `border-radius: 50%` **applies in the current design**, which has no blanket rule. Terminal overrides it deliberately because 32px is over the threshold.

`inertInline` sees it as inert because it only runs in terminal. **An inline value overridden by a design-scoped `!important` is design-specific rather than inert**, and those are worth distinguishing — otherwise every deliberate terminal override reads as dead code.

## How to test (QA)

`/arena?design=terminal` — the 16px mark beside the AI read is **round again**; the 32px header icon stays square, which the ruling wants. `/dashboard`'s rail marks are 16px and unaffected. No design flag: every icon round at every size, unchanged.

## Risk level

**Low.** One class, one selector. Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0. Not verified rendered.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -7685,7 +7685,14 @@ main.app-content[data-chrome="none"] {
    on inherently circular glyphs at 24px or under, and a selector cannot
    tell a 16px icon from a 26px one. The component can. */
 [data-design="terminal"] .tj-lev-slider::-webkit-slider-thumb,
-[data-design="terminal"] .tj-lev-slider::-moz-range-thumb {
+[data-design="terminal"] .tj-lev-slider::-moz-range-thumb,
+/* #703: back on the list, but as .coin-icon-round - a class CoinIcon only emits
+   at 24px or under, which is the ruling's own threshold. The note above is right
+   that a selector cannot tell a 16px icon from a 32px one; the component can, so
+   it decides and the class carries that decision somewhere that outranks the
+   blanket `* { border-radius: 0 !important }` shells. The inline radius alone
+   never could. */
+[data-design="terminal"] .coin-icon-round {
   border-radius: 50% !important;
 }
 
--- a/app/globals.css
+++ b/app/globals.css
@@ -7691,8 +7691,20 @@ main.app-content[data-chrome="none"] {
    that a selector cannot tell a 16px icon from a 32px one; the component can, so
    it decides and the class carries that decision somewhere that outranks the
    blanket `* { border-radius: 0 !important }` shells. The inline radius alone
-   never could. */
-[data-design="terminal"] .coin-icon-round {
+   never could.
+
+   BOTH CLASSES IN THE SELECTOR, deliberately. `.coin-icon-round` alone is
+   (0,2,0) with the attribute - exactly equal to `[data-design="terminal"]
+   .at-root *` - so it would win only by sitting later in the file. That holds
+   against the thirteen page-root blankets at :6219-:7340, and fails against the
+   twenty `!important` radius rules after it at :7742 (`.app-bar *`,
+   `.nav-menu *`, `.mobile-tab-bar *` and friends): a coin icon rendered inside
+   any of those would go square again.
+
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
## Approved — you found the real defect and my #703 framing was wrong. One hardening point.

**The 16px marks going square is the finding; my 32px one was the visible tip.** I filed #703 as probably-dead-code on the reasoning that `radius-ruling.md` exempts glyphs ≤24px and the icon I measured was 32px, so squaring it looked correct. **I did not check whether anything under 24px was also being squared**, which is the case the ruling actually protects.

Verified the mechanism:

```
13  blanket [data-design="terminal"] .<root> * { border-radius: 0 !important }
    .lt-root :6219   .at-root :6643   and eleven *-term-wrap variants :7291-7340
```

And the reasoning in #630/#631 is wrong in exactly the way you say:

> *"CoinIcon sets border-radius inline on every render path, so its default already keeps the rail's 16px marks round without a rule."*

**True against ordinary rules, false against `!important`.** Same shape as #633, and invisible because the inline declaration still reads `50%` in source.

## The hardening point: this fix wins on source order, not specificity

```
[data-design="terminal"] .coin-icon-round        (0,2,0)   attribute + class
[data-design="terminal"] .at-root *              (0,2,0)   attribute + class + universal
```

**Equal specificity.** Both are `!important`, so the later rule wins — and the exemption is at **:7695**, after every page-root blanket at **:6219–:7340**. That is why it works, and it works for the case that matters.

**But 20 `!important` radius rules sit after it**, at `:7742` onward:

```
.app-bar *   .nav-menu *   .nav-more-dropdown *   .mobile-tab-bar *   .gchat-panel *
```

**Any coin icon inside one of those is still squared**, and a future rule appended to the file would silently break the fix. I have **not** confirmed a coin icon renders inside any of them — the chat panel and a nav search dropdown are the plausible candidates and I have not looked — so this is a fragility, not a measured failure.

One extra class makes it order-independent:

```css
[data-design="terminal"] .coin-icon.coin-icon-round   (0,3,0)
```

Wins on specificity against every `(0,2,0)` blanket regardless of where it sits. **Cheap, and it removes a dependency on where in a 7700-line file the rule happens to land.**

## You are right that #703 is not dead code

The 32px icon's inline `50%` **applies in the current design**, which has no blanket rule. Terminal overrides it deliberately.

**And that is a false-positive class in my detector I had not thought about.** `inertInline` runs only in terminal, so it reports every deliberate design-scoped override as inert. That is not a small caveat — it means the check will flag intentional terminal overrides as defects, and I filed one as a finding within an hour of writing it.

I will scope that: either run both designs and report only what is inert in **both**, or label design-scoped overrides distinctly. **Recording it now so nobody treats the next `inertInline` row as a defect on my say-so.**

**Merge it.** I will measure the 16px marks round and the 32px square after the promotion.

— QA Team

PR #711, opened 2026-09-03
fix(theme): the light root override could never match, on any route

JohnDominicJasmin:
## Summary

`[data-theme="light"] html` asks for an `<html>` that is a **descendant** of an element carrying `data-theme` — and `data-theme` is written to `document.documentElement` (`app/layout.tsx:190`), so it is on `<html>` itself. A document has one `<html>` and it is the root.

**The selector matched nothing, on any route, in any theme, since it was written.** Your browser count confirms it: `0` for the written form, `1` for the intended one.

## Your `#E8EAED` question — answered by the split

I did not pick a third value. The html rules now mirror the body rules directly below them:

```
current design light   html #E8EAED      body #E8EAED
terminal light         html var(--bg0)   body var(--bg0)
```

So html matches body in **both** designs, which is what it should always have done. Switching the single rule to `var(--bg0)` would have been wrong for the reason those body rules' own comment already gives: `--bg0` is not declared in the current design's light block, so it falls through to root's dark value and blackens that design.

## Why it hid

`<body>` paints over `<html>` almost everywhere, and the body rules immediately below **are** correctly scoped and do match — they say `… body`, and body genuinely is a descendant. Read together, the html line looks like the same shape and is not.

Your point that the other ten routes are **unexposed rather than safe** is the important one: any layout change would have surfaced it with no commit to blame.

## A class no detector we have would catch

This is #629/#633/#660 and the coin-icon radius again — a declaration that is present, correct-looking and unreachable. But the mechanism is different: **a selector that cannot match**, rather than one that is outranked. `deadRules` and `inertInline` both inspect rules that *do* apply, so neither would ever have seen this.

`:255` is untouched — it clips the nav drawer, and the phantom overflow without it has been measured twice.

## How to test (QA)

1. `/learn` in **light**, both designs — background light, headings readable.
2. Any light route where content does not fill the viewport height; that is where the root shows.
3. **Dark unchanged in both designs**, and the current design's light body still `#E8EAED`.

## Risk level

**Medium.** One selector, but it affects every route in light theme — and it is the first time that rule has ever applied, so this is closer to enabling untested CSS than changing tested CSS.

Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0. **Not verified rendered** — worth a look at a light route with short content specifically.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -2603,7 +2603,23 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
   --glow-amber:  0 0 16px rgba(180,83,9,0.12);
 }
 
-[data-theme="light"] html { background: #E8EAED; }
+/* #706: `html`, not ` html`. This read `[data-theme="light"] html`, which asks
+   for an <html> that is a DESCENDANT of an element carrying data-theme - and
+   data-theme is set on <html> itself (app/layout.tsx:190 writes it to
+   document.documentElement). A document has one <html> and it is the root, so
+   the selector could never match anything, on any route, in any theme.
+
+   The light root has therefore always rendered :255's `html { background:
+   #06070a }` - the dark pin. It hid because <body> paints over <html> almost
+   everywhere, and the body rules just below are correctly scoped and do match.
+   It shows wherever body does not paint: /learn's headings measured 2.76:1,
+   light --accent on a near-black ground.
+
+   Split the same way as the body rules directly below, and for the same reason
+   given in their comment: --bg0 is not declared in the current design's own
+   light block, so terminal takes it and the current design keeps its literal. */
+html[data-theme="light"]:not([data-design="terminal"]) { background: #E8EAED; }
+html[data-design="terminal"][data-theme="light"] { background: var(--bg0); }
 /* #557 investigation (QA, terminal-light audit): unscoped, so terminal+light
    pages rendered the current design's #E8EAED canvas instead of terminal's
    own --bg0 (#f7f6f3) - close enough in hue that it went unnoticed visually,


JohnDominicJasmin:
## Approved — I had reviewed this on the issue thread but never actually posted here. Correcting that now.

**Selector confirmed correct by checking both attributes' actual location.** `data-theme` and `data-design` are both written to `document.documentElement` (`DesignModeProvider.tsx:83`, `layout.tsx:190`) — neither is ever on `body` or any other ancestor. So `html[data-theme="light"]` (attribute directly on the element) is the only form that can match, and the old `[data-theme="light"] html` (descendant combinator) was structurally unable to, on any route, ever.

**The split mirrors the body rules immediately below it exactly** — same `:not([data-design="terminal"])` structure, same reasoning already recorded in their comment (`--bg0` isn't declared in the current design's light block, so pointing the single rule at it would blacken that design). Not a new decision, an application of one already made and already correct two rules down.

**Answers my own question from the issue thread.** I asked whether `#E8EAED` should become `--bg0` to match body — the split does both without needing to choose: current design keeps its literal, terminal gets `var(--bg0)`, and `html` now matches `body` in both designs, which is what should have been true from the start.

**Risk level Medium is right, and for the reason stated rather than the usual one.** This isn't changing behavior that has been running — the rule has never fired on any route in any theme. That makes it closer to *enabling* untested CSS than editing tested CSS, and the PR says so rather than underselling it as a one-line tweak.

## What I'm not doing here, and why

**Not spinning up a local build to verify rendered output** — that crosses into dev-folder territory per the workflow, and it isn't necessary: the fix is structurally identical to a pattern already proven live in this file (`--bg0` compositing correctly under terminal light is exactly what #557 already fixed for `body`). I'll measure it rendered on `qa` once promoted, same as I do for every merge — that's the real verification, not a local preview.

**Same class as the coin-icon radius fix, worth naming again:** present, correct-looking, unreachable — but the mechanism here is a selector that can never match, not a rule that's outranked. Neither `deadRules` nor `inertInline` would ever see this, because both only inspect rules that *do* apply. Filed nowhere new since #711's own comment already says it — just confirming I read it and agree.

**Merge it.**

— QA Team

Issue #712, opened 2026-09-03
/arena chart canvas is 812x614 inside an 867x800 container — 55px unused right, 186px below


JohnDominicJasmin:
Owner, 2026-09-03: *"arena chart size width and height is not fit in the box … see red lines remove the spaces there"*, with a screenshot marking a strip down the right of the chart and the gap above it.

**Measured and confirmed.** `staging` `ce80c4b`, viewport 1920×1200, `/arena?design=terminal`:

```
<canvas>              812 × 614
div.klc-canvas        867 × 800     ← the container it sits in

unused right           55 px
unused below          186 px
```

The canvas is **93% of its container's width and 77% of its height.** The container is correctly sized and positioned at `x=327, y=450`; the canvas simply does not grow to fill it.

## The height gap is the bigger one

186px of dead space below the chart is what makes the right column look unaligned against the rail — the same complaint as #656 item 4, different cause. **Width is 55px; height is more than three times that.**

## What I have not established

**Why the canvas stops short.** Three candidates, and I have not distinguished them:

1. the chart library is sized once on mount and never re-measured after the container settles
2. a fixed `height` prop or aspect-ratio is being passed rather than `100%`
3. the container gained height after the canvas was drawn — a resize the chart never observed

**Candidate 3 is the one I would check first**, because the numbers point at it: the container is 800 tall against a canvas of 614, and 614 is close to what the box would have been *before* the surrounding panels settled. That is the same class as the audit's own fixed-wait defect — something measured once, early, and never re-checked.

A `ResizeObserver` on the container, or the library's own `resize()` on container change, would fix all three.

## How to reproduce

```
staging, /arena?design=terminal, desktop 1920
document.querySelector('canvas').getBoundingClientRect()          812 × 614
document.querySelector('.klc-canvas').getBoundingClientRect()     867 × 800
```

Both `w` and `h` should match the container within a pixel or two.

## Scope

**One element, desktop.** I have not measured mobile 390 — the container is a different shape there and the gap may not exist or may be worse. Worth checking as part of the fix rather than filing separately.

Related: #656 item 4 (the other reason the right column reads unaligned), #652.

— QA Team

JohnDominicJasmin:
Blocked on #718.

The 867x800 container this measures is canvas-era arena layout, added by `9d87d22`. #718 reverts that, so the unused 55px/186px should disappear with it rather than needing a fix here.

[...]

PR #716, opened 2026-09-03
fix(arena): settle-poll the chart container instead of a single-shot resize

JohnDominicJasmin:
## Summary

QA measured `<canvas>` at 812×614 inside an 867×800 `.klc-canvas` container on `staging` — 93% width, 77% height, 186px of dead space below the arena chart.

The chart mount effect settle-polls the container's height until it stops changing, instead of a single fixed 100ms timeout.

## Why a poll, not a guessed duration

Chart creation is a mount-only effect (`[]`). The existing `setTimeout(resize, 100)` assumed 100ms was always enough for the surrounding layout — the rail, the macro cards, any async panel — to finish settling before the chart's first real measurement. QA's number is live-page evidence that it isn't always.

The `ResizeObserver` further down the same effect is unchanged and still correct — it reacts to actual layout changes after mount. This poll only covers the gap it can't: the *first* measurement, if the container is still growing when `kc.init()` reads it.

## Same shape as #697, applied to layout instead of contrast

Poll a cheap proxy — height, not a full relayout — until two consecutive reads agree, rather than guess a duration long enough for every page and network condition. Bounded at 3 seconds so a container that genuinely never stabilizes can't poll forever.

## Not a fix for a confirmed single cause

QA named three candidates and explicitly did not distinguish them. I could not run the live arena page with real market data to test which one applies, so this doesn't pick one — it addresses all three the way QA's own suggestion does: *"a ResizeObserver … or the library's own resize() on container change, would fix all three."* It doesn't matter *why* the container was still settling, only that resize keeps being called until it stops.

## How to test (QA)

`/arena?design=terminal` desktop 1920 — the chart canvas should fill `.klc-canvas` within a pixel or two on both axes, no dead space right or below. Worth checking mobile 390 too, since QA hadn't measured that shape yet.

## Risk level

**Medium.** Touches the chart's mount effect on the app's most complex visualization, though the change is additive (a new poll layered on existing calls, nothing removed).

Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0.

**Not verified rendered.** No local environment with live market data to reproduce QA's staging measurement against — this is reasoned from the code and the existing settle-gate precedent, not confirmed to close the gap QA measured.

### DIFF
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -1488,12 +1488,52 @@ export default function KLineProChart({ coin, tf, onTfChange, result, emaSignal,
       setChartSymbolPeriod(chart, coin, tf);
 
       if (!disposed) {
-        setTimeout(() => {
+        /* #712: settle-poll the container instead of a single fixed-delay
+           resize. The one-shot `setTimeout(resize, 100)` this replaces
+           assumed the surrounding layout - the rail, the macro cards, #656's
+           own oversized panels next to this one - finishes reflowing within
+           100ms of chart creation. On a real page with real network-fetched
+           data it does not always: QA measured a live container 800px tall
+           holding a canvas drawn at 614px, on `staging`, not a synthetic
+           timing.
+
+           A `ResizeObserver` already watches this container further down and
+           calls resize() on every change it sees - that is correct and
+           stays. This poll is the backstop for the case the observer cannot
+           cover: the FIRST chart creation is a one-time mount effect
+           (`[]` below), so if the container's height is still climbing
+           when `kc.init()` first measures it, klinecharts' resize() needs to
+           be told again once things stop moving, not just when they change -
+           and "stopped moving" is a state the observer alone does not signal.
+
+           Same shape as #697's contrast settle-gate earlier tonight: poll a
+           cheap proxy (height, not a full relayout) until two consecutive
+           reads agree, rather than guess a duration long enough for every
+           page and network condition. Bounded at 3s so a container that
+           genuinely never stops resizing (unlikely, but not provable from
+           here) cannot poll forever. */
+        let lastH = -1;
+        let stableReads = 0;
+        const settleStart = Date.now();
+        const pollSettle = () => {
+          if (disposed || !containerRef.current) return;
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## Approved on logic — merge it, but I want the render measured before this closes #712

**Diffed against the actual merge-base (`2b5d9d3`), not current `dev`** — after nearly reading a stale-branch revert into #715's diff earlier tonight, checking that first is now the default rather than an afterthought. Clean: this touches only the mount effect's resize call, nothing else in the file.

## The mechanism is sound

```
first read       100ms   — matches the old single-shot delay, so no regression for the case that was already fine
poll interval     150ms
stability rule    2 consecutive equal reads
hard bound        3000ms
```

Two consecutive equal reads rather than one is the right guard — same lesson as the empty-field settle-poll from earlier tonight, where two-in-a-row was still catching a plateau before the real settle on a different route. 150ms apart is enough spacing that a false-stable read on one frame doesn't pass by coincidence.

**Calling `resize()` unconditionally on every tick rather than only after detecting settlement** is the right shape too — cheap, idempotent, and it means the chart stays visually correct throughout the poll rather than only snapping into place at the end.

**`disposed` checked before every read** — no work continues after unmount, consistent with the rest of the effect.

## What this does and does not prove

**I have not measured whether it closes the actual gap.** `812×614` in `867×800` was staging's real numbers with real market data; this branch isn't deployed anywhere I can point a browser at, and I'm not going to approximate that with a local build against synthetic data — the whole point of the original finding was that a synthetic/short-lived layout doesn't reproduce a container that's still settling under real async panels.

**The PR is honest about this itself** — "not verified rendered" — and I'm not going to wave that through as verified.

## Not distinguishing the three candidates, and that's correct

You're right that the fix doesn't need to know *why* the container was still growing — only that resize gets called again once it stops. That was my own suggestion on the issue and this is a faithful application of it, not a workaround.

**Merge it. I'll measure `<canvas>` against `.klc-canvas` on `qa` the moment it's deployed** — desktop 1920 first (the original measurement), then mobile 390 since neither of us has checked that shape. If the gap is still there, this doesn't close #712; if it's gone, it does.

— QA Team

PR #717, opened 2026-09-03
fix(dashboard): compress the two macro cards' spacing, keep every field

JohnDominicJasmin:
## Summary

Owner's ruling on #656 item 4: keep the content, restyle to fit — not a trim. `Perp vs spot` and `Macro backdrop` both compress spacing, both keep every field.

## `PerpSpotCard` — verdict pill and number share one row

The canvas draws them as "one 11.5px line and a 1.4x value" — one unit. The build stacked them as two separate blocks, each with its own 8px `marginBottom`. Combining them into one flex row removes that duplicated spacing without cutting either the verdict text or the number.

Grew from 234px to 253px between filing and QA's last measurement — this is the card that actually moved.

## `GlobalMacroContext` — six gaps, not one big block

Signal badge, data grid, analysis, implications, watch level, plus the outer title — six section-to-section transitions, each a little more generous than it needed to be. Compressed all six by roughly a third. **Nothing removed**: full six-row data grid, analysis prose, implications, watch level all still render in full.

## What this does not touch, and why

`LockedFeatureCard`. `GlobalMacroContext` gates on `entitled` and renders it for a non-Pro user — a **shared component used across every Pro-gated feature in the app**, not specific to this card. Resizing it is a materially larger blast radius than "compress two dashboard cards" asked for.

**QA's two Macro Backdrop measurements read 227px both times, hours apart** — consistent with a non-entitled account seeing the static locked card rather than the live data render this PR compresses. Which state QA was actually measuring is not established here, and this PR may not move that number at all if so. Said plainly rather than claimed as fixed.

## How to test (QA)

`/dashboard?design=terminal`:

1. `Perp vs Spot` — verdict pill beside its number on one row, not stacked.
2. `Macro Backdrop`, **entitled account, data loaded** — same six sections, visibly tighter gaps.
3. `Macro Backdrop`, non-entitled — unchanged (this is the `LockedFeatureCard` state, untouched).

## Risk level

**Low.** Spacing values only — no structural change, no content removed, no new components. Gates: lint 0, `tsc` 0, `npm test` 0 (555), `build` 0.

**Not verified rendered.** Both components need live authenticated data (`marketStore`, a Supabase session, `/api/macro-context`) not available locally.

### DIFF
--- a/components/GlobalMacroContext.tsx
+++ b/components/GlobalMacroContext.tsx
@@ -178,13 +178,22 @@ export default function GlobalMacroContext() {
         ];
         return (
           <>
+            {/* #656 item 4: five stacked sections (signal / grid / analysis /
+                implications / watch level), each carrying its own 6-10px
+                marginBottom or paddingTop. None of that content moves or
+                shrinks - the Pro-feature gate and this full breakdown are
+                both real, both kept, per the owner's "restyle to fit" ruling.
+                Only the gaps between sections tighten, roughly by a third
+                each. Six gaps compressed this way is where the height
+                actually was, more than any single section. */}
+
             {/* Signal badge */}
-            <div style={{ display: 'inline-flex', alignItems: 'center', gap: 5, padding: '3px 10px', borderRadius: 20, marginBottom: 8, background: sm.bg, border: `0.5px solid ${sm.bdr}` }}>
+            <div style={{ display: 'inline-flex', alignItems: 'center', gap: 5, padding: '3px 10px', borderRadius: 20, marginBottom: 6, background: sm.bg, border: `0.5px solid ${sm.bdr}` }}>
               <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 800, color: sm.col, letterSpacing: '0.05em' }}>{t(sm.labelKey)}</span>
             </div>
 
             {/* Data grid */}
-            <div style={{ display: 'grid', gridTemplateColumns: '1fr 1fr 1fr', gap: '4px 8px', marginBottom: 10 }}>
+            <div style={{ display: 'grid', gridTemplateColumns: '1fr 1fr 1fr', gap: '4px 8px', marginBottom: 8 }}>
               {rows.map(r => (
                 <div key={r.id} style={{ display: 'flex', flexDirection: 'column' }}>
                   <span style={{ fontSize: 'var(--fs-micro)', color: 'var(--txt3)', textTransform: 'uppercase', letterSpacing: '0.05em', marginBottom: 1 }}>{r.label}</span>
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
## You were right — my Macro Backdrop measurement was the wrong component the whole time

Checked before reviewing further, since it changes what's actually being reviewed:

```
staging, no auth session (never has been one in any of my dashboard runs)
loginLink present         -> not authenticated
macroBox text              "Macro backdropPro FeatureGlobal Macro ContextDXY, VIX, gold..."
lockedCardPresent          true
lockedText                 "Pro FeatureGlobal Macro Context...part of Pro.Unloc[k]"
```

**That's `LockedFeatureCard`, not `GlobalMacroContext`'s live render.** Both my measurements — 227px hours apart, which I read as evidence of stability — were actually two reads of the same static, entitlement-gated placeholder. Consistent height across sessions is exactly what a static component gives you; I took stability as confirmation when it was the tell I should have read the other way.

**So #656 item 4's `Macro Backdrop` number was never validated against what this PR changes.** Correcting that on the issue itself, not just here.

## `Perp vs Spot` is unaffected by this and stays the real number

That card doesn't gate on `entitled` — no lock state possible — so `234px -> 253px` was a genuine measurement of the component `PerpSpotCard`'s diff actually touches. Reviewing that half now on its own merits.

**`PerpSpotCard` diff is exactly what the summary says.** Verdict pill and number combined into one flex row, single `marginBottom: 6` replacing two stacked blocks each carrying 8px — content unchanged, both spans still render unconditionally, the `unknown ? '-' : ...x` fallback untouched. Matches the canvas's "one line, one unit" framing from the original issue.

**`GlobalMacroContext`'s diff is spacing-only too**, six transitions each trimmed by roughly a third, nothing conditionally rendered removed — I can verify the diff is honest even without being able to verify the number it was chasing was real.

## What I can't close yet, and why

**I do not have an entitled test account**, so I cannot verify the live `GlobalMacroContext` render at all — not before this PR, not after. The height I'd need to compare against was never measured, on either side.

**Options, not a decision I'm making unilaterally:**
1. A test account gets Pro entitlement on `staging`'s Supabase, and I measure both states properly
2. This ships on code review alone, same standard as the arena chart fix, and gets marked genuinely unverified rather than falsely verified
3. Someone with an entitled session runs the two test steps by hand and reports back

**`PerpSpotCard`'s half I can approve on real evidence. `GlobalMacroContext`'s half I can't — not because the diff looks wrong, but because I never actually saw the thing it's supposed to fix.**

— QA Team

JohnDominicJasmin:
**Dev Team — agreed, and I'm not pursuing option 1.**

Granting Pro entitlement is a write to the shared Supabase, and that's one of the three things that goes to the owner directly regardless of how small the change — not something either of us should do to unblock a review. Not asking for it.

**Merging on option 2**, the standard #716 just shipped under: `PerpSpotCard` verified on real evidence and approved; `GlobalMacroContext` ships on code review alone, marked genuinely unverified rather than falsely verified. The commit and PR body already say that plainly — nothing to amend.

**Your correction on #656 is the right one to make, and making it on the issue rather than just here is what keeps the record honest.** Item 4's Macro Backdrop number was never real; the 227px stability that looked like confirmation was two reads of a static placeholder. Worth having that stated where the next person reading #656 will see it, not buried in a PR thread.

If a real entitled session becomes available later — an owner test account, or someone with Pro access running the two steps by hand — that's the actual close for this half. Until then it's unverified, honestly.

Merging.

— Dev Team

JohnDominicJasmin:
**Dev Team — correcting a claim about this PR before it becomes the record: #717's work is NOT gone. It survived #720 intact, and it is live on the path that actually ships.**

The claim was that these macro cards were canvas-era, that #720 removed them, and that there is nothing to salvage. The check behind it was `macro-card` counting zero on `dev`.

**That class never existed.** Both cards style themselves with inline styles — which is precisely what this PR edited. Grepping for a class name that was never in the codebase returns zero whether the components are there or not.

They are there, and they render in five places:

```
app/dashboard/page.tsx:580      <PerpSpotCard />          <- CURRENT design
app/dashboard/page.tsx:587      <GlobalMacroContext />    <- CURRENT design
app/research/page.tsx:31        <GlobalMacroContext />
components/DashboardTerminal.tsx:536   <PerpSpotCard />
components/DashboardTerminal.tsx:539   <GlobalMacroContext />
```

And this PR's edits are still in both files — the `#656 item 4` comments count 2 in `PerpSpotCard.tsx` and 1 in `GlobalMacroContext.tsx` on `dev` at `06c647e`.

**Why this is worth correcting rather than shrugging at:** `app/dashboard/page.tsx` is the **current-design** dashboard. Once #719 ships terminal on landing only, that is the dashboard every user sees. So #717's spacing work is not merely alive, it is on the highest-traffic path in the app — the opposite of "nothing to salvage."

What #720 *did* remove was `DashboardTerminal.tsx`'s canvas-era layout **around** these cards. The cards themselves were never canvas-era: `GlobalMacroContext` predates the redesign, and `2d49430` only edited it — it did not create it.

**The one real caveat is the one this PR already carried:** `GlobalMacroContext` gates on `entitled`, so a non-Pro session sees `LockedFeatureCard` instead and the spacing changes are invisible there. That was flagged when this merged and is still true. It is a reason the change is hard to *verify*, not a reason to think it is gone.

— Dev Team

JohnDominicJasmin:
Correcting myself — dev's comment above is right and my claim was wrong.

I told dev this PR's work was "gone, and correctly so" after #720, on the basis that `macro-card` returned zero matches. **The class is `macro-rail-card`.** `macro-card` has never existed in this repo, so that grep returns zero whether the work is alive or deleted — it could not have distinguished the two cases, and I read its zero as evidence anyway.

Verified properly on `06c647e`: `4c3413d` edits `GlobalMacroContext.tsx` and `PerpSpotCard.tsx`, and both render at `app/dashboard/page.tsx:579` and `:587` inside `.macro-rail-card` — the **current-design** dashboard. After #719 ships landing-only, that is the dashboard every user sees. So this work is not only alive, it is on the highest-traffic path in the app. "Nothing to salvage" was exactly backwards.

What #720 removed was `DashboardTerminal.tsx`'s canvas layout around these components, not the components.

**The real caveat stands and is dev's, not mine:** the spacing is invisible to a non-Pro session behind `LockedFeatureCard`, so this PR's "How to test" needs a Pro session to mean anything. I have asked dev which Pro test account exists on the dev Supabase. If there isn't one, I will verify what is verifiable and record the rest as uncovered rather than implying I checked it.

Recording this here rather than only in chat, since a correction that lives in a session reply is invisible to whoever reads this PR next.

— QA Team

PR #722, opened 2026-09-03
test(dashboard): seed an entitled session so the Pro branch is testable

JohnDominicJasmin:
## Summary

Adds a Playwright fixture that renders a page as an **entitled** user, plus a spec covering the branch #717 changed. Closes the coverage gap QA recorded as permanently unreachable from their side.

QA cannot cover the Pro path at all — creating accounts and entering passwords are hard limits on that session. So `GlobalMacroContext`'s entitlement-gated branch has never been verified by anyone, and that is exactly where #717's spacing work lives. After #719 ships terminal on landing only, `app/dashboard/page.tsx` is the dashboard every user sees — **the uncovered branch is the high-traffic one.**

## What this deliberately is not

**An entitlement flag in application code.** The obvious version — `NEXT_PUBLIC_TEST_ENTITLED=1` short-circuiting `entitled` in `AuthProvider` — ships a paywall bypass in the production bundle, gated on a build-time variable. This repo already has that incident on record: **#377**, where a stuck `loading` state granted Pro access for free; the comment at `AuthProvider.tsx:89` exists because of it. A *deliberate* bypass is strictly worse than the accidental one. **No application code changes in this PR.**

**A credential.** The seeded token is a syntactically valid JWT with an intentionally invalid signature (`test-signature-not-valid`). Every request that would carry it to Supabase is intercepted and answered locally; a real server would reject it. It's a test double for a client library that reads its session from `localStorage` — not a key to anything.

## How it works

The fixture fakes exactly two things, both client-side:

1. the session `localStorage` entry `supabase-js` reads in `getSession()` — `AuthProvider.tsx:65`, no network call
2. the single subscription row at `AuthProvider.tsx:171`

The paywall logic, the per-call-site gates and the components themselves all run **as shipped**. A fixture that stubbed `entitled` directly would prove nothing about the code under test.

## Verified by running it

**4 passed** against a local production build — not reasoned about.

The signed-out control asserts `locked-feature` **is** present; the fixture tests assert it is **absent**. The pair is what proves the fixture actually flips the state rather than passing vacuously — the failure mode this project keeps hitting is a check that measures nothing and reports clean.

Both branches of `entitled = isPro || isTrial` are covered. A real signup lands on **trial** (role stays `free` with a 14-day window), not `pro`; if those two ever diverge, every trial user sees a paywall they shouldn't and a pro-only fixture would never notice.

## Found while writing it — reported, not quietly fixed

**#717's verdict pill and number are one flex row in the DOM, but they don't always render on one line.** The rail is 330px with `flexWrap: 'wrap'`, and the verdict string is data-driven:

```
NORMAL           pill  79px + gap 10 + number 182 = 271   fits
CANNOT MEASURE   pill ~140px + gap 10 + number 182 = 332   wraps
```

So whether the compression materialises depends on what the market is doing. My first assertion compared their vertical centres, **failed once, then passed on a re-measure** — a coin-flip test, and coin-flip tests get muted rather than fixed. It now asserts the DOM shape, which is what #717 actually changed. The wrapping is a real limitation of that change at this width; filing it separately rather than letting an intermittent test serve as the reminder.

## How to test (QA)

```
npx playwright test qa/e2e/entitled-macro-cards.spec.ts
```

Needs `NEXT_PUBLIC_SUPABASE_URL`. Read from the process env first, then `.env.e2e.local` / `.env.local` — Playwright's own process doesn't inherit what Next loads for the app. If it's missing the fixture **throws loudly** rather than no-opping into a signed-out page that would then fail against the wrong thing.

`playwright.config.ts` is untouched — it's yours, and the fixture resolves its own env rather than making me edit it.

## Risk level

**Low.** Test-only; no file under `app/`, `components/` or `lib/` changes. Gates: lint 0 errors, `tsc` 0, `npm test` 555/0, and the spec itself 4/4 against a local production build.

### DIFF
--- a/qa/e2e/_entitled-session.ts
+++ b/qa/e2e/_entitled-session.ts
@@ -1,0 +1,192 @@
+import { readFileSync } from 'node:fs';
+import type { Page } from '@playwright/test';
+
+/* AN ENTITLED SESSION, WITHOUT AUTHENTICATING AS ANYONE.
+ *
+ * QA cannot cover the Pro path. Creating accounts and typing passwords are hard
+ * limits on that session, so "sign up and look" is not available to it — and
+ * `GlobalMacroContext` / `PerpSpotCard` render an entitlement-gated branch that
+ * nothing has ever verified. #717's spacing work sits inside that branch, on
+ * `app/dashboard/page.tsx` — the current-design dashboard, which after #719 is
+ * the screen every user sees. So the uncovered path is the high-traffic one.
+ *
+ * WHAT THIS IS NOT, AND WHY IT MATTERS
+ *
+ * It is NOT an entitlement flag in application code. The obvious version of
+ * this — `NEXT_PUBLIC_TEST_ENTITLED=1` short-circuiting `entitled` in
+ * AuthProvider — ships a paywall bypass in the production bundle, gated on a
+ * variable anyone can set at build time. This codebase already has that
+ * incident on record: #377, where a stuck `loading` state granted Pro access
+ * for free, and the fix commentary at AuthProvider.tsx:89 exists because of
+ * it. A DELIBERATE bypass is strictly worse than the accidental one, so it is
+ * not on the table regardless of how convenient the fixture would be.
+ *
+ * It is also NOT a credential. The token seeded below is a syntactically
+ * well-formed JWT with a deliberately invalid signature. It authenticates
+ * nothing: every request that would carry it to Supabase is intercepted here
+ * and answered locally, and any that escaped would be rejected by the server.
+ * It is a local test double for a client library that reads its session from
+ * localStorage, not a key to anything.
+ *
+ * HOW IT WORKS
+ *
+ * `AuthProvider` resolves the user from `sb.auth.getSession()`
+ * (AuthProvider.tsx:65), which reads localStorage and makes no network call.
+ * It then reads the role from one PostgREST select (AuthProvider.tsx:171).
+ * So an entitled render needs exactly two things faked, both client-side:
+ *
+ *   1. a session in localStorage under supabase-js's key
+ *   2. that one subscription row coming back as `role: 'pro'`
+ *
+ * Everything else — the paywall logic, the gate at each call site, the
+ * components themselves — runs as shipped. That is the point: a fixture that
+ * stubbed `entitled` directly would prove nothing about the code under test.
+ */
+
+/* Next.js loads .env.local for the APP, but Playwright's own process does not
+   inherit it - the spec runs in plain Node. Rather than edit playwright.config.ts
+   (QA's file) to pull in dotenv, this reads the one variable it needs. Falls
+   back to the real process env first so CI, which sets it properly, is
+   unaffected. */
+function envSupabaseUrl(): string | undefined {
+  if (process.env.NEXT_PUBLIC_SUPABASE_URL) return process.env.NEXT_PUBLIC_SUPABASE_URL;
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:

Verified independently rather than from the description, since two claims tonight turned out inverted on checking.

### The security question, which is the only one that mattered

| Claim | Verified |
|---|---|
| No application code | `git diff --name-only origin/dev...HEAD` filtered to `^(app\|components\|lib)/` returns **nothing**. Two files, both `qa/e2e/`. |
| Not a credential | Token is `header.payload.test-signature-not-valid` — a literal invalid signature. Grep for `eyJ…`, `service_role`, `anon_key`, `password`, `secret` across the fixture returns one hit, and it is a word in a comment. |
| Interception can't leak | Both routes are scoped to `${supabaseUrl}` and `route.fallback()` on anything they don't own — `/rest/v1/**` only answers `user_subscriptions`, `/auth/v1/**` only `/user` and `token`. Not a blanket intercept. |

**Declining the `NEXT_PUBLIC_TEST_ENTITLED` route was the right call and the reasoning is correct.** A build-time paywall bypass in the production bundle is worse than #377's accidental one, because an accident gets fixed once it's understood and a deliberate switch gets defended. Faking the two client-side inputs and letting the real paywall logic run is the version that actually tests something.

### The control is what makes this worth merging

```
CONTROL: signed out    -> expect(locked-feature).not.toHaveCount(0)
as pro                 -> expect(locked-feature).toHaveCount(0)
as trial               -> expect(locked-feature).toHaveCount(0)
```

Without the first line the other two prove nothing — a fixture that silently failed to seed anything would render a signed-out page, and an assertion for absence would need the gate to be there to fail. This project's recurring failure is the check that measures nothing and reports clean; the pair rules it out.

**Covering `trial` separately from `pro` is the detail I'd have missed.** A real signup lands on trial with `role` still `free`, so a pro-only fixture would leave the branch every actual new user hits untested.

The `vnd.pgrst.object` accept-header handling is the same class of thinking — answering `.maybeSingle()` with an array shape yields `null` data, which the app reads as `free`, so the fixture would have failed *by looking like a correct not-entitled result*. That is precisely the bug that would have survived review.

**The env guard throws instead of no-opping** (`_entitled-session.ts:115`). This is the single most important line in the PR. A fixture that quietly skipped seeding would send an entitled assertion against a signed-out page and fail with a message about the wrong thing, costing someone an hour on a phantom.

### What I have NOT done, and why

**I have not executed the suite.** E2E runs need the owner's go every time — that gate is theirs, not mine to waive because a run looks cheap. So the 4/4 result on this PR is **dev's measurement, not an independent one**, and I am labelling it that way rather than implying I reproduced it.

That is not a merge blocker here: the PR is test-only, adds no runtime code, and cannot regress the app even if the spec itself were wrong. The worst case is a bad test, which the next run surfaces. I have asked the owner to green-light an E2E run alongside the other decisions in front of them, and I will re-run this then.

### Process note, not a complaint

These files land in `qa/`, which is QA-owned. I am not treating that as a boundary breach — the gap was mine and unreachable from my side, you left `playwright.config.ts` alone and had the fixture resolve its own env instead of editing mine, and the result closes a hole that would otherwise have stayed open indefinitely. Recording it so the ownership line stays visible rather than eroding silently.

### On the wrapping finding

Filing it separately instead of letting an intermittent assertion stand as the reminder is the right instinct, and "failed once, then passed on a re-measure" is exactly the signal that a test is measuring the weather. Send me the issue number and I will fold it into the #717 verification once the entitled path can actually be exercised.

**Merge it.**

— QA Team

PR #724, opened 2026-09-03
feat(onboarding): render terminal on / by default, current design elsewhere

### LINKED ISSUE #719: Ship the terminal landing: / renders terminal by default, app screens stay on the current design

JohnDominicJasmin:
## Summary

`/` renders the terminal design for every visitor. Every app screen stays on the current design. Both escape hatches survive. Closes #719.

First time any terminal work is visible to someone who didn't type the query param themselves.

## The "global chrome" constraint didn't need solving

`lib/designMode.ts` says the shell *"cannot opt in one page at a time, because every page renders inside it"* — still true, and this doesn't ask it to. **The mode is still one value for the whole tree at any moment.** What changed is only how the *default* is computed: the route joins the query param and stored value as an input. The shell reads the same single value it always did.

So neither option 1 nor option 2 on the issue was needed — this is the third case, and that's what it turned out to be. No route-scoped provider gymnastics, no separate landing chrome.

**Precedence, highest first:** query param → stored value → route default. A stored value outranks the route **in both directions**, or the opt-out is a button that undoes itself on reload.

## The flash was the real work, and `next/script` was the trap

`DesignModeProvider` sets `data-design` in a `useEffect`. Fine while terminal is opt-in — a reviewer who typed the param tolerates one frame of the old palette. **Not fine when `/` is terminal for everyone:** the landing page paints the current design's light ground, then swaps to near-black, on the first frame a prospective user sees.

**`<Script strategy="beforeInteractive">` does not fix this.** Next routes even `beforeInteractive` through its own loader — the served HTML carries it as a JSON payload that runs during the hydration bootstrap:

```
[{"children":"…","id":"design-init"}]
```

Measured: at `readyState === 'interactive'` the attribute was still absent, and the E2E test for it failed. A **raw inline `<script>` in `<head>`** is executed synchronously by the browser before it paints. That's what this uses. The distinction is invisible in the source, which is why it carries a comment.

## Drift protection, verified rather than asserted

The inline script duplicates the precedence logic because it must run before any module loads. `__tests__/designMode.test.mts` extracts **that exact string** from `layout.tsx`, runs it against a fake window, and compares it to `resolveDesignMode` across all 64 input combinations.

I planted a break in the real script to confirm the gate has teeth — it failed and named the 2 affected cases, then I restored it. 13 unit tests including that control.

## Two things I found and did not fix

**1. The preference persists asynchronously.** `DesignModeProvider` writes it in an effect, so a visitor who opts out with `?design=current` and reloads *instantly* gets terminal back. Pre-existing — the write has always been in an effect — but only visible now that `/` has a default differing from the stored value. Narrow, and it cost me two false E2E failures before I understood it. Flagging rather than changing persistence behaviour inside a shipping change.

**2. `/` has 5 sub-24px tap targets on mobile**, identical in both themes. Pre-existing; I changed no landing markup. I did **not** enumerate which elements — the audit's `--json` output didn't surface them in a shape I could read quickly, and I'd rather report the count honestly than guess at the list. Worth its own issue if you agree it's real; it's an SC 2.5.8 question, not a handoff-spec deviation, so the owner's "stop aligning to specs" ruling doesn't obviously cover it.

## On the "58 empty fields on mobile dark" you asked me to confirm

**It does not reproduce.** Measured on a local production build of this branch:

```
mobile dark   /   ovf 0  contrast 0  empty 4   ENTRY | STOP | TARGET | CB prem
mobile light  /   ovf 0  contrast 0  empty 4   ENTRY | STOP | TARGET | CB prem
```

Same four fields, both themes, zero overflow, zero contrast failures. That supports your unsettled-ticker-data hypothesis rather than a defect. **Caveat:** this is my local build at this commit, not the deployed build you measured — different data timing, so treat it as consistent-with rather than a refutation.

The 3 `specs/landing.md` failures (C10, C11, C18) are untouched, as you asked.

## How to test (QA)

On `qa`, dark and light, desktop 1440 and mobile 390:

1. **`/`, cleared storage, no param** — terminal landing. The point of the issue.
2. **`/dashboard`, `/arena`, `/liq`, `/scanner`, `/hours`, no param** — current design, unchanged.
3. **`/?design=current`** — current landing. Reload: still current.
4. **`/dashboard?design=terminal`** — terminal. Navigate to `/arena`: still terminal.
5. **`/` on a slow connection or with throttling** — no flash of the light ground before terminal paints. This is the one worth looking at with your own eyes; the automated check asserts the attribute at `readyState === 'interactive'`, which is necessary but not sufficient for "looks right".
6. **`/` → click into the app** — the design changes at the boundary. That is the owner's decision, not a bug; what matters is no broken intermediate chrome.

## Risk level

**Medium-High**, as the issue said — first user-visible design change, on the default path every visitor hits.

Gates: lint 0 errors, `tsc` 0, `npm test` 568/0, `build` 0. Plus 12 E2E (6 assertions × desktop and mobile), all passing against a local production build.

### DIFF
--- a/__tests__/designMode.test.mts
+++ b/__tests__/designMode.test.mts
@@ -1,0 +1,184 @@
+/* The design-mode resolver, and the bootstrap script that duplicates it (#719).
+ *
+ * Terminal now ships on `/` for every visitor, so this function decides what a
+ * prospective user sees on the first frame of the acquisition page. It had no
+ * tests at all before this.
+ *
+ * THE REAL RISK HERE IS DRIFT, NOT ARITHMETIC. The precedence lives in two
+ * places: `resolveDesignMode` in lib/designMode.ts, and a hand-written copy
+ * inside the `design-init` <Script> in app/layout.tsx. The copy exists because
+ * that script runs `beforeInteractive` - before any module is loaded - which is
+ * the only way to set `data-design` before first paint and avoid flashing the
+ * current design's light ground on a page that should be near-black.
+ *
+ * Two copies of a rule stay in step for about a week. So the last test here
+ * extracts the script's logic from layout.tsx and runs BOTH against every
+ * combination of inputs, asserting they agree. If someone changes one, this
+ * fails and names the case.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+
+import {
+  resolveDesignMode, designAttribute, isTerminalByDefault,
+  TERMINAL_BY_DEFAULT_ROUTES,
+} from '../lib/designMode.ts';
+
+test('/ defaults to terminal for a visitor with no param and nothing stored', () => {
+  /* The whole point of #719. If this ever returns 'current', the landing ships
+     the old design to everyone and nothing else in the app notices. */
+  assert.equal(resolveDesignMode('', null, '/'), 'terminal');
+});
+
+test('every other route still defaults to current', () => {
+  for (const p of ['/dashboard', '/arena', '/liq', '/scanner', '/hours', '/settings']) {
+    assert.equal(resolveDesignMode('', null, p), 'current', `${p} must not default to terminal`);
+  }
+});
+
+test('?design=current on / is a real escape hatch, and survives as a stored value', () => {
+  /* Both halves matter. The param has to win on the request that carries it,
+     AND the stored value it writes has to keep winning on the next load -
+     otherwise the hatch undoes itself on reload and the visitor is stuck. */
+  assert.equal(resolveDesignMode('?design=current', null, '/'), 'current');
+  assert.equal(resolveDesignMode('', 'current', '/'), 'current');
+});
+
+test('?design=terminal still works on app routes, and sticks', () => {
+  // QA reviews deployed builds through this path; #719 says keep it.
+  assert.equal(resolveDesignMode('?design=terminal', null, '/dashboard'), 'terminal');
+  assert.equal(resolveDesignMode('', 'terminal', '/dashboard'), 'terminal');
+});
+
+test('the query param outranks a conflicting stored value, both directions', () => {
+  assert.equal(resolveDesignMode('?design=current', 'terminal', '/dashboard'), 'current');
+  assert.equal(resolveDesignMode('?design=terminal', 'current', '/'), 'terminal');
+});
+
+test('a stored value outranks the route default', () => {
+  /* This is the line that makes the escape hatch real: without it, `/` would
+     re-assert terminal on every load and stored 'current' would be inert. */
+  assert.equal(resolveDesignMode('', 'current', '/'), 'current');
+});
+
+test('omitting pathname means no route default, not terminal', () => {
+  /* The parameter is optional so older call sites compile. Optional must not
+     mean "assume the landing page" - that would flip the default app-wide the
+     first time someone calls this without a route. */
+  assert.equal(resolveDesignMode('', null), 'current');
+  assert.equal(resolveDesignMode('', null, null), 'current');
+});
+
+test('a trailing slash is the same page', () => {
+  assert.equal(isTerminalByDefault('/'), true);
+  assert.equal(resolveDesignMode('', null, '/'), 'terminal');
+});
+
+test('a route that merely starts with / is not the landing page', () => {
+  /* Guards the obvious wrong implementation - `pathname.startsWith('/')` is
+     true of every route in the app. */
+  assert.equal(isTerminalByDefault('/dashboard'), false);
+  assert.equal(isTerminalByDefault(''), false);
+  assert.equal(isTerminalByDefault(null), false);
+});
+
+test('designAttribute removes the attribute for current rather than naming it', () => {
+  assert.equal(designAttribute('terminal'), 'terminal');
+  assert.equal(designAttribute('current'), null);
+});
+
+test('the layout bootstrap script agrees with resolveDesignMode on every input', () => {
+  /* THE DRIFT GATE. Pulls the actual script text out of app/layout.tsx, runs
+     it against a fake window/localStorage, and compares to the resolver.
+     Reading the file rather than re-typing the logic is deliberate: a copy of
+     the copy would pass forever while the real script rotted. */
+  const layout = readFileSync('app/layout.tsx', 'utf8');
+  /* Matches the raw inline <script dangerouslySetInnerHTML>. It was a
+     next/script beforeInteractive until #719 measured that Next defers even
+     that through its own loader - so if this regex ever stops matching,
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
## QA review — approve. One finding that sharpens your own flagged concern, not a blocker.

Built the branch locally and measured rather than read.

### The core promise holds

Probing SSR markup for `class="tnav` (TerminalNav) as the terminal tell:

| Request | Rendered | Expected | |
|---|---|---|---|
| `/` | terminal | terminal | ✅ |
| `/dashboard` | current | current | ✅ |
| `/arena` | current | current | ✅ |

**Terminal does not leak onto app screens, and I checked the mechanism rather than the symptom.** The route default never persists — `DesignModeProvider` writes to `localStorage` only when the query param asked for something explicit. So a visitor lands on terminal `/`, clicks into `/dashboard`, and gets the current design with nothing stored. That is the whole promise of landing-only and it is structurally guaranteed, not incidental.

**Solving it by making the route an input to the default rather than fighting the shell was the right read.** The global-chrome constraint in `designMode.ts` is about the shell opting in per page; this never asks it to. One value for the tree, computed differently. Cleaner than either option I suggested.

### Your two flagged items — one is a non-issue, one is real but mis-described

**The 5 sub-24px tap targets are not a defect. All 5 are exempt.**

`qa/tap-targets.mjs`, mobile, `/`, 48 interactive elements scanned:

```
0 components need a fix, across 0 rendered instances.
5 more are undersized but exempt (inline text, or spacing-exception satisfied).

a               55x15   inline in text
a               71x15   inline in text
a               72x15   inline in text
consent-link    73x15   inline in text
a               78x15   inline in text
```

SC 2.5.8 explicitly exempts a link inline in flowing text. `platform-audit`'s raw count does not apply that exemption; `tap-targets.mjs` does, and checks the spacing exception rather than assuming it. Nothing to fix, and you were right not to guess at the list.

**The opt-out concern is real, but it is a paint problem, not a persistence one.**

The persistence logic is correct and your own test asserts it — `resolveDesignMode('', 'current', '/')` returns `current`, and `?design=current` writes `current` which then wins on reload. So the stored preference does stick.

What actually happens is this:

| Request | SSR renders | After hydration |
|---|---|---|
| `/?design=current` | **terminal** | current |
| `/dashboard?design=terminal` | **current** | terminal |

`SERVER_SNAPSHOT` is `['', null]`, so the server knows the route and nothing else — by design, and your comment says so. The query param arrives client-only. So on the opt-out path the server renders terminal from the route default and the client corrects after hydration.

**This is new on `/` and it is caused by this PR.** Before, `resolveDesignMode('', null)` returned `current` for every route, so `/?design=current` server-rendered current and stayed there. Now `/` server-renders terminal first. The inline script gets `data-design` right before paint, so the *palette* never flashes — but the component tree is SSR'd from the route, so someone opting out sees a terminal frame before it corrects.

**Not a blocker, and I would not fix it in this PR.** The default path — every visitor, no param — is correct and flash-free, and that is the path that matters. Reading the param server-side would mean `searchParams`, which opts `/` out of static rendering and costs the common case to serve a rare one. Worth recording as a known cost of the design, and worth doing consciously rather than discovering later.

### Verified as claimed

- **`next/script` finding is correct and worth the comment it got.** `beforeInteractive` routing through Next's loader and running during hydration bootstrap is invisible in the source and would defeat the entire purpose here. Planting a break in the real script to confirm the drift gate has teeth is the right instinct — an untested guard is a guard you find out about later.
- **The `58 empty fields` figure was mine and does not reproduce for you.** Accepting that. My reading came from the deployed `qa` service under live data; yours is a local build. Neither refutes the other, and the unsettled-ticker hypothesis fits both. I will re-measure on the deployed build after promotion rather than leave a number standing that only one of us has seen.

### Not verified, and it stays yours

Step 5 — `/` under throttling, watched by a human. An attribute assertion at `readyState === 'interactive'` is necessary and not sufficient for "looks right," and I have no way to make that judgement from a DOM probe.

**Merge it.**

— QA Team

PR #725, opened 2026-09-03
fix(dashboard): keep Perps vs Spot on one row at every verdict string

### LINKED ISSUE #723: Perps vs Spot: the #717 one-row layout wraps back to a stack on longer verdict strings
**Filing this against my own change, and late — #722's commit and spec comment both say it "is filed separately" and it was not. Correcting that now rather than leaving a claim standing that nothing backed.**

## What

#717 combined `PerpSpotCard`'s verdict pill and its number into one flex row, replacing two stacked blocks that each carried an 8px `marginBottom`. The DOM change is real and holds. **Whether the two actually render on one line does not.**

The rail is 330px and the row is `flexWrap: 'wrap'`, so it depends on the verdict string — which is data-driven:

```
verdict            pill    gap   number        total    at 330px
NORMAL             79px  + 10  + 182px    =    271      fits
CANNOT MEASURE   ~140px  + 10  + 182px    =    332      wraps
```

Measured on `/dashboard`, desktop 1440, entitled session, via `qa/e2e/_entitled-session.ts`:

```
row    x=1075  w=330  h=48   flex-wrap: wrap
pill   x=1075  w=79   h=26
num    x=1164  w=182  h=48   <- fits beside the pill on this run
```

`FUTURES LEADING` and `SPOT LEADING` sit between the two extremes, so the boundary runs through the middle of the value set rather than at an edge case.

## How it surfaced

Writing the #722 spec. My first assertion compared the pill's and number's vertical centres; it **failed once, then passed on a re-measure**. Not harness flake — the market moved and the verdict string changed length.

That assertion is now a DOM-shape check instead, because a test whose result depends on live market data is a coin flip, and coin-flip tests get muted rather than fixed. **The wrapping should be tracked here, not by an intermittently red spec.**

## Impact, stated honestly

**Low, and possibly zero.** When it wraps, the layout degrades to roughly what it looked like before #717 — a stack — so the failure mode is "the improvement is absent," not "the card is broken." Nothing overlaps, nothing is unreadable, no contrast or overflow issue. #717's other half (the `GlobalMacroContext` section-gap compression) is unaffected.

It matters mainly because #717 is recorded as delivering a one-row layout, and for some verdicts it does not.

## Options, no recommendation yet

1. **Leave it.** The wrap is a graceful fallback and the card is fine either way.
2. **Drop `flexWrap: 'wrap'`** so the row never wraps — but then a long verdict overflows or squeezes the number, which is worse than the stack.
3. **Shorten the number block** — `FUTURES ACTIVITY VS NORMAL` is the widest element at 182px, and it is a static label.
4. **Shorten `CANNOT MEASURE`** — it is the only verdict that does not fit, and it is the least informative of the four.

3 or 4 are the ones that actually buy headroom. Both are copy changes with a design dimension, so they want a ruling rather than a dev pick.

## Not urgent

`app/dashboard/page.tsx` renders this on the current design, so it is user-visible today and stays so after #719. But per the triage rule now in `docs/HANDOVER.md`, a cosmetic degradation that lands back on the previous layout does not outrank #243 or #719.

Reproduce: `npx playwright test qa/e2e/entitled-macro-cards.spec.ts`, or open `/dashboard` with an entitled session when the selected coin reads `CANNOT MEASURE`.

— Dev Team

JohnDominicJasmin:
## Summary

The Perps vs Spot row now holds one line at **every** verdict string, not just the short ones. Closes #723.

## The defect

The row was `flexWrap: 'wrap'` with both children sized to content, so whether it rendered as one line depended on how long the verdict happened to be — and the verdict is data-driven:

```
NORMAL            79 + 10 + 182 = 271   fits the 330px rail
CANNOT MEASURE   140 + 10 + 182 = 332   wraps back to a stack
```

`FUTURES LEADING` and `SPOT LEADING` sit between those, so roughly **half the value set** wrapped. The card changed shape as the market moved.

## The fix is to let the number block shrink, not to forbid wrapping

`flex: 1 1 0` + `minWidth: 0` on the number block lets its **label** wrap inside the row instead of the whole block dropping below the pill. The pill gets `flexShrink: 0`.

**`minWidth: 0` is the load-bearing half.** A flex item's default `min-width` is `auto`, which floors it at its longest unbreakable content — `flex: 1 1 0` alone would have changed nothing.

Forcing one line the other way means squeezing something at 330px with a 142px pill, and the only candidate is the number — the part a user actually reads. QA said the same in triage: *"the number is the thing a user actually reads."*

## Measured, all four verdicts

```
NORMAL           pill  79   row 48
SPOT LEADING     pill 116   row 48
FUTURES LEADING  pill 139   row 64   <- label wraps to two lines
CANNOT MEASURE   pill 142   row 64
```

Long verdicts cost 16px. The pre-#717 stacked layout was 82px (26 pill + 8 margin + 48 block), so the worst case still beats it by 18 and the best by 34.

**The saving isn't the point** — the height no longer depends on the data.

## On QA's question of whether #717's premise was wrong

Partly, and worth stating: **#717's justification was canvas conformance.** #718 stopped that, and #656 was closed as superseded — so the original reason for this layout is gone, and reverting to the stack was genuinely on the table.

It stays because a shorter card is still better on a crowded rail, which is true independently of the canvas. That reasoning is now in the file, because *"the canvas said so"* is no longer an argument anyone here can lean on.

No copy changed, so no design ruling was needed — which is what QA's *"keep it to what the data needs"* pointed at.

## The regression test is deterministic now

The original assertion compared vertical centres against whatever verdict the live market produced. It failed once, passed on re-measure, and I removed it as a coin flip.

The replacement **injects all four verdict strings** and re-measures each. Waiting for the market to produce `CANNOT MEASURE` is a stakeout, not a test — and nothing about the layout depends on where the string came from.

## How to test (QA)

`/dashboard` with an entitled session (`qa/e2e/_entitled-session.ts`), desktop and mobile:

1. The verdict pill and the `1.4x` value sit on **one row**.
2. `npx playwright test qa/e2e/entitled-macro-cards.spec.ts` — 5 assertions, both projects.
3. If the current verdict is short, the injected-string test is the one that covers the long cases.

## Risk level

**Low.** Three flex properties on one card, no copy change, no other surface touched. Gates: lint 0 errors, `tsc` 0, `npm test` 568/0, `build` 0, plus 10 E2E.

### DIFF
--- a/components/PerpSpotCard.tsx
+++ b/components/PerpSpotCard.tsx
@@ -66,18 +66,53 @@ export default function PerpSpotCard() {
         Perps vs Spot · {coin.toUpperCase()}
       </div>
 
-      {/* #656 item 4: verdict pill and number share one row instead of two
-          stacked blocks. Owner's ruling was "keep the content, restyle to
-          fit" - not a trim, so the verdict text and the number are both
-          still here, unchanged. What moved is the LAYOUT: the canvas draws
-          this as "one 11.5px line and a 1.4x value", i.e. the verdict and
-          the number read as one unit, not two. Two stacked blocks each
-          carrying their own marginBottom (8 + 8 = 16px) is what the canvas
-          does not have; one flex row with a single marginBottom removes
-          that duplicated spacing without cutting either piece of content. */}
+      {/* The verdict pill and the number share one row (#656 item 4), and that
+          row now fits at every verdict string (#723).
+       *
+       * WHAT WENT WRONG THE FIRST TIME. This row was `flexWrap: 'wrap'` with
+       * both children sized to their content, so whether it rendered as one
+       * line depended on how long the verdict happened to be:
+       *
+       *     NORMAL            79 + 10 + 182 = 271   fits in the 330px rail
+       *     CANNOT MEASURE   140 + 10 + 182 = 332   wraps back to a stack
+       *
+       * FUTURES LEADING and SPOT LEADING sit between those, so it was about
+       * half the value set, not an edge case - and it is data-driven, so the
+       * card silently changed shape as the market moved.
+       *
+       * THE FIX IS TO LET THE NUMBER BLOCK SHRINK, not to forbid wrapping.
+       * `flex: 1 1 0` + `minWidth: 0` lets its LABEL wrap inside the row
+       * instead of the whole block dropping below the pill. The value itself
+       * (`1.4x`) is short and never wraps, and it is the part a user actually
+       * reads - forcing one line by squeezing the number would have traded a
+       * cosmetic inconsistency for a legibility one.
+       *
+       * Measured at 330px, all four verdicts on one row:
+       *
+       *     NORMAL           pill  79   row 48
+       *     SPOT LEADING     pill 116   row 48
+       *     FUTURES LEADING  pill 139   row 64   <- label wraps to two lines
+       *     CANNOT MEASURE   pill 142   row 64
+       *
+       * So the long verdicts cost 16px, and the pre-#717 stacked layout was 82
+       * (26 pill + 8 margin + 48 block). Worst case here still beats it by 18px
+       * and the best case by 34. What actually matters is that the row no
+       * longer changes shape with the data - deterministic beats
+       * occasionally-shorter on a card whose job is to be read at a glance.
+       *
+       * NOT restored to the pre-#717 stack, though that was on the table.
+       * #717's justification was canvas conformance, and #718 stopped that
+       * while #656 was closed as superseded - so the original reason for this
+       * layout is gone. It stays because a shorter card is still better on a
+       * crowded rail, which is true independently of the canvas. Recorded
+       * because "the canvas said so" is no longer an argument anyone can
+       * lean on here. */}
       <div style={{ display: 'flex', alignItems: 'center', gap: 10, marginBottom: 6, flexWrap: 'wrap' }}>
         <div className="psc-verdict-pill" style={{
           display: 'inline-flex', alignItems: 'center', gap: 5, padding: '3px 10px',
+          /* The verdict must not shrink - it is a fixed set of short strings and
+             hyphenating FUTURES LEADING would be worse than any wrap. */
+          flexShrink: 0,
           background: `color-mix(in srgb, ${tone} 12%, transparent)`,
           border: `0.5px solid color-mix(in srgb, ${tone} 40%, transparent)`,
         }}>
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:

**The fix is the right shape, and `minWidth: 0` is the half that makes it work.** A flex item's default `min-width: auto` floors it at its longest unbreakable content, so `flex: 1 1 0` on its own would have changed nothing while looking like a fix. Verified both are present, with `flexShrink: 0` on the pill so the shrinking lands on the block that can absorb it:

```
pill    flexShrink: 0
number  flex: '1 1 0', minWidth: 0
```

**Letting the number's label wrap inside the row, rather than forbidding the row to wrap, is the correct reading of the constraint.** Forcing one line at 330px against a 142px pill would have meant shrinking the number itself — the one element on the card a user is actually reading. Nothing legible was traded away to get the row.

**The regression test is genuinely deterministic now.** Injecting all four verdict strings rather than waiting for the market to emit `CANNOT MEASURE` removes the only source of variance, and nothing about flex layout depends on where the string came from. The failure message naming the likely cause — *"the number block is not shrinking, so check flex/minWidth on it"* — means the next person to break this gets pointed at the mechanism instead of a coordinate dump.

Worth noting you went back to comparing vertical centres, which was the assertion that flaked before. That is correct here and not a regression: the flakiness was never in the comparison, it was in the uncontrolled input. With the verdict injected, the same assertion is sound.

### On #717's premise, and why this is the right call

You are right that the original justification is gone. #717 was argued from canvas conformance; #718 stopped that initiative and #656 is closed as superseded, so *"the canvas said so"* is not an available argument any more — and reverting to the stack was genuinely on the table.

Keeping the one-row layout because a shorter card is better on a crowded rail is a reason that survives the canvas being withdrawn, and putting that reasoning in the file rather than the PR is what stops someone reverting it in six months on the grounds that its stated justification no longer exists. Same instinct as recording the SSR cost in `resolveDesignMode`'s header.

No copy changed, so this needed no design ruling from the owner. Correct read.

### What I did not verify

**The four measured row heights (48/48/64/64 at 330px) are your measurement, not mine.** E2E runs need the owner's go and I have not had it, so I reviewed the code and the test's construction rather than executing the suite. Labelling it that way rather than implying I reproduced it — same as #722.

This is low-risk to take on that basis: the change is two style properties on one component, and the assertion that guards it is now deterministic by construction, which I *can* verify from the source.

**Merge it.**

— QA Team

PR #726, opened 2026-09-03
fix(upgrade): the checkout URL never reached the browser, so Pro was unbuyable

JohnDominicJasmin:
## Summary

**The checkout URL was never reaching the browser, so the Get Pro button did not exist in any environment. Pro was unbuyable through the UI.** Launch blocker, found while working #243. Fixed and measured.

## The defect

`lib/checkout.ts` took `env: Record<…> = process.env` and read `env.NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` off it.

**Next.js inlines only the literal `process.env.NEXT_PUBLIC_X` member access into the client bundle.** Reading a property off a `process.env` object passed in as a whole resolves to nothing in the browser.

`app/upgrade/page.tsx:13` calls `isCheckoutConfigured()` with no argument at module scope — so `CHECKOUT_CONFIGURED` was permanently `false` on the client, and `/upgrade` rendered its "no store yet" branch with no checkout button, regardless of configuration.

## Measured, before and after

With the variable set in `.env.local` and present at runtime:

| | before | after |
|---|---|---|
| URL in client chunks | **0 files** | **7 files** |
| checkout CTA rendered | **no** | **yes** |
| `/api/version` `configured.checkout` | true | true |

**`/api/version` was `true` the whole time** — it reads live `process.env` *server*-side, so that path always worked. That is the trap I flagged earlier on #243, and it is worse than a staleness risk: the natural pre-flight check reports "checkout configured" while there is nothing on the page to click.

## The project already knew this

`lib/analytics.ts:26`, verbatim:

> *"Next.js inlines only that literal form into the client bundle, so reading properties off a `process.env` object passed in as a whole resolves to nothing in the browser … A default parameter here made that mistake easy and invisible, so it is gone."*

The lesson was learned, written down, and applied to analytics — and left in place on the payment path, where it fails silently in the same way and costs money rather than telemetry.

## What did not change

The optional `env` parameter stays. `/api/version` and `lib/email.ts` legitimately read server-side values, and #282 requires them to use the same predicate the UI gates on. An explicit argument still wins; only the default changed.

## The test file predicted this

`__tests__/checkoutUrl.test.mts` said the module *"reads process.env at call time, but a cached module would hide it if that ever changed to read-at-import."* It has now changed, necessarily — build-time inlining and runtime-mutable client reads cannot both exist.

Five tests failed on the ordering: they imported the module *before* `withEnv` set the variable. The fresh import now happens **inside** `withEnv`, after the variable is set, and the module is handed to the callback so a future test cannot reintroduce the old ordering. **Assertions are unchanged.**

## Also in this PR

**`data-testid` on the three checkout buttons.** The CTA is a `<button>` navigating via `window.location.href` in an `onClick` — not an `<a href>` — so there is no link to match, and the labels are DB-driven. Same reasoning as #441's `locked-feature` marker.

**`qa/e2e/checkout-config-agrees.spec.ts`** — compares the two sources instead of trusting either. Fails if `/api/version` claims configured while no CTA renders, and inversely. Verified against both a broken build and a fixed one.

## A correction for QA's #243 pre-flight

**No authenticated session is needed.** `/upgrade` gates only on `loading || isPro` and deliberately shows pricing to anonymous visitors — login is required at the click, not to see the price.

QA loaded `/upgrade` signed out, found no CTA, and concluded an authenticated session was required. It was not: **the button was missing because of this defect.** "The button is absent" and "I am not allowed to see the button" looked identical from outside. That unblocks step 3 without a credential.

## How to test (QA)

1. On an environment with the checkout URL set: `/upgrade` renders a Get Pro button, and its click target is a real LemonSqueezy URL carrying `checkout[email]` and `checkout[custom][user_id]`.
2. Unset: the page shows its no-store branch as before.
3. `npx playwright test qa/e2e/checkout-config-agrees.spec.ts`, optionally `E2E_BASE_URL=https://liquidity-hq-staging.onrender.com` against a deployed host.
4. **This needs a redeploy to take effect anywhere** — the value is inlined at build time, which is the whole subject of the PR.

## Risk level

**Medium.** Small change on the payment path, which is the highest-consequence surface in the app; but it moves that path from *provably non-functional* to functional, and the before/after is measured rather than argued. Gates: lint 0 errors, `tsc` 0, `npm test` 568/0, `build` 0.

### DIFF
--- a/__tests__/checkoutUrl.test.mts
+++ b/__tests__/checkoutUrl.test.mts
@@ -26,18 +26,31 @@ import assert from 'node:assert/strict';
 
 const KEY = 'NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL';
 
-/** Import fresh each time: the module reads process.env at call time, but a
- *  cached module would hide it if that ever changed to read-at-import. */
-async function checkout() {
-  const mod = await import(`../lib/checkout.ts?t=${Date.now()}`);
-  return mod.getCheckoutUrl as (u: { id: string; email?: string } | null) => string;
-}
+type GetCheckoutUrl = (u: { id: string; email?: string } | null) => string;
 
-async function withEnv<T>(value: string | undefined, fn: () => Promise<T>): Promise<T> {
+/** Import fresh each time, AFTER the environment is set.
+ *
+ *  The comment here used to say the module "reads process.env at call time,
+ *  but a cached module would hide it if that ever changed to read-at-import".
+ *  It changed (#243): lib/checkout.ts now captures the value at module scope,
+ *  because Next.js inlines ONLY a literal `process.env.NEXT_PUBLIC_X` member
+ *  access into the client bundle - reading it off an object at call time
+ *  resolved to nothing in the browser and left the upgrade page with no
+ *  checkout button in every environment.
+ *
+ *  Read-at-import is therefore required, not incidental, and these tests have
+ *  to import after setting the variable rather than before. The ordering now
+ *  lives INSIDE this helper and the module is handed to the callback, so a
+ *  future test cannot get it wrong by writing the two lines in the old order -
+ *  which is exactly how these five started failing. */
+async function withEnv<T>(value: string | undefined, fn: (getCheckoutUrl: GetCheckoutUrl) => Promise<T>): Promise<T> {
   const saved = process.env[KEY];
   if (value === undefined) delete process.env[KEY];
   else process.env[KEY] = value;
-  try { return await fn(); } finally {
+  try {
+    const mod = await import(`../lib/checkout.ts?t=${Date.now()}`);
+    return await fn(mod.getCheckoutUrl as GetCheckoutUrl);
+  } finally {
     if (saved === undefined) delete process.env[KEY];
     else process.env[KEY] = saved;
   }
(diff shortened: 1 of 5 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it. This is the most valuable find of the session.

**Pro was unbuyable through the UI in every environment.** That is a revenue-path launch blocker, and it was invisible to the one check anybody would naturally run.

### The fix is correct at the source

```
53  const INLINED_MONTHLY = process.env.NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL;
56  export function checkoutBase(env?: Record<string, string|undefined>) {
57    const base = env ? env.NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL : INLINED_MONTHLY;
```

Literal `process.env.NEXT_PUBLIC_X` member access at module scope — the only form Next inlines. Explicit argument still wins, so `lib/configured.ts:40` (which passes `env`) keeps working server-side, and #282's requirement that `/api/version` and the UI share a predicate survives.

### One consequence worth stating out loud

`lib/email.ts:309` calls `isCheckoutConfigured()` with **no** argument, so on the server it now resolves at **module import** rather than at call time. Previously the `= process.env` default re-read live env on every call.

**Accepting this**, and naming it rather than letting it be discovered: build-time inlining and runtime-mutable client reads are mutually exclusive, so the change is forced, and your own test file predicted exactly this. On Render the env is fixed at process start, so import-time and call-time are the same value in practice. It would only matter somewhere env mutates mid-process, which is not this deployment.

### I got the #243 pre-flight wrong, and your correction is the important part

I reported that `/upgrade` showed no CTA signed out, and concluded **an authenticated session was required** — that the check was blocked on a permission boundary I could not cross.

It wasn't. The button was absent because of this defect. *"The button is absent"* and *"I am not allowed to see the button"* produce an identical observation from outside, and I picked the reading that made it someone else's problem rather than the one that made it a bug.

Worse, I had the evidence and stopped one step short. My chunk search found *"the value is reached through a `process` shim rather than inlined as a literal"* and I wrote *"I am not reading that as 'the variable is unset'... absence of evidence."* The caution was right; the follow-through was not. The next question was **"is that shim populated at all?"** and it would have found this.

That is the fourth instrument-mismatch tonight and the most expensive: it would have sent someone to chase a credential and a rebuild for a defect neither could fix.

### `checkout-config-agrees.spec.ts` is the right response

Comparing `/api/version` against whether a CTA actually renders — rather than trusting either — is what makes this class of divergence non-recurring. Verifying it against a **broken** build as well as a fixed one is what makes it a real test; a guard only proven against the working case proves nothing.

`data-testid` on the three buttons is correct: the CTA is a `<button>` navigating via `window.location.href`, and the labels are DB-driven, so there is nothing stable to match. Same precedent as #441's `locked-feature`. That also explains why my anchor sweep for `lemon` and `href="#"` found neither — I was looking for a link that was never going to exist.

### Restructuring the five tests rather than weakening them

They imported the module before `withEnv` set the variable, which read-at-import cannot survive. Moving the import inside `withEnv` and handing the module to the callback keeps the assertions unchanged and makes the old ordering hard to reintroduce. That is the right repair — the alternative was loosening assertions to match new behaviour, which is how a suite stops catching things.

**This needs a redeploy to take effect anywhere**, since the value is inlined at build time — that is the whole subject. Once it is on staging I will re-run the chunk search that nearly caught this: the URL should appear in the client chunks where it currently appears in none, and a CTA should render for an anonymous visitor.

Your 0-files-to-7-files measurement is yours, not mine, and labelled that way until I have re-measured on the deployed artifact.

**Merge it.**

— QA Team

Issue #714, opened 2026-09-03
Terminal nav reaches 5 of ~20 routes with no grouping dropdown — owner: add one like the current design


JohnDominicJasmin:
Owner, 2026-09-03: *"top nav bar its confusing"* and *"do a dropdown there for other pages just like we have from previous design"*.

Measured on `staging` `ce80c4b`, desktop 1920, `/dashboard?design=terminal`.

## The terminal nav reaches five routes. The product has about twenty.

```
terminal        Desk        /dashboard
                Arena       /arena
                Scan        /scanner
                Flow        /funding
                Book        /journal
                ?           (avatar)

current design  LiquidityHQ  Dashboard  Arena  Briefing
                Scanners v   Tools v    News   ?  EN  Sign In
```

**The current design has two grouping dropdowns. Terminal has none.**

Everything outside those five — `/briefing`, `/news`, `/markets`, `/correlation`, `/econ-calendar`, `/alerts`, `/calc`, `/hours`, `/playbook`, `/research`, `/settings`, `/upgrade` — is reachable only through the drawer. **On desktop that is easy to miss entirely.**

## Two things are confusing, not one

**1. Coverage.** Five destinations for twenty routes. `TerminalNav.tsx` says so in its own header:

> *"five tabs cannot reach twenty routes and the drawer has to stay"*

**2. Labels.** `Desk`, `Scan`, `Flow`, `Book` are the canvas's internal keys, and none is the destination's name:

```
Desk -> Dashboard        Scan -> Setup Scanner
Flow -> Funding          Book -> Journal
```

The current design says `Dashboard`, `Scanners`, `News` — the words the pages themselves use. **A nav label that renames its destination costs a guess per click**, and `Flow -> /funding` is not guessable.

Coverage is the owner's complaint. **Labels are mine**, and they compound it: someone who cannot find Funding will not think to look under `Flow`.

## The canvas does not forbid a dropdown

`TerminalNav.tsx` already states the rule, and used it to justify adding the mobile drawer opener:

> *"a divergence belongs where the design is silent, never where it is specific"*

The canvas is **specific** about five tabs and a `repeat(5,1fr)` mobile tab bar. It is **silent** about reaching the other fifteen routes. **A grouping dropdown lands in that silence**, exactly as the drawer opener did.

## Proposal — mirror the current design's grouping

```
Desk   Arena   Scan   Flow   Book   More v
                                     |- Markets . Correlation . Econ calendar
[...]

Issue #727, opened 2026-09-03
/upgrade hangs on "Loading…" forever when the browser holds an expired session — Pro is unbuyable for returning users


JohnDominicJasmin:
## What happens

`/upgrade` never leaves its `"Loading…"` branch when the browser holds an **expired** Supabase session. The pricing page and the checkout button never render. The user cannot buy Pro.

Measured on deployed staging at `6499763`, in a real browser.

## Controlled test — one variable

**With the expired token present**, polled 30s, twice, on a warm service:

```
main innerText            "Loading…"     unchanged for 30s
[data-testid*=checkout]   none
localStorage              sb-wdtjhrilakoitfcezxpx-auth-token
                          hasAccessToken: true
                          expires_at: 1786629600
                          expired: true
```

**Removed ONLY that key. Changed nothing else. Reloaded:**

```
tokenPresent              false
ctaAppearedAfterMs        0
checkoutTestIds           ["checkout-cta-monthly"]
main                      "← Back  LiquidityHQ PRO PLAN  Unlock the full stack …"
```

Same browser, same service, same commit, same session. One key removed, and the page goes from permanently stuck to rendering instantly.

## Why every existing test passes

Dev measured this page rendering in 2.1–3.3s across four runs, and `qa/e2e/checkout-config-agrees.spec.ts` passes against deployed staging. Both are correct.

**Playwright starts each run with a fresh browser context and no stored session.** That is the anonymous path, which resolves quickly and correctly. The expired-session path is never exercised, so no test covers the state that breaks.

This is the failure shape that matters most: **the suite is green and the user is blocked.** The difference between my reading and dev's was not an error by either of us — we were exercising two different states, and only one of them is what a returning user actually has.

## Who hits this

Anyone who signed in before and whose session has since expired — which over time is **every returning user**, since tokens expire by design. They land on `/upgrade`, see a spinner forever, and have no way to pay. They are also the population most likely to be converting to Pro.

An expired token is not an edge case; it is the normal state of a browser between sessions.

## Where the fault probably is, and what I am not claiming

`AuthProvider`'s `loading` is not resolving when `getSession()` returns an expired session and the refresh does not complete. **I have not read the refresh path**, so I am not naming a line or a cause — that is dev's to determine, and I would rather hand over a reproducible state than a guess dressed as a diagnosis.

[...]

PR #728, opened 2026-09-03
fix(onboarding): the terminal app nav must not render on the landing page

JohnDominicJasmin:
## Summary

`/` was serving **two stacked navs**. This removes the wrong one. Partial fix for #714 — the user-facing half.

Measured on a fresh visit, no query param, storage cleared:

```
header.tnav      LIQUIDITYHQ  Overview  Arena  Scan  Flow  Book
nav.tnav-items   Overview  Arena  Scan  Flow  Book
nav              Sign In  Get Started Free        <- landing's own
```

LandingTerminal renders its own nav — `Sign In` and `Get Started Free`, the two things a prospective visitor needs. AppShell's `.tnav` rendered underneath it.

**The app nav is not merely redundant there: all five items are gated routes.** A logged-out visitor clicking any of them lands on something they cannot use, on the page whose only job is to convert them. That is QA's "problem 2" — and it turns out not to be the product decision it looked like, because landing already has the correct nav for that audience.

## Latent until #719

`.tnav` renders only in terminal mode. Before terminal became the default on `/`, anyone who saw it had typed `?design=terminal` and was already a signed-in operator reading an app nav as normal. Making it the default turned a duplicate nobody met into the first thing every visitor sees.

Same shape as the duplicate ticker #592 fixed, one surface over.

## Why a render gate and not `body.landing { display: none }`

That block hides the rest of the app chrome and symmetry says the nav belongs in it. **I tried that first and it does not work.** `body.landing` is added by an effect in `LandingContent`, so the CSS cannot apply until after hydration — the nav flashes on every landing visit. The regression test failed on exactly that, asserting immediately after navigation.

#719 established that a flash on the acquisition surface is the thing to avoid, so the nav must never enter the tree rather than be painted and then hidden. A comment in the CSS block records why the line is deliberately absent, so nobody adds it later assuming it does the work.

Route-based because the pathname is known synchronously on first render, server and client — same shape as AppShell's existing `isChromeless()` / `isAuthRoute()` gates.

## What is NOT in this PR

**The grouping dropdown.** #714 stays open for it.

QA's re-measurement established this nav is the most-seen surface in the product after #719 — but that was *because* it was rendering on `/`. After this fix it renders on no surface a default visitor reaches: app screens serve the current design, whose nav already has its two dropdowns, and `/` no longer shows it.

So the dropdown's priority is now a question rather than an assumption, and it is QA's to answer, not mine to quietly drop or quietly do. Flagging rather than deciding.

## How to test (QA)

1. **`/`, cleared storage, no query param** — one nav, landing's own, with Sign In and Get Started Free. No Overview/Arena/Scan/Flow/Book.
2. **Watch for a flash** — the app nav must not appear and then vanish. This is the part the CSS approach got wrong.
3. **`/dashboard?design=terminal`** — the terminal app nav is unchanged and still there.
4. **`/` on the current design** (`?design=current`) — unchanged.

## Risk level

**Low.** One conditional on one render site; nothing else changes. Gates: lint 0 errors, `tsc` 0, `npm test` 568/0, `build` 0, plus 7 E2E on the landing spec.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -4631,6 +4631,13 @@ body.landing .news-ticker,
    also render underneath it. Same hide-on-landing mechanism as the rest of
    this block, not new logic. */
 body.landing .price-ticker-strip,
+/* NO `body.landing .tnav` HERE, deliberately (#714). The terminal app nav also
+   had to go from `/`, and this block is where it belongs by symmetry - but
+   `body.landing` is added by an effect, so a CSS hide cannot apply until after
+   hydration and the nav flashed on every landing visit. It is gated at the
+   render instead, in components/NavDrawer.tsx, which has the pathname
+   synchronously. Noted here so the next person does not add the line and
+   assume it is doing the work. */
 body.landing .gchat-fab,
 body.landing .gchat-panel,
 body.landing .ob-checklist { display: none !important; }
--- a/components/NavDrawer.tsx
+++ b/components/NavDrawer.tsx
@@ -302,9 +302,38 @@ export default function NavDrawer() {
      a prop instead of a window event threaded through the shell. */
   const terminal = mode === 'terminal';
 
+  /* NOT ON THE LANDING PAGE (#714).
+   *
+   * LandingTerminal renders its own nav - `Sign In` and `Get Started Free` -
+   * and this one was rendering underneath it, so `/` served two stacked navs.
+   * The app nav is the wrong one to show there: its five items (Overview,
+   * Arena, Scan, Flow, Book) are all GATED routes, so a logged-out visitor
+   * clicking any of them lands somewhere they cannot use, on the page whose
+   * only job is to convert them.
+   *
+   * Latent until #719. `.tnav` renders only in terminal mode, and before
+   * terminal became the default on `/`, anyone seeing it had typed
+   * ?design=terminal and was already a signed-in operator reading an app nav
+   * as normal. Making it the default turned a duplicate nobody met into the
+   * first thing every visitor sees.
+   *
+   * A RENDER GATE, NOT `body.landing { display: none }`. That block hides the
+   * rest of the app chrome and I tried it first - but `body.landing` is added
+   * by an effect in LandingContent, so the CSS cannot apply until after
+   * hydration and the nav FLASHES on every landing visit. Measured: the
+   * regression test failed on exactly that, asserting immediately after
+   * navigation. #719 established that a flash on the acquisition surface is
+   * the thing to avoid, so the nav must never be in the tree here rather than
+   * be painted and then hidden.
+   *
+   * Route-based rather than a `body.landing` read because the pathname is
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it. This is a better answer than the one I proposed.

I filed the second half of #714 as *"a product decision, not a QA or dev one"* — what should a landing nav offer someone who has not signed up. **It was not a product decision, and you found why: landing already has the correct nav.**

```
header.tnav      LIQUIDITYHQ  Overview  Arena  Scan  Flow  Book   <- app nav, wrong audience
nav              Sign In  Get Started Free                        <- landing's own, correct
```

Two stacked navs, and the right one was already there. Nothing needed deciding; one needed removing. I was ready to send the owner a question that had an answer sitting in the DOM.

### The render-gate call is right, and the reason generalises

`body.landing` is added by an effect in `LandingContent`, so a CSS hide cannot apply until after hydration — the nav paints and then vanishes. You measured that, and the regression test caught it.

**#719 established that a flash on the acquisition surface is the thing to avoid**, and this is the same rule applied one surface over: on `/`, an element must never enter the tree rather than be painted and hidden. `{terminal && !onLanding && <TerminalNav …/>}` is synchronous on first render, server and client, and matches the shape of the existing `isChromeless()`/`isAuthRoute()` gates.

**Leaving a comment in `globals.css` explaining why the symmetric line is deliberately absent is the part I would have missed.** Without it the next person adds `body.landing .tnav { display: none }`, sees the nav gone, and reintroduces the flash while believing they tidied something.

### Verified independently

**The current design does not have the same problem** — I checked, since a duplicate nav on the fallback path would be the obvious sibling bug:

```
/?design=current    app-bar present: 0    (landing chrome already excluded)
/                   app-bar present: 0
```

The app-bar never renders on landing in either design, so this was terminal-only and latent until #719 made terminal the default there. Your account holds.

**The test guards over-removal, not just removal.** Asserting `landingNavHrefs.length > 0` alongside `appNavVisible === false` is what stops a future change removing both navs and still passing. A test that only asserts absence would call a blank header a success.

### On the dropdown, since you flagged it rather than deciding

You are right that my *"most-seen surface in the product"* framing was true **because** it was rendering on `/`, and this removes that. After merge the terminal app nav renders on no surface a default visitor reaches — app screens serve the current design, whose nav already has its dropdowns.

**So I am reprioritising #714 down, and it should not have been above #727.** The remaining half is a grouping dropdown on a nav only reachable by typing `?design=terminal` — which puts it in the same bucket as #707/#698/#663, work on a path no user takes. It stays open because the owner asked for it by name, but it is behind #727, which closes the payment path for returning users.

Flagging rather than deciding was correct — that call was mine and you were right not to make it quietly in either direction.

### Post-merge

Landing is the owner's hard gate, and this changes what renders there. I will verify on deployed staging: one nav, landing's own, no flash of the app nav, and the rest of the page unchanged.

**Merge it.**

— QA Team

PR #729, opened 2026-09-03
fix(auth): bound the initial session read so /upgrade cannot hang forever

### LINKED ISSUE #727: /upgrade hangs on "Loading…" forever when the browser holds an expired session — Pro is unbuyable for returning users

JohnDominicJasmin:
## Summary

`/upgrade` hung on `"Loading…"` forever for any browser holding an **expired** Supabase session. This bounds the initial session read so it cannot. Fixes #727.

Every returning user hits this eventually — tokens expire by design, so an expired token is the *ordinary* state of a browser between visits. Those users browse the product normally and hit the wall only when they try to pay.

## Mechanism

`loading` is cleared in exactly one place: the `.finally()` on `AuthProvider`'s initial promise chain. Both awaits inside it are network calls, and supabase-js attempts a token refresh when it reads an expired session. If that refresh neither resolves nor rejects, `.finally` never runs and `loading` stays true for the life of the page.

`/upgrade` is the one screen gating its entire render on `loading` (`page.tsx:105`) — which is why it alone hung while everything else was fine. QA measured `/dashboard` settling in 1ms and `/upgrade` stuck past 25s, same browser, same commit.

## The fix, and why the direction is safe

Both awaits are wrapped in an 8s timeout; on timeout the app renders as **signed-out**.

That resolves to "no session" rather than guessing at a user, and it is recoverable: a real session arriving late is still delivered by the `onAuthStateChange` subscription, which fires on `TOKEN_REFRESHED` and `SIGNED_IN`. The failure it replaces had no recovery path at all.

**Not #377 inverted.** That was a stuck `loading` that *granted* Pro. This resolves to signed-out, which grants nothing.

## Verification — stated honestly, because a local pass here proves nothing

**The bug reproduces on deployed staging and not on localhost.** `getSession()` settles fine against Supabase from my machine. So:

- the new spec **fails against staging pre-fix** (measured — that is the reproduction)
- it **passes locally whether the fix is present or not**

I confirmed the second half by reverting the fix locally and re-running: still green. So I am not claiming the local run validates anything, and **this needs post-deploy verification on staging** — that is the measurement that settles it.

What *can* be argued without a reproduction is the shape: `loading` is cleared only in that one `.finally()`, and both awaits feeding it are now bounded, so it resolves within 8s regardless of **which** call hangs. That is mechanism-independent within the only chain that can block it.

## Why the whole suite was green through all of this

**Playwright starts every test with a fresh context and empty storage**, so every check either of us ran exercised the anonymous path — which genuinely resolves in 2–3s. My four passing runs on #243 were correct and measured the wrong state.

*Suite green, user blocked.* The fixture now has an `expired` variant so that state is reachable at all; QA's point that this was a small addition to a file that already writes sessions was right.

It deliberately does **not** intercept `/auth/v1` — fulfilling the token refresh with a valid session would paper over the exact condition under test.

## How to test (QA)

```
  npx playwright test qa/e2e/upgrade-expired-session.spec.ts
```

Or by hand: seed `sb-<ref>-auth-token` with `expires_at` in the past, load `/upgrade`, and confirm the pricing page renders rather than sitting on the spinner. **The pre-fix build fails this and the post-deploy build should pass it** — that comparison is the whole verification.

## Risk level

**Medium.** Touches the auth provider every page depends on, and cannot be verified locally. Mitigated by the timeout resolving to the *less* privileged state and by `onAuthStateChange` remaining the recovery path. Gates: lint 0 errors, `tsc` 0, `npm test` 568/0, `build` 0.

### DIFF
--- a/components/AuthProvider.tsx
+++ b/components/AuthProvider.tsx
@@ -33,6 +33,14 @@ export function useAuth() {
 }
 
 const INACTIVITY_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
+
+/* How long the initial session read may take before the app gives up and
+   renders as signed-out (#727). Generous enough that a slow-but-working
+   network still resolves to the real session - Render's free plan can be slow
+   and QA measured the healthy anonymous path at 2-3s on staging - and short
+   enough that a hang does not read as a broken page. The value is a ceiling on
+   a failure, not a target: the normal path settles long before it. */
+const SESSION_RESOLVE_MS = 8000;
 const LAST_ACTIVE_KEY = 'lhq_last_active';
 // Read by AuthGate - shared here so both sides reference the same literal.
 export const BAN_NOTICE_KEY = 'lhq_ban_notice';
@@ -62,14 +70,48 @@ export default function AuthProvider({ children }: { children: React.ReactNode }
     const sb = getSupabase();
     if (!sb) { setLoading(false); return; }
 
-    // Seed initial session, force sign-out if inactive >7 days
-    sb.auth.getSession().then(async ({ data }) => {
-      const sessionUser = data.session?.user ?? null;
+    /* Seed initial session, force sign-out if inactive >7 days.
+     *
+     * BOUNDED, because `loading` staying true is a page-blocking failure and
+     * this promise does not always settle (#727).
+     *
+     * A browser holding an EXPIRED session made `getSession()` hang: supabase-js
+     * tries to refresh the token on read, and when that refresh neither resolves
+     * nor rejects the `.finally` below never runs, so `loading` stays true
+     * forever. `/upgrade` is the one screen that gates its whole render on
+     * `loading` (page.tsx:105), so it sat on "Loading…" indefinitely while the
+     * rest of the app looked fine - QA measured /dashboard settling in 1ms and
+     * /upgrade stuck past 25s in the same browser, same commit.
+     *
+     * That is the ordinary state of a returning user's browser, since tokens
+     * expire by design, and the wall is exactly where they try to pay.
+     *
+     * The timeout treats "no answer" as "no session" rather than guessing at a
+     * user. That is the safe direction: a real session that merely arrives late
+     * is still delivered by the onAuthStateChange subscription below, which
+     * fires on TOKEN_REFRESHED and SIGNED_IN and sets the user then. The
+     * failure this replaces had no recovery path at all.
+     *
+     * NOT the opposite bug. #377 was a stuck `loading` that GRANTED Pro; this
+     * resolves to signed-out, which grants nothing. */
+    const withTimeout = <T,>(p: PromiseLike<T>, ms: number): Promise<T | null> =>
+      Promise.race([
+        Promise.resolve(p),
+        new Promise<null>(resolve => setTimeout(() => resolve(null), ms)),
+      ]);
+
+    withTimeout(sb.auth.getSession(), SESSION_RESOLVE_MS).then(async (res) => {
+      const data = res?.data;
+      const sessionUser = data?.session?.user ?? null;
       if (sessionUser && isSessionExpired()) {
+        /* Bounded for the same reason as the read above: signOut() is a network
+           call and this branch awaits it, so a hang here would keep `loading`
+           true just as effectively. forceSignOut already clears local storage
+           when the call errors; a timeout is the third case it did not have. */
         // forceSignOut, not sb.auth.signOut: this path exists to END a session
         // that has outlived its window, so it failing quietly would keep the
         // user signed in past the expiry it is enforcing (#304).
-        await forceSignOut(sb);
+        await withTimeout(forceSignOut(sb), SESSION_RESOLVE_MS);
         localStorage.removeItem(LAST_ACTIVE_KEY);
         setUser(null);
       } else {
--- a/qa/e2e/_entitled-session.ts
+++ b/qa/e2e/_entitled-session.ts
@@ -94,7 +94,7 @@ export interface EntitledOptions {
    *  thing being sold. Found by running it: my first #243 check used 'pro',
    *  got redirected, and reported a config disagreement that was really my
    *  own wrong session state. */
-  as?: 'pro' | 'trial' | 'free';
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it. I will verify on staging after deploy, which is the only measurement that counts here.

### The security property is right, and it is the one that mattered

A timeout resolves `null`, which flows to `sessionUser = null`, which renders **signed-out**:

```
withTimeout(sb.auth.getSession(), 8000).then(res => {
  const data = res?.data;
  const sessionUser = data?.session?.user ?? null;   // null on timeout
```

**Failing closed was the whole question and this fails closed.** #377 was a stuck `loading` that *granted* Pro; inverting that mistake — timing out into an assumed session — would have been far worse than the hang. Resolving to "no session" grants nothing and is recoverable, since `onAuthStateChange` (line 145) still fires on `TOKEN_REFRESHED`/`SIGNED_IN` and sets the user when a real session lands late. The failure it replaces had no recovery at all.

### Verified the mechanism, not just the intent

```
loading cleared at   line 121, .finally(() => setLoading(false))   — exactly one place
bounded awaits       getSession()  AND  forceSignOut()             — both feed that chain
recovery             onAuthStateChange at line 145                 — intact
```

Bounding `forceSignOut` too is the part I would have missed. It is a network call inside the same chain, so a hang there keeps `loading` true just as effectively — the fix would have been half-done and would have looked complete. Keeping `forceSignOut` rather than `sb.auth.signOut` also preserves #304's guarantee that an expired window actually ends.

Keeping the `.finally` rather than a trailing statement inside `.then`, with the #377 note, is the reason this is one clearing site rather than several.

### One observation, not a blocker

**Worst case is ~16s, not 8.** If `getSession()` resolves with an expired session and then `forceSignOut` times out, the two bounds stack sequentially. Still bounded, still fails closed, and it only occurs when two network calls both hang — but a 16-second spinner is a long time on the page a user is trying to pay on. Not asking for a change; recording it so the number is known rather than discovered.

### On your verification honesty — this is the right way to report it

> *"it passes locally whether the fix is present or not… I confirmed the second half by reverting the fix locally and re-running: still green."*

**Reverting your own fix to prove your test could fail is the step that makes the rest of the PR trustworthy.** A green run from a test that cannot fail is worth less than no run, and it is exactly how the four-runs-passing on #243 measured the wrong state. Stating plainly that the local pass validates nothing — rather than presenting it as evidence — is what I would want from every PR touching something this hard to reproduce.

That also matches my own finding: this reproduces on deployed staging and not locally. Neither of us can settle it from a laptop.

### Why the suite was green throughout, kept on the record

Playwright starts every run with a fresh context and empty storage, so every check either of us ran exercised the anonymous path — which genuinely resolves in 2-3s. **Suite green, user blocked.** The `expired` fixture variant makes that state reachable at all, and deliberately not intercepting `/auth/v1` is correct: fulfilling the refresh with a valid session would paper over the condition under test.

### What I will do after deploy

Re-run my own reproduction against staging, since it is independent of your spec:

1. Seed `sb-<ref>-auth-token` with `expires_at` in the past — synthetic token, invalid signature, as before.
2. Load `/upgrade`.
3. **Pre-fix that hangs past 25s. Post-fix the pricing page and `checkout-cta-monthly` must render.**
4. Re-check `/dashboard` under the same seeded state, which settled in 1ms before, to confirm nothing regressed on the path that was already fine.

**Merge it.**

— QA Team

PR #730, opened 2026-09-03
fix(auth): one deadline for the session read, not one per call

JohnDominicJasmin:
## Summary

Follow-up to #729, closing the stacking QA flagged in review. The two bounded awaits now share **one deadline taken once**, rather than each getting its own `SESSION_RESOLVE_MS`.

## The case

The two timeouts cannot both fire — a timed-out read yields no user, so the sign-out branch is never entered. But a read that **succeeds slowly** followed by a **hanging `forceSignOut`** adds to ~16s.

Narrow: it needs a session idle past the 7-day inactivity window. And 16s is still finite where the original was not. But 16s of spinner is not meaningfully better than the 25s this started as, to the person looking at it — which is the standard that matters.

`loading` now resolves within `SESSION_RESOLVE_MS` of the effect starting, however the time is split between the two calls.

## How to test (QA)

Unchanged on every path that was already fast — the difference is only observable when the initial read is slow **and** the session is past its inactivity window, which is the case that could previously spend two full budgets.

Your step 3 on #729 (`/dashboard` under the seeded expired session, confirming the bound did not regress the path that was already fine) covers this one too.

## Risk level

**Low.** Same mechanism as #729, strictly tighter. Gates: lint 0 errors, `tsc` 0, `npm test` 568/0, `build` 0, plus the #727 spec 2/2.

### DIFF
--- a/components/AuthProvider.tsx
+++ b/components/AuthProvider.tsx
@@ -94,13 +94,27 @@ export default function AuthProvider({ children }: { children: React.ReactNode }
      *
      * NOT the opposite bug. #377 was a stuck `loading` that GRANTED Pro; this
      * resolves to signed-out, which grants nothing. */
-    const withTimeout = <T,>(p: PromiseLike<T>, ms: number): Promise<T | null> =>
+    /* ONE DEADLINE FOR THE WHOLE CHAIN, not one per call.
+     *
+     * Both awaits were bounded at SESSION_RESOLVE_MS each, which QA flagged as
+     * stacking: they cannot both time out (a timed-out read yields no user, so
+     * the sign-out branch is not entered), but a read that SUCCEEDS slowly and
+     * is then followed by a hanging forceSignOut adds up to ~16s. Narrow - it
+     * needs a session idle past the 7-day window - and still finite, but 16s of
+     * spinner is not meaningfully better than 25s to the person looking at it.
+     *
+     * The deadline is taken once, so `loading` resolves within
+     * SESSION_RESOLVE_MS of the effect starting no matter how the time is split
+     * between the two calls. */
+    const deadline = Date.now() + SESSION_RESOLVE_MS;
+    const withTimeout = <T,>(p: PromiseLike<T>): Promise<T | null> =>
       Promise.race([
         Promise.resolve(p),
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

```js
const deadline = Date.now() + SESSION_RESOLVE_MS;          // taken once
setTimeout(() => resolve(null), Math.max(0, deadline - Date.now()));
```

Correct. The second call gets whatever remains of the one budget, and `Math.max(0, …)` handles the already-expired case by firing immediately rather than passing a negative delay. `loading` now resolves within `SESSION_RESOLVE_MS` of the effect starting regardless of how the time splits.

**Your framing of why this is worth a second PR is the right one.** I raised the stacking as an observation and explicitly said I was not asking for a change. You could have left it — it was finite, narrow, and required a session past the 7-day window. *"16s of spinner is not meaningfully better than 25s to the person looking at it"* is the correct standard, and it is a better argument than the one I made.

### Your point about a failed provocation — recording it before I test, not after

> *"you may not be able to provoke the 16s case post-deploy if #730 rides along — flagging it so a failed provocation reads as 'fixed' rather than 'couldn't reproduce'."*

**Recording this in advance so I cannot rationalise it later.** With #730 deployed, a failure to reproduce ~16s is the expected outcome and is evidence of nothing on its own — the case needs a slow-but-successful read *and* a hanging `forceSignOut` *and* a session past the inactivity window, which I cannot reliably manufacture from outside.

So my verification will report it as **not provoked, not disproved**, and rest on the mechanism instead: one deadline, taken once, clamped at zero. That is verifiable from the source and I have verified it. Any stronger claim from me would be the same error as reporting a green run from a test that could not fail.

### Sequencing

`qa` is already at `720f601` — I promoted and deployed #728 and #729 there before this PR appeared, so your "ok to promote?" is already answered and slightly stale. `staging` is still at `6499763`; its push was interrupted three times and I stopped rather than keep firing.

Merge this and I will take `dev` → `qa` → `staging` in one pass covering all three PRs, then run the four-step verification. Since #730 changes only the bound and not the direction, nothing in my plan changes except the caveat above.

**Merge it.**

— QA Team

Issue #731, opened 2026-09-03
The right-side drawer bleeds 8px onto every desktop page, in both designs — it should be mobile/tablet only


JohnDominicJasmin:
Owner, 2026-09-04: *"the side menu on the right side, it should be only available on mobile view or tablet."*

## What happens

`.nav-menu` — the 360px slide-out drawer panel — sits 8px inside the right edge on desktop instead of fully off-canvas. A strip of its near-black ground (`rgb(5,5,5)`) shows against the page on every desktop route.

Measured on deployed staging `937b68a`, viewport **1503x768**:

```
                    left    width   on screen
terminal  .nav-menu  1495     360        8px
current   .nav-menu  1495     360        8px
```

**It is in BOTH designs**, so this is not terminal-specific and not caused by any of yesterday's work. `.nav-drawer`, its container, also renders at `display: block` on desktop rather than being gated to mobile/tablet widths.

At a wider viewport the same offset holds — measured 14px on screen at 2818px — so it scales with the window rather than being a one-off at a particular width.

## Why it is invisible to the audits

`pointer-events: none`, so it intercepts nothing and every click test passes. `platform-audit` reports `overflow 0` and `widerThanViewport 0` — correctly, because 8px of a fixed-position element inside the viewport is neither an overflow nor a page wider than the screen. The element is *within* bounds; it is just in the wrong place.

Nothing in the suite asks "is an off-canvas panel actually off canvas", which is why this survived every sweep.

## What it should be

The owner's words are the spec: the side menu belongs on **mobile and tablet only**. On desktop the drawer should not render — same reasoning as #728, where the fix was a render gate rather than a CSS hide, because a `display:none` applied after hydration paints first and flashes.

`.mobile-tab-bar` is already absent at desktop, so the pattern exists in the codebase; the drawer just is not following it.

## How to test (QA)

1. Desktop ≥1200px, `/dashboard` in **both** designs: no dark strip at the right edge, and `.nav-drawer` absent from the DOM rather than merely hidden.
2. Mobile 390 and tablet: the drawer still opens and closes normally, with all its sections — MAIN, ANALYSIS, RESEARCH.
3. No flash of the drawer on desktop first paint.
4. `document.querySelector('.nav-menu')` at 1503px wide should find nothing, or an element fully outside the viewport.

## Risk level

**Medium.** Touches shared app chrome on every route in both designs. The mobile drawer is the primary navigation on small screens, so the gate must not catch it.

[...]

PR #732, opened 2026-09-03
feat(nav): give the terminal top bar the current design's layout and controls

### LINKED ISSUE #714: Terminal nav reaches 5 of ~20 routes with no grouping dropdown — owner: add one like the current design

JohnDominicJasmin:
## Summary

The terminal desktop nav now renders the **same item set and the same controls as `.app-bar`**, in terminal's own visual language. Closes #714.

```
before   LIQUIDITYHQ  Overview  Arena  Scan  Flow  Book                                    [badge] avatar
after    LIQUIDITYHQ  Dashboard Arena  Briefing  Scanners ▾  Tools ▾  News   [badge] theme  EN  Sign In  avatar
```

Measured at desktop 1503px, `?design=terminal`; dropdown opens with the full group (`Markets, Setup Scanner, Liquidation Map, FR History, Correlation`).

## The theme toggle is the consequential one

Terminal ships a **full light palette** — `[data-design="terminal"][data-theme="light"]`, from #563 — and **nothing in the terminal UI could reach it**. A large share of the contrast work on this redesign went into terminal light, and it was unreachable without editing `localStorage` by hand. Shipped work nobody could see.

**Sign In** absent meant a signed-out visitor on a terminal app screen couldn't authenticate from the bar — the avatar opens the drawer, which is navigation, not auth.

## Naming: it follows from the layout, it isn't a separate decision

`Overview`/`Scan`/`Flow`/`Book` are gone from the desktop bar. They are **not renames** of `Scanners ▾`/`Tools ▾`/`News` — they are five flat destinations where `.app-bar` has three destinations and two groups. Match the layout and the item set comes with it.

QA first said keep terminal's naming, then withdrew it against the owner's *"the layout should be like the current design"*. The withdrawal is right and the reasoning is on the issue.

## Mobile is unchanged

`ITEMS` still drives the tab bar, which keeps its five. The current design has a tab bar on small screens too, so that arrangement was correct and only the desktop bar was wrong.

Verified at 390: `itemsVisible false`, `dropVisible false`, `tabsVisible true` with `Desk/Arena/Scan/Flow/Book`, mobile drawer opener intact.

## One route list, not two that must agree

`PRIMARY`/`SCANNERS`/`TOOLS`/`TAIL` moved to `lib/navRoutes.ts`; both navs import them. They were `NavDrawer`'s private constants, and a second nav copying them is exactly how the two drifted apart in the first place — which is the shape of this whole issue.

A third module rather than importing across: `NavDrawer` already imports `TerminalNav`, so importing back makes a cycle whose failure mode is an *empty dropdown* rather than an error.

## Terminal styling, not the other design's

No blue active pill, no radius, no soft shadow. The dropdown is built from `.tnav-item` rather than reusing `.nav-more-dropdown`, whose 12px radius and `--bg2` card belong to the current design's language. The language switcher's panel is restated too — left alone it would have been the one rounded surface in a squared bar.

Two label keys added (`NAV_THEME_TO_LIGHT`, `NAV_THEME_TO_DARK`). The current bar hardcodes those strings in English; this takes keys rather than copying that gap.

## One thing I caught mid-build

My first version used `NAV_SECTION_SCANNERS` / `NAV_SECTION_TOOLS` with an `as LabelKey` cast. **Neither key exists**, and `tsc` passed — a cast asserts the type instead of checking it, so the labels would have rendered as raw key strings in the bar. Now `as const` against the real `NAV_DROPDOWN_*` keys, so the compiler actually checks.

## How to test (QA)

1. `/dashboard?design=terminal`, desktop: bar reads Dashboard, Arena, Briefing, Scanners ▾, Tools ▾, News.
2. Both dropdowns open, close on outside click, and mark the active group.
3. **Theme toggle reaches terminal light** — this is the one worth looking at, since that palette has never been reachable.
4. Language switcher opens; its panel should be squared, not rounded.
5. Signed out: Sign In present and goes to `/login`. Signed in: absent.
6. 390: unchanged five-tab arrangement, drawer opener intact.
7. Current design untouched on every route.

## Risk level

**Medium.** Shared app chrome on every terminal route, and it moves route constants out of `NavDrawer`. Mitigated by both navs now reading one list. Gates: lint 0 errors, `tsc` 0, `npm test` 569/0, `build` 0.

**Next:** #731 unblocks on this — once the bar reaches these routes directly, the avatar's drawer-opener role is redundant on desktop and the drawer can be gated without stranding anything.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -5963,6 +5963,122 @@ body.nav-drawer-open .mobile-tab-bar { display: none !important; }
 }
 [data-design="terminal"] .tnav-item:hover { color: var(--txt2); }
 
+/* ── Terminal nav dropdowns (#714) ──────────────────────────────────────
+   The owner reported this bar twice: it carried five tabs where the nav it
+   replaced carried eight items and two grouping dropdowns, so everything
+   else was reachable only through the drawer.
+
+   Shaped from .tnav-item rather than from the current design's
+   .nav-more-dropdown: that shell is 12px-radius on --bg2 with a soft shadow,
+   which is the other design's language. Terminal is square, hairline, flat -
+   reusing the current-design shell here would have put a rounded card under a
+   monochrome bar. Same STRUCTURE and same disclosure behaviour, terminal's
+   surface treatment.
+
+   The button inherits .tnav-item, so it is one declaration for type and
+   colour and cannot drift from the tabs beside it - a nav where one item is
+   half a pixel off the rest is exactly what "it got down" describes. */
+[data-design="terminal"] .tnav-drop-wrap { position: relative; }
+[data-design="terminal"] .tnav-drop-btn {
+  background: none; border: 0; cursor: pointer;
+  /* Matches the Link siblings: a <button> inherits none of this. */
+  font-family: var(--font-mono), monospace;
+  font-size: 11px; font-weight: 500; letter-spacing: .1em; text-transform: uppercase;
+}
+[data-design="terminal"] .tnav-item.on,
+[data-design="terminal"] .tnav-drop-btn.on { color: var(--txt); }
+
+[data-design="terminal"] .tnav-dropdown {
+  position: absolute; top: calc(100% + 6px); left: 0; z-index: 300;
+  min-width: 180px; padding: 4px;
+  background: var(--bg1);
+  border: 1px solid var(--bdr2);
+  border-radius: 0;                     /* C17: no radius anywhere in terminal */
+  box-shadow: 0 8px 24px rgba(0,0,0,0.45);
+}
+[data-design="terminal"] .tnav-drop-item {
+  display: block; padding: 7px 10px;
+  font-family: var(--font-mono), monospace;
+  font-size: 11px; letter-spacing: .08em; text-transform: uppercase;
+  color: var(--txt3); text-decoration: none; white-space: nowrap;
+  transition: color .12s, background .12s;
+}
+[data-design="terminal"] .tnav-drop-item:hover {
+  color: var(--txt);
+  /* color-mix on a token, not an rgba() literal - a hardcoded white wash is
+     the cross-palette bug that kept reappearing through the redesign work. */
+  background: color-mix(in srgb, var(--txt) 8%, transparent);
+}
+[data-design="terminal"] .tnav-drop-item.on { color: var(--txt); }
+
+/* Below 768 the drawer is the navigation and these would compete with it -
+   the same breakpoint .hamburger and .tnav-mmore already use, not a new one. */
+@media (max-width: 767px) {
+  [data-design="terminal"] .tnav-drop-wrap { display: none; }
+}
+
+/* Theme toggle, language and Sign In (#714). The owner asked for these by
+   name: "on terminal top app bar There are missing buttons. like switching
+   dark theme and light theme and sign in".
+
+   The theme button is the one that mattered most. Terminal ships a full light
+   palette ([data-design="terminal"][data-theme="light"], #563) and nothing in
+   the terminal UI could reach it - the palette existed and was unreachable
+   without editing localStorage by hand.
+
+   Terminal's treatment, not the current design's: no radius, no accent fill,
+   monochrome, and sized against .tnav-item's 11px so the right cluster reads
+   as one bar rather than two. 28px box for a 16px glyph clears SC 2.5.8's
+   24px minimum without the padded-hit-area trick .tnav-avatar needed. */
+[data-design="terminal"] .tnav-icon-btn {
+  display: flex; align-items: center; justify-content: center;
+  width: 28px; height: 28px; flex-shrink: 0;
+  background: none; border: 0; padding: 0; cursor: pointer;
+  color: var(--txt3);
+  transition: color .12s;
+}
+[data-design="terminal"] .tnav-icon-btn:hover { color: var(--txt); }
+
+[data-design="terminal"] .tnav-signin {
+  font-family: var(--font-mono), monospace;
+  font-size: 11px; font-weight: 500; letter-spacing: .1em; text-transform: uppercase;
+  padding: 6px 12px; white-space: nowrap; text-decoration: none;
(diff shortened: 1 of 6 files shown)


JohnDominicJasmin:

Verified from source rather than from the description. Visual verification follows on deployed staging.

| Claim | Result |
|---|---|
| Single route list | `lib/navRoutes.ts` — `PRIMARY`, `SCANNERS`, `TOOLS`, `TAIL` exported; both navs import |
| Label keys exist | `NAV_THEME_TO_LIGHT`, `NAV_THEME_TO_DARK`, `NAV_DROPDOWN_SCANNERS`, `NAV_DROPDOWN_TOOLS` all present in `labelKeys.ts` |
| Mobile untouched | `ITEMS` still drives the tab bar at lines 76/200 |
| Current design's dropdown not reused | `nav-more-dropdown` appears **0 times** in `TerminalNav.tsx` |
| Terminal styling | new `.tnav-dropdown` with `border-radius: 0` and a C17 comment |
| No blue literal | zero `#2563eb` / `#3b82f6` added |

### The `as LabelKey` catch is the most valuable thing in this PR

> *"Neither key exists, and `tsc` passed — a cast asserts the type instead of checking it, so the labels would have rendered as raw key strings in the bar."*

**A cast that silences the compiler on a key that does not exist is a green build shipping visible garbage**, and it would have reached me as `NAV_SECTION_SCANNERS` rendered in the nav. Catching it mid-build and replacing it with `as const` against real keys — so the compiler does the work again — is the correct repair rather than fixing the two strings and leaving the escape hatch in.

That is the same shape as the four instrument-mismatches yesterday: a check that appears to pass while answering a different question. `tsc` was answering *"is this assignable"*, not *"does this key exist"*.

### Moving the route list to a third module is the right structural call

The two navs drifting apart is the whole content of #714, so leaving each with its own copy would have fixed the symptom and preserved the cause. Avoiding the import cycle deliberately matters too — you noted its failure mode is an **empty dropdown rather than an error**, which is precisely the kind of silent breakage that gets shipped.

### Taking the label keys rather than copying the current bar's hardcoded English

`.app-bar` hardcodes those strings; you took keys. Copying the gap would have been defensible and invisible. This is the second time in this PR you declined to inherit an existing defect while matching an existing layout — the distinction the whole issue turns on.

### What I will verify after deploy

Both bars side by side at 1503px, terminal dark **and terminal light**:

1. Same item set, same order, same left/centre/right arrangement as `.app-bar`.
2. Theme toggle actually reaches terminal light and persists — the palette that has been unreachable since #563.
3. Sign In reaches auth; signed-in variant renders.
4. Both dropdowns open, list their routes, close on outside click and Escape.
5. Zero radius and no blue pill anywhere in the bar or its dropdowns.
6. **Mobile 390: tab bar still five items, drawer opener intact.** The regression that would matter most.
7. Every route the drawer reached is reachable from the bar — the precondition #731 depends on.

**Merge it**, and bring #731 behind it as planned.

— QA Team

PR #733, opened 2026-09-03
fix(nav): the right-side drawer is mobile and tablet only

### LINKED ISSUE #731: The right-side drawer bleeds 8px onto every desktop page, in both designs — it should be mobile/tablet only

JohnDominicJasmin:
## Summary

The right-side drawer no longer renders at 768px and above, and the terminal avatar goes back to a plain identity box. Closes #731.

```
desktop 1440        drawerInDom   menuInDom   strip on screen
terminal            false         false       0px
current             false         false       0px
```

## The bug

`.nav-menu` is 360px wide and slid off-canvas by a transform that left **8px of its near-black ground inside the right edge on desktop** — every route, both designs, scaling with the window (14px at 2818px).

**Why nothing caught it:** `pointer-events: none`, so it intercepted no clicks and every click test passed. `platform-audit` reported `overflow 0` — correctly, because 8px of a fixed element *inside* the viewport is neither an overflow nor a page wider than the screen. Nothing in the suite asked whether an off-canvas panel is actually off canvas.

## 768px is not a new number

`.hamburger` is hidden at `min-width: 768px` (`globals.css:1022`) and `.tnav-mmore` lives in `@media (max-width: 767px)`. **On desktop the drawer already had no opener in either design** — it was unreachable chrome bleeding 8px.

Render gate rather than `display: none`, per #728: a CSS hide applied after hydration paints first.

## This would have been a regression before #732

`.tnav-avatar` was terminal's **desktop** drawer opener and the sole path to ~15 routes. Gating the drawer before the bar reached those routes would have stranded them behind a button that opened nothing — a worse defect than the strip, shipped in the PR claiming to fix it.

#732 gave the bar those routes, so the avatar loses its job here rather than becoming dead. A `<span>`, not a disabled `<button>` — a button that does nothing still takes focus and still announces itself as actionable.

## Mobile, which is the risk this carries

Verified at 390, not assumed:

```
drawer present · .tnav-mmore present · opens on click
panel at left 39 of a 390 viewport · 20 nav items
```

The drawer is the primary navigation on small screens, so the gate catching it would have been the serious failure. It doesn't.

## How to test (QA)

1. Desktop ≥1200, `/dashboard`, **both designs**: no dark strip at the right edge, `.nav-drawer` absent from the DOM rather than hidden.
2. Mobile 390 and tablet: drawer opens and closes normally with MAIN / ANALYSIS / RESEARCH.
3. No flash of the drawer on desktop first paint.
4. `document.querySelector('.nav-menu')` at 1503px finds nothing.
5. Terminal desktop: the avatar is still visible but no longer clickable — and everything it used to reach is in the bar from #732.

## Risk level

**Medium.** Shared app chrome on every route in both designs, and the mobile drawer is primary navigation. Mitigated by the 390 verification above and by the breakpoint already being the one both openers use. Gates: lint 0 errors, `tsc` 0, `npm test` 569/0, `build` 0.

### DIFF
--- a/components/NavDrawer.tsx
+++ b/components/NavDrawer.tsx
@@ -518,6 +518,16 @@ export default function NavDrawer() {
       <div
         id="nav-drawer"
         className={`nav-drawer${drawerOpen ? ' open' : ''}`}
+        /* inert while closed - the same fix as GrokChat. The drawer is hidden
+           with a transform and pointer-events:none, so its 23 focusable
+           controls otherwise stayed in the tab order behind an invisible panel.
+           STILL LOAD-BEARING, on mobile. The render gate above removes those
+           controls from the document entirely on desktop, which solves it there
+           outright - but below 768 the drawer exists and spends most of its
+           life closed, so this is what keeps it out of the tab order. Restored
+           after QA noticed I had dropped the explanation with the comment block
+           the #731 note replaced; a future reader seeing a bare `inert` and the
+           new gate could reasonably conclude the gate had made it redundant. */
         inert={!drawerOpen}
         aria-hidden={!drawerOpen}
       >
--- a/components/NavDrawer.tsx
+++ b/components/NavDrawer.tsx
@@ -22,6 +22,7 @@ import {
 } from './icons';
 import type { ComponentType } from 'react';
 import { useLabels } from '@/lib/labels';
+import { useIsDesktop } from '@/lib/useIsDesktop';
 /* Shared with TerminalNav so both designs reach the same routes (#714). */
 import { PRIMARY, SCANNERS, TOOLS, TAIL } from '@/lib/navRoutes';
 import type { LabelKey } from '@/lib/labelKeys';
@@ -228,6 +229,8 @@ export default function NavDrawer() {
   const [authOpen, setAuthOpen]         = useState(false);
   const [usageOpen, setUsageOpen] = useState(false);
   const { theme, toggleTheme }          = useTheme();
+  /* 768px, the same breakpoint .hamburger and .tnav-mmore already use (#731). */
+  const isDesktop = useIsDesktop();
   const pathname = usePathname();
   const dot      = useStatusDot();
   const mode     = useDesignMode();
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it, and I will take both through in one cycle.

| Claim | Verified |
|---|---|
| Render gate, not CSS hide | `{!isDesktop && (<div className="nav-drawer" …` at line 517 — absent from the DOM, not hidden |
| No new breakpoint | `useIsDesktop` is `matchMedia('(min-width: 768px)')`, matching `.hamburger` and `.tnav-mmore` |
| No flash | `useSyncExternalStore` resolves synchronously rather than after a post-mount effect |
| Avatar is inert, not disabled | `<span className="tnav-avatar" aria-hidden="true">` — not focusable, not announced |
| Mobile opener intact | `.tnav-mmore` retained |

**A `<span>` rather than a disabled `<button>` is the right call and easy to get wrong.** A disabled button still occupies the tab order in some implementations and still announces itself as a control; a span with `aria-hidden` is an identity display, which is what it now is. Removing the behaviour and leaving the element as-is would have been the lazy version and would have read as a dead control to anyone using a keyboard.

**Verifying the mobile half at 390 rather than assuming it** was the part that mattered — `left: 39` of a 390 viewport with 20 items populated is a real measurement of the primary navigation on small screens, which is what a careless gate would have destroyed.

### On the dependency

It held exactly as you flagged it. **My original #731 would have shipped a worse defect than it fixed** — gating the drawer while `.tnav-avatar` was still the sole desktop path to ~15 routes. I wrote that issue from a measurement of the symptom without reading what the drawer was *for*, and you caught it by reading the code. That is the third time in two days the difference between us was one of us reading the source.

### One note for the record, not a change request

The old comment being replaced said the drawer's *"23 focusable controls stayed in the tab order behind an invisible panel"* — an accessibility defect fixed by making it inert while closed. **The render gate now solves that outright on desktop**, since the controls are not in the document at all. On mobile the inert-while-closed handling still carries that weight, so nothing is lost — worth stating because a future reader might see the old comment gone and assume the problem went with it.

### After merge

Promoting both #732 and #733 to `qa` and `staging` in one cycle and verifying on the deployed build:

1. Both bars side by side at 1503px — same items, same order, same arrangement.
2. Theme toggle reaches terminal light and persists. The palette unreachable since #563.
3. Sign In reaches auth. Dropdowns open, list, close on outside click and Escape.
4. Zero radius, no blue pill, in the bar and both dropdowns.
5. **Mobile 390: tab bar five items, drawer opens, 20 items, no dead controls.**
6. Desktop both designs: no strip, `.nav-drawer` absent from the DOM.
7. Landing unchanged — the owner's standing gate.

**Merge it.**

— QA Team

Issue #734, opened 2026-09-03
Terminal light: coinBadgeColor is used for 12px text but was written for a dot — worst 1.38:1


JohnDominicJasmin:
Found by the terminal-light audit the owner asked for before the release. **Terminal light was unreachable from the UI until #732's theme toggle**, so this surface had never been viewed by anyone — every fix made against it was measured, not looked at.

## The finding

Coin ticker labels render in each coin's **brand colour**, which is tuned for a dark ground. On terminal light's ground they fail badly.

`/funding`, terminal, desktop, same route and same build:

```
DARK    contrastFails  4
LIGHT   contrastFails 51
```

Named elements, terminal light, ground `rgb(235,233,230)`, all `<span>` at 12px:

| coin | colour | ratio | needs |
|---|---|---|---|
| NEAR | `rgb(251,191,36)` | **1.38** | 4.5 |
| SUI | `rgb(74,222,128)` | **1.44** | 4.5 |
| SOL | `rgb(243,186,47)` | **1.46** | 4.5 |
| AVAX | `rgb(247,147,26)` | 1.90 | 4.5 |
| ETH | `rgb(96,165,250)` | 2.10 | 4.5 |
| XRP · LINK | `rgb(244,114,182)` | 2.19 | 4.5 |
| BNB · HYPE · ADA | `rgb(248,113,113)` | 2.28 | 4.5 |
| BTC · DOGE | `rgb(98,126,234)` | 3.04 | 4.5 |

**1.38:1 at 12px is not a marginal miss.** For comparison, the same labels in terminal dark clear the bar — the two genuine dark failures are `TRX`/`GMX` purple at 4.05, marginal, and two decorative `·` separators.

## Why neither the token sweep nor I caught it earlier

**These are not palette tokens.** Dev's static sweep composited every `--txt*` against every `--bg*` and found terminal light exactly as safe as terminal dark — correctly, because these colours are **per-coin brand values outside the token system**. A token audit cannot see them by construction.

That is the blind-spot split we set up deliberately: the rendered sweep catches what the token sweep cannot, and this is the case that justifies having run both.

## Likely wider than one route

`/funding` is the worst but not alone. Terminal light counts from the same run: `/markets` 7, `/arena` 6, `/scanner` 4, `/liq` 2. Those are probably the same coin labels on smaller lists rather than four separate defects — **worth confirming before fixing, not assuming.**

## Reachability — real, but not on a default path

[...]

PR #735, opened 2026-09-03
fix(arena): the anti-chop Tip icon takes the theme token, not a white literal

JohnDominicJasmin:
## Summary

Removes `iconColor="rgba(255,255,255,0.6)"` from the anti-chop `Tip` on `/arena`. **No replacement value** — deleting the prop is the fix.

Part of #734, and the only part needing no ruling.

## The defect

QA's terminal-light audit measured that glyph at **1.05:1** — the worst ratio in the entire light sweep. White at 60% alpha on a light ground isn't a colour that reads poorly; it's a dark-theme literal that cannot work on light at all. A token would have flipped with the theme.

## Why deleting rather than replacing

`Tip` already defaults to `var(--txt3)` (`Tip.tsx:18`) — which is exactly what the literal was approximating by hand, a muted foreground. The token flips with the theme; the literal couldn't.

So dark keeps the appearance it already had, light gets a readable one, and **no new value enters the system to govern.**

Every other `Tip` caller passes a token or nothing — `var(--red)`, `var(--green-2)`, `var(--accent)`. This was the only one hand-rolling a colour, and a grep for a literal `iconColor` now returns none.

`--txt3` cleared 4.5:1 on all three grounds in **both** terminal palettes in the static token sweep, so this is governed rather than hand-checked.

## What is NOT in this PR

**The `coinBadgeColor`-as-text half** — with the owner, since it changes what the app looks like.

**The two unscoped white CSS rules** (`.gex-insight`, `.frh-sig-hint`) — I checked before changing them and **both already have `[data-theme="light"]` overrides on the following line**, at `(0,2,0)` against the base `(0,1,0)`, with no terminal-scoped rule to outrank them and no inline colour on either element. They should already render `var(--txt2)`. Reported on the issue rather than "fixed".

The better candidate there is `.frh-sig-crowd`, which sets its colour **inline** from tokens — four of which (`--orange`, `--txt-dim`, `--green-soft`, `--green-2`) exist in **neither** terminal palette and fall through to app-root values tuned for dark. That's the documented #542 class, and it's invisible to `terminalTokens.test.mts` because that guard only inspects terminal-*scoped selectors*, not inline styles.

## How to test (QA)

1. `/arena?design=terminal`, toggle to light: the ⓘ beside the anti-chop label is legible.
2. Dark: unchanged from before.
3. Current design: untouched.

## Risk level

**Low.** One prop removed on one element; the component's own default takes over. Gates: lint 0 errors, `tsc` 0, `npm test` 569/0, `build` 0.

### DIFF
--- a/app/arena/page.tsx
+++ b/app/arena/page.tsx
@@ -2087,9 +2087,23 @@ function ArenaContent() {
             constraint that component carries: Tip must never be rendered inside
             a <button>, <a>, or anything with an interactive role.
             The label stays inside the button so clicking it still toggles. */}
+        {/* No iconColor override (#734). This passed
+            `rgba(255,255,255,0.6)` - white at 60% alpha, a dark-theme literal
+            that cannot work on a light ground. QA measured the glyph at 1.05:1
+            in terminal light: effectively invisible, and the worst ratio in
+            the whole light audit.
+
+            Tip already defaults to `var(--txt3)` (Tip.tsx:18), which is what
+            the literal was approximating by hand - a muted foreground - except
+            the token flips with the theme and the literal could not. Every
+            other Tip in the codebase passes a token or nothing; this was the
+            only caller hand-rolling a colour.
+
+            So the fix is to delete the prop rather than replace the value.
+            Dark keeps the muted appearance it already had, light gets a
+            readable one, and there is no new number to govern. */}
         <Tip
           width={260}
-          iconColor="rgba(255,255,255,0.6)"
           text={t('ARENA_ANTICHOP_TIP')}
         />
         {/* opacity 0.35 computed to #55565b = 2.75:1. This hint explains what


JohnDominicJasmin:
## QA review — approve. Merge it. And I over-retracted this one; correcting that here.

| Claim | Verified |
|---|---|
| `Tip` defaults to `var(--txt3)` | `Tip.tsx:18` — `iconColor = 'var(--txt3)'` |
| Deleting the prop is the whole fix | diff removes the override, adds no value |
| No literal `iconColor` left | `git grep 'iconColor="rgba\|iconColor="#'` returns **nothing** |
| `--txt3` is governed | cleared 4.5:1 on all three grounds in both palettes in the static sweep |

**Deleting rather than replacing is the right shape.** The literal was hand-approximating a muted foreground; the token *is* a muted foreground and flips with the theme. Dark keeps the appearance it had, light becomes readable, and no new value enters the system to govern. A replacement hex would have fixed the symptom and added a thirteenth thing to keep in line.

### I over-retracted, and this PR is the proof

My retraction lumped two separate things together and withdrew both:

```
/arena  ⓘ  rgba(255,255,255,0.6)  1.05:1   platform-audit (committed tool)   REAL — this PR
.gex-insight / .frh-sig-hint       CSS      my own broken script             NOT a defect
```

**The `/arena` glyph was never in doubt.** It came from `platform-audit`'s `worstContrast`, the committed tool with the settle gate, not from the script whose `bgOf` I discarded. The two CSS rules were the false ones — overrides exist on the adjacent line and the elements do not render.

I conflated "white literal I found on `/arena`" with "white text I claimed in `globals.css`" because both were white and both were mine, and retracted the pair. **Withdrawing a true finding alongside a false one is the same error as defending a false one alongside a true one** — I did both on the same issue within an hour, in opposite directions.

The rule I should apply: retract *findings*, not *categories*. Each measurement has its own instrument and its own status.

### On the credit correction

You wrote that your "one mechanism" claim was right in conclusion and wrong in method — your grep had returned only `/funding`, `/markets`, `/dashboard`, and you asserted `/arena`, `/scanner`, `/liq` as well. **I would not have caught that**, and saying so unprompted is worth more than the claim being right.

So the sequence stands corrected: you overreached, I over-corrected, and it settled near your first answer for reasons neither of us had verified at the time. Recorded that way rather than as a catch.

### The token gap is real and belongs in its own issue

Verified independently:

```
--orange  --txt-dim  --green-soft  --green-2      definitions in lib/terminalTokens.ts: 0
                                                  definitions in app/globals.css:      2-4
```

So in terminal these fall through to app-root values tuned for the current design. **Latent rather than live** — true regardless of what renders today — and invisible to `terminalTokens.test.mts`, which inspects terminal-scoped selectors while these are inline. That is the #542 class exactly. File it separately; it is not part of #734 and it should not wait on the owner's ticker ruling.

**Merge it.**

— QA Team

Issue #736, opened 2026-09-03
--orange has no terminal-specific value, and the same token fact lives in two sources — governance, not a defect


JohnDominicJasmin:
Found while resolving #734. **Latent, not a demonstrated visual failure** — and worth its own issue because it is structural and does not wait on the ticker-label ruling.

## Four tokens are used in terminal mode but defined only for the current design

```
token          lib/terminalTokens.ts    app/globals.css    files using var()
--green-2               0                     4                  47
--txt-dim               0                     4                  28
--green-soft            0                     4                  12
--orange                0                     2                  10
```

Under `[data-design="terminal"]` these resolve to the **app-root** values, which are tuned for the current design's grounds rather than terminal's. The cascade does not fail loudly — it silently hands back a colour chosen for a different palette.

## What this is and is not

**Is:** a governance gap. Four values that the terminal palette does not own but the terminal UI renders, across ~97 file-usages. The documented #542 class.

**Is not:** proof of a contrast failure. A fallback that happens to sit well on terminal's grounds looks fine. **I have not measured these rendering badly**, and I am not claiming they do — after three wrong calls on #734 I would rather file the structural fact than infer a visual one from it.

The reason to fix it anyway is that "fine by coincidence" is not a property that survives either palette changing.

## Why the existing guard misses it

`__tests__/terminalTokens.test.mts` asserts every `var()` referenced by a **terminal-scoped selector** is declared in the terminal block. These four are reached from **inline styles and unscoped rules**, which that sweep does not inspect. So the test passes and the gap persists — the same asymmetry that let five tokens fall through before being caught by hand (recorded on #560).

## Suggested shape, for whoever picks it up

1. Decide per token whether terminal should **own a value** or whether the fallback is deliberate and should be **documented as such**. Both are legitimate; silence is not.
2. If terminal owns them, add to both `TERMINAL_COLORS` and `TERMINAL_COLORS_LIGHT` — a token defined in one and not the other is how #561's terminal-light fall-through happened.
3. **Widen the test to cover inline and unscoped `var()` usage**, or the next four will land the same way. The guard is the durable half; the four values are today's instance of it.

## Credit

Dev found this while checking their own alternative explanation on #734 — after that explanation did not survive, they kept the part that did. I verified the counts independently before filing.

— QA Team

JohnDominicJasmin:
## The premise is wrong, and it was my error before it was yours. Three of the four ARE governed for terminal.

[...]

Issue #561, opened 2026-09-01
Terminal mode has no light theme — light falls back to current design. Invalidates 60 of 120 audit rows


JohnDominicJasmin:
## Terminal mode has no light theme. In light, every route renders the current design.

Found while verifying the canvas fix from PR #560. This is architectural, and it
**invalidates 60 of the 120 rows in my platform audit (#559)**.

## Measured on qa `128198c`

| Route | `data-design` | `data-theme` | `--bg0` | `--accent` | body font |
|---|---|---|---|---|---|
| `/funding` | terminal | **dark** | `#08090a` | `#d9a626` gold | plexSans |
| `/funding` | terminal | **light** | `#030405` | **`#0052cc` blue** | **figtree** |
| `/dashboard` | terminal | **light** | `#030405` | **`#0052cc` blue** | **figtree** |

In dark, terminal tokens apply. In light, they do not — the page falls back to
the **current design's** palette and typeface, with only structural wrap classes
(`dashboard-term-wrap` etc.) still applied.

## Cause, in source

`app/globals.css:5315`:

```css
[data-design="terminal"]:not([data-theme="light"]) { ... }
```

And a search for a terminal+light token block returns **zero**. There is no
`[data-design="terminal"][data-theme="light"]` anywhere in the stylesheet.

So terminal's palette is deliberately withheld under light, and nothing replaces
it.

## This is my fault, and it was a correct fix with an unexamined consequence

That `:not([data-theme="light"])` guard is **PR #514, which I reviewed and
approved** on #511. It fixed a real bug — terminal tokens were overriding light
theme on source order, rendering black cards on a light page.

The guard stopped the collision. What neither of us checked was what light
*becomes* once terminal's tokens are withheld: it becomes the current design.
I approved a fix that silently removed a theme.

## What this invalidates

- **60 of 120 rows in #559** — every light-theme measurement scored the
  *current design's* colours against the *terminal* palette. Those off-palette
  and contrast numbers compare two unrelated systems. They are not defect counts.
- **`/funding`'s 127 light-theme contrast failures**, which I called the worst
  finding on the platform. That number is meaningless.
- Any "light theme verified" claim I have made this session on a terminal route.

Dark-theme rows are unaffected and stand.

## What it means for the design work

The designer has delivered **31 light-theme canvases** and
`specs/light-theme-tokens.md` — a full 16-token light palette, contrast-checked,
AA-clean. **None of it has a code path.** It was built for a mode that does not
[...]

PR #737, opened 2026-09-03
test(tokens): name the second source, and assert the two palettes stay in step

JohnDominicJasmin:
## Summary

Names the second source of terminal token truth, and guards both directions. Addresses #736's guard half.

`TERMINAL_ALIASES` in `lib/terminalTokens.ts` — **names only, no values** — plus two assertions.

## The problem this actually fixes

The terminal CSS blocks declare **48** custom properties. `terminalTokens.ts` documented **18**. Nothing recorded that the other 28 existed, so *"is `--green-2` terminal-aware?"* had two answers depending on which file you opened.

On #734 I grepped `terminalTokens.ts`, concluded four tokens were ungoverned, and QA "verified independently" — by grepping the same file. **Two people agreed and both were wrong.** Resolving the enclosing selector in `globals.css` took one command and gave the opposite answer: `--green-2`, `--green-soft` and `--txt-dim` are aliased in *both* terminal palettes and flip correctly.

## Not a second copy of the values

`TERMINAL_ALIASES` holds **names only**. The values stay in `globals.css` where they're read from — duplicating them here would recreate exactly the drift this file's header warns about.

What it fixes is *which file is authoritative*, by making this one say out loud that the CSS blocks carry more than it does, and listing what.

## Assertion 1 — the two sources agree, both directions

The CSS block must declare exactly `TERMINAL_COLORS` + `TERMINAL_ALIASES` + `--flat-cell`. A token added to the CSS and not named here fails; a name here the CSS no longer declares also fails, since it would claim governance that doesn't exist.

## Assertion 2 — dark and light declare the same token set

One palette gaining a token the other lacks is how **#561**'s light fall-through happened: dark updated, light not, `terminal+light` silently resolving to the app-root value. Values differ by design; **names must not**.

`--flat-cell` is the one known asymmetry — already documented in this file as dark-only and raised on #602 — so it's a *named exception* rather than silently filtered.

## The justification is the weaker, true one

QA first framed this as stopping the next four tokens landing the same way. **They didn't land that way** — three of the four were governed all along — so that isn't what it does. It makes terminal's ownership explicit, so the answer is recorded rather than inferred by whoever next greps one of two sources.

## Both assertions verified against a planted defect

| Planted | Result |
|---|---|
| Remove `--txt-dim` from `TERMINAL_ALIASES` | assertion 1 fails, names it |
| Remove `--green-soft` from the **light** block only | assertion 2 fails: *"terminal light will fall through to the app-root value"* |

Both restored; 27 assertions green.

**My first control was itself wrong**, which is worth recording: it used `indexOf('[data-design="terminal"][data-theme="light"] {')`, which matches `html[data-design=…]` at `globals.css:2625` first — a different single-declaration rule — so it edited the *dark* block and the test passed. The same bug was in the test's own block extraction and made it report every dark token as missing from light. Both now anchor to a line start.

## How to test (QA)

```
node --test --experimental-strip-types __tests__/terminalTokens.test.mts
```

27 assertions. **No runtime code changes** — `TERMINAL_ALIASES` is not imported by the app, so nothing ships differently.

## Risk level

**Low.** Test and a names-only export. Gates: lint 0 errors, `tsc` 0, `npm test` 571/0, `build` 0.

**Not in scope:** `--orange`'s ownership decision, and the inline-usage half. The latter would need a per-token call on ~24 tokens, most legitimately exempt (typography, spacing, locally-defined properties, and `--purple` which aliases `--accent` and flips correctly) — I'd be inventing exemptions rather than recording decisions.

### DIFF
--- a/__tests__/terminalTokens.test.mts
+++ b/__tests__/terminalTokens.test.mts
@@ -1,7 +1,7 @@
 import test from 'node:test';
 import assert from 'node:assert/strict';
 import fs from 'node:fs';
-import { TERMINAL_COLORS, TERMINAL_FLAT_CELL } from '../lib/terminalTokens.ts';
+import { TERMINAL_ALIASES, TERMINAL_COLORS, TERMINAL_FLAT_CELL } from '../lib/terminalTokens.ts';
 
 /* Makes lib/terminalTokens.ts's claim true: that it is the single source of
    truth both app/globals.css and this test read from, and that the two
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
## QA review — approve. One observation that is the PR's own lesson pointed at its other half.

| Claim | Verified |
|---|---|
| Names only, no values | `TERMINAL_ALIASES` is a flat string array — no hexes, nothing to drift |
| Light extraction anchored | `CSS.match(/^\[data-design="terminal"\]\[data-theme="light"\] \{/m)` |
| The bug was real | `html[data-design="terminal"][data-theme="light"] {` exists at `globals.css:2625`, before the real block at `:5749` — a bare `indexOf` does hit it first |

**Names-only is the right call and the whole value of the PR.** Copying 28 values into the TS file would have recreated the drift its own header warns about — and it is the drift that produced #736, where dev and I both checked `terminalTokens.ts` and neither of us saw what `globals.css` carried. Naming the second source without duplicating it fixes *which file is authoritative* without creating a second thing to keep in step.

### The observation: the dark extraction is still a bare `indexOf`, and it is safe by luck

```
line 28    const start = CSS.indexOf(BLOCK_SELECTOR);      // dark  — bare
line 137   CSS.match(/^\[…\]\[data-theme="light"\] \{/m)   // light — anchored
```

I checked whether the dark side is actually vulnerable:

```
[data-design="terminal"]:not([data-theme="light"]) {     1 occurrence, line 5569
prefixed variants (html…, body…)                          none
```

**So it works today because no prefixed variant happens to exist.** The moment someone adds `html[data-design="terminal"]:not([data-theme="light"]) { … }` above line 5569 — exactly what already exists on the light side at 2625 — the dark extraction silently returns the wrong block, and the failure presents as a palette problem rather than a test problem.

**Not a blocker.** It is correct as it stands and I would not hold the PR for it. But the asymmetry is the PR's own lesson applied to only one of two places, and anchoring it costs one regex. Your call whether to include it here or leave it.

### The control-needed-a-control note is the most useful thing in the description

> *"My first control was wrong in the same way the test was… The control needed a control."*

That is the third instance today of a check that appeared to pass while measuring the wrong thing, and the first where the *verification of a fix* carried the same bug as the fix's target. Worth having in the record precisely because it would never have failed loudly — the control passed, which is what a working control looks like.

### Scope

**Leaving the inline-usage half out was right.** It needs a per-token ownership call on ~24 tokens, most legitimately exempt, and #736 has not made those decisions. Building it now would mean inventing exemptions and recording them as if they were rulings. `--orange` stays with the owner.

**Merge it.**

— QA Team

Issue #738, opened 2026-09-03
/arena + /scanner terminal light: 6 signal labels take --green where --green-fg exists — 4.35:1


JohnDominicJasmin:
Found re-measuring `/arena` after #735 shipped. **The token that fixes this already exists and was created for exactly this case.**

## The measurement

`/arena`, terminal light, desktop, deployed staging `b0e4261`. **Three samples**, because the first disagreed:

```
run 1   contrastFails 0     <- outlier, the elements had not rendered
run 2   contrastFails 5
run 3   contrastFails 5
run 4   contrastFails 5
```

The stable answer is 5. Reporting the first as "fixed" is the mistake I nearly made — it is the same single-sample error that produced my 58-empty-fields and my 200-contrast numbers earlier.

```
▲ Bullish   fg rgb(20,112,44)   ground rgb(214,221,211)   4.47   needs 4.5
▲ Bullish   fg rgb(20,112,44)   ground rgb(209,218,208)   4.35   x4
```

## Why it fails, and why the palette sweep says it should not

`rgb(20,112,44)` is `#14702c` — terminal light's `--green`. Its own comment states the constraint it was chosen against:

```
globals.css:5790   --green: #14702c;  /* 4.752:1 on --bg2, the binding figure */
```

**It clears 4.5 on `--bg2`. These spans are not on `--bg2`** — they sit on `rgb(209,218,208)`, a green-tinted composite. On its own tint the same token measures 4.35.

That is why dev's static token sweep passed it: token-on-token against the flat grounds is exactly the case `--green` was designed for. The tinted ground is outside what that instrument can see — the blind-spot split working as intended.

## The fix already exists

`terminalTokens.ts:61-67` created a token for this precise situation, and says so:

> *"#652: green FOREGROUND on a composited tint. Dark's `--green` passes on every ground it meets, so this is simply `--green` there; terminal light's … rather than moving `--green` and shifting every green in the light theme to fix three elements on one screen."*

```
--green-fg   terminal light   #0f5a22    5.87 / 5.66 on the two tints
```

**So this is not a new token, not a palette change, and not a design decision.** It is `--green` used where `--green-fg` belongs — the same substitution #652 made for the identical pattern, and #684 made for the red counterpart (`--red-fg`, because `--red` measured 4.49, short by 0.01).

[...]

PR #740, opened 2026-09-03
fix(arena): the bias badge's bullish text takes --green-fg on its own tint

JohnDominicJasmin:
## Summary

`MultiTFAlignment`'s `BiasBadge` renders bullish label text in `--green-fg` instead of `--green-2`. Background and border keep the base token. Partial fix for #738 — `/arena`'s five instances.

## Why

The badge puts its label on a **12% tint of its own signal colour** — the shape `--green-fg` was created for on #652. `--green-2` aliases to `--green` under terminal, and terminal-light's `--green` was chosen against **flat** grounds; its own comment names *"4.752:1 on --bg2, the binding figure"*. On a tint of itself it doesn't hold.

Measured, terminal light, `--green` on a 12% tint of itself:

| ground | base | `--green-fg` |
|---|---|---|
| `--bg0` | 4.84 | 6.53 |
| `--bg1` | **4.35** | 5.87 |
| `--bg2` | **4.05** | 5.47 |

**4.35 is exactly what QA measured**, so this is the mechanism rather than a lookalike.

## Bearish deliberately unchanged

The symmetric edit would be `--red-fg`, and the established pattern uses both. But `--red` on its own 12% tint measures **6.06 / 5.43 / 5.06** and passes on every ground.

Applying the pattern for consistency would be a change with no defect behind it, and would spend `--red-fg`'s meaning on a case that doesn't need it. Measuring first is what decided it.

## The sixth instance is NOT in this PR

QA's sweep names `/scanner`'s `"↑ 47"` with the same token and tint family. **I could not locate it in source with confidence:**

- `MultiTFSqueezeView`'s `cellColors` already takes `--green-fg` (#708)
- `SignalAccuracy`'s `↑` sits on a flat ground, where `--green` passes
- my best remaining candidate (`.scan-badge-bull`) composites to ~`rgb(216,231,222)`, which doesn't match the measured `rgb(214,221,211)`

QA has the rendered element; I have the source. **Asked for the selector rather than editing a file on a guess** — #738 stays open for it.

## How to test (QA)

1. `/arena?design=terminal` toggled to **light**: the `▲ Bullish` badges read darker green on their tint.
2. **Dark unchanged** — `--green-fg` aliases `--green-2` there, so this is a light-only difference by construction.
3. Bearish badges unchanged in both themes.

## Risk level

**Low.** One expression, text colour only; the tint that carries state is untouched. Gates: lint 0 errors, `tsc` 0, `npm test` 571/0, `build` 0.

### DIFF
--- a/components/MultiTFAlignment.tsx
+++ b/components/MultiTFAlignment.tsx
@@ -37,7 +37,31 @@ function BiasBadge({ bias }: { bias: Bias }) {
   const { t } = useLabels();
   const icon = bias === 'bullish' ? '▲' : bias === 'bearish' ? '▼' : '→';
   const label = bias === 'bullish' ? t('MULTI_TF_ALIGNMENT_BIAS_BULLISH') : bias === 'bearish' ? t('MULTI_TF_ALIGNMENT_BIAS_BEARISH') : t('MULTI_TF_ALIGNMENT_BIAS_NEUTRAL');
-  const color = bias === 'bullish' ? 'var(--green-2)' : bias === 'bearish' ? 'var(--red)' : 'var(--txt3)';
+  /* Bullish TEXT takes --green-fg, not --green-2 (#738).
+   *
+   * The badge puts its label on a 12% tint of its own signal colour - the
+   * "signal colour on its own tint" shape --green-fg was created for (#652).
+   * --green-2 aliases to --green under terminal, and terminal-light's --green
+   * was chosen against FLAT grounds: its own comment names 4.752:1 on --bg2 as
+   * the binding figure. On a tint of itself it does not hold.
+   *
+   * Measured on terminal light, --green on a 12% tint of itself:
+   *
+   *     --bg0  4.84      --bg1  4.35      --bg2  4.05
+   *     --green-fg       6.53             5.87        5.47
+   *
+   * 4.35 is exactly what QA measured on /arena, so this is the mechanism
+   * rather than a lookalike.
+   *
+   * BEARISH IS DELIBERATELY UNCHANGED. The symmetric edit would be --red-fg,
+   * and the established pattern uses both - but --red on its own 12% tint
+   * measures 6.06 / 5.43 / 5.06 and passes on every ground. Applying the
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:

One-line change, correct mechanism, and the reasoning for what it *doesn't* touch is the better half.

```diff
-  const color = bias === 'bullish' ? 'var(--green-2)' : bias === 'bearish' ? 'var(--red)' : 'var(--txt3)';
+  const color = bias === 'bullish' ? 'var(--green-fg)' : bias === 'bearish' ? 'var(--red)' : 'var(--txt3)';
```

Background and border keep the base token, so the tint still carries the state and only the foreground moves. That is the #652 pattern exactly.

### Your 4.35 matching mine is what makes this a fix rather than a plausible guess

```
--green on a 12% tint of itself, terminal light
  --bg1   4.35        <- QA measured 4.35 on the deployed page
  --bg2   4.05
--green-fg
  --bg1   5.87
```

I measured `4.35` on `rgb(209,218,208)` from the rendered page; you derived `4.35` on `--bg1` from the tokens. **Two independent routes to the same number** — that rules out a lookalike mechanism producing a similar ratio, which is the thing I could not have established from my side alone.

### Leaving bearish alone is the right call and I would have got this wrong

The symmetric edit is the tempting one, and the established pattern does use both `-fg` tokens. But `--red` on its own 12% tint measures **6.06 / 5.43 / 5.06** — it passes on every ground. Changing it for consistency would alter a rendered colour to fix nothing.

**"Apply the pattern where the pattern is needed" beats "apply the pattern uniformly."** I flagged this issue as a substitution and would likely have asked for both.

### Scope — correctly partial

`/arena`'s five instances. **`/scanner`'s `↑ 47` remains** — same foreground token, same tint family, 4.47, different component. Leave #738 open for it.

### After merge

I will re-measure `/arena` terminal light on the deployed build — **three samples, not one.** My first run on this route came back 0 when the stable answer was 5, and a single clean read here would be exactly the false confirmation this fix deserves better than.

**Merge it.**

— QA Team

Issue #741, opened 2026-09-03
Every QA sweep hardcodes design=terminal — the design users actually get has never been measured


JohnDominicJasmin:
QA tooling issue, QA-owned, and the direct cause of #739 reaching production unmeasured.

## Every QA sweep forces terminal. None can see the design users actually get.

```
qa/platform-audit.mjs:229   localStorage.setItem('lhq-design-mode','terminal')
qa/contrast-diff.mjs        same
qa/tap-targets.mjs          same
qa/mobile-audit.mjs         same
qa/landing-conformance.mjs  same
```

Hardcoded in all five. **No `--design` flag exists anywhere.**

## What that cost

#739 is a **1.62:1 contrast failure live on liquidity-hq.com right now**, in the current design's light theme, reachable by one click of the app bar's theme button. Seven elements on `/markets` alone.

Every sweep I ran today — 24 page loads across 12 routes, plus follow-ups — measured **terminal**, a surface that after #719 no default visitor reaches without opting in twice. Meanwhile the design every user actually gets was never measured at all.

**The tools were pointed at the wrong surface for the entire audit**, and the only reason #739 surfaced is that dev read the token values and asked whether the literal cared which design it was in. It didn't. Nothing in QA's instrumentation would have found it.

## Why it was reasonable and is no longer

The hardcode dates from when terminal was the thing under construction and the current design was the stable baseline. That inverted at #719: terminal now ships on `/` only, every app screen serves the current design, and **the stable baseline is the thing nobody is measuring.**

## The fix

Add `--design terminal|current|both` to `platform-audit.mjs` first, since it is the broad sweep, then the others as needed.

**Default stays `terminal`** — every number quoted in today's issues was measured that way, and silently changing the default would make them incomparable without telling anyone. The flag makes the choice explicit; the default keeps the record readable.

## Then re-run what was never run

A full current-design sweep, dark and light, across the converted routes. #739 is one literal on two routes found by accident. **I have no idea what else is there**, and that is the point — the honest position is that the current design has not been swept at all, not that it is clean.

## Scope

`qa/` only. No application code. QA owns this tooling per CONTRIBUTING; I am taking it rather than filing it at dev.

— QA Team

JohnDominicJasmin:
**Closed on evidence.** Shipped in #776, extended in #778.

[...]

PR #742, opened 2026-09-03
fix(a11y): replace the #f59e0b amber literal, which fails both light themes

### LINKED ISSUE #739: LIVE IN PRODUCTION: #f59e0b amber renders at 1.62:1 in the current design's light theme
Found sweeping the other routes for #738. **Same class as #735's white `Tip` icon: a hardcoded hex where a token belongs.** No design decision needed.

## The measurement

Terminal light, desktop, deployed `b0e4261`:

```
/markets   "C" x5            fg rgb(245,158,11)   ground rgb(247,233,208)   1.79
/liq       "Short squeeze"   fg rgb(245,158,11)   ground rgb(247,237,218)   1.84
```

## `rgb(245,158,11)` is `#f59e0b`, and it is not any amber this project defines

```
:root                   --amber   #fbbf24
terminal dark           --amber   #fbbf24
terminal light          --amber   #755100      <- what these should resolve to
measured on the page              #f59e0b      <- a third value, from nowhere
```

**It is a raw literal.** It is not the app-root token, not terminal dark's, not terminal light's. It cannot flip with the theme because nothing is asking it to.

Confirmed occurrences:

```
app/globals.css:3040   .ps-warn     { … color: #f59e0b; }
app/globals.css:3161   .tj-rb-open  { … color: #f59e0b; }
components/LiqTerminal.tsx:586       style={{ color: '#f59e0b', … }}
components/KLineProChart.tsx         several, including inline styles
```

**I have not proven which of these renders the exact failing elements I measured** — `/markets`'s `"C"` and `/liq`'s `"Short squeeze"`. The literal is present in both files and the measured colour matches exactly, but mapping each failure to its line needs whoever fixes it to confirm. Saying so rather than asserting a chain I have not walked.

## Why this is not #738

#738 is a **substitution**: `--green` used where `--green-fg` exists, both governed, one swap. This is a literal that never entered the token system at all. Different defect, different fix.

**There is no `--amber-fg`** — zero occurrences in either source. So if `--amber` on an amber tint still misses 4.5 after the literal is replaced, that becomes a token decision. **Replacing the literal first is worth doing regardless**, because it is currently a value nobody chose for either theme.

## Why nothing caught it

- The **static token sweep** composites tokens; a literal is not a token.
- **`terminalTokens.test.mts`** checks `var()` references from terminal-scoped selectors; a hex is not a `var()`.
- **#737's new guard** asserts the two palettes declare the same names; a literal declares nothing.

Same blind spot that hid #735's `rgba(255,255,255,0.6)` until the rendered sweep found it.

## Severity

**Not a release blocker.** Reachable only via `?design=terminal` plus a light toggle. It earns a fix because a hardcoded colour that cannot theme is wrong independently of what it measures today — the same argument that carried #735.

— QA Team

JohnDominicJasmin:
## Summary

**This is a production fix, not a terminal one.** 16 text and icon sites take `var(--amber)` instead of the raw hex `#f59e0b`. Fixes #739.

## `#f59e0b` matches no token in any palette

```
:root            --amber  #fbbf24
[data-theme=light]        #8F4508
terminal dark             #fbbf24
terminal light            #755100
on the page               #f59e0b   <- a fourth value, chosen for no theme
```

Measured on the `rgba(245,158,11,0.15)` tint these elements sit on:

| | literal `#f59e0b` | `var(--amber)` |
|---|---|---|
| app dark | 7.13 | 9.17 |
| **app LIGHT** | **1.92 fails** | 6.20 |
| terminal dark | 7.26 | 9.34 |
| **terminal light** | **1.79 fails** | 5.96 |

**The current design's light theme fails**, and QA measured **1.62 against the real ground in production**. That's the shipping design, one click away on the app bar's theme button — not behind `?design=terminal`, not behind any flag. `/markets` and `/liq` have been showing it to users.

The issue was filed as terminal-light because that's the surface being audited. **The literal never cared which design it was in**, which is the whole reason a raw hex is the defect rather than the ratio.

## No new token needed

QA flagged that no `--amber-fg` exists, so a substitution might not suffice. **It does** — `--amber` already carries per-theme values chosen against light grounds, which is exactly what the literal lacked.

## Appearance changes in dark too, deliberately

`#f59e0b` → `#fbbf24`, a slightly brighter amber, and the token the system already names for *"weak/caution state"*. Flagged on the issue before building rather than left to be discovered in review.

## Not changed, each for a reason

| Site | Why |
|---|---|
| `KLineProChart:1018,1739` | klinecharts **canvas strokes** — graphics at 3:1, not DOM text |
| `news/page.tsx:171-173` | Reuters **source-identity** colours — same class as `coinBadgeColor` on #734; identity, not state, and that question is the owner's |

`UsageRings:14` **is** included though it's a ring rather than text: its three sibling entries already use `var(--green-2)`/`var(--accent-2)`, so the literal was the odd one out in its own array. **Consistency with its neighbours, not a measured failure** — stated rather than folded into the count.

## How to test (QA)

1. **`/markets` and `/liq` in the CURRENT design, theme toggled to light** — the `C` grade chip and the squeeze warnings are readable. This is the half that matters.
2. Same two routes in terminal light.
3. Dark, both designs: amber is marginally brighter than before.
4. The chart's dashed amber line and the Reuters news chips are unchanged.

## Risk level

**Medium.** Touches nine files across both designs and changes a visible colour in dark as well as light. Mitigated by every site being the same one-value substitution to a token the system already owns. Gates: lint 0 errors, `tsc` 0, `npm test` 571/0, `build` 0.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -3037,7 +3037,7 @@ html[data-design="terminal"][data-theme="light"] { background: var(--bg0); }
 .ps-result-profit .ps-rval { color: var(--green); }
 .ps-result-good   .ps-rval { color: var(--green); }
 .ps-result-danger .ps-rval { color: var(--red); }
-.ps-warn { background: rgba(217,119,6,.1); border: 0.5px solid rgba(217,119,6,.3); border-radius: 8px; padding: 8px 12px; font-size: var(--fs-caption); color: #f59e0b; margin-bottom: 8px; }
+.ps-warn { background: rgba(217,119,6,.1); border: 0.5px solid rgba(217,119,6,.3); border-radius: 8px; padding: 8px 12px; font-size: var(--fs-caption); color: var(--amber); margin-bottom: 8px; }
 .ps-log-btn { width: 100%; padding: 13px; border-radius: var(--radius-card); border: 0.5px solid var(--green-bdr); background: var(--green-bg); color: var(--green); font-size: var(--fs-label); font-weight: 700; cursor: pointer; margin-top: 4px; transition: box-shadow .15s; }
 .ps-log-btn:hover { box-shadow: 0 0 16px rgba(52,211,153,.2); }
 /* ── Shared empty state pattern ── */
@@ -3158,7 +3158,7 @@ html[data-design="terminal"][data-theme="light"] { background: var(--bg0); }
 .tj-trade-dir-tag.tj-short { background: var(--red-bg);   color: var(--red);   border: 0.5px solid var(--red-bdr); }
 .tj-trade-setup-tag { font-size: var(--fs-caption); color: var(--txt3); padding: 2px 7px; border-radius: 20px; border: 0.5px solid var(--bdr); }
 .tj-result-badge { font-size: var(--fs-caption); font-weight: 700; padding: 2px 8px; border-radius: 20px; }
-.tj-rb-open { background: rgba(217,119,6,.12); color: #f59e0b; border: 0.5px solid rgba(217,119,6,.3); }
+.tj-rb-open { background: rgba(217,119,6,.12); color: var(--amber); border: 0.5px solid rgba(217,119,6,.3); }
 .tj-rb-win  { background: var(--green-bg); color: var(--green); border: 0.5px solid var(--green-bdr); }
 .tj-rb-loss { background: var(--red-bg);   color: var(--red);   border: 0.5px solid var(--red-bdr); }
 .tj-rb-be   { background: var(--bg3); color: var(--txt3); border: 0.5px solid var(--bdr); }
--- a/app/liq/page.tsx
+++ b/app/liq/page.tsx
@@ -602,16 +602,16 @@ export default function LiqPage() {
                     border: '0.5px solid rgba(245,158,11,0.3)',
                     display: 'flex', gap: 8, alignItems: 'flex-start',
                   }}>
-                    <span style={{ color: '#f59e0b', flexShrink: 0, lineHeight: 0 }}><Warn size={13} /></span>
+                    <span style={{ color: 'var(--amber)', flexShrink: 0, lineHeight: 0 }}><Warn size={13} /></span>
                     <div style={{ fontSize: 'var(--fs-caption)', color: 'var(--txt3)', lineHeight: 1.6 }}>
                       {longSqueezeRisk ? (
                         <>
-                          <span style={{ color: '#f59e0b', fontWeight: 700 }}>{t('LIQ_WHALE_LONG_SQUEEZE_TITLE')}</span>
+                          <span style={{ color: 'var(--amber)', fontWeight: 700 }}>{t('LIQ_WHALE_LONG_SQUEEZE_TITLE')}</span>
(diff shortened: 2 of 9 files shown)


JohnDominicJasmin:
## QA review — approve. Merge and promote. This is the most consequential fix of the session.

**It fixes a defect that is live on liquidity-hq.com**, not on a staging surface and not behind a flag. Seven elements on `/markets` at **1.62:1**, reachable by one click of the app bar's theme button.

### Verified

| | |
|---|---|
| Both measured failure sites covered | `MarketsTerminal.tsx` (the `"C"` badge) and `LiqTerminal.tsx` (`"Short squeeze"`) both in the diff |
| Literal removed everywhere it was text | 5 occurrences remain, and they are **exactly** the 5 you named |
| No new token | `--amber` already carries `#8F4508` app-light and `#755100` terminal-light |

```
remaining #f59e0b on the branch
  app/news/page.tsx:171-173      Reuters source identity  -> #734's coinBadgeColor question
  components/KLineProChart.tsx:1018,1739   canvas strokes -> graphics at 3:1, not DOM text
```

**Each exclusion has a reason rather than being an omission**, and the reasons are the right ones. Canvas strokes are not text and the 4.5 bar does not apply; source-identity colours are the same class as the coin badges and belong with that decision, not this one. Splitting them out rather than sweeping them in is what keeps #734's ruling coherent.

### On `UsageRings` — labelling it honestly is the right call

You flagged it as **consistency, not contrast**: a ring is not text, so it was never a measured failure; its three siblings already take `var(--green-2)`/`var(--accent-2)`, so the literal was the odd one out in its own array.

Including it is fine. **Saying which of the 16 sites are fixes and which are tidying is what stops the next reader inferring that all 16 were failing.** That distinction is exactly what I lost on #734 when I collapsed four causes into one story.

### The dark appearance change is correctly surfaced

`#f59e0b` → `#fbbf24` in dark. Small, and `--amber` is the token the system already names for *"weak/caution"* — but it is a visible change on a shipping surface, and putting it in the PR rather than leaving review to find it is the difference between a disclosed change and a discovered one.

### Your estimate versus my measurement

You computed **1.92** against a clean `#ffffff`; I measured **1.62** by compositing the real ancestor chain from the rendered page. **Yours was the right method with an assumed ground; mine had the actual ground.** Neither is near 4.5 and the fix is the same — recording both so the number in the issue is not mistaken for a single authoritative reading.

### On #741 being the finding of the day

Agreed, and your framing sharpens it: the blind-spot pattern had a **third layer** we had not seen. Not the tool answering an adjacent question, not the fact living in two sources — the *configuration every tool was run with*. Five scripts, one hardcoded line each, and the design users actually get was never measured by any of them.

**This amber is the proof.** It surfaced only because it happened to also appear on a surface nobody reaches. I have added `--design terminal|current|both` to `platform-audit.mjs` and the first current-design sweep is running now — 48 page loads of a surface that has never been audited. I do not know what it will find, which is the point.

**Merge it.** I will promote and re-measure `/markets` on the deployed build, three samples.

— QA Team

Issue #748, opened 2026-09-03
Make terminal the default design on every route — reverses #719's landing-only scope


JohnDominicJasmin:
Owner, 2026-09-04: *"remove ?design terminal make terminal design default"*

**This reverses #719's landing-only scope.** Terminal becomes the default on every route, for every visitor, with no query param.

## The change

`lib/designMode.ts` currently resolves: query param → stored preference → `TERMINAL_BY_DEFAULT_ROUTES` (which contains only `'/'`) → `current`.

The owner wants terminal as the default everywhere. Mechanically that is the route list becoming every route, or the final fallback becoming `terminal` rather than `current` — **dev's call which, and the `?design=current` escape hatch must survive either way.** QA and the owner both need it to compare designs.

`app/layout.tsx`'s inline pre-paint script duplicates the precedence and is kept in step by `__tests__/designMode.test.mts` across all 64 input combinations. That test is the guard; it must be updated deliberately rather than left to fail.

## What this changes about everything parked today — the important half

I closed or parked several issues on the explicit reasoning that **terminal was reachable only by opting in twice, so defects there were on a path no user takes.** That reasoning dies with this change. Every one of these becomes live:

| | | |
|---|---|---|
| #707 | `/hours` white labels on pastel bands — **1.52:1** in terminal light | closed *"not planned"* |
| #698 | `/scanner` 26 dark contrast failures, worst 3.51 | closed *"not planned"* |
| #663 | terminal typography defined in two places | closed *"not planned"* |
| #734 | `coinBadgeColor` used for 12px text — worst **1.38:1** | open, awaiting owner ruling |
| #738 | 6 signal labels at 4.35 where `--green-fg` exists | open |

**#707's 1.52:1 and #734's 1.38:1 would ship to every visitor.** I closed the first as not planned. That was correct under landing-only and is wrong under this instruction.

**Recommendation: reopen #707, #698 and #663 before this ships**, and treat #734's ruling as blocking rather than deferred. Not asking the owner to re-decide those individually — flagging that my closure reasoning was scope-dependent and the scope just moved.

## What does not change

The landing page keeps its own nav (#728) and its own design; it was already terminal. Nothing about `/` moves.

## Sequencing

**This should ship after the terminal-light work, not before.** The audit found real defects on surfaces that were invisible; making them visible first would put them in front of users. #742 (the amber literal) is already merged and is a production fix regardless.

— QA Team

JohnDominicJasmin:
[...]

Issue #744, opened 2026-09-03
Terminal dashboard: the .scc-card coin row is fully transparent in both themes


JohnDominicJasmin:
Owner, 2026-09-04: *"the coin item. Why does it have no background even in dark theme."*

The wide BTC price row under BEST SETUP TODAY is `.scc-card`. Measured on deployed staging `dc6a4e3`, terminal, **both themes**:

```
                dark                       light
ownBg           rgba(0, 0, 0, 0)           rgba(0, 0, 0, 0)
border          none                       0px none
sits on         body rgb(6,7,10)           body rgb(247,246,243)
size            1079 x 66
```

**Transparent in both themes, with no border.** It is a card in name only — every other panel on the screen has a ground, so this one reads as unstyled rather than deliberately flat.

More visible in dark because the near-black body offers no contrast at all against a transparent block.

Every sibling panel takes a terminal ground token, so the fix is presumably `--bg1`/`--bg2` plus the standard border — but **which** is a design call rather than something to guess. Flagging that rather than picking one.

— QA Team

JohnDominicJasmin:
Verified on deployed staging `0a00c24`, terminal, **both themes**:

```
              .scc-card bg        body bg             border-bottom      radius
dark          rgb(20, 21, 23)     rgb(6, 7, 10)       -                  -
light         rgb(235, 233, 230)  rgb(247, 246, 243)  0.8px rgb(213,210,205)   0px
```

Was `rgba(0,0,0,0)` in both. It now reads as a surface in each theme rather than sitting invisibly on the page ground.

**#607's hairline and zero radius survived**, which was the part worth checking: the row gained a ground without becoming a floating card again. `border-radius: 0px` and the 0.8px separator are both intact.

**On the #607 question** — dev flagged this as reversing a prior ruling. It was not: #607's stated basis was conformance to `Dashboard 2a.dc.html`, and the owner cancelled the canvas-mirror initiative on 2026-09-03 (#718). The 'no fill' half was canvas-derived and lost its authority; the hairline and radius come from terminal's live design rules, which is why they stay.

— QA Team

PR #749, opened 2026-09-03
fix(dashboard): restore the section gap, and give the coin row a ground

### LINKED ISSUE #743: Terminal dashboard: no spacing between the market-read panel and BEST SETUP TODAY
Owner, 2026-09-04, on the terminal dashboard: *"I want to add spacing between the market read and best setup today."*

Measured on deployed staging `dc6a4e3`, terminal, desktop:

```
.dash-section (the "BEST SETUP TODAY" block)
  margin-top     0px      <- no separation from the panel above
  margin-bottom  10px
```

The section owns bottom spacing but not top, so the market-read panel and the BEST SETUP label sit flush.

**Not a token or theme issue** — one spacing value on a shared class. Check whether `.dash-section` is used elsewhere before setting `margin-top`, since a change there moves every section that uses it; a scoped rule may be safer than editing the base.

— QA Team

JohnDominicJasmin:
## Summary

Two owner-reported items on the terminal dashboard. Fixes #743, addresses #744.

## #743 — the section gap

`.dash-section` already declares `margin: 20px 0 10px`. An inline `marginTop: 0` on the BEST SETUP TODAY label outranked it and closed the gap, so the market-read panel and the label sat flush.

**Deleting the override restores the 20px the rule already intended** — the same inline-owns-a-property-the-stylesheet-owns shape as #681 and #660.

## #744 — the coin row's ground, and this one is a proposal

`.scc-card` takes `background: var(--bg1)` under terminal instead of `transparent`.

**This partly reverses #607, and I haven't hidden that.** That decision made the whole main column *banded rather than carded* — no per-band fill, hairline separators, column ground throughout — and `background: transparent` was part of it, not an oversight. So this is a ruling overriding a ruling, not a bug fix. QA asked me to propose rather than pick; here's what I picked and what I deliberately left.

**What I changed:** `--bg1` lifts the row off `--bg0` enough to read as a surface while staying inside terminal's palette. #607's hairline separator and zero radius are **kept** — so the row gains a ground without becoming a floating card again, which is the half of #607 the owner didn't object to.

**What I left:** the sibling band (`.dash-conditions-row`) gets no fill. The owner named this element specifically, and changing the others to match would be me deciding how far the reversal goes. **If the column should be carded throughout, that's a bigger call and it's theirs.**

Hover still reads: resting `--bg1`, hover `--bg2` from the base rule, distinct in both themes.

## How to test (QA)

1. `/dashboard?design=terminal`: a 20px gap above BEST SETUP TODAY.
2. The BTC coin row below it has a visible surface in **both** themes rather than sitting flush on the body.
3. Hovering the row still changes its ground.
4. The neighbouring conditions row is unchanged — **that's the open question, not an omission.**

## Risk level

**Low** for #743, **Low but visible** for #744 — one background value on one element, reversing part of a documented decision. Gates: lint 0 errors, `tsc` 0, `npm test` 571/0, `build` 0.

Per your reordering these are items 5 and 6, behind the four that block #748 — no rush on the review.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -7181,9 +7181,28 @@ main.app-content[data-chrome="none"] {
    per-band fill. Terminal only - the current design's dashboard keeps its
    spaced cards, and .scc-card/.dash-conditions-row both render there too. */
 [data-design="terminal"] .dash-main { gap: 0; }
+/* The coin row keeps a surface (#744). Owner: "the coin item. Why does it have
+   no background even in dark theme."
+
+   THIS PARTLY REVERSES #607, WHICH IS WHY IT IS SPELLED OUT. That decision made
+   the whole main column banded rather than carded - no per-band fill, hairline
+   separators, column ground throughout - and `background: transparent` here was
+   part of it, not an oversight. The owner is asking for the opposite for this
+   one element, so this is a ruling overriding a ruling rather than a bug fix.
+
+   --bg1, not the base rule's card treatment: it lifts the row off --bg0 enough
+   to read as a surface while staying inside terminal's palette, and the
+   hairline separator and zero radius from #607 are kept. So the row gains a
+   ground without becoming a floating card again, which is the half of #607
+   the owner did not object to.
+
+   The sibling bands (.dash-conditions-row below) are deliberately NOT given a
+   fill - the owner named this element specifically, and changing the others to
+   match would be me deciding how far the reversal goes. If the column should
+   be carded again throughout, that is a bigger call and theirs. */
 [data-design="terminal"] .dash-main > .scc-card {
   border: 0; border-bottom: 1px solid var(--bdr);
-  background: transparent; border-radius: 0;
+  background: var(--bg1); border-radius: 0;
 }
 [data-design="terminal"] .dash-main > .dash-conditions-row {
   gap: 0; border-bottom: 1px solid var(--bdr);
--- a/components/DashboardTerminal.tsx
+++ b/components/DashboardTerminal.tsx
@@ -512,7 +512,15 @@ export default function DashboardTerminal() {
         <MarketRead />
 
         <TPanel id="tour-best-setup">
-          <div className="dash-section dash-section-hot" style={{ marginTop: 0 }}>{t('DASH_BEST_SETUP_TODAY_HEADER')}</div>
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
## QA review — approve, and your #744 framing is more cautious than it needs to be. #607's authority was withdrawn.

### #743 is clean

`.dash-section` already declares `margin: 20px 0 10px`. An inline `marginTop: 0` outranked it. **Deleting the override restores the gap the rule already intended** — same inline-beats-stylesheet shape as #681 and #660, and the third instance of that pattern on this dashboard.

### #744 — you called it "a ruling overriding a ruling". It is not, and the distinction matters

You wrote that `background: transparent` was deliberate under #607 and the owner is now asking for the opposite, so this is a contested override. **I checked what #607 rested on:**

> *"conflicts with the canvas's hairline-only divider pattern"*
> *"`Dashboard 2a.dc.html`'s main column has **no gap** on its flex container"*

**#607 was canvas conformance.** Its entire justification was matching `Dashboard 2a.dc.html`.

**The owner cancelled the canvas-mirror initiative on 2026-09-03** (#718): *"we are repackaging the platform without users feedback… we are wasting time, money, and resources."* Every canvas-derived decision lost its authority that day — that is why #656 and #652 were closed as superseded and why #718 exists at all.

**So there is no competing ruling here.** #607's premise is void, the owner has asked for a ground, and giving it one is straightforwardly correct rather than a contested trade.

**Your caution was still right in method.** Surfacing the conflict rather than quietly reversing a prior decision is what let me go and check the basis. Had #607 been an owner ruling or a live design decision, your framing would have been exactly correct and the change would have needed their word. It just happens that the authority it rested on is one the owner has already withdrawn.

### On keeping #607's hairline and zero radius

Right call, and worth stating why it survives the above: **the hairline separator and squared corners are terminal design rules, not canvas conformance.** They come from the palette and the radius ruling, which are live. Only the "no fill" half was canvas-derived. Splitting those two rather than reverting #607 wholesale is the correct read.

### One thing in the diff that is not in the description

The two `#f59e0b` → `var(--amber)` lines in `globals.css` are #742's work, already merged. Presumably a rebase artifact rather than a re-application. **Harmless — same change, same value** — but worth a glance to confirm it is not a stale copy that could conflict later.

**Merge it.**

— QA Team

PR #751, opened 2026-09-03
test(tokens): detect declaration collisions in both directions (#663)

JohnDominicJasmin:
## Summary

Closes the detector half of #663. The terminal declaration-ownership guard now
covers seven properties instead of five, and reports **which side of a
collision wins** — because the fix is opposite depending on the answer.

This is deliberately **only** the detector. #663 asks two questions: a tooling
question (how do we stop this recurring) and a convention question (which side
*should* own these declarations). This PR answers the first. The second is
still open and is not something dev should settle unilaterally — see "What this
does NOT do".

## What changed

`__tests__/terminalTypographyOwnership.test.mts`:

- `Owned` gained `important: Set<string>`; the CSS parser now records
  `!important` per declaration, not just which properties a rule sets.
- The reporter emits direction-specific text:
  - `inline <prop> kills the terminal rule's <prop>` — the CSS rule is dead
    (#629, #660's shape)
  - `the terminal rule's <prop> !important kills inline <prop>` — the inline
    value is dead (#633's shape)
- `terminalRules()` and `collisions()` take the property list as a parameter,
  so one parser serves both checks.
- New `WIDE_PROPS` (background, border, border-radius, color, display, margin,
  padding) checked against `WIDE_BASELINE`, an **enumerated** list of the 25
  collisions that exist today.
- New "the wide check is armed" test — asserts the CSS walk parsed a plausible
  number of rules **and** that at least one carries `!important`, so the
  direction half cannot silently measure nothing.

The original typography assertion is untouched and still strict at **0**.

## Why an enumerated baseline, not zero

The 25 are not all defects. `.edge-card-value` / `.edge-card-signal` set
`color` inline from live data with the CSS colour as a deliberate fallback —
that is correct code. Several `.card` entries are components legitimately
styling a shared class. Asserting zero would mean either 25 edits I cannot
justify or an exemption list longer than the check itself.

Enumerating instead makes it a ratchet:

| Change | Result |
|---|---|
| New collision appears | fails, naming file, class, property, direction |
| Existing one fixed | fails until its baseline line is deleted |
| One swapped for another | fails both ways |

## What this found

**Two instances of #633's shape that nothing could see before.** Both set an
inline `borderRadius` that a terminal `!important` overrides:

- `components/Skeleton.tsx` — `.skel-bar`
- `components/UpgradeGateModal.tsx` — `.locked-card-term-wrap`

Neither is harmful — terminal wants zero radius and gets it. But both are
inline values that do nothing, which is exactly the class of thing that cost
three separate computed-value investigations this cycle. They are left in place
and recorded in the baseline rather than fixed here, so the convention ruling
decides them rather than me.

## Negative result

My first control probe did **not** trip the check, and that was correct
behaviour, not a bug. I probed `.card` with an inline `display` and `.skel-bar`
with an inline `padding` — neither terminal rule declares those properties, so
there is no collision to report. The check is scoped to what the CSS actually
owns, not to the property list. Recorded because "I added a probe and nothing
happened" is otherwise indistinguishable from a broken detector, and the second
probe is what proves it is not.

## How to test (QA)

This is test-only — no application code, no user-visible change. Verify the
guard rather than the UI.

1. On this branch, `npm test`. Expect **573+ pass / 0 fail**, including
   `no NEW declaration collision, in either direction`.
2. **Prove the forward direction fires.** Create `components/__probe.tsx`:
   ```tsx
   export default function P() {
     return <div className="card" style={{ background: 'red' }} />;
   }
   ```
   `npm test` must now fail with
   `components/__probe.tsx: .card — inline background kills the terminal rule's background`.
3. **Prove the reverse direction fires.** Change that element to
   `<div className="skel-bar" style={{ borderRadius: 4 }} />`.
   The failure must now read
   `the terminal rule's border-radius !important kills inline borderRadius`
   — the opposite sentence. Delete the probe file; tests pass again.
4. **Prove the baseline is a ratchet, not a suppression.** Delete any one line
   from `WIDE_BASELINE` and re-run — it must fail as a NEW collision. Restore it.

Step 3 is the one worth doing: the whole point of this PR is that the two
directions read differently.

## Risk level

**Low.** Test-only. Four gates pass on this branch: lint 0 errors (243
pre-existing warnings, unchanged), `tsc --noEmit` clean, 573/573 tests, build
succeeds.

**Not verified:** nothing about this runs in the browser, so there is no
rendered check to do. The 25 baseline entries were triaged by reading, not by
inspecting computed styles on a live page — I am asserting they are *detected*,
not that all 25 are *harmless*.

## What this does NOT do

- **It does not pick a convention.** #663 lists three candidates for who should
  own these declarations. QA said "take #663" but has not ruled among them, and
  that ruling is not dev's to make. The detector works under any of the three.
- **It does not fix the 25.** Including the two new ones. Each is a judgement
  call the convention decides.
- **It does not widen past seven properties.** Anything with a shorthand /
  longhand relationship the parser does not model (e.g. `padding` vs
  `padding-top`) would need that handled first.

### DIFF
--- a/__tests__/terminalTypographyOwnership.test.mts
+++ b/__tests__/terminalTypographyOwnership.test.mts
@@ -17,11 +17,38 @@
  * flagging those would bury the real case, which is the mistake #666's first
  * version made in the other direction by flagging `button, input`.
  *
- * The reverse collision - a CSS `!important` beating a component's inline value,
- * which is #633's shape - is NOT covered here. 53 terminal rules use
- * `!important` across background, border, border-radius, color, display, margin
- * and padding. That is a live exposure with no detector, and it is recorded on
- * #663 rather than silently implied to be handled.
+ * The reverse collision - a CSS `!important` beating a component's inline
+ * value, which is #633's shape - IS covered, by the second assertion at the
+ * foot of this file. It was not, when the paragraph above was first written;
+ * that gap is what #663 called "a live exposure with no detector".
+ *
+ * ── THE CONVENTION THIS ENFORCES ──────────────────────────────────────────
+ *
+ * Ruled on #663 by QA on 2026-09-04, after dev declined to pick it. Written
+ * here rather than left in an issue comment, because the rule has to be
+ * findable from the code it governs:
+ *
+ *   1. A TERMINAL-ONLY component puts all typography and constant colour in
+ *      CSS. There is no second design to serve, so a constant has no reason
+ *      to be inline. `DashboardTerminal` and the `.csb2-*` rules are the
+ *      pattern; #660 already did this.
+ *
+ *   2. In a DUAL-DESIGN component, inline is for COMPUTED values only - a
+ *      colour derived from data, a width from a percentage, a transform from
+ *      a measurement. Anything constant belongs in CSS. The test to apply:
+ *      could this value have been written in a stylesheet? Then it should be.
+ *
+ * A third rule - banning `!important` under `[data-design="terminal"]`
+ * outright - was DECLINED on sequencing, not on merit. Converting 53
+ * declarations to higher-specificity selectors is a large mechanical change
+ * across shared chrome, and #748 is about to make terminal the default on
+ * every route; the risk lands exactly where the blast radius widens. It is
+ * the right end state and gets its own issue once terminal-default has been
+ * live long enough for a regression to be attributable.
+ *
+ * Both adopted rules describe which side SHOULD own a declaration. This file
+ * detects only that both sides claim one - which is what makes it valid under
+ * either rule, and why it does not try to auto-decide the fix.
  */
 import test from 'node:test';
 import assert from 'node:assert/strict';
--- a/__tests__/terminalTypographyOwnership.test.mts
+++ b/__tests__/terminalTypographyOwnership.test.mts
@@ -54,9 +54,9 @@ const PROPS: Array<[css: string, jsx: string]> = [
  *  Only the last sequence matters: ancestors restrict WHICH elements match, so
  *  ignoring them can over-report but never under-report, and over-reporting on
  *  an ancestor is rare enough to accept against parsing full descendant paths. */
-interface Owned { need: string[]; props: Set<string> }
+interface Owned { need: string[]; props: Set<string>; important: Set<string> }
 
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it. And the convention you declined to pick is now ruled on #663.

### Verified

```
tests 5  pass 5  fail 0        on the current tree
Owned.important: Set<string>   !important recorded per declaration
messages: "inline <prop> kills the terminal rule's <prop>"
          "the terminal rule's <prop> !important kills inline <prop>"
```

**Reporting which side wins is the whole value.** #629 and #660 were the CSS rule dying; #633 was the inline value dying. Same symptom, opposite fixes, and a detector that only said "these collide" would leave the reader to work out which — which is the step that has failed three times.

### Your negative control result is the part I want on the record

> *"My first control probe did not trip the check, and that was correct behaviour, not a bug… neither terminal rule declares those properties, so there is no collision to report."*

**"I added a probe and nothing happened" is indistinguishable from a broken detector**, and the temptation is to assume the guard works and move on. Probing again against a property the CSS actually owns is what turns a silent pass into evidence. That is the second time today you have caught a control that proved nothing — the first was #737's `indexOf` matching the wrong block.

### The convention — ruled, so this has something to enforce

**Adopted: rules 1 and 2, together.** Terminal-only components put constants in CSS; dual-design components may use inline *only* for computed values. Complementary rather than alternatives — rule 1 settles terminal-only, rule 2 settles the rest.

**Declined: rule 3** (banning `!important` under `[data-design="terminal"]`). Your own words — *"most correct, most invasive, needs its own pass"* — and I agree with all three. Declining on **sequencing**: 53 declarations converted to higher-specificity selectors, across shared chrome, in the week #748 makes terminal the default everywhere. A specificity change that looks equivalent and is not would be found the way the last three were: by reading computed values, after shipping. Reopen once terminal-default has settled.

### The 25 collisions

**Do not open a PR fixing them.** They are ownership ambiguities, not measured failures — none is a `1.38:1`. Work them under the convention as files are touched. A dedicated pass is churn on shared chrome for no measured user-facing gain, and it is not a #748 blocker.

**Merge it.** #738's `/scanner` half next, unless the green token question lands first.

— QA Team

Issue #752, opened 2026-09-03
LIVE IN PRODUCTION: /news source pills at 1.99-2.00:1 — includes the #f59e0b that #742's sweep missed


JohnDominicJasmin:
**Dev Team** — found while checking whether #739's sweep was complete. It was not. Five `#f59e0b` literals survive PR #742, and looking at where they live turned up a larger problem next to them.

Measured from tokens, not rendered — stated up front, because that is the instrument's limit and the last three defects came from not stating it. Direction is not in doubt at 1.68:1; the exact figure could move by a few tenths.

---

## 1. `/news` source pills — 6 of 11 source colours fail as text

`app/news/page.tsx:170` `SOURCE_COLORS` feeds `SourcePill`, which renders the **source name as text** at `--fs-micro`, `font-weight: 700`, `opacity: 0.9`, on `.ncard-grid`'s `var(--bg1)`. Half the map is raw hex chosen as a brand colour, with no theme awareness.

Contrast of the pill text against `--bg1`, all four contexts, `opacity: 0.9` composited:

```
source              value        cur dark   cur light   term dark   term light
Reuters (x3)        #f59e0b        7.68       2.00        7.08        1.68
BBC (x2)            #e11d48        3.62       4.26        3.33        3.60
CoinDesk            #1a7aff        4.26       3.47        3.93        2.94
The Block           #38bdf8        7.70       1.99        7.10        1.68
CryptoSlate         #818cf8        5.61       2.64        5.18        2.23
AP News (x2)        var(--accent-2) 6.43      5.58        6.84        4.92
CoinTelegraph       var(--green-2) 8.55       5.26          -         4.27
Decrypt             var(--orange)  7.30       7.30        6.73        6.21
Bitcoin Magazine    var(--amber)   9.80       5.55        9.03        4.78
Finnhub             var(--txt-dim) 5.26       5.09        4.78        4.48
(unmapped source)   var(--txt3)    4.43       6.43        4.05        4.14
```

**Worst: 1.68:1** — Reuters and The Block in terminal light, against a 4.5 bar for text.

Three findings in that table, in severity order:

1. **`#e11d48` and `#1a7aff` fail in all four contexts.** Not a light-theme problem — a colour problem. BBC and CoinDesk are unreadable everywhere.
[...]

PR #753, opened 2026-09-03
fix(hours): derive each session-band label from the band's own luminance (#707)

### LINKED ISSUE #707: /hours: white labels on pastel session bands, 1.52:1 in light — plus two pills at 4.28

JohnDominicJasmin:
## Summary

Closes #707, third attempt, on the owner's ruling: **the label colour is derived per band from that band's own luminance**, black or white, and never written down.

Verified by rendering in all four theme × design combinations, not only by arithmetic — and the rendered check caught a real bug the arithmetic could not see. Details under "What the rendered check found".

## What changed

**`lib/readableOn.ts`** (new) — `parseCssColor`, `compositeOver`, `relativeLuminance`, `contrastRatio`, and `readableOn(surface, ground)` returning whichever of black/white measures higher once `surface` is flattened onto `ground`. `ground` is required: a translucent band measured against nothing is measured against white, which is the light-theme answer, which is the bug.

**`app/hours/page.tsx`** — the strip's real ground (`--bg3`) is read from the live stylesheet with `getComputedStyle`, watched with a `MutationObserver` on `data-theme` / `data-design` plus the existing `theme-change` event. Each band label takes `readableOn(seg.bg, ground)`. **The `opacity: 0.9` is removed.**

**`__tests__/readableOn.test.mts`** (new) — 9 tests. Band colours are parsed out of `page.tsx` and grounds resolved out of `globals.css`, so neither is restated here.

## Why one colour could never work

```
              six bands, four contexts
'#fff'        fails all six in light          1.43 - 1.78
var(--txt)    fails three in dark             2.33 - 4.42   ← what I shipped in #750
```

The bands are translucent, so they land **light in light mode and mid-tone in dark**. A mid-tone band is equidistant from both ends of the scale — no single token reaches it from either side.

## The measured result

Every band, every context, opacity removed:

```
band     current dark    current light   terminal dark   terminal light
DEAD     white  8.34     black 11.57     white  8.17     black 11.28
LONDON   white  5.42     black 12.66     white  5.33     black 12.37
PRE_NY   white  7.81     black 12.43     white  7.63     black 12.10
NY       white  6.23     black 11.43     white  6.12     black 11.15
PRIME    black  6.89     black 13.97     black  6.96     black 13.77
ASIA     white  4.60     black 14.10     black  4.64     black 13.84
```

**ASIA is a 0.12 tie in dark** — 4.57 black against 4.60 white in the current design, and the other way round in terminal — so the derived answer differs by design on that one band. Both clear. Forcing them to agree would need a tiebreak rule, which is a table wearing a different hat.

**The fade had to go, and that is not tidying.** At `opacity: 0.9`, ASIA in dark fails *both* ways: 4.27 black, 4.07 white. No colour clears that band through the fade, so the fade was the thing that had to move. There is a test asserting the label carries no opacity, so it cannot come back as a "restore the original look" change.

## What the rendered check found

The arithmetic said terminal dark ASIA should be **black**. Rendered, it came out **white**.

Cause: React runs child effects before parent effects, so this page's effect fired **before `DesignModeProvider` had put `data-design` on `<html>`**. The read returned `#0f1115` (current design) instead of `#111416` (terminal). Two grounds four RGB steps apart — invisible on every band except the one that is a 0.12 tie, where it picks the other colour.

Fixed by watching the attributes with a `MutationObserver` instead of assuming when they settle — the pattern `GrokSignalChart` already uses. Re-verified: terminal dark ASIA now renders `rgb(0,0,0)`.

**This is the argument for QA's guardrail, made against my own work.** Token arithmetic predicted #752's production numbers to two decimal places, and it still could not see this, because the defect was not in the colour maths — it was in *which ground the maths was handed*.

## Honest note on one of the tests

`every band label clears 4.5:1 in all four contexts` **cannot fail on a band colour**, and I found that by control-testing it rather than by trusting it. `contrast(white,L) = 1.05/(L+0.05)` and `contrast(black,L) = (L+0.05)/0.05` cross at L = 0.1791, where both are **4.58** — so `max(black, white)` has a floor of 4.58 for every opaque colour that exists.

That is the guarantee the ruling rests on, so it is now asserted directly: a dense sweep of the sRGB cube confirms the floor at 4.58. The original test is kept for what it *does* catch (an unresolvable `--bg3`, a band literal that stops parsing, a future change to the candidate set) with its ceiling written into the test rather than left for the next reader to discover.

## How to test (QA)

On `qa`, `/hours`, the 24-hour session bar at the top.

1. **Dark, current design.** Every band label legible. PRIME reads **black** where the others read white — that is correct, not a bug.
2. **Light, current design.** Every label black and legible. The two that failed at 1.46–1.78 (PRIME, ASIA) are the ones to look at.
3. **Terminal, dark** (`?design=terminal`). Same as 1, except **ASIA is also black**. The 0.12 tie flips with the ground.
4. **Terminal, light.** All black.
5. **Toggle the theme with the page open**, in both designs. Labels must re-derive immediately — this is the `MutationObserver`, and it is the part that was broken until the rendered check.
6. **Hard-reload directly on `?design=terminal`** and check ASIA without touching the toggle. That is the exact case the effect-ordering bug got wrong; on a stale ground ASIA renders white.

## Risk level

**Medium.** Four gates pass: lint 0 errors, `tsc --noEmit` clean, 583/583 tests, build succeeds. Verified by rendering in all four contexts on a local dev server, reading computed `color` off each label rather than eyeballing.

**What is still unverified:** not measured on a deployed environment — I am under a standing instruction not to deploy, so the rendered check is localhost against this branch. `--bg3` reads `#0f1115` / `#e4e6e9` / `#111416` / `#e3e1dd` there, matching the stylesheet, so I do not expect a deployed difference; that is an expectation, not a measurement.

**Thinnest margin:** ASIA in dark at 4.57 / 4.64, about 0.1 above the bar. If a band colour is ever adjusted, that is the one that crosses first.

**Third attempt on this issue.** The first two each passed review and each broke the theme nobody checked, so the useful question on this one is not "does light work" but "was every ground measured" — steps 3, 5 and 6 are where the previous two would have failed.

### DIFF
--- a/__tests__/readableOn.test.mts
+++ b/__tests__/readableOn.test.mts
@@ -1,0 +1,212 @@
+/* The /hours session-band labels, measured against the real palette (#707).
+ *
+ * Two fixes shipped here and each one fixed a theme and broke the other:
+ * `'#fff'` fails all six bands in light, `var(--txt)` fails three in dark.
+ * Both were reasoned about correctly and measured on one axis. This is the
+ * test that makes the second axis non-optional.
+ *
+ * It reads the band colours from app/hours/page.tsx and the grounds from
+ * app/globals.css - not from constants restated here - so a palette change
+ * moves the assertion with it. Restating either would be the two-sources bug
+ * that #736 and #663 are both about, in the file meant to prevent it.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import {
+  readableOn, parseCssColor, compositeOver, contrastRatio, relativeLuminance,
+  BLACK, WHITE, type Rgb,
+} from '../lib/readableOn.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const CSS = readFileSync(path.join(ROOT, 'app', 'globals.css'), 'utf8');
+const PAGE = readFileSync(path.join(ROOT, 'app', 'hours', 'page.tsx'), 'utf8');
+
+const BAR = 4.5;
+
+/** Custom-property declarations inside one top-level rule. */
+function tokens(selector: string): Record<string, string> {
+  const lines = CSS.split('\n');
+  const idx = lines.findIndex(l => l.trim() === selector + ' {');
+  assert.ok(idx >= 0, `token block not found in globals.css: ${selector}`);
+  const from = lines.slice(0, idx).join('\n').length;
+  let i = CSS.indexOf('{', from), depth = 0, end = i;
+  for (; i < CSS.length; i++) {
+    if (CSS[i] === '{') depth++;
+    else if (CSS[i] === '}') { depth--; if (depth === 0) { end = i; break; } }
+  }
+  const body = CSS.slice(CSS.indexOf('{', from) + 1, end);
+  const out: Record<string, string> = {};
+  for (const m of body.matchAll(/(--[a-z0-9-]+)\s*:\s*([^;]+);/g)) out[m[1]] = m[2].trim();
+  return out;
+}
+
+function resolve(map: Record<string, string>, value: string, depth = 0): string | null {
+  if (depth > 10) return null;
+  const m = /^var\((--[a-z0-9-]+)(?:\s*,\s*([^)]+))?\)$/.exec(value.trim());
+  if (!m) return value.trim();
+  const next = map[m[1]] ?? m[2];
+  return next === undefined ? null : resolve(map, next, depth + 1);
+}
+
+const root = tokens(':root');
+const light = tokens('[data-theme="light"]');
+const termDark = tokens('[data-design="terminal"]:not([data-theme="light"])');
+const termLight = tokens('[data-design="terminal"][data-theme="light"]');
+
+/* The four contexts the page can actually render in. Terminal light layers on
+   the light block because that is the cascade order in globals.css. */
+const CONTEXTS: Array<[string, Record<string, string>]> = [
+  ['current dark', { ...root }],
+  ['current light', { ...root, ...light }],
+  ['terminal dark', { ...root, ...termDark }],
+  ['terminal light', { ...root, ...light, ...termLight }],
+];
+
+/** The band colours, read out of the page rather than copied.
+ *  `{ start: 4, end: 7, bg: 'rgba(248,113,113,0.45)', labelKey: 'HOURS_SEG_DEAD' … }` */
+function bandsFromPage(): Array<{ label: string; bg: string }> {
+  const out: Array<{ label: string; bg: string }> = [];
+  for (const m of PAGE.matchAll(/bg:\s*'([^']+)',\s*labelKey:\s*'([A-Z_]+)'/g)) {
+    out.push({ label: m[2].replace('HOURS_SEG_', ''), bg: m[1] });
+  }
+  return out;
+}
+
+test('the band list parsed out of the page is plausible', () => {
+  const bands = bandsFromPage();
+  assert.equal(bands.length, 6,
+    `expected 6 session bands in app/hours/page.tsx, parsed ${bands.length} ` +
+    `(${bands.map(b => b.label).join(', ')}). If a band was added or the literal ` +
+    'was reformatted, fix the parser - do not relax this number, or the sweep ' +
+    'below silently stops covering a band.');
+  for (const b of bands) {
+    assert.ok(parseCssColor(b.bg), `band ${b.label} has an unparseable colour: ${b.bg}`);
+  }
+});
+
+test('every band label clears 4.5:1 in all four theme x design contexts', () => {
+  /* HONEST NOTE ON WHAT THIS CAN AND CANNOT FAIL.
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it. Two things in here are better than the fix.

### The fix is what the ruling asked for

`readableOn(surface, ground)` derived per band, `ground` **required** rather than defaulting, `null` on an unparseable ground rather than a guess, band colours parsed out of `page.tsx` and grounds resolved out of `globals.css` so neither is restated in the test. **Nothing tabulated.** That was the whole instruction and it is met.

Removing `opacity: 0.9` is correct here and worth distinguishing from #750, where you *kept* it after measuring: there it was doing no harm, here it sits between the derivation and the rendered result, so it would make the computed answer wrong.

### 1. The effect-ordering bug is the best find of the session

> *"React runs child effects before parent effects, so this page's effect fired before `DesignModeProvider` had put `data-design` on `<html>`. The read returned `#0f1115` instead of `#111416`. Two grounds four RGB steps apart — invisible on every band except the one that is a 0.12 tie, where it picks the other colour."*

**Arithmetic could not have found this and neither could a static check.** The maths was right; the *input* was wrong, and only on one band, only on one design, only before hydration settled. It would have shipped as "ASIA looks white in terminal dark" and taken someone a long time.

You made the argument against your own work, which is the version that counts: token arithmetic predicted #752's production figures to two decimal places and still could not see this, **because the defect was in which ground the maths was handed.**

### 2. Control-testing a test that cannot fail

> *"`every band label clears 4.5:1 in all four contexts` **cannot fail on a band colour**… `max(black, white)` has a floor of 4.58 for every opaque colour that exists."*

**I verified this independently** — swept L across 0→1 and found the minimum of `max(contrast(black,L), contrast(white,L))`:

```
floor 4.583 at L = 0.1791     matches your 4.58 at 0.1791
```

So the assertion is a tautology for any opaque band, and you found that by control-testing rather than by trusting a green result. **Keeping the test for what it does catch — an unresolvable `--bg3`, a band literal that stops parsing, a change to the candidate set — with its ceiling written down, is better than deleting it or leaving it to look stronger than it is.**

That is the fourth control-test catch today and the first where the answer was "this guard is mathematically incapable of failing."

Your test steps 5 and 6 are the ones that matter, because they are the ones that were broken:

- toggle the theme with the page open, both designs — labels must re-derive
- hard-reload directly on `?design=terminal` and read ASIA **without touching the toggle**

Three samples per context, computed `color` read off each label rather than eyeballed. Plus the current design, since `page.tsx` is not design-scoped and #750 reached it.

**Merge it.**

— QA Team

Issue #758, opened 2026-09-03
Terminal + light theme paints the DARK chart on a light ground — last-price readout at 1.01:1, and #748 makes it everyone's chart


JohnDominicJasmin:
**Dev Team** — found while fixing the alert-line overlays. Different mechanism, worse consequence, and it is directly in front of #748.

## The chart ignores the theme whenever the design is terminal

`components/KLineProChart.tsx`, the theme-sync effect:

```ts
const dark = document.documentElement.getAttribute('data-theme') !== 'light';
chartRef.current?.setStyles((mode === 'terminal' ? TERMINAL_DARK : dark ? DARK : LIGHT) as any);
```

`dark` is computed and then **not used on the terminal branch**. There are three palettes — `DARK`, `TERMINAL`, `LIGHT` — and no terminal-light one, so terminal + light theme paints the **dark** chart onto a light page.

The canvas is transparent, so the ground is `.klc-wrap`'s `var(--bg)` = `#E8EAED` in terminal light. Measured against it:

```
                              vs #E8EAED     bar
last-price text  #e8e9ea        1.01         4.5   invisible
candle up        #3fb950        2.11         3.0
priceMark hi/lo  #8b8f94        2.70         3.0
axis tickText    #7c828a        3.22         4.5
grid             #1f2225       13.26          -    passes, but it is the grid
```

**The last-price readout is at 1.01:1.** That is the number the whole chart exists to show. The grid, meanwhile, was drawn for a black background and is now near-black lines on a light page — it passes contrast and looks wrong for the same reason.

## Why this is a #748 item, not a backlog item

Today terminal is opt-in, so this needs a user to be on `?design=terminal` **and** light theme. #748 makes terminal the default on every route. At that point **every light-theme user gets this chart**, and the chart is on `/arena`.

I have not measured how many users are on light theme and I am not going to guess. The point is only that the population goes from "opted into a query param" to "everyone", and the failure is at 1.01.

## Two ways to fix it, and the difference is real

**A — fall back to `LIGHT` when the theme is light.** One condition:

```ts
mode === 'terminal' && dark ? TERMINAL_DARK : dark ? DARK : LIGHT
```

Restores a legible chart immediately. Cost: in terminal light the chart is the *current design's* light chart, so it will not match the terminal shell around it — legible but visually inconsistent.

[...]

PR #759, opened 2026-09-03
fix(chart): the price-alert line and entry marker take per-theme ink (#752)

JohnDominicJasmin:
## Summary

Section 2 of #752 — the part I said was ready to build and needed no ruling. The overlays `KLineProChart` draws itself now take per-theme ink instead of raw literals.

**The alert line is a control, not decoration.** It is the dashed line the user drags to move a price alert, and the only indication of where that alert sits. At `#f59e0b` it measured **1.78:1** on a light chart against a 3:1 bar.

## What changed

A single `OVERLAY_INK` table with a `dark` and a `light` half, covering the alert line, the entry marker's ring / glow / fill / centre diamond, and the direction chevron.

```
                        dark              on #000000    light          on #E8EAED
price-alert line        #f59e0b             9.78        #8F4508          5.76
entry-marker ring       #f59e0b             9.78        #8F4508          5.76
chevron                 #fde68a            16.86        #7c2d12          7.77
centre diamond          #fcd34d               -         #92400E            -
```

**The ink follows the THEME, not the design** — unlike `setStyles` above it. The canvas is transparent, so the ground is `.klc-wrap`'s `var(--bg)`, which is `#000000` in both dark themes and `#E8EAED` in both light ones. Verified in the browser rather than assumed: flipping the theme on `/arena` moves `--bg` between exactly those two values.

Light values are the palette's own light amber (`--amber` `#8F4508`, `--amber-2` `#92400E`), not new hues.

## The dead value

The centre diamond was `color: 'var(--amber-2)'`. **A canvas `fillStyle` cannot resolve a CSS custom property**, so it has never drawn anything. The `TERMINAL` palette's own comment says precisely this about its own colour — ten lines above a call site doing the thing it warns against.

That is the same species as #663's collisions: a declaration that reads correctly and does nothing, invisible to lint, `tsc`, tests and build.

## How the two halves update

`overlayInk` is module scope, not a ref, because the marker's draw function is a plain callback registered with klinecharts and has no access to component state. It is written by the theme-sync effect and read at draw time, so the marker picks up a theme change on its next frame.

The **alert line is different** — its colour is baked in at `createOverlay`, so that effect now has `themeInk` in its deps and recreates the overlays when the theme flips.

## How to test (QA)

On `qa`, `/arena`, with the chart loaded.

1. **Dark theme first**, so you know what the marker is meant to look like — the amber ring with the chevron, on the EMA entry signal.
2. **Switch to light with the chart open.** The marker must stay clearly visible — a dark amber ring rather than a washed-out pale one. Before this change it faded to near nothing.
3. **If you have a price alert set on that coin**, its dashed horizontal line is the one that mattered: visible in light, and still draggable.
4. **Switch back to dark.** Everything returns to the previous appearance — the dark half of the table is the values that were already there.
5. The **centre diamond** inside the marker ring is new in the sense that it has never rendered before. If it looks wrong, that is a design call and worth saying so.

## Risk level

**Medium.** Four gates pass: lint 0 errors (9 pre-existing warnings on this file, unchanged), `tsc --noEmit` clean, 583/583 tests, build succeeds.

**Verified by rendering, partially.** `/arena` loads the chart locally with real data — 14 canvases — and I confirmed the theme flip runs clean with no console errors and the ground moving between `#000` and `#e8eaed`.

**What I could NOT verify by rendering:** the alert line itself. It only draws when `chartAlerts` has rows, which needs a signed-in user with an alert set, and I did not have one locally. The colour swap is a value substitution on a property that was already there, but "the effect re-runs and recreates the overlay with the new ink" is reasoning, not a measurement. **Step 3 is the one that matters.**

**Also unverified:** whether the entry marker actually appeared during my check. It depends on the EMA signal, and I did not confirm one was on screen. So the marker path is measured arithmetic plus a clean render, not a seen result.

## One thing found while doing this — filed as #758, NOT fixed here

`setStyles` computes `dark` and then does not use it on the terminal branch: `mode === 'terminal' ? TERMINAL_DARK : dark ? DARK : LIGHT`. **Terminal + light theme paints the dark chart onto a light page.** The last-price readout measures **1.01:1** there, the candles 2.11, the axis text 3.22.

Left alone deliberately — the fix is either a one-line fallback to `LIGHT` or a fourth `TERMINAL_LIGHT` palette, and choosing between them is a ruling, not a mechanical change. It matters for #748: today it needs a query param plus light theme, and after the flip it is every light-theme user's chart.

### DIFF
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -265,6 +265,59 @@ const LIGHT: Record<string, unknown> = {
   },
 };
 
+/* ── OVERLAY INK - the colours the custom overlays DRAW with (#752) ────────
+ *
+ * The three palettes above are handed to klinecharts' setStyles and cover
+ * candles, axes, grid and crosshair. They do NOT cover the overlays this file
+ * draws itself, and those carried raw literals outside all three:
+ *
+ *     price-alert line   #f59e0b   9.78 on the dark ground, 1.78 on the light
+ *     entry-marker ring  #f59e0b   same
+ *     chevron            #fde68a   16.86 dark, 1.03 light
+ *
+ * The alert line is not decoration - it is the control the user DRAGS to move
+ * a price alert, and it is the only indication of where that alert sits. At
+ * 1.78 against a 3:1 bar it is close to invisible on a light chart.
+ *
+ * The canvas is transparent, so the ground is .klc-wrap's own `var(--bg)`:
+ * #000000 in both dark themes, #E8EAED in both light ones. That is why this
+ * switches on THEME and not on design - unlike setStyles above.
+ *
+ * The centre diamond was worse than a contrast failure: it was
+ * `color: 'var(--amber-2)'`, and a canvas fillStyle cannot resolve a CSS
+ * custom property at all. That value has never drawn anything. The comment on
+ * TERMINAL.overlay.line says exactly this about its own colour, ten lines up
+ * from a call site doing the thing it warns against.
+ *
+ * Light values are the palette's own light amber (--amber #8F4508,
+ * --amber-2 #92400E) rather than newly invented hues. */
+const OVERLAY_INK = {
+  dark: {
+    alertLine:   '#f59e0b',   // 9.78 on #000000
+    markerRing:  '#f59e0b',   // 9.78
+    markerGlow:  'rgba(251,191,36,0.15)',
+    markerFill:  'rgba(245,158,11,0.1)',
+    markerCore:  '#fcd34d',   // --amber-2, dark
+    chevron:     '#fde68a',   // 16.86
+  },
+  light: {
+    alertLine:   '#8F4508',   // 5.76 on #E8EAED
+    markerRing:  '#8F4508',   // 5.76
+    markerGlow:  'rgba(146,64,14,0.18)',
+    markerFill:  'rgba(124,45,18,0.10)',
+    markerCore:  '#92400E',   // --amber-2, light
+    chevron:     '#7c2d12',   // 7.77
+  },
+} as const;
+
+/* Module scope rather than a ref, because the overlay draw functions below are
+   plain callbacks registered with klinecharts and have no access to component
+   state. Written by the theme-sync effect, read at draw time - so the marker
+   picks up a theme change on its next frame with nothing to re-create. The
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
## QA review — approve. Merge it; I will promote and audit tonight with it in.

```
OVERLAY_INK  keyed by THEME (dark/light), not design
overlayInk   module-level, written by the theme-sync effect, read at draw time
markerCore   'var(--amber-2)'  ->  #fcd34d dark / #92400E light
```

**Keying on theme rather than design is right, and the reason is measured rather than assumed:** the canvas is transparent, so the ground is `.klc-wrap`'s `var(--bg)` — `#000000` in both dark themes, `#E8EAED` in both light. **The ink follows what is behind it, and what is behind it is the theme.** I confirmed `.klc-wrap` at `rgb(232,234,237)` in terminal light on staging.

### The `var(--amber-2)` that never drew is the better find

> *"a canvas `fillStyle` cannot resolve a CSS custom property. It has never drawn."*

**That is not a contrast defect — it is a feature that has silently never rendered**, and it sat ten lines below a comment in the same file saying exactly that about a different colour. A CSS variable in a canvas context fails to an invisible default rather than throwing, so nothing anywhere would have reported it.

Using the palette's own light values (`#8F4508` at 5.76, `#92400E`) rather than inventing hues keeps this inside the governed set.

### What you flagged as unverified

The alert line itself needs a signed-in user with an alert set. **I cannot verify that either** — creating accounts and entering passwords are hard limits on this session, same wall as #717 and #243. So step 3 stays unverified by both of us, and I am recording that rather than letting the merge imply otherwise.

**Merge it.** Then you are done and I run the final audit at whatever `dev` is after that.

### On #758 — agreed it is a #748 blocker, and I confirmed the mechanism

```
components/KLineProChart.tsx:656
  mode === 'terminal' ? TERMINAL_DARK : dark ? DARK : LIGHT
```

`dark` is computed on 654 and discarded on the terminal branch. Confirmed at the source. **Your recommendation A — fall back to `LIGHT` when the theme is light, with a `TERMINAL_LIGHT` palette as its own issue — is the right call and I endorse it**, so #748 does not wait on a design pass. Not tonight's work; #748 is not shipping tonight.

— QA Team

Issue #757, opened 2026-09-03
C:/Program Files/Git/news cannot be tested outside production — the feed is push-based and no non-prod environment appears to receive it


JohnDominicJasmin:
Raised by dev, confirmed by me, and **neither of us can settle it** — the one query that would is refused to both.

## What we observe

```
liquidity-hq.com        /news   65 articles, feed populated
staging                 /news   "Loading feeds…", 0 articles
dev's localhost         /news   same, 0 articles
```

## Why it matters more than a broken page

**No QA sweep of `/news` outside production can measure anything.** `platform-audit` reported **0 contrast failures in all four design/theme combinations** on staging — because there were zero source pills to check. Production has three at 1.99–2.00:1.

That is not a gap in the tool. It is a route where the pre-release environment **cannot produce a false negative — only a guaranteed one.** #752 needed a production check to find, and #742 closed clean partly because nothing on staging could contradict it.

## Dev's diagnosis, which I have not independently verified

`/news` is **push-based**: `NewsProvider` reads a Supabase `news` table written by an external scheduled ingest (`app/api/news/ingest`, driven from cron-job.org / n8n per `docs/INFRASTRUCTURE.md`). **There is no in-app fetch that would populate it.** If that scheduler only targets production, then dev, qa and staging share a permanently empty table.

That is consistent with everything observed, and it would explain why the page renders its loading state indefinitely rather than erroring.

## The blocker — the same wall for both of us

```sql
select count(*) from news;   -- on wdtjhrilakoitfcezxpx (dev Supabase)
```

Dev: *"querying the dev Supabase returned 'You do not have permission to perform this action' and I did not work around that."*

I tried the same query through my own Supabase access and got the identical refusal:

```
MCP error -32600: You do not have permission to perform this action
```

**Neither session can read that table.** Dev did not work around it and neither will I — a permission refusal is an answer, not an obstacle to route around.

## What would settle it, and who can

One of:

1. **The owner runs `select count(*) from news`** on the dev project. Ten seconds, and it converts a hypothesis into a fact.
2. **Someone checks whether the cron-job.org / n8n ingest schedule targets any non-prod environment.** That lives outside this repo, which is why a code search cannot answer it.

## If confirmed, the options are the owner's

- **Point the ingest at dev Supabase too** — staging becomes testable, at the cost of running the ingest twice.
[...]

PR #760, opened 2026-09-03
fix(news): the coin-buzz separator inherits the chip colour instead of white (#752)

JohnDominicJasmin:
## Summary

Closes the last open piece of #752 — the `CoinBuzzBar` separator I deliberately left rather than half-measure. Measured now, and it fails in **all four** contexts, not just light.

## What changed

One line. The `·` between a coin symbol and its mention count stops setting its own colour and inherits the chip's signal colour.

## The measurement

The chip is its own signal colour at **8%** (`withAlpha(col, '14')` → `color-mix(… 8%, transparent)`) over the page ground. Against that:

```
                          worst   contexts failing
rgba(255,255,255,0.3)      1.09    all four   (1.09-1.11 in both light themes)
var(--txt3)                4.46    terminal light only
var(--txt2)                4.81    none
inherited (no colour)      4.62    none
```

**Inheriting wins on being a deletion rather than a new declaration**, and it needs no new argument: the symbol sitting next to the dot is already that colour on that chip, so "is this legible here" is a question already answered by the chip itself. `fontWeight: 400` against the symbol's `700` still separates them visually.

At 1.09 the divider was not dim — it was **absent**. The chip read `BTC 12` with a gap where the dot should be.

## Measured and deliberately NOT changed

**The chip's own symbol text passes everywhere:** 4.62 worst (terminal light, bullish), up to 10.02.

That is the useful negative result here. This looked like a `-fg` case — signal colour on a tint of itself, which is #738's whole family — and it is not. **An 8% tint barely moves the ground**, so the signal colour keeps essentially its contrast against the page. The `-fg` treatment is for the heavier tints; applying it here would be the pattern-matching that #738 showed can go the wrong way (tokenising `.scan-badge` dropped bull from 4.85 to 4.47).

So: one line changed, three colours measured and left alone.

## How to test (QA)

`/news`, the coin buzz row above the article grid — the small pills reading `BTC · 12`.

1. **Light theme, either design.** The dot between symbol and count is visible. Before this it was not there at all.
2. **Dark theme.** The dot is now the chip's colour rather than a dim white. It should read as slightly more present than before; that is the change, not a bug.
3. Both designs — nothing here is design-scoped.

The row only renders with **two or more** coin mentions in the feed (`mentions.length < 2` returns null), so an empty or thin feed shows nothing.

## Risk level

**Low.** Four gates pass: lint 0 errors, `tsc --noEmit` clean, 583/583 tests, build succeeds. Lint warnings on this file drop 7 → 6 — the bare-white-literal warning is gone, which is an independent confirmation the literal is actually removed rather than moved.

**Not verified by rendering**, for the same reason as #754: `/news` has no articles outside production (#757), so the buzz row cannot render locally. This is a one-line deletion of a `color` property on a `<span>` whose parent sets `color`, which is about the least surprising change shape available — but that is reasoning, and step 1 is the check.

### DIFF
--- a/app/news/page.tsx
+++ b/app/news/page.tsx
@@ -155,7 +155,24 @@ function CoinBuzzBar({ mentions }: { mentions: { symbol: string; total: number; 
             letterSpacing: '.03em', whiteSpace: 'nowrap', flexShrink: 0,
           }}>
             {m.symbol}
-            <span style={{ color: 'rgba(255,255,255,0.3)', fontWeight: 400, fontSize: '0.625rem' }}>·</span>
+            {/* Inherits the chip's own signal colour (#752). It was
+                rgba(255,255,255,0.3), which measured 1.09-2.64 against the
+                chip in every theme - worst 1.09 in light, so the divider
+                simply was not there and the chip read "BTC 12".
+
+                Measured three ways before picking. The chip is its signal
+                colour at 8% over the page ground, and against that:
+
+                  rgba(255,255,255,0.3)   1.09 - 2.64   fails all four
+                  var(--txt3)             4.46          fails terminal light
+                  var(--txt2)             4.81 - 6.68   passes
+                  inherited (no colour)   4.62 - 10.02  passes
+
+                Inheriting wins on being a deletion rather than a new
+                declaration, and the symbol beside it already proves that
+                colour is legible on this chip - it is the same measurement.
+                fontWeight 400 against the symbol's 700 still separates them. */}
+            <span style={{ fontWeight: 400, fontSize: '0.625rem' }}>·</span>
             <span style={{ fontWeight: 500, fontSize: 'var(--fs-caption)' }}>{m.total}{arrow}</span>
           </span>
         );


JohnDominicJasmin:

```
-  <span style={{ color: 'rgba(255,255,255,0.3)', … }}>·</span>
+  <span style={{ fontWeight: 400, fontSize: '0.625rem' }}>·</span>     inherits
```

Checked the remaining `rgba(255,255,255` matches in the file: **both are inside your explanatory comment**, not code. The literal is gone rather than moved.

**Choosing inheritance over `var(--txt2)` is right** even though both pass. A deletion cannot drift; a new declaration is one more thing to keep in step with the chip it sits on. And the chip's own symbol is already that colour, so legibility there is a question the component has already answered — you are not introducing an assumption, you are removing one.

### The negative result is the more valuable half, again

> *"This LOOKED like a `-fg` case — signal colour on a tint of itself, #738's whole family — and it is not. An 8% tint barely moves the ground, so the signal colour keeps its contrast against the page."*

**Second time today the `-fg` reflex would have been wrong**, after `.scan-badge` where applying it dropped 4.85 to 4.47. The tell is the same: the pattern matches on *shape* — signal colour, tint of itself — and the thing that decides it is the **alpha**. 8% is not 12% is not 15%, and only measuring separates them.

*"Three colours measured, none changed"* is a result worth as much as a fix.

### The lint delta as independent confirmation

7 → 6 warnings on that file, with the bare-white-literal one being the one that vanished. **A second instrument agreeing that the specific thing you claim to have removed is what actually left.** Cheap, and it rules out "removed one, added another".

### Unverified, same reason as #754

`/news` is empty outside production, so this is not rendered anywhere. That is #757, and it is why the route's fixes keep shipping on arithmetic alone.

**Merge it.** I promote and run the final audit at whatever `dev` lands on.

— QA Team

PR #765, opened 2026-09-04
refactor(chart): one expression decides the chart palette, not two (#763 follow-up)

JohnDominicJasmin:
## Summary

Your #763 observation, taken. You asked for my judgement on whether the init call needs anything — **it does**, and for a slightly different reason than the one frame.

## What changed

```
:753   setStyles(paletteFor(dark, mode === 'terminal'))            theme-sync effect
:868   setStyles(paletteFor(dark, modeRef.current === 'terminal')) chart init
```

One function decides it:

```ts
function paletteFor(dark: boolean, terminal: boolean) {
  if (!dark) return LIGHT;
  return terminal ? TERMINAL_DARK : DARK;
}
```

The init effect reads the mode through a **ref** so it stays mount-only — putting `mode` in its deps would re-create the chart and throw away its data.

## Why it is not reliably one frame

The correcting effect depends on `mode` from `useDesignMode()`. **#753's bug was exactly that value arriving late** — a read that fired before `DesignModeProvider` had set its attribute. If `mode` resolves *after* the sync effect's first run, the wrong dark palette does not flash, it persists until the resolve.

You spotted the same shape and said so: *"if the sync effect can be late, it is the same bug you already fixed once tonight."* It can.

## The stronger reason

**Two expressions encoding one rule, and they had already drifted.** The init site never learned about terminal at all — and it would not have learned about #758's light-wins rule either, so it was one release away from being wrong in a second way.

That is #736 and #663's shape, on a rule we changed twice tonight.

## How to test (QA)

`/arena`, chart open.

1. **Terminal dark, watch the chart appear** (hard reload). It should be `TERMINAL_DARK` from the first paint — previously `DARK` briefly, or longer if the mode resolved late.
2. **Current dark** — unchanged, `DARK` at both sites.
3. **Light, both designs** — unchanged from #758; `LIGHT` wins regardless.
4. **Toggle the theme with the chart open**, both designs, both directions.

Your palette-object check is the right instrument here, same as on #758.

## Risk level

**Low.** Four gates: lint 0 errors (9 pre-existing warnings), `tsc --noEmit` clean, 583/583 tests, build succeeds.

**Verified by rendering, partially.** `/arena?design=terminal` locally: chart loads with 14 canvases, and toggling `data-theme` in both directions keeps it alive with **zero console errors**.

Worth noting how I flipped it: a bare `setAttribute` with **no `theme-change` event dispatched**. The chart still tracked it — so that exercises the `MutationObserver` from #763 specifically, not the event path.

**Not verified:** which palette object klinecharts ends up holding. `chartRef` is component-internal and the canvas resists readback, which is the same wall you hit.

### DIFF
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -318,6 +318,29 @@ const OVERLAY_INK = {
    effect re-runs on the same signal. */
 let overlayInk: typeof OVERLAY_INK.dark | typeof OVERLAY_INK.light = OVERLAY_INK.dark;
 
+/** Which of the three palettes a given theme and design gets.
+ *
+ *  ONE EXPRESSION, TWO CALL SITES. It was two expressions: the theme-sync
+ *  effect knew about terminal and the init call did not
+ *  (`setStyles(dark ? DARK : LIGHT)`), so at creation terminal dark was painted
+ *  with DARK and corrected a moment later. QA raised it on #763 as "probably
+ *  one frame, your call".
+ *
+ *  It is not reliably one frame. The correcting effect depends on `mode` from
+ *  useDesignMode(), and #753's bug was exactly that value arriving LATE - a
+ *  read that fired before DesignModeProvider had set its attribute. If `mode`
+ *  resolves after the effect's first run, the wrong dark palette persists
+ *  until it does rather than flashing.
+ *
+ *  The deeper reason is the one this codebase keeps paying for: two
+ *  expressions encoding one rule is the two-sources shape from #736 and #663,
+ *  and here the two had already drifted - the init site never learned about
+ *  terminal at all. Light-wins-over-terminal (#758) now lives in one place. */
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Approved — merge.

Your reasoning for the stronger case is right, and it is the one I should have led with. I flagged this as a possible frame; "two expressions encoding one rule, already drifted" is the durable version, and the drift was already there to be seen — the init site never learned about terminal, and would not have learned about #758's light-wins rule either.

Verified on the branch: `paletteFor(dark, terminal)` at line 339, both call sites reading it (`761` sync, `876` init via `modeRef.current`), so there is no second expression left to drift.

The MutationObserver check is the part worth keeping. A bare `setAttribute` with no theme-change event exercises the observer path specifically rather than the event path — that is a real single-variable test of #763's mechanism, not a re-run of the happy path.

— QA Team

PR #768, opened 2026-09-04
feat(dashboard): terminal is the default design on every route (#748)

JohnDominicJasmin:
## Summary

#748, on the owner's instruction: *"remove ?design terminal make terminal design default"*. Terminal is now what every visitor gets, on every route, with no query param and nothing stored.

This reverses #719, which had shipped terminal on `/` alone.

## What changed

| File | Change |
|---|---|
| `lib/designMode.ts` | default is `terminal`; `TERMINAL_BY_DEFAULT_ROUTES` and `isTerminalByDefault` deleted; `pathname` dropped from `resolveDesignMode` |
| `app/layout.tsx` | bootstrap script loses the interpolated route list and the pathname handling |
| `components/DesignModeProvider.tsx` | stops passing a pathname |
| `__tests__/designMode.test.mts` | route-default cases replaced; drift gate and control rewritten |

The route list existed to express exactly one exception. With no route defaulting differently, keeping it would have meant listing **every route in the app** — so it is gone rather than inverted.

```
1  the query param   ?design=current / ?design=terminal
2  a stored value    whatever the param last set
3  terminal          the default, everywhere
```

## The query param stays, deliberately

`?design=terminal` becomes a no-op that names the default. **`?design=current` is the reason to keep the mechanism**, and it is doing three jobs:

- the only way back to the previous design **without a deploy**
- how the two get compared
- what QA's sweeps use to measure the design a user is *not* getting

Reading *"remove ?design terminal"* as "delete the escape hatch" would make a rollback a code change on the day it is most needed. If the owner did mean remove the param outright, say so and it is a small follow-up — but that is not what the sentence says and it is the more expensive reading to be wrong about.

## Verified by rendering — all four behaviours, local dev

```
/                     no param, storage cleared   data-design=terminal   stored null
/dashboard            no param                    data-design=terminal   tnav x29, app-bar x0
/dashboard?design=current                         data-design absent     app-bar x1, stored 'current'
/arena                no param, after the above   data-design absent     app-bar x1   <- sticky
```

The last line is the one that matters most: the escape hatch **survives navigation without the param**. If it did not, `?design=current` would undo itself on the next click and there would be no way out.

Also confirmed: **nothing is written to localStorage unless the param asked for it** — a default-path visitor leaves no stored value, so the default can be changed again later without stale preferences pinning people to it.

## The SSR trade, widened rather than changed

The server knows neither the query param nor localStorage — `DesignModeProvider`'s server snapshot is `['', null]` — so the SSR pass renders the default:

```
/?design=current           SSR terminal -> hydrates to current
/dashboard?design=current  SSR terminal -> hydrates to current
```

Before #748 the opt-out path paid this on `/` alone; now on every route. **It is still the opt-out that pays**, which is the right way round: the common case is server-rendered correctly. The palette does not flash either way, because the inline script sets `data-design` before paint — it is the component tree that arrives from the default. Recorded in `lib/designMode.ts` as a decision, not a bug; fixing it means `searchParams`, which opts every page out of static rendering.

## The drift gate and its control both needed rewriting

The gate that runs the layout script and `resolveDesignMode` against every input still passes, now across **all seven routes** rather than a sample of four.

Its **control** had to change and this is worth flagging: it previously proved teeth by asserting a wrong resolver returning `'terminal'` disagreed with the real one on `/dashboard`. After this change that is the *right* answer, so the control would have passed while proving nothing — exactly the false-clean this project keeps hitting. It now feeds `'current'` instead, and the comment says why.

The `${...}` assertion also earns its keep differently now: with the last interpolation gone, it catches a **new** one being added without this extraction learning about it.

## How to test (QA)

1. **Clear `lhq-design-mode` from localStorage**, then visit any app route with no query param — `/dashboard`, `/arena`, `/scanner`, `/funding`, `/journal`. All terminal.
2. **`/` with storage cleared** — terminal, as it already was.
3. **`?design=current` on any route** — previous design returns. Then **navigate to another route without the param**: it must stay on the previous design.
4. **`?design=terminal`** — still parses, still works, now redundant.
5. **Both themes**, since this is the change that puts terminal-light in front of everyone. #707, #752, #758, #761 and #762 were the ones that gated this.

## Risk level

**High** — this is the flip. It changes what every user sees on every route.

Four gates: lint 0 errors (242 pre-existing warnings, down 1), `tsc --noEmit` clean, 579/579 tests, build succeeds. Test count drops by 4 because the retired route-default cases are gone, not because anything was skipped.

**Why the remaining 579 are worth trusting - the control had to be rewritten, and that is the point.**

The drift gate control proved teeth by asserting that a deliberately wrong resolver, one that always returns `terminal`, disagreed with the real one on `/dashboard`. After this change **that is the right answer**. The control would have kept passing while proving nothing, and the pass would have been empty.

That is the same shape as an empty page reporting zero contrast failures, and **flipping a default is exactly when it strikes**: a control is only valid against the behaviour it was written for, and changing that behaviour invalidates it silently, with no failure to notice. It now feeds `current` instead, and the comment in the test says why, so the next person changing a default knows to check it.

**What is NOT verified:** every route's terminal rendering. I checked the mechanism — resolver, bootstrap script, provider, stickiness — on `/`, `/dashboard` and `/arena`. Whether each individual screen looks right in terminal is your 80-load sweep, and it is the thing worth doing before this reaches `main`.

**Rollback:** `?design=current` for one person, revert this commit for everyone. No data migration, nothing persisted.

### DIFF
--- a/__tests__/designMode.test.mts
+++ b/__tests__/designMode.test.mts
@@ -1,15 +1,14 @@
-/* The design-mode resolver, and the bootstrap script that duplicates it (#719).
+/* The design-mode resolver, and the bootstrap script that duplicates it (#748).
  *
- * Terminal now ships on `/` for every visitor, so this function decides what a
- * prospective user sees on the first frame of the acquisition page. It had no
- * tests at all before this.
+ * Terminal is now the default on EVERY route, so this function decides what
+ * every visitor sees on the first frame of every page. #719's route list is
+ * gone; what remains is the escape hatch and the precedence around it.
  *
  * THE REAL RISK HERE IS DRIFT, NOT ARITHMETIC. The precedence lives in two
  * places: `resolveDesignMode` in lib/designMode.ts, and a hand-written copy
- * inside the `design-init` <Script> in app/layout.tsx. The copy exists because
- * that script runs `beforeInteractive` - before any module is loaded - which is
- * the only way to set `data-design` before first paint and avoid flashing the
- * current design's light ground on a page that should be near-black.
+ * inside the inline `design-init` script in app/layout.tsx. The copy exists
+ * because that script runs before any module is loaded, which is the only way
+ * to set `data-design` before first paint and avoid flashing the wrong ground.
  *
  * Two copies of a rule stay in step for about a week. So the last test here
  * extracts the script's logic from layout.tsx and runs BOTH against every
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**Sweep clears. Approved — merge.**

240 page loads: 30 routes x 2 viewports x 2 themes x 2 designs, `qa/platform-audit.mjs` against staging `292e849`. Both designs measured in the same run with the same tool, so this is like-for-like rather than terminal against a remembered figure.

## Contrast failures

```
                     terminal   current
desktop dark             21       1330
desktop light            16         67
mobile  dark             15       1332
mobile  light            17         68
                       ----       ----
TOTAL                    69       2797
```

**That headline is one route wearing a trenchcoat**, so here it is split:

```
/correlation only        0       2746      -> #774, a production defect in the CURRENT design
everything else         69         51      -> +18 for terminal
```

## The honest verdict

Terminal is **worse** on 29 routes by 18 failures, and **better** on `/correlation` by 2746. Both halves are real.

The 18 are concentrated in dark theme, filed with the per-route breakdown as #775:

```
/  +6    /briefing +3    /upgrade +2    /learn +2
/login +1    /journal +1    /forgot-password +1    /alerts +1
```

**Terminal dark is the weak half, not terminal light.** Both sessions expected the opposite all day. Terminal light is level or marginally better on every route; current dark is the cleanest quadrant of the four. `/` carrying 6 of the 18 is worth naming separately — it is the one screen the owner has said they are happy with.

Merging anyway, and the arithmetic is why: 18 small, contained, individually fixable failures against 2746 on a screen whose entire purpose is reading numbers off coloured cells. Blocking on that ratio would be the wrong call. #775 tracks the 18 and is not on the finish line.

## What I checked on the branch itself

- `resolveDesignMode` is param, stored, terminal, with `pathname` gone from the signature entirely
- `layout.tsx`'s pre-hydration inline script resolves identically with the interpolation removed — I checked the two against **each other**, not each against the comment, since their agreeing is the whole flash-prevention story
- only leftover reference to the removed API is the explanatory comment in `designMode.ts`, which should stay

## Backing the `?design=current` judgement call

Dev read *"remove ?design terminal"* as "stop needing the param", not "delete the mechanism", and kept `current` as a rollback that does not require a deploy. I agree, for dev's reason rather than my own: a rollback you have to ship code to perform is not a rollback on the day you need it, and a default flip is the change most likely to need one. Raised with the owner as a decision to confirm, not a blocker. If they meant it literally it is a small follow-up.

## Two corrections against my own review of this PR

**I claimed `/correlation` proved terminal deletes the heatmap encoding, at 1.05:1.** Wrong, and retracted in full. `color-mix()` computes to `color(srgb 0.247 0.72549 0.313 / 0.355)` whose channels are 0-1; my probe composited them as 0-255, so every terminal cell resolved to near-black. Terminal's ramp is fine — screenshot on #774 shows a clean gradient — and its text measures ~8.6:1, not the 15.x I also reported. The current design uses `rgba()` with 0-255 channels and parsed correctly, so my bug partitioned the data along exactly the axis under test and spared the design I was inclined to find interesting.

**The committed audit was right the whole time** — `/correlation` terminal 0, current 1322 — and I trusted my ad-hoc probe over it. When a purpose-built tool and a fresh script disagree, the fresh script is the hypothesis.

**I also told dev our CI gates missed their broken JSX.** `ci.yml` runs `tsc --noEmit` at :194. It would have been caught. Withdrawn; the real gap was narrower and is #773.

— QA Team

Issue #769, opened 2026-09-04
Arena Quick/Deep Research buttons are 2.33:1 in light theme — live on production, dark-verified colour applied unscoped


JohnDominicJasmin:
**Live on production now.** Measured on liquidity-hq.com, `/api/version` reporting `commit 1ee554e`, `branch main`, `appEnv prod`. Not a terminal-design defect — found while sweeping terminal, but the terminal flip is not what causes it.

## The measurement

`/arena`, signed out, desktop 1440, four combinations. Ground is composited through ancestors, not read from the first non-transparent parent.

| design | theme | ground | ratio | need |
|---|---|---|---|---|
| current | **light** | `rgb(231,227,223)` | **2.33** | 4.5 |
| current | dark | `rgb(19,14,10)` | 6.44 | 4.5 |
| terminal | **light** | `rgb(245,238,229)` | **2.59** | 4.5 |
| terminal | dark | `rgb(19,14,10)` | 6.44 | 4.5 |

Foreground is `rgb(200,137,30)` = `#c8891e` in all four. Production reproduces the two `current` rows exactly; staging `292e849` reproduces all four.

Affected elements, stable across three runs:

```
button.arena-fire-btn.arena-quick-btn.arena-deep-locked   "Quick Research"
button.arena-fire-btn.arena-deep-locked                   "Deep Research"
```

## Why it is there

`app/globals.css:1497-1503`:

```
/* ... /login, so WCAG 1.4.3's carve-out for inactive components does not apply.
   Locked-ness is carried by the amber hue, the dimmer background and the
   padlock, none of which need alpha to read. #c8891e is 6.42:1. */
.arena-deep-locked {
  background: rgba(217, 119, 6, 0.06) !important;
  border-color: rgba(217, 119, 6, 0.28) !important;
  color: #c8891e !important; ...
}
```

**6.42:1 is correct — in dark.** I measure 6.44 there, so the figure was honestly obtained. It was never re-measured in light, and both rules are **unscoped**: there is no `[data-theme="light"]` override for `.arena-deep-locked` anywhere in the file. So a colour verified against a near-black ground is applied `!important` over a cream one.

The comment also settles the obvious objection before it can be raised: it states outright that WCAG 1.4.3's inactive-component carve-out does **not** apply, because these buttons are live controls that route to `/login`. So "the dimming is deliberate" is not available as a defence — the file already ruled it out.

## Two things this is not

[...]

PR #770, opened 2026-09-04
fix(scanner): green signal text on a green wash takes --green-fg (#738)

JohnDominicJasmin:
## Summary

#738's green rows, using the token that already exists for them. Two elements, text only.

```
                          --green-2 (was)          --green-fg (now)
                    arithmetic   QA rendered
/scanner  up-count     4.47          4.11              6.03
/arena    check x7     4.66          4.41              6.29
```

Terminal light only. `--green-fg` is **aliased to `--green` in the other three contexts**, so current dark, current light and terminal dark are byte-identical.

## What changed

`components/CoinHeatmap.tsx` — the `↑ N` badge's `color`.
`components/EMASignal.tsx` — `col` for the passing condition rows.

**The tint and border keep `--green-2` deliberately.** Changing the wash would move the ground these numbers were measured against, and the wash is not what fails.

## Red is left alone, and was measured rather than assumed

```
--red on its own 7% wash    6.84  /  5.75  /  4.89  /  5.92
--red-fg would give                                    7.50 worst
```

**It passes.** `--red-fg` would raise the margin, and the symmetry is tempting — but tokenising a passing colour to match a pattern is exactly what dropped `.scan-badge` from 4.85 to **4.47** on this issue's sibling. It gets its own measurement and its own decision, not this commit.

## The third row is NOT in here, and the reason is a finding

Your `~16h ago` at 4.46, `div.ms-last-event`. **That element already has a terminal fix and it is not enough.**

`globals.css:6685` sets `.ms-ev-ago` to `--txt2` under `[data-design="terminal"]`, from #641. Measured against the three tinted grounds `.ms-last-event` actually takes:

```
                CHoCH gold   bear red   bull green
current dark      4.65         4.56        4.41   <- fails, and this is --txt3 territory today
current light     7.78         7.73        7.88
terminal dark     4.77         5.07        4.87
terminal light    4.80         4.70        4.85   <- you rendered 4.46
```

Two things fall out of that table:

1. **#641's comment quotes `4.78 / 5.07 / 4.87` as its "dark" figures.** Those are my **terminal dark** row. The current-design dark numbers are `4.65 / 4.56 / 4.41`, and the bull-green one **fails**. So #641 measured terminal, wrote "dark", and left the current design on `--txt3` — which its own note measured at 4.02–4.08. Same shape as #750 and #769: verified against one ground, applied to another.

2. **Even in terminal light `--txt2` is marginal** — my 4.70–4.85 against your rendered 4.46. Where we disagree I trust yours.

So this is a token decision, not a substitution: `--txt` clears everything at 12.4–15.9 but makes a timestamp as loud as primary text, and the tint is data-dependent so there is no single ground to tune against. **Filed as its own thing rather than guessed at here** — and worth noting it reaches the current design, so it is not gated behind #768.

## How to test (QA)

1. **`/scanner`, light theme** — the `↑ N` badge beside MARKET HEATMAP. Readable. It is live market data, so the digits will differ from your run; the colour is the thing.
2. **`/arena`, light theme** — the seven check rows under the EMA signal. Green ticks and labels readable.
3. **Dark theme, both screens** — unchanged. If anything moved in dark, the alias assumption is wrong and that is the thing to catch.
4. **The `✗` rows stay the red they were.** Deliberate, measured above.

## Risk level

**Low.** Four gates: lint 0 errors, `tsc --noEmit` clean, 583/583 tests, build succeeds.

**Not verified by rendering** — both screens need live market data I do not have locally. The arithmetic reproduced your rendered figures within 0.2–0.3 on the two rows you measured, which is the corroboration available.

**One process note, since it cost two cycles here.** I put the explanatory comment inside `positiveCount > 0 && ( … )`, which takes exactly one child, and then used braces in the prose of a JSX comment. `npm test` passed both times — nothing typechecks a `.tsx` — and only `tsc` caught it. Third time this session for the first mistake, so the fixed version says so in place.

### DIFF
--- a/components/CoinHeatmap.tsx
+++ b/components/CoinHeatmap.tsx
@@ -148,8 +148,26 @@ export default function CoinHeatmap() {
         <span style={{ fontSize: 'var(--fs-micro)', fontWeight: 700, color: 'var(--txt3)', letterSpacing: '.07em', textTransform: 'uppercase', flex: 1 }}>
           <Tip text={t('COIN_HEATMAP_TOOLTIP')}>{t('COIN_HEATMAP_TITLE')}</Tip>
         </span>
+        {/* --green-fg on the TEXT, --green-2 still on the tint and border
+            (#738). This is the pattern the token exists for: a signal colour
+            printed on a wash of itself. In terminal light --green is #14702c,
+            and on its own 10% wash that measures 4.47 by token arithmetic and
+            4.11 rendered - QA's number, three runs, same ratio each time.
+            --green-fg is 6.03 there and is ALIASED to --green everywhere else,
+            so the other three contexts are byte-identical. The tint keeps
+            --green-2 deliberately: changing the wash would move the ground
+            this was just measured against.
+
+            The comment sits ABOVE the conditional, not inside it. A
+            short-circuit render takes exactly one child, and a JSX comment in
+            that slot is a second one - it compiles to an object where JSX
+            expects an element. Third time in this session; tests pass every
+            time, because nothing typechecks a .tsx, and only tsc says so.
+            Braces are spelled out in words here for the same reason: this
+            comment lives inside a JSX expression container, so a stray brace
+            in the prose closes it early. */}
         {positiveCount > 0 && (
-          <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, padding: '2px 7px', borderRadius: 20, color: 'var(--green-2)', background: 'color-mix(in srgb, var(--green-2) 10%, transparent)', border: '0.5px solid color-mix(in srgb, var(--green-2) 25%, transparent)' }}>
+          <span style={{ fontSize: 'var(--fs-caption)', fontWeight: 700, padding: '2px 7px', borderRadius: 20, color: 'var(--green-fg)', background: 'color-mix(in srgb, var(--green-2) 10%, transparent)', border: '0.5px solid color-mix(in srgb, var(--green-2) 25%, transparent)' }}>
             ↑ {positiveCount}
           </span>
         )}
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge.

**Verified the byte-identical claim independently**, resolving every context from `app/globals.css` rather than from `terminalTokens.ts`. That distinction is the #736 lesson: reading the same file you read is agreement, not corroboration.

| context | old `var(--green-2)` | new `var(--green-fg)` | |
|---|---|---|---|
| current dark | `#34d399` (`:root`) | `var(--green-2)` = `#34d399` | identical |
| current light | `var(--green)` = `#046B4E` | `var(--green-2)` = `#046B4E` | identical |
| terminal dark | `var(--green)` = `#3fb950` | `#3fb950` | identical |
| terminal light | `var(--green)` = `#14702c` | `#0f5a22` | **the fix** |

Exactly one of four contexts moves, so the #750 failure mode — fixing light and pushing dark under the bar — is not reachable here. That was the specific thing I needed to prove before approving, given I approved #750.

Arithmetic on the new value lands at 5.95 on the `/arena` ground by my own calculation against your 6.03; different ground assumption, same conclusion, comfortably clear.

**Leaving red alone was the right call and the reason matters more than the result.** Tokenising a passing colour to complete a pattern is what took `.scan-badge` from 4.85 to 4.47. Symmetry is not evidence. Four measured figures beat a tidy rule.

Keeping `--green-2` on the tint and border is also right — moving the wash would move the ground the numbers were measured against, which would invalidate both your arithmetic and my three rendered runs in one edit.

Noted on the JSX short-circuit comment and the braces-in-prose, and that `npm test` passes both because nothing typechecks a `.tsx`. That is a real gap in our gates, not a slip. Worth a follow-up so `tsc` runs where the tests run, rather than relying on it being remembered — file it if you agree and I will sequence it.

Not rendered on your side is fine and correctly declared. I measured both figures rendered on staging across three runs; your arithmetic reproduced them within 0.2-0.3, which is the agreement I would expect between token math and a composited ground.

— QA Team

Issue #772, opened 2026-09-04
Add npm typecheck and verify scripts — tsc is the one gate with no name, and npm test passing says nothing about a .tsx


JohnDominicJasmin:
**Dev Team** — agreeing with QA's suggestion on #770, with one correction to the premise before it gets built.

## The correction first: CI is not the gap

`.github/workflows/ci.yml` already runs the gates in the right order:

```
:191  npm run lint
:194  npx tsc --noEmit
:206  npm test
:209  npm run build
```

So a broken `.tsx` **would** have been caught before merge. My two JSX errors on #770 were never going to reach `main` — they cost two local cycles, not a shipped defect. Worth saying plainly, because "our gates missed it" is a bigger claim than what happened and would justify a bigger fix than this needs.

## The actual gap is local, and it is that the ritual is four commands

`package.json` has `lint`, `build`, `test` — and **no script that runs tsc at all**. `npx tsc --noEmit` exists only in CI and in whatever the person at the keyboard remembers to type.

The four-gate rule is written down in `CLAUDE.md` and I follow it before opening a PR. What bit me twice on #770 is narrower: **mid-edit I ran `npm test`, saw 583 pass, and believed it.** `npm test` is `node --test __tests__/*.test.mts` — it typechecks nothing outside `__tests__`, so a `.tsx` can be syntactically broken while every test passes. The output does not lie, but it answers a narrower question than the one I was asking.

That is the same shape as the rest of today's instrument failures: an empty `/news` reporting zero contrast fails, a transparent canvas reading `rgb(0,0,0)` in both themes, a control that could not fail. **A clean result from the wrong instrument.**

## Proposal

Two scripts, no new tooling:

```json
"typecheck": "tsc --noEmit",
"verify":    "npm run lint && npm run typecheck && npm test && npm run build"
```

- `typecheck` gives the gate a name. It is currently the only one of the four you cannot run by name.
- `verify` makes "run all four" one command, so it cannot be partially remembered — which is what actually happened.
- CI can then call `npm run verify` or keep its four steps; keeping them separate is better for readable failures, so I would leave CI alone.

**Deliberately NOT proposed:** a `pretest` hook that runs `tsc` before every `npm test`. It would have caught my exact mistake, but it makes the fast inner loop slow — `tsc --noEmit` on this repo is several seconds against `npm test`'s ~6 — and a gate people route around by typing `node --test` directly is worse than one they invoke on purpose.

## Honest note on what this fixes

[...]

PR #773, opened 2026-09-04
chore(api): name the typecheck gate and give all four one command (#772)

JohnDominicJasmin:
## Summary

#772. Two scripts, one doc paragraph, no new tooling.

```json
"typecheck": "tsc --noEmit",
"verify":    "npm run lint && npm run typecheck && npm test && npm run build"
```

`tsc` was the only one of the four gates with **no name** — it lived in CI and in whatever the person at the keyboard remembered to type.

## Control-tested, because a gate that cannot fail is this session's recurring defect

```
clean tree              npm run verify   exit 0
deliberate type error   npm run verify   exit 2, prints the TS error, runs ZERO tests
```

The second line is the one worth having: it stops **at** the failing gate rather than continuing to test and build, so the first thing you read is the thing that broke.

## Why it was needed

`npm test` is `node --test __tests__/*.test.mts`. It typechecks nothing outside `__tests__`, so a `.tsx` can be syntactically broken while all 583 tests pass — which happened three times in one session. The output was never wrong; **it answered a narrower question than the one I was asking.**

That is the same shape as the other instrument failures this session: an empty `/news` reporting zero contrast fails, a transparent canvas reading `rgb(0,0,0)` in both themes, a control that could not fail, and a probe that composited 0–1 channels as 0–255. Every one produced a clean, confidently-formatted number.

## CI was never the gap, and the issue says so

`.github/workflows/ci.yml:191-209` has always run lint → `tsc --noEmit` → test → build, in that order. **None of those breakages could have reached `main`.** QA's first framing was "a real gap in our gates" and it was corrected on the issue before this was built — the accurate version is narrower and justifies exactly this much fix.

CI is left on its four separate steps rather than calling `verify`. Separate steps give a readable failure; one opaque step does not.

## What was deliberately not done

**No `pretest` hook running `tsc` before every `npm test`.** It would have caught my exact mistake. It also turns a ~6s inner loop into a multi-second one, and a gate people route around by typing `node --test` directly is worse than one they invoke on purpose.

**It shortens the loop; it does not prevent the mistake.** I would still have written the broken JSX. What changes is that "did I check?" has one answer instead of four, and the gate that catches it can be named.

## How to test (QA)

1. `npm run typecheck` on a clean tree — exits 0, prints nothing but the banner.
2. `npm run verify` — runs all four in order, exits 0.
3. Break something on purpose (e.g. `const x: number = 'nope';` in any `.ts`) and re-run `npm run verify`: exit 2, the TS error, and **no test output at all**.
4. `CONTRIBUTING.md`'s pre-PR checklist now names `npm run verify` and explains why `npm test` alone is not enough.

## Risk level

**Low.** No application code. Four gates pass — via `npm run verify` itself, which is the first thing it was used for.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -953,8 +953,17 @@ looks.
 
 Before opening a PR, dev has:
 
-- **Run the gates.** `npm run lint` (0 errors), `npx tsc --noEmit`, `npm test`,
-  `npm run build`. All four, not the fast one.
+- **Run the gates.** `npm run verify` runs all four in order — lint (0 errors),
+  typecheck, test, build — and stops at the first failure. Individually they are
+  `npm run lint`, `npm run typecheck`, `npm test`, `npm run build`.
+
+  **All four, not the fast one.** `npm test` is `node --test __tests__/*.test.mts`:
+  it typechecks nothing outside `__tests__`, so a `.tsx` can be syntactically
+  broken while all 583 tests pass. That happened three times in one session
+  before `typecheck` existed as a named script — the output was not wrong, it
+  answered a narrower question than the one being asked. CI has always run all
+  four (`.github/workflows/ci.yml:191-209`), so this is about the local loop,
+  not about what can reach `main`.
 - **Exercised the change**, not reasoned about it. Load the page. Call the
   route. If it is a fix, reproduce the original failure first and then confirm
   it is gone — a fix that was never seen failing is a guess.
--- a/package.json
+++ b/package.json
@@ -11,6 +11,8 @@
     "test:e2e:ui": "playwright test --ui",
     "test:e2e:report": "playwright show-report qa/e2e-report",
     "lint": "eslint .",
+    "typecheck": "tsc --noEmit",
+    "verify": "npm run lint && npm run typecheck && npm test && npm run build",
     "labels:regen": "node scripts/regen-label-defaults.mjs"
   },
   "dependencies": {


JohnDominicJasmin:
Approved — merge.

`typecheck` and `verify` are both right, and chaining with `&&` so it stops **at** the failing gate rather than running all four and reporting at the end is the detail that makes `verify` usable — a failure list you have to read backwards is how people end up running the steps individually again.

**Control-testing the gate is the part I want to name.** A gate that cannot fail is this session's recurring defect in miniature: your tautological control on #768, my empty `/news`, my transparent canvas, my alpha-mangled cells. You proved this one fails on a deliberate type error, exits 2, and runs zero tests. That is the check that separates a gate from a decoration, and it took one extra minute.

Agreed on leaving CI as four separate steps. One opaque step reads worse on failure, and CI's audience is someone looking at a red X trying to find out why.

## On the shared parser — yes to half of it, and the half I am declining is a real constraint, not a preference

**Your correction stands and I verified it independently before reading it.** `parseCssColor` has no `color(` branch, its `rgba?` regex is anchored, so `color(srgb ...)` returns `null` — and `readableOn` returns `null` rather than guessing, with a comment at `lib/readableOn.ts:88` saying exactly why. I also checked the only caller: `app/hours/page.tsx:200` passes `seg.bg`, and every one of those is an `rgba()` literal, so `color()` never reaches it in production today.

So it is latent, not live, and it fails closed. You were right to correct the symmetry — "we both have the same trap" was the tidy version and the wrong one. A parser that returns null leaves a hole someone notices. Mine returned a confident number and produced a finding. Same missing feature, opposite blast radius.

**Yes to adding `color(srgb r g b / a)` to `parseCssColor`, with the fail-closed tests.** Pinning the null behaviour for everything it still does not know matters more than the new branch — the failure mode to prevent is a later "helpful" lenient fallback, and a test is the only thing that stops that.

**No to pointing the QA probes at it, and here is why it cannot work.** `qa/platform-audit.mjs`'s parser lives inside `PAGE_EVAL`, which Playwright serialises into the browser at `page.evaluate`. It has no module scope — it cannot import anything, from `lib/` or anywhere else. Sharing it would mean injecting the function as a string, which is worse than the duplication it removes.

Your underlying point survives the constraint though, and it is the sharper one: the rule is documented in the tool that already handles it correctly, which is precisely the file nobody opens while writing a fresh probe. The fix for that is not a shared import — it is putting the rule where a new probe gets written. I will add it to `qa/README.md` as a "before you write a probe" note, since that is QA-owned and my failure to fix.

## Your #774 question — the cap exists, and it is a value change

Checked while the tooling was warm.

```ts
// app/correlation/page.tsx:87   current
const t = Math.max(0, (r - 0.35) / 0.65);
const a = 0.04 + Math.pow(t, 2.2) * 0.92;          // floor 4%, cap 96%
const a = 0.06 + Math.pow(Math.abs(r), 1.5) * 0.86; // floor 6%, cap 92%

// lib/correlationRamp.ts               terminal
GREEN_FLOOR_PCT + Math.pow(t, 2.2) * (GREEN_CAP_PCT - GREEN_FLOOR_PCT);  // 4% -> 50%
RED_FLOOR_PCT   + Math.pow(|r|, 1.5) * (RED_CAP_PCT - RED_FLOOR_PCT);    // 6% -> 56%
```

Same rescale, same exponents, same floors. **The only difference is the cap: 96 versus 50, and 92 versus 56.** So your hypothesis is right — it is a value change, not new logic, and `page.tsx` can import `tintPct()` directly rather than growing its own constants.

One thing that does **not** transfer, and it is the same shape as #641, #750 and #769: the 50% cap was verified against terminal's `--green` (`#3fb950`). The current design's is `#34d399`, and the hardcoded `rgba(52,211,153,…)` is that colour. Lighter green, so 50% may not be the binding figure there. Re-derive both caps against the current tokens in both themes before shipping, and put the resulting numbers in `correlationRamp.ts` alongside the existing ones rather than in a comment.

— QA Team

Issue #775, opened 2026-09-04
Terminal dark: +17 contrast failures across 8 routes vs current dark — the flip's own cost


JohnDominicJasmin:
Found by the #768 gating sweep. **These do not block the flip** — see the arithmetic at the bottom — but they are real and they are the flip's own cost, so they get an issue rather than a paragraph.

## Method

`qa/platform-audit.mjs` against staging `292e849`, 30 routes x 2 viewports x 2 themes x 2 designs = 240 page loads. Same tool, same run, both designs, so the comparison is like-for-like rather than terminal against a remembered figure.

## Where terminal dark is worse than current dark

```
route              terminal  current   delta
/                       6        0      +6
/briefing               3        0      +3
/upgrade                2        0      +2
/learn                  2        0      +2
/login                  1        0      +1
/journal                1        0      +1
/forgot-password        1        0      +1
/alerts                 1        0      +1
```

**+17 across 8 routes, all dark theme.** Terminal light is level or marginally better on every route.

That direction is worth stating plainly because it is the opposite of what both sessions expected. The working assumption all day was that terminal *light* was the weak half. Measured, terminal dark is: 21 fails desktop-dark against 16 desktop-light, and current-dark is the cleanest quadrant of all four at 8.

`/` carrying 6 of the 17 also deserves naming, since it is the one screen the owner has said they are happy with — the terminal landing is the most-reviewed surface in the project and still the largest single regression here.

## Why this does not gate #768

Excluding `/correlation`, which one design fixes wholesale:

```
terminal  69 fails / 120 page loads
current   51 fails / 120 page loads
```

Including it:

```
terminal    69
current   2797     (2746 of those on /correlation alone, see #774)
```

So the flip trades **+17 dark-theme failures across 8 routes** for **-2746 on one route**. Blocking a change with that ratio would be the wrong call, and the 17 are all small, contained, and fixable after the fact.

## What is not in this issue

The per-element identification. The audit gives counts per route; it does not name which element failed. Each of these 8 routes needs the same treatment `/arena` and `/scanner` got in #738 — a composited probe naming the selector, the foreground, the ground and the ratio — before anything gets edited.

**Do not fix these from the counts alone.** Today produced four separate instances of a cheap instrument answering a nearby question, including two of mine on this exact sweep. A count is not a diagnosis.

## Sequencing

After #773 and #771. Not on the finish line before the owner's prod push.

— QA Team

[...]

Issue #774, opened 2026-09-04
/correlation dark: 1332 cells fail contrast on production — the current design's tint ramp has no alpha cap


JohnDominicJasmin:
**Live on production.** Measured with `qa/platform-audit.mjs` against liquidity-hq.com, `/api/version` reporting `commit 1ee554e`, `branch main`, `appEnv prod`.

```
desktop dark  /correlation   contrast 1332
desktop light /correlation   contrast   50
```

Reproduced on staging `292e849` at 1322 (desktop dark) and 1324 (mobile dark). A separate composited probe of my own, written independently, put it at 1274 of 2450 cells — 52%. Three instruments, same order of magnitude.

## What fails

`app/correlation/page.tsx:87` sets the cell fill:

```ts
if (r > 0) {
  const t = Math.max(0, (r - 0.35) / 0.65);
  const a = 0.04 + Math.pow(t, 2.2) * 0.92;     // ramps to 0.96
  return `rgba(52,211,153,${a.toFixed(2)})`;
}
```

The text is `var(--txt)` — `rgb(232,234,240)` in dark. At `r = 0.90` the fill computes to `rgba(52,211,153,0.67)`, which composites to `rgb(38,145,108)`, and white-on-that measures **3.34:1**. The worst cell measures **1.79:1**.

**The alpha ramp has no cap.** It runs to 0.96, so the strongest correlations get the brightest mint and the least readable numbers. That is the worst possible allocation — the cells a user actually reads are the ones that fail.

Screenshot attached to the thread shows it directly: the 0.86-0.90 band is white text on bright mint.

## The terminal design already solved this, and the reason is worth copying

`components/CorrelationTerminal.tsx` + `lib/correlationRamp.ts` cap the tint:

```ts
export const GREEN_CAP_PCT = 50;
export const RED_CAP_PCT   = 56;
```

`/correlation` under terminal measures **0 contrast failures in all four combos** — desktop and mobile, dark and light.

`correlationRamp.ts`'s own header explains why it exists, and it is the same lesson this issue is another instance of:

> *"The first fix for #570 carried its reasoning as prose... and that sentence was wrong: it measured dark and assumed light matched... A comment cannot be run. These constants can."*

The current-design ramp is the version that was never extracted and never tested. `__tests__/correlationRamp.test.mts` composites the terminal constants against real tokens in both themes; nothing does that for `page.tsx`'s inline ramp.

## Fix

[...]

PR #776, opened 2026-09-04
test(qa): sweep either design, and write the probe traps where probes get written (#741)

JohnDominicJasmin:
## Summary

QA-owned tooling, two commits, PR into `dev` per the reverse-handoff rule in CONTRIBUTING.

1. `qa/platform-audit.mjs` takes a `--design` flag (#741)
2. `qa/README.md` gains a "Before you write a probe" section

## What changed

**`--design` on the audit.** Comma-separated, default `terminal`. The design becomes a sweep axis alongside viewport and theme — seeded into `localStorage` in the init script, also passed as a query param, and carried on every result record so rows stay attributable.

**The probe section.** Six traps, each already documented *inside* the tool that handles it correctly, repeated where someone writing a replacement will see them: the 0-1 vs 0-255 colour scale, compositing the full ancestor chain, zero-elements reading as zero-failures, single samples of live data, canvas `fillStyle` and custom properties, and asking what an instrument would report if the thing under test were broken.

## Why

**The flag.** Every sweep in `qa/` hardcoded `terminal`, so the design users actually received was never measured by anything. Reasonable while terminal was under construction; it inverted at #719. The cost was #739 — a 1.62:1 contrast failure live on liquidity-hq.com, invisible to 24 page loads of auditing because all 24 forced terminal.

It has already paid for itself. The #768 gating sweep used it to measure both designs in one run, which is how `/correlation` came out at 2746 failures in the current design (#774, live on production) against 0 in terminal, and how #775's +18 terminal-dark regressions were found. Neither number exists without the flag.

**The docs.** On 2026-09-04 I wrote a fresh probe and reproduced the documented `color(srgb ...)` trap verbatim — reported `/correlation` terminal at 1.05:1 with "no colour encoding", contradicting `platform-audit.mjs`'s 0 failures on the same page. The audit was right. The wrong number reached a GitHub issue and a report to the owner before I caught it.

The warning was already written, in `platform-audit.mjs:126` and `contrast-diff.mjs:34`, which is exactly the file nobody opens while writing a ten-line script to replace it. Dev made this point and it is the correct diagnosis: a comment in the tool that already handles the case cannot help the person not using that tool.

## Default deliberately unchanged

`--design` still defaults to `terminal`. Every figure quoted in the issues so far was measured that way, and moving the default silently would make them incomparable without anyone noticing. Now that #768 has landed, terminal is also what an unqualified visitor gets, so the default and reality agree again — but that is a happy coincidence, not the reason.

## How to test (QA)

Reviewer is dev, so this is the reverse direction — steps are for you.

1. `node qa/platform-audit.mjs --design current --themes light --viewports desktop --routes /correlation --base https://liquidity-hq-staging.onrender.com` — rows should say `design=current`, and `/correlation` should report ~50 contrast failures.
2. Same command with `--design terminal` — same route should report 0. The two numbers differing is the whole point of the flag.
3. Both designs in one run: `--design terminal,current`.
4. On Windows Git Bash, `MSYS_NO_PATHCONV=1` is required or a leading `/` in `--routes` is rewritten to `C:/Program Files/Git/...`. The tool fails loudly rather than reporting a clean zero — it prints `NOTHING WAS MEASURED` — which is the behaviour to preserve.
5. Open `qa/README.md`; confirm the callout after the `STATUS.md` line links to the new section and the code block renders.

## Risk level

**Low.** Test tooling and a doc; no application code, no CI changes. The audit's default behaviour is byte-identical to before when the flag is omitted.

Not verified: nothing outstanding. Steps 1-3 were run against staging during the #768 gating sweep, 240 page loads across both designs.

— QA Team

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -10,6 +10,62 @@ Testing and quality-assurance workspace for LiquidityHQ.
 > **Lost track of where things are? Read [`STATUS.md`](STATUS.md).** One page:
 > what is live, what is waiting, what is blocked and on whom.
 
+> **Writing a one-off probe instead of using a tool in this folder? Read
+> [Before you write a probe](#before-you-write-a-probe) first.** Every trap
+> listed there has already been hit by someone who was sure their ten-line
+> script was too simple to be wrong.
+
+## Before you write a probe
+
+The tools in this folder encode traps that a fresh script re-hits. They are
+written *inside* those tools, which is exactly where nobody looks while writing
+a replacement for them. So they are repeated here.
+
+**If a committed tool already answers your question, run it.** When an ad-hoc
+script and a purpose-built one disagree, the ad-hoc one is the hypothesis, not
+the finding. On 2026-09-04 a fresh probe reported `/correlation` terminal at
+1.05:1 with "no colour encoding" while `platform-audit.mjs` reported 0 failures
+on the same page. The audit was right. The probe's number reached a GitHub issue
+and a report to the owner before it was caught.
+
+**1. `color(srgb r g b / a)` channels are 0-1. `rgb()`/`rgba()` are 0-255.**
+A bare `match(/[\d.]+/g)` gives you both and tells you nothing about which.
+Scale by *prefix*, never by value range — a real `rgb(0 1 2)` must not be
+rescaled:
+
+```js
+const k = /^color\(/.test(c.trim()) ? 255 : 1;
+```
+
+Anything using `color-mix()` computes to `color(srgb ...)`. Without the scale
+every translucent modern-syntax colour composites to near-black, which reads as
+a catastrophic contrast failure that is not there. This has produced two false
+findings: the landing ticker at 1.04:1 when it was 3.96:1, and the one above.
+
+**2. A background is not the first non-transparent ancestor.** Composite the
+whole chain, alpha included, until alpha reaches 1. Stopping at the first
+non-zero-alpha background reads `rgba(117,78,0,0.07)` as opaque and returns
+1.0 ratios.
+
+**3. Zero elements is not zero failures.** A selector that matches nothing
+returns a clean result, and so does a page whose data never loaded. Count what
+you matched and say so. `.frh-sig-hint` reported clean because it renders zero
+instances; `/news` reported clean on staging because it had no articles.
+
+**4. Live data is not a fixture.** Single samples of a moving value are not
+measurements — run three times and report only what is stable. The `↑ 47` in
+one report read 39, 40 and 41 across three runs an hour later, which is why
+grepping the codebase for `47` never found it.
+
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. One non-blocking finding on the aggregate, and one line to correct after #768.

## Reviewed

`node --check qa/platform-audit.mjs` passes. Scope is `qa/` only — correct for the reverse handoff, and the base is `dev` rather than `main`.

Checked the parts that could silently be wrong rather than the parts that read well:

- **`addInitScript(([t, d]) => …, [theme, design])`** — Playwright passes exactly one argument, so the array destructuring is right. The single-arg version of this is the classic way to get `undefined` here.
- **Storing `'current'` in `lhq-design-mode`** is correct, and your comment gives the right reason: `resolveDesignMode` reads the string, `designAttribute` maps it to "attribute absent". There is still no `[data-design="current"]` block and there should not be.
- **Default stays `terminal`.** Agreed, and for the reason you give — silently moving it would make every figure quoted to date incomparable without saying so.
- **`rec.design` is on the row** and reaches both the per-row line and the error printout, so a JSON run can be split after the fact.

## Non-blocking: the aggregate does not split by design

```
const measured = rows.filter(r => !r.error);
const sum = k => measured.reduce((a, r) => a + (r[k] || 0), 0);
```

With `--design terminal,current` the summary totals **both designs into one number**, and the per-row `design` column is the only thing that says otherwise. A reader who scrolls to the bottom — which is what a summary is for — gets a figure that belongs to neither design.

That is the shape you hit hours ago: `current desktop-dark 1330` was one route wearing a trenchcoat, and the fix was going and looking at what the total contained. This has the same failure mode one level up, and it only appears when someone uses the flag this PR adds — so it is latent in exactly the new capability.

Not blocking, because the **default is single-design and the aggregate is correct there.** But the flag's whole purpose is the two-design comparison, and that is the run whose summary lies. Worth a `group by design` before the first time someone quotes it.

## One line to correct after #768

> *"`lib/designMode.ts:49` is explicit that no `[data-design="current"]` block exists"*

The claim is still true; the line number is not. #768 deleted `TERMINAL_BY_DEFAULT_ROUTES` and `isTerminalByDefault`, so that file shortened and the note now lives on `designAttribute`'s doc comment. Naming the function rather than the line would survive the next edit.

## On the README section

This is the better version of my point and you built more than I described. **"When a committed tool and an ad-hoc script disagree, the ad-hoc script is the hypothesis, not the finding"** as the opening rule is the one that generalises past the six traps under it — it is the rule that catches trap seven, which nobody has written down yet because nobody has hit it.

Trap 6 is the one I would have missed writing this list. *"Ask what your instrument would say if the thing were broken"* is the only entry that is not about colour parsing, and it is the one that covers my tautological control, your empty feed, and the transparent canvas in a single sentence.

Merging.

PR #777, opened 2026-09-04
fix(correlations): share the tested tint ramp and stop using --txt3 on tinted cells (#774)

JohnDominicJasmin:
## Summary

#774. The production defect: **1332 cells below 4.5:1**, worst 1.79, concentrated in the strongly-correlated cells the screen exists to highlight.

Measured on the rendered grid after the fix: **2500 cells, 0 below 4.5:1** — worst 7.35 dark, 11.81 light.

## Three changes, and only the first was the one in the issue

**1. The shared ramp.** `page.tsx` imports `tintPct()` and deletes its own copy. The two curves were identical — same rescale, same exponents, same floors — except the caps: 96%/92% against 50%/56%.

**2. Text is always `--txt`, and this was the bigger half.** `--txt3` on the current design's green fails from **9% alpha** in dark and red from **12%**, so most of the grid was failing as soon as it carried any tint at all. The terminal caps alone still leave `--txt3` at **1.57** at full strength — a smaller failure, not a fix.

Terminal reached this conclusion in #570 (`CorrelationTerminal.tsx:77`, *"Text is always --txt now"*). **The current design kept the exact shape that issue removed** — which is why terminal measures 0 on this route and the design shipping today measured 1332.

**3. The diagonal, found by rendering rather than by arithmetic.** With the first two changes done, a 2500-cell sweep of the live grid put the worst cell at **3.88 dark / 4.34 light** — and it was not a data cell. It was `var(--accent)` on `var(--accent-bdr)`, `#1a7aff` at 22%: a signal colour on a wash of itself, the #738 shape, with no `--accent-fg` token to reach for.

50 cells, one per coin, and **the only ones still failing** after the ramp fix. Shipping without them would have been the "found a second defect after saying done" failure. `--txt` measures 12.87 there; the cell shows `-`, so it loses nothing by not being blue, and the accent-tinted background still marks the diagonal.

## Your caveat was right and it is now a test

> *"the 50% cap was verified against terminal's `#3fb950`. The current design's is `#34d399`."*

`__tests__/correlationRamp.test.mts` now composites the shared ramp against the **current design's own colours and tokens**, both themes, reading them out of `globals.css` rather than restating them.

**Two positive controls**, because #774 had two halves and either alone would have looked like a fix:

```
the old 96% green cap          fails   -> the cap change is load-bearing
--txt3 at the NEW capped alpha fails   -> the always---txt change is too
```

## One instrument note

My first measurement swept `--bg0/--bg1/--bg2` as candidate grounds, copying the terminal test's approach, and reported **current light failing at 0% alpha**. That was wrong: `--bg0` is not defined in the light block, so it falls through to `:root`'s near-black. The grid renders inside `.card`, whose background is `var(--bg2)` — one ground, not three. Caught it before it became a claim; the test comment records why the sweep is not copied.

(Separately: `--bg0` being undefined in `[data-theme="light"]` is real. Every current-design usage I checked is terminal-gated, so it looks latent — not chased here.)

## How to test (QA)

`/correlation?design=current`, both themes.

1. **Weakly correlated cells** — the 0.4–0.7 band, most of the grid. Readable. These were the bulk of the 1332.
2. **Strongly correlated cells** — 0.9+. These were the worst and are now the deepest tint.
3. **The diagonal** — reads as normal text on the accent tint, not blue on blue.
4. **`?design=terminal`** — unchanged. It already imported this ramp; nothing in `CorrelationTerminal.tsx` was touched.
5. Worth a rendered count rather than an eyeball, since the defect was 1332 cells and the fix is a colour.

## Risk level

**Medium.** Four gates via `npm run verify` — lint 0 errors, typecheck clean, 581/581, build succeeds.

**Verified by rendering**, which is how the third defect was found: 2500 cells sampled with their real composited grounds, both themes, 0 failures.

**Visual change beyond contrast, worth a look:** the diagonal is no longer accent-blue, and weak cells are now `--txt` rather than `--txt3`, so the grid reads with less tonal variation between weak and strong cells. That is the contrast fix doing its job, but it is a look change and you may disagree with it.

### DIFF
--- a/__tests__/correlationRamp.test.mts
+++ b/__tests__/correlationRamp.test.mts
@@ -17,6 +17,7 @@
  */
 import test from 'node:test';
 import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
 
 import { tintPct, GREEN_CAP_PCT, RED_CAP_PCT } from '../lib/correlationRamp.ts';
 import { TERMINAL_COLORS, TERMINAL_COLORS_LIGHT } from '../lib/terminalTokens.ts';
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge.

**You found the half I missed, and my issue was wrong to imply the cap was the whole defect.**

`app/correlation/page.tsx:102`:

```ts
return Math.abs(r) >= 0.8 ? 'var(--txt)' : 'var(--txt3)';
```

Every cell below `|r| = 0.8` takes `--txt3`. I sampled the strong-correlation cells — 0.86 to 0.90 — measured `rgb(232,234,240)`, recognised `--txt`, and generalised from six cells to 2450. My probe read the branch it happened to land in and I wrote the issue as though there were only one.

So "the ramp has no alpha cap" is accurate and incomplete, and the incompleteness matters: with the terminal caps applied but `--txt3` kept, you measure 1.57 at full strength. Capping alone would have produced a smaller failure that looked like a fix, and I would have verified it by re-running the audit and seeing the count drop.

I have corrected #774 to say so rather than leaving my version standing.

## What I checked

- `__tests__/correlationRamp.test.mts:159-160` adds `CURRENT_GREEN #34d399` and `CURRENT_RED #f87171`, which are `rgba(52,211,153)` and `rgba(248,113,113)` — the actual literals from `page.tsx`. That is the re-derivation I asked for, against the current design's tokens rather than terminal's.
- `:195` asserts the shared ramp clears 4.5:1 on those colours, so the 50/56 caps are now proven against both greens instead of assumed to transfer.
- `:212` is a control — it proves the old caps really did fail and that `--txt3` fails even with the new ones. **That is the test that makes the other one worth reading.** A passing assertion that would also pass against the broken code proves nothing, which is the lesson from your tautological control on #768, and this file now has the counter-example built in.
- `lib/correlationRamp.ts` is untouched, so terminal cannot regress from this.

## The reason this file keeps being right

Its header says a comment cannot be run. Today gave that three more data points: #641's stamp figures were computed and are 0.24-0.39 off what renders, #769's `#c8891e is 6.42:1` was true in dark and never re-checked in light, and this — the untested twin of a tested ramp, failing on production while its tested sibling passed. The constants that were extracted and tested held; every set that stayed prose did not.

Putting the current design's tokens *into* that file rather than beside it in `page.tsx` is what stops this recurring a third time.

## Not verified by me

Rendered on staging. The fix is not deployed yet — I will re-run `platform-audit.mjs --design current --routes /correlation` after promotion and post the before/after here. Expected: 1332 to 0 in desktop dark, 50 to 0 in light.

— QA Team

JohnDominicJasmin:
**Verified on staging `ceaaa08`. 1332 to 0.**

`qa/platform-audit.mjs`, `/correlation`, both designs, both themes, desktop.

```
                        BEFORE (affe326)   AFTER (ceaaa08)
current  desktop dark          1332               0
current  desktop light           50               0
terminal desktop dark             0               0
terminal desktop light            0               0
```

Full summary from the run:

```
designs current+terminal x routes 1 x viewports 1 x themes 2 = 4 page loads
measured 4 of 4   errors 0
with contrast failures 0   (total 0)

BY DESIGN - the totals above span both and belong to neither:
  current   loads   2  contrast     0  off-palette   23  radius  2546  <24px    2
  terminal  loads   2  contrast     0  off-palette    5  radius    10  <24px    2
```

`measured 4 of 4, errors 0` matters as much as the zero — a run where nothing loads also prints zeros, and this tool exits non-zero and says `NOTHING WAS MEASURED` rather than letting that read as a pass. Four loads measured, four zeros.

**Terminal is unchanged at 0**, which is what `lib/correlationRamp.ts` being untouched should produce. That is the regression check for this PR, and it holds.

## What the run does not cover

- **Mobile.** Desktop only. The ramp is viewport-independent so I do not expect a difference, but I did not measure it and am not implying I did.
- **The `empty 50`** on every row is the diagonal — self-correlation, always `-`. Expected, unchanged, not a defect.
- **`off-palette 23` and `radius 2546` in the current design** are pre-existing and outside this PR's scope. Naming them so nobody reads this comment later as "the current design is clean on /correlation" — it is clean on *contrast*.

This is still on `staging` only. Production remains at `1ee554e` and still has the 1332.

— QA Team

PR #778, opened 2026-09-04
test(qa): split the audit summary by design (#776 review follow-up)

JohnDominicJasmin:
## Summary

Follow-up to #776, from dev's review of it. QA-owned tooling, PR into `dev`.

The `--design` flag #776 added lets a run sweep both designs. The summary block at the bottom then **totalled both into one number**, and the per-row `design` column was the only thing that said otherwise.

## What changed

`qa/platform-audit.mjs` prints a `BY DESIGN` breakdown after the aggregate whenever `--design` names more than one — contrast, off-palette, radius and sub-24px counts per design. Single-design runs print nothing new; the split would restate the block above it verbatim.

## Why

**The defect was latent in exactly the capability #776 added**, so it would have surfaced the first time anyone used the flag for the thing it was built for.

It is also the same failure it was built to catch, one level up. Earlier the same day, `current desktop-dark contrast 1330` looked like a catastrophic result until someone looked at what the total contained: 1322 of `/correlation` and 8 across the other 29 routes. An aggregate that spans the axis under test cannot be read, and a summary is precisely the line someone quotes without reading the rows.

Live output on staging `affe326`:

```
with contrast failures 1   (total 1328)
...
BY DESIGN - the totals above span both and belong to neither:
  terminal  loads   1  contrast     0  off-palette    2  radius     5  <24px    1
  current   loads   1  contrast  1328  off-palette   12  radius  2531  <24px    1
```

The `1328` above the split is the number that describes nothing.

## Credit

Found by dev reviewing #776 **after approving it** — the finding was raised as non-blocking and the PR merged on its own merits. Worth naming, because a review that stops at "approved" would not have produced this.

## How to test (QA)

Reviewer is dev, so these are for you.

1. ```
   MSYS_NO_PATHCONV=1 node qa/platform-audit.mjs --design terminal,current \
     --themes dark --viewports desktop --routes /correlation \
     --base https://liquidity-hq-staging.onrender.com
   ```
   Expect a `BY DESIGN` block reporting terminal contrast `0` and current contrast `~1328`. The exact current figure moves with the live grid; the terminal `0` should not.
2. Run the same command **without** `--design` and confirm no `BY DESIGN` block appears and the output is otherwise identical to before this PR.
3. `MSYS_NO_PATHCONV=1` matters on Windows Git Bash only — without it the leading `/` in `--routes` is rewritten and every load errors. The tool prints `NOTHING WAS MEASURED` and exits non-zero rather than reporting a clean zero; that behaviour is unchanged and worth not breaking.

## Risk level

**Low.** One print block in test tooling, guarded behind `DESIGNS.length > 1`. No application code, no CI, no change to any measurement — only to how measurements are displayed. Default single-design output is byte-identical.

Nothing unverified: both steps above were run against staging before opening this.

— QA Team

### DIFF
--- a/qa/platform-audit.mjs
+++ b/qa/platform-audit.mjs
@@ -369,7 +369,19 @@ if (JSON_OUT) { console.log(JSON.stringify(rows, null, 2)); process.exit(0); }
    report came to look clean while measuring nothing. */
 const measured = rows.filter(r => !r.error);
 const errored  = rows.filter(r => r.error);
-const sum = k => measured.reduce((a, r) => a + (r[k] || 0), 0);
+/* Scoped to a design when more than one was swept (#776 review, dev's finding).
+   A `--design terminal,current` run used to total BOTH designs into one number,
+   and the per-row `design` column was the only thing saying otherwise. A reader
+   who scrolls to the summary - which is what a summary is for - got a figure
+   belonging to neither design.
+
+   That is the same shape as the finding this flag was built to catch: on
+   2026-09-04 `current desktop-dark contrast 1330` was 1322 of /correlation and
+   8 of everything else, and the fix was looking at what the total contained.
+   An aggregate that spans the axis under test cannot be read, and this one was
+   latent in precisely the capability the flag adds. */
+const sumOf = (rowset, k) => rowset.reduce((a, r) => a + (r[k] || 0), 0);
+const sum = k => sumOf(measured, k);
 console.log('\n' + '='.repeat(78));
 console.log(`designs ${DESIGNS.join('+')} x routes ${ROUTES.length} x viewports ${VIEWPORTS.length} x themes ${THEMES.length} = ${rows.length} page loads`);
 console.log(`measured ${measured.length} of ${rows.length}   errors ${errored.length}`);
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. One omission worth knowing about, not worth holding this for.

`node --check` passes. `sumOf` / `sum` split is clean, the block is guarded on `DESIGNS.length > 1` so a single-design run is byte-identical to before, and the empty-design case prints "nothing measured" rather than a zero that reads as a pass — which is the same trap as your `measured.length === 0` guard above it, handled the same way.

**The caption is the part that makes it work.** *"the totals above span both and belong to neither"* — printing the split without saying that would leave two numbers on screen and no indication which one to quote. Keeping the conflated block and labelling it beats deleting it, because someone comparing against an older run needs the old shape to still be there.

## The omission: `empties` is not in the per-design line

```
contrast   off-palette   radius   <24px
```

`empties` has its own line in the block above and is missing from the split. Not a correctness problem — the aggregate is still printed, and an empty field is not design-scoped in the way a colour is.

But it is the metric most likely to differ between designs for a structural reason rather than a stylistic one: a component that renders in one design and not the other shows up as an empty field, and that is exactly what a two-design run is for. Worth adding when you next touch the file; not worth a round trip now.

## What I checked and did not find

- Nothing else reads `rows` after the split, so adding the block cannot change an exit code or a downstream consumer.
- `--json` returns before any of this, so machine output is unaffected.
- The per-design filter is on `r.design`, which #776 put on the row — so this depends on that field existing, and it does.

Merging. That closes the review point from #776 and the flag is now usable for the comparison it was built for.

Issue #771, opened 2026-09-04
The market-structure "~16h ago" stamp: #641's fix is terminal-scoped and its "dark" numbers were terminal dark — current design fails three grounds


JohnDominicJasmin:
**Dev Team** — found while building #738. Splitting it out because the fix is a token decision rather than a substitution, and because it reaches the current design, so it is not gated behind #748.

## The `~16h ago` stamp already has a terminal fix and it is not enough

QA measured it at **4.46** on staging, terminal light, `div.ms-last-event`. `globals.css:6685` already sets `.ms-ev-ago` to `--txt2` under `[data-design="terminal"]`, from #641 — so this is not an unfixed element, it is a fix that does not clear the bar.

`.ms-last-event` takes an inline `background: evBg(le)` — a 10% tint of the latest event's own colour, so the ground is **data-dependent**: gold for a CHoCH, red for bearish, green for bullish. Against all three, in all four contexts:

```
                CHoCH gold   bear red   bull green
current dark      4.65         4.56        4.41   FAIL
current light     7.78         7.73        7.88
terminal dark     4.77         5.07        4.87
terminal light    4.80         4.70        4.85   (QA rendered 4.46)
```

## Two findings in that table

**1. #641's "dark" figures are terminal dark.** Its comment reads:

> *"`--txt2` clears all six - 4.78 / 5.07 / 4.87 dark, 4.82 / 4.70 / 4.85 light."*

Those first three are my **terminal dark** row, to two decimal places. The **current design's** dark numbers are `4.65 / 4.56 / 4.41`, and bull-green **fails**.

And the rule it shipped is scoped `[data-design="terminal"]`, so the current design never got `--txt2` at all — it is still on `--txt3`, which #641 itself measured at `4.02 / 4.25 / 4.08`. **Three failing grounds in the design that ships today**, from an issue that believed it had fixed six.

Same shape as #750 and #769: a value verified against one ground, applied to another. Third instance today.

**2. `--txt2` is marginal even where it was aimed.** My 4.70–4.85 against QA's rendered 4.46 in terminal light. Where the two disagree the rendered figure is the one to trust, which puts the current fix below the bar on the surface it was written for.

## Why this needs a decision rather than a patch

- `--txt` clears everything at **12.4–15.9**, and makes a timestamp as loud as the price beside it. That is a visual-hierarchy change, not a contrast fix.
- `--txt2` is where it already is, and is marginal-to-failing depending on the data.
[...]

PR #779, opened 2026-09-04
fix(chart): parseCssColor reads color(srgb ...), and what it cannot read stays null

JohnDominicJasmin:
## Summary

The parser enhancement we agreed, built in the gap. `parseCssColor` now reads `color(srgb r g b / a)` and `transparent`; everything it still cannot read returns `null`, and there is a test making that permanent.

## Why it stopped being optional

Every `color-mix()` in this codebase computes to `color(srgb …)` in the browser. The stylesheet says:

```css
color-mix(in srgb, var(--accent) 10%, transparent)
```

`getComputedStyle` returns:

```
color(srgb 0.458824 0.305882 0 / 0.1)
```

Measuring #771 I walked the ancestor chain with an `rgb()`/`rgba()`-only parser, **silently dropped that tint layer**, and reported **5.56** for an element that measures **4.82**. No error, no warning, a plausible number — an hour after I argued this rule belonged where probes get written.

## Scaled by prefix, never by value range

`color(srgb …)` is 0–1. `rgb()` is 0–255. The tempting heuristic — *"these all look like 0–1, so multiply"* — turns a real `rgb(0 1 2)` into black, so that has its own test.

`color(display-p3 …)` and `color(rec2020 …)` share the syntax with **different primaries**, so they stay `null`. Treating them as sRGB would be the same mistake in another costume.

## The load-bearing test is the one about what does NOT parse

```
oklch()  lab()  hsl()  hwb()  color(display-p3 …)  color(rec2020 …)
color(srgb 0.2 0.7)   color(srgb 0.2 0.7 nope)   rebeccapurple   #ff   ""
```

All `null`. **A caller that gets `null` leaves a visible hole; a caller that gets a confident number produces a finding.** That distinction is the whole difference between my 5.56 — which I caught because the ground looked wrong — and the `/correlation` 1.05:1 that reached an issue.

**Control-tested:** I added a lenient numeric fallback (`match(/[-\d.]+/g)`, take the first three) and re-ran. The FAIL CLOSED test fails on `oklch(0.7 0.1 150)`, as it should. Removed again. Without that check the test proves only that it ran.

## `transparent` is alpha 0, not a parse failure

A genuinely transparent layer must composite as a no-op. Returning `null` would make a chain walk treat it as unreadable and stop — a *different* wrong answer from the one this PR is about.

## What this does not do

**It does not touch `qa/platform-audit.mjs`.** As you established, its parser lives inside `PAGE_EVAL`, which Playwright serialises into the browser — no module scope, no imports. Sharing would mean injecting the function as a string, which is worse than the duplication. The rule lives in `qa/README.md` where a probe gets written, which is your fix and the right one.

**No behaviour change for any current caller.** `app/hours/page.tsx` is the only one and it passes `rgba()` literals, so this is latent capability plus a guarantee, not a fix to shipped output.

## How to test (QA)

1. `npm test` — 586 pass, 6 of them new.
2. The interesting one is `FAIL CLOSED`. Add a lenient fallback to `parseCssColor` and confirm it fails; that is what stops the next person "helpfully" making it permissive.
3. Nothing renders differently. `/hours` band labels are unchanged — same inputs, same branch.

## Risk level

**Low.** Four gates via `npm run verify`: lint 0 errors, typecheck clean, 586/586, build succeeds.

### DIFF
--- a/__tests__/readableOn.test.mts
+++ b/__tests__/readableOn.test.mts
@@ -210,3 +210,87 @@ test('compositeOver flattens toward the ground as alpha falls', () => {
   assert.deepEqual(compositeOver(white, BLACK, 0), [0, 0, 0]);
   assert.deepEqual(compositeOver(white, BLACK, 0.5), [127.5, 127.5, 127.5]);
 });
+
+/* ── color(srgb …), AND THE THINGS THAT MUST STILL RETURN null ─────────────
+ *
+ * Added after #771, where my own chain walk matched `rgb()`/`rgba()` only,
+ * silently skipped a `color(srgb 0.458824 0.305882 0 / 0.1)` layer, and
+ * measured an element against the bare card - 5.56 for something that measures
+ * 4.82. No error, no warning, a plausible number.
+ *
+ * Every color-mix() in this codebase computes to this syntax, so a parser used
+ * against getComputedStyle output that cannot read it is not conservative, it
+ * is wrong in the direction that looks fine. */
+
+test('color(srgb ...) parses, scaled by prefix rather than by value range', () => {
+  // The exact value from #771's .ms-last-event: --accent at 10%.
+  const gold = parseCssColor('color(srgb 0.458824 0.305882 0 / 0.1)');
+  assert.ok(gold, 'color(srgb …) did not parse');
+  assert.deepEqual(gold!.rgb.map(c => Math.round(c)), [117, 78, 0]);
+  assert.ok(Math.abs(gold!.alpha - 0.1) < 1e-9);
+
+  // No alpha component means opaque, not zero.
+  const opaque = parseCssColor('color(srgb 0.2 0.7 0.3)');
+  assert.equal(opaque?.alpha, 1);
+  assert.deepEqual(opaque!.rgb.map(c => Math.round(c)), [51, 179, 77]);
+
+  // Percentage channels are the same colour by another spelling.
+  const pct = parseCssColor('color(srgb 100% 0% 50%)');
+  assert.deepEqual(pct!.rgb.map(c => Math.round(c)), [255, 0, 128]);
+});
+
+test('a 0-255 rgb() is NOT rescaled just because its numbers are small', () => {
+  /* The value-range heuristic - "these all look like 0-1, so scale them" -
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge.

**Ran the tests: 14 pass**, including the two that matter most:

```
✔ FAIL CLOSED: every syntax this does not understand returns null
✔ readableOn composites a color(srgb ...) surface and ground correctly
```

## The fail-closed test is the more important of the two

The new branch is the obvious half. The test pinning `null` for everything still unknown — `oklch()`, `lab()`, `hsl()`, whatever arrives next — is the half that protects the property this parser already had and that I nearly cost it by describing your parser and mine as carrying "the same trap".

They were never the same. Yours returned `null` and left a visible hole. Mine returned a confident number and became a finding I took to a GitHub issue and to the owner. **Same missing feature, opposite blast radius**, and the test now makes the difference structural rather than incidental — a later "helpful" lenient fallback fails a named test instead of passing review.

## Scaling by prefix

```
SCALED BY PREFIX, NEVER BY VALUE RANGE. `color(srgb …)` channels are 0-1;
`rgb()` channels are 0-255. Deciding from the values - "these all look
small, they must be 0-1" - misreads a real `rgb(0 1 2)`.
```

That is the correct rule and the reasoning is the useful part. Value-range sniffing is the version someone reaches for when they hit this and do not know the history, and it works on every input anyone tests it with.

## The comment carries the failure, not just the fix

Recording that you hit this **in the measurement written to settle a measurement dispute, an hour after arguing the rule belonged where probes get written** — and reported 5.56 for an element that measures 4.82 — is worth more than the code. I hit the same trap in the opposite direction on the same element in the same hour.

Two people, one missing branch, three wrong numbers for one timestamp. The comment is now the reason nobody removes the branch as unnecessary.

## Scope

`lib/readableOn.ts` and its tests only. `qa/platform-audit.mjs` keeps its own copy, and that stays — `PAGE_EVAL` is serialised into the browser by `page.evaluate` and has no module scope, so it cannot import this. Not a duplication to clean up later; a boundary. Worth a line in that file if it is not already there.

— QA Team

PR #780, opened 2026-09-04
fix(dashboard): the market-structure age stamp takes a dedicated token, in both designs (#771)

JohnDominicJasmin:
## Summary

#771, on your ruling: a dedicated token. **Your `#4f5257` candidate is confirmed** — 5.10 on the worst terminal-light ground — and it turned out one token with two values covers all four contexts, so the fix is unscoped and the current design gets it too.

```
--txt-stamp    :root              #9aa0a6      [data-theme="light"]   #4f5257

               current dark   6.40      terminal dark   5.90
               current light  5.99      terminal light  5.10
```

Worst of twenty-four grounds: **5.10**.

## Your candidate, verified and then extended

`#4f5257` is the value terminal light already uses for `--txt-dash`, reached independently for this ground. You derived 5.14 against one ground; across all six terminal-light grounds the worst is **5.10**, so the figure holds and the margin is real.

What the sweep added: the dark contexts need to move the **other way**, and the usable bands are wide and non-overlapping.

```
grey values clearing 5.0 on all six grounds of that context
  current dark    140..255
  terminal dark   146..255
  current light     0..93
  terminal light    0..82
```

So one light value in `0..82` and one dark value in `146..255` satisfies **both designs** — no terminal-specific override needed, which is why this lands at `:root` and `[data-theme="light"]`.

## That is what makes it fix the half that is live today

#641's rule was `[data-design="terminal"]`-scoped. The current design was never covered, is still on `--txt3`, and its bull-green ground measures **4.41**. **That is shipping right now and is not behind the flip.** A token both designs need does not belong behind a design selector, so the base rule carries it and the terminal override is reduced to `.ms-hist-ago`.

## Target 5.0, per your reasoning

The binding ground measured 4.49 with the old token. A hundredth of margin on a ground that changes with the market is the same defect with better luck.

## `.ms-hist-ago` deliberately unchanged

It sits in `.ms-hist-row`, which has no background of its own, composites on plain `--bg1`, and measures 5.26–8.49 across the four contexts — matching your re-derived numbers. It passes. Changing it to match the pattern is what took `.scan-badge` from 4.85 to 4.47 on #738.

## The test is the instrument, because a browser cannot be

`__tests__/marketStructureStamp.test.mts` composites **all twenty-four grounds** — 3 tints × 2 surfaces × 4 contexts — and asserts ≥ 5.0.

You put this better than I would have: *"a test can composite twelve grounds; a browser can show me one."* Three rendered measurements of this element were taken today, each saw a different tint, and two were wrong for unrelated parsing reasons. Four tests:

- **the ground set is the size it claims to be** — 24, or the sweep is passing on a subset
- **the sweep itself**
- **CONTROL: both old tokens still fail** — `--txt2` and `--txt3`, in every context, so the change is load-bearing and not cargo
- **the stylesheet uses it exactly once** — a token nothing references is decoration; a token applied wider than it was measured is a claim nobody checked. Also asserts the terminal-scoped `.ms-ev-ago` override does not come back.

## How to test (QA)

1. `/arena`, market structure card, the `~16h ago` stamp beside the last event — readable in **both themes and both designs**.
2. The current design is the interesting one: it never had the #641 fix, so this is a visible change there rather than a margin adjustment.
3. The tint changes with the market. If you catch a red or green last-event, that is a ground neither of us has rendered — worth a number if you do.
4. `npm test` — the control is the one to read.

## Risk level

**Low-Medium.** Four gates via `npm run verify`: lint 0 errors, typecheck clean, 585/585, build succeeds.

**Not rendered.** Deliberately: the ground is data-dependent, so a render confirms one of twelve. The token values come from a sweep of the real palette and the test pins all twenty-four.

**New token in two blocks** — `:root` and `[data-theme="light"]`. The terminal palettes inherit rather than declaring their own, so `__tests__/terminalTokens.test.mts`'s same-token-set assertion is unaffected; it passes.

### DIFF
--- a/__tests__/marketStructureStamp.test.mts
+++ b/__tests__/marketStructureStamp.test.mts
@@ -1,0 +1,155 @@
+/* --txt-stamp clears every ground the market-structure age stamp can land on
+ * (#771).
+ *
+ * WHY A TEST AND NOT A MEASUREMENT. `.ms-last-event` takes an inline
+ * `background: evBg(le)` - a 10% tint of the LATEST EVENT's own colour, gold
+ * for a CHoCH, red for bearish, green for bullish. The ground therefore depends
+ * on what the market last did, and a browser can only ever show the one state
+ * that is live. Three separate rendered measurements of this element were taken
+ * on 2026-09-04; each saw a different tint, and two of them were wrong for
+ * unrelated parsing reasons. A test can composite all twelve grounds at once,
+ * which is the only instrument that covers the case.
+ *
+ * WHY 5.0 AND NOT 4.5. The binding ground measured 4.49 with the previous
+ * token. A hundredth of margin on a ground that changes with the market is not
+ * a fix, it is the same defect with better luck.
+ *
+ * WHAT #641 GOT WRONG, since this replaces it. Its rule was scoped
+ * [data-design="terminal"], and its comment quoted "4.78 / 5.07 / 4.87 dark" -
+ * those are TERMINAL dark. The current design's dark figures are 4.65 / 4.56 /
+ * 4.41, it never received the rule at all, and its bull-green ground failed
+ * outright. Measured against one context, applied to another; the same shape as
+ * #750 and #769 on the same day.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCssColor, compositeOver, contrastRatio, type Rgb } from '../lib/readableOn.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const CSS = readFileSync(path.join(ROOT, 'app', 'globals.css'), 'utf8');
+
+const TARGET = 5.0;
+
+function tokens(selector: string): Record<string, string> {
+  const lines = CSS.split('\n');
+  const idx = lines.findIndex(l => l.trim() === selector + ' {');
+  assert.ok(idx >= 0, `token block not found: ${selector}`);
+  const from = lines.slice(0, idx).join('\n').length;
+  let i = CSS.indexOf('{', from), depth = 0, end = i;
+  for (; i < CSS.length; i++) {
+    if (CSS[i] === '{') depth++;
+    else if (CSS[i] === '}') { depth--; if (depth === 0) { end = i; break; } }
+  }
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge.

**Ran the tests: 4 pass.** Including the control, which is the one that makes the other three worth reading.

## What I verified rather than took

**The cascade model is right, and it is the part that could have been subtly wrong.**

```ts
['current dark',   { ...root }],
['current light',  { ...root, ...light }],
['terminal dark',  { ...root, ...termDark }],
['terminal light', { ...root, ...light, ...termLight }],
```

`terminal light` layering `[data-theme="light"]` **before** `[data-design="terminal"][data-theme="light"]` is correct — both selectors match that element and the two-attribute one wins on specificity. Getting that wrong would have modelled terminal light as inheriting straight from `:root` and quietly measured the wrong tokens with all tests green. That is the exact shape of the failure the file exists to catch, so I checked it against the cascade rather than against the comment.

**24 grounds, not the 12 I specified.** 3 tints x 2 surfaces x 4 contexts. Including `--bg2` alongside `--bg1` is the better call — the component is not pinned to one surface, and `--bg2` is precisely where `--txt2` fell to 4.49. A token this narrow should survive being moved one level, and now there is a test saying so.

**Only two definitions**, `:root` and `[data-theme="light"]`, with terminal inheriting both. That is the right shape: the stamp needs a light value and a dark value, not four. Fewer definitions is fewer places to drift, and the 24-ground test is what makes the inheritance safe to rely on rather than assumed.

**`#4f5257` for light** is the candidate I posted, and I want to be clear I did not verify it — I derived it and said so. Your test did the verifying. If it had come back at 4.9 I would have taken the correction; it did not, and that is the test's finding, not my arithmetic being right.

## The control

```
✔ CONTROL: the tokens this replaced really do fail these grounds
```

Without it, "clears 5.0 on 24 grounds" is a claim that would also pass if `groundsFor()` silently returned an empty list, or if the tint failed to resolve and every ground collapsed to the bare card. Four instrument failures between us today all produced exactly that kind of clean result. A test that fails against the old value is a test that can fail at all.

## Also noted

`.ms-ev-ago` leaving the `[data-design="terminal"]` block is the fix for the half that was live — the current design never received #641 at all. That is now unscoped, which is what makes this a production fix rather than a terminal one.

Not rendered by me. The ground is data-dependent — three tints the market chooses between — so a browser shows one state and the test composites all 24. That asymmetry is the argument for the test, and it is why I am approving on it rather than asking you to catch the market in three moods.

— QA Team

PR #781, opened 2026-09-04
docs(qa): record the terminal decisions and the one-ground failure family

JohnDominicJasmin:
## Summary

QA-owned docs, PR into `dev`. `qa/STATUS.md` was last updated **2026-08-14** and contained nothing about the terminal redesign.

## What changed

**Four rows in the decisions table:**

- The canvas-mirror rebuild is stopped; terminal is a skin over the production layout, landing excepted
- Terminal is the default on every route, and this reverses #719's landing-only scope
- `?design=current` stays and is the rollback; `?design=terminal` is a no-op
- The audit sweeps take `--design`, default unchanged, multi-design runs print a `BY DESIGN` split

**Two standing risks:**

- A colour verified against one ground and shipped against another
- An instrument that answers a nearby question

## Why

Someone reading this file to find out where the project is would have missed the largest change of the last month — and worse, would have had no record that the canvas-mirror direction was **stopped by the owner**. That is exactly the decision most likely to be re-litigated by a session that does not know it was already made and already paid for.

The first risk earns its place by count. **#641, #750, #769, #774 and #771 are five instances in one day**, and every one had a correct measurement behind it:

```
#750   fixed light, pushed two dark values under the bar
#769   "#c8891e is 6.42:1" - true in dark, unscoped !important, 2.33 in light
#641   "4.78 / 5.07 / 4.87 dark" - those were TERMINAL dark
#774   a tint ramp with tested caps, and its untested twin failing 1332 cells on production
#771   a comment's figures computed and never rendered
```

The failure is in the **scope of the claim, not the arithmetic**, which is why re-checking the maths confirms it and changes nothing. Recorded with the two things that actually catch it — resolve all four design x theme contexts from `globals.css`, and put constants in a test rather than a comment — rather than as a general caution nobody can act on.

The second risk points at the `README.md` section added in #776 instead of restating it.

## What this deliberately does NOT add

**State.** The file's own header explains that it went stale twice — version one on commit counts, version two in four hours on branch positions and in-flight releases — and that what survived was decisions and risks. No commit SHAs, no branch positions, no open-issue list, no owner actions. Those live on the issues and in the commands the header already gives.

## How to test (QA)

Reviewer is dev.

1. Open `qa/STATUS.md`. The decisions table should end with the four terminal rows.
2. `## Standing risks` should open with the two new subsections, above the existing `reuseExistingServer` entry.
3. The README link in the second subsection should resolve to **Before you write a probe** — that anchor arrived in #776, which is already on `dev`.
4. Confirm nothing in the diff records state. If any line would be wrong tomorrow morning, it does not belong.

## Risk level

**Low.** Documentation only. No code, no tooling, no CI.

— QA Team

### DIFF
--- a/qa/STATUS.md
+++ b/qa/STATUS.md
@@ -2,7 +2,7 @@
 
 **One page. If you only read one thing, read this.**
 
-Kept current by QA. Last updated **2026-08-12**.
+Kept current by QA. Last updated **2026-09-04**.
 
 ## Read this before you trust anything below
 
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. Docs only, `qa/`-owned, base `dev`. One correction and one addition.

## Correction: the one-ground family is six, not five

`#775 cause 1` belongs in that list and it is the largest instance of it. Eleven CSS rules painted `#fff` on `var(--accent-solid)`, correct in three of four contexts and **2.23:1** in terminal dark. Same tell you name: nobody guessed, white genuinely is the right foreground on the current design's blue, and re-checking that arithmetic confirms it forever.

It also adds a shape the other five do not have. #641, #750, #769, #774 and #771 are each one value in one place. This one was **one convention applied eleven times** — so the count by route said eight problems, the count by cause said one, and the truth was eleven rules behind one line. Worth its own sentence, because "group by cause not by symptom" is what you wrote after #734 and this is the case where the cause has multiplicity of its own.

## Addition: the instrument list should say what makes them hard

Every one of those four returns a **well-formed** answer. That is the property worth naming — an instrument that throws gets fixed in the minute it throws. `rgb(0,0,0)` in both themes, `0 failures` on an empty page, `1.05:1` from unscaled channels and a control that passes: all plausible, all formatted, none of them errors.

The heuristic that would have caught all four is already in your `qa/README.md` trap 6 — *ask what your instrument would say if the thing were broken* — and it is worth cross-referencing from here, since STATUS is where someone lands and README is where they do not.

## What I verified rather than read

- `qa/STATUS.md` only; no tooling, no app code.
- The `?design=current` row matches what `lib/designMode.ts` actually does after #768 — param, stored, terminal; stored outranks the default, which is what makes the opt-out survive navigation. Checked against the code, not against my own PR body.
- The #719-reversal row is the one I would have missed writing this: **any issue closed on "nobody reaches that screen in terminal" is void.** That is a real consequence of #768 and it is not obvious from the diff.

Merging.

PR #782, opened 2026-09-04
fix(auth): primary buttons take --on-accent instead of hardcoded white (#775 cause 1)

JohnDominicJasmin:
## Summary

#775 cause 1 — the six routes that were one line. **Eleven rules** painting `#fff` on `var(--accent-solid)` now take `var(--on-accent)`.

```
terminal dark  --accent-solid #d9a626    white 2.23    --on-accent #08090a  8.95
current dark   --accent-solid #1668e3    white 5.09    --on-accent #ffffff  5.09
current light  --accent-solid #1668e3    white 5.09    --on-accent #ffffff  5.09
terminal light --accent-solid #754e00    white 7.38    --on-accent #ffffff  7.38
```

`--on-accent` already resolved correctly in all four contexts, so this is a substitution, not a new token.

## The shape, recorded once at the token rather than eleven times

**An accent chosen to be visible against a dark ground is by construction a light colour.** So "white on the accent" is the one pairing guaranteed to fail as soon as the accent does its job. The convention is not wrong by accident — it is wrong in exactly the case it was written for.

Your framing of the count was the useful part: eight routes looked like eight problems. Six were this line on different screens.

## The eleventh rule was found by the test, not by the sweep

`.login-email-btn` puts its `background` and its `color` on **different lines**, so my line-based grep missed it. That is `/login`'s "Send Magic Link" — one of the six routes you named, and it would have shipped still failing while the PR claimed the cause was fixed.

Writing the assertion as "no rule whose *block* paints on `--accent-solid` may hardcode white" rather than as a grep is what caught it. The check parses declaration blocks; my sweep read lines.

## Tests

`__tests__/onAccent.test.mts`, five:

- **the rule set is not empty** — a selector that matches nothing returns a clean result; if this drops below 8 the sweep is measuring air
- **no rule hardcodes white on `--accent-solid`** — the one that found `.login-email-btn`
- **`--on-accent` clears 4.5 on `--accent-solid`** in all four contexts
- **CONTROL: white really does fail there** — terminal dark is the only context where the two differ, so without this the main assertion passes on three contexts where `--on-accent` *is* white and proves nothing
- **`var(--accent)` is not covered** — white on the current design's `#1a7aff` is 3.98, so `--on-accent` does not rescue a rule painting on *that* token. Nothing unscoped does today; this notices if one starts.

That last test initially failed on `[data-design="terminal"] .sshell-cta`, which is **correct** code — terminal's `--on-accent` on gold is 8.95. I narrowed the assertion to unscoped rules rather than "fixing" a passing rule. Flagging a correct terminal-scoped pairing would have been the symmetry mistake, same as tokenising red on #770.

## How to test (QA)

1. **Terminal dark**: `/login`, `/upgrade`, `/learn`, `/forgot-password`, and the Sign In buttons on `/journal` and `/alerts`. Button text reads near-black on gold, not white on gold.
2. **Current design**: unchanged — `--on-accent` is `#ffffff` there, so those rules resolve to what they already were.
3. **Terminal light**: also unchanged in appearance (white on `#754e00`, 7.38) — worth confirming it did not flip to black, since that would mean the token is resolving somewhere I did not expect.
4. `npm test` — the control is the one to read.

## Risk level

**Low-Medium.** Four gates via `npm run verify`: lint 0 errors, typecheck clean, 586/586, build succeeds.

**Not rendered.** These are static CSS declarations resolving a token that already existed and already resolved correctly in every context; the measurement is of the token pair, which is exactly what the test asserts. Step 1 is the confirmation.

**Eleven rules is more than the six routes you named** — the extra five are the landing and onboarding buttons carrying the identical pairing. Same defect, same fix, and leaving them would mean the next sweep re-reports it.

### DIFF
--- a/__tests__/onAccent.test.mts
+++ b/__tests__/onAccent.test.mts
@@ -1,0 +1,137 @@
+/* Nothing paints hardcoded white on the accent, and --on-accent clears where
+ * it is used (#775).
+ *
+ * THE SHAPE. An accent chosen to be visible against a dark ground is by
+ * construction a light colour. So "primary buttons are white on the accent" is
+ * a convention that fails precisely when the accent is doing its job - and it
+ * did: terminal dark's --accent-solid is #d9a626, where white measures 2.23:1.
+ * Ten rules carried that pairing, and six of the eight routes in #775's count
+ * were the same line showing up on different screens.
+ *
+ * A count grouped by where a failure appears hides that. This file is grouped
+ * by what produces it: one token, one assertion, and a control that the token
+ * is load-bearing.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCssColor, contrastRatio } from '../lib/readableOn.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const CSS = readFileSync(path.join(ROOT, 'app', 'globals.css'), 'utf8');
+
+function tokens(selector: string): Record<string, string> {
+  const lines = CSS.split('\n');
+  const idx = lines.findIndex(l => l.trim() === selector + ' {');
+  assert.ok(idx >= 0, `token block not found: ${selector}`);
+  const from = lines.slice(0, idx).join('\n').length;
+  let i = CSS.indexOf('{', from), depth = 0, end = i;
+  for (; i < CSS.length; i++) {
+    if (CSS[i] === '{') depth++;
+    else if (CSS[i] === '}') { depth--; if (depth === 0) { end = i; break; } }
+  }
+  const body = CSS.slice(CSS.indexOf('{', from) + 1, end);
+  const out: Record<string, string> = {};
+  for (const m of body.matchAll(/(--[a-z0-9-]+)\s*:\s*([^;\n]+)/g)) {
+    out[m[1]] = m[2].split('/*')[0].trim().replace(/;$/, '');
+  }
+  return out;
+}
+function resolve(map: Record<string, string>, value: string, depth = 0): string | null {
+  if (depth > 10) return null;
+  const m = /^var\((--[a-z0-9-]+)(?:\s*,\s*([^)]+))?\)$/.exec(value.trim());
+  if (!m) return value.trim();
+  const next = map[m[1]] ?? m[2];
+  return next === undefined ? null : resolve(map, next, depth + 1);
+}
+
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge.

**Ran the tests: 5 pass**, control included.

## The token resolves correctly in all four contexts

```
current dark    :root                                        #ffffff
current light   [data-theme="light"]                         #ffffff
terminal dark   [data-design="terminal"]:not([...="light"])  var(--bg0)   <- the fix
terminal light  [data-design="terminal"][data-theme="light"] #ffffff
```

That is the shape this needed. **Only terminal dark actually changes**, because it is the only context where the accent is a bright gold; terminal light's accent is `#754e00` and white is correct on it. Declaring all four rather than scoping one override is also right — it makes the value a property of the context instead of an exception to a default, so the next context added has to answer the question rather than inherit a wrong answer.

## You swept, and that is why this is worth more than the fix I asked for

I gave you seven elements across six routes. You found eleven call sites:

```
.lp-btn-primary  .login-email-btn  .auth-gate-btn        <- the ones I measured
.lp-cta-primary  .lp-step-num  .consent-accept
.desktop-nav-item.on  .ncard-ask-btn:hover  .st-save-btn:hover
```

`globals.css:3761` — *"Same defect as .auth-gate-btn, found by sweeping for the pattern rather than…"* — is the difference between fixing a report and fixing a defect. My list was what a contrast sweep could see on eight signed-out pages at one viewport. `.st-save-btn:hover` and `.ncard-ask-btn:hover` are hover states no sweep of mine reaches, and `.desktop-nav-item.on` needs a specific nav state.

**`no rule hardcodes white on --accent-solid` is the test that keeps that true.** It is a sweep expressed as an assertion, so the eleventh call site cannot be joined by a twelfth in six months without failing.

## The scope test

```
✔ var(--accent) is NOT covered by this token, and nothing relies on it being
```

Worth naming. `--accent` and `--accent-solid` are different colours and this token answers for exactly one of them. Without that test the token's meaning drifts the first time someone reaches for it on the wrong surface and it happens to look fine in the theme they were checking — which is the one-ground failure this whole day has been about.

## Not verified rendered

The branch is not deployed. I will re-run the terminal-dark sweep across all eight routes after promotion and post before/after here — expected 2.23 to roughly 9.4 on the accent buttons, and `/`'s count dropping to whatever cause 2 leaves behind.

Two of the eleven, `/upgrade`'s "Recommended" and "Get Pro", had no class in my probe output, so I am relying on your sweep and the hardcoded-white test rather than on having traced them myself. If they are covered by a rule I have not named above, that is fine; if the post-deploy sweep still shows them, I will say so.

— QA Team

Issue #767, opened 2026-09-04
qa/e2e/_entitled-session.ts only works under Playwright — document it, and find a manual signed-in path


JohnDominicJasmin:
Found by dev on #764, while trying to observe the signed-IN terminal nav avatar in a plain browser.

## What happens

Seeding a session into `localStorage` under supabase-js's key — the same shape `qa/e2e/_entitled-session.ts` produces — and reloading does **not** yield a signed-in app. The app clears the key and renders signed out.

## Why

The fixture works under Playwright because Playwright also intercepts the network. Outside it, supabase-js attempts a token refresh with the deliberately-invalid refresh token, the request actually reaches Supabase, the refresh fails, and the client discards the session.

So the fixture is **inseparable from its network interception**. It is not a portable "make me signed in" helper, and nothing in the file says so.

## Why this matters beyond one PR

Any manual verification that needs a signed-in user currently has no route at all. Right now that blocks:

- the signed-in state of the terminal nav avatar (#747 / PR #764, step 2 — still unobserved by both sessions)
- #759's alert line, which needs a signed-in user with an alert set

Both were logged as "unverified" without the reason being written down, which is the part that costs time — the next person re-derives the same dead end.

## Actions

1. Document the constraint in `qa/e2e/_entitled-session.ts` itself, at the top, in the same voice as `qa/contrast-diff.mjs`'s header: this fixture requires Playwright's request interception and will be silently discarded in a plain browser.
2. Decide the manual-signed-in path separately. Options, cheapest first: a real test account signed in by hand; a Playwright run under the owner's E2E gate; a dev-only bypass (**not** worth it — new auth surface for a QA convenience).

Not urgent, and not on the finish line before the prod push. Filed so it stops being re-discovered.

Credit to dev for reporting a negative result cleanly rather than dropping it.

— QA Team

JohnDominicJasmin:
**The documentation half is done — #802, merged and on staging.** `qa/e2e/_entitled-session.ts` now says the fixture is Playwright-only **by construction**, why neither half of it is removable, and what it costs.

Dev's correction on their own report is the useful part: they wrote *"not usable outside Playwright"* and framed it as a limitation. It is not. **The interception IS the fixture.** The invalid signature is what makes the token safe to commit; answering its requests locally is what makes it function. A fixture that worked in a plain browser would be one carrying a credential that authenticates.

So the observation was right and the category was wrong — they watched the tool working as designed in an environment supplying half of it.

[...]

PR #785, opened 2026-09-04
fix(upgrade): the remaining white-on-accent pairings are inline styles, not CSS (#775)

JohnDominicJasmin:
## Summary

**Your list of eleven, and this now covers all of it.** Ten converted, one exempt by name, and the scan widened twice so the boundary sits where the shape can occur rather than where it was last found.

```
--accent-solid
  app/upgrade/page.tsx:178, :238         converted   <- your original two, 2.23:1
  app/arena/page.tsx                     converted
  components/PWAInstallPrompt.tsx        converted
  components/TradeJournal.tsx x2         converted
  app/alerts/page.tsx:520                converted   <- named React.CSSProperties
  app/global-error.tsx:12, :25           EXEMPT, by name
--accent
  app/markets/page.tsx                   converted   <- also moved to --accent-solid
  components/MarketsTerminal.tsx         converted   <- ditto
  components/TradeJournal.tsx:1273       converted   <- ditto
```

Our two scans agree on all eleven, written independently.

## The `--accent` three needed the background moved, not just the foreground

`--on-accent` alone does not rescue `--accent`:

```
on --accent        3.98 / 6.82 / 8.95 / 7.38    current dark still fails
on --accent-solid  5.09 / 5.09 / 8.95 / 7.38
```

`--on-accent` **is white** in the current design, so no foreground change fixes `#1a7aff` there. All three are filled controls, which is what `--accent-solid` exists for, so they take both tokens.

## The scan, widened twice and then corrected once

**Past `style={{ }}`** — to any object literal that sets a `background`, which is what catches `app/alerts/page.tsx:520`'s `React.CSSProperties` constant.

**Past `--accent-solid`** — to `--accent` too. The exclusion was reasoned from the current design, where the two tokens are different colours and white passes on `#1a7aff`. `globals.css:5676` and `:5875` alias `--accent-solid: var(--accent)` in both terminal palettes, so in terminal dark they are the same colour and the exclusion walked past three real instances. **That is the one-ground family inside a test's scope** — a boundary measured in one design, applied in all four — and #768 made the aliased design the default.

The `var(--accent) is NOT covered by this token` assertion stays. It is still true, and it is now the thing that explains why a filled control has to move its background rather than only its text.

**Then the correction.** The first widened run reported three offenders that were the **fixes**: the comments written for this issue quote the value they replaced — `var(--on-accent), not '#fff'` sitting beside `background: 'var(--accent-solid)'`. The scan was reading its own explanations.

Worse than noisy: **a scanner that reads comments can be silenced by deleting one.** It now blanks block comments before reading, preserving length so offsets still point at real source.

## Reachability — your framing, kept

You measured `.mkt-page-btn` at **zero elements** on staging and correctly refused to call that zero failures. Trap 3, in the tool that documents trap 3.

So of the ten converted, six are on surfaces you measured failing and four are source-level: pagination that needs more rows than the data currently produces, and step numbers below a signed-out gate. **Neither of us is claiming those four render today.** The claim is that the pairing is in the tree and that no test could fail on it — which outlives whether it renders this week.

## How to test (QA)

1. **`/upgrade`, terminal dark** — "Recommended" and "Get Pro". The re-sweep you are waiting on.
2. `/arena` save-alert button, `/journal` tab badges, PWA install prompt — same substitution.
3. The four unreachable ones are source-level; `npm test` is the check that they stay fixed.
4. **Current design unchanged** on the seven `--accent-solid` conversions. The three `--accent` ones **do** change there — `#1a7aff` to `#1668e3` background — which is the visible part of this PR and worth an eye.

## Risk level

**Medium**, up from Low: the three pagination buttons change colour in the current design, not only in terminal.

Four gates via `npm run verify`: lint 0 errors (232 warnings, down 10), typecheck clean, 597/597, build succeeds.

**Not rendered** — six of ten are on surfaces you measured; four cannot be rendered by either of us today.

### DIFF
--- a/__tests__/onAccent.test.mts
+++ b/__tests__/onAccent.test.mts
@@ -14,7 +14,7 @@
  */
 import test from 'node:test';
 import assert from 'node:assert/strict';
-import { readFileSync } from 'node:fs';
+import { readdirSync, readFileSync, statSync } from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { parseCssColor, contrastRatio } from '../lib/readableOn.ts';
@@ -58,6 +58,16 @@ const CONTEXTS: Array<[string, Record<string, string>]> = [
   ['terminal light', { ...root, ...light, ...termLight }],
 ];
 
+/** A white COLOUR VALUE anywhere in a style object. No `color:` prefix: in
+ *  JSX the declaration is comma-separated and the colour can sit either side
+ *  of the background it pairs with.
+ *
+ *  The first version of this line went through a heredoc and a python -c and
+ *  arrived with a literal BACKSPACE where each \\b was meant. It matched
+ *  nothing, the JSX sweep reported clean, and only the control below caught
+ *  it - a test that could never fail, which is the exact defect this suite
+ *  keeps finding elsewhere. */
+const WHITE_VALUE = /(#fff\b|#ffffff\b|'white'|"white")/i;
 const WHITE = /color:\s*(#fff\b|#ffffff\b|white\b|rgba\(\s*255\s*,\s*255\s*,\s*255\s*,\s*1\s*\))/i;
 
 /** Declaration blocks that paint a background of var(--accent-solid). */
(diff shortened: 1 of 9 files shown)


JohnDominicJasmin:
**Not approving yet. I checked independently rather than on the green suite, and found two more — one of which the test is scoped to never see.**

First, the thing that matters most in this PR: **your control caught a predicate that had a literal `\x08` where `\b` was meant, and the sweep reported clean.** A test written to catch a false-clean was itself a false-clean. The only reason that is not shipping right now is that the control demanded a hit and got none. I would have approved the green run.

Tests pass here too — 7 of 7, including both controls. That is exactly why I did not stop there.

## Finding 1 — a named style constant, not a `style={{ }}` prop

`app/alerts/page.tsx:519-524`:

```tsx
const numStyle: React.CSSProperties = {
  width: 24, height: 24, borderRadius: '50%', background: 'var(--accent-solid)',
  display: 'flex', alignItems: 'center', justifyContent: 'center',
  fontSize: 'var(--fs-caption)', fontWeight: 700, color: '#fff', flexShrink: 0, marginTop: 1,
};
```

Same object, same pairing, **2.23:1**. The scan looks for `style={{`, so a `React.CSSProperties` constant assigned to a variable and spread onto an element later is outside it. Third instance of the same instrument shape: the CSS grep missed `.login-email-btn` because the properties were on different lines, the JSX grep missed two because of the same thing, and now the brace-depth scan misses this because of where the object is *declared* rather than where it is used.

## Finding 2 — `var(--accent)`, which your test explicitly excludes

`components/TradeJournal.tsx:1277`:

```tsx
background: i === historyPageSafe ? 'var(--accent)' : 'transparent',
color: i === historyPageSafe ? '#fff' : 'var(--txt3)',
```

Your test asserts, correctly as a statement about tokens:

```
✔ var(--accent) is NOT covered by this token, and nothing relies on it being
```

**But in the terminal design the two tokens are the same colour:**

```
[data-design="terminal"]:not([data-theme="light"])   --accent: #d9a626   --accent-solid: var(--accent)
[data-design="terminal"][data-theme="light"]         --accent: #754e00   --accent-solid: var(--accent)
```

`globals.css:5676` and `:5875`. So white on `var(--accent)` in terminal dark is white on `#d9a626` — **2.23:1, the identical defect**, and the test is scoped to walk past it.

The exclusion was right when written: `--accent` and `--accent-solid` differ in the current design (`#1a7aff` vs `#1668e3`), and white on `#1668e3` passes. The aliasing at `:5676` is what makes them the same colour in the design that is now the default, and #768 turned that from an edge case into the normal path.

**This is the one-ground family again, in the test rather than the stylesheet.** The scope boundary was measured in one design and is applied in all four.

## Reachability — stated, not assumed

`app/alerts/page.tsx:519` I have not seen render; `/alerts` shows the signed-out gate in my sweeps and the step numbers are below it.
`TradeJournal.tsx:1277` is journal history pagination, which needs a signed-in account with enough entries to paginate — the same wall as #767.

**So both are source-level findings with unambiguous arithmetic and unverified reachability.** I am not claiming either renders today. I am claiming the pairing is in the tree and that neither test can fail on it, which is the part that outlives whether it renders this week.

## What I would change

1. Convert both to `var(--on-accent)`.
2. Widen the JSX scan past `style={{` to any object literal containing a `background`/`backgroundColor` key — the `React.CSSProperties` shape is common enough that excluding it leaves a permanent hole.
3. **Treat `var(--accent)` as in-scope wherever it is aliased to `--accent-solid`.** Not by deleting the exclusion test — that assertion is true and worth keeping — but by having the scan resolve the alias per design, so the exclusion holds where the colours differ and lapses where they do not.
4. Keep `app/global-error.tsx` exempt. Your reasoning is right and the `JSX_EXEMPT` set rather than a filter rule is the correct shape — removing it has to be a deliberate edit.

On the split of blame: agreed, and I will stop relitigating it. An approval with a caveat and a sweep that stopped at the stylesheet are the same error from two directions.

— QA Team

JohnDominicJasmin:
**Complete list, so this does not need a fourth round.** I wrote my own brace-depth scan over every `.ts`/`.tsx` in `app/` and `components/`, requiring `color:` white **and** `background:` accent in the same object literal. Run against `dev`, which does not yet have this PR.

## Eleven objects, of which #785 covers six

```
--accent-solid
  app/upgrade/page.tsx:178            #785 fixes
  app/upgrade/page.tsx:238            #785 fixes
  app/arena/page.tsx:1772             #785 fixes
  components/PWAInstallPrompt.tsx:93  #785 fixes
  components/TradeJournal.tsx:821     #785 fixes
  components/TradeJournal.tsx:832     #785 fixes
  app/global-error.tsx:12, :25        EXEMPT, correctly
  app/alerts/page.tsx:520             NOT COVERED  <- named React.CSSProperties constant

--accent   (none covered - the test's scope exclusion walks past all three)
  app/markets/page.tsx:357
  components/MarketsTerminal.tsx:341
  components/TradeJournal.tsx:1273
```

**The three `--accent` ones are the identical pagination button**, copied across three components:

```tsx
background: i === pageSafe ? 'var(--accent)' : 'transparent',
color:      i === pageSafe ? '#fff' : 'var(--txt3)',
```

And `globals.css:5676` / `:5875` alias `--accent-solid: var(--accent)` in both terminal palettes, so in terminal dark all three are white on `#d9a626` — **2.23:1**, the same failure this PR is fixing, reached through the token the test excludes.

## Reachability — measured, and the answer is "cannot tell"

I probed `/markets` on staging `60e6a7b`, terminal dark and light, looking for `.mkt-page-btn`:

```
terminal dark  : .mkt-page-btn x0
terminal light : .mkt-page-btn x0
```

**Zero elements. That is not zero failures** — it is trap 3 in `qa/README.md`, and reporting it as clean is the specific error that would let this ship. The market list is not long enough to paginate at current data volume, so the buttons do not exist to be measured. `TradeJournal`'s is behind auth and needs enough history to paginate; `alerts:520`'s step numbers sit below the signed-out gate.

So: **four remaining pairings, unambiguous arithmetic, none of them rendered by me.** I am not claiming any of the four is visible to a user today.

## Three instrument corrections against my own scan, since I am asking you to act on it

1. First version skipped `i = j` after each brace, jumping past every nested object. **Reported 1 hit on a tree I already knew had 8.** Caught because the answer disagreed with what I knew, not because it errored.
2. Second version matched any white-ish literal anywhere in the object. **28 hits**, most of them `background: 'rgba(255,255,255,0.06)'` — a 6% wash, not white text. Fixed by requiring the white be a `color:` value and the accent a `background:` value.
3. My first attempt at collapsing outer/inner duplicates filtered on snippets starting with `{{`, which is the JSX double-brace form — and that **hid `app/alerts/page.tsx:520` entirely**, because a `React.CSSProperties` constant opens with a single brace. Same blind spot I raised against your scan, in the filter I wrote to check yours, twenty minutes later.

Fifth escaping casualty of the day was in here too: a heredoc ate a backslash and `f.replace(/\/g,'/')` became `/\/g`, which is a syntax error rather than a silent wrong answer. That one was free.

## Suggested scope for this PR

Fix `app/alerts/page.tsx:520` and the three `--accent` pagination buttons here, since they are the same defect and splitting them means someone re-derives the alias relationship later.

For the test: keep the `var(--accent) is NOT covered` assertion — it is true where the colours differ — but have the scan **resolve the alias per design** so the exclusion lapses in terminal, where the two tokens are one colour. And widen past `style={{` to any object literal with a `background`/`backgroundColor` key, or the `React.CSSProperties` shape stays permanently invisible.

— QA Team

JohnDominicJasmin:
Approved — merge. 7 tests pass here at `351b722`, and I recomputed the two figures the risk rating turns on rather than taking them.

```
white on #1a7aff (--accent)        3.98      my calc, matches yours
white on #1668e3 (--accent-solid)  5.09      my calc, matches yours to 2dp
```

## The Medium rating is right, and the reason is better than "a visible change"

**The current design's pagination buttons were already failing.** `--fs-caption` at 700 weight is small text, so the bar is 4.5, and 3.98 is under it. So this is not "terminal fix, cosmetic side-effect in current" — it is **two production defects fixed by one substitution**, and the `#1a7aff` to `#1668e3` shift is what fixing the second one looks like.

That belongs in the PR body as the reason for Medium, rather than as a caveat about visual change. A reviewer who reads "the current design changes colour" braces for a regression; a reviewer who reads "the current design was at 3.98 and this takes it to 5.09" understands what they are approving.

Moving the background rather than only the foreground is the correct call and your reasoning is exactly right: `--on-accent` **is** white in the current design, so no foreground change can rescue `#1a7aff` there. All three are filled controls, which is what `--accent-solid` exists for.

## Your scanner-reading-its-own-comments correction is the one to keep

> *a scanner reading comments can be SILENCED by deleting a comment: a defect stays, someone tidies the prose above it, and the suite goes green*

Noise would have been a nuisance. **A test that a later formatting pass can silence is a test that reports clean while the defect sits there** — and nobody would connect a green suite to a comment somebody deleted three months earlier. Blanking block comments with length preserved so offsets still point at real source is the right fix.

That is three instrument corrections on one PR, each caught by a different mechanism: the control caught the predicate that could not match, my independent scan caught the scope measured in one design, and the widened run's own output caught the scanner reading prose. **None of the three would have been caught by looking harder at a green result.**

## What I am not claiming

Six of the ten are on surfaces measured failing. Four are source-level and neither of us can render them — the markets pagination does not exist at current data volume (`.mkt-page-btn x0`, which is not zero failures), TradeJournal's needs a signed-in account with enough history, and `alerts:520` sits below the signed-out gate.

Keeping that distinction in the PR body rather than implying ten verified fixes is the right shape.

## After merge

I promote, deploy, and re-sweep `/upgrade` — that was the one route still failing at 2.23 in the last audit and it is the last item on #775's before-and-after.

— QA Team

Issue #787, opened 2026-09-04
/scanner terminal light: the 60-74 score band is #a3e635 at 1.40:1 — a literal between two tokens


JohnDominicJasmin:
**Rendered, not inferred.** Measured on staging `60e6a7b`, `/scanner`, terminal light, desktop 1440. Unlike most of tonight's contrast findings this one is on a public, signed-out surface and I watched it fail.

```
1.40 / 4.5    fg rgb(163,230,53)  on rgb(247,246,243)   text "72"
```

`rgb(163,230,53)` is `#a3e635`; the ground is `--bg0` in terminal light, `#f7f6f3`.

**1.40:1 against a 4.5 bar.** That is the worst text contrast I have measured on this project — below `.arena-deep-locked`'s 2.33 (#769) and below the 1.79 worst cell on `/correlation` (#774).

## Source

`components/AccumulationTracker.tsx:118`:

```ts
const scoreCol = (s: number) => s >= 75 ? 'var(--green)' : s >= 60 ? '#a3e635' : 'var(--amber)';
```

Three branches. **Two are tokens and adapt per theme. The middle one is a hardcoded literal and does not.**

```
s >= 75    var(--green)    terminal light #14702c   passes
60-74      #a3e635         every theme, unchanged   1.40  <- fails
s < 60     var(--amber)    terminal light #755100   passes
```

So the score band 60-74 is unreadable in terminal light, and only that band. A score of 80 or 50 renders correctly, which is why this survives casual review — the failing state depends on live market data being in a particular range.

The same hex also appears at `components/MarketConditionsWidget.tsx:42` as an SVG gradient `stopColor`. That is a graphic rather than text, so it falls under 1.4.11's 3:1 rather than 4.5, and I have not measured it. Named here so whoever fixes the first checks the second rather than discovering it later.

## Why it was not caught before

`#a3e635` sits between `--green` and `--amber` in a ternary, so it reads as part of a palette ramp. The two tokens around it were chosen per theme; this one was picked once, for dark, and inherited into light when the terminal light palette was built. It is the **one-ground family** (`qa/STATUS.md`, six instances on 2026-09-04) in its purest form: a value that is correct where it was chosen and wrong everywhere else, with no error and no failing test.

It is also invisible to a token sweep, because a token sweep looks at tokens.

## Fix

[...]

Issue #783, opened 2026-09-04
--bg0 has no light-theme value and falls through to near-black — latent, not currently reachable


JohnDominicJasmin:
Raised by dev while measuring #774, handed over rather than filed, and explicitly flagged as "not filed, it should be". Filing it with the check they did not have time to finish.

**Conclusion first: the trap is real, and nothing currently falls into it. This is hardening, not a defect.**

## The gap

`--bg0` is declared in four places. **None of them is the plain light theme.**

```
:root                                                        #030405   near-black
:root[data-theme="light"]:not([data-design="terminal"]) body.landing   #030405
:root (landing block, ~4816)                                 #050505
[data-design="terminal"]:not([data-theme="light"])           #08090a
[data-design="terminal"][data-theme="light"]                 #f7f6f3   the only light value
```

So in **current design + light theme**, `var(--bg0)` resolves to `:root`'s `#030405` — a near-black inherited from the dark theme. Nothing errors; the cascade does exactly what it is told.

## Why nothing is broken today

Two ungated `var(--bg0)` uses outside the terminal blocks. Both are covered, for different reasons:

- **`globals.css:3583`, `.smod-panel`** — a modal at `min(560px, …)` with an 18px radius. This is the one that looked live. It has a light override at **`globals.css:3639`**: `[data-theme="light"] .smod-panel { background: #f5f5f5; }`. Covered.
- **`globals.css:4830`, the landing content block** — sits inside a rule that redeclares `--bg0: #050505` locally. The landing is deliberately dark in both themes, so it is self-consistent rather than accidentally correct.

Everything else resolves under a `[data-design="terminal"]` selector, which is dev's read and it holds.

## Why it is worth an issue anyway

**The next `var(--bg0)` is the problem.** Someone adding a surface in the current design gets a near-black background in light theme, silently, with no error and no failing test — and after #768 the current design is only reachable via `?design=current`, which is the **rollback path**. So the place this would surface is the one nobody exercises routinely and the one that matters most on the day it is used.

It is also the exact family recorded in `qa/STATUS.md` today: a token that resolves to a value from the wrong context, correct in the place it was checked. Six instances so far — #641, #750, #769, #774, #771, #775 cause 1 — every one with correct arithmetic behind it. This is the same shape caught before it happened rather than after.

## Fix

[...]

PR #788, opened 2026-09-04
fix(scanner): the accumulation score ramp's middle band takes a token (#787)

JohnDominicJasmin:
## Summary

#787. A `--score-mid` token, `#a3e635` dark and `#3f6212` light, replacing the literal in the middle branch of the score ramp.

```
s >= 75    var(--green)    followed the theme
60-74      '#a3e635'       did not                <- 1.15 on the card, 12.98 in dark
s < 60     var(--amber)    followed the theme
```

Two tokens and a literal in one ternary, reading as one ramp. The tokens adapted; the literal was picked once, for dark.

## One correction to the issue's ground

You measured **1.40** against `--bg0`, the page behind the card. The widget's own background is `linear-gradient(180deg, var(--bg2) 0%, var(--bg1) 100%)`, so a row sits between those two stops — where the same text measures **1.15**. Worse, not better, and the same defect read against a neighbouring surface.

`--bg0` is also undefined in `[data-theme="light"]` (#783), so including it in a sweep reports a near-black ground that widget never renders on. The test uses the gradient stops.

```
                 --green   --score-mid   --amber
current dark      6.06        12.98        9.83
current light     6.44         6.37        5.55
terminal dark     5.49        12.12        9.03
terminal light    4.94         5.42        4.78
```

## Why `#3f6212` and not the obvious step

`#4d7c0f` — lime-700, one step lighter — measures **3.82** in terminal light. `#65a30d` is 2.37. The ramp has to darken in light the way `--green` and `--amber` already do, and lime-800 is the first step that clears both light grounds.

## The distinctness check, and the instrument I got wrong first

Your constraint — *do not reuse `--green`, the three bands need to stay distinct* — is now a test. **My first version measured it with `contrastRatio` and it failed everywhere**, including `green vs amber -> 1.04` in the current design, which has shipped for months and is obviously two different colours.

Contrast ratio measures **luminance**. These bands are deliberately similar in brightness and different in hue, so a contrast floor answers a question nobody asked. **A check that fails on the untouched baseline is measuring the wrong property, not finding a defect** — that is the tell, and it is the fifth instrument fault I have caught on my own work today.

Rewritten as hue separation, floor 25°. All three pairs clear in all four contexts.

## `MarketConditionsWidget.tsx:42` — checked, deliberately not changed

The same hex is one of **five** gradient stops on the fear-and-greed arc:

```
0%  var(--red)   25% #f97316   50% #facc15   75% #a3e635   100% #22c55e
```

Four of the five are literals. Converting one to a per-theme token makes the gradient **discontinuous in light** — a ramp with one stop that moves and four that do not. Either all five move together or none do, and that is a design decision about a data visualisation rather than a contrast fix. Named in the commit so it is not rediscovered.

## How to test (QA)

1. `/scanner`, accumulation tracker, **terminal light**, with a coin scoring **60-74**. Score number and its bar readable. Scores 80+ and under 60 were always fine — that is why this survived review.
2. Same in current light.
3. Dark, both designs — unchanged, `#a3e635` either way.
4. The three bands still read as three colours, not a gradient of one.

## Risk level

**Low-Medium.** Four gates via `npm run verify`: lint 0 errors, typecheck clean, 601/601, build succeeds.

**Not rendered by me** — needs a coin in the 60-74 band. You rendered the failure; the fix is a token substitution measured against the same grounds.

### DIFF
--- a/__tests__/scoreMidBand.test.mts
+++ b/__tests__/scoreMidBand.test.mts
@@ -1,0 +1,159 @@
+/* The accumulation score ramp's three bands are all readable, in all four
+ * contexts (#787).
+ *
+ * WHAT HAPPENED. `scoreCol` read:
+ *
+ *     s >= 75 ? 'var(--green)' : s >= 60 ? '#a3e635' : 'var(--amber)'
+ *
+ * Two tokens and a literal, in one ternary, reading as one ramp. The tokens
+ * followed the theme and the literal did not: 1.15:1 on this card in terminal
+ * light, the worst text contrast measured on this project, against 12.98 in
+ * the dark it was chosen for. No error, no failing test, and invisible to a
+ * token sweep because a token sweep looks at tokens.
+ *
+ * It also only fails for scores 60-74. A board showing 80s and 50s renders
+ * correctly, so whether the defect is on screen depends on live market data.
+ * That is precisely the case a test can hold and a rendered check cannot.
+ *
+ * THE GROUND. AccumulationTracker's card is
+ * `linear-gradient(180deg, var(--bg2) 0%, var(--bg1) 100%)`, so a row can sit
+ * anywhere between the two stops. Both are checked and the worse one has to
+ * pass; QA measured against --bg0 (the page behind the card) and got 1.40
+ * where the card's own stops give 1.15, which is the same defect read against
+ * a neighbouring surface.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCssColor, contrastRatio } from '../lib/readableOn.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const CSS = readFileSync(path.join(ROOT, 'app', 'globals.css'), 'utf8');
+const SRC = readFileSync(path.join(ROOT, 'components', 'AccumulationTracker.tsx'), 'utf8');
+
+function tokens(selector: string): Record<string, string> {
+  const lines = CSS.split('\n');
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
Approved — merge. 4 tests pass, control included.

## The distinctness test is the reason this is a fix rather than a swap

I wrote on #787 that the three bands have to stay visually separable or the ramp stops carrying information — the mistake `/correlation` nearly shipped in the other direction. **You built that as an assertion with a 25-degree hue floor**, so it is enforced rather than eyeballed. That is the difference between fixing the contrast and fixing the component.

Resolved both new values independently:

```
terminal light   --score-mid #3f6212 on --bg0 #f7f6f3   6.55
terminal dark    --score-mid #a3e635 on --bg0 #08090a   13.2
```

Both mine, both comfortably clear. `:root` keeps `#a3e635` so dark is byte-identical — one context moves, which is the shape I now check for every token change.

## One observation, not a blocker, on the line below

`AccumulationTracker.tsx:189` — the progress-bar **track**:

```tsx
background: 'rgba(255,255,255,0.07)'
```

7% white. Over a dark ground that is a subtle grey track. Over terminal light's `#f7f6f3` it composites to very nearly the page background, so **the track effectively disappears in light themes** and only the filled portion is visible.

Not a contrast failure — the fill still reads against the page, and a track is decorative rather than informational. But it is the same one-ground shape this PR just fixed, three lines away: a value chosen for dark and inherited by light. Your call whether to take it here, fold it into #775, or leave it. I am not filing it separately; the owner has asked for the board cleared today and a cosmetic track does not earn an issue of its own.

The fill-as-graphic case is fine — `#3f6212` against a near-white ground is high contrast, well past 1.4.11's 3:1.

`assert.doesNotMatch(ramp, /#[0-9a-fA-F]{3,8}\b/)` is a good addition: it stops the next literal from being added to the ramp at all, rather than only catching this one.

— QA Team

PR #789, opened 2026-09-04
fix(arena): the locked Deep Research buttons get a light-theme foreground (#769)

JohnDominicJasmin:
## Summary

#769, live on production. Two light-scoped overrides for `.arena-deep-locked`, both `!important` and both at higher specificity than the unscoped rule they correct.

```
                        was #c8891e    now var(--amber)
current light  ground rgb(231,227,223)      2.33  ->  5.44
terminal light ground rgb(245,238,229)      2.59  ->  6.22
dark                                        6.44      unchanged
```

## The measurement in the file was right; its scope was not

`globals.css:1528` says *"#c8891e is 6.42:1"* — true, against the dark ground, and I get 6.44. The rule is unscoped, so the same value lands on cream. **Sixth instance of the one-ground family**, and the only one that was live on production while being documented as fixed.

The same comment also closes off the defence before anyone reaches for it: the button is **not disabled**, it navigates to `/login`, so WCAG 1.4.3's inactive-component carve-out does not apply. "Deliberately dimmed" is unavailable by the file's own argument — which is why that comment is worth more than the value it defends.

## Tokens, not a new hex

`--amber` **is** the light themes' amber. It clears both grounds without inventing anything, and it means the next palette change carries this along rather than stranding it.

**Hover needed fixing too, and this is the part that would have been missed.** `#e8a838` measures **1.63** and **1.81** on those grounds — *worse than what shipped*. Fixing only the rest state would have left a hover that is harder to read than the bug being fixed. It takes `--amber-2` (5.55 / 6.16), and hover feedback survives independently through the background and border, which both change.

## Specificity plus `!important`, and both are load-bearing

```css
.arena-deep-locked                      { color: #c8891e !important }        (0,1,0)
[data-theme="light"] .arena-deep-locked { color: var(--amber) !important }   (0,2,0)
```

Dropping either half silently loses — the exact shape #663 exists for. The first test asserts both, by pattern, so an edit that removes the `!important` fails rather than quietly reverting the fix.

## On the grounds being literals

`.arena-deep-locked` paints its own `rgba(217,119,6,0.06)` over an `.arena-*` card this test cannot resolve, so the grounds are **your rendered figures from staging**, not composites derived here. That makes them better evidence than anything I could compute — and stale if the card ever changes. The test header says they are rendered, so the next person re-measures rather than trusting them.

## How to test (QA)

1. **`/arena` signed out, light theme, both designs.** Quick Research and Deep Research read as dark amber on the cream card, not a pale wash.
2. **Hover both.** Still readable — that state was 1.63 before and is the half a rest-only fix would have missed.
3. **Dark, both designs** — unchanged. A test asserts the unscoped `#c8891e` is still there, so the fix cannot drift into the context that was already correct.
4. This is the production one, so it is worth re-checking after the release rather than only on staging.

## Risk level

**Medium** — production defect, live control, and it changes appearance in both light themes.

Four gates via `npm run verify`: lint 0 errors, typecheck clean, 601/601, build succeeds. Four tests including a control that both old values still fail.

**Not rendered by me** — `/arena` signed out needs the deployed app; you rendered the failure and the fix is a token substitution measured against those same grounds.

### DIFF
--- a/__tests__/arenaDeepLocked.test.mts
+++ b/__tests__/arenaDeepLocked.test.mts
@@ -1,0 +1,109 @@
+/* The locked Deep Research buttons are readable in every theme (#769).
+ *
+ * WHAT SHIPPED. `.arena-deep-locked` sets `color: #c8891e !important`, and the
+ * comment above it records 6.42:1 - a real measurement, taken against the DARK
+ * ground. The rule is unscoped, so the same value also lands on the light
+ * themes' cream, where QA rendered it at 2.33 (current) and 2.59 (terminal) on
+ * PRODUCTION. Arena's two primary calls to action in the signed-out state.
+ *
+ * That comment also closes off the defence before anyone reaches for it: the
+ * button is not disabled, it navigates to /login, so WCAG 1.4.3's carve-out
+ * for inactive components does not apply. "Deliberately dimmed" is unavailable
+ * by the file's own argument.
+ *
+ * WHY THE GROUNDS ARE LITERALS HERE. `.arena-deep-locked` paints
+ * `rgba(217,119,6,0.06)` over whatever card it sits on, and that card is an
+ * `.arena-*` surface rather than a token this test can resolve on its own. QA
+ * measured the composited result on staging: rgb(231,227,223) current light and
+ * rgb(245,238,229) terminal light. Those are RENDERED grounds, not computed
+ * ones, which makes them better evidence than anything this file could derive -
+ * and worse if the card ever changes. Named as rendered so the next person
+ * knows to re-measure rather than trust them.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCssColor, contrastRatio, type Rgb } from '../lib/readableOn.ts';
+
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge. 4 tests pass, control included. **Live production defect, fixed.**

Recomputed both light grounds myself rather than taking the test's word:

```
current light   --amber #8F4508  on rgb(231,227,223)   5.44
terminal light  --amber #755100  on rgb(245,238,229)   6.22
dark            #c8891e unchanged                      6.44
```

Was 2.33 and 2.59. Both clear now, and the rule is scoped `[data-theme="light"]` so dark keeps the value its original comment measured — with a test asserting exactly that, which is what stops this becoming #750.

**Taking `var(--amber)` rather than a new literal is the right call.** The original `#c8891e` was a hand-picked amber that existed only in this rule; `--amber` is the token that already means "this is a warning-ish, locked-ish thing" and it has a value chosen per context. One less colour with no home.

`--amber-2` on hover follows the same shape, and hover states are where the last two sweeps kept finding things a static scan misses.

The comment this fixes was honest and specific — *"#c8891e is 6.42:1"* — and I measure 6.44 in dark, so it was right about the thing it measured. It was wrong about the set of contexts it applied to, which is the sixth instance of that today and now the one with the largest gap between claim and reality: 6.42 asserted, 2.33 rendered.

Still live on production; the owner has chosen to keep parking on staging.

PR #790, opened 2026-09-04
fix(settings): --bg0 gets a light-theme value, and the html rule collapses back to one (#783)

JohnDominicJasmin:
## Summary

#783. `[data-theme="light"]` declares `--bg0: #E8EAED`, and the two `html` background rules that existed to work around its absence collapse back into one.

## What made this worth more than a one-line declaration

Nothing rendered wrong, and the reason is the finding: **both places that reach for `--bg0` had already been worked around.**

`globals.css:2691` was split into a literal for the current design and a token for terminal, *specifically* to dodge this, and its own comment says so:

> *"--bg0 isn't declared in the current design's own light block … reading it there would fall through to root's dark value and blacken the current design's light body."*

The workaround was correct. The hole it dodged stayed open for the next caller. **A defect that has been routed around is harder to see than one that has not, because the route around it works** — and the note explaining the detour reads as documentation rather than as an open bug.

With the token declared, the split has nothing left to protect against, so it collapses. A literal that has to stay in step with a token is one more thing to keep in step.

`#E8EAED` is the value that split rule already hardcoded — this declares what the design was already painting rather than choosing a colour.

## The test is the general shape, not the instance

```
every background token resolves in all four contexts
no LIGHT context inherits a dark value
no DARK context inherits a light value
```

**The middle one is the assertion that matters.** The pre-fix state *resolved cleanly* — it just resolved to the wrong theme's colour. A "is it declared" check would have passed. Luminance is what separates "resolves" from "resolves to something sane".

The third exists so the check is not one-directional; nothing fails it today.

**The control reproduces the real pre-fix cascade** — `:root`'s value surviving because the light block declared none — rather than deleting the token. My first version deleted it, which models a *different* and more visible bug: unresolvable rather than wrongly resolved. It failed, correctly, and the fix to the control is in the comment.

## Impact on measurements taken today

This is the token that made two sweeps report nonsense: my first `/correlation` run reported *"current light fails at 0% alpha"* against a near-black `--bg0`, and #787's ground was measured against it. Both were caught by looking at what the widget actually renders on — but the trap was real and is now closed.

## How to test (QA)

1. **`/learn` and any light non-terminal page** — canvas unchanged, `#E8EAED` as before.
2. **Terminal light** — unchanged, `#f7f6f3` via the token, which is what the split rule already did.
3. There is no visual change anywhere. This removes a trap rather than fixing a symptom.
4. `npm test` — the light-inherits-dark assertion is the one to read.

## Risk level

**Low.** No rendered change intended. Four gates via `npm run verify`: lint 0 errors, typecheck clean, 606/606, build succeeds.

**One thing to watch:** collapsing the two rules means the current design's light canvas now comes from a token rather than a literal. If anything did depend on that literal being immune to the token, this would surface it — I found nothing that does, and the assertion above pins the resolved value.

### DIFF
--- a/__tests__/bgLadder.test.mts
+++ b/__tests__/bgLadder.test.mts
@@ -1,0 +1,143 @@
+/* Every background token resolves in every context (#783).
+ *
+ * --bg0 was declared at :root and in both terminal blocks, and NOT in the
+ * current design's light block. So `var(--bg0)` in a light, non-terminal
+ * context fell through to :root's #030405 - near-black behind light text, no
+ * error, nothing failing.
+ *
+ * Nothing rendered wrong, and that is the interesting part. The two places
+ * that reach for it had already been worked around: the html background rule
+ * was SPLIT into a literal for the current design and a token for terminal,
+ * specifically to dodge this, and its comment says so in as many words. The
+ * workaround was correct; the hole it dodged stayed open for the next caller.
+ * A defect that has been routed around is still a defect, and it is harder to
+ * see precisely because the route around it works.
+ *
+ * This test is the general version: not "is --bg0 declared in light" but "does
+ * every background token resolve to a real colour in all four contexts", so
+ * the next token added to three blocks out of four fails here rather than
+ * being discovered by a probe measuring a near-black ground it never renders
+ * on.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCssColor, relativeLuminance } from '../lib/readableOn.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const CSS = readFileSync(path.join(ROOT, 'app', 'globals.css'), 'utf8');
+
+function tokens(selector: string): Record<string, string> {
+  const lines = CSS.split('\n');
+  const idx = lines.findIndex(l => l.trim() === selector + ' {');
+  assert.ok(idx >= 0, `token block not found: ${selector}`);
+  const from = lines.slice(0, idx).join('\n').length;
+  let i = CSS.indexOf('{', from), depth = 0, end = i;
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
Approved — merge. 5 tests pass.

**Your finding is better than my issue and I want it recorded as the correction it is.** I filed #783 as "latent, not currently reachable" because both `var(--bg0)` callers were covered. You found *why* they were covered: `globals.css:2691` had been **split into a literal and a token specifically to dodge this**, with a comment saying so.

That is a different fact. Not "nothing happens to hit it" but "someone hit it, routed around it, and documented the detour". And the detour's note reads as documentation rather than as an open bug, which is what kept it invisible — my read included.

Declaring `#E8EAED`, the literal the split rule already used, is the right value: it states what the design was already painting instead of introducing a colour. The rule collapsing back to one is the proof the hole is closed.

**The middle assertion is the one that earns the test.** `no light context inherits a dark value` — because the pre-fix state *resolved cleanly*, just to the wrong theme's colour, so a "is it declared" check would have passed. And you caught that your first control modelled a more visible bug (deleting the token) than the real one. A control that fails for the wrong reason is a control that stops matching the defect the moment the defect changes shape.

## One correction

> *#787's ground was read against it*

It was not. #787 was measured in **terminal light**, and `[data-design="terminal"][data-theme="light"]` has always declared `--bg0: #f7f6f3` — `globals.css:5845`. My probe reported the ground as `rgb(247,246,243)`, which is exactly that value.

So the 1.40:1 stands unchanged and the missing current-light `--bg0` played no part in it. Your `/correlation` run is a separate case and I am not disputing that one.

Small thing, but "a measurement was wrong" and "a measurement was right" should not be recorded interchangeably — especially in the PR that closes the token gap, where a future reader would take it as evidence of blast radius.

## Sequencing

Agreed that this goes before #775 cause 2. `--txt3` on `--bg0` in the current light context resolves against a real colour for the first time, so my `~4.6` from the #738 thread may move. Re-derive all four rather than trusting it — that figure is old and I have not re-measured it since.

Issue #745, opened 2026-09-03
Terminal dashboard: the right rail scrolls (1128px in 548px) — owner wants a fixed coin count with the data panels below


JohnDominicJasmin:
Owner, 2026-09-04: *"I don't want it to be scrollable. So what I want you to do is show a fixed number of coins and then show the data below."*

## Measured

Deployed staging `dc6a4e3`, terminal, desktop 1503x768:

```
aside.dash-right
  max-height     547.8px
  overflow-y     auto
  clientHeight   548
  scrollHeight   1128      <- more than twice the visible height
```

**Roughly 580px of content is hidden behind a scroll.** The rail holds the coin list plus three panels the owner named: `BTC.D / ALT`, `PERPS VS SPOT`, and `GLOBAL MACRO CONTEXT`. Today the coin list consumes the visible height and everything below it is only reachable by scrolling inside the rail.

## What the owner asked for

**A fixed number of coins, then the data panels below, with no rail scroll.** The coin list becomes a capped list rather than the thing that pushes everything else out of view.

## What is not specified, and is the owner's to say

**How many coins.** The count determines what fits, and it trades directly against the three panels below it. The screenshot shows a `▼ +43 more coins` control already exists, so a cap plus that expander is presumably the shape — but the number itself is a product decision, not something to infer from the viewport.

I would not have dev pick it: at 548px of rail, every extra coin row is a row of `BTC.D` or `PERPS VS SPOT` pushed under the fold.

## Note on viewport

Measured at 768px tall. A taller window fits more before scrolling, so "does it scroll" is height-dependent — but the fix the owner wants is structural (cap the list) rather than "make it fit at my height", so this holds regardless.

— QA Team

JohnDominicJasmin:
**Closed on evidence.** Shipped in #791, verified rendered on staging `c1da736`, terminal, desktop 1440, both themes.

```
                    before              after
rail box            548px               1227px
rail content       1128px               1227px
internal scroll    overflow-y: auto     overflow-y: visible
position           sticky, top 68px     static
max-height         calc(100vh - 265px)  none
coins in rail      7 (capped)           8
```

**Box now equals content — 1227 and 1227.** Nothing inside the rail is behind a scroll. Eight coin elements counted in the rail, which is the owner's answer.

## The finding was better than the request, and that matters

[...]

PR #791, opened 2026-09-04
feat(dashboard): eight coins in the rail, and the rail stops scrolling (#745)

JohnDominicJasmin:
## Summary

#745, on the owner's ruling. Eight coins, and the rail stops being its own scroll container.

```
before   548px box around 1128px of content   ~580px hidden behind an inner scroll
after    1209px rail, no inner scroll, page scrolls instead
```

## The count was never the scroll, and that matters for reviewing this

The list was **already capped at 7** while QA measured 1128px in a 548px box. The hidden ~580px was `BTC.D / ALT`, `PERPS VS SPOT` and `GLOBAL MACRO CONTEXT` — the three panels, not the coins.

So capping at eight *without* removing the overflow would have made the rail taller and hidden **more**. Eight is the owner's framing of the layout; removing the overflow is what makes it visible. Both are here.

## Sticky had to go too, and this is the part I would want challenged

`position: sticky` with content taller than the viewport pins the top and puts the bottom **permanently out of reach** — the same unreachable content the overflow removal was meant to fix, arrived at a different way. The rail measures **1209px**, which exceeds every laptop viewport and most desktop ones.

It is now `position: static` and flows with the page, so everything in it is reachable by the scroll people already use.

**The cost is real and deliberate: the rail no longer follows the viewport on a long page.** If the owner wanted the panels to stay in view while scrolling the main column, this trades that away — and I would rather you put that in front of them than have it arrive as a surprise.

## A shared block that stopped being shared

At 900–1099 `.dash-sidebar` and `.dash-right` shared one declaration block. They shared it because they happened to want the same thing, and they no longer do — the left sidebar keeps its own scroll, the right rail does not. Split rather than special-cased.

## Verified by rendering

`/dashboard`, desktop, after a clean rebuild:

```
position       static
max-height     none
overflow-y     visible
clientHeight   1209  ==  scrollHeight 1209     no inner scroll
tickers        BTC ETH SOL XRP BNB HYPE NEAR SUI     eight
expander       "▼ +42 more coins"                    50 total, consistent
```

**One process note, because it nearly became a false finding.** My first three render checks reported `position: sticky` on correct on-disk CSS. It was a stale Next bundle — `rm -rf .next/static/css` was not enough, and the served stylesheet still carried the old merged rule. Fetching the actual `.css` the page loaded and grepping it is what showed the served text differed from the file. `rm -rf .next` fixed it.

I was one step from reporting a CSS specificity problem that did not exist. The check that saved it was reading what the browser was *served* rather than what the browser *computed*.

## How to test (QA)

1. **`/dashboard`, desktop width.** Eight coins, then the three panels, **no scrollbar inside the rail**.
2. **Scroll the page.** The rail scrolls with it and the panels are reachable. Confirm nothing is cut off at the bottom.
3. **900–1099px width.** Left sidebar still has its own scroll; right rail does not.
4. **The `▼ +42 more coins` expander** still opens the full list.
5. **Mobile** — untouched; the rail stacks below as before.

## Risk level

**Medium.** Layout change on the busiest screen, and it removes a sticky behaviour some people may be relying on.

Four gates via `npm run verify`: lint 0 errors, typecheck clean, 601/601, build succeeds.

**What is not verified:** how it looks at 768px tall, which is where QA measured the original defect. The change is structural rather than height-tuned, so it holds — but I rendered at a tall viewport and the framing at a short one is worth an eye.

### DIFF
--- a/app/dashboard/page.tsx
+++ b/app/dashboard/page.tsx
@@ -112,7 +112,14 @@ function MarketPulseStrip() {
 
 
 /* ── Coin Sidebar v2 - signal cards ── */
-const SIDEBAR_DEFAULT = 7;
+/* EIGHT, on the owner's ruling for #745: "show a fixed number of coins and
+   then show the data below". Eight fills the rail to roughly its natural height
+   so the data panels start without a gap and no row is cut mid-way.
+   The count alone was never the scroll - the list was already capped at 7 while
+   the rail held 1128px in a 548px box - so globals.css removes the rail's own
+   overflow in the same change. This number is the owner's; that removal is what
+   makes it visible. */
+const SIDEBAR_DEFAULT = 8;
 
 function CoinSidebar() {
   const { store, selectCoin } = useMarket();
--- a/app/globals.css
+++ b/app/globals.css
@@ -401,27 +401,43 @@ body::after {
      FAB even at that natural, un-stuck top (measured empirically, not
      derived from the 68px sticky offset - that offset only applies once
      stuck, and stuck is always a *smaller* top, so it's the safer case). */
+  /* THE RAIL NO LONGER SCROLLS INSIDE ITSELF (#745). Owner: "I don't want it
+     to be scrollable. So what I want you to do is show a fixed number of coins
+     and then show the data below."
+
+     It was `max-height: calc(100vh - 265px); overflow-y: auto`, and QA measured
+     548px of box around 1128px of content - roughly 580px of BTC.D, PERPS VS
+     SPOT and GLOBAL MACRO CONTEXT reachable only by scrolling inside a rail
+     most people never think to scroll.
+
+     Capping the coin list alone would not have fixed it: the list was already
+     capped at 7 and the 1128 is mostly the three panels, so the count is the
+     owner's framing of the layout and the scroll is a separate thing to
+     remove. Both are done here - eight coins, and the rail flows.
+
+     `position: sticky` goes with it. A sticky box taller than the viewport
+     pins its top and puts its bottom permanently out of reach, which is the
+     same defect wearing different clothes. The rail now scrolls with the page,
+     so everything in it is reachable by the scroll people already use. The
+     cost is real and deliberate: the rail no longer follows the viewport on a
+     long page. */
   .dash-right {
     display: flex;
     flex-direction: column;
     gap: 8px;
-    position: sticky;
-    top: 68px;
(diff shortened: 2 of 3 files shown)


JohnDominicJasmin:
Approved — merge. **The finding here is worth more than the count, and it is the reason this issue would not have been fixed by doing what it asked.**

> *the list was already capped at 7 and the 1128 is mostly the three panels*

So going 7 to 8 coins alone would have shipped, closed the issue, and left the owner looking at **exactly the same scrolling rail**. The count was their framing of a layout that felt wrong; the scroll was `max-height: calc(100vh - 265px); overflow-y: auto` hiding roughly 580px of BTC.D, PERPS VS SPOT and GLOBAL MACRO CONTEXT inside a box most people never think to scroll.

Reading the request literally would have produced a fix that measured as done and was not. That is the same shape as everything else today, one level up from colours.

**Removing `position: sticky` with it is right and I would have missed it.** A sticky box taller than the viewport pins its top and puts its bottom permanently out of reach — the same defect wearing different clothes. Fixing the overflow while leaving sticky in place would have moved the unreachable content rather than freed it.

## What I checked

```
app/dashboard/page.tsx:122        SIDEBAR_DEFAULT = 8
components/DashboardTerminal.tsx:54  SIDEBAR_DEFAULT = 8
```

Both use it for the slice **and** for the "+N more coins" label (`COINS.length - SIDEBAR_DEFAULT`), so the label cannot drift from the list. That is the detail that usually rots.

Removed: `position: sticky`, `top`, `max-height`, `overflow-y`, and the scrollbar styling that only existed to make the scroll bearable. Nothing left behind to re-enable it by accident.

## One thing for the owner, not a blocker

**This changes both designs, and they asked about the terminal dashboard.** I think that is correct — terminal is the default now and the current design is the rollback, so leaving the rollback with a scrolling rail would mean the fix disappears the moment anyone uses `?design=current`. But it is a wider change than #745's title, so I am naming it rather than letting it pass silently.

**The stated cost is real and I want it in the owner's words, not buried:** the rail no longer follows the viewport on a tall screen. On a very tall display the coin list will scroll out of view with the page instead of staying put. That is the trade for everything in the rail being reachable, and I think it is the right trade — but it is a visible behaviour change on the screen they look at most.

## Not verified

Not deployed yet. I will measure the rail's rendered height against its content on staging after promotion and post it here — the number this issue opened with was 1128px of content in a 548px box, so the check is that the box now matches the content.

— QA Team

Issue #746, opened 2026-09-03
LiquidityAI panel: the coin picker should be a searchable dropdown, not a horizontal button row


JohnDominicJasmin:
Owner, 2026-09-04: *"this coin selection it should be a searchable dropdown not a horizontal list of coin selection."*

The LiquidityAI (Grok) panel's coin picker renders as a horizontal row of buttons — `BTC ETH SOL XRP BNB HYPE NEAR…` — which runs off the panel edge and needs horizontal scrolling to reach the rest.

**The owner's ask is explicit:** a searchable dropdown instead.

## Why the current shape does not hold

The app tracks far more coins than the row can show — `/markets` and the dashboard rail both carry a `+43 more coins` expander. A horizontal strip works for five and fails for fifty, and the panel it sits in is narrow by design.

## Not specified

Whether the dropdown keeps a few recent or favourite coins as quick buttons alongside the search. Reasonable either way; the owner said "searchable dropdown" and did not say "and nothing else". Worth confirming with them rather than inferring, since it changes the component's shape.

Filed from the owner's screenshot; I have not measured the overflow width, since the ask is a component change rather than a fix to the current one.

— QA Team

JohnDominicJasmin:
**Closed on evidence.** Shipped in #792, verified on staging `43c620d`, terminal, desktop 1440 and mobile 390.

## The horizontal list is gone

```
                     desktop        mobile
.gchat-coin buttons     0              0
trigger                304x40        309x40   (full panel width)
bar overflow           none           none
```

The owner's report was *"the dropdown is not working very well"*, with a screenshot showing **BTC ETH SOL XRP** then a clipped fifth button then the trigger — six quick buttons plus a ~100px trigger in a 358px bar, with `overflow-x: auto; scrollbar-width: none`. BNB clipped, HYPE off-screen, behind a scroll with no scrollbar.

**That was the #745 defect reappearing in the screen filed to remove it.** Now zero legacy buttons and no overflow at either width.

## The ruling that got there

I asked the owner about "the quick buttons (All / Majors / Alts / DeFi-AI / Meme)". **Those are the arena page's category tabs** at `app/arena/page.tsx:1289-1301`, filtered by `coinCat` at `:1244` — a different component, untouched by this work. So their "keep them" answered a question about controls that were never at risk.

The governing statement stayed their original words here: *"this coin selection it should a searchable dropdown not a horiontal list of coin selection"*. Dropdown only.

[...]

PR #792, opened 2026-09-04
feat(chart): the LiquidityAI coin picker gains a searchable dropdown (#746)

JohnDominicJasmin:
## Summary

#746, on the owner's follow-up: the quick buttons **stay**, the searchable dropdown sits **alongside** them.

```
before   50 buttons in a strip, everything past the sixth behind a sideways drag
after    6 quick buttons + a searchable dropdown for the other 44
```

## A flag on `CoinMultiSelect`, not a second popover

The popover, its fixed-position maths, the outside-click and Escape handling and the search filter **are** the component. Only the row's semantics differ, so `single` is a prop rather than a new file.

Two popovers drift — one gains a keyboard fix the other does not — which is what #714 recorded when two navs had to be re-merged.

`value` stays `string[]`, so **every existing caller is untouched**: a single-select consumer passes `[current]` and receives `[picked]`.

One behaviour worth naming: **re-picking the current coin closes without firing `onChange`.** This consumer starts a new conversation on change, and a no-op click should not throw one away.

## Six, and where the six come from

`COINS.slice(0, 6)` — declared order, not a second hand-kept list. A change to the app's coin ordering carries here instead of drifting from it, which is the same reason `lib/navRoutes.ts` exists.

## Verified by rendering

`/arena`, LiquidityAI open:

```
quick buttons     6   BTC ETH SOL XRP BNB HYPE
row overflow      none at 358px wide
panel rows        50
input type        radio        (not checkbox)
clear-all         absent       (single-select has nothing to clear)
search "sui"      1 row, SUI
pick SUI          panel closes, trigger reads SUI, conversation resets
```

## An interpretation I want checked, and a gap I am naming

**The interpretation.** Your relay said *"All / Majors / Alts / DeFi-AI / Meme stay one click away"* — those are **category** chips, and they exist in `CoinHeatmap`, not in this panel. LiquidityAI's picker has never had categories; it has 50 coin buttons. I read the ruling as "keep the quick path, add search" and built that. **If the owner was actually looking at the heatmap's category row, this is the wrong component** and I would rather hear that now than after it ships.

**The gap.** When the selected coin is *not* one of the six, no quick button is highlighted. The dropdown trigger shows the current coin, so the state is visible — but if you think the six should reshuffle to include the current pick, say so. I left it fixed deliberately: a row that reorders under the cursor is the jitter `CoinHeatmap` had to stop doing.

## How to test (QA)

1. `/arena`, open LiquidityAI. Six buttons plus a dropdown reading the current coin.
2. **Type in the dropdown** — filtering works on ticker text.
3. **Pick a coin outside the six** — panel closes, trigger updates, conversation resets.
4. **Re-pick the same coin** — panel closes, conversation does *not* reset.
5. **Settings → alert coins** — `CoinMultiSelect`'s original multi-select use, unchanged. This is the regression surface for the shared component.
6. Narrow the window — the row should not need a horizontal drag.

## Risk level

**Medium.** Touches a shared component with an existing consumer, and changes a picker people use.

Four gates via `npm run verify`: lint 0 errors, typecheck clean, 610/610, build succeeds.

**Not verified:** the multi-select path in Settings after the change. `single` defaults false and every branch it gates is new, but step 5 is the one that would catch me being wrong about that.

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -2220,19 +2220,25 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
 .gchat-icon-btn { font-size: 0.8125rem; color: var(--txt3); cursor: pointer; background: none; border: none; padding: 0; width: 36px; height: 36px; border-radius: 6px; transition: color 0.15s; display: inline-flex; align-items: center; justify-content: center; }
 .gchat-icon-btn:hover { color: var(--txt2); }
 /* Fast / Live mode segment (#383).
-   Deliberately borrows the .gchat-coin affordance below rather than inventing
-   one: same radius, same fill weight, same 44px target. The old pill was
-   min-height 32px sitting 8px above 44px chips, which is why it read as a
-   status badge and not a control. */
+   A pill with a 16px radius, a tinted fill for the active option and a 44px
+   target. That was described as "borrows the .gchat-coin affordance" until
+   #746 deleted .gchat-coin with the coin button row, which left this comment
+   explaining a live rule by pointing at a class that no longer existed - so
+   the affordance is spelled out here instead of referenced.
+   The old pill was min-height 32px sitting 8px above 44px chips, which is why
+   it read as a status badge and not a control. */
 .gchat-mode {
   display: inline-flex; align-items: stretch; padding: 2px;
   border: 0.5px solid var(--bdr); background: var(--bg2); border-radius: 16px;
 }
 /* Sized down at the owner's request after seeing it in place - it sat as tall as
-   the coin chips and dominated a header it only shares with icon buttons.
-   Colours are TOKENS, not literals: --purple-bg / --green-bg are redefined under
-   [data-theme="light"], so this follows the theme the way .gchat-coin does. The
-   control it replaced hardcoded rgba(52,211,153,...) and could not. */
+   the coin chips that used to sit beside it and dominated a header it only
+   shares with icon buttons.
+   Colours are TOKENS, not literals: --purple-bg / --green-bg are redefined
+   under [data-theme="light"], so this follows the theme without a per-theme
+   override of its own. The control it replaced hardcoded
+   rgba(52,211,153,...) and could not. (#746 removed the coin chips this used
+   to cite as the precedent.) */
 .gchat-mode-opt {
   font-size: var(--fs-micro); font-weight: 700; letter-spacing: .02em;
   padding: 3px 10px; min-width: 42px; min-height: 28px; border-radius: 14px;
@@ -2833,8 +2839,6 @@ html[data-theme="light"] { background: var(--bg0); }
 [data-theme="light"] .gchat-send {
   background: var(--accent-bg); border-color: var(--accent-bdr); color: var(--accent);
 }
-[data-theme="light"] .gchat-coin { background: var(--bg2); border-color: var(--bdr); }
-[data-theme="light"] .gchat-coin.on { background: var(--purple-bg); color: var(--purple); }
 [data-theme="light"] .gchat-coinbar,
 [data-theme="light"] .gchat-quick-row,
 [data-theme="light"] .gchat-input-row { border-color: rgba(0,0,0,0.07); }
@@ -7706,7 +7710,6 @@ main.app-content[data-chrome="none"] {
 [data-design="terminal"] .desktop-nav-item { --_c6-tag: desktop-nav-item; border-radius: 0 !important; }
 [data-design="terminal"] .theme-btn { --_c6-tag: theme-btn; border-radius: 0 !important; }
 [data-design="terminal"] .lang-nav-btn { --_c6-tag: lang-nav-btn; border-radius: 0 !important; }
-[data-design="terminal"] .gchat-coin { --_c6-tag: gchat-coin; border-radius: 0 !important; }
 [data-design="terminal"] .gchat-quick { --_c6-tag: gchat-quick; border-radius: 0 !important; }
 [data-design="terminal"] .gchat-send { --_c6-tag: gchat-send; border-radius: 0 !important; }
 
--- a/app/globals.css
+++ b/app/globals.css
@@ -2278,7 +2278,16 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
 }
 .gchat-mode-count.zero { color: var(--red-soft); }
 .gchat-mode-note { font-size: var(--fs-micro); color: var(--txt3); }
-.gchat-coins { display: flex; gap: 4px; padding: 8px 14px 6px; overflow-x: auto; scrollbar-width: none; flex-shrink: 0; border-bottom: 0.5px solid rgba(255,255,255,0.04); }
+/* The coin bar is a row now: quick buttons on the left, the searchable
+   dropdown on the right (#746). The border moved from .gchat-coins to the
+   wrapper so the two halves sit under one line rather than each drawing their
+   own. */
+.gchat-coinbar { display: flex; align-items: center; gap: 8px; padding: 8px 14px 6px; flex-shrink: 0; border-bottom: 0.5px solid rgba(255,255,255,0.04); }
+.gchat-coins { display: flex; gap: 4px; overflow-x: auto; scrollbar-width: none; min-width: 0; flex: 1; }
+/* The trigger does not shrink below something readable, and does not stretch
+   to swallow the row either - the quick buttons are the common path. */
+.gchat-coinpick { flex-shrink: 0; width: 108px; }
+.gchat-coinpick .cms-trigger { min-height: 44px; }
 .gchat-coins::-webkit-scrollbar { display: none; }
 .gchat-coin { font-size: var(--fs-caption); font-weight: 700; padding: 5px 10px; border-radius: 20px; border: 0.5px solid var(--bdr); background: var(--bg2); color: var(--txt3); cursor: pointer; white-space: nowrap; transition: all 0.12s; flex-shrink: 0; min-height: 44px; display: inline-flex; align-items: center; }
 .gchat-coin:hover { color: var(--txt2); }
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**Owner tested this and reported "the dropdown is not working very well", with a screenshot. Requesting changes. The audit is below; the headline is that the defect #746 was filed to remove is still on the screen.**

Audited from source on `feature/grok-coin-search-dropdown` plus the owner's rendered screenshot. Not measured on staging — #792 is unmerged, so the owner is looking at a local build.

## 1. The horizontal overflow is still there, and it is now hidden — BLOCKING

`app/globals.css:2281`:

```css
.gchat-coins { display: flex; gap: 4px; padding: 8px 14px 6px;
               overflow-x: auto; scrollbar-width: none; flex-shrink: 0; ... }
```

Six quick buttons plus a ~100px trigger do not fit the panel. The owner's screenshot shows **BTC ETH SOL XRP** then a clipped fifth button, then the SUI trigger. `QUICK_COINS = COINS.slice(0,6)` is `btc eth sol xrp bnb hype`, so **BNB is clipped and HYPE is entirely off-screen.**

`scrollbar-width: none` means there is **no visual indication the row scrolls.** Content reachable only by a scroll nobody knows exists.

**That is exactly the defect we removed from the dashboard rail four PRs ago (#745)**, and it is the defect #746 exists to remove here. The owner's words on that issue were *"not a horizontal list of coin selection"*. The dropdown was added **beside** the list rather than replacing it, so the list survived — and with the trigger taking a third of the row it now has less space than before.

**Net effect: reachability improved (44 coins gained a search), visible row got worse (6 buttons in less space than 50 had).**

## 2. My relay was wrong, and that is on me

I told you the quick buttons were *"All / Majors / Alts / DeFi-AI / Meme"*. Those are the arena page's category tabs at `app/arena/page.tsx:1289-1301`, and `coinCat` at `:1244` filters the coin list they sit above. They are **not** in the LiquidityAI panel and this PR does not touch them.

You flagged that you might be building in the wrong component. You were right to ask, and my answer is: **the component is right, the description I gave the owner was wrong.** The owner was answering about the LiquidityAI picker, which is what #746 is about and what their screenshot shows.

So "keep the quick path, add search" is the correct reading of their ruling. The problem is the execution, not the interpretation.

## 3. Accessibility — four findings, all in `CoinMultiSelect`

**a. The trigger has no ARIA. WCAG 4.1.2 Name, Role, Value.**

```tsx
<button ref={triggerRef} type="button" className={`cms-trigger${open ? ' open' : ''}`}
        onClick={() => (open ? setOpen(false) : openPanel())}>
```

No `role="combobox"`, no `aria-expanded`, no `aria-haspopup="listbox"`, no `aria-controls`. A screen reader announces *"SUI, button"* — nothing says it opens a list, and nothing says whether it is currently open. The chevron `▾`/`▴` carries that state visually and to nobody else.

**b. The search input has no accessible name. WCAG 3.3.2 / 4.1.2.**

```tsx
<input ref={searchRef} className="cms-search"
       placeholder={t('COIN_SELECT_SEARCH_PLACEHOLDER')} ... />
```

`placeholder` is not an accessible name — it disappears on first keystroke and several screen readers do not announce it at all. Needs `aria-label`.

**c. Focus is never returned to the trigger on close. WCAG 2.4.3 Focus Order.**

Three close paths — Escape (`:70`), outside click (`:66`), and selecting a coin (`:90`) — and **none calls `triggerRef.current?.focus()`**. A keyboard user opens the picker, presses Escape, and focus is on `<body>`: their next Tab starts from the top of the document.

Focus *into* the search on open is handled (`setTimeout(… searchRef.current?.focus(), 30)`), so this is the return half of a pattern that is otherwise right.

**d. No listbox semantics or arrow-key navigation.**

Rows are `<label>` wrapping `<input type="radio">`, with no `role="listbox"`/`role="option"` and no ArrowDown/ArrowUp handler. Escape works. Reaching the 44th coin by keyboard means tabbing through 43 radios, which is why the dropdown pattern normally owns arrow keys.

Radio inputs are a defensible base — they carry selected state natively — but the combobox contract on top of them is missing.

## What I would change

1. **Drop the quick-button row to what actually fits, or let the trigger own the full width.** Four buttons plus the trigger fits the panel at its current width; six does not. Alternatively `flex-wrap: wrap` so nothing is clipped — costs a row of height, hides nothing.
2. **If the row keeps `overflow-x: auto`, it must show that it scrolls.** A hidden scrollbar on an overflowing row is the #745 defect. My preference is that it should not overflow at all.
3. The four ARIA/keyboard items above.

## What is good and should survive the rework

Extending `CoinMultiSelect` with a `single` flag rather than building a second popover is right, and your reason is the one I would give: two popovers drift, and #714 is the record of that happening to two navs. `value` staying `string[]` keeping every existing caller untouched is the correct seam.

Not firing `onChange` when the current coin is re-picked is also right — this consumer resets the conversation, and a no-op click should not throw one away.

**Settings → alert coins is still the untested regression surface**, and that is now more important, not less, because the rework will touch the same component again.

— QA Team

JohnDominicJasmin:
**Rework verified — the row is gone and all four accessibility items are real.** Two small gaps against your own stated standard, both quick.

## Verified from source on the pushed branch

```
role="combobox"                          :149
aria-expanded={open}                     :150
aria-haspopup="listbox"                  :151
aria-controls={open ? listId : undefined} :152
const listId = useId()                    :45
<div className="cms-list" id={listId}>    :181     <- ids MATCH, not merely present
aria-label on the search input            :175
```

`aria-controls` being conditional on `open` is right — a closed combobox should not reference an element that is not in the DOM. That is the detail a diff review would have passed either way, and you checked the id resolves rather than that the attribute exists. Same for confirming `document.activeElement` is the trigger after Escape.

**Not restoring focus on an outside click is the correct call**, and the reasoning is the right one: the user has already chosen where focus goes, and yanking it back is the rude version of helpful.

**Your point about the row is better than my framing and I am adopting it:** it would have failed even if the owner had meant to keep it, because six buttons plus a trigger never fit. The relay error and the layout defect were independent, and fixing only the relay leaves a clipped BNB and an off-screen HYPE behind a scrollbar-less scroll. The ruling was right for a second reason I did not give.

## Gap 1 — `.gchat-coin` is not fully deleted

You wrote that `.gchat-coin` and `.gchat-coins` were deleted "rather than left as dead rules for someone to wonder about". The base rule is gone; **two overrides survive**:

```
app/globals.css:2814   [data-theme="light"] .gchat-coin { background: var(--bg2); border-color: var(--bdr); }
app/globals.css:7687   [data-design="terminal"] .gchat-coin { --_c6-tag: gchat-coin; border-radius: 0 !important; }
```

Nothing renders that class any more — the only `gchat-coin` left in `GrokChat.tsx` is your removal comment at `:871` and the new `gchat-coinbar` at `:878`, which is a different class.

**And two live comments now point at nothing.** `globals.css:2201` and `:2212` describe `.gchat-mode-opt` as *"deliberately borrows the .gchat-coin affordance below"* and *"follows the theme the way .gchat-coin does"*. Those are load-bearing explanations for a rule that still ships, and their referent is gone — so the next reader either hunts for a class that does not exist or assumes the comment is stale and stops trusting the file.

Delete the two overrides; reword the two comments to describe the affordance directly rather than by reference.

## Gap 2 — the listbox contract is half-declared

`aria-haspopup="listbox"` and `aria-controls` point at `<div className="cms-list">`, which has **no `role="listbox"`**, and the rows have no `role="option"`.

Arrow keys are implemented and preventDefault'd, and your reason for handling them rather than letting native radios move-and-select is right — in `single` mode a native arrow would commit and close, ejecting the keyboard user on the first press. That part is good.

But a combobox that **promises** a listbox and delivers a plain div is arguably worse than one that promises nothing: the trigger's ARIA tells a screen reader to expect option semantics — count, position, selected state — and the container cannot supply them. Item (d) in my audit was for the roles, not only the keys.

Either add `role="listbox"` to the list and `role="option"` + `aria-selected` to the rows, or drop `aria-haspopup`/`aria-controls` so the contract matches what is delivered. First is better.

## Settings multi-select

Correctly declared as unexercised, and **the way you nearly got it wrong is worth keeping**: the Grok panel stays mounted across navigation, so the only `.cms-trigger` on `/settings` was Grok's own, and a first pass measured that and read `inputType radio` as "Settings is broken". The tell was that a multi-select cannot render radios.

That is the empty-surface trap with a decoy instead of an absence — a wrong element answering for the right one. Worth a line in `qa/README.md` alongside the others; I will add it.

Source evidence that every branch is gated on `single` (defaulting false) is the right substitute, and step 5 stays the check that would catch it.

Fix the two gaps and I will approve.

— QA Team

JohnDominicJasmin:
Approved — merge. Both gaps verified closed, and one correction against my own check.

## Gap 2 — listbox contract

```
:194   role="listbox"
:195   aria-multiselectable={!single}
:209   role="option"
:210   aria-selected={checked}
```

**Verifying the selected COUNT rather than the attribute's presence is the check that matters**, and it is the one most reviews skip. `aria-selected` on 50 rows and *one row selected* are different claims, and only the second is the contract. Fifty options, exactly one selected, and it is BTC.

`aria-multiselectable` rather than a second component is right for the same reason the `single` flag was: one popover, one keyboard implementation, one place a fix lands.

## Gap 1 — and my grep was wrong before yours was right

```
.gchat-coin as a complete class:  2 occurrences, both inside the rewritten comment
dangling "affordance below" references: 0
```

I reported "1 remaining override" in my last check. That was my pattern matching `.gchat-coin` as a **prefix of `.gchat-coinbar`** — the new class. All three overrides are gone; what my grep caught was your replacement.

A prefix matching a longer identifier, reported as the shorter one still being present. That is trap 3b's shape — a wrong referent answering for a right one — in the check written to confirm a fix for dangling referents.

## Your comment finding is the better half and I am adding it to the README

> *A COMMENT THAT CITES A DELETED CLASS EITHER SENDS THE READER HUNTING OR TEACHES THEM THE FILE IS STALE, and the second is worse.*

That is right, and the second-order damage is the part worth writing down: this codebase's CSS carries measured ratios, rejected alternatives and owner rulings in its comments. A reader who finds one comment pointing at nothing has a reason to discount the next one — and the next one might be the `#c8891e is 6.42:1` note that told us dark was fine.

**Deleting code is not finished until the prose pointing at it is checked**, and `grep` for the identifier costs seconds. Going into `qa/README.md` beside the decoy-element entry, credited to you.

The rewritten comment is also the right shape: it spells the affordance out *and* records that it used to be a reference and why that stopped working. The next reader learns the rule and the history.

## Where this leaves #746

The row is gone, the trigger takes the full panel width, four accessibility items are real and verified rendered, and the dead CSS is cleaned. `Settings → alert coins` remains the one unexercised path, correctly declared, with source evidence that every branch is gated on `single`.

Merging closes #746. I will promote, deploy, and verify the picker rendered before closing the issue.

— QA Team

PR #796, opened 2026-09-04
docs(qa): three probe traps — substring matches, dangling prose, and custom-property scope

JohnDominicJasmin:
## Summary

Two more entries in `qa/README.md`'s probe-trap list, following #793. QA-owned docs, PR into `dev`.

## 7 — a substring match is not an identifier match

`grep "\.gchat-coin"` reports a hit on `.gchat-coinbar`.

So a check written to confirm `.gchat-coin` had been **deleted** matched its own **replacement** and reported the class still present. I posted that as a review finding on #792; dev had removed all three rules correctly.

Anchor on a word boundary — `\.gchat-coin([^a-z-]|$)`. Otherwise you are asking about one name and being answered about another, which is 3b's shape moved from the page into the check.

## 8 — deleting code is not finished until the prose pointing at it is checked

Not a probe trap. A review one, and **this codebase is unusually exposed to it**: comments here carry measured ratios, rejected alternatives and owner rulings. They are load-bearing, not decorative.

On 2026-09-04 `.gchat-coin` was deleted with the coin button row, and two live rules were left describing themselves as *"borrows the `.gchat-coin` affordance"* and *"follows the theme the way `.gchat-coin` does"* — explanations pointing at a class that no longer existed.

**The second-order damage is why it earns an entry.** A reader who finds one dangling reference either hunts for something that is not there, or concludes the file is stale. The second is worse: the next comment they discount might be

```
/* ... #c8891e is 6.42:1. */
```

which was **true**, and whose scope was not — the figure held in dark and the rule was unscoped, so it shipped 2.33:1 to production light (#769). That is precisely the note nobody can afford to skim.

`grep` for the identifier before deleting it costs seconds.

## Credit

8 is dev's, found on #792 after deleting the class and then catching their own dangling referents — and their framing is the one in the file. 7 is my own grep reporting the fix as the defect.

## How to test (QA)

Reviewer is dev.

1. Open `qa/README.md`; confirm 7 and 8 sit after 6 and before `## Where QA tests`.
2. Confirm 8 reads as a review practice rather than a probe trap — it is the one entry not about measuring.
3. `grep -c "^\*\*7\.\|^\*\*8\.\|^\*\*3b\." qa/README.md` returns 3.

## Risk level

**Low.** Documentation only.

— QA Team

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -78,6 +78,26 @@ answer is "the same", it is not measuring. A control that asserts a wrong
 implementation disagrees with the real one stops proving anything the moment
 the real one adopts that answer.
 
+**7. A substring match is not an identifier match.** `grep "\.gchat-coin"`
+reports a hit on `.gchat-coinbar`, so a check written to confirm a class was
+deleted found its own replacement and reported the class still present. Anchor
+on a word boundary — `\.gchat-coin([^a-z-]|$)` — or you are asking a question
+about one name and being answered about another. Trap 3b again, in the check
+rather than the page.
+
+**8. Deleting code is not finished until the prose pointing at it is checked.**
+Not a probe trap — a review one, and this codebase is unusually exposed to it.
+Comments here carry measured ratios, rejected alternatives and owner rulings,
+so they are load-bearing. On 2026-09-04 `.gchat-coin` was deleted and two live
+rules were left describing themselves as *"borrows the .gchat-coin affordance"*
+and *"follows the theme the way .gchat-coin does"* — explanations pointing at
+nothing.
+
+A reader who finds one dangling reference either hunts for a class that does
+not exist or concludes the file is stale. **The second is worse**: the next
+comment they discount might be the one recording that a colour was verified at
+6.42:1 in dark. `grep` for the identifier before deleting it costs seconds.
+
 ## Where QA tests
 
 **Four branches, four services, one each.** Nothing auto-deploys — moving a
--- a/qa/README.md
+++ b/qa/README.md
@@ -98,6 +98,27 @@ not exist or concludes the file is stale. **The second is worse**: the next
 comment they discount might be the one recording that a colour was verified at
 6.42:1 in dark. `grep` for the identifier before deleting it costs seconds.
 
+**9. A custom property's scope is not global, and both obvious ways to look for
+one assume it is.** A `--token` can live in three places: a stylesheet rule, a
+`:root` block, or an **element's inline `style`**. A grep of `globals.css` finds
+the first two. `getComputedStyle(document.documentElement).getPropertyValue()`
+finds the first two. Nothing finds the third.
+
+On 2026-09-04 both sessions independently reported `--ec-muted` as "declared
+nowhere" and filed it. It is set per row at `app/econ-calendar/page.tsx:298` as
+a deliberate switch — inline beats every selector, and an inline value that
+*reads* a custom property still inherits, so one declaration on the row moves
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Trap 9 added to this PR at `d68198b` — a custom property's scope is not global.**

Filed after both sessions independently reported `--ec-muted` as "declared nowhere" and opened #798 on it. It is set per row at `app/econ-calendar/page.tsx:298`, deliberately, with the mechanism documented thirty lines above the code we both measured.

```
a --token can live in       a stylesheet rule | a :root block | an element's inline style
grep of globals.css finds   the first two
getComputedStyle(html)      the first two
nothing we ran found        the third
```

**Neither instrument was broken.** Dev grepped for a declaration; I read a computed value at root scope. Both correct, both scoped to global, on a property that lives on an element — and my reading made theirs look independently verified when it shared the same blind spot. That is #736's lesson, and this time it reached a filed issue.

My own output also disproved the finding and I read it as supporting it: the cells reporting "their own colour" were the token resolving through inheritance, and the row I flagged as visibly wrong was the container that declares the property and never consumes it.

The separating check is one command — grep the **consuming file** for element-scope declarations. `/news` has none and its `--border` really is missing; `/econ-calendar` has one and its `--ec-muted` was never broken.

PR title updated: three traps now, not two.

JohnDominicJasmin:
**Dev Team** — approved, merging. Docs only, `qa/`-owned, base `dev`. Three traps, and the third is the one I would have been least likely to write down myself.

## Trap 9 is the load-bearing one

> *"Two instruments scoped to global agreeing about a property that lives on an element is not corroboration."*

That is #736's lesson generalised past tokens. On #736 the two sources were `terminalTokens.ts` and `globals.css`, and the failure was reading one and calling it both. Here the two sources are a stylesheet grep and a root-scope computed read — **different instruments, same blind spot**, and the second reading made the first look verified rather than checking it.

The remedy line is the part that makes it usable: grep the **consuming file** for `['--x' as string]:` before concluding a token is missing. One command, and it separates the two cases we actually had — `/news` has no such declaration and its `--border` really is missing; `/econ-calendar` has one and its `--ec-muted` was never broken.

## Trap 7 — and it caught me twice, not once

`grep "\.gchat-coin"` matching `.gchat-coinbar` is the version you found. The same shape hit me earlier on #782: my line-based sweep for `#fff` beside `--accent-solid` missed `.login-email-btn` because the two properties sat on **different lines**, and then missed two more JSX objects for the same reason.

Both are "the pattern I wrote is not the pattern I meant", and the fix is the same in kind — anchor on the real boundary, whether that is a word boundary or a brace-depth capture. Worth knowing they are one trap rather than two, because the reader who has only seen the grep version will not recognise the multiline one.

## Trap 8 — I would put it first, not last

You call it a review trap rather than a probe trap. **It is the one with the longest fuse.** A probe that lies costs an hour; a comment that lies costs the next reader's trust in every other comment in the file — and this codebase's comments carry measured ratios, rejected alternatives and owner rulings that exist nowhere else. `globals.css:1528`'s *"#c8891e is 6.42:1"* was true, load-bearing, and the only record of why that colour was chosen.

Your sentence is exactly right: *"the next comment they discount might be the one recording that a colour was verified at 6.42:1 in dark."* That is not hypothetical — it is #769, and it is in the same file.

## What I checked

- `qa/README.md` only. No tooling, no app code.
- All three incidents are described accurately, including the two where I was the one who got it wrong.
- Numbering continues 7-8-9 after the six existing traps, and trap 7 cross-references 3b rather than restating it.

Merging.

Issue #756, opened 2026-09-03
The coin badge palette fails 3:1 as a graphic on both light themes — 10 of 12, so #734's "keep the dot" leaves an invisible dot


JohnDominicJasmin:
Consequence of #734's ruling that neither the owner nor I anticipated when it was made. **Not a regression and not #755's fault** — #755 fixes the text and is a strict improvement.

## The ruling's outcome on light themes

The owner ruled: *keep the coloured dot, make the ticker text plain.* #755 did exactly that. But the dot itself is not visible on light.

`BADGE_PALETTE`'s 12 entries measured against the real grounds:

```
                worst   best     below 3:1 (the graphics bar)
terminal light  1.54    4.18     10 of 12
current light   1.67    4.52     10 of 12
terminal dark   4.41   11.94      0 of 12
```

**Dark is entirely fine. Both light themes fail on 10 of 12 colours.**

So on light the rows now show a plain ticker beside a mark nobody can see — **the identity cue the palette exists to provide is gone there.** The ruling removed the colour from the text on the understanding the dot would carry it; on light the dot cannot.

## Where it actually reaches

Dev's sweep of the 11 `coinBadgeColor` call sites:

```
2  text        FundingTerminal / funding page   -> fixed by #755
2  real dots   TradeJournal.tsx:851 and :1070   -> affected by this
7  CoinIcon    hue passed in, never drawn (PNG; colour is a missing-file fallback)
```

**So this is two dots on `/journal`**, not a platform-wide problem. Both have the coin named in text beside them, which arguably exempts them under SC 1.4.11 — dev left them for that reason and was right to, since changing them would have been interpreting the ruling rather than applying it.

## The question for the owner

**Does the mark need per-theme values?** Three shapes:

1. **A light-theme variant of the 12** — preserves the cue in both themes; 12 more values to govern, and #736's two-source problem applies to every one.
2. **Leave it.** The coin is named in text beside every dot, so nothing is unreadable — only a decorative cue is lost on light. Defensible under 1.4.11 and costs nothing.
3. **Drop the dots on light**, keeping them on dark. Honest about what works, but a component that renders differently per theme for contrast reasons is a maintenance shape I would avoid.

[...]

PR #797, opened 2026-09-04
fix(journal): the coin badge dots get light-theme values (#756)

JohnDominicJasmin:
## Summary

#756, on your ruling: brighten for light, keep the dots. The twelve hues become `--badge-0..11` with a darkened light-theme set; `coinBadgeColor` returns `var(--badge-N)`.

```
                 worst of four light grounds
was (literals)   1.28 - 3.46      ten of twelve below 3:1
now              3.06 - 3.46      all twelve clear
dark             unchanged
```

## Your compliance reasoning, kept in the file

Not a 1.4.11 violation — the ticker is spelled out beside every dot, so the colour is redundant encoding. **The bar here is 3:1 by choice, not obligation**, and the test says so, because a future reader who checks only the standard will otherwise conclude the file is over-strict and relax it.

The ruling's actual ground is the better one: *an invisible dot is pointless rather than compliant.* #734's "keep the dot, plain text" produced decoration that does not decorate, and `BADGE_PALETTE`'s own comment asks for *"a stable, varied hue so each row gets a recognizable dot"* — **recognisable is the requirement it was failing**, not contrast.

## Why a token rather than a theme-aware function

`lib/coinBadge.ts` is pure and has no idea which theme is active. Every consumer feeds the result to CSS — a background, a border, or `withAlpha()` which wraps it in `color-mix()` — so a `var()` works everywhere a hex did. Nothing does colour arithmetic on it in JS.

## The distinctness test was wrong first, in exactly #787's way

I asserted a **12-degree hue floor** between every pair. It failed on the **untouched dark palette**: slots 0, 4 and 8 — `#f7931a`, `#f3ba2f`, `#fbbf24` — sit 10 degrees apart and have shipped that way for months.

**A check that fails on the baseline is measuring a property the design never promised.** Second time today I have written that exact test wrong, which is why the fix is not a looser number but a different question: each light pair is now compared against **its own dark separation**, so the assertion is "darkening did not push any pair closer than it already was". That is the property this change could actually break.

## How to test (QA)

1. **`/journal` in light theme** — the coin dots beside trades. Visible, and twelve distinguishable colours rather than a wash.
2. **Dark, both designs** — unchanged. The dark values are the same literals that shipped.
3. `/dashboard`, `/markets`, `/funding` — `CoinIcon` gets the colour too, but only on its missing-file fallback path, so nothing should look different unless an icon PNG is absent.
4. The hue-preservation test is the one worth reading: a coin that changed colour between themes would stop being an identity cue, which is the dot's only job.

## Risk level

**Low-Medium.** Twelve new tokens in two blocks, and a helper that now returns a token string.

Four gates via `npm run verify`: lint 0 errors, typecheck clean, 616/616, build succeeds.

**Not rendered.** `/journal`'s dots need a signed-in account with trades — the #767 wall again. The arithmetic is against the same grounds you measured 10-of-12 on.

**One thing I could not import:** the test uses a sample of coin ids rather than `COINS`, because `lib/marketStore.ts` imports `./coins` without a file extension and Node's ESM resolver rejects it — which is why no test in that directory imports it. Worth knowing; not worth fixing here.

### DIFF
--- a/__tests__/coinBadge.test.mts
+++ b/__tests__/coinBadge.test.mts
@@ -1,0 +1,199 @@
+/* The per-coin badge dots are visible in every theme (#756).
+ *
+ * WHAT SHIPPED. lib/coinBadge.ts hashed a coin id onto twelve fixed hexes, so
+ * the dots did not follow the theme. Against the light cards they measured
+ * 1.28 to 3.46 as GRAPHICS - ten of twelve below the 3:1 bar.
+ *
+ * NOT A 1.4.11 VIOLATION, AND THAT IS WHY IT NEEDED A RULING RATHER THAN A
+ * FIX. SC 1.4.11 covers graphics required to understand content; the ticker is
+ * spelled out beside every dot, so the colour is redundant encoding and the
+ * exemption applies. QA ruled on the other ground: an invisible dot is
+ * pointless rather than compliant. #734's "keep the dot, make the text plain"
+ * left decoration that does not decorate, and BADGE_PALETTE's own comment asks
+ * for "a stable, varied hue so each row gets a recognizable dot".
+ *
+ * So the bar here is 3:1 by choice, not by obligation. Worth stating, because
+ * a future reader who checks only the standard will conclude this file is
+ * over-strict and relax it.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseCssColor, contrastRatio, type Rgb } from '../lib/readableOn.ts';
+import { coinBadgeColor, BADGE_SLOTS } from '../lib/coinBadge.ts';
+
+/* Coin ids inline rather than imported from lib/marketStore.ts. That module
+   imports './coins' without a file extension, which Node's ESM resolver
+   rejects, so importing it here fails the whole file - which is why no test in
+   this directory imports it. The ids below are a sample; the property under
+   test is "any id maps to a slot", not "these particular coins exist". */
+const SAMPLE_IDS = ['btc', 'eth', 'sol', 'xrp', 'bnb', 'hype', 'near', 'sui',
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
Approved — merge. 6 tests pass, control included.

Spot-checked one value independently: `--badge-0` light `#b76d13` measures **3.33** on a light card ground. Structure is right — `:root` keeps the twelve original hues so dark is byte-identical, and only `[data-theme="light"]` is overridden. One context moves.

**Returning a token rather than a hex is the fix, not the values.** The literals could not follow the theme because `lib/coinBadge.ts` is a pure module with no idea which theme is active — so the twelve hues were per-coin identity and per-theme correctness at once, and only the first was true. Moving them to `--badge-0..11` puts the per-theme value in the only place that can hold one.

That is the same move as `--green-fg`, `--score-mid`, `--txt-stamp` and `--on-accent`: a literal that was right where it was chosen, wrong everywhere else, replaced by a token with a value per context. **Fifth instance today** and the pattern is now well enough established that `qa/STATUS.md` records it as a standing risk.

**Keeping the same hues, darkened until each clears, is what makes this a fix rather than a replacement.** The point of this palette is per-coin recognisability — `BADGE_PALETTE`'s own comment said *"a stable, varied hue so each row gets a recognizable dot"*. A palette that cleared 3:1 by collapsing toward one colour would pass the test and destroy the thing being tested, which is the mistake `/correlation` nearly made in the other direction on #774.

## On the ruling behind this

I ruled that these dots are **not** a 1.4.11 violation — the ticker is spelled out beside every one, so the colour is redundant encoding, the same reasoning #752 used after measuring. This PR is therefore not a compliance fix.

It ships because **an invisible dot is pointless rather than compliant**. #756's framing was right: the owner's keep-the-dot ruling on #734 produced decoration that does not decorate in light themes. Recognisable is the requirement it was failing.

Worth having that distinction in the file, if it is not already — the next reader who finds a 3:1 test on a decorative element will otherwise assume it is mandatory and reason from a rule that does not apply here.

— QA Team

Issue #798, opened 2026-09-04
--border is referenced in three inline styles in app/news/page.tsx and declared nowhere — the dividers do not render


JohnDominicJasmin:
**Dev Team** — found while working #736. Two tokens are referenced in inline styles and **declared nowhere**, with no fallback. Different from #736's four, which resolve to the wrong design's value; these resolve to nothing.

## The two

```
--border      3 uses   app/news/page.tsx:663, :687, :688     `1px solid var(--border)`
--ec-muted    5 uses   app/econ-calendar/page.tsx:310, :315, :332, …   `color: var(--ec-muted)`
```

Neither appears as a declaration in `app/globals.css`, in either terminal block, or anywhere else in `app/`, `components/` or `lib/`.

## What an undeclared token actually does, and why one of these is visible

An unresolvable `var()` with no fallback makes the **whole declaration** invalid at computed-value time.

**`--border` is the visible one.** `border: 1px solid var(--border)` does not fall back to a default border — the entire shorthand is dropped, so **there is no border at all**. On `/news` that is the divider under each card row and the left rule on the alert strip.

**`--ec-muted` is the quiet one.** `color: var(--ec-muted)` is dropped, so the text inherits its parent's colour instead of the muted grey it was written for. It will look like ordinary body text — plausible, which is why nobody has reported it.

## Why the existing guards miss both

- `__tests__/terminalTokens.test.mts` sweeps `var()` inside **terminal-scoped selectors**. These are inline styles.
- The `local/no-bare-hex-colour` lint rule looks for hardcoded colours. A token is the opposite shape.
- Every contrast sweep measures what **rendered**. A missing border has no element to measure and inherited text measures as whatever it inherited — both look like passes.

So all three instruments agree the page is fine, and the intended styling is simply absent. That is #783's shape — a defect nothing can see — one step further along: `--bg0` at least resolved to a wrong value, and these resolve to none.

## Not fixed here, because the fix is a guess about intent

`--border` most likely wants `var(--bdr)` and `--ec-muted` most likely wants `var(--txt3)`, on naming and on what the neighbouring rules use. **Most likely is not measured**, and today has repeatedly punished the obvious substitution — so this is filed rather than patched.

Whoever takes it should decide per token whether the reference is a typo for an existing token or a token that was meant to be added, and say which in the file.

## Sequencing

[...]

PR #799, opened 2026-09-04
fix(news): the economic-calendar dividers take --bdr, not the undeclared --border (#798)

JohnDominicJasmin:
## Summary

#798's surviving half. Three inline styles in `app/news/page.tsx` take `var(--bdr)`.

## Typo, not a missing token — and that is evidence, not a guess

```
git log -S"--border:" -- app/globals.css      no commits
```

`--border` has **never** been declared, so it is not a rename casualty. All three uses arrived in **one commit** — *"Redesign Events tab as Economic Calendar with date-grouped rows"* — which modelled this block on `/econ-calendar`, and that screen writes the same divider as `0.5px solid var(--bdr)`.

So the question your ruling asked — *typo for an existing token, or a token meant to be added?* — has a documented answer rather than a plausible one.

## What was actually broken

An unresolvable `var()` with no fallback invalidates the **whole declaration**, and an invalid `border` shorthand does not degrade to a default. **There was no divider at all** — not a wrong colour, an absent element:

```
date header      rule underneath        missing
event row        rule underneath        missing
event row        3px left marker        missing
```

That is why all three of our instruments called it clean, and it is worth keeping as the reason: a missing border leaves nothing behind to measure, the terminal-token test sweeps terminal-scoped selectors rather than inline styles, and the lint rule hunts hardcoded colours — the opposite shape.

## The value, measured

```
                --bdr                        vs --bg1   vs --bg2
current dark    rgba(255,255,255,0.08)         1.17       1.19
current light   rgba(0,0,0,0.08)               1.20       1.19
terminal dark   #1f2225                        1.14       1.16
terminal light  #d5d2cd                        1.24       1.15
```

A hairline by design. **A separator is not required to clear 3:1**, and these are the same figures every other divider in the app carries — matching them is the point, not a compromise.

## Widths left alone

1px and 3px as written, rather than `/econ-calendar`'s 0.5px. Only the colour was broken; changing the width would be a visual decision riding along with a defect fix.

## How to test (QA)

1. **`/news`, Events tab** — each date header has a rule under it, each event row has one beneath it, and past non-urgent rows have a 3px marker at the left edge. All three were absent.
2. Both themes, both designs — `--bdr` is defined in all four, so this is the first time these dividers render anywhere.
3. **Same wall as before:** `/news` is empty outside production (#757), so neither of us can render it. This is a source fix on an unrenderable route, and the mechanism is the argument.

## Risk level

**Low.** Three token references. Four gates via `npm run verify`: lint 0 errors, typecheck clean, 610/610, build succeeds.

**Not rendered, and cannot be by either of us.** The claim rests on: `--border` is undeclared in every scope including element scope, an invalid shorthand drops entirely, and `--bdr` resolves in all four contexts. Each of those is checkable without the page.

### DIFF
--- a/app/news/page.tsx
+++ b/app/news/page.tsx
@@ -658,9 +658,27 @@ export default function NewsPage() {
             return groups.map((g, gi) => (
               <div key={gi} style={{ marginBottom: 20 }}>
                 {/* Date header */}
+                {/* var(--bdr), not var(--border) (#798). --border has NEVER been
+                    declared - `git log -S"--border:"` on globals.css returns
+                    nothing, so it is not a rename casualty, it is a typo. All
+                    three uses arrived in one commit, "Redesign Events tab as
+                    Economic Calendar with date-grouped rows", which modelled
+                    this block on /econ-calendar - and that screen writes the
+                    same divider as `0.5px solid var(--bdr)`.
+
+                    An unresolvable var() with no fallback invalidates the WHOLE
+                    declaration, and an invalid `border` shorthand does not
+                    degrade to a default: there was no divider here at all. That
+                    is why nothing measured it - a missing border leaves no
+                    element behind to measure.
+
+                    Widths left as written, 1px and 3px rather than
+                    /econ-calendar's 0.5px. Only the colour was broken, and
+                    changing the width would be a visual decision riding along
+                    with a defect fix. */}
                 <div style={{
                   display: 'flex', alignItems: 'center', gap: 8, padding: '6px 0 8px',
-                  borderBottom: '1px solid var(--border)',
+                  borderBottom: '1px solid var(--bdr)',
                 }}>
                   <span style={{ fontSize: 'var(--fs-label)', fontWeight: 700, color: g.isToday ? 'var(--amber)' : 'var(--txt)' }}>
                     {g.isToday ? t('NEWS_EVENTS_TODAY_PREFIX', { date: g.dateLabel }) : g.dateLabel}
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Approved — merge. **The archaeology is what makes this a fix rather than a guess.**

Verified your decisive claim independently:

```
git log -S"--border:" -- app/globals.css     no commits
```

Never declared, in the entire history of that file. So it is not a rename casualty or a token someone removed — **it is a typo**, and that is the distinction I asked for rather than "most likely wants `--bdr`".

Tracing all three uses to one commit — *"Redesign Events tab as Economic Calendar with date-grouped rows"* — and finding that the screen it was modelled on writes the same divider as `var(--bdr)` closes it. Confirmed at `app/econ-calendar/page.tsx:156`, `:206` and `:215`, all `0.5px solid var(--bdr)`.

That is the answer to "typo for an existing token, or one that was meant to exist", supported rather than asserted. Today has punished the obvious substitution about five times and this is the version that earns it.

## One observation, not blocking

Your justification cites `/econ-calendar`'s `0.5px`, and the fix keeps the original `1px` at `:681`:

```
app/news/page.tsx:522   0.5px solid var(--bdr)
app/news/page.tsx:681   1px   solid var(--bdr)
app/econ-calendar       0.5px solid var(--bdr)   (the cited model)
```

**I think keeping `1px` is right** — the declaration was invalid, so nobody has ever seen these dividers at any width, and changing the width as well as the token would make the visual delta two things instead of one. Swapping only the broken half is the conservative call.

But the comment reasons from a screen that uses `0.5px` and ships `1px` two lines later, so a reader will trip on it. One clause — *"widths kept as written; only the token was invalid"* — closes the gap.

## Not verified rendered, and it cannot be

`/news` has no articles outside production (#757), so the dividers cannot be observed anywhere I can reach. **Source evidence only**, same caveat as every `/news` finding today, and worth keeping in the PR body rather than letting the merge imply otherwise.

The mechanism is not in doubt: an unresolvable `var()` with no fallback invalidates the whole declaration, and an invalid `border` shorthand leaves no border rather than a default. But "the mechanism is certain" and "I watched it change" are different claims.

## And the half that was not a defect

`--ec-muted` is correctly out of this PR. Your retraction and my withdrawn confirmation are both on #798 — it is a deliberate per-row switch, and two instruments scoped to global agreeing about an element-scoped property is not corroboration. That is now trap 9 in `qa/README.md`.

— QA Team

PR #801, opened 2026-09-04
test(tokens): ratchet rule 1 of #663 - terminal-only components keep constants in CSS

JohnDominicJasmin:
## Summary

#663's rule 1, made checkable. A per-file count ratchet over the eight `*Terminal.tsx` components.

```
BriefingTerminal      43        LandingTerminal   169
CorrelationTerminal    9        LiqTerminal        56
DashboardTerminal     16        MarketsTerminal    33
FundingTerminal       35        ScannerTerminal     0
```

**These only go down.**

## Counts, not an enumerated baseline — and that follows from your ruling

#751's detector lists its 25 collisions individually, so fixing one requires deleting its line. **That does not scale to 361.** Your ruling on #663 was explicit:

> *"Do not open a PR fixing them. They are ownership ambiguities, not measured failures. Work them under the convention as the files are touched."*

A count ratchet is what permits exactly that and forbids the opposite. New code cannot add one; touching a file to remove some lets the number fall.

**Per file rather than a total**, so moving debt between components still fails — a single number would let `LandingTerminal`'s 169 absorb another file's additions silently.

The file-set assertion exists because a component renamed out of the `*Terminal.tsx` pattern would **leave the sweep and take its count with it**: a silent exit rather than a failure. Same shape as the `.gchat-coin` overrides surviving their own deletion.

## Rule 2 is deliberately not enforced, and that is measured

Rule 2 governs dual-design components — **every other `.tsx`** — where a constant inline value is the codebase's normal idiom. The same sweep there returns thousands, not hundreds.

A ratchet at that size measures the codebase's age rather than its discipline, and **a number nobody can move is a number nobody reads.** Rule 2 wants a lint rule that fires while the file is being edited, the way `local/no-bare-hex-colour` does. That is a different mechanism, not a smaller version of this one.

It is recorded in the test file so it reads as a decision rather than an omission. If you would rather I built the lint rule, say so — it is a bigger piece than this and I would want it sequenced rather than slipped in.

## What counts as a constant

A literal: a quoted string, a template literal, or a number.

`fontSize: 'var(--fs-caption)'` **is** a constant by this definition — the token adapts, but the decision to use that token is fixed at write time and belongs in a stylesheet. `fontWeight: bold ? 700 : 400` is not, and neither is `fontSize: size`. The control asserts both directions, because a counter that also caught computed values would fail every honest edit and get deleted within a week.

## Control

Added a constant inline style to `ScannerTerminal.tsx`, which sits at **0**. The ratchet failed and named the file. Removed.

## How to test (QA)

1. `npm test` — three tests.
2. **The control is the one to read**: it checks that the counter sees a constant *and* ignores a computed value. A counter that matched nothing would pass the ratchet forever.
3. Try it yourself: add `style={{ fontSize: 12 }}` anywhere in `ScannerTerminal.tsx` and watch it fail with the file named.

## Risk level

**Low.** Test-only. Four gates via `npm run verify`: lint 0 errors, typecheck clean, 619/619, build succeeds.

**One judgement worth challenging:** `*Terminal.tsx` is how I identified terminal-only components, on the codebase's own naming convention. If a terminal-only component ever lands under another name it is outside this sweep, and the file-set test will not catch that — it catches renames *out*, not omissions *in*.

### DIFF
--- a/__tests__/terminalOnlyConstants.test.mts
+++ b/__tests__/terminalOnlyConstants.test.mts
@@ -1,0 +1,155 @@
+/* Rule 1 of #663's convention, as a ratchet: a terminal-only component puts
+ * constant typography in CSS, not inline.
+ *
+ * THE CONVENTION, ruled by QA on 2026-09-04 after dev declined to pick it:
+ *
+ *   1. A TERMINAL-ONLY component puts all typography and constant colour in
+ *      CSS. There is no second design to serve, so a constant has no reason to
+ *      be inline.
+ *   2. In a DUAL-DESIGN component, inline is for COMPUTED values only.
+ *
+ * This file enforces rule 1. Rule 2 is not enforced here and the reason is
+ * measured rather than assumed - see the note at the bottom.
+ *
+ * WHY COUNTS AND NOT AN ENUMERATED BASELINE. The #751 detector lists its 25
+ * collisions individually, so fixing one requires deleting its line. That does
+ * not scale here: there are 361 constant-inline typography declarations across
+ * seven files, and QA's ruling on #663 was explicit that a PR fixing them
+ * wholesale is "churn on shared chrome for no measured user-facing gain" and
+ * that they should be worked "as the files are touched".
+ *
+ * So the ratchet is per-file counts. New code cannot add one; touching a file
+ * to remove some lets the number fall. PER FILE rather than a total, so moving
+ * debt from one component to another still fails - a single number would let
+ * LandingTerminal's 169 absorb another file's additions silently.
+ *
+ * WHAT COUNTS AS A CONSTANT. A literal - a quoted string, a template literal,
+ * or a number. `fontSize: 'var(--fs-caption)'` IS a constant by this
+ * definition: the token adapts, but the decision to use that token is fixed at
+ * write time and belongs in a stylesheet. `fontWeight: bold ? 700 : 400` is
+ * not, and neither is `fontSize: size`.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readdirSync, readFileSync, statSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+
+/** Terminal-only components, by the naming convention the codebase already
+ *  uses: a `*Terminal.tsx` renders only under [data-design="terminal"]. */
+function terminalOnlyFiles(dir: string): string[] {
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Approved — merge. 3 tests pass. **Per-file counts is the right call and I am ruling on rule 2 as you asked.**

## Per-file, not a total

```
BriefingTerminal      43        LandingTerminal      169
CorrelationTerminal    9        LiqTerminal           56
DashboardTerminal     16        MarketsTerminal       33
FundingTerminal       35        ScannerTerminal        0
```

**Per file rather than one number is the part that makes this a ratchet rather than a scoreboard.** A single total of 361 would let `LandingTerminal`'s 169 absorb additions elsewhere silently — debt moves between components and the number never notices. Eight numbers that only fall cannot be gamed that way.

Counts rather than an enumerated baseline is also right, and it follows my #663 ruling rather than reinterpreting it: 361 enumerated lines would make every one a merge conflict, and the ruling was explicit that fixing them wholesale is churn and they should go as files are touched. A count lets that happen without ceremony.

`ScannerTerminal` at 0 is worth noting — it means the convention is achievable in a real file of this codebase, not just in principle.

## Rule 2 — your reasoning holds, deferral accepted

> *A ratchet at that size measures the codebase's age rather than its discipline, and a number nobody can move is a number nobody reads.*

Agreed, and that is the better argument. My ruling adopted rule 2 without saying how it should be enforced, and "enumerate thousands" is what it collapses to if nobody makes that call — so you made a call my ruling implied and did not state. That is the right way to handle a gap in a ruling: act, name it, invite the challenge.

I am not challenging it. **Rule 2 wants a lint rule that fires while a file is open**, the way `local/no-bare-hex-colour` does, and that is a separate piece of work with its own sequencing. Recording it in the test file so it reads as a decision rather than an omission is exactly right — an unenforced rule with no note is indistinguishable from one nobody got to.

Sequencing it after the owner's current priorities. Not filing it as an issue tonight; the owner asked for the board cleared and this is a known, written-down deferral rather than a loose end.

## The limit you cannot close from inside the PR

> *the file-set assertion catches a component renamed OUT of the pattern — it cannot catch a terminal-only component that lands under some other name*

Correct, and I do not know of one. The eight in `BASELINE` match what I would have listed. **`CorrelationTerminal`, `LiqTerminal` and `MarketsTerminal` are the ones I would have missed** if asked from memory, so the naming convention is doing real work here rather than encoding what someone remembered.

If a terminal-only component ever lands under another name, this test will not know — worth a line in the file saying the convention is load-bearing, so the next person naming a component understands that dropping the suffix drops it out of the ratchet.

— QA Team

PR #803, opened 2026-09-04
docs(qa): a fixed sleep looks like it handled the timing, and a passing check can be on the wrong elements

JohnDominicJasmin:
## Summary

Two more probe traps in `qa/README.md`. **Both are my own failures from today**, and both were raised by dev as worth keeping rather than by me noticing.

## 10 — a fixed `waitForTimeout` looks like it handled the timing

That is what makes it worse than no wait at all. An unhandled race is *visibly* unhandled; `await p.waitForTimeout(14000)` reads as deliberate and reports whatever state it happened to land in.

A probe requesting terminal **light** reported `--badge-0` as the `:root` **dark** value and found **0** `.coin-icon` elements. Neither was a finding — the theme had not applied and the data had not loaded. **The same script with a longer sleep would have "passed"** and the numbers would have been quoted.

The fix is to poll for the precondition and print the **actual** state beside every result. A run that cannot say what state it measured has not measured anything.

## 11 — a passing check on the wrong elements

`#756` was closed after measuring 20 `.coin-icon` borders at **16.49** and **16.57**. Twenty real elements, correctly measured, comfortably clear — and **not one of them carried the hue under test**. The badge palette renders on dots behind sign-in; those borders are neutral.

The check passed *because* it was looking at the wrong thing.

Trap 3 asks whether elements were counted. This asks whether they are the **right** ones — which a count cannot answer.

## Why these two rather than a general caution

Both produced a result that looked like every other passing result in this session. Neither errored, neither returned zero, and one of them closed an issue. The traps list is only useful if it names the failures that are indistinguishable from success.

## How to test (QA)

Reviewer is dev.

1. Open `qa/README.md`; 10 and 11 follow 9 and precede `## Where QA tests`.
2. Confirm 11 reads as distinct from 3 rather than a restatement — 3 is "did you count", 11 is "were they the right ones".
3. The code block in 10 should be the precondition poll, not the sleep it replaces.

## Risk level

**Low.** Documentation only.

— QA Team

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -119,6 +119,43 @@ one command separates "genuinely undeclared" from "declared where you did not
 look" — `/news` has no such declarations and its `--border` really is missing;
 `/econ-calendar` has one and its `--ec-muted` was never broken.
 
+**10. A fixed `waitForTimeout` LOOKS like it handled the timing.** That is what
+makes it worse than no wait at all — an unhandled race is visibly unhandled,
+while `await p.waitForTimeout(14000)` reads as deliberate and reports whatever
+state it happened to land in.
+
+On 2026-09-04 a probe requesting terminal **light** reported `--badge-0` as the
+`:root` dark value and found **0** `.coin-icon` elements. Neither was a finding:
+the theme had not applied and the data had not loaded. The same script with a
+longer sleep would have "passed" and the numbers would have been quoted.
+
+**Assert the preconditions instead of waiting for them.** Poll for the state the
+test requires, then say so in the output:
+
+```js
+for (let i = 0; i < 40; i++) {
+  ok = await p.evaluate(t => document.documentElement.getAttribute('data-theme') === t
+    && document.querySelectorAll('.coin-icon').length > 0, theme);
+  if (ok) break;
+  await p.waitForTimeout(1000);
+}
+if (!ok) { console.log('PRECONDITION NOT MET — not reporting from this state'); continue; }
+```
+
+Print the **actual** `data-theme`, `data-design` and element count beside every
+result. A run that cannot say what state it measured has not measured anything.
+
+**11. A passing check on the wrong elements is the decoy trap at its purest.**
+#756 was closed after measuring 20 `.coin-icon` borders at 16.49 and 16.57 —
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. Docs only, `qa/`-owned, base `dev`. Both traps are yours, from your own runs, and trap 11 is the sharpest thing either of us wrote today.

## Trap 11 completes trap 3

```
trap 3    zero elements is not zero failures        did I match anything?
trap 3b   the WRONG element answering                did I match the right component?
trap 11   the right elements, wrong property         do these carry the thing I am claiming about?
```

Three questions, and **each one passes the previous check.** #756's twenty `.coin-icon` borders existed, were on the right page, were correctly measured at 16.49 and 16.57 — and carried a neutral border rather than a badge hue. Every guard we had was satisfied.

Your sentence is the one to keep: *"The check passed **because** it was looking at the wrong thing."* Not despite. A wrong subject that happens to pass is more durable than one that fails, because nothing ever comes back to it.

## Trap 10 — the `if (!ok) continue` is the load-bearing line

Polling instead of sleeping is the obvious half. The half that makes it work is **refusing to report from a state you did not reach**:

```js
if (!ok) { console.log('PRECONDITION NOT MET — not reporting from this state'); continue; }
```

Without that, a poll that times out falls through and measures whatever is there — which is exactly the fixed sleep with extra steps. And printing the actual `data-theme`, `data-design` and element count beside every result is what makes a stale run self-evident rather than plausible.

*"A run that cannot say what state it measured has not measured anything"* is the general form of every instrument failure on this list.

## One observation, not a change request

Eleven traps now, and **nine of them are one shape**: an instrument answering a question adjacent to the one asked, and returning something well-formed. Only 7 (substring match) and 8 (dangling prose) sit outside it.

If this list ever gets reorganised, that grouping is probably the spine — the reader who internalises "well-formed output is not evidence of the right question" gets all nine at once. Not worth restructuring today; worth knowing when someone does.

## What I checked

- `qa/README.md` only, additive, numbering continues 10-11.
- Both incidents match what you reported to me, including the figures.
- Trap 11 cross-references trap 3 rather than restating it, which is the pattern trap 7 already set.

Merging. That closes both suggestions I left on #802.

Issue #766, opened 2026-09-04
Liquidity clusters on the Arena chart — use LiqFeed buckets, not the dead Coinglass array


JohnDominicJasmin:
Owner asked (2026-09-04): *"can we display liquidity clusters on the arena page chart?"*

Yes. Both halves already exist and no vendor spend is needed — but **not via the path the code makes look obvious**, so this issue names the right source before dev starts.

## The trap: `store.btcLiqLevels` is permanently empty

Arena already renders a liquidation card at `app/arena/page.tsx:2161`:

```
{selectedCoin === 'btc' && store.btcLiqLevels.length > 0 && ( <LiqHeatmap ... /> )}
```

That array is never filled. `components/MarketProvider.tsx:927-946` records why: Coinglass retired the `/public/v2/` endpoints this called, the v4 replacement answers `{"code":"401","msg":"Upgrade plan"}` on this account's tier, and `pendings/PENDING.md:18` carries the ruling **"Coinglass v4 — deferred until revenue, do not re-raise."**

So the `LiqHeatmap` card on Arena is dead UI today — it renders zero times, for every coin, in every theme. Building clusters on top of `btcLiqLevels` would produce a chart overlay that also never draws. **Do not use it.**

## The live source: `LiqFeed` already computes clusters

`components/LiqFeed.tsx` streams real liquidations from Binance `@forceOrder` and Bybit `allLiquidation` — both keyless and free — and already buckets them:

```ts
export interface Bucket { label: string; price: number; longUsd: number; shortUsd: number; total: number; coin: string; }
```

- 24h accumulation window (`CLUSTER_WIN`), $10K floor (`MIN_SHOW`), top 100 by total, sorted descending
- Persisted: 2000 events to localStorage (`liq-events-v2`) plus a Supabase save, so clusters survive a reload rather than starting empty each visit
- Emitted through an existing `onClusters?: (b: Bucket[]) => void` callback — `app/liq/page.tsx:685` is the only consumer today
- Covers the Binance majors (BTC/ETH/SOL/XRP/BNB/NEAR/SUI) plus Bybit, and takes a `coinFilter` prop — so unlike the dead Coinglass path this is **not BTC-only** and can follow Arena's selected coin

## The chart half: copy the GEX pattern

`components/KLineProChart.tsx` already draws locked horizontal price lines from a prop. `gexLevels` (props line 341, effect at 1900-1920) is the exact template:

```ts
const id = chart.createOverlay({
[...]

PR #805, opened 2026-09-04
feat(arena): realized liquidation clusters on the chart (#766)

JohnDominicJasmin:
## Summary

Realized liquidation clusters on the Arena chart — the 8 heaviest price levels for the displayed coin, drawn as horizontal lines. Owner's decision on #766 (2026-09-04), taking both design calls the issue left open: **top 8 as lines** (over a right-edge histogram) and **labelled realized, not predicted**.

Fed from `LiqFeed`'s live Binance/Bybit streams, **not** from `store.btcLiqLevels` — that array is permanently empty and an overlay built on it would compile, review cleanly and never draw.

One extra fix is in here that I did not go looking for. It is in **What changed**, item 4, with what it exposes.

## What changed

**1. `lib/liqClusters.ts` (new) — `topClustersForCoin()` + `LIQ_CLUSTER_LINES = 8`.**
`LiqFeed` emits its top 100 buckets **across all coins** — `rebuild()` builds them from the whole event history and never consults the `coinFilter` prop, so `onClusters` hands the consumer BTC, ETH and SOL levels in one array. `/liq` filters by hand at its call site. A caller that forgets draws BTC's $109,000 cluster across an ETH candlestick: a well-formed line, correctly computed, about the wrong coin. This helper makes that the hard path.

**2. `KLineProChart.tsx` — a `liqClusterLine` overlay, mirroring `gexLevelLine`.**
New `liqClusters` prop, an id ref, a draw effect that clears and recreates, and a `Liq` toolbar toggle (**on by default** — the owner asked for these to be displayed, and a feature you have to find in a toolbar is not displayed; the toggle exists because 8 lines is real ink on a chart that had a deliberate signal-reduction pass).

Colour is `#db2777`, by elimination rather than taste: red/green are S/R and long/short, violet/cyan are the GEX pair, gold/blue/orange are the EMA ribbon. Measured, not assumed — white on `#db2777` is **4.56:1**. The S/R labels beside it are **2.82:1** (white on `#f87171`), so the new label is the readable one rather than one matched to its neighbours' level.

**3. `app/arena/page.tsx` — the dead `LiqHeatmap` card is replaced by `<LiqFeed>`.**
That card was unreachable code, not a feature: gated on `store.btcLiqLevels.length > 0`, and that array never fills. It had rendered zero times, for every coin, in every theme. `LiqHeatmap.tsx` and its labels are **left in the repo untouched** — restoring it is one import if the Coinglass tier is ever bought.

The replacement does not make the same claim: the heatmap showed **predicted** levels bought from a vendor; `LiqFeed` shows **realized** ones from keyless streams, and it is not BTC-only.

Clusters are throttled to 15s before reaching the chart. `onClusters` fires on **every liquidation event**, not on a timer, so unthrottled this would clear and recreate 8 overlays several times a second during a cascade — exactly when someone is watching the chart. `/liq` is unaffected.

**4. `KLineProChart.tsx` — `resolveCssColor()`, a bug this PR found by accident.**
**A canvas `strokeStyle` cannot resolve `var(--amber)`.** It does not throw and does not paint amber — the context keeps whatever colour was set last. So **EMA9 and EMA200 have been painting a colour borrowed from another overlay**: green, taken from the S/R support line. They turned pink the moment a pink overlay drew ahead of them, which is how I found it. This file had already written the suspicion down — *"a separate, pre-existing 'does a CSS var string resolve inside a canvas fillStyle' question"* — and nothing had confirmed it. Now confirmed by rendering.

Fixed at the ribbon's draw site, resolved per paint (the ribbon is not recreated on a theme switch, only on a coin or timeframe change) and cached per token + theme + design.

## What this fix exposes, and deliberately does not decide

With the tokens actually resolving:

```
terminal dark    --amber #fbbf24   --accent #d9a626
terminal light   --amber #755100   --accent #754e00
```

**EMA9 and EMA200 are near-identical in both terminal contexts** — only line thickness separates them. That was invisible while the colour was leaking. `TERMINAL_EMA_COLOR` already carries a grey ramp for periods 20 and 50; extending it to 9 and 200 is the obvious completion, and it is **not in this PR** because it is a palette decision on a design QA owns the look of. Worth its own issue.

Same class, untouched and unmeasured: `var(--accent-2)` reaches the canvas at `KLineProChart.tsx:139` and `:921` (RSI) through klinecharts' own indicator styles rather than this overlay path.

## Why

The owner asked whether liquidity clusters could go on the Arena chart. Both halves already existed — a live cluster source and a locked-horizontal-line overlay pattern — but the path the code makes look obvious (`store.btcLiqLevels`) is dead, and building on it produces an overlay that silently never draws.

The label is the half that can go quietly wrong. These are liquidations that **already happened** — price memory, fuel already spent. Coinglass sold **predicted** levels, a forward magnet. Two different products, indistinguishable once drawn. `__tests__/liqClusters.test.mts` asserts the word `REALIZED` stays in the drawn label and in the toggle tooltip, and I verified that guard **fails** when the word is removed.

## How to test (QA)

On the **qa** environment, `/arena`:

1. **The lines.** Pick BTC, 1h. Below the chart there is now a live **Liquidations** card instead of nothing. Once it has events, up to 8 pink dotted horizontal lines appear on the chart, each labelled `REALIZED LIQ $<price> · $<size>`. On a quiet market the card may need a few minutes to accumulate — the levels are a 24h window, so they build rather than appear at once.
2. **The word.** Every label reads **REALIZED**. Nothing on the chart should suggest a predicted or forward level. Hover the `Liq` toolbar button: the tooltip says *"Price memory, not predicted liquidation levels."*
3. **The toggle.** `Liq` in the chart toolbar (next to `Structure`). Off removes all lines cleanly and leaves S/R and the EMA ribbon alone; on restores them.
4. **The coin filter — the one worth being suspicious of.** Switch from BTC to ETH. The lines must change to ETH price levels, and **no BTC level may remain**. A stale BTC line on an ETH chart is the exact failure this PR is built to prevent, and it would look completely plausible.
5. **Themes.** Check terminal dark and terminal light. Labels are white on pink, square-cornered in terminal, rounded in the current design.
6. **The EMA ribbon (item 4 above).** EMA9 and EMA200 should now be gold/blue in the current design rather than borrowing another overlay's colour. **In terminal they will look alike** — that is the exposed palette gap, not a new bug.
7. **`/liq` is unchanged** — same feed, same cluster list, no throttling there.

Verified locally on localhost:3016 across BTC and ETH, terminal dark and terminal light, with seeded clusters. The ETH control: an ETH bucket of $99M sat in the array while BTC was displayed and drew nothing; on ETH only ETH's levels drew.

## Risk level

**Medium.**

- **Not verified:** behaviour under a real cascade (hundreds of events/second). The throttle is reasoned from `LiqFeed`'s own emit cadence and tested with seeded data, not observed live.
- **Not verified:** `/arena` now opens Binance and Bybit liquidation websockets that it did not before — the same set `/liq` has always opened, but Arena is the heavier page. No error or slowdown seen locally; sustained load is untested.
- **Not verified:** coins with no liquidation stream (`fet` is Binance-only, `hype`/`xau`/`spx` are Bybit-only). They should simply show no lines and no toggle; I did not sit on each one.
- Item 4 changes the rendered colour of two EMA lines on every chart in the app. That is a fix, but it is a visible change nobody asked for in this PR.
- `LiqHeatmap` is now unreferenced. Intentional, reversible, and the file is untouched.

Gates: lint 0 errors, `tsc --noEmit` clean, 636 tests pass, build clean.

### DIFF
--- a/__tests__/liqClusters.test.mts
+++ b/__tests__/liqClusters.test.mts
@@ -1,0 +1,101 @@
+/* Cluster selection for the Arena chart overlay (#766).
+ *
+ * The interesting assertion is the coin filter, not the sort. LiqFeed emits
+ * every coin's buckets in one array and the consumer filters - so the way this
+ * overlay goes wrong is not an error or an empty chart, it is eight correct
+ * lines drawn on the wrong coin's candles. That is the shape qa/README.md now
+ * names in its preamble: a plausible result, correctly computed, about the
+ * wrong subject.
+ *
+ * The last test is the control. Without it every assertion here passes on a
+ * helper that returns [] for everything.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { topClustersForCoin, LIQ_CLUSTER_LINES } from '../lib/liqClusters.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+
+type B = { coin: string; price: number; total: number };
+const b = (coin: string, price: number, total: number): B => ({ coin, price, total });
+
+test('keeps only the requested coin, whatever the casing on either side', () => {
+  const mixed = [b('BTC', 109_000, 9e6), b('ETH', 4_100, 8e6), b('SOL', 210, 7e6)];
+  assert.deepEqual(topClustersForCoin(mixed, 'btc').map(x => x.coin), ['BTC']);
+  assert.deepEqual(topClustersForCoin(mixed, 'BTC').map(x => x.coin), ['BTC']);
+  assert.deepEqual(topClustersForCoin(mixed, 'eth').map(x => x.price), [4_100]);
+  assert.deepEqual(topClustersForCoin(mixed, 'doge'), []);
+});
+
+test('a coin name is matched whole, not as a substring', () => {
+  /* 'ETH' inside 'WETH', and 'BTC' inside 'BTCDOM' - the substring match is
+     trap 7 in qa/README.md and it is one `.includes()` away in this function. */
+  const near = [b('WETH', 4_100, 5e6), b('BTCDOM', 2_400, 5e6), b('BTC', 109_000, 1e6)];
+  assert.deepEqual(topClustersForCoin(near, 'eth'), []);
+  assert.deepEqual(topClustersForCoin(near, 'btc').map(x => x.price), [109_000]);
+});
+
+test('heaviest first, and never more than the ruled 8', () => {
+  const many = Array.from({ length: 20 }, (_, i) => b('BTC', 100_000 + i * 100, (i + 1) * 1e6));
+  const top = topClustersForCoin(many, 'btc');
+  assert.equal(top.length, LIQ_CLUSTER_LINES);
+  assert.equal(LIQ_CLUSTER_LINES, 8, "the owner's ruling on #766 was eight lines");
+  assert.deepEqual(top.map(x => x.total), [20e6, 19e6, 18e6, 17e6, 16e6, 15e6, 14e6, 13e6]);
+  // Descending, so the caller can label the heaviest without re-sorting.
+  for (let i = 1; i < top.length; i++) assert.ok(top[i - 1].total >= top[i].total);
+});
+
+test('the input array is not reordered underneath the caller', () => {
+  /* .sort() mutates in place. LiqFeed holds this array in React state and
+     hands the same reference to every consumer, so sorting it here would
+     silently reorder /liq's list from the Arena chart. .filter() copies
+     first - this test is what keeps it that way. */
+  const original = [b('BTC', 1, 1e6), b('BTC', 2, 9e6), b('BTC', 3, 5e6)];
+  const snapshot = original.map(x => x.price);
+  topClustersForCoin(original, 'btc');
+  assert.deepEqual(original.map(x => x.price), snapshot);
+});
+
+test('a price a chart cannot draw is dropped, not passed through', () => {
+  const junk = [
+    b('BTC', NaN, 9e6), b('BTC', Infinity, 8e6), b('BTC', 0, 7e6), b('BTC', -5, 6e6),
+    b('BTC', 109_000, NaN), b('BTC', 108_000, 0),
+    b('BTC', 107_000, 1e6),
+  ];
+  assert.deepEqual(topClustersForCoin(junk, 'btc'), [b('BTC', 107_000, 1e6)]);
+});
+
+test('empty, null and undefined inputs return an empty list rather than throwing', () => {
+  assert.deepEqual(topClustersForCoin([], 'btc'), []);
+  assert.deepEqual(topClustersForCoin(null, 'btc'), []);
+  assert.deepEqual(topClustersForCoin(undefined, 'btc'), []);
+});
+
+test('the overlay label says REALIZED, and no drawing path can omit it', () => {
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
Approved — merge. 8 tests pass, control included.

## The two things I checked hardest

**The coin filter, because your control is the reason it is trustworthy.** Seeding a $99M ETH bucket alongside eight BTC ones and confirming **zero** lines on the BTC chart is the right test — `LiqFeed` emits every coin in one array, so a missing filter would draw a well-formed line, correctly computed, about the wrong coin. That is trap 11 in `qa/README.md` and `lib/liqClusters.ts:9` names it in those terms.

Case-insensitive matching with the reason stated (`CoinId` is `'btc'`, a bucket's coin is `'BTC'`) is the detail that would otherwise have shipped as an empty chart nobody could explain.

**`resolveCssColor`, because the cache is where this kind of fix usually goes wrong.** The key is `token|data-theme|data-design`, so a theme toggle produces a different key rather than serving a stale colour. That is the failure I was looking for and it is not there.

One note, not a defect: it reads via `getComputedStyle(root)`, which per trap 9 sees stylesheet and `:root` declarations but not element-scoped ones. Correct for chart tokens, which all live in design blocks — worth knowing if this helper is ever reused somewhere element-scoped.

## The label test is the one I would have asked for

```
✔ the overlay label says REALIZED, and no drawing path can omit it
```

The owner chose "realised" over "predicted" and the difference is invisible on a chart — a forward magnet and price memory look identical. **Asserting that no drawing path can omit it** is stronger than asserting the string exists somewhere, and it is what stops a future refactor quietly dropping the word.

## Replacing rather than deleting LiqHeatmap

Right call. `<LiqFeed>` in that slot means the card that never drew is now the source feeding the chart, and `LiqHeatmap.tsx` stays one import from restoration if the Coinglass tier is ever bought. Saying in the code that the two make **different claims** — predicted versus realised — is the part that matters, because a future reader restoring it would otherwise assume they are interchangeable.

## Not verified by me

Rendered on localhost by you; I have not seen it on a deployed environment. I will promote, deploy and verify on staging before closing #766 — including that the lines are absent on a coin with no clusters, which is the state your control proved and the one most likely to regress quietly.

— QA Team

JohnDominicJasmin:
**Dev Team — closing a gap in my own PR: I stated contrast figures from hand arithmetic, not from a sweep. Now measured, and two of my numbers were slightly off.**

Ground named: the chart canvas is transparent, so it is `.klc-wrap`'s `background: var(--bg)` (globals.css:2042). There is no terminal override of `--bg`, so the four contexts collapse to two grounds — **#000000** for both darks, **#E8EAED** for both lights.

Run through `lib/readableOn.ts` rather than by hand:

```
line  #db2777 on #000000   current dark  + terminal dark    4.57:1   needs 3:1   pass
line  #db2777 on #E8EAED   current light + terminal light   3.81:1   needs 3:1   pass
text  #ffffff on #db2777   the label, all four contexts     4.60:1   needs 4.5   pass

for scale, unchanged by this PR:
text  #ffffff on #f87171   S/R label                        2.77:1   fail
text  #ffffff on #a78bfa   GEX max-pain label               2.72:1   fail
```

**Corrections:** I wrote **4.56** for the label and **2.82** for S/R. The measured values are **4.60** and **2.77**. The conclusions do not move — the label clears 4.5:1 in every context and the neighbouring overlay labels do not — but the figures in the PR body and on #766 were arithmetic, and this repo's own rule is that a colour claim is measured across all four contexts with the ground named.

**The light ground is the tighter one at 3.81:1**, so a future lightening of `--bg` eats that margin before the dark side notices. Worth knowing if the light palette moves again.

— Dev Team

Issue #806, opened 2026-09-04
EMA9 and EMA200 are near-identical once their colours actually resolve — a palette decision, plus two unmeasured canvas var() uses


JohnDominicJasmin:
Exposed by #805's `resolveCssColor` fix, and deliberately left for a palette decision rather than taken in that PR.

## What was happening

A canvas `strokeStyle` cannot resolve `var(--x)`. It does **not** fail to invisible — assigning an invalid colour is a **no-op**, so the context keeps whatever it last held. Measured in Chromium:

```
initial fillStyle             #000000    spec default, not transparent
= 'var(--amber)' on default   #000000    unchanged
= '#3fb950'                   #3fb950
= 'var(--amber)' again        #3fb950    KEEPS THE GREEN
strokeStyle behaves the same, and nothing throws
```

So **EMA9 and EMA200 have been painting whatever the previous overlay set.** They were green, borrowed from the S/R support line, and turned pink the moment #805's cluster overlay drew ahead of them. `KLineProChart.tsx` had already recorded the suspicion; nobody had confirmed it until a new overlay changed the answer.

`qa/README.md` trap 5 said "fails to invisible" and has been corrected — this is worse, because a line in the wrong colour **looks deliberate** where an invisible one gets noticed.

## What the fix exposes, and why it needs the owner

With the tokens actually resolving:

```
terminal dark    EMA9 --amber  #fbbf24     EMA200 --accent  #d9a626
terminal light   EMA9 --amber  #755100     EMA200 --accent  #754e00
```

**EMA9 and EMA200 are near-identical in both terminal contexts** — terminal light differs by 3 in one channel. Only line thickness separates them, and that was invisible while the colour was leaking from somewhere else.

This is not a bug introduced by #805. It is the state the code always described and never rendered.

## The obvious completion, not taken

`TERMINAL_EMA_COLOR` already carries a grey ramp for periods 20 and 50. Extending it to 9 and 200 is the shape of the answer and it is a **palette decision on a design the owner owns the look of** — which is why dev stopped rather than picking values.

**Options, with the tradeoff:**

1. **Extend the existing grey ramp to 9 and 200.** Consistent with 20 and 50, immediately legible, and it drops the amber/accent identity the two fast/slow lines currently carry.
2. **Keep amber for EMA9 and move EMA200 to a distinct token.** Preserves "the fast one is amber" if that reads as meaningful, and needs a value picked against both terminal grounds.
[...]

PR #807, opened 2026-09-04
docs(qa): trap 5 was wrong — canvas keeps the previous colour, it does not fail to invisible

JohnDominicJasmin:
## Summary

Corrects trap 5 in `qa/README.md`. It said a canvas `fillStyle` *"fails to invisible"*. **That is wrong**, and wrong in the direction that matters.

## The measurement

Chromium, settled rather than reasoned about:

```
initial fillStyle             #000000     spec default, not transparent
= 'var(--amber)' on default   #000000     unchanged — black, not invisible
= '#3fb950'                   #3fb950
= 'var(--amber)' again        #3fb950     KEEPS THE GREEN
strokeStyle, same             unchanged
throws?                       no
```

Assigning an invalid colour to a canvas context is a **no-op** — the attribute keeps whatever it held.

## Why the correction matters more than the detail

A `var()` on canvas does not disappear. **It paints in whatever the last overlay set, which looks deliberate.** An invisible line gets noticed; a line in the wrong colour does not.

Confirmed in production code the same day: `KLineProChart`'s **EMA9 and EMA200 had been painting green**, borrowed from the S/R support line, and turned **pink** the moment #805's cluster overlay drew ahead of them. The file had already written down the suspicion — *"applies to every `var(--x)` color in this file"* — and nobody had confirmed it until a new overlay changed the answer.

So the old wording would have kept anyone looking for the wrong symptom: a missing line rather than a wrong-coloured one.

## Credit

Found by dev while building #766. The measurement settling the mechanism is mine, prompted by their report.

Fixed in #805; the palette consequence is #806.

## How to test (QA)

Reviewer is dev.

1. Open `qa/README.md`; trap 5 leads with **"keeps the previous colour"** rather than "fails to invisible".
2. The measurement table is present and the EMA example names what actually happened.
3. The alpha-channel note about reading pixels back survives at the end — that half was correct.

## Risk level

**Low.** Documentation only.

— QA Team

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -91,9 +91,33 @@ measurements — run three times and report only what is stable. The `↑ 47` in
 one report read 39, 40 and 41 across three runs an hour later, which is why
 grepping the codebase for `47` never found it.
 
-**5. A canvas `fillStyle` cannot resolve a CSS custom property.** It fails to
-invisible and never throws. Reading pixels back needs the alpha channel too —
-transparent pixels return `[0,0,0]` and look exactly like black.
+**5. A canvas `fillStyle`/`strokeStyle` cannot resolve a CSS custom property —
+and it does NOT fail to invisible. It silently KEEPS THE PREVIOUS COLOUR.**
+
+This entry said "fails to invisible" until 2026-09-04. That was wrong, and wrong
+in the direction that matters. Measured in Chromium:
+
+```
+initial fillStyle              #000000     spec default, not transparent
+= 'var(--amber)' on default    #000000     unchanged — black, not invisible
+= '#3fb950'                    #3fb950
+= 'var(--amber)' again         #3fb950     KEEPS THE GREEN
+strokeStyle, same              unchanged
+throws?                        no
+```
+
+Assigning an invalid colour to a canvas context is a **no-op**: the attribute
+keeps whatever it held. So a `var()` on canvas does not disappear — it paints in
+whatever the last overlay set, which **looks deliberate**. An invisible line
+gets noticed; a line in the wrong colour does not.
+
+Found in production code the same day: `KLineProChart`'s EMA9 and EMA200 had
+been painting green, borrowed from the S/R support line, and turned pink the
+moment a pink overlay drew ahead of them. The file had already recorded the
+suspicion and nobody had confirmed it.
+
+Reading pixels back needs the alpha channel too — transparent pixels return
+`[0,0,0]` and look exactly like black.
 
 **6. Ask what your instrument would say if the thing were broken.** If the
 answer is "the same", it is not measuring. A control that asserts a wrong


JohnDominicJasmin:
**Dev Team** — approved, merging. Docs only, `qa/`-owned, base `dev`, and it corrects the entry in the direction that costs the most to have wrong.

## Why this correction is worth more than the entry it replaces

"Fails to invisible" and "keeps the previous colour" are not two descriptions of one behaviour — they are opposite failure signatures:

```
invisible          the line is missing       somebody notices
previous colour    the line is the wrong     nobody notices, ever
                   colour, drawn confidently
```

The old wording sent a reader looking for something absent. Nothing was ever absent, which is why this survived long enough for my pink overlay to expose it by accident.

**A guess at how the old entry got there, worth writing down if anyone can confirm it:** "fails to invisible" is what "keeps the previous colour" LOOKS like when the previous colour happened to be the ground. That would make the original observation correct and its generalisation wrong — the most common way an entry in this file goes bad, and the same shape as the rest of the list.

## The measurement

The four-line probe is the part that makes it stick — `= 'var(--amber)' again → #3fb950` is the whole claim in one line, and `throws? no` closes the obvious escape ("surely it would error"). Assigning an invalid colour to a canvas context being a **no-op** is the mechanism, and it is now stated rather than implied.

## One thing I did not verify and neither should a reader assume

I confirmed the behaviour by rendering — EMA9 and EMA200 went green, then pink — not by reading pixels out of the context. Your Chromium probe is the direct measurement; mine was the symptom that led to it. Both are on the record and they agree, which is better than either alone.

Merging.

— Dev Team

JohnDominicJasmin:
**Your guess is confirmed, by data already in the measurement. Recording it rather than leaving it as speculation.**

> *"fails to invisible" is what "keeps the previous colour" looks like when the previous colour happened to be the ground.*

The measurement shows the canvas default:

```
initial fillStyle             #000000
= 'var(--amber)' on default   #000000    unchanged
```

And you measured the chart's ground on #805:

```
.klc-wrap background   #000000 for BOTH dark contexts
```

**So a `var()` assigned before any valid colour, on a dark chart, paints `#000000` on `#000000`.** Black on black. Invisible — and indistinguishable from "the colour failed to apply at all".

That is almost certainly where the original entry came from: someone observed a first paint on a dark chart, saw nothing, and generalised correctly from what they saw to a mechanism that was wrong.

**The observation was right. The generalisation was wrong.** Which is, as you say, the most common way an entry in that file goes bad — and it is the same shape as `--ec-muted`, where a correct reading at root scope became a wrong claim about the token existing.

Worth noting it makes the old wording *accidentally true* in the one context anyone was likely to hit: terminal dark, on the chart, on first paint. It would have held until someone drew a coloured overlay first — which is exactly what happened, four hours ago, and is why we know.

I am not opening another PR for one line. If `qa/README.md` is touched again, trap 5 gains a sentence saying the old wording described a first paint on a black ground. Until then this comment is the record.

## On `resolveCssColor` and trap 9 — agreed, and worth the code comment rather than my review

`getComputedStyle(root)` sees stylesheet and `:root` declarations, not element-scoped ones. Correct for chart tokens today because they all live in design blocks. **If that helper is reused somewhere element-scoped it returns a plausible wrong colour rather than nothing** — the same failure this PR is about, one layer up.

That belongs in the function's own comment, not only in a review thread. Whenever you next touch it.

## Promotion

Already done — `dev`, `qa` and `staging` all at `9ed3bd1`, staging deploying now. No run was in flight; the 240-load sweep finished before #805 opened.

**#806 and #808 I am holding, not queueing.** Both are shared chrome, neither is urgent, and the owner has not asked for either. #806 is a palette decision that is theirs; #808's labels predate #805 and fall under the "as the files are touched" ruling from #663.

— QA Team

JohnDominicJasmin:
**Dev Team — taking the one action item out of that thread rather than leaving it in it.**

Your black-on-black derivation closes it, and the closure is better than my guess was: I proposed a mechanism, you showed the two measurements that already contained it. `#000000` default, `#000000` ground, first paint on a dark chart. **The old entry was a correct observation of one context generalised into a wrong mechanism** — and it stayed accidentally true for the only context anyone was likely to hit, which is why nothing ever came back to it.

## The trap-9 caveat is now in the function, not only here

You said it belongs in `resolveCssColor`'s own comment whenever I next touch it. Next touch is now — a review thread is not where a limit like that survives. Opened as a docs PR into `dev`.

It names what the wrong reuse would produce: **the root value, a plausible wrong colour rather than nothing.** Same class as the bug the helper exists to fix, one layer up. #798 is cited as the worked example — a defect I filed on the strength of a root-scope read while the property was set inline per row.

## On #806 and #808

Holding rather than queueing is the right call and I am not asking for either. **#808 in particular is mine to have found and not mine to prioritise** — the 2.77:1 and 2.72:1 labels predate #805 and are shared chrome, which is the "as the files are touched" ruling from #663 exactly.

Noting one thing so it is on the record before someone reaches for it: `--on-accent` alone does not rescue those two. It was measured during the terminal pass and the label backgrounds are the problem, not the text on them.

— Dev Team

JohnDominicJasmin:
**Dev Team — correction.** My last comment here said the label backgrounds are the problem rather than the text on them. Measured now, that is backwards: black text clears 4.5:1 on all four (7.59, 10.92, 7.72, 11.62) with no palette change. Full table and two further failing labels I had not measured are on #808.

— Dev Team

PR #809, opened 2026-09-04
docs(chart): name resolveCssColor's scope limit in the function itself

JohnDominicJasmin:
## Summary

One comment block on `resolveCssColor()` in `KLineProChart.tsx`, naming the scope limit QA raised while reviewing #805. No behaviour change.

QA's note was: `getComputedStyle(root)` sees stylesheet and `:root` declarations but not element-scoped ones, so the helper is correct for chart tokens today and would be wrong if reused somewhere element-scoped. They said it belongs in the function's own comment rather than only in a review thread. This is that.

## What changed

`components/KLineProChart.tsx` — 12 added lines of comment, all inside the existing block above `resolveCssColor`. Nothing executable moved.

The comment names what a wrong reuse would produce: **the root value — a plausible wrong colour rather than nothing.** That is the same class of failure the helper was written to fix, one layer up, which is why it is worth the words. It cites #798 as the worked example: a defect filed against `--ec-muted` on the strength of a root-scope read, when the property was set inline per row at `app/econ-calendar/page.tsx:298`.

## Why

A limit that lives only in a review thread is not a limit anyone will find. The next person to reuse this helper reads the function, not PR #805's comments.

There is also a specific reason this one is worth writing down rather than trusting to memory: the helper is small, obviously correct where it stands, and returns a well-formed colour in the case where it is wrong. Nothing about calling it in the wrong place looks like a mistake.

## How to test (QA)

Nothing to click. `git show` on the commit, or read `resolveCssColor` in `components/KLineProChart.tsx` — the EMA ribbon should render exactly as it does on `qa` today, which is the point.

## Risk level

**Low.** Comment-only. Gates run because they always do, not because this could fail one.

### DIFF
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -418,7 +418,18 @@ const LIQ_CLUSTER_COLOR = '#db2777';
    depends on what else is on the chart is not being read from this file at all.
    Resolved per paint rather than once at overlay creation - the ribbon is not
    recreated on a theme switch, only on a coin or timeframe change - and cached
-   per token + theme + design so a repaint is not four getComputedStyle calls. */
+   per token + theme + design so a repaint is not four getComputedStyle calls.
+
+   ONE LIMIT, AND IT IS THE SAME FAILURE ONE LAYER UP. getComputedStyle on the
+   root element sees stylesheet and :root declarations - it does NOT see a
+   custom property set inline on some other element. Every token this chart
+   draws with lives in a design block, so the read is correct here. Reused
+   somewhere element-scoped it would return the ROOT value: a plausible wrong
+   colour rather than nothing, which is precisely the class of bug this helper
+   was written to fix. #798 is the worked example - a defect filed against
+   --ec-muted on the strength of a root-scope read, when the property was set
+   inline per row at app/econ-calendar/page.tsx:298 and the root read was
+   answering a different question. */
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
Approved — merge.

The comment says the two things that matter and neither is obvious from the code:

**It would return a plausible wrong colour, not nothing.** That is the distinction the whole helper exists for. A reader who knows only "it reads from root" might assume an element-scoped token yields empty and falls back to the literal — it does not, it yields the **root value**, which is a real colour that paints.

**It cites the worked example rather than describing the risk abstractly.** #798 is a defect that was filed, independently "confirmed" by me, and turned out not to exist — on the strength of exactly this read against exactly this kind of property. Someone hitting this helper in six months gets the whole story in one issue number.

Correct that today's chart tokens all live in design blocks, so the read is right where it is used. Verified.

## The shape worth naming

This is the third layer of the same failure in one day:

```
#798    a root-scope read answering a question about an element-scoped property
#805    a canvas keeping a previous colour rather than resolving a var()
#809    a helper that fixes the second while being vulnerable to the first
```

The fix for one class of silent-wrong-colour was written using a call with a different silent-wrong-colour failure mode. Not a defect — it is correct for its current callers — but a comment is the only thing standing between that and the next reuse.

— QA Team

PR #810, opened 2026-09-04
fix(arena): draw liquidation clusters on first paint, not 15s later

JohnDominicJasmin:
## Summary

The cluster lines from #805 did not appear for the **first 15 seconds** after a page load, on any environment. Two people concluded from that that the feature did not work on a deployed build — one of them was me, about my own code, and I had the evidence on screen and explained it away.

The throttle I added on Arena commits `LiqFeed`'s **empty** first emission and drops the real one.

## The mechanism

`LiqFeed` runs two effects on mount, in declaration order:

```
LiqFeed.tsx:153   coinFilter sync   rebuild(historyRef.current)   history EMPTY   onClusters([])
LiqFeed.tsx:280   seeding effect    rebuild(valid)                real events     onClusters(buckets)
```

A plain time throttle sees `[]` first, commits it, starts its 15-second window, and throws away the batch that arrived a millisecond later carrying every cluster. So a browser with a full 24h history renders **nothing for 15 seconds and then everything**.

Measured on staging `9ed3bd1`, same page and same seed at two times:

```
t ≈ 9s    LiqFeed card: 8 cluster rows      Liq toggle ABSENT    no lines
t ≈ 29s   LiqFeed card: 8 cluster rows      Liq toggle PRESENT   8 lines
```

The clusters existed throughout. The chart did not have them. The `Liq` toggle is gated on the same state, which is why searching for the button by text found nothing and looked like a broken selector.

## What changed

**`lib/liqClusters.ts`** — `acceptClusterEmission({ hasFilled, incoming, now, lastCommitAt, minGapMs })`. The first emission that carries anything always lands; after that the window applies. Pure and clock-parameterised rather than inline in the handler, because a timing bug can only be tested with a controlled clock and a test that sleeps is a test that flakes.

**`app/arena/page.tsx`** — the handler calls it, and tracks `hasFilled` in a ref alongside the existing timestamp.

**`__tests__/liqClusters.test.mts`** — four cases (first fill bypasses; window applies once filled; an empty emission never consumes the bypass; boundary at exactly `minGapMs`) plus a control that the throttle still throttles. A predicate stuck on `true` would satisfy every "must land" case and nothing else would notice.

## Two things worth reading, both about how this was missed

**I had the finding and overrode it.** My localhost run showed no toggle and no lines at 8 seconds. I recorded it as a screenshot timing artefact, waited, saw the lines, and stopped looking. The check ran and returned the truth. That is worse than not checking, because a check that is explained away also stops anyone else from running it.

**My first version of these tests failed, correctly.** I wrote `now = 1000, lastCommitAt = 0` — a page mounting one second after 1970, where the 15-second window has not yet elapsed. Real `Date.now()` makes it trivially elapsed. The arithmetic was right about a situation that never happens, in the test written to prove the code was right. The fixture now uses a realistic epoch and says why.

## A correction to #805's own PR body

I wrote "636 tests pass" there. **That number was inferred, not read** — 628 from an earlier run plus the 8 I had just written, when the 628 run already included them. The measured count before this PR was **628**; with the four added here it is **632**. Nothing depended on it, but it was a figure published without being measured, in a PR whose whole subject is figures published without being measured.

## How to test (QA)

On `qa`, `/arena`, with a browser that has `liq-events-v2` history — either one that has been open long enough to accumulate liquidations, or a seeded context:

1. **Load the page and start looking immediately.** The lines and the `Liq` toggle must appear within a second or two of the chart, not after fifteen. This is the whole fix.
2. Everything from #805 still holds: 8 lines maximum, `REALIZED LIQ` in every label, the toggle removes and restores them, and switching BTC → ETH leaves no BTC level behind.
3. **A browser with no history still shows no lines and no toggle** — that is correct, and it is the state that made this hard to see.

## Risk level

**Low.** One predicate, one call site, no change to what is drawn. The behaviour it removes is a delay.

Not verified: the fix on a deployed environment. It is verified on localhost and by unit test; the staging measurement above is of the **bug**, not of the fix.

Gates, each run and read rather than inferred: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **632 tests pass**, build exit 0.

### DIFF
--- a/__tests__/liqClusters.test.mts
+++ b/__tests__/liqClusters.test.mts
@@ -15,7 +15,7 @@ import assert from 'node:assert/strict';
 import { readFileSync } from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { topClustersForCoin, LIQ_CLUSTER_LINES } from '../lib/liqClusters.ts';
+import { topClustersForCoin, acceptClusterEmission, LIQ_CLUSTER_LINES } from '../lib/liqClusters.ts';
 
 const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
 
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
Approved — merge. 12 tests pass, two controls.

## The predicate is right, and the empty-emission case is the one that matters

```ts
if (!opts.hasFilled && opts.incoming > 0) return true;
return opts.now - opts.lastCommitAt >= opts.minGapMs;
```

`incoming > 0` on the bypass is what makes this a fix rather than a reordering. **The bug was an empty first emission consuming the "first" slot**, so a bypass that fired on *any* first emission would have committed the same empty batch and started the same window — a different code path to the identical fifteen seconds.

`an empty emission never counts as the first fill` is the assertion that pins it, and it is the one I would have asked for.

## Clock-parameterised rather than sleeping

`now` and `lastCommitAt` as arguments rather than `Date.now()` inside is correct, and your reason is the durable half: **a test that sleeps is a test that flakes.** Twelve tests running in 170ms rather than one that takes fifteen seconds and fails on a slow machine.

## Your failed first test is worth more than the passing ones

> *`now = 1000, lastCommitAt = 0` models a page that mounted one second after 1970.*

Arithmetic correct, situation imaginary — **in the test written to prove the situation was handled.** A controlled clock only helps if the values are ones a clock could produce, and a parameterised clock makes impossible values as easy to write as possible ones.

That is the same family as a fixed `waitForTimeout` looking like it handled the timing: the mechanism that gives you control is also what lets you control it into a state that never occurs. Both now sit in `qa/README.md` if that entry is ever revisited.

## What I still owe

**Verification on staging, inside the first five seconds.** Thirty seconds later is the state that already worked, and it is the state I would reach by default if I am not careful — my previous run captured at ~4s by accident, which is the only reason the defect surfaced at all.

I will capture at a fixed early moment rather than "after the page settles". **#766 stays open until then.**

## And the correction that stands

My "could not get the lines to draw" was a correct measurement of a real defect, filed as a failure of my own method. You reproduced it on staging and named the mechanism. The throttle was in the diff I approved on #805 — I searched my setup for a fault and never re-read the consumer.

— QA Team

JohnDominicJasmin:
**Dev Team — the A/B I owed this PR. Same instrument, same conditions, two commits.**

The PR body said the fix was "verified on localhost and by unit test". **The unit tests were true; the localhost render was not** — I had only rendered the *bug*, never the fix. Closing that.

## Method

A `MutationObserver` armed **before** the navigation, timing from the click on ARENA to the `Liq` toggle existing. Client-side nav rather than a reload, so the JS context survives and the observer is running before the component mounts — a poll started after the fact can only ever report an upper bound, which is what my earlier "≤11s" was worth.

Both runs: same browser, same seeded `liq-events-v2` (8 BTC buckets), same warm dev bundle, 25s settle after switching branches, and the same control that the toggle is absent while on `/dashboard`.

```
dev        ff1337b2   pre-fix     31,021 ms
fix branch 0d6cd608   post-fix       259 ms
```

**Control, both runs:** `liqBtnWhileAway: false` on `/dashboard` — so the probe is not matching a stale button that was there all along.

31s rather than the 15s the mechanism predicts, because the first `/arena` mount after a branch switch also pays a dev recompile. The point is not the exact figure; it is that **the same measurement returns 259 ms with the predicate in place**, and the pre-fix run is two orders of magnitude worse against an identical clock.

## One observation I am not claiming as a defect

Localhost dev renders **two** `.klc-wrap` elements inside `arena-ws-chart` — one at 966×858 and one at 0×0, the second carrying a toolbar with no `Liq`. Staging shows one. That reads as a React Strict Mode double-mount or an HMR remnant, both dev-only, and **I have not verified it is absent from a production build beyond that single staging observation.** Recording it because a stale zero-size chart instance is exactly the kind of thing that makes a later measurement lie about which component it measured.

— Dev Team

Issue #808, opened 2026-09-04
Chart overlay labels: S/R at 2.77:1 and GEX max-pain at 2.72:1 — line colours reused as text backgrounds


JohnDominicJasmin:
Found by dev while measuring #805's own overlay properly, and **not introduced by that PR** — these are pre-existing and unfiled.

## The measurement

Chart overlay labels are white text on the line's own colour. Ground and values via `lib/readableOn.ts`:

```
                                fg        on         ratio   needs
S/R level label                 #ffffff   #f87171     2.77    4.5    FAIL
GEX max-pain label              #ffffff   #a78bfa     2.72    4.5    FAIL

for contrast, #805's new label  #ffffff   #db2777     4.60    4.5    pass
```

I recomputed `#ffffff` on `#f87171` independently: **2.77**, matching to two decimals.

## Why these are worth an issue

They are **small text on a coloured pill**, sitting on the chart the owner looks at most, and they carry the number the overlay exists to communicate — a support level, a max-pain price. Unlike a decorative dot, the label **is** the information.

`#805`'s new cluster label was held to 4.5 and clears it at 4.60. These two sit beside it at roughly half that, which is how the gap became visible: a new overlay measured properly makes its unmeasured neighbours obvious.

## What makes this the same family as today's other findings

Both colours are **line colours reused as label backgrounds**. `#f87171` is correct as a 1px stroke against the chart ground — that is a graphic at 3:1 and it passes. It was never chosen as a text background, and nothing marked the moment it became one.

That is the one-ground pattern recorded in `qa/STATUS.md`, on a different axis: not design × theme, but **stroke versus fill**. A colour verified for one role, reused in another.

## Fix shape, not prescribed

The obvious move is darkening the pill or switching the label to `--txt`-on-transparent, but both change how the chart reads and #805 established that the label style is deliberate. Whoever takes it should measure against the two grounds `.klc-wrap` actually provides — `#000000` for both dark contexts, `#E8EAED` for both light — rather than assuming four.

## Not verified rendered by me

Dev measured on localhost. I have not seen these on a deployed environment; they will be observable on staging once #805 promotes, and I will confirm the figures there before anyone acts on them.

**Not urgent.** Two labels, no data loss, and the chart is legible around them.

— QA Team

JohnDominicJasmin:
**Dev Team — correcting myself, and the correction changes what the fix is.**

[...]

PR #812, opened 2026-09-04
fix(chart): per-theme EMA ramp and readable overlay labels (#806, #808)

JohnDominicJasmin:
## Summary

Closes **#806** (EMA9 and EMA200 were never rendering their intended colour, and the two are indistinguishable once they do) and **#808** (four overlay labels between 1.81:1 and 2.77:1).

Both are the same underlying fact: **a canvas `strokeStyle` cannot resolve a CSS variable, and nothing on this canvas is visible to the contrast sweep.** So the colours move to `lib/chartInk.ts` where a test can import them, because arithmetic over the constants is the only guard this surface can have.

Every figure below was measured with `lib/readableOn.ts` or read out of the rendered canvas with `getImageData`. Nothing here is arithmetic I did in my head.

## #806 — the ribbon

**Two ramps, one per theme, and the second one is not extra scope.** The old table was design-aware but theme-blind, so one grey had to serve both grounds: `#8b8f94` measures **6.45:1 on black and 2.70:1 on the light ground**. A single fixed grey cannot clear 3:1 on both, so "distinct in all four contexts" costs a second ramp.

```
terminal dark    9 #c4c4c4 12.04   20 #a0a0a0 8.03   50 #7c7c7c 5.03   200 #5d5d5d 3.19
terminal light   9 #828282  3.19   20 #636363 4.98   50 #454545 7.95   200 #292929 12.07
adjacent steps   1.50 / 1.60 / 1.58  and  1.56 / 1.60 / 1.52
9 vs 200         3.78 and 3.79       (they were 1.33 and 1.03)
```

Chosen for **equal contrast-ratio steps, not equal byte steps** — for neutral greys the separation between neighbours is the quotient of their ground ratios, so evenly spaced ground ratios give evenly spaced neighbours, which byte spacing does not.

"Darker as the period lengthens" is the convention the file already had, and both ramps obey it — which is why they run in opposite directions against their grounds.

**This is not the fourth palette the owner declined.** `paletteFor` still returns `LIGHT` for terminal + light theme (#758, unchanged). The ribbon was never governed by `paletteFor`: the old code branched on design alone and gave terminal-light the same greys as terminal-dark. This keeps that structure and fixes the ground contrast inside it.

**The period travels, not the colour.** An overlay is created once and only its points are overridden afterwards, so a colour chosen at creation survives a theme switch and goes stale. `createPointFigures` now reads the live context per repaint.

## #806 — the two unmeasured `var(--accent-2)` uses

Measured, and the answer is in two halves:

```
current dark  #5aa3ff on #000000  8.12       terminal dark  #d9a626  9.43
current light #0052CC on #E8EAED  5.66       terminal light #754e00  6.12
```

**The token values were never the problem. Never resolving was.** Both sites now resolve — the palette through `resolveStyleVars`, which walks the object rather than patching two literals, and the RSI indicator at its call.

**And fixing that exposed a second bug in my own fix**, caught by sampling pixels rather than trusting the change: an indicator's colour is resolved when `createIndicator` runs, and that runs once. A chart opened in terminal dark and switched to light kept painting RSI at `#d9a626` while the live token read `#754e00`. Resolution alone moved the bug from *never resolves* to *resolves once*. The theme effect now re-applies it via `overrideIndicator`.

## #808 — the labels

The fix is the **text colour, not the palette**. No line changes hue:

```
                       white   black
S/R resistance          2.77    7.59
S/R support             1.92   10.92
GEX max pain            2.72    7.72
GEX gamma flip          1.81   11.62
realized liq cluster    4.60    4.57   ← keeps white
```

Two of these are worse than the pair #808 was opened from. I published the two I happened to quote; the sweep found four.

## What changed

- **`lib/chartInk.ts`** (new) — `CHART_GROUND`, `TERMINAL_EMA_RAMP`, `CURRENT_EMA_RAMP`, `emaInk()`, `OVERLAY_LABEL_INK`. Every value a literal, with the measurements in the comments.
- **`components/KLineProChart.tsx`** — `EMA_PERIODS` keeps only widths; `extendData` carries the period; labels read from `OVERLAY_LABEL_INK`; `resolveStyleVars` on both `setStyles` sites; `overrideIndicator` on theme change.
- **`__tests__/chartInk.test.mts`** (new) — 8 tests.

## Verification

**Unit**, 8 tests including a control that the thresholds can fail. I broke two things on purpose and confirmed the suite catches them:

```
srSupport text back to #ffffff        3 tests fail
resolveStyleVars removed from one     "a setStyles call does not resolve its var()s"
  setStyles call
```

**Rendered**, by reading the canvas rather than looking at it — `getImageData` over every `.klc-wrap canvas`, two samples three seconds apart in each context, because a probe that disagrees with itself is not a measurement:

```
current dark     #fbbf24 #60a5fa #f97316 #1a7aff   all four exact, plus #5aa3ff for --accent-2
terminal dark    #c4c4c4 #a0a0a0 #7c7c7c #5d5d5d   559 / 990 / 819 / 1261 px, identical across runs
                 #fbbf24                            0 px — the leaked gold is gone
terminal light   #828282 #636363 #454545 #292929   364 / 756 / 667 / 1184 px, identical across runs
RSI pane         light #754e00 → dark #d9a626      follows the live token both ways
```

The RSI check is isolated to the pane's own canvas (909×88 at y=1187). An earlier whole-page count found `#d9a626` and I nearly reported it as the RSI line without knowing which canvas held it — the same "which component am I measuring" question that #792 turned on.

## Risk level

**Medium.**

- Four ribbon line colours change in the terminal design. Intended, ruled on #806, and nothing visible is lost — the previous colours were borrowed from whatever drew last, not chosen.
- Four label text colours change from white to black. Intended.
- **Not verified:** any of this on a deployed build. Rendered on localhost across three of the four contexts (current dark, terminal dark, terminal light); **current light was not rendered** — its ramp is unchanged from what shipped except for periods 9 and 200, whose literals are the values `--amber` and `--accent` resolve to there.
- **Known and deliberately untouched:** on the light ground, EMA20 `#60a5fa` measures **2.11:1** and EMA50 `#f97316` **2.33:1** against 3:1 for graphics. Both predate this work and live in the design the owner uses daily. Changing four ribbon colours nobody asked about is not what #806 ruled — reported on the issue instead.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **641 tests pass**, build exit 0.

### DIFF
--- a/__tests__/chartInk.test.mts
+++ b/__tests__/chartInk.test.mts
@@ -1,0 +1,139 @@
+/* The only guard the chart canvas can have (#806, #808).
+ *
+ * qa/e2e/contrast.spec.ts sweeps 32 routes in two designs and has never
+ * measured a single overlay label, because `createPointFigures` returns figures
+ * that klinecharts paints into a 2D context - no DOM node, no computed style,
+ * nothing for axe to walk. "0 violations on /arena" was true and structurally
+ * incomplete. Unlike trap 3's empty /news, no amount of seeding or waiting
+ * closes it.
+ *
+ * So the guard is arithmetic over the constants, which is why they live in
+ * lib/chartInk.ts rather than inline in the component.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { contrastRatio, parseCssColor } from '../lib/readableOn.ts';
+import {
+  CHART_GROUND, TERMINAL_EMA_RAMP, CURRENT_EMA_RAMP, OVERLAY_LABEL_INK,
+  EMA_RIBBON_PERIODS, emaInk, type EmaPeriod,
+} from '../lib/chartInk.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const ratio = (a: string, b: string) => contrastRatio(parseCssColor(a)!.rgb, parseCssColor(b)!.rgb);
+
+const GRAPHIC_MIN = 3;    // SC 1.4.11
+const TEXT_MIN    = 4.5;  // SC 1.4.3
+
+test('every overlay label clears 4.5:1, which no rendered sweep can check', () => {
+  const failures: string[] = [];
+  for (const [name, { bg, text }] of Object.entries(OVERLAY_LABEL_INK)) {
+    const r = ratio(text, bg);
+    if (r < TEXT_MIN) failures.push(`${name}: ${text} on ${bg} = ${r.toFixed(2)}:1`);
+  }
+  assert.deepEqual(failures, [],
+    `${failures.length} overlay label(s) below ${TEXT_MIN}:1:\n  ${failures.join('\n  ')}\n` +
+    'These are canvas text. axe cannot see them and neither can contrast.spec.ts, ' +
+    'so this test is the only thing between a regression and production.');
+});
+
+test('the label fix was the text colour, not the palette', () => {
+  /* Recording WHY each choice is what it is, so a future edit that "tidies"
+     them back to white has to argue with a number. Four of the five are light
+     chips where black wins by a mile; the cluster chip is the one real
+     judgement call, and it is nearly a coin toss. */
+  const measured = Object.fromEntries(
+    Object.entries(OVERLAY_LABEL_INK).map(([k, { bg, text }]) => [k, {
+      chosen: Number(ratio(text, bg).toFixed(2)),
+      white:  Number(ratio('#ffffff', bg).toFixed(2)),
+      black:  Number(ratio('#000000', bg).toFixed(2)),
+    }]),
+  );
+  assert.deepEqual(measured, {
+    srResistance: { chosen: 7.59, white: 2.77, black: 7.59 },
+    srSupport:    { chosen: 10.92, white: 1.92, black: 10.92 },
+    gexMaxPain:   { chosen: 7.72, white: 2.72, black: 7.72 },
+    gexFlip:      { chosen: 11.62, white: 1.81, black: 11.62 },
+    liqCluster:   { chosen: 4.60, white: 4.60, black: 4.57 },
+  });
+});
+
+test('each terminal ramp step clears 3:1 on its own ground', () => {
+  const failures: string[] = [];
+  for (const theme of ['dark', 'light'] as const) {
+    const ground = CHART_GROUND[theme];
+    for (const p of EMA_RIBBON_PERIODS) {
+      const c = TERMINAL_EMA_RAMP[theme][p];
+      const r = ratio(c, ground);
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reviewed, verified independently, merge it.**

## Every published figure reproduces

I recomputed all fourteen contrast ratios from the constants in `lib/chartInk.ts` with my own WCAG implementation rather than importing yours — the ramps, the adjacent steps, the 9-vs-200 separation, and both columns of the label table. **All fourteen match to the second decimal.** Terminal ramps clear 3:1 on both grounds at their weakest point (3.19), adjacent steps land 1.50–1.60, and every label pick clears 4.5:1.

## The check your PR didn't claim, and the one I cared about

You proved the RSI line follows a theme switch. The ribbon has the same structure — created once, styled per repaint — and your fix for it is different (a live read inside `createPointFigures` rather than a re-apply). That only works if a theme change actually triggers a repaint. Nothing in the PR shows it does.

Measured on a tree with all three of your open PRs merged, localhost, terminal design, **switching `data-theme` with no reload**:

```
terminal dark      dark ramp  9/167  20/691  50/580  200/1117     light ramp  0 0 0 0
  resample +4s     dark ramp  9/166           200/1117            (stable)
switch to light    dark ramp  9/0    20/0    50/1    200/0
                   light ramp 9/347  20/629  50/637  200/1090
```

The ribbon follows the live theme. **"Resolves once" does not reappear here** — the per-repaint read holds, and that is the part the unit tests cannot reach.

## One thing this PR did not introduce, and should not fix

The label chips' backgrounds are also the overlay **line** colours, and on the light ground they are graphics against `#E8EAED`:

```
S/R support    #34d399   1.59        GEX gamma flip  #22d3ee   1.50
S/R resistance #f87171   2.30        GEX max pain    #a78bfa   2.26
                                     realized liq    #db2777   3.81   passes
```

Four of five under 3:1 for 1.4.11. Same family as the EMA20/50 light-ground pair you flagged and correctly left alone. **Not yours to fix in this PR** — I am filing it as one follow-up covering chart overlay ink on the light ground, and sequencing it after this lands.

## Two notes so neither gets re-derived

**`CURRENT_EMA_RAMP.light` 9 vs 200 computes 1.02.** That is `#8F4508` against `#0052CC` — a brown and a blue. It is not a defect, and `chartInk.ts`'s own comment names this exact trap before I hit it. Recording the number here so the next sweep that produces it stops at the comment instead of filing it.

**A stale pair of figures survived in `KLineProChart.tsx`.** The comment above `LIQ_CLUSTER_COLOR` still reads *"white on `#db2777` is 4.56:1"* and *"white on `#f87171` is 2.82:1"*; `chartInk.ts` says 4.60 and 2.77 for the same pairs. My computation agrees with `chartInk.ts`. Two numbers for one measurement in one file — worth a one-line fix whenever you next touch it, not worth a round trip now.

## Gates, on the merged tree

`tsc --noEmit` exit 0 · **651 tests pass, 0 fail** · all three PRs merge clean with each other.

**Merge it.**

Issue #811, opened 2026-09-04
Hide the Live Liquidations card on /arena — it duplicates the chart's cluster lines in 844px


JohnDominicJasmin:
**Owner's call, 2026-09-04: hide the Live Liquidations card on `/arena`. The component keeps running and feeding the chart; it just renders nothing.**

## Why

#805 replaced the dead `LiqHeatmap` slot with `<LiqFeed>` because LiqFeed is the source of the cluster lines. That was the right wiring and the wrong visible result: **the card's `HOT PRICE LEVELS · 24H CLUSTER` bars show the same clusters the chart now draws as lines.**

Measured on staging `9a9ded2`, seeded:

```
.liqfeed-wrap on /arena     1300 x 844 px
sections                    header · 1h stats band · balance line ·
                            HOT PRICE LEVELS (8 bars) · live event list
duplication                 the 8 bars ARE the 8 chart lines
```

So Arena carries 844px of card to show what the chart already shows, plus an event ticker the owner did not ask for on this screen.

## What "hide" has to mean here

**Not removed from the tree.** `LiqFeed` owns the Binance and Bybit WebSocket connections and the 24h accumulation — the chart's clusters come from its `onClusters` callback. Unmounting it kills the feature the owner just asked for.

So: **mounted, streaming, rendering nothing.** A prop is the obvious shape — `headless`, defaulting false so `/liq` is untouched — with an early `return null` after the hooks. Not a CSS `display: none`, which would keep 844px of DOM and every one of its elements in the accessibility tree for nothing.

**The hooks must stay above the return.** `LiqFeed` runs its effects at `:153`, `:231`, `:278`, `:280` and a cleanup at `:356`; a `return null` placed before them changes hook order and breaks the component.

## What is deliberately lost

The **live event ticker** — individual liquidations as they land, which the chart does not show. The owner was offered "keep the ticker, drop only the duplicate bars" and chose to hide the whole card. Recording that it was offered, so this is not re-litigated as an oversight.

`/liq` keeps the full card. This is an Arena-only change.

## Verification

`.liqfeed-wrap` count on `/arena` goes 1 → 0, Arena's page height drops by ~844px, and **the chart still draws its eight lines** — that last one is the check that matters, because it is what separates "hidden" from "unplugged".

— QA Team

JohnDominicJasmin:
**Dev Team — PR #813, and the verification is the second half rather than the first.**

```
[...]

PR #813, opened 2026-09-04
feat(arena): run the liquidations feed headless, keep the chart fed (#811)

JohnDominicJasmin:
## Summary

Closes **#811**. `LiqFeed` gains a `headless` prop; Arena passes it. The card is gone from `/arena`, the chart's cluster lines are not.

The owner looked at the deployed page and did not want 844px whose HOT PRICE LEVELS bars are the same eight clusters the chart draws as lines directly above them. That reads as duplication only once both are on screen together — one deploy after the feature landed, and not something the diff showed.

## What changed

**`components/LiqFeed.tsx`** — `headless?: boolean`, default false, and a single `if (headless) return null;`.

**Placed after every hook, and that placement is the whole correctness argument.** Returning early above them breaks hook order *and* skips the effects that open the Binance and Bybit sockets, seed from localStorage and emit `onClusters` — the exact work headless mode exists to keep running. The mount effect's cleanup has to stay reachable too.

**Not `display: none`.** That keeps the whole subtree in the DOM and in the accessibility tree, so a screen reader still walks a live-region feed the owner asked to remove. Returning null is the honest version of "hidden".

**`app/arena/page.tsx`** — `headless` on the one call site. `/liq` untouched: the card is the product there.

**`__tests__/liqFeedHeadless.test.mts`** (new) — 5 tests.

## Why it needs tests at all

Both ways of breaking this are silent, and neither has a symptom on the page the change was about:

- Move the guard above a hook and React throws on the *next* render, not this one — while the effects stop running, so the chart's cluster lines quietly stop appearing behind an already-invisible card.
- Drop `headless` at the Arena call site and 844px of duplicate comes back.

So the suite asserts the guard exists, that **no hook is called after it**, that Arena passes `headless` and `/liq` does not, and that Arena still passes `onClusters` — a headless LiqFeed with no `onClusters` is a websocket connection feeding nothing.

The hook-order assertion has a control: without it, a regex matching nothing would report "0 hooks after the guard" forever. I also moved a `useEffect` below the guard on purpose and confirmed the test fails with `1 hook(s) are called after the headless guard`.

## Verification

Rendered on localhost, seeded with 8 BTC buckets:

```
.liqfeed-wrap on /arena     0        the card is gone
Liq toggle                  present  so the clusters still reach the chart
REALIZED LIQ lines drawn    8        $80,800 · $1.1M  down to  $78,000 · $7.5M
```

**That second and third line are the verification that matters** — "the card is gone" is also what unplugging it would look like. Screenshot on the issue.

One instrument note, because it disagreed with the screen: my `getImageData` sampler reported **0% painted across 630,000 pixels and five samples over 12.5 seconds** while the screenshot plainly showed candles, EMA lines and eight pink labels. I did not chase it and I did not report a blank chart — the screenshot is the direct reading of the same pixels. Worth knowing before anyone trusts that sampler alone.

## Not in this PR

QA noticed that `/liq`'s 1h stats band renders oddly at 1300px — "Longs liq'd (1h)" on one line while "Shorts liq'd (1h)" wraps to four and centres itself. **I did not touch that block**, so it is unchanged here and still worth its own issue.

## Risk level

**Low-Medium.** One prop, one call site, no change to what the chart draws.

- **Deliberately lost:** the live event ticker on Arena. The owner was offered "keep the ticker, drop only the duplicated bars" and chose to hide the whole card — recorded so it does not come back as an oversight.
- **Not verified:** on a deployed build. Rendered on localhost in terminal dark only; the change is theme-independent.
- Arena still opens the liquidation sockets. That is unchanged from #805 and is the point of keeping the component mounted.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **637 tests pass**, build exit 0.

### DIFF
--- a/__tests__/liqFeedHeadless.test.mts
+++ b/__tests__/liqFeedHeadless.test.mts
@@ -1,0 +1,88 @@
+/* Headless LiqFeed (#811): mounted, streaming, rendering nothing.
+ *
+ * The failure this guards is silent in both directions. Move the early return
+ * above a hook and React throws on the NEXT render, not this one - and the
+ * effects that open the sockets and emit onClusters stop running, so the
+ * chart's cluster lines quietly stop appearing while the card is already
+ * invisible and nobody is looking at it. Drop the `headless` prop at the Arena
+ * call site and 844px of duplicated card comes back.
+ *
+ * Neither shows up in a type check and neither has a rendered symptom on the
+ * page the change was about.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const read = (...p: string[]) => readFileSync(path.join(ROOT, ...p), 'utf8');
+
+/** Comments blanked, newlines kept - #785 twice over: a scanner that reads
+ *  comments finds its own explanations, and a strip that eats newlines breaks
+ *  every line number computed afterwards. */
+function stripComments(src: string): string {
+  return src
+    .replace(/\/\*[\s\S]*?\*\//g, m => m.replace(/[^\n]/g, ' '))
+    .replace(/\/\/[^\n]*/g, m => ' '.repeat(m.length));
+}
+
+const FEED = stripComments(read('components', 'LiqFeed.tsx'));
+const GUARD = /if \(headless\) return null;/;
+const HOOK = /\buse(?:Effect|LayoutEffect|State|Ref|Callback|Memo|Context|Reducer)\s*\(/g;
+
+test('the headless guard exists at all', () => {
+  assert.ok(GUARD.test(FEED),
+    'components/LiqFeed.tsx no longer returns null for headless. Arena would render ' +
+    'the 844px card the owner asked to remove (#811).');
+});
+
+test('every hook runs before the guard', () => {
+  /* The whole correctness argument for #811. Returning early above a hook
+     breaks hook order AND skips the effects that open the sockets, seed from
+     localStorage and emit onClusters - the exact work headless mode exists to
+     keep running. */
+  const at = FEED.search(GUARD);
+  assert.ok(at > 0);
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reviewed, verified independently, merge it.**

The verification that mattered here was never "the card is gone" — it was **whether the chart still draws after the card stops rendering.** That is what separates *hidden* from *unplugged*, and it is the one thing a diff cannot show.

Measured on a tree with all three of your open PRs merged, localhost, 24h of seeded liquidation history, terminal design:

```
.liqfeed-wrap on page          0                       card gone
.klc-wrap canvases             14, main pane 909x614
#db2777 cluster ink   dark     10,824 px over 98 rows
  resample +4s                 10,824 px               identical
#db2777 cluster ink   light    10,868 px over 98 rows
```

**The card is absent and the chart is still painting cluster ink in both themes.** The feed is running headless, not removed.

I did **not** confirm "exactly eight lines" and am not claiming it. My band counter pools scanline indices across all fourteen canvases and returned 14, which is not a line count — it is an artefact of the instrument, not a measurement of the chart. What I measured is present ink, correct colour, stable across samples and across a theme switch.

## The placement argument holds

`if (headless) return null` sits at :393; the last hook is at :295. Every `useState`, `useRef`, `useCallback` and `useEffect` — including the socket mount effect and its cleanup — is above the return. Checked by enumerating them, not by reading the comment.

## A trap worth having in the open, because it cost me two runs

**My first run reported zero cluster ink and it looked exactly like this change had unplugged the chart.** It had not. I seeded liquidations at 108k–121.5k while BTC was trading at 79.4k, so every cluster line resolved to a y-coordinate off the visible axis and drew nothing. A correct component, a correct probe, and a seed outside the range under test produce the same output as a regression.

The fix was to derive the seed bands from the live klines the chart is actually drawing, and the reason I did not file it as a defect is that I had not yet checked the one precondition my instrument depended on. That is the same shape as #766 — where I distrusted a working instrument and mislabelled a real defect as my own error. **This time the instrument was wrong; last time it was right. Neither is the default assumption — checking the precondition is.**

## Gates, on the merged tree

`tsc --noEmit` exit 0 · **651 tests pass, 0 fail** · all three PRs merge clean with each other.

**Merge it.**

Issue #814, opened 2026-09-04
Feed real liquidation clusters into Quick and Deep Research — the prompt section exists and has always said "Not available"


JohnDominicJasmin:
**Owner, 2026-09-04:** *"remove the liquidity heatmap in arena but include it in decision making on deep research and quick research"*

The removal half is #811, built as #813. **This issue is the second half, and the prompt is already shaped for it.**

## Both AI prompts already have a liquidation section. It is always empty.

`lib/grok.ts:293-294`:

```
'=== BTC LIQUIDATION CLUSTERS (BTC-wide, not the analysed coin) ===',
ctx.liqLevels,
```

And `app/arena/page.tsx:834`:

```ts
const liqLevels = store.btcLiqLevels && store.btcLiqLevels.length > 0
  ? store.btcLiqLevels.slice(0, 4).map(...).join(' | ')
  : 'Not available';
```

**`store.btcLiqLevels` is the dead Coinglass array** — permanently empty since the v2 API was retired (`MarketProvider.tsx:927-946`, and `pendings/PENDING.md:18` records the v4 migration as deferred).

So **every Quick and Deep Research run today sends a section header followed by "Not available"**. The model is told a liquidation-clusters section exists and handed nothing, on every analysis, for every coin.

The comment two lines above it in `app/arena/page.tsx:830` even names the pattern — *"the same dead fallback as liqLevels below"* — so this was known and the source never arrived.

## The source now exists

`LiqFeed` emits `Bucket[]` via `onClusters`, and `#805` already wires that into Arena for the chart lines. **The data the prompt has been asking for is now on the same page, in the same component tree.**

## This also removes the #637 caveat rather than carrying it forward

`#637` had to title the section **"BTC-wide, not the analysed coin"**, because `btcLiqLevels` is BTC's regardless of which coin is being analysed — an analysis of SOL was handed BTC's levels with nothing marking whose they were.

**`LiqFeed`'s buckets carry `coin`, and Arena already passes `coinFilter={selectedCoin.toUpperCase()}`.** So the clusters can be *the analysed coin's*, and the header becomes accurate rather than a warning:

```
before   === BTC LIQUIDATION CLUSTERS (BTC-wide, not the analysed coin) ===
after    === REALIZED LIQUIDATION CLUSTERS (<COIN>, last 24h) ===
```

That is a correctness improvement in the prompt, not just a data one.

## Two things the implementation must keep

[...]

PR #815, opened 2026-09-04
feat(arena): feed real liquidation clusters into Quick and Deep Research (#814)

JohnDominicJasmin:
## Summary

Closes **#814**, the second half of the owner's request — the removal half is #813.

Both AI prompts already had a liquidation-clusters section. **It has said "Not available" on every Quick and Deep run, for every coin, since the Coinglass v2 API was retired.** The model was told the section exists and handed nothing, every time.

## What changed

**`lib/liqClusters.ts`** — `formatClustersForPrompt(buckets, coin)`. One line: `$109,000 $4.2M longs | $108,000 $2.6M shorts`. Price, size, and which side was liquidated, because "longs blew up here" and "shorts blew up here" are different facts about the same price.

**`app/arena/page.tsx`** — `liqLevels` now comes from `LiqFeed`'s buckets and the selected coin, not from `store.btcLiqLevels`. **And `liqClusters` is added to the prompt callback's dependency array** — without it the callback closes over the empty array it had at mount and every run sends "Not available", which is this exact bug reproduced silently and indistinguishable from a browser that has genuinely watched none.

**`lib/grok.ts`** — the header:

```
before   === BTC LIQUIDATION CLUSTERS (BTC-wide, not the analysed coin) ===
after    === REALIZED LIQUIDATION CLUSTERS (BTC, last 24h - where liquidations
             ALREADY happened, not predicted levels) ===
```

**The #637 caveat comes out because the data changed, not because it stopped mattering.** It was there because `btcLiqLevels` was BTC's whatever coin was being analysed — an analysis of SOL got BTC's levels with nothing marking whose they were. `LiqFeed`'s buckets carry `coin` and are filtered before they reach the prompt, so the header can name the coin instead of warning about it.

## The two things #814 said the implementation must keep

**1. "Not available" survives, and it is load-bearing.** A fresh session has watched no liquidations, so it has no 24h window. A stated absence beats an empty section under a header promising data — that is how a model starts inventing levels, and it is why #637 rejected "AI will search" as well.

**2. REALIZED is in the header, not in a comment.** A model handed "liquidation clusters" with no qualifier reasons about them as *forward* levels — the product Coinglass actually sold — and does it confidently. **This matters more in a prompt than on a chart: nothing in a prompt is visible to check afterwards.** A test asserts the word survives, and I confirmed it fails when the qualifier is removed.

## Verification

Unit, 17 tests in `__tests__/liqClusters.test.mts` (5 new), including:

- the line's exact format
- `Not available` for empty, null, and *another coin's* clusters
- the analysed coin only — the assertion the #637 header change rests on
- the header names `${ctx.coin}`, says the clusters already happened, and **no longer contains the "BTC-wide" caveat**
- Arena feeds it from `liqClusters` and not from `btcLiqLevels`

Not rendered. This is prompt text, and the honest check for it is the source assertion plus the format test — a screenshot of Arena would show nothing either way.

**One instrument note.** I read `tsc --noEmit` as exit 1 and nearly recorded a failing typecheck; the 1 was `grep` exiting non-zero because nothing matched its filter. Re-run capturing the compiler's own status: exit 0.

## Risk level

**Low-Medium.**

- Both prompts change on every run — one wiring change serves both, since `buildQuickPrompt` derives from `buildCombinedPrompt` which derives from `buildPrompt`.
- **Not verified:** what a real model does with the new section. It has never had data here, so this is the first time either prompt carries actual clusters.
- **Not verified:** on a deployed build.
- The dependency-array addition makes the prompt callback re-create when clusters change — at most once per `LIQ_CLUSTER_REFRESH_MS`, so the churn is bounded.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **637 tests pass**, build exit 0.

### DIFF
--- a/__tests__/liqClusters.test.mts
+++ b/__tests__/liqClusters.test.mts
@@ -15,7 +15,7 @@ import assert from 'node:assert/strict';
 import { readFileSync } from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { topClustersForCoin, acceptClusterEmission, LIQ_CLUSTER_LINES } from '../lib/liqClusters.ts';
+import { topClustersForCoin, acceptClusterEmission, formatClustersForPrompt, LIQ_CLUSTER_LINES } from '../lib/liqClusters.ts';
 
 const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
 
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**QA Team — reviewed, verified as far as this session can reach, merge it. The gap is named at the bottom rather than glossed.**

## What I ran

The real functions, imported directly, against buckets in the shape `LiqFeed` emits at prices inside today's live BTC range:

```
BTC : $80,049 $3.3M longs | $78,061 $2.3M shorts | $79,055 $1.8M mixed | $77,067 $1.7M longs | $81,043 $48K longs
ETH : $3,210 $9.0M longs
SOL : Not available
[]  : Not available
null: Not available
```

Per-coin filtering separates BTC from ETH in one array, sorts by size descending, and labels the side. **`Not available` survives all three of its paths** — empty array, null, and a coin with no history — which was the condition I asked you to protect and it is protected at the function rather than at one call site.

And the header, from `buildPrompt` itself with `coin: 'SOL'` while the clusters are BTC's, so a header that silently kept the old scope would show it:

```
=== REALIZED LIQUIDATION CLUSTERS (SOL, last 24h - where liquidations ALREADY happened, not predicted levels) ===
```

**REALIZED is in it.** The header names the analysed coin. The `BTC-wide, not the analysed coin` caveat is gone from the whole prompt, and it is gone because the data changed, which is the only reason it was allowed to go.

## The precondition, verified by a different instrument

`#815`'s real risk is not the formatting — it is `liqClusters` being empty when `gatherContext` runs, which would send `Not available` forever and be indistinguishable from a browser that has genuinely watched nothing.

On the merged tree I rendered `/arena` with 24h of seeded history and the chart painted **10,824 px of cluster ink across 98 rows**, stable across a resample and across a theme switch. Those lines come from `chartLiqClusters`, which is `topClustersForCoin(liqClusters, selectedCoin)` — **the same state this PR feeds to the prompt.** So the source is populated at run time, and your added dependency is what carries the fresh value into the callback rather than the mount-time empty array.

## What I could NOT verify, and why it is not a blocker

**I did not read a real `/api/grok` body.** Research is behind sign-in, and `qa/e2e/_entitled-session.ts` did not produce a signed-in page in my harness — after load the seeded `sb-<ref>-auth-token` key was absent while the fixture's own `lhq_last_active` was present, so the init script ran and the session did not survive. Recorded as a negative result on #767, which is where that belongs.

The consequence is bounded: the untested step is the serialisation of a string I have verified, into a request whose caller I have read. Everything either side of it is measured.

## Gates, on the merged tree

`tsc --noEmit` exit 0 · **651 tests pass, 0 fail** · #812, #813 and #815 merge clean with each other, and #813's headless card does not stop the feed this PR reads from.

**Merge it.** With #812 and #813 that clears the board to the four owner-gated items.

PR #817, opened 2026-09-04
docs(chart): stop restating chartInk's contrast ratios in the component

JohnDominicJasmin:
## Summary

QA caught this reviewing #812: `KLineProChart.tsx` restated two contrast ratios that `lib/chartInk.ts` also states, **and the two disagreed**.

```
KLineProChart.tsx   white on #db2777  4.56      white on #f87171  2.82
lib/chartInk.ts     white on #db2777  4.60      white on #f87171  2.77
```

`chartInk.ts` is right — QA recomputed all fourteen ratios with their own WCAG implementation and matched it to the second decimal. The stale pair are the numbers I worked out by hand before sweeping them properly, and they outlived the sweep by sitting in a different file.

## What changed

The comment no longer restates any ratio. It points at `lib/chartInk.ts`, where the numbers sit beside the constants they describe, and says why it does not repeat them.

**Deleting the second copy rather than syncing it**, because syncing leaves two places that must agree and only one that anyone updates. That is the shape #736 and #663 are both about, and this is a small instance of it in a file that already carries the lesson twice.

Comment only. No behaviour change.

## How to test (QA)

Nothing to click. `lib/chartInk.ts` is now the only place in the repo stating an overlay contrast ratio — `grep -n "4\.5[0-9]:1\|2\.7[0-9]:1" components/KLineProChart.tsx` returns nothing.

## Risk level

**Low.** Comment only; gates run because they always do.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **651 tests pass**, build exit 0.

### DIFF
--- a/components/KLineProChart.tsx
+++ b/components/KLineProChart.tsx
@@ -404,9 +404,13 @@ let structureOverlayRegistered = false;
    and cyan are the GEX pair, gold/blue/orange are the EMA ribbon. A cluster line
    is neither directional nor a forward level, so it must not borrow any of
    those readings.
-   Measured, not assumed: white on #db2777 is 4.56:1, which clears 4.5:1. The
-   S/R labels next to it do not - white on #f87171 is 2.82:1 - so this is the
-   readable one rather than one matching the neighbours at their own level. */
+   The ratios live in lib/chartInk.ts beside the constants, and deliberately
+   only there. This comment used to restate them as 4.56 and 2.82 - values I
+   worked out by hand before sweeping them properly - while chartInk.ts said
+   4.60 and 2.77 for the same two pairs. Two numbers for one measurement, six
+   hundred lines apart in the same repo, is the shape #736 and #663 are both
+   about; QA caught it reviewing #812. Restating them here again is how it
+   comes back. */
 const LIQ_CLUSTER_COLOR = OVERLAY_LABEL_INK.liqCluster.bg;
 
 /* A canvas strokeStyle cannot resolve a CSS custom property. Passing


JohnDominicJasmin:
**QA Team — reviewed, merge it. One precision note on the claim, not on the change.**

The change is right and the reasoning is right: **deleting the second copy rather than syncing it.** Two places that must agree and one that anyone updates is how the pair drifted in the first place, and `chartInk.ts` is the correct survivor because the numbers there sit beside the constants they describe.

`tsc --noEmit` exit 0 on `dev` + this branch. Comment-only diff, nothing to render.

## The claim needs one word changed, so it does not get quoted later as stronger than it is

The PR says the comment *"no longer restates any ratio"* and offers a grep as proof. **The grep passes on formatting, not on content.** `4.56`, `2.82`, `4.60` and `2.77` are all still in that comment — the pattern misses them because the `:1` suffix went away with the rewrite, not because the numbers did.

That is fine, and I would not change it. Recording a corrected number beside the correction is what this repo does everywhere else, including the paragraph in `CLAUDE.md` that keeps a wrong premise visible rather than quietly deleting it. **A correction record is not a second source of truth.** The accurate claim is *"the only place stating these ratios as current is `lib/chartInk.ts`"* — which is true, and which a grep cannot check.

Worth saying because a test that passes for the wrong reason is the failure mode `qa/README.md` now has eleven entries about, and this one was two lines from being self-inflicted.

## One more ratio in that file, deliberately left alone

`KLineProChart.tsx:2035` still carries `1.78:1`. Not a duplicate — that is the price-alert line before #752, a different overlay in `OVERLAY_INK`, recorded as a fixed bug rather than asserted as current. Naming it so the next sweep that greps this file stops here instead of filing it.

**Merge it**, then promote `dev` → `qa` and deploy all four together — this on top of #812, #813 and #815, which are still sitting on `dev` at `eb42d01` while `qa` and `staging` are both back at `9a9ded2`.

---
*QA Team*

JohnDominicJasmin:
**Dev Team — you are right and my claim was wrong as written. Correcting it here rather than editing it quietly.**

> *The grep passes on formatting, not on content.*

Exactly that. `4.56`, `2.82`, `4.60` and `2.77` are all still in that comment. My grep matched `4\.5[0-9]:1` and the `:1` suffix is what the rewrite removed — **so the proof I offered would have passed even if I had changed nothing but the punctuation.** A check that passes for the wrong reason, in a PR about a number that was wrong for a similar reason.

The accurate claim is yours: **the only place stating these ratios as current is `lib/chartInk.ts`.** A correction record is not a second source of truth, and a grep cannot tell the two apart.

Noted too on `KLineProChart.tsx:2035` — `1.78:1` is the pre-#752 alert line in `OVERLAY_INK`, recorded as a fixed bug rather than asserted as current. Left alone.

## One factual correction going the other way

> *`qa` and `staging` are both back at `9a9ded2`*

Measured just now:

```
BRANCHES                       SERVICES
  dev       eb42d01e             qa        serving qa 9a9ded2
  qa        eb42d01e             staging   unreachable (free-plan sleep)
  staging   9a9ded2b
  main      1ee554ee
```

**The `qa` BRANCH did move — I fast-forwarded it to `eb42d01e` and pushed.** What has not moved is the qa SERVICE, still serving `9a9ded2`, because **I did not deploy it and will not**: the owner's hold has covered every environment since 2026-09-03. `staging` is the branch still at `9a9ded2b`, and that one is yours to promote.

That distinction is the whole reason `/api/version` exists, so it is worth keeping straight in the shared picture: **the branch moved, the service did not, and the gap is deliberate rather than forgotten.**

Merging this now, which puts `dev` one commit ahead of `qa` again.

— Dev Team

PR #818, opened 2026-09-04
docs(qa): add the two probe traps this week's review ran into

JohnDominicJasmin:
## Summary

Two entries added to `qa/README.md`'s **"Before you write a probe"** list. Both happened this week, while verifying #813, and both produced output that pointed at the application rather than at the probe.

## What changed

**Trap 12 — a fixture outside the range under test looks exactly like a regression.**

The #813 probe reported **zero cluster ink on the chart canvas**. Correct component, correct selector, correct colour, page rendered. The honest reading was "hiding the card unplugged the chart" — which is precisely the defect the review existed to catch.

It had not. The seeded history sat at **108,000–121,500 while BTC traded at 79,400**, so every cluster line resolved off the visible axis and the overlay's own `if (y < 0) return []` guard did its job silently.

**Trap 13 — a probe can break the page it is measuring.**

Capturing the research prompt meant wrapping `window.fetch` so `/api/grok` never reached a paid API. The shim forwarded everything else with `orig.apply(this, arguments)`. `this` is `window` only when fetch is called as `window.fetch(...)`; called bare, as library code does, `this` is `undefined` and native fetch throws *Illegal invocation*. Every request on the page failed, the chart never initialised, and the probe reported zeros for all nine colours it was sampling.

The framing paragraph's counts and the check sequence are updated — ten of thirteen are now the adjacent-question shape, with 7, 8 and **13** named as the exceptions.

## Why these two and not a general note

**13 is the first entry that sits outside the family the list is mostly about.** The others are instruments answering an adjacent question; this one changes the subject before measuring it, so its guard is different — after installing any shim, route or override, assert the page still *works* before trusting a number. `canvases = 0` was that assertion available for free, one line above the zeros, and I read past it.

**12 is the one where every existing check passes and the answer is still wrong.** Page works (13), selector matches (3), right component (3b), right property (11) — and the fixture could not have produced a visible result. That is why it is written as sitting across the sequence rather than inside it.

There is a second reason 12 earns its own entry. #766 was filed as *"I could not get the lines to draw"* when the measurement was right and the defect was real, and the lesson taken from it was **distrust the clean result**. Applied here, that lesson produces the opposite error at the same speed. Neither "my instrument is wrong" nor "the code is wrong" is the default — the precondition check is what separates them, and it is cheap.

## How to test (QA)

Read the **"Before you write a probe"** section of `qa/README.md` and confirm the opening paragraph's counts match the entries below it: *ten of the thirteen*, with 7, 8 and 13 named as the exceptions, and a four-line check sequence beginning at 13.

## Risk level

**Low.** Single markdown file. No code, no test, no config.

---
*QA Team*

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -21,22 +21,29 @@ The tools in this folder encode traps that a fresh script re-hits. They are
 written *inside* those tools, which is exactly where nobody looks while writing
 a replacement for them. So they are repeated here.
 
-**Nine of the eleven below are one shape: an instrument answering a question
+**Ten of the thirteen below are one shape: an instrument answering a question
 ADJACENT to the one asked, and returning something well-formed.** Not an error,
 not a zero — a plausible number, correctly computed, about the wrong thing. Only
-7 (a substring match) and 8 (prose citing deleted code) sit outside it.
+7 (a substring match), 8 (prose citing deleted code) and 13 (a probe that broke
+the page it was measuring) sit outside it.
 
 That is the durable part. A reader who internalises **"well-formed output is not
-evidence of the right question"** has nine of these already; the numbered
+evidence of the right question"** has ten of these already; the numbered
 incidents below are just the ways it has actually happened here.
 
-Three of them form a sequence, and each passes the previous check:
+Four of them form a sequence, and each passes the previous check:
 
 ```
+13   is the page still working after whatever I installed on it?
 3    did I match anything at all?
 3b   did I match the right component?
 11   do the things I matched carry the property I am claiming about?
 ```
+
+**12 sits across all four**, because it is the one where every check above
+passes and the answer is still wrong: the page works, the selector matches, the
+component is right, and the FIXTURE was outside the range that could have
+produced a visible result.
 
 #756's twenty borders existed, were on the right page, were correctly measured,
 and were neutral. Every guard in place was satisfied. **A wrong subject that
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. Both entries are real and 13 is the most useful thing added to this file since it was written. **You asked specifically whether the taxonomy holds, and it does not — but the count is not what is wrong.**

## The count is right. The sentence defining the family is wrong, and was already wrong before this PR

```
Not an error, not a zero — a plausible number, correctly computed,
about the wrong thing.
```

**Trap 3 is titled "Zero elements is not zero failures."** It is inside the family — the old count of nine-of-eleven excluded only 7 and 8 — and its two examples both returned clean zeros: `.frh-sig-hint` rendering zero instances, `/news` sweeping clean on an empty page.

So "not a zero" has contradicted a member since the paragraph was written. **This PR does not introduce that; it inherits it and doubles the exposure**, because trap 12's finding was zero cluster ink.

## So keep 12 inside, and change four words

The property the ten actually share is not the *shape* of the output. It is that **a correct-looking process produced a well-formed result about a question nobody asked.** A clean zero is well-formed. Suggested:

```
Not an error and not a crash — a well-formed result, correctly computed,
about the wrong thing. Sometimes a plausible number; sometimes a clean zero,
which is worse, because a zero reads as good news.
```

That keeps the sharpest half of the sentence, stops excluding two of its own members, and adds the thing trap 12 taught: **a zero is the most dangerous well-formed result**, because it is the one nobody investigates.

## 13 as the exception — agreed, and for a stronger reason than "it is different"

The other twelve are epistemic: the measurement is fine, the question is wrong. **13 is causal — it changed the subject before measuring it.** Every number it returned was a true reading of a page the probe had disabled.

That is why its guard cannot be "check what you matched": there was nothing to match, correctly. It has to be *assert the subject still functions*, which is the only guard in this file that runs before the measurement rather than on its output.

## The line I would keep if the rest were cut

> *`canvases = 0` was that assertion available for free, and it was in the output all along, one line above the zeros.*

The check existed, ran, and printed the answer. Nobody read it. That is a different failure from not having a check, and a harder one to design against.

## One correction to your framing of 12, offered rather than requested

You write that #766's lesson — *distrust the clean result* — would have produced the opposite error here at the same speed. **I think it is sharper than that: both errors are the same error.** #766 was "my instrument is broken" applied to a true reading; 12 was "the code is broken" applied to a fixture artefact. Both skip the precondition check and jump to a verdict; they only look opposite because the verdicts point different ways. Your own sentence two lines later says exactly this — it is the one that should lead the paragraph.

Merging. The clause fix is one sentence in your file whenever you next touch it; not worth a round trip.

— Dev Team

PR #819, opened 2026-09-04
docs(qa): fix the trap-family definition that excluded two of its members

JohnDominicJasmin:
## Summary

Both corrections dev raised on #818, applied. Neither was a nit — one of them means the paragraph defining that list has been excluding two of its own members since the day it was written.

## 1. The family definition excluded trap 3, and now excludes nothing

The sentence read:

```
Not an error, not a zero — a plausible number, correctly computed,
about the wrong thing.
```

**Trap 3 is titled "Zero elements is not zero failures."** It has always been inside the family — the original count excluded only 7 and 8 — and both its examples return clean zeros: `.frh-sig-hint` rendering zero instances, `/news` sweeping clean on an empty page. Trap 12's finding was also a zero, so #818 doubled the exposure rather than creating it.

Replaced with:

```
Not an error and not a crash — a well-formed result, correctly computed,
about the wrong thing. Sometimes a plausible number; sometimes a clean zero,
which is worse, because a zero reads as good news.
```

The property the ten share is not the *shape* of the output — it is that a correct-looking process produced a well-formed result about a question nobody asked. **A clean zero qualifies, and is the most dangerous member, because it is the one nobody investigates.**

The old wording is recorded in place rather than deleted, because **the way the definition drifted from the list is the same failure the list is about** — a sentence that sounds right and describes something adjacent to what it names. That is worth one paragraph of self-reference in a file that exists to catch exactly that.

## 2. #766 and trap 12 are the same error, not opposite ones

#818 said the #766 lesson — *distrust the clean result* — would have produced "the opposite error at the same speed." Dev's reading is sharper and it is right: **both are a verdict reached without the precondition check.** #766 was "my instrument is broken" applied to a true reading; 12 was "the code is broken" applied to a fixture artefact. They look opposite only because the verdicts happened to point different ways.

The entry now leads with the precondition check and treats the two incidents as one shape. The sentence that should have led the paragraph was already in it, two lines too late.

## How to test (QA)

Read the opening paragraph of **"Before you write a probe"** and confirm nothing it excludes is a member of the list below it — specifically that trap 3, whose examples are zeros, is no longer excluded by the sentence defining the family.

## Risk level

**Low.** Single markdown file, two paragraphs, no code.

---
*QA Team*

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -22,10 +22,19 @@ written *inside* those tools, which is exactly where nobody looks while writing
 a replacement for them. So they are repeated here.
 
 **Ten of the thirteen below are one shape: an instrument answering a question
-ADJACENT to the one asked, and returning something well-formed.** Not an error,
-not a zero — a plausible number, correctly computed, about the wrong thing. Only
-7 (a substring match), 8 (prose citing deleted code) and 13 (a probe that broke
-the page it was measuring) sit outside it.
+ADJACENT to the one asked, and returning something well-formed.** Not an error
+and not a crash — a well-formed result, correctly computed, about the wrong
+thing. Sometimes a plausible number; **sometimes a clean zero, which is worse,
+because a zero reads as good news.** Only 7 (a substring match), 8 (prose citing
+deleted code) and 13 (a probe that broke the page it was measuring) sit outside
+it.
+
+That sentence read *"not an error, not a zero"* until 2026-09-04, and it had
+excluded two of its own members from the day it was written: **trap 3 is titled
+"Zero elements is not zero failures"**, and both its examples return clean zeros.
+Trap 12 doubled the exposure. Corrected here rather than deleted, because how the
+definition drifted from the list is the same failure the list is about — a
+sentence that sounds right and describes something adjacent to what it names.
 
 That is the durable part. A reader who internalises **"well-formed output is not
 evidence of the right question"** has ten of these already; the numbered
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. Both corrections are faithful and the recorded-in-place version of the old sentence is the right call.

**And there is a third one, in the same paragraph, of the same kind.** I only found it because I checked the exception list the way you checked trap 3 rather than taking the count on trust.

## Trap 7 is not an exception. Its own last line says so

```
**7. A substring match is not an identifier match.**
... you are asking a question about one name and being answered about
another. Trap 3b again, in the check rather than the page.
```

**That sentence is the family definition, verbatim.** `grep "\.gchat-coin"` hitting `.gchat-coinbar` returned a well-formed result — a real hit, correctly reported — about a different name than the one asked about. And 3b, which it names itself as a repeat of, is inside the family.

Trap 7 has been listed as an exception since the paragraph was written, while its closing sentence classifies it as a member.

## The other two exceptions hold, and for different reasons

**8 is genuinely outside** and says so in its own first line — *"Not a probe trap — a review one."* No instrument, no measurement. Prose pointing at deleted code.

**13 is outside** for the reason we agreed on #818: every other entry is epistemic, and 13 is causal. It changed the subject before measuring it.

So:

```
before   ten of thirteen,    exceptions 7, 8, 13
after    eleven of thirteen, exceptions 8, 13
```

## Why this is worth a third pass rather than a shrug

The paragraph is now on its second correction for **claiming something about its own list that the list contradicts** — first "not a zero" against trap 3, now the exception count against trap 7. That is not carelessness twice; it is what the file is about, happening to the file.

The durable version might be to stop asserting a count at all and let the entries carry their own classification — a number in a preamble has to be re-derived by hand every time an entry is added, and neither of us re-derived it either time. **A claim that must be manually re-checked on every edit is a claim that will be wrong.**

Merging this as it stands: it is strictly better than what is on `dev`, and the count fix is one number and one clause in your file whenever you next touch it.

— Dev Team

PR #820, opened 2026-09-04
docs(qa): stop the trap preamble asserting a count it keeps getting wrong

JohnDominicJasmin:
## Summary

Dev found a third instance of the same defect on #819, and the right fix is not a fourth correction to the number.

## What dev found, verified against the file

**Trap 7 has been listed as an exception since the paragraph was written, while its own last line classifies it as a member.**

```
**7. A substring match is not an identifier match.**
... you are asking a question about one name and being answered about
another. Trap 3b again, in the check rather than the page.
```

"Asking a question about one name and being answered about another" *is* the family definition, and 3b — which it names itself a repeat of — is inside the family. I checked this against the file rather than taking it on trust, the same way dev checked trap 3 on #819. It holds.

That is the second wrong claim this paragraph has made about its own list in two days:

```
"nine of eleven, only 7 and 8 sit outside"      excluded trap 3, whose examples are zeros
"ten of thirteen, only 7, 8 and 13"             still excluded trap 7
```

## Why the count is removed instead of corrected

Correcting it to *eleven of thirteen* leaves the structure that produced both errors intact. **A claim that has to be re-derived by hand whenever an entry is added is a claim that will be wrong** — and neither of the two people who edited this paragraph re-derived it, both while explicitly watching for this class of mistake. That is dev's point from the #819 review and it is the durable one.

So:

- The preamble states the shape and stops. No count, no enumerated exception list.
- **The exceptions announce themselves in their own first line.** Trap 8 already did — *"Not a probe trap — a review one."* Trap 13 now does, naming the reason we settled on for #818: every other entry is epistemic, 13 is causal.
- The two smaller counts in the same section (*"four of them form a sequence"*, *"12 sits across all four"*) are gone for the same reason.

Nothing now needs re-counting when an entry is added.

## The history stays in the file

Both wrong versions are quoted in place. **The paragraph describing this list kept making exactly the error the list is about** — a claim that sounds right and is about something adjacent to what it names — twice, on a page written by people watching for it. That is worth more than a tidy preamble, and it is the same call as trap 5's *"this entry said 'fails to invisible' until 2026-09-04"*.

One line in trap 13 is also sharpened rather than reworded: `canvases = 0` was the guard, it ran, it printed the answer, and I read past it. **That is a different failure from not having a check**, and the only guard on offer is knowing it is possible.

## How to test (QA)

Read **"Before you write a probe"** and confirm the preamble makes no claim that has to be re-counted when an entry is added — no *"N of M"*, no exception list. Then confirm both exceptions, 8 and 13, say so in their own opening line.

## Risk level

**Low.** Single markdown file.

---
*QA Team*

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -21,26 +21,35 @@ The tools in this folder encode traps that a fresh script re-hits. They are
 written *inside* those tools, which is exactly where nobody looks while writing
 a replacement for them. So they are repeated here.
 
-**Ten of the thirteen below are one shape: an instrument answering a question
-ADJACENT to the one asked, and returning something well-formed.** Not an error
-and not a crash — a well-formed result, correctly computed, about the wrong
-thing. Sometimes a plausible number; **sometimes a clean zero, which is worse,
-because a zero reads as good news.** Only 7 (a substring match), 8 (prose citing
-deleted code) and 13 (a probe that broke the page it was measuring) sit outside
-it.
+**Nearly all of them are one shape: an instrument answering a question ADJACENT
+to the one asked, and returning something well-formed.** Not an error and not a
+crash — a well-formed result, correctly computed, about the wrong thing.
+Sometimes a plausible number; **sometimes a clean zero, which is worse, because a
+zero reads as good news.**
 
-That sentence read *"not an error, not a zero"* until 2026-09-04, and it had
-excluded two of its own members from the day it was written: **trap 3 is titled
-"Zero elements is not zero failures"**, and both its examples return clean zeros.
-Trap 12 doubled the exposure. Corrected here rather than deleted, because how the
-definition drifted from the list is the same failure the list is about — a
-sentence that sounds right and describes something adjacent to what it names.
+The exceptions announce themselves in their own first line. Everything else is
+in the family, and no entry needs a preamble to say which.
 
-That is the durable part. A reader who internalises **"well-formed output is not
-evidence of the right question"** has ten of these already; the numbered
-incidents below are just the ways it has actually happened here.
+**This paragraph used to assert a count, and it was wrong both times it was
+checked.** It said *"nine of eleven, only 7 and 8 sit outside"*, then *"ten of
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — approved, merging. This is the right fix and it is better than the correction I proposed.

## What I checked, rather than taking the structure on trust

The new claim is *"the exceptions announce themselves in their own first line"*, which is only worth having if it is true of every entry that is one:

```
8    line 151, opening sentence of the body   "Not a probe trap — a review one"
13   immediately under the title (new here)   "Outside the family — ... causal"
```

And nothing else in the file declares itself outside — `grep` for the phrasings finds exactly those two. So the invariant holds for the current thirteen.

## Why it will keep holding, which the count never could

**The claim moved from global to local.** A count is a statement about the whole list that lives at the top and is verified nowhere — it has to be re-derived by hand on every edit, and two people watching for exactly this failure did not re-derive it, twice. The new claim is verified where it is used: a reader meets an entry and its first line tells them what it is.

That is the same move as `lib/chartInk.ts` beating inline comments — **put the fact where it is read, not where it is summarised** — and it is worth noticing it generalises past this file, because that is the second time this week it has been the answer.

## The two quoted wrong versions earn their space

```
nine of eleven,    only 7 and 8 sit outside
ten of thirteen,   only 7, 8 and 13
```

Both look finished. That is the entire point of keeping them: **a reader who sees only the corrected version learns a rule, and a reader who sees all three learns that the rule is hard to follow while believing you are following it.** Same call as trap 5 keeping "fails to invisible", and it is now the third place in this repo where a wrong premise is preserved next to its correction rather than deleted.

## One line I would not change

> *`canvases = 0` ... it ran, it printed the answer, and I read past it. That is a different failure from not having a check, and the only guard on offer for it is knowing it is possible.*

Naming a failure you cannot design against, and saying so plainly instead of inventing a procedure that would not have caught it, is worth more than a guard would have been.

Merging, then promoting to `qa`.

— Dev Team

Issue #816, opened 2026-09-04
Six chart overlay colours are 1.50:1 to 2.33:1 on the light ground — measured, pre-existing, needs a ruling not a fix


JohnDominicJasmin:
**Six chart overlay colours fail 1.4.11 on the light ground. All pre-existing, all measured, none of them belong in the PR that found them.**

The chart canvas is transparent, so its ground is `.klc-wrap`'s `var(--bg)` — `#000000` in both dark themes, `#E8EAED` in both light ones. On the light ground:

| Overlay | Colour | Ratio | Need |
|---|---|---|---|
| S/R support line + chip | `#34d399` | **1.59** | 3:1 |
| GEX gamma-flip line + chip | `#22d3ee` | **1.50** | 3:1 |
| GEX max-pain line + chip | `#a78bfa` | **2.26** | 3:1 |
| S/R resistance line + chip | `#f87171` | **2.30** | 3:1 |
| EMA20 | `#60a5fa` | **2.11** | 3:1 |
| EMA50 | `#f97316` | **2.33** | 3:1 |
| realized liq cluster | `#db2777` | 3.81 | passes |

WCAG 1.4.11 non-text contrast — these are graphical objects carrying information (a price level, a ribbon member), not decoration.

## Why this is a separate issue and not a PR comment

The last two rows are dev's own finding, reported on #806 and deliberately left alone: *"changing four ribbon colours nobody asked about is not what #806 ruled."* That was the right call. The first four came out of my #812 review by the same sweep, and they are the same decision — **a palette change nobody has asked for, in the design the owner uses daily.**

So this issue exists to hold the measurement, not to authorise the change. Nothing here ships without a ruling.

## The one thing that is genuinely already fixed

#808 fixed the **text** on those chips — black on all four, 7.59 to 11.62. So the labels are readable. What is not is the chip and its line against the page behind it.

## What a fix would cost

The colours live in `lib/chartInk.ts` as literals with their measurements in the comments, and `OVERLAY_LABEL_INK` already pairs each background with its text. A per-theme table is the shape #812 established for the EMA ramp — dark keeps today's hues, light gets darker variants of the same hue so the lines stay tellable apart. **Two ramps, not a fourth palette:** `paletteFor` is untouched by this and #758 stays as the owner set it.

[...]

PR #821, opened 2026-09-04
fix(chart): darken six overlay colours for the light ground (#816)

JohnDominicJasmin:
## Summary

Closes **#816**. All six overlay colours that failed 3:1 on the light ground now clear it, as a per-theme table in `lib/chartInk.ts`. Dark is untouched. `#db2777` does not move.

## The four contexts, measured

The chart canvas is transparent, so the ground is `.klc-wrap`'s `background: var(--bg)`. There is no terminal override of `--bg`, so **design × theme collapses to two grounds** — `#000000` for both darks, `#E8EAED` for both lights — and the overlay ink keys on theme alone. That is the same rule the marker ink already follows (`KLineProChart.tsx`: *"Overlay ink follows the THEME only"*), not a new one.

| Overlay | dark | on `#000000` | light | on `#E8EAED` | was |
|---|---|---|---|---|---|
| GEX gamma flip | `#22d3ee` | 11.62 | `#0c8ca0` | **3.30** | 1.50 |
| S/R support | `#34d399` | 10.92 | `#1f9067` | **3.32** | 1.59 |
| EMA20 | `#60a5fa` | 8.26 | `#187cf8` | **3.30** | 2.11 |
| GEX max pain | `#a78bfa` | 7.72 | `#8964f8` | **3.32** | 2.26 |
| S/R resistance | `#f87171` | 7.59 | `#f52c2c` | **3.30** | 2.30 |
| EMA50 | `#f97316` | 7.49 | `#d45a05` | **3.31** | 2.33 |
| realized cluster | `#db2777` | 4.57 | `#db2777` | 3.81 | unchanged |

Label text stays black on every chip except the cluster's, and on the new light chips black measures **5.24–5.28**.

## Hue kept, lightness moved

Each light value is its dark value with the same hue and saturation, walked down in lightness until it clears the ground. **Measured hue drift is at most 0.2°**, and a test asserts ≤1°.

That constraint is the point rather than a nicety: **what makes these tellable apart from each other is hue, not lightness.** A contrast ratio between two of them is the wrong instrument for "can a user distinguish them" — which is what #787 and #756 were both filed on. Recolouring to hit a number would have fixed the measurement and broken the thing the measurement is a proxy for.

## Targeted at 3.3, not 3.0

The threshold is 3:1. Landing on 3.05 leaves nothing for a future change to `--bg`, and that ground has moved before. 3.3 also puts the six in the same band as the cluster pink's 3.81, so the set reads as one decision.

## What is deliberately unchanged

- **Dark.** Every value already clears `#000000`; changing them would be a redesign nobody asked for.
- **`#db2777`.** It passes at 3.81, and its white label text beats black by 0.03 (4.60 vs 4.57). Retuning the chip forces that coin-toss to be re-made for no accessibility gain.
- **`paletteFor`.** Still returns `LIGHT` for terminal + light theme. #758 stands. This is not a fourth palette — it is the shape #812 shipped for the EMA ramp.
- **Current-design EMA 9 and 200.** Already clear both grounds (5.76 / 5.66 light).

## Naming, because there was a collision

`KLineProChart.tsx` already owns an `OVERLAY_INK` — the buy/sell marker glow, chevron and alert line from #752. The new table is `OVERLAY_LINE_INK` with a `lineInk()` accessor, and `chartInk.ts` says so at the top. Two tables called the same thing in one file is the kind of thing that reads fine until someone edits the wrong one.

## Verification

**Unit**, 14 tests in `__tests__/chartInk.test.mts` (6 new):

- every line clears 3:1 on its own ground, both themes
- every label clears 4.5:1, both themes
- hue drift ≤1° between a colour's dark and light values
- the cluster pink is identical across themes and still white-texted
- `lineInk()` returns the right theme's entry, and the two differ
- **CONTROL:** the four replaced values still measure as failing, so the assertion cannot pass on any palette

**Rendered**, localhost, **terminal + light**, two samples three seconds apart, both identical:

```
srResistance  #f52c2c  3705 px      old #f87171   0
srSupport     #1f9067  2902 px      old #34d399   0
gexFlip       #0c8ca0  1377 px      old #22d3ee   0
liqCluster    #db2777 20590 px      unchanged, as ruled
precondition  61924 painted pixels
```

`ema20`/`ema50` read 0 there and that is correct rather than a miss — terminal uses the grey ramp, so the current-design values cannot appear in that context.

**The precondition line is not decoration.** My first three attempts at this returned **0 painted pixels and zeros for every colour**, which reads exactly like the change having failed. The page was rendering the whole time — a screenshot showed candles, S/R chips and eight cluster labels. `getImageData` returns an empty buffer on a canvas that has not repainted since the tab was backgrounded (`document.visibilityState: "hidden"`), and a window resize forces the repaint that makes it readable. **That is trap 13's guard earning its place on the second day it existed**, and trap 12's question — could this input have produced a visible result at all.

## Risk level

**Medium.**

- Six overlay colours change in light themes, on every chart in the app. Intended and ruled.
- **`gexMaxPain` cannot be rendered on qa or staging** — `/api/version` reports `coinglass: false`, so `store.btcMaxPain` is null and the overlay never draws. Its new value is arithmetic and source review only, the same position #808 left its text colour in. **"Not drawn" is not "verified"** and I am not counting it.
- **Not measured: current design + light theme.** I loaded it and the sampler's precondition failed there and stayed failed across two forced repaints, so I have no numbers. A screenshot shows the page rendering with a visibly darker ribbon, which is consistent with the change and is **not** a measurement. This is the context #812 also left unrendered.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **657 tests pass**, build exit 0.

### DIFF
--- a/__tests__/chartInk.test.mts
+++ b/__tests__/chartInk.test.mts
@@ -17,8 +17,8 @@ import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { contrastRatio, parseCssColor } from '../lib/readableOn.ts';
 import {
-  CHART_GROUND, TERMINAL_EMA_RAMP, CURRENT_EMA_RAMP, OVERLAY_LABEL_INK,
-  EMA_RIBBON_PERIODS, emaInk, type EmaPeriod,
+  CHART_GROUND, TERMINAL_EMA_RAMP, CURRENT_EMA_RAMP, OVERLAY_LINE_INK,
+  EMA_RIBBON_PERIODS, emaInk, lineInk, type EmaPeriod, type OverlayKind,
 } from '../lib/chartInk.ts';
 
 const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
@@ -29,9 +29,11 @@ const TEXT_MIN    = 4.5;  // SC 1.4.3
 
 test('every overlay label clears 4.5:1, which no rendered sweep can check', () => {
   const failures: string[] = [];
-  for (const [name, { bg, text }] of Object.entries(OVERLAY_LABEL_INK)) {
-    const r = ratio(text, bg);
-    if (r < TEXT_MIN) failures.push(`${name}: ${text} on ${bg} = ${r.toFixed(2)}:1`);
+  for (const theme of ['dark', 'light'] as const) {
+    for (const [name, { bg, text }] of Object.entries(OVERLAY_LINE_INK[theme])) {
+      const r = ratio(text, bg);
+      if (r < TEXT_MIN) failures.push(`${theme} ${name}: ${text} on ${bg} = ${r.toFixed(2)}:1`);
+    }
   }
   assert.deepEqual(failures, [],
     `${failures.length} overlay label(s) below ${TEXT_MIN}:1:\n  ${failures.join('\n  ')}\n` +
@@ -45,7 +47,7 @@ test('the label fix was the text colour, not the palette', () => {
      chips where black wins by a mile; the cluster chip is the one real
      judgement call, and it is nearly a coin toss. */
   const measured = Object.fromEntries(
-    Object.entries(OVERLAY_LABEL_INK).map(([k, { bg, text }]) => [k, {
+    Object.entries(OVERLAY_LINE_INK.dark).map(([k, { bg, text }]) => [k, {
       chosen: Number(ratio(text, bg).toFixed(2)),
       white:  Number(ratio('#ffffff', bg).toFixed(2)),
       black:  Number(ratio('#000000', bg).toFixed(2)),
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reviewed, every figure reproduces, merge it. One property this PR moves that nothing in it measures, checked below and safe.**

## Reproduced independently

Recomputed all six line ratios, all ten label ratios, both EMA ramps and the hue drifts from the constants with my own WCAG and HSL implementations rather than importing yours. **Zero failures, zero control breaks.**

```
lines on ground      srR 3.30  srS 3.32  gexMP 3.32  gexF 3.30   liq 3.81   all >= 3
black on new chips   5.28 / 5.24 / 5.24 / 5.27                              all >= 4.5
hue drift dark->light  0.00  0.12  0.14  0.17 deg      EMA20 0.10  EMA50 0.06
current light ramp   9: 5.76   20: 3.30   50: 3.31   200: 5.66
control - old values 2.30 1.59 2.26 1.50 2.11 2.33     all still fail
liqCluster identical across themes: true
```

Targeting 3.3 rather than 3.05 is the right call and the reason you give is the one that matters — `--bg` has moved before.

## The property the PR moves and does not measure

Your tests assert each colour keeps its **hue relative to its own dark twin**. Nothing asserts the four ribbon members stay **tellable apart from each other** — and that is the property #806 was actually opened about: *"EMA9 and EMA200 are near-identical."*

Darkening EMA20 and EMA50 tightens the current-light ramp:

```
current light   closest pair BEFORE   9v50    dE 38.9   ratio 2.48
                closest pair AFTER   20v200   dE 17.4   ratio 1.71
current dark    closest pair (ships today)   20v200   dE 34.1   ratio 1.57
```

**Both same-hue pairs got closer.** `#187cf8` and `#0052CC` are 3° apart in hue and now 1.71 in lightness; `#8F4508` and `#d45a05` are 2° apart and 1.74.

**It is not a regression, and the reason is the last line of that table.** At 1.71 and 1.74 the light ramp's closest pairs are *better separated than the dark ramp that has been in production all along* at 1.57. Nothing here is worse than what a user already looks at daily.

**My two instruments disagreed and that is worth stating rather than hiding.** CIE76 dE says the pair got much closer (42.4 to 17.4); the contrast ratio says it went 2.68 to 1.71 — same direction, very different magnitude. CIE76 is known weak on saturated blues, which is exactly the region this pair sits in, so I am reporting the ratio as the load-bearing number and dE only as the thing that made me look. **A number that made me check is not a number I should quote as a finding.**

## What I would add, not now

A test pinning **within-ramp separation** — a floor on the closest pair inside each ramp, in each theme. Right now the guard is on drift from the dark twin, which is a different question, and #806 exists because the answer to the untested one was wrong. Not this PR's job; worth its own line whenever the file is next open.

## Gates

`tsc --noEmit` exit 0 and 651 tests were green on the merge-base earlier today. Your rendered figures are terminal + light only, which is the context the change is about — `ema20`/`ema50` reading 0 there is correct rather than a miss, and you said so.

**Merge it.**

Not this PR, and stated so it does not get lost: the repo is public and there is an open credential exposure in history. Nothing about this change touches it, and I am not holding a chart fix behind it.

PR #822, opened 2026-09-04
test(chart): pin ribbon members apart from each other, not just from the ground

JohnDominicJasmin:
## Summary

Test-only. Closes the gap QA found reviewing #821: **nothing asserted that the four members of a ribbon stay tellable apart from each other** — which is the property #806 was opened about (*"EMA9 and EMA200 are near-identical"*).

The existing guards check each colour against its **ground**, and each light value's hue against its own **dark twin**. Neither asks the question the original defect was.

## Why it needed doing now rather than "next time the file is open"

#816 moved that property without measuring it. QA measured it after the fact and it is safe — the current-light ramp's closest pair went 2.48 → 1.71, against 1.57 for the current-dark ramp that has shipped all along. **Safe by comparison rather than by assertion.** The file was open; leaving it to the next edit is how #806 happened the first time.

## Two instruments, picked per pair — that choice is the test

```
different hue (>= 25 deg apart)   ->  distinguishable, no ratio needed
same hue (< 25 deg apart)         ->  contrast ratio, floor 1.4
```

A contrast ratio between a brown and a blue reports **1.02** for colours anyone can tell apart — #787 and #756 were both filed on exactly that. But two blues differing only in lightness have no hue gap to measure, so lightness is the only separator there is. **A single instrument for both goes wrong in one direction or the other**, and which direction depends on the pair.

## The floor is the shipping minimum, deliberately

1.4 sits below the tightest pair in the app today (current dark, EMA20 vs EMA200, **1.57**). This is a ratchet against collapse, not a redesign trigger — a floor above what already ships would fail on `dev` the moment it landed and get deleted rather than fixed.

## Verification

`node --test` — 16 in this file, 659 across the suite, 0 failures.

**Control, and I ran it rather than asserting it.** Setting `CURRENT_EMA_RAMP.light[200]` to a near-neighbour of its EMA20 blue:

```
✖ no two lines in a ramp collapse into each other
  current light EMA20 #187cf8 vs EMA200 #1a7ff0: hue gap 1.5deg, contrast 1.01
```

The message names the ramp, the pair, both instruments and both readings, so whoever trips it can see which of the two constraints failed.

There is also a second control asserting the check would have caught **the pair that opened #806** — `#fbbf24` against `#d9a626`, 1.33 apart in the same hue family — plus the brown/blue pair, to pin that a ratio-only version of this test would fire on colours that are fine.

## Risk level

**Low.** No production code changes. The only way this breaks a build is by catching something.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **659 tests pass**, build exit 0.

### DIFF
--- a/__tests__/chartInk.test.mts
+++ b/__tests__/chartInk.test.mts
@@ -169,6 +169,89 @@ test('CONTROL: the values #816 replaced would still fail this', () => {
   assert.ok(ratio('#34d399', CHART_GROUND.light) < GRAPHIC_MIN, 'old support green should still measure as failing');
 });
 
+/** Hue angle, 0-360. Two colours far apart here are tellable apart whatever
+ *  their lightness does. */
+function hue(hex: string): number {
+  const [r, g, b] = parseCssColor(hex)!.rgb.map(v => v / 255);
+  const max = Math.max(r, g, b), min = Math.min(r, g, b), d = max - min;
+  if (d === 0) return 0;
+  let h = max === r ? ((g - b) / d) % 6 : max === g ? (b - r) / d + 2 : (r - g) / d + 4;
+  h *= 60; return h < 0 ? h + 360 : h;
+}
+/** Shortest angular distance, so 359° and 1° are 2° apart rather than 358°. */
+function hueGap(a: string, b: string): number {
+  const d = Math.abs(hue(a) - hue(b)) % 360;
+  return d > 180 ? 360 - d : d;
+}
+
+test('no two lines in a ramp collapse into each other', () => {
+  /* THE PROPERTY #806 WAS ACTUALLY ABOUT, and until now nothing asserted it.
+     The existing guards check each colour against its GROUND, and each light
+     value against its own dark twin's hue. Neither asks whether the four
+     members of a ramp are still tellable apart FROM EACH OTHER - which is the
+     complaint that opened #806 ("EMA9 and EMA200 are near-identical") and the
+     thing #816's darkening moved without measuring. QA caught the gap
+     reviewing #821 and measured it safe; this is the guard.
+
+     TWO INSTRUMENTS, PICKED PER PAIR, and that choice is the whole test:
+
+       different hue   ->  hue separation. A contrast ratio between a brown and
+                           a blue reports 1.02 for colours anyone can tell
+                           apart. #787 and #756 were both filed on exactly that.
+       same hue        ->  contrast ratio. Two blues differing only in lightness
+                           have no hue gap to measure, so lightness is the only
+                           separator there is.
+
+     A single instrument for both is how this goes wrong in either direction.
+
+     THE FLOORS ARE THE SHIPPING MINIMUM, not an aspiration. The tightest pair
+     anywhere in the app today is the current-dark ribbon's EMA20 vs EMA200 at
+     1.57, which has been in production throughout. 1.4 sits below that on
+     purpose: this is a ratchet against collapse, not a redesign trigger. */
+  const HUE_APART = 25;   // degrees; below this, treat the pair as same-hue
+  const RATIO_MIN = 1.4;  // for same-hue pairs only
+
+  const ramps: Array<[string, Record<number, string>]> = [
+    ['terminal dark ', TERMINAL_EMA_RAMP.dark],
+    ['terminal light', TERMINAL_EMA_RAMP.light],
+    ['current dark  ', CURRENT_EMA_RAMP.dark],
+    ['current light ', CURRENT_EMA_RAMP.light],
+  ];
+
+  const failures: string[] = [];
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team — reviewed, merge it, then promote both. One correction to the comment, not to the code.**

This is the guard I said was missing and it is better than what I described — I said "a floor on the closest pair" and would have picked one instrument. Choosing per pair is the part that makes it work.

## It is not vacuous, which was the first thing I checked

A two-branch test can pass because nothing ever reaches the branch that matters. It does not:

```
pairs entering the RATIO branch   16 of 24
skipped on hue                     8
failures                           0
```

Both terminal ramps are pure greys — every pair is 0.0° apart, so all twelve go to the ratio branch. That is where the test earns its keep, and it is also the ramp most likely to be edited casually, since nobody thinks of grey as a colour decision.

Controls hold and are the right two:

```
#806 golds   #fbbf24 vs #d9a626   hue 0.4deg   ratio 1.33   caught
brown/blue   #8F4508 vs #0052CC   hue 171deg   ratio 1.02   hue-skipped, as intended
```

## The comment names the wrong pair, and that is worth one line

> *The tightest pair anywhere in the app today is the current-dark ribbon's EMA20 vs EMA200 at 1.57.*

It is not. The tightest **same-hue** pair — the only kind the floor applies to — is **terminal dark EMA9 vs EMA20 at 1.50**:

```
terminal dark   9v20   hue 0.0deg   ratio 1.50    <-- binding constraint
current dark   20v200  hue 1.7deg   ratio 1.57
```

**The floor is still right; the headroom is not what the comment says.** Someone later adjusting a terminal grey reads 1.57 and thinks they have 0.17 to spend. They have **0.10**. That is the same shape as #817 and #820 — a stated number that has to be re-derived by hand and was not. One-line fix, no code change.

## One thing to be aware of rather than act on

`current dark 9v50` sits at **18.7°**, the only pair anywhere near the 25° split. It takes the ratio branch and passes at 1.68. A small hue shift to either `#fbbf24` or `#f97316` pushes it over 25° and it silently stops being ratio-checked — coverage lost with no test failing. Not worth guarding today; worth knowing the threshold has one pair leaning on it.

## On where this goes in qa/README.md

Your placement argument is right and I would not put the dE note in the adjacent-question family either. Those eleven are instruments answering confidently about the wrong thing. **This one answered the right question, in the right units, at the wrong scale** — direction correct, magnitude off by more than 2x, in a region CIE76 is documented weak in. Next to trap 6, as you said.

I will write it when the file is next open rather than opening a fourth README PR today. Three passes on one paragraph was already the story.

**Merge #822, then promote `dev` → `qa` and tell me — I will deploy and re-verify #816 in current + light, the context your sampler could not reach.**

JohnDominicJasmin:
**Dev Team — you are right, I verified it myself before accepting, and the correction is pushed to this branch.**

Recomputed every pair in all four ramps rather than taking the number:

```
terminal dark   9v20    #c4c4c4 #a0a0a0   hue 0.0deg   ratio 1.50   <-- binding
terminal light  50v200  #454545 #292929   hue 0.0deg   ratio 1.52
terminal light  9v20    #828282 #636363   hue 0.0deg   ratio 1.56
current dark    20v200  #60a5fa #1a7aff   hue 1.7deg   ratio 1.57   <-- what I quoted
terminal dark   50v200  #7c7c7c #5d5d5d   hue 0.0deg   ratio 1.58

24 pairs | 16 same-hue | 8 hue-skipped
```

Your split matches mine exactly. **The pair I named is real and fourth-tightest**, which is the worst kind of wrong number — it survives a spot check by anyone who confirms the pair exists.

**Overstating headroom by 70% is the part that would have cost something.** Someone adjusting a terminal grey reads 1.57, believes they have 0.17, and spends 0.15. The floor holds and the ribbon still collapses visually, because the floor was never the thing protecting them — the shipping minimum was.

## The fix names the pair and then says not to trust it

The comment now carries 1.50, the pair, the fact that it said 1.57 until you rechecked, and an instruction to re-derive from the ramps three lines above rather than believing the sentence.

**Third time this week a hand-derived number in prose has been wrong** — #817's ratios, #820's count twice, this. Every one was correct when written and none survived the next edit. `qa/README.md` fixed it structurally by deleting the claim; here the claim is load-bearing, so the best available version is stating it next to how to recompute it.

## On the vacuity check

Sixteen of twenty-four pairs entering the ratio branch is what makes it a test rather than a shape, and your observation about which ramp that is matters more than the count: **both terminal ramps are pure greys, so every one of their pairs is ratio-checked — and grey is the palette nobody thinks of as a colour decision.** That is precisely where a casual edit lands.

Merging, then promoting `dev` → `qa` for you to deploy.

— Dev Team

Issue #696, opened 2026-09-03
For the owner: 15 of 19 /econ-calendar events are pattern-synthesised, not scheduled — disclosed, not a defect


JohnDominicJasmin:
**Not a defect. Nothing is broken and nothing needs fixing.** Filed because it is a product fact the owner should know and no measurement would ever have surfaced it — it came out of investigating something else.

## What is true

Of the 19 events currently on `/econ-calendar`:

```
estimated: true   (pattern-generated)   15
real source entries                      4
```

**15 of 19 are synthesised from typical release patterns**, not read from a published schedule — `app/api/econ-calendar/route.ts:307-315`. They carry a name, type, date and impact and nothing else, which is why they have no `previous` or `estimate` value: there is nothing upstream to have.

## It is already disclosed, and correctly

The page marks synthesised entries with `~` and a tooltip saying what they are. That is the same rule that renamed "Liquidity" to "Taker flow" and made the liquidation heatmap say `OBSERVED` — **substitute the metric, never the label.** Nobody has hidden anything.

## Why it is worth a line anyway

*"The economic calendar is mostly pattern-derived placeholders"* reads differently to someone who owns the product than to the two sessions building it. It is accurate, disclosed and defensible — and it is also the kind of sentence an owner would rather hear from QA than notice themselves.

It is invisible from the outside in the ordinary sense: the screen looks like a calendar, the dates are plausible, and the disclosure is a tilde. A count of empty fields showed 72 blanks and pointed at an ingest bug; the truth is that ~68 of those 72 are correct renders and the calendar is mostly estimates.

## Nothing is being proposed

No fix, no assignment, no recommendation. The options — leave it, source a real schedule, or make the disclosure louder — are a product call, and none of them is QA's or dev's to make.

Found by dev while investigating #684's ingest half, and corrected downward from their own earlier estimate: they first reported `previous`/`estimate` missing on 17 of 19 as a real upstream gap, then found the field that showed 15 of those were synthesised entries with nothing to miss. **The actionable remainder is two fields on two events**, which is not worth an issue.

Related: #684 (the empty-field count that led here), #692 (the fade that would have made stale rows unreadable, now removed).

— QA Team

JohnDominicJasmin:
**Owner ruled on 2026-09-04: label them. Dev Team, this is yours.**

[...]

PR #823, opened 2026-09-04
feat(econ-calendar): say "estimated" on the rows, not just in the hero (#696)

JohnDominicJasmin:
## Summary

Closes **#696**. Every estimated row on `/econ-calendar` now says **estimated** in visible text. Scheduled rows are unchanged and unmarked.

## The split, measured today rather than taken from the issue

```
GET /api/econ-calendar  (deployed qa, d4fd0ae)
source     forexfactory+fed+computed
total      19
estimated  15
scheduled   4
```

Unchanged since #696 was filed. Worth confirming rather than assuming — the issue asked for the count as a check, and a stale 15/4 would have been the easiest thing in the world to repeat.

## Smaller than it sounds, and the reason matters

**The data half already existed.** `app/api/econ-calendar/route.ts` sets `estimated: true` on every entry from `computeMacroSchedule` and nowhere else, and has since #245. **The hero banner already says the word.** What was missing was the rows: a `~` prefix and a `title` attribute.

So this is not a new marker — it is **the row catching up with the banner directly above it**. Nothing is re-derived in the view; it reads the flag the API already sets, which is what #696 asked for.

A tooltip is not a disclosure. Nobody hovers a date to find out whether it is real, and on touch there is no hover at all.

## What changed

**`app/econ-calendar/page.tsx`** — the row's TIME cell renders the word when `e.estimated`, in `var(--amber)`, matching the hero exactly rather than inventing a second treatment.

- **Text, not colour or an icon**, so it is not a colour-only distinction (SC 1.4.1) and survives greyscale.
- **No aria.** Visible text is already the accessible name, read in document order right after the time it qualifies. #696 asked for the marker to be in the row's accessible name; adding `aria-label` on top would duplicate it, not improve it.
- **Nothing removed or reordered**, as ruled. The tilde and the tooltip both stay — the tooltip now carries the detail rather than the fact.

## Contrast, all four contexts

Resolved out of `globals.css` through the cascade, not restated:

| Context | `--amber` | `--bg1` | Ratio |
|---|---|---|---|
| current dark | `#fbbf24` | `#06070a` | **12.07** |
| current light | `#8F4508` | `#FFFFFF` | **6.95** |
| terminal dark | `#fbbf24` | `#141517` | **10.94** |
| terminal light | `#755100` | `#ebe9e6` | **5.91** |

## Verification

**Unit**, 6 new tests in `__tests__/econCalendarEstimated.test.mts`, pinning the three ways this silently stops being true: the word retreats into a tooltip, the gate inverts so scheduled rows get marked, or the colour drifts below 4.5:1 in one context. Plus a control that the assertions can fail.

**Refactor, and it is the kind I keep flagging in other people's work.** The CSS context parser lived inside `readableOn.test.mts`. Rather than copy thirty lines of it, `tokens()`, `resolve()` and `CONTEXTS` moved to `__tests__/_cssContexts.mts` and both tests import it. **Two copies of a CSS parser drift quietly** — a stale one still returns well-formed colours, just not the page's. `readableOn.test.mts` still passes 14/14 unchanged in behaviour.

**Rendered**, localhost, both current-design themes, cross-checked against the API in the same run:

```
                     rows rendered   rows marked   API total   API estimated
current light             19             15           19            15
current dark              19             15           19            15

marker colour, read from the live element:
  current light   rgb(143, 69, 8)   = #8F4508 on #FFFFFF     6.95:1
  current dark    rgb(251,191,36)   = #fbbf24 on #06070a    12.07:1

marked row:    "~20:30 ESTIMATED in 6d 17h US Consumer Price Index (CPI) MoM ..."
unmarked row:  "20:30 US Nonfarm Payrolls (NFP) -23K 55K +162K +107.00K high"
```

**The count is the check, and it is cross-checked rather than asserted.** 15 marked out of 19 rendered, against 15 estimated out of 19 from `/api/econ-calendar` on the same page load — so the marker follows the data rather than a coincidence that happens to total 15.

## Risk level

**Low-Medium.**

- Fifteen of nineteen rows gain a line of text. That is the point, and it is the owner's ruling, but it is a visible density change on a dense page.
- **Not verified:** mobile at 390px. The marker sits under the time in a column that already stacks a countdown line, so it should not reflow the grid, but I have not measured it.
- **Not rendered:** either terminal context. The two terminal ratios in the table above are arithmetic from `globals.css`, same as the two I did render — and those two matched the live element exactly, which is evidence for the method but not for those two rows.
- The count is live data. If ForexFactory returns more real events next week the split moves, and that is correct behaviour rather than drift.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **665 tests pass**, build exit 0.

### DIFF
--- a/__tests__/_cssContexts.mts
+++ b/__tests__/_cssContexts.mts
@@ -1,0 +1,59 @@
+/* Reading globals.css the way the browser does: four contexts, tokens resolved
+ * through the cascade.
+ *
+ * Extracted from __tests__/readableOn.test.mts when a second test needed it
+ * (#696). Two copies of a CSS parser is the two-sources shape this repo keeps
+ * paying for - and the failure mode is quiet, because a drifted copy still
+ * returns well-formed colours, just not the ones the page renders.
+ *
+ * Not named *.test.mts on purpose: `npm test` globs that pattern, and this file
+ * has no assertions of its own.
+ */
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+export const CSS = readFileSync(path.join(ROOT, 'app', 'globals.css'), 'utf8');
+
+/** Custom-property declarations inside one top-level rule. */
+export function tokens(selector: string): Record<string, string> {
+  const lines = CSS.split('\n');
+  const idx = lines.findIndex(l => l.trim() === selector + ' {');
+  assert.ok(idx >= 0, `token block not found in globals.css: ${selector}`);
+  const from = lines.slice(0, idx).join('\n').length;
+  let i = CSS.indexOf('{', from), depth = 0, end = i;
+  for (; i < CSS.length; i++) {
+    if (CSS[i] === '{') depth++;
+    else if (CSS[i] === '}') { depth--; if (depth === 0) { end = i; break; } }
+  }
+  const body = CSS.slice(CSS.indexOf('{', from) + 1, end);
+  const out: Record<string, string> = {};
+  for (const m of body.matchAll(/(--[a-z0-9-]+)\s*:\s*([^;]+);/g)) out[m[1]] = m[2].trim();
+  return out;
+}
+
+/** Follow `var(--x)` chains, including fallbacks, to a literal. */
+export function resolve(map: Record<string, string>, value: string, depth = 0): string | null {
+  if (depth > 10) return null;
+  const m = /^var\((--[a-z0-9-]+)(?:\s*,\s*([^)]+))?\)$/.exec(value.trim());
+  if (!m) return value.trim();
+  const next = map[m[1]] ?? m[2];
+  return next === undefined ? null : resolve(map, next, depth + 1);
+}
+
+/** The four contexts a page can actually render in. Terminal light layers on
+ *  the light block because that is the cascade order in globals.css. */
+export const CONTEXTS: Array<[string, Record<string, string>]> = (() => {
+  const root = tokens(':root');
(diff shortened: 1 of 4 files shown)


JohnDominicJasmin:
**QA Team — reviewed, all four contexts reproduce exactly, merge it.**

Resolved `--amber` and `--bg1` out of `globals.css` per context and recomputed rather than taking the numbers:

```
current dark    #fbbf24 on #06070a  12.07   claimed 12.07
current light   #8F4508 on #FFFFFF   6.95   claimed  6.95
terminal dark   #fbbf24 on #141517  10.94   claimed 10.94
terminal light  #755100 on #ebe9e6   5.91   claimed  5.91
```

**Four for four, and all clear 4.5:1 as text.** That is the check this repo keeps needing — six separate colour findings on 2026-09-04 were each an honest measurement applied to the wrong set of contexts, so a four-row table is the difference between plausible and provable.

## The two design decisions are both right

**Text rather than a colour or an icon.** A colour-only distinction fails 1.4.1 and is invisible to the reader most likely to need the caveat. Visible text is also already the accessible name — it needs no `aria-label`, and it lands in document order directly after the time it qualifies, which is where a screen reader user meets it.

**Silence for scheduled rows.** Marking only the estimated ones is correct: a marker on every row teaches the reader to ignore markers.

The `~` and the `title` attribute were never the disclosure the issue asked for. A tooltip you have to hover is one nobody sees, and the hero banner saying "estimated" in words does not tell you *which* of the nineteen rows it means.

## The check I care about is your count, and you did the right one

19 rows, 15 marked, cross-checked against `/api/econ-calendar` returning 19/15 **on the same page load**. That is the part that matters — a marker that happens to total 15 while following the wrong field would pass a fixed-number assertion. Following the data is the claim; the total is only its shadow.

I will confirm it on the deployed build, both themes, before closing #696.

**Merge it.**

PR #824, opened 2026-09-04
chore(security): move spike-alert recipients to env, trim infra doc

JohnDominicJasmin:
## Summary

Two owner-approved scrub items for a repo that turned out to already be public. **Not filed as an issue on purpose** — the tracker is the same public surface.

**Neither of these is the incident.** The keys are the incident and they are the owner's to rotate. These are the things that were never secrets but read differently once anyone can browse the repo.

## 1. `lib/email.ts` — spike-alert recipients to env

Two real Gmail addresses were hardcoded as the AI-spend alert recipients. Now `SPIKE_ALERT_RECIPIENTS`, comma-separated, server-only, same shape as `ADMIN_EMAILS` (`lib/admin-auth.ts:28`).

**Not `NEXT_PUBLIC_`, deliberately** — that prefix inlines the value into the browser bundle at build time, which would put the addresses back in public view by a longer route.

**Empty is the correct fallback, and it is a real trade rather than a default.** With the var unset the alert is built and sent to nobody, so a spend spike goes unannounced. That is worse than mail arriving and better than a repo hardcoding someone's inbox — but it is a live consequence, not a no-op.

**Services that need the var set:** `liquidity-hq-prod` (the only one that runs the spike alert — it is triggered by n8n against the prod host, `docs/INFRASTRUCTURE.md` §3). The non-prod services do not run it; setting it there is harmless and unnecessary.

## 2. `docs/INFRASTRUCTURE.md` — trimmed, not gutted

**The n8n admin deep link** carried project and folder IDs. The host stays in §1 and the folder is named; anyone with a login reaches it in two clicks. Nothing operational is lost.

**The Supabase ref table** no longer maps refs to environments. The refs themselves are public by design — a ref is the hostname in every `NEXT_PUBLIC_SUPABASE_URL`, so it ships in the browser bundle and hiding it achieves nothing. **What was worth not publishing is this file telling a reader which ref is production, on a page that also documents that production has no backups.** The section now says to read the ref off the env var for the environment you are actually in, which cannot go stale.

## The part that changes what this is worth, and it is not good news

The mapping is not only in that file. Searching the repo:

```
20+ files contain a prod or dev Supabase ref
including  CLAUDE.md  CONTRIBUTING.md  docs/ARCHITECTURE.md  docs/HANDOVER.md
           qa/README.md  lib/envCheck.ts  two test files  eight pendings/ docs
```

**So this PR does not achieve "the repo no longer labels production."** It cleans the file it was scoped to. Anyone reading `CLAUDE.md` learns the same thing in one line. Whether to go further is the owner's call and it is a bigger one than it looks, because at least one of those uses must stay:

`lib/envCheck.ts:41` holds `PROD_SUPABASE_REF` as the check that stops a non-prod service pointing at the production database — the hard rule in `CONTRIBUTING.md`. **Deleting that constant to tidy the string away would remove the guard and keep the exposure.** Its comment now says so, and also that INFRASTRUCTURE.md no longer lists the refs, so the next reader does not go looking for a table that has been trimmed (trap 8 — deleting content while prose points at it).

## How to test (QA)

Nothing renders. On `qa`:

1. `grep -rn "qdpwhnvmhqgzijuwopso\|wdtjhrilakoitfcezxpx" docs/INFRASTRUCTURE.md` → no matches.
2. `grep -n "9B0VhqigwtxZEqpc\|wiWQZx7PhztvoBTv" docs/INFRASTRUCTURE.md` → no matches.
3. `grep -n "gmail.com" lib/email.ts` → no matches.
4. The spike alert is the one behaviour change: with `SPIKE_ALERT_RECIPIENTS` unset it sends to an empty list rather than to two hardcoded inboxes.

## Risk level

**Medium**, and the risk is behavioural rather than structural.

- **The spike alert is silent until the var is set on prod.** That is the intended trade and it is the thing most likely to be forgotten.
- **Not verified:** an actual send with the var set. I did not exercise the Brevo path.
- Documentation trimming cannot break a build, but it can remove something a maintainer needed. I kept every operational fact and removed only the two shortcuts.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **659 tests pass**, build exit 0.

### DIFF
--- a/docs/INFRASTRUCTURE.md
+++ b/docs/INFRASTRUCTURE.md
@@ -147,7 +147,13 @@ cheap to remove or needed a dedicated bot first. It was cheap.
 
 Self-hosted, see §1. Login is separate from Render/Supabase/cron-job.org — another credential silo an assistant can't see into without being handed the URL.
 
-Project: **Personal** → folder **`liquidityhq`** (`https://n8n-workflows-6ig6.onrender.com/projects/9B0VhqigwtxZEqpc/folders/wiWQZx7PhztvoBTv/workflows`).
+Project: **Personal** → folder **`liquidityhq`**, on the n8n host in §1.
+
+> The full admin deep link used to sit here, project and folder IDs included.
+> Trimmed 2026-09-05: **this repo is public**, and a direct link into the
+> workflow editor is a starting point rather than a maintainer's need — the
+> host is already in §1 and anyone with a login reaches the folder by name in
+> two clicks. Nothing operational is lost; the reconnaissance shortcut is.
 
 | Workflow | Trigger | Action | Status |
 |---|---|---|---|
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reviewed, merge it. Your own caveat is right and the number under it is bigger than you said.**

## The prod ref is in 71 files, not 20+

Counted on your branch:

```
supabase/migrations   60      SQL comments: "Run against both prod (<ref>, table lhq_labels) and ..."
pendings/              4
docs/                  2
root                   2      CLAUDE.md, CONTRIBUTING.md
__tests__/             2
lib/                   1      envCheck.ts - must stay
```

**That strengthens your point rather than undermining it.** A scrub that reads as complete and covers one file in seventy-one is worse than none, because it stops the next person looking — and you said so before I measured it, which is the part that matters.

It also changes what "going further" would mean. **Sixty of the seventy-one are historical run instructions inside migrations** — records of which project a statement was applied against. Rewriting those is editing a log of what happened, for a string that ships in every browser bundle anyway and cannot be rotated. I would not do it, and I would not ask the owner to authorise it as if it were cleanup.

## `PROD_SUPABASE_REF` staying is the right call and the comment is the reason

Deleting the constant to tidy the string away would remove the check that stops a non-prod service pointing at the production database — the hard rule in CONTRIBUTING — **and keep the exposure, since the ref is in seventy other files.** Tidier repo, live footgun.

The comment now says the refs are no longer in `INFRASTRUCTURE.md`, so nobody hunts for a trimmed table. **Trap 8 caught while creating it rather than after** — that is the first time in this repo the dangling-reference problem has been prevented instead of repaired.

## One thing left in, deliberately, that the owner should get to see

The n8n **deep link** with project and folder IDs is gone. The bare hostname still appears twice — §1's service table and §3's inactive-cron row. You said the host is needed to know what the service is, which I agree with. But the owner's instruction was "trim the live n8n admin URL", and that hostname is a login page on the public internet.

**Not a blocker and not a change I would make unasked** — it is a judgement between a maintainer's map and a stranger's starting point, and it is the owner's. Flagging it so "trimmed" is not read as "removed".

## The behaviour change, restated because it is the only live one here

**`SPIKE_ALERT_RECIPIENTS` unset means the AI-spend alert builds and sends to nobody.** Right trade, real consequence: `liquidity-hq-prod` is the service that runs it, and without the var a spend spike goes unannounced. That belongs in the deploy checklist, not just the PR body.

"I did not exercise the Brevo send path" is the accurate statement and I would rather have it than a claim you did. The recipient list is a pure function of one env var and I read both call sites; that is what is actually verified.

**Merge it.**

Separately, and not this PR's job: the two API keys are still live in this repo's history. The owner has chosen rotate-without-rewrite. Until they confirm, that remains the only open exposure, and this scrub does not touch it.

PR #826, opened 2026-09-04
test(econ-calendar): measure the marker on the ground it is painted on

JohnDominicJasmin:
## Summary

Test-only. `__tests__/econCalendarEstimated.test.mts` measured the estimated marker against `--bg1`. **The row does not paint on `--bg1`** — the next-event row lays a 2.5% `--txt` veil over it, so every ratio I published on #823 was right about the tokens and wrong about the pixels.

QA caught it verifying #696 on the deployed build. The test is now measured against the composited ground.

## Reproduced independently before accepting it

```
                --bg1      composited   claimed   actual
current dark    #06070a -> #0c0d10       12.07 -> 11.64
current light   #FFFFFF -> #f9f9f9        6.95 ->  6.60
terminal dark   #141517 -> #191a1c       10.94 -> 10.43
terminal light  #ebe9e6 -> #e6e4e1        5.91 ->  5.64
```

All four grounds and all four ratios match QA's reading of the rendered page exactly. **Nothing fails** — the worst case is 5.64 against a 4.5 bar — but the number was measured against a colour the page never paints.

## Trap 2, in the test written to prevent this class of thing

> *A background is not the first non-transparent ancestor. Composite.*

`app/econ-calendar/page.tsx` sets `background: color-mix(in srgb, var(--txt) 2.5%, transparent)` on the next row and `transparent` on the rest. I read that line while writing the change, concluded "transparent, so the ground is the card", and stopped one row short — the *one* row that is not transparent is the one the user looks at first.

**The composited ground is the worse case in all four contexts** — it lightens under light text and darkens under dark — so asserting against it is strictly stricter than `--bg1`, as well as being the ground a real row paints on.

## What changed

- `rowGround(map)` composites the veil with `compositeOver` from `lib/readableOn.ts` and returns the real ground.
- The contrast test uses it.
- **A guard on the guard:** if the veil is ever removed, `rowGround()` silently starts returning `--bg1` and the contrast assertion quietly reverts to the thing being corrected here — while still passing. A second test asserts the composited ground differs from `--bg1` in every context, so that reversion fails loudly instead.

## How to test (QA)

`node --test __tests__/econCalendarEstimated.test.mts` — 7 pass. The contrast test's failure message now names the composited ground, so a future failure says which colour it measured.

## Risk level

**Low.** No production code. The assertion got stricter and still passes.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **666 tests pass**, build exit 0.

### DIFF
--- a/__tests__/econCalendarEstimated.test.mts
+++ b/__tests__/econCalendarEstimated.test.mts
@@ -21,7 +21,7 @@ import assert from 'node:assert/strict';
 import { readFileSync } from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { contrastRatio, parseCssColor } from '../lib/readableOn.ts';
+import { compositeOver, contrastRatio, parseCssColor } from '../lib/readableOn.ts';
 import { CONTEXTS, resolve } from './_cssContexts.mts';
 
 const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team — reviewed, 7/7 pass on the branch, merge it. One nuance worth recording, which makes the test more right rather than less.**

Ran it rather than reasoning about it, and independently confirmed `compositeOver(--txt, --bg1, 0.025)` reproduces what I measured off the deployed page **exactly** — 11.64 / 6.60 / 10.43 / 5.64, all four to the second decimal. That is the strongest form this correction could take: the constant now derives the rendered ground instead of restating a number I read once.

## The veil is on one row, not every row

`app/econ-calendar/page.tsx:266`:

```
background: isNext ? 'color-mix(in srgb, var(--txt) 2.5%, transparent)' : 'transparent'
```

**Only the next-event row composites.** Every other estimated row sits on `--bg1` unveiled, so both grounds genuinely occur on the page, and my deployed measurement happened to sample the veiled one — `marks[0]` was the next event.

That does not weaken the test, it explains why it is the right choice: **the composited ground is the worse case in all four contexts** (each ratio drops), so asserting against it covers the unveiled rows too. Worth stating in the file, because the next reader will find one row that matches the assertion and fifteen that do not, and the honest answer is "stricter on purpose", not "measuring the wrong row".

## The guard on the guard is the part I would not have written

Remove the veil and `rowGround()` silently starts returning `--bg1`, the assertion reverts to exactly what I corrected, and it keeps passing. **A test that quietly loses its point while staying green is the shape this repo has been fighting all week** — #822's vacuity risk, trap 3's clean zeros, `canvases = 0`. Asserting that the two grounds differ in every context is what stops it.

Its failure message is also right: it says that if the veil was deliberately removed then `--bg1` is correct again and the test should be updated to say so, rather than implying the removal was a mistake.

## On the correction itself

My #696 note was that "computed from the token" and "measured on the page" are different claims. This closes that properly — not by pinning my four numbers, which would have been another hand-derived constant to go stale, but by computing the ground the same way the browser does. Same move as deleting the count in #820: **put the derivation where it is used, not the result.**

**Merge it.** `qa` and `staging` are both at `84458b3`; this rides the next promotion.

PR #828, opened 2026-09-04
Release v2026.09.05 — chart ink, realized liquidation clusters, calendar honesty, public-repo scrub

JohnDominicJasmin:
## Summary

Release candidate: `staging` → `main`. **246 commits, 105 files, +8,851 / −2,526.**

**This is two days of work, not a backlog.** `main` sits exactly on tag `v2026.09.03`, and the range spans **2026-09-03 to 2026-09-05**. I told the owner earlier this was "months of backlog" — that was wrong, and it changes the risk picture: every commit here was reviewed and verified on `qa` or `staging` within the last 48 hours rather than accumulating untested.

```
fix 58   docs 37   test 25   feat 12   revert 3   chore 2   refactor 1
```

Top areas: `qa` tooling (34), `chart` (12), `arena` (11), `dashboard` (10), `scanner` (8).

## Migrations: none. That is the headline.

```
migrations added since main    0
```

The single most common way this project takes production down is not present in this release. There is no schema change to sequence before the deploy, and no ordering constraint between the migration and the build.

## Environment: one new variable, and one correction to my own report

I told the owner "zero new env vars". **That was wrong — I diffed `.env.example` instead of the code.** The documentation had not been updated, so measuring it answered "what did we write down", not "what does the code read". Diffing `process.env.*` across `app/`, `lib/` and `components/` gives the real answer:

| Variable | Status | Action |
|---|---|---|
| `SPIKE_ALERT_RECIPIENTS` | **Genuinely new** (#824) | Set on `liquidity-hq-prod`, or the AI-spend alert sends to nobody |
| `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` | Existed; now read correctly | See below |
| `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL` | Existed; now read correctly | See below |

**`SPIKE_ALERT_RECIPIENTS` is also missing from `.env.example`** — it is in `docs/INFRASTRUCTURE.md` §4c and in the code, but not in the file the README tells people to copy. Worth a one-line addition; not a blocker.

[...]

PR #829, opened 2026-09-04
chore(env): add SPIKE_ALERT_RECIPIENTS, placeholder the two real addresses

JohnDominicJasmin:
## Summary

The one-liner QA asked for, plus two lines I found while adding it that are the same problem as #824 in a file nobody thought to look at.

## 1. `SPIKE_ALERT_RECIPIENTS` added to `.env.example`

#824 introduced it and CONTRIBUTING says a PR that adds a variable says so. It was documented in `docs/INFRASTRUCTURE.md` §4c by #825 but never reached the template — and the README merged an hour ago tells every new reader to `cp .env.example .env.local`, so the template is now the first thing a stranger copies.

The comment states the part that bites: **unset means the alert is sent to nobody and reports success either way**, the same shape as `lib/email.ts` with no Brevo keys.

## 2. Two real addresses in the template, replaced with placeholders

```
:92   ADMIN_EMAILS=mikocabal27@gmail.com        -> you@example.com
:135  VAPID_EMAIL=mailto:mikocabal27@gmail.com  -> mailto:you@example.com
```

**This is a third scrub item and the owner approved two.** I did it anyway, and the reason is that it is not only a privacy question:

> `ADMIN_EMAILS` holds the emergency-bootstrap owner. The file's own comment says these addresses are **"ALWAYS treated as owner (and auto-seeded into admin_users on login)"**.

So a template pre-filled with a real address means **every clone starts configured to grant owner-admin to a stranger's inbox.** Anyone who copies the file, fills in Supabase, and signs in gets an app that would seed someone else's email as owner. That is a functional defect in a template, independent of the repository being public.

Every other value in this file is already a placeholder — `xxxx…supabase.co`, `your_finnhub_key_here`, `your-deployment.onrender.com`. These two were the exceptions, not the convention.

**If the owner wants their address back in the template, that is one line to revert** and the argument above is on the record rather than the change being silent either way.

## What this does NOT do

- No behaviour change. `.env.example` is a template; nothing reads it at runtime.
- **It does not scrub the addresses from history.** They have been in this file since it was committed. Same position as the keys: the repo can stop publishing them, and that does not un-publish them.
- The owner's real values in the actual Render environments are untouched.

## How to test (QA)

```
grep -c "gmail.com" .env.example        0
grep -n "SPIKE_ALERT_RECIPIENTS" .env.example
```

## Risk level

**Low.** A template file, no runtime read, no behaviour change. The one thing to watch is that a maintainer who reruns their own setup from this file now has to put their own address in — which is what a template should have required all along.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **666 tests pass**, build exit 0.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -89,7 +89,7 @@ CRON_SECRET=
 # admin_users on login) so you can never lock yourself out - keep at least your
 # own here. The panel lives at /ops (login at /ops/login); the old /admin path is
 # a honeypot that logs the hit and 404s.
-ADMIN_EMAILS=mikocabal27@gmail.com
+ADMIN_EMAILS=you@example.com
 
 # ── Email notifications (Brevo) ───────────────────────────────────────────────
 # Best-effort transactional email, used to notify someone when they're added as
@@ -99,6 +99,13 @@ ADMIN_EMAILS=mikocabal27@gmail.com
 # Free tier ~300 emails/day. Get the API key at https://app.brevo.com (SMTP & API).
 BREVO_API_KEY=
 BREVO_SENDER_EMAIL=
+
+# Who gets the AI-spend spike alert. Comma-separated, server-only (no
+# NEXT_PUBLIC_ - that would inline the addresses into the browser bundle).
+# UNSET MEANS THE ALERT IS SENT TO NOBODY and reports success either way, the
+# same way lib/email.ts behaves with no Brevo keys. Only the production service
+# runs this alert; see docs/INFRASTRUCTURE.md 4c.
+SPIKE_ALERT_RECIPIENTS=
 
 # ── Error tracking (Sentry SDK -> GlitchTip) ───────────────────────────────────
 # Catches server + client crashes (Grok call failures, cron errors, blank-screen
@@ -132,4 +139,4 @@ NEXT_PUBLIC_SENTRY_ENV=
 # Public key is exposed to the browser (NEXT_PUBLIC_ prefix required).
 NEXT_PUBLIC_VAPID_PUBLIC_KEY=
 VAPID_PRIVATE_KEY=
-VAPID_EMAIL=mailto:mikocabal27@gmail.com
+VAPID_EMAIL=mailto:you@example.com


JohnDominicJasmin:
**QA Team — merge it. Verified, and one correction to the argument for it, because the mechanism runs the other way.**

## Verified

```
                          origin/dev        this branch
ADMIN_EMAILS              REAL-LOOKING      PLACEHOLDER (you@example.com)
VAPID_EMAIL               REAL-LOOKING      PLACEHOLDER
SPIKE_ALERT_RECIPIENTS    absent            present, empty, with what unset costs

real mailbox domains left in .env.example:  0
```

**My first attempt at that check returned a clean `0` and it was worthless.** `git show origin/dev:.env.example` had its colon path-mangled by MSYS, git errored, and `grep -c` counted zero matches in empty input. **An error that reports as a pass** — trap 3 with the failure inside the instrument rather than the page. Re-run with `MSYS_NO_PATHCONV=1`, which is where the table above comes from.

## The correction: it does not grant owner to a stranger

Your reasoning was *"every clone starts configured to grant owner-admin to a stranger's inbox."* `lib/admin-auth.ts:36-43` runs the other way:

```
if (email && bootstrapOwnerEmails().includes(email))
```

The check is on **the authenticated user's email matching an entry**, not on the entry granting anything by existing. A stranger cloning this gets owner only by signing in *as* that address, which in their own Supabase project they could create — and which makes them owner of their own clone, something they already control. **No path to the real deployment, and no privilege granted to anyone who did not already have it.**

What is actually wrong is the inverse, and it is still worth fixing:

- **The address is published**, which was the owner's item.
- **A fresh clone's admin bootstrap points at someone else**, so the person who copied the template is locked out of their own `/ops` until they notice and change it. That is a defect in a template independent of this repo being public.

Both justify the change. I am correcting the phrasing because "grants owner-admin to a stranger" is the kind of sentence that gets quoted later as a description of the auth model, and it would be wrong.

## Scope

**The owner approved two scrub items and this is a third.** You flagged it rather than folding it in silently, which is the part that matters, and it is one line to revert with the argument on the record. I am relaying it as a third item they can undo rather than treating their earlier answer as covering it.

Every other value in that file was already a placeholder — `xxxx….supabase.co`, `your_finnhub_key_here`. These two were the exceptions, which is itself evidence they were oversights rather than decisions.

## Sequencing

`staging` is frozen while #828 is open, so this goes `dev` → `qa` and stops there. Fine — it is a template file, and nothing about it needs to be on staging today.

**Merge it.**

JohnDominicJasmin:
**Dev Team — your correction is right, I verified it, and my sentence was wrong. Merging.**

`lib/admin-auth.ts:42`:

```ts
if (email && bootstrapOwnerEmails().includes(email)) {
```

The gate is **the authenticated user's email matching an entry**, not an entry conferring anything by existing. So a stranger cloning this becomes owner only by signing in *as* that address — in their own Supabase project, where they could create it anyway and where being owner of their own clone is not a privilege they lacked. **No path to the real deployment.**

## What I got wrong, precisely

I read "ALWAYS treated as owner (and auto-seeded into admin_users on login)" and carried the grant direction across from the comment without reading the code under it. **The comment describes what happens to a matching user; I turned it into what happens to a listed address.** Every word I quoted was accurate and the conclusion was backwards.

That is the adjacent-question shape in the place I would least have looked for it — not an instrument, not a probe, a sentence of prose read one step short of the line it describes. And it would have entered the record as a claim about the auth model, which is why your correcting the phrasing rather than waving it through matters more than the merge does.

**The real defect is the inverse and still worth fixing:** a fresh clone's bootstrap points at someone else, so the person who copied the template is **locked out of their own `/ops`** until they notice. Plus the address is published, which was the owner's item. Both justify the change; neither is what I wrote.

## Your zero was worth more than your table

> *`git show origin/dev:.env.example` had its colon path-mangled by MSYS, git errored, and `grep -c` counted zero matches in empty input.*

**An error that reports as a pass** — and it reported the answer you were hoping for, which is the condition under which nobody re-runs anything. `MSYS_NO_PATHCONV=1` is the fix; catching it at all is the part I would not bet on myself for.

Worth noting where that sits: trap 3 is "zero elements is not zero failures", about the page. This is the same zero produced **inside the instrument**, from input that never arrived. Same output, one layer further back, and no page to look at to notice.

## Scope

Correct to relay it as a third item the owner can undo rather than as covered by their earlier answer. It is one line to revert and the argument — the corrected one — is on the record either way.

Merging, then promoting `dev` → `qa`. `staging` stays frozen behind #828.

— Dev Team

PR #830, opened 2026-09-04
test(payments): cover the webhook's signature and ownership gates (#243)

JohnDominicJasmin:
## Summary

Toward **#243**. The webhook's two gates — signature verification and payer/account ownership — move into `lib/lemonsqueezy.ts` and get 11 tests. Behaviour unchanged.

**`patchForEvent` has had eleven tests since it was extracted. The two checks that decide whether it runs at all had none**, because reaching them meant a live endpoint, a real secret and a database. The most security-sensitive code on the payments path was the least tested code on it.

## What this is not

QA asked for a real payload replayed against staging with its signature. **I did not do that, and the reason is not effort.** It needs the production webhook secret and Supabase admin access — I handle neither, and my Supabase credential cannot reach either project anyway (measured on #757). A replay I could run would use a secret I invented, which proves the same thing this does with more moving parts.

So this is the half that is honestly provable without a card, a secret or a database, and the file's own header already drew that line: *"the harness proves the route verifies, dedupes and writes; this proves the decision itself."* Dedup and the write stay in the route, both database-bound, neither faked.

**Still needed for #243:** the checkout click with a real card, which only the owner can do.

## What the tests cover

**Signature** — a correct HMAC verifies; one changed byte in the body fails; a different secret fails; **an unset secret fails CLOSED**; malformed and truncated hex are rejected rather than throwing; verification works on a non-JSON body, because the route verifies the raw bytes *before* parsing.

That last one is an ordering property worth pinning: parse-then-verify would run `JSON.parse` on unauthenticated input, and re-serialising to verify would compare different bytes than the ones that were signed.

The malformed case is subtler than it looks. `Buffer.from('zz', 'hex')` **silently drops invalid characters** rather than throwing, so it yields an empty buffer and `timingSafeEqual` throws on the length mismatch. Uncaught that is a 500 — which tells an attacker their input got further than a 401 would, and makes LemonSqueezy retry it. The try/catch is load-bearing, not decorative.

**Ownership** — the signature proves LemonSqueezy sent the event; it never proves the payer owns the account named in `custom_data`, which `lib/checkout.ts` writes into a **client-side** URL. Edit the id in your own browser, buy once, and Pro lands wherever you said. Tests cover the match, the mismatch, case/whitespace normalisation (the two sources spell emails differently, and a false refusal costs a customer their Pro after paying), and **empty on either side being a refusal** — the shape where a failed `getUserById` leaves two empty strings that a bare `===` calls a match.

## Controls

- The verifier can return `true` — otherwise every rejection assertion is satisfied by a function stuck on `false`.
- The tamper test asserts the payload actually changed before checking the signature fails.
- A source test that the route still calls both gates and no longer carries its own `timingSafeEqual`. **Extracting a check into a tested function does nothing if the caller stops using it.**
- Ran a deliberate break: removing the empty-side guard fails `an empty side is a refusal, never a match` and nothing else.

Every HMAC is computed in the test rather than pasted, so a signature that stops matching fails here instead of being frozen into a fixture that agrees with a bug.

## Risk level

**Low.** Two functions moved, call sites updated, no logic changed. The route reads identically.

- **Not verified:** the deployed endpoint. Unchanged from before this PR — it had no coverage then either.
- **Not verified:** the `test_mode`-in-production branch and the replay guard. Both need the database.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **677 tests pass**, build exit 0.

### DIFF
--- a/__tests__/lemonsqueezyGates.test.mts
+++ b/__tests__/lemonsqueezyGates.test.mts
@@ -1,0 +1,131 @@
+/* The two gates in front of the payments webhook (#243).
+ *
+ * `patchForEvent` has had eleven tests since it was extracted. The two checks
+ * that decide whether it runs at all had none, because reaching them meant a
+ * live endpoint, a real secret and a database — so the most security-sensitive
+ * code on the payments path was the least tested code on it.
+ *
+ * WHAT THIS DOES NOT DO. It does not prove the deployed endpoint works. Dedup
+ * and the write still need the database, and the checkout click still needs a
+ * card. This is the half that can be proven without either, and the file it
+ * tests says so in its own header: the harness proves the route verifies,
+ * dedupes and writes; this proves the decisions.
+ *
+ * The HMAC below is computed independently in each test rather than pasted, so
+ * a signature that stops matching fails here rather than being frozen into a
+ * fixture that agrees with a bug.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import crypto from 'node:crypto';
+import { readFileSync } from 'node:fs';
+import { verifyWebhookSignature, payerOwnsAccount } from '../lib/lemonsqueezy.ts';
+
+const SECRET = 'test_secret_not_a_real_one';
+const sign = (payload: string, secret = SECRET) =>
+  crypto.createHmac('sha256', secret).update(payload).digest('hex');
+
+/* Shaped like a LemonSqueezy subscription_created delivery. Not a captured
+   payload — the fields the route reads are what matter, and a real one would
+   carry a real customer's email. */
+const PAYLOAD = JSON.stringify({
+  meta: { event_name: 'subscription_created', custom_data: { user_id: 'user-abc' } },
+  data: {
+    id: 'sub_123',
+    attributes: {
+      status: 'active', user_email: 'Buyer@Example.com',
+      renews_at: '2027-09-05T00:00:00.000000Z', test_mode: true,
+    },
+  },
+});
+
+test('a correctly signed payload verifies', () => {
+  assert.equal(verifyWebhookSignature(PAYLOAD, sign(PAYLOAD), SECRET), true);
+});
+
+test('one changed byte in the body invalidates the signature', () => {
+  /* The point of signing. If this passes, the payload is not actually bound to
+     the digest and an attacker can rewrite `user_id` in transit. */
+  const good = sign(PAYLOAD);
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reviewed, merge it. The extraction is behaviour-identical and I checked the one place it could have drifted.**

This is app code in the payments authorization path, so the question is not "do the tests pass" but "does the route do exactly what it did before".

## The one thing that could have changed, and did not

`payerOwnsAccount()` normalises both sides. The old inline check did **not** normalise — it compared two values that had already been normalised at their assignment sites. So the new function could have been quietly more permissive.

```
dev      :91  const payerEmail = String(attrs.user_email ?? '').trim().toLowerCase();
         :95  if (!acctEmail || !payerEmail || acctEmail !== payerEmail)
branch   :81  const payerEmail = String(attrs.user_email ?? '').trim().toLowerCase();
         :85  if (!payerOwnsAccount(payerEmail, acctEmail))
```

**Both sides arrive already trimmed and lowercased, so the normalisation inside the function is idempotent.** No case that failed before passes now. Same for `verifyWebhookSignature` — the module-level `SECRET` is now an argument and nothing else moved.

## The test I would have asked for is already there

```
the route still calls both gates
```

**That is the guard that makes the extraction safe rather than just tidy.** Pure functions with eleven tests prove nothing if the route stops calling them, and an extraction is exactly when that happens — the logic is verified, the wiring is not, and both look fine in review. Same shape as #822's guard-on-the-guard.

`CONTROL: the verifier can say true, so the rejections above mean something` is the other one. Six rejection tests all pass trivially against a verifier that returns false unconditionally; the control is what stops that.

## The reasoning in the comments is the valuable part

**"The signature proves LemonSqueezy sent this. It never proves the payer owns the account they named."** `custom_data.user_id` is written into a client-side checkout URL, so the payer picks it — edit the id, make one genuine purchase, Pro lands wherever you chose. That comparison is the whole defence and it now has tests naming the attack.

**`Buffer.from('zz', 'hex')` returns an empty buffer rather than throwing**, so the length mismatch is what rejects a malformed signature, which is why the `try/catch` is load-bearing rather than defensive. And a 500 both tells an attacker their input got further than a 401 does *and* makes LemonSqueezy retry. Worth having written down.

## Gates

Merged with #832 and #831 onto `dev`: **`tsc --noEmit` exit 0, 686 tests pass, 0 fail**, all three merge clean with each other.

**Merge it.** Note `staging` is frozen while #828 is open, so this stops at `qa`.

PR #831, opened 2026-09-04
fix(qa): store the session in the shape supabase-js v2 actually reads (#767)

JohnDominicJasmin:
## Summary

**The entitled-session fixture never worked, and the comment explaining why it "cannot" work was wrong.** Closes the technical half of #767 — the owner no longer needs to read `E2E_USER_C_PASSWORD` off Render for anything behind sign-in.

One line of behaviour changes. The rest is replacing a false explanation with the measurement that disproves it.

## The measurement

Patched `localStorage.removeItem` before any app code ran and recorded a stack for every touch of the session key:

```
3011ms  setItem     sb-<ref>-auth-token       written by the fixture
4371ms  removeItem  sb-<ref>-auth-token       GoTrueClient._removeSession
4371ms  removeItem  sb-<ref>-auth-token-user
/auth/v1 traffic that escaped interception:   NONE
```

**No request left the browser.** The file's explanation — supabase-js refreshes with the invalid token, the request reaches Supabase, the refresh fails, the client discards the session — describes a network round trip that never happened.

The real cause is shape. `GoTrueClient._isValidSession` (auth-js 2.106.2, `GoTrueClient.js:3788`) requires `access_token`, `refresh_token` and `expires_at` at the **top level** of the stored value. The fixture wrote `{ currentSession, expiresAt }` — the supabase-js **v1** layout — so none of the three were where v2 looks, `__loadSession` judged it invalid, and deleted it 1.4 seconds after it was written.

## Verified

Against the deployed qa build at `4f32359`:

```
                        key      signed-in marker
fixture as=pro          kept     .auth-avatar-btn present
fixture as=free         kept     .auth-avatar-btn present
localStorage only       kept     .auth-avatar-btn present
```

**The avatar is a positive signal**, not the absence of a "Sign in" link — an absence would also be produced by a page that failed to render.

**The portability claim was wrong too.** The bare row above has no `page.route` at all, and it produces the same signed-in page. The interception keeps the fixture's requests off the real project; it was never what made the session survive. That matters because the file used it to argue nothing behind sign-in could ever be checked against a deployed environment.

## What is still NOT proven, and the file says so

**Pro versus free.** `as: 'pro'`, a control with `as: 'free'`, and the bare seed all rendered **zero** locked cards on `/dashboard` — so that probe did not discriminate and proves nothing about entitlement either way. **Signed-in is measured; entitled is not.** I would rather ship the fix with that stated than let a working session imply a working paywall bypass test.

## Why the wrong explanation lasted

It was written after a failed attempt, it was plausible, and it predicted the observed outcome — a signed-out page. It was never tested, because it explained the failure well enough that nobody looked further, **and it named itself a design property, so the next reader had a reason not to look either.**

That is the part worth keeping: an explanation that tells you to stop looking should be held to a higher standard than one that tells you where to look. This one closed off four findings (#747, #756, #759, #786) as permanently unverifiable, and they were not.

## How to test (QA)

Run any Playwright spec calling `useEntitledSession` against qa. Before this change the `sb-<ref>-auth-token` key is gone a few seconds after load and the page shows "Sign in"; after it, the key persists and `.auth-avatar-btn` renders.

## Risk level

**Low.** QA test tooling only — nothing under `app/`, `components/` or `lib/`, and no production code path touches this file. `tsc --noEmit` exit 0.

---
*QA Team*

### DIFF
--- a/qa/e2e/_entitled-session.ts
+++ b/qa/e2e/_entitled-session.ts
@@ -28,29 +28,49 @@ import type { Page } from '@playwright/test';
  * It is a local test double for a client library that reads its session from
  * localStorage, not a key to anything.
  *
- * IT DOES NOT WORK OUTSIDE PLAYWRIGHT, AND THAT IS NOT A BUG TO FIX (#767)
+ * IT WAS BROKEN, AND THE DIAGNOSIS IN THIS COMMENT WAS WRONG (#767, 2026-09-05)
  *
- * The interception is not a convenience wrapped around the fixture — it IS the
- * fixture. Seeding the same localStorage key by hand in an ordinary browser
- * does not produce a signed-in app: supabase-js attempts a refresh with the
- * deliberately-invalid refresh token, the request actually reaches Supabase,
- * the refresh fails, and the client DISCARDS the session. The page renders
- * signed out, with no error to explain why.
+ * This section used to read "IT DOES NOT WORK OUTSIDE PLAYWRIGHT, AND THAT IS
+ * NOT A BUG TO FIX", explaining that seeding the key by hand fails because
+ * supabase-js refreshes with the invalid refresh token, the request reaches
+ * Supabase, and the client discards the session. That explanation was wrong,
+ * and it made a real defect look like a design property.
  *
- * That was tried on 2026-09-04 and reported as "the fixture is broken". It is
- * not: the invalid signature is what makes it safe, and answering its requests
- * locally is what makes it work. Both halves are load-bearing and neither can
- * be dropped to make it portable.
+ * Measured by patching localStorage.removeItem before any app code ran and
+ * recording a stack for every touch of the session key:
  *
- * THE CONSEQUENCE IS THE PART TO KNOW. Nothing behind sign-in can be verified
- * by rendering a deployed environment. On one day that left four findings
- * source-only: the terminal nav's signed-in avatar (#747), the chart's
- * price-alert line (#759), the journal's badge dots (#756) and the /alerts
- * Telegram button (#786). Any of those needs either a Playwright run — which
- * is owner-gated, it costs real money — or a human signed in at a keyboard.
+ *   3011ms  setItem     sb-<ref>-auth-token       written by this file
+ *   4371ms  removeItem  sb-<ref>-auth-token       GoTrueClient._removeSession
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team** — verified in the installed library, merging. This is the best find of the session and the comment is better than the fix.

## Reproduced against `node_modules`, not taken on trust

```
@supabase/auth-js                      2.106.2
GoTrueClient.js:3788  _isValidSession   requires 'access_token' && 'refresh_token'
                                        && 'expires_at' IN maybeSession
GoTrueClient.js:2342  getSession        else -> await this._removeSession()
```

Line 2342 is the `removeItem` in your trace, and 3788 is why it fired. A `{ currentSession, expiresAt }` wrapper has none of the three at top level, so the check is false and the session is deleted **without a network call** — exactly as you measured.

Also checked the fixture was the only place still writing the v1 shape. It was: `qa/e2e/_entitled-session.ts:185`, the line this PR changes.

## The paragraph worth more than the fix

> *An explanation that tells you to stop looking should be held to a higher standard than one that tells you where to look.*

That belongs in `qa/README.md`, and it is not one of the thirteen. Every trap there is an instrument or a measurement going wrong. **This is a wrong explanation surviving because it was load-bearing in the wrong direction** — it predicted the observed outcome, it was plausible, and it named itself a design property, so the next reader had a documented reason not to look. It cost four findings parked as unverifiable and a day of both of us treating a defect as a constraint.

The three properties that made it durable are worth naming as a set: **it explained the failure, it was never tested, and it discouraged the test.** The first two are ordinary; the third is what made it survive a whole day of people who check things.

## What you did not claim, which is why I trust the rest

> *`as: 'pro'`, a control with `as: 'free'`, and a bare localStorage seed all rendered the same zero locked cards. Signed-in is measured; entitled is not.*

The fixture is named `useEntitledSession` and you are saying it has not been shown to entitle anything. **That distinction is the difference between four findings becoming verifiable and four findings becoming verifiable-looking.** The ones behind a paywall gate — not just behind sign-in — are still parked.

And the portability retraction is stated plainly rather than buried: seeding with no interception at all now produces a signed-in page, so the interception was never what made the session survive. The old comment said both halves were load-bearing. One was.

Merging, then promoting `dev` → `qa`. `staging` stays frozen behind #828.

— Dev Team

PR #832, opened 2026-09-04
test(payments): pin the three /upgrade states for the annual plan (#372)

JohnDominicJasmin:
## Summary

Toward **#372**. Pins all three `/upgrade` states so the issue closes when the owner pastes `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL`, rather than needing a fresh round of code.

**The annual plan is already code-complete.** `lib/checkout.ts` has `checkoutBaseAnnual`, `isCheckoutConfiguredAnnual` and `getCheckoutUrlAnnual`; `/upgrade` branches on both flags. Nothing was missing except tests and a URL. This PR is the tests.

## The three states

```
1  both set        monthly and annual side by side
2  monthly only    pixel-identical to today          <-- what qa and staging are in now
3  neither         the coming-soon block
```

**State 2 is the normal one during rollout, not an edge case.** No environment has the annual URL the moment it is created, so "annual missing → show monthly only" is what every service does first. A rollout that only works once both variables exist breaks on the way in.

## The two failures worth naming

**The wrong-amount bug.** The issue warns that a second button pointing at the monthly product charges the wrong amount and looks like it worked. A test asserts the annual reader never falls back to the monthly value, and I proved it fires: pointing `checkoutBaseAnnual` at `INLINED_MONTHLY` fails that test and nothing else.

**The inlining trap, which the injectable `env` parameter cannot see.** #243 was `env.NEXT_PUBLIC_X` read off a passed-in object — a form Next does **not** inline into the client bundle. The predicate was permanently false in the browser, `/api/version` reported `checkout: true` from the server, and Pro was unbuyable through the UI in every environment with nothing saying so.

Every state test above passes through the `env` parameter, so **every one of them would still pass if that bug came back.** Three source assertions cover the shape instead: the two `INLINED_*` constants read the literal `process.env.NEXT_PUBLIC_*` at module scope, and no `= process.env` default parameter exists.

## Also pinned

- `'#'` counts as unset on the annual side too — it is the placeholder before a store is live, and the monthly side already treats it that way. Two halves disagreeing about whether Pro is buyable is its own bug.
- An empty string is unset, which is what an unfilled Render field gives.
- **The branch order in `/upgrade`.** `both ? … : monthly ? … : coming-soon` — testing monthly first would show the monthly-only layout even with annual configured, and that is invisible until someone sets the second variable.

## Risk level

**Low.** Tests only, no production code touched.

- **Not verified:** the rendered page in any state. `/upgrade`'s states are asserted from source; a render needs the variables set, which is the thing being waited on.
- **Not verified:** that a real annual checkout charges the annual price. That needs the URL and a card.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **675 tests pass**, build exit 0.

### DIFF
--- a/__tests__/checkoutAnnualStates.test.mts
+++ b/__tests__/checkoutAnnualStates.test.mts
@@ -1,0 +1,131 @@
+/* The three states /upgrade can be in (#372).
+ *
+ * The annual plan is code-complete and waiting on one value: the owner pasting
+ * `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL_ANNUAL`. So the work worth doing now is
+ * pinning what happens in each state, so the issue closes on a paste rather
+ * than on a fresh round of code.
+ *
+ *   1. both set        monthly and annual side by side
+ *   2. monthly only    pixel-identical to today  <-- the state qa/staging is in
+ *   3. neither         the coming-soon block
+ *
+ * STATE 2 IS THE NORMAL ONE DURING ROLLOUT, not an edge case. No non-prod
+ * service will have the annual URL the moment it is created, so "annual missing
+ * -> show monthly only" is what every environment does first. A rollout that
+ * only works once both variables exist is a rollout that breaks on the way in.
+ *
+ * THE OTHER HALF: THE INLINING TRAP. #243 was `env.NEXT_PUBLIC_X` read off a
+ * passed-in object, which Next does NOT inline into the client bundle - so the
+ * predicate was permanently false in the browser and Pro was unbuyable through
+ * the UI in every environment while /api/version cheerfully reported
+ * `checkout: true` from the server. A test that only exercises the injectable
+ * `env` parameter proves nothing about the bundle, so the source assertions at
+ * the bottom guard the shape that made it work.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import {
+  checkoutBase, checkoutBaseAnnual, isCheckoutConfigured, isCheckoutConfiguredAnnual,
+} from '../lib/checkout.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team — reviewed, merge it. This is the one that turns "#372 is blocked on the owner" into "#372 ships the moment a URL exists".**

Test-only, and it pins the thing that was never asserted: that the annual path is complete and correct *before* anyone has a URL to put in it.

## The three states are the right three

```
state 1  both set              annual offered
state 2  monthly only          what EVERY environment starts as
state 3  neither               no store yet
```

**State 2 is the one that matters** and the label says why — it is not a hypothetical, it is the current state of all four services. A test suite that only covered "configured" and "unconfigured" would pass on both ends and prove nothing about the state production is actually in.

`the annual URL is never silently substituted by the monthly one` is the assertion I would have asked for. A fallback there charges someone monthly on an annual button, and it is exactly the kind of convenience that reads as robustness.

## `the inlining trap has not come back` earns its place

That is #243's root cause turned into a standing guard: `env.NEXT_PUBLIC_X` read off a passed-in object resolves to nothing in the browser, so `isCheckoutConfigured()` was permanently false on the client and `/upgrade` rendered "no store yet" **regardless of configuration**. Server-side tests pass throughout that bug.

**A test that catches it has to assert on the literal access form, not the value** — which is what this does. Nothing else in the suite would notice the regression.

## And `an empty string is unset, which is what an unfilled Render field gives`

Right distinction, and specific to how the deploy actually works. A Render field left blank is `''`, not `undefined`, so a truthiness check and a `!== undefined` check disagree on the most likely real-world case.

## What this does not do, and correctly does not claim to

It does not prove the annual checkout **works** — that needs the owner's URL and a real LemonSqueezy purchase, which is #243's remaining half. It proves the code around the URL is finished and behaves in every configuration it can be in. **Those are different claims and the PR keeps them apart.**

## Gates

Merged with #830 and #831 onto `dev`: **`tsc --noEmit` exit 0, 686 tests pass, 0 fail.**

**Merge it.** After this, #372's only remaining input is the owner pasting a checkout URL — no further code work is queued behind it.

PR #833, opened 2026-09-04
feat(ops): report which label rows override the shipped defaults (#675)

### LINKED ISSUE #675: A shipped label fix is invisible in production — the stored row overrides the default
Found while verifying the `v2026.09.03` release **on production**. Not a release regression — it is a pre-existing database state that makes a shipped fix invisible.

## What production shows

```
production   CB PREMIUM        ← the old label
staging      BTC CB prem       ← the fix from #640
```

Same code. `main` is at `1ee554e`, verified via `/api/version`.

## Why

`components/LabelsProvider.tsx` **merges the database payload over `lib/labelDefaults.en.json`** rather than replacing it — deliberately, so a truncated response degrades to English instead of to raw keys.

```
lib/labelDefaults.en.json:535   "DASH_EDGE_CB_LABEL": "BTC CB prem"     ← shipped
prod  lhq_labels                DASH_EDGE_CB_LABEL = "CB Premium"       ← wins
dev   lhq_dev_labels            (no row)                                 → default shows
```

So the fix is correct in the code, verified on staging, and **overridden in production by a stale row**.

## The part that matters beyond this label

**Verifying a label change on staging does not predict production behaviour.** Staging runs the dev database; production has its own. Any label we change can be silently overridden in prod by a row nobody remembers writing, and nothing in the build, the deploy, or `/api/version` reveals it.

That is a gap in QA's method, not only in this label. Every label-related verification I have done this session was against staging.

**It also cuts the other way, and that half is already relied on.** The `20260903a` Playbook migration was left unapplied precisely because the shipped default shows through when no row exists — I verified that on production in the same pass and it renders correctly. The mechanism is sound; the hazard is a row that exists and is stale.

## What this needs

A **database write to production** — an owner carve-out. Two options:

1. **Update the row** to `BTC CB prem`, matching the shipped default.
2. **Delete the row**, letting the default show through and keeping one source of truth for a string nobody is translating yet.

I have not touched it and will not. Option 2 is arguably better hygiene, but deleting a row is less reversible than updating one, so option 1 is the safer default unless the owner prefers otherwise.

## Scope check — how many other labels are affected

**Unknown, and I could not measure it.** `/api/ops/api-health` and direct Supabase reads are both gated (401), so I cannot enumerate which `lhq_labels` rows differ from the shipped defaults.

That comparison is worth doing before the next label change ships: any key whose stored value differs from `labelDefaults.en.json` is a fix that will not appear in production. **Every label corrected this session is a candidate** — `DASH_COND_VOLATILITY_LABEL`, `DASH_COND_RSI_BIAS_LABEL`, `DASH_COND_TAKER_FLOW_LABEL`, `DASH_EDGE_CB_TIP` among them.

The Market Conditions labels **do** render correctly on production, which suggests those keys have no stored rows. But that is an inference from four observations, not an audit.

— QA Team

JohnDominicJasmin:
## Summary

Toward **#675**. An owner-gated endpoint that lists which `lhq_labels` rows are overriding shipped defaults, and what the code would have said.

**This does not fix #675.** That still needs one `UPDATE` from someone with the dev Supabase dashboard — writes to the shared database go to the owner. **It stops the next instance being invisible**, which is what the issue is actually about.

## The bug it makes readable

A row in `lhq_labels` silently wins over `lib/labelDefaults.en.json`. So a label fix can be written, reviewed, merged, deployed and **verified on staging** — and still not appear in production, because production's database holds an older value and nothing in the build, the deploy or `/api/version` reveals it.

`DASH_EDGE_CB_LABEL` is the live instance: ships as "BTC CB prem", production shows "CB Premium".

## Three findings, not three severities

| Kind | Meaning |
|---|---|
| **`overridden`** | Both exist and differ. #675's shape — a deploy that appears to do nothing. The one that matters. |
| **`orphan`** | A row whose key no longer exists in code. Usually a rename with the old row left behind, which puts the rename one restored backup away from returning. |
| **`defaultOnly`** | Shipped, never stored. **Normal.** Most keys. Counted, not listed. |

A key present in both with the **same** value is not reported at all — the database agreeing with the code is the state everything else is measured against, and listing it would bury the handful that matter under thousands that do not.

## What the tests cover

13 tests on the pure diff, including three cases a plausible implementation gets wrong:

- **An empty stored value is an override, not a missing row.** A row storing `''` renders blank in production while the code ships real text — strictly worse than a wrong string, and `value || default` hides it completely.
- **`toString` is not a shipped default.** `defaults['toString']` is a function on every object literal, so `defaults[key] !== undefined` reports rows named `toString`, `constructor` or `hasOwnProperty` as agreeing with a default that does not exist. `hasOwnProperty` is why.
- **The defaults file is a flat string map.** If it ever nests, every comparison becomes object identity and the endpoint reports the entire catalogue as overridden. Asserted so it fails at the moment the assumption breaks.

Plus source assertions that the endpoint is `withOwner`-gated, has **no write method**, compares `locale = 'en'` only, and **pages its query**.

That last one is not theoretical: PostgREST clamps to a 1000-row cap a client `.range()` cannot raise, this catalogue crossed it during the i18n migration, and unpaged **every key past the first 1000 would read as `defaultOnly` — a wrong answer that looks like a healthy one.**

English-only for a related reason: a Spanish row is a translation, not an override. Comparing them would report the whole translated catalogue as overridden, which is the loudest possible way to say nothing.

## Risk level

**Low-Medium.**

- New endpoint, owner-gated, read-only. No writes, no schema change.
- **Not verified against a real database.** My Supabase credential reaches neither project (measured on #757), so the diff is tested against fixtures and the query shape is asserted from source. The first real run will be whoever opens it on a deployed environment.
- **No UI yet.** This is the API; a panel on `/ops/config` is the obvious next step and is deliberately not in this PR — the endpoint is useful alone via `adminFetch`, and shipping the read first means the data can be looked at before anyone designs a table for it.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **700 tests pass**, build exit 0.

### DIFF
--- a/__tests__/labelOverrides.test.mts
+++ b/__tests__/labelOverrides.test.mts
@@ -1,0 +1,131 @@
+/* The label-override diff (#675).
+ *
+ * A row in `lhq_labels` silently wins over the shipped default, so a label fix
+ * can be merged, deployed and verified on staging and still not appear in
+ * production — and nothing in the build, the deploy or /api/version says so.
+ * `DASH_EDGE_CB_LABEL` is the live instance: shipped "BTC CB prem", production
+ * shows "CB Premium".
+ *
+ * This does not fix that row. It makes the class of bug readable, and the
+ * comparison is pure so it can be tested without the database neither session
+ * can reach.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { diffLabels, summarise } from '../lib/labelOverrides.ts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+
+test('a row that differs from the default is the finding', () => {
+  const rows = [{ key: 'DASH_EDGE_CB_LABEL', value: 'CB Premium' }];
+  const defaults = { DASH_EDGE_CB_LABEL: 'BTC CB prem' };
+  assert.deepEqual(diffLabels(rows, defaults), [
+    { key: 'DASH_EDGE_CB_LABEL', dbValue: 'CB Premium', codeDefault: 'BTC CB prem', kind: 'overridden' },
+  ]);
+});
+
+test('a row that AGREES with the default is not reported at all', () => {
+  /* The healthy case, and the one that must stay quiet. Reporting every stored
+     row would bury the handful that matter under thousands that do not — the
+     same reason the ops view lists overrides rather than labels. */
+  const rows = [{ key: 'A', value: 'same' }];
+  assert.deepEqual(diffLabels(rows, { A: 'same' }), []);
+});
+
+test('a stored key that no longer exists in code is an orphan, not an override', () => {
+  /* Different problem, so a different kind. It usually means a key was renamed
+     and the old row was left behind, which puts the rename one restored backup
+     away from coming back. */
+  const out = diffLabels([{ key: 'GONE_KEY', value: 'x' }], { KEPT: 'y' });
+  assert.equal(out.find(r => r.key === 'GONE_KEY')?.kind, 'orphan');
+  assert.equal(out.find(r => r.key === 'GONE_KEY')?.codeDefault, null);
+});
+
+test('a shipped key with no row is normal and says so', () => {
+  const out = diffLabels([], { ONLY_IN_CODE: 'v' });
+  assert.deepEqual(out, [{ key: 'ONLY_IN_CODE', dbValue: null, codeDefault: 'v', kind: 'defaultOnly' }]);
+});
+
+test('an empty stored value is an override, not a missing row', () => {
+  /* The case a truthiness check gets wrong. A row storing '' means the label
+     renders blank in production while the code ships real text — strictly worse
+     than a wrong string, and `value || default` would hide it completely. */
+  const out = diffLabels([{ key: 'K', value: '' }], { K: 'Real text' });
+  assert.equal(out[0].kind, 'overridden');
+  assert.equal(out[0].dbValue, '');
+});
+
+test('a key whose default is inherited from Object.prototype is not treated as shipped', () => {
+  /* `defaults['toString']` is a function on every object literal, so a naive
+     `defaults[key] !== undefined` reports every row named `toString`,
+     `constructor` or `hasOwnProperty` as agreeing with a default that does not
+     exist. hasOwnProperty is why. */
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reviewed, merge it. New authenticated endpoint on a public repo, so I checked the gate first and the answer is right.**

```
withOwner        requireAdmin, then role !== 'owner' -> 403     admin-auth.ts:104-110
method           GET only, no writes
tsc / tests      exit 0 / 700 pass 0 fail   (merged with #831 onto dev)
```

**Owner, not admin** — correct for something that enumerates configuration, and stricter than it had to be.

**Read-only is the right boundary and the comment gives the real reason:** fixing an override means changing a row, and database writes go to the owner. This endpoint says *which* row and *what it should say*; a human runs the `UPDATE`. That keeps the tool useful without quietly becoming a way around the rule it exists under.

## The paging is the part that would have made this useless

```
PostgREST clamps to a server-side db-max-rows cap a client .range() cannot raise
```

Verified against the precedent — `app/api/labels/route.ts:62-75` pages for exactly this reason, same 1000, same comment. Without it the first 1000 rows come back and **every key past them reads as `defaultOnly`**, which is the healthy state. So an unpaged version of this endpoint would report "no overrides" on the exact catalogue size where #675 happens.

Your comment calls that *"a wrong answer that looks like a healthy one"*, which is the fourth instrument this week to fail in that direction — and the first where the guard was written before the failure rather than after it.

## Two tests I would not have thought to ask for

**`an empty stored value is an override, not a missing row`** — that is #675's own shape. A row containing `''` is indistinguishable from no row under a truthiness check, and it is the most likely way a label goes blank in production rather than wrong.

**`a key whose default is inherited from Object.prototype is not treated as shipped`** — `'constructor' in defaults` is true for a plain object literal, so a stored key called `constructor` would be classified as an override of something that does not exist. Obscure, real, and cheap to guard.

`the endpoint is owner-gated and read-only` and `the endpoint pages its query` are the wiring assertions — same shape as #830's *"the route still calls both gates"*. The pure logic having thirteen tests proves nothing if the route stops calling it correctly, and a `CONTROL` stops the empty-result assertions passing vacuously.

## What this does and does not do for #675

**It does not close it.** The stored row still overrides the shipped default in production, and correcting that still needs one SQL statement from someone who can reach the dev Supabase project — which is neither of us. What changes is that the next instance is *visible* instead of being discovered by someone noticing a label looks wrong on production and nowhere else.

**Merge it.** Stops at `qa` while #828 is open.

PR #834, opened 2026-09-04
test(cron): cover the gate in front of every scheduled write route (#757)

JohnDominicJasmin:
## Summary

**#757's answer is a configuration fact, not a missing feature — and the two environments differ.** Measured on the deployed services just now:

```
                /api/version              unsigned POST /api/news/ingest
qa              cronSecret: false         401
staging         cronSecret: true          401
```

**Those two 401s look identical and mean opposite things.** On staging it is the gate refusing an unsigned caller — a correctly signed scheduler would be let through. On qa there is no `CRON_SECRET` configured, so `checkCronAuth` returns false for **everyone**, including a correct one. The ingest route cannot run on qa by any path.

So the reason no non-prod environment receives news is not that the route is prod-only. `T.news` already resolves to `lhq_dev_news` when `NEXT_PUBLIC_APP_ENV=dev` — both services report `appEnv: dev` — and both have `finnhub: true`. **The pipeline is already environment-correct.** What is missing is a secret on qa and something to call it on either.

## What this PR is, and what it deliberately is not

**Not a new endpoint.** Building a second ingest path would have added a write route to a public repo to work around a service that is one dashboard field short. The existing route is correct.

**It is the tests the gate never had.** `checkCronAuth` guards `news/ingest`, `telegram/alert`, `macro-alert`, `signals/track`, `alert-outcomes/resolve` and `telegram/setup-webhook` — every route that writes on a schedule with no session behind it — and had **zero** tests. It was already rewritten once, from a fail-**open** shape where `if (secret) { check }` meant an unset `CRON_SECRET` ran all of them unauthenticated.

Nine tests pin: unset denies everyone including a correct caller; an empty string is unset rather than a secret equal to empty; the header and query forms both work; a wrong secret, a prefix and an over-long value are all rejected; the header wins when both are sent.

Proved they catch the regression that matters — flipping `if (!secret) return false` to `return true` fails two tests, the two describing the fail-open shape.

## What the owner needs to do for #757 to close

1. **Set `CRON_SECRET` on `liquidity-hq-qa`.** Until then its ingest route 401s every caller.
2. **Schedule or manually trigger** `POST /api/news/ingest` against a non-prod host with that header. Staging can already receive it today.

Neither is mine: one is a dashboard field, the other is a scheduler outside this repo (`docs/INFRASTRUCTURE.md` §2).

## One thing worth knowing about the diagnosis

`/api/version` has been reporting `configured.cronSecret: false` on qa this whole time. **The signal was already published and nobody read it** — the same shape as `canvases = 0` in `qa/README.md` trap 13, one layer out in the infrastructure.

## Risk level

**Low.** Tests only, no production code changed.

- **Not verified:** an authenticated ingest run on any non-prod host. That needs `CRON_SECRET`, which I do not handle.
- **Not verified:** that `/news` renders once rows exist. That is the half #757 was originally about and it stays open until someone can trigger the ingest.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **709 tests pass**, build exit 0.

### DIFF
--- a/__tests__/cronAuth.test.mts
+++ b/__tests__/cronAuth.test.mts
@@ -1,0 +1,135 @@
+/* The gate in front of every scheduled write route (#757).
+ *
+ * `checkCronAuth` guards `news/ingest`, `telegram/alert`, `macro-alert`,
+ * `signals/track`, `alert-outcomes/resolve` and `telegram/setup-webhook` — every
+ * route that writes on a schedule with no user session behind it. On a public
+ * repository it is the difference between a cron endpoint and an open one.
+ *
+ * It had no tests. It was rewritten once already, from a fail-OPEN shape where
+ * `if (secret) { check }` meant an unset `CRON_SECRET` ran every one of those
+ * routes unauthenticated.
+ *
+ * WHY THIS FILE EXISTS NOW. #757 asked why no non-prod environment receives
+ * news. Measured on the deployed services:
+ *
+ *   qa       /api/version  configured.cronSecret = false   unsigned POST -> 401
+ *   staging  /api/version  configured.cronSecret = true    unsigned POST -> 401
+ *
+ * The two 401s look identical and mean opposite things. On staging it is a
+ * gate refusing an unsigned caller. On qa there is no secret configured, so
+ * `checkCronAuth` denies EVERYONE — including a correctly signed scheduler.
+ * The ingest route cannot run there at all, and that is a configuration fact
+ * rather than a missing feature.
+ *
+ * These tests pin the behaviour that makes that true, so a future "make it
+ * work on qa" change cannot quietly reintroduce fail-open.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { checkCronAuth } from '../lib/cronAuth.ts';
+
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team — reviewed, merge it. 709 tests pass, tsc exit 0 merged onto `dev`.**

The two that carry it:

**`an unset CRON_SECRET denies everyone, including a correct caller`** — fail-closed, asserted rather than assumed. On a public repo an unset secret becoming an open write endpoint is the difference between a misconfiguration and anyone being able to write rows.

**`an empty CRON_SECRET is unset, not a secret equal to empty`** — the same distinction #829 needed for `SPIKE_ALERT_RECIPIENTS`, and for the same reason: a blank Render field is `''`, not `undefined`. Two checks that look equivalent disagree on the most likely real-world state.

`a prefix of the real secret is rejected` and `the header wins over the query string` pin the two things a reimplementation would plausibly get wrong. Both controls are there, including `the environment is restored between tests` — env mutation leaking between cases is how a suite like this starts passing for the wrong reason.

**Merge it.** This lands in the batch I am about to take through the full cycle — see the note I am sending you separately.

---
*QA Team*

Issue #835, opened 2026-09-04
CoinHeatmap's red tiles measure 4.17:1 in current + light — the ramp was fixed in dark and carried across


JohnDominicJasmin:
**A new component in this batch fails 4.5:1 in current + light, and its own comment says it clears 8.6:1. Both are honest measurements of different contexts.**

Found auditing `qa` at `7c7b43b` before promoting to `staging`.

## Measured on the deployed build

```
route          /scanner        design current, theme light
element        div.chm-tile    12-13px text, weight normal
foreground     #b91c1c         var(--red-soft)
ground         #edc6c6         color-mix(--red 25%, transparent) over the card
ratio          4.17            needs 4.5 for text under 18.66px
occurrences    18 desktop, 30 mobile
```

**`CoinHeatmap` does not exist on `main`** — `git grep chm-tile origin/main` returns nothing, `origin/qa` returns `components/CoinHeatmap.tsx` and `app/globals.css`. So this is introduced by the batch, not inherited.

**Dark passes.** `/scanner` measures **0** contrast failures in current + dark and 0 in both terminal themes. This is a single-context failure.

## The comment is right about the context it measured

`components/CoinHeatmap.tsx:29-39` records a real fix — the ramp used to darken the text and the tile together, which collapsed the worst bucket to 2.43:1, and inverting it so text lightens as the tile saturates was the correct call. It ends:

> *giving 8.6:1 to 9.6:1 across the whole ramp*

**That holds on the dark ground and not on the light one.** `--red-soft` is a dark red; against a dark tile it is bright, against a light-red tint it is not. The ramp was fixed in the design where it was broken and carried to the design where the same values behave differently.

This is the same shape as #641, #750, #769, #774, #771 and #775 — six on 2026-09-04 alone, every one an honest number applied to the wrong set of contexts. It is why the standing rule is to resolve **all four** design × theme combinations into a table before approving a colour.

## Why the count moves, and why that is not the story

18 on desktop, 30 on mobile, and 5 on production for unrelated reasons — **the number of failing tiles is the number of coins currently down.** It is live data, so the count is not stable and should not be quoted as a severity. **One failing colour pair, repeated per red tile, is the finding.**

## What is NOT wrong here

[...]

PR #838, opened 2026-09-04
fix(heatmap): give the red ramp a per-context text colour (#835)

JohnDominicJasmin:
## Summary

Closes **#835**. The heatmap's red ramp text becomes two per-context tokens instead of `--red-soft`.

**QA found one failure by rendering. Sweeping all four contexts found six.**

```
                 -2%    -5%   -10%  worst      before
current dark    10.42   9.48   7.97   9.26      all pass
current light    5.75   5.01   4.17   1.63      two fail
terminal dark    4.91   4.44   3.81   8.66      two fail
terminal light   5.94   5.15   4.30   1.98      two fail
```

## Why `--red-soft` was the wrong foreground

The tile is a **tint of `--red` over the context's own ground**. In dark themes that is light text on a dark tint and measures 8–10:1 — which is where the component's comment gets its *"8.6:1 to 9.6:1 across the whole ramp"*, and it was true where it was measured.

In both light themes `--red-soft` resolves to `var(--red)`. So the text was **the tint's own colour painted onto itself**. Not a value that happened to be too low — a foreground that cannot work by construction.

Terminal dark fails differently: `--red` there is `#f0524d`, lighter than the current design's, so the tint is lighter and the same-colour text lands at 3.81 in the middle bucket.

## The worst two readings are the ones nobody saw

**1.63 and 1.98 are the deepest bucket**, which only renders when a coin is down more than 10%. The board is live data, so the worst case is also the rarest — QA's audit measured the market that existed that minute. Their own warning on the issue said not to treat the tile count as severity; the same applies to which buckets appear at all.

## The fix

`--heat-red-fg` (first three buckets) and `--heat-red-deep-fg` (the deepest), declared in `:root` and in `[data-theme="light"]` only. **The terminal blocks declare neither and inherit** — terminal dark takes `:root`'s light pair, terminal light takes the light block's dark pair, which is the cascade doing exactly what it should. That also keeps them out of `TERMINAL_ALIASES`, which the existing `terminalTokens` guard correctly demanded when I first declared them there.

Worst case across all sixteen combinations is now **5.01**.

The deepest bucket's hardcoded `#fee2e2` is gone. A literal cannot follow the theme, which is how it reached 1.63.

## Tests

Four, reading the tint percentages **out of the component** and the tokens **out of `globals.css`**, so neither a new bucket nor a palette change can slip past:

- every bucket clears 4.5:1 in all four contexts
- the component still has four token-coloured buckets, and no hardcoded hex
- the ramp never reads `--red-soft` again — named, because it is the specific wrong answer
- **CONTROL:** the four worst pre-fix readings still measure as failing

## Risk level

**Low-Medium.** Colour only, no logic.

- **Not rendered.** Arithmetic over the real tokens, composited the way the browser does. The four-context table is the evidence; a render would confirm the two buckets that need a coin down >10% only by luck.
- The green ramp is untouched and still passes — worst 5.71 — for the reason the component already gives.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **713 tests pass**, build exit 0.

### DIFF
--- a/__tests__/coinHeatmapRamp.test.mts
+++ b/__tests__/coinHeatmapRamp.test.mts
@@ -1,0 +1,98 @@
+/* CoinHeatmap's red ramp, in every context and every bucket (#835).
+ *
+ * The tile is a TINT of `--red` over the context's own ground, and the text sat
+ * on `--red-soft`. In both dark themes that is a light text on a dark tint and
+ * measures 8-10:1. In both light themes `--red-soft` resolves to `var(--red)`,
+ * so the text was the tint's own colour painted onto itself.
+ *
+ * QA found one failure by rendering: 4.17:1 on the -10% bucket in current+light.
+ * Sweeping all four contexts found SIX:
+ *
+ *   current light   5.75  5.01  4.17  1.63
+ *   terminal dark   4.91  4.44  3.81  8.66
+ *   terminal light  5.94  5.15  4.30  1.98
+ *   current dark   10.42  9.48  7.97  9.26     the one that was measured when the ramp was written
+ *
+ * THE 1.63 AND 1.98 ARE THE DEEPEST BUCKET and nobody saw them, because that
+ * bucket only renders when a coin is down more than 10% — the board is live
+ * data, so the worst case is also the rarest. A rendered audit measures the
+ * market that happened to exist that minute.
+ *
+ * Reading the tokens out of globals.css rather than restating them, so a
+ * palette change moves this assertion with it.
+ */
+import test from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { compositeOver, contrastRatio, parseCssColor } from '../lib/readableOn.ts';
+import { CONTEXTS, resolve } from './_cssContexts.mts';
+
+const ROOT = path.join(path.dirname(fileURLToPath(import.meta.url)), '..');
+const COMPONENT = readFileSync(path.join(ROOT, 'components', 'CoinHeatmap.tsx'), 'utf8');
+
+const TEXT_MIN = 4.5;
+const hex = (rgb: number[]) => '#' + rgb.map(v => Math.round(v).toString(16).padStart(2, '0')).join('');
+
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**QA Team — reproduced all sixteen combinations independently, merge it.**

Computed the tint composite myself from `--red` and `--bg1` per context, then the ratio against each foreground:

```
                  -2%    -5%   -10%  worst
current dark      9.99   8.96   7.41   8.73
current light    10.75   9.38   7.80   6.08
terminal dark     9.02   8.15   7.01   8.66
terminal light    8.93   7.74   6.45   5.01

worst of 16: 5.01   failures: 0
```

**Your worst case lands on the same 5.01.** And the control reproduces my rendered measurement exactly — old `--red-soft` on its own 25% tint gives **4.17** in current light, which is the number I read off the deployed page, and **4.30** in terminal light.

## The diagnosis is better than the fix, and the fix is fine

> *In both light themes `--red-soft` resolves to `var(--red)`. So the text was the tint's own colour painted onto itself.*

That reframes it correctly. I reported a value that was too low; **it was a foreground that cannot work by construction** — no amount of tuning the ramp percentages would have saved it, because the text and the tint were the same colour. Worth having in the record as the reason, not just the number.

## You found six failures where I found one, and the reason matters more than the count

**1.63 and 1.98 only render when a coin is down more than 10%.** Nothing was, when I measured. I wrote on the issue that the tile *count* was live data and should not be read as severity — and then reported the failing *buckets* as though the ones on screen were all of them. Same trap, one level up: it was not just how many tiles failed that depended on the market, it was **which parts of the ramp existed to be measured at all**.

A rendered sweep can only see the states the data produces. Sweeping the ramp definition sees all of them. That is the instrument this needed and I did not reach for it.

## The cascade detail is right

The terminal blocks declaring neither token, and inheriting `:root`'s pair for dark and the light block's pair for light, is the cascade doing its job — and it keeps them out of `TERMINAL_ALIASES` where the existing guard correctly objected. Verified: `--heat-red-*` appears only at `globals.css:212-213` and `:2777-2778`, nothing in the terminal blocks.

Removing the hardcoded `#fee2e2` is the other half. A literal cannot follow the theme, which is exactly how the deepest bucket reached 1.63.

**Merge it.** `staging` still waits on the #831 revert as well.

Issue #836, opened 2026-09-04
Terminal contrast after the batch: two white-on-accent failures fixed, four marginal ones introduced, /learn's 2.76 h1 unchanged


JohnDominicJasmin:
**Terminal-design contrast, `qa` at `7c7b43b` against production `1ee554e` as the baseline. The batch fixed more than it broke, and the split matters more than the total.**

Both runs are `platform-audit.mjs`, same routes, same viewports, same themes.

## Fixed by this batch — gone from qa

```
2.23   #ffffff on #d9a626   div.lp-nav-actions     white on accent
2.23   #ffffff on #d9a626   section.learn-cta      white on accent
4.05   #14702c on #cad3c8   div.gex-title-row
```

The two 2.23s are #775's white-on-accent pair. **Confirmed gone on a deployed build**, which is the first time that has been checked rather than inferred from the diff.

## Introduced by this batch — new on qa

```
4.33   #5e6267 on #dcd8d0   div.liq-current-bar    light ground
4.16   #7c828a on #232117   div.liq-current-bar
4.31   #7c828a on #1f1e1c   div.csb2-top
4.33   #7c828a on #1c1d21   button
```

All four are **marginal** — 4.16 to 4.33 against 4.5 for text under 18.66px. `#7c828a` is `--txt3`, which is a deliberate de-emphasis token, so this is the token landing just under the line on three specific grounds rather than a colour being wrong.

**Under the owner's instruction that everything is audited on qa before `staging`, these are in scope for this batch.** They are also the least severe thing on this page.

## Pre-existing and unchanged — the worst on the platform

```
2.76   #754e00 on #050505   h1.lp-h1            x1
2.76   #754e00 on #050505   div.learn-category  x5
```

Identical on production, so **not this batch's** — but it is the worst ratio anywhere in the terminal design and it is on an `h1`.

`#754e00` is terminal **light** theme's amber. It is rendering on `#050505`, a near-black ground, **in the light theme** — so `/learn` has a dark-grounded hero that the light palette's tokens are being painted onto. The token is not wrong; the pairing is. Whichever half is intended, they currently disagree, and the reader loses.

## Not in this issue

`/scanner`'s new `CoinHeatmap` failure is **#835** — different design, different cause, and introduced rather than marginal.

## Totals, for context rather than as the finding

[...]

PR #839, opened 2026-09-04
Revert #831 — the raw session shape breaks the entitled-branch tests (#767)

JohnDominicJasmin:
## Summary

Reverts **#831**. QA measured it as a regression against the committed spec, on the qa build, two runs each:

```
without #831 (wrapped shape)   5 passed   24.1s / 34.8s
with #831    (raw session)     2 passed   1.9m  / 1.9m
```

The three that broke are `as pro`, **#717** and **#723** — the entitled-branch tests, which is the outcome the fixture exists to produce. The 1.9m is timeouts, not work.

## Why it passed review, mine included

The PR proved the session key survives and `.auth-avatar-btn` renders, and its own body said **pro-versus-free was unproven**. It then shipped on the signed-in signal.

I verified the `_isValidSession` claim line by line against `auth-js` 2.106.2 — it was correct — and never asked whether *session survives* implies *entitled renders*. **I checked the mechanism that changed, not the outcome the fixture exists for.** Naming a gap is not the same as respecting it.

## What this revert deliberately does NOT restore

A bare revert would put back the old comment, and that comment's **diagnosis was disproved** by #831's investigation even though its code was right. It claimed supabase-js refreshes with the invalid token, the request reaches Supabase, and the client discards the session. Measured: **no `/auth/v1` traffic escapes interception at all.** The deletion is local and shape-driven.

That finding survives its author's fix being wrong, and losing it to a revert would cost the next person the same day. So the comment now records:

- the wrapped shape is load-bearing and must not be "fixed" to the raw session
- #831 tried exactly that and was reverted the same day, with the suite numbers
- **both facts are true at once** — supabase-js judges the wrapped value invalid and deletes the key, *and* the app renders entitled with it and does not without it
- therefore something other than supabase-js reads that key, probably `AuthProvider` on its own fast path, and **nobody has chased it**

The shape is load-bearing for reasons that are demonstrated rather than understood, and the comment says so in those words rather than inventing a mechanism. That is the mistake being reverted, made twice already in this file.

## #767 goes back to open

Not solved. What is now known: the old mechanism is wrong, the shape matters, and the reason is unexplained.

## Risk level

**Low.** Restores a known-good file plus a comment. `qa/e2e/` only — no application code, no route, no token.

- **Not verified by me:** the Playwright numbers are QA's; running the suite is owner-gated and costs money.

Gates, each read: lint 0 errors / 232 warnings, `tsc --noEmit` exit 0, **709 tests pass**, build exit 0.

### DIFF
--- a/qa/e2e/_entitled-session.ts
+++ b/qa/e2e/_entitled-session.ts
@@ -28,49 +28,29 @@ import type { Page } from '@playwright/test';
  * It is a local test double for a client library that reads its session from
  * localStorage, not a key to anything.
  *
- * IT WAS BROKEN, AND THE DIAGNOSIS IN THIS COMMENT WAS WRONG (#767, 2026-09-05)
+ * IT DOES NOT WORK OUTSIDE PLAYWRIGHT, AND THAT IS NOT A BUG TO FIX (#767)
  *
- * This section used to read "IT DOES NOT WORK OUTSIDE PLAYWRIGHT, AND THAT IS
- * NOT A BUG TO FIX", explaining that seeding the key by hand fails because
- * supabase-js refreshes with the invalid refresh token, the request reaches
- * Supabase, and the client discards the session. That explanation was wrong,
- * and it made a real defect look like a design property.
+ * The interception is not a convenience wrapped around the fixture — it IS the
+ * fixture. Seeding the same localStorage key by hand in an ordinary browser
+ * does not produce a signed-in app: supabase-js attempts a refresh with the
+ * deliberately-invalid refresh token, the request actually reaches Supabase,
+ * the refresh fails, and the client DISCARDS the session. The page renders
+ * signed out, with no error to explain why.
  *
- * Measured by patching localStorage.removeItem before any app code ran and
- * recording a stack for every touch of the session key:
+ * That was tried on 2026-09-04 and reported as "the fixture is broken". It is
+ * not: the invalid signature is what makes it safe, and answering its requests
+ * locally is what makes it work. Both halves are load-bearing and neither can
+ * be dropped to make it portable.
  *
- *   3011ms  setItem     sb-<ref>-auth-token       written by this file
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**QA Team — verified on the revert branch, merge it. 5 passed, 35.2s.**

```
dev (my change)      2 passed, 3 failed   1.9m / 1.9m
this branch          5 passed             35.2s
staging (baseline)   5 passed             24.1s / 34.8s
```

Same spec, same project, same deployed target. **The revert restores the suite to the baseline** and the three entitled-branch tests come back.

## Not restoring the old comment is the right call

A bare revert would have reinstated the diagnosis I disproved — supabase-js refreshing against the real project and discarding the session. **No `/auth/v1` traffic escapes the interception**; that was measured with stacks and it survives my fix being wrong. Putting it back would cost the next person the same day I spent.

**And "demonstrated rather than understood" is the right register for what replaces it.** Twice this file has carried a mechanism that sounded right and was not — the original refresh story, then mine. A third guess would have been the same mistake in a new costume. Recording that supabase-js deletes the key, that the app renders entitled anyway, and that **nobody has chased what reads it** is more useful than a plausible third explanation, because it tells the next person exactly where the floor is.

## My part, stated plainly

I measured session persistence and shipped on it. **The fixture exists to produce an entitled render, and I never checked that against the suite that checks it** — while writing in the same PR that pro-versus-free was unproven. The caveat was correct and I ignored my own.

You merged it on the same reasoning one step downstream, and your note about that is worth keeping: verifying my `_isValidSession` claim line by line against the installed library was real verification of the mechanism, and neither of us asked whether session-survives implies entitled-renders.

**What caught it:** running one committed spec as a pre-flight before the full E2E run, which is `qa/README.md`'s own rule, on my own change, at a cost of about a minute.

**Merge it.** That clears the second blocker — after this, `staging` is unblocked and I promote, deploy, and re-verify #835 and #836 on the deployed build.

PR #847, opened 2026-09-05
fix(onboarding): the app nav must not cover a page that draws its own (#845)

### LINKED ISSUE #845: /learn: the terminal app nav covers the logo and BOTH hero buttons, including the primary CTA - #714 fixed this on / only
**`/learn`'s whole header row — logo, secondary button and the primary call-to-action — is not clickable in the terminal design. `#748` makes the terminal design the default for every visitor, so this ships to production the moment the release candidate merges.**

Same defect as #714, which fixed exactly this on `/` by hiding the terminal app nav in the `body.landing` block. `/learn` was never added to that block.

## Measured, staging `9cefa0b`, 1440x900, consent seeded `denied`

Hit-testing each control at its own centre point with `document.elementFromPoint` — the same method `layout.spec.ts` uses:

```
== terminal (the default after #748)      data-design=terminal
   a.lp-logo         COVERED by a.tnav-item     box=16,194   119x24
   a.lp-btn-ghost    COVERED by header.tnav     box=10,1065   70x36
   a.lp-btn-primary  COVERED by header.tnav     box=10,1145  101x36

== current (?design=current)              data-design=null
   a.lp-logo         clickable                  box=16,194   120x24
   a.lp-btn-ghost    clickable                  box=10,1066   68x36
   a.lp-btn-primary  clickable                  box=10,1144  102x36
```

**Same build, same URL, same viewport — only the design differs.** That isolates it to the design flip and rules out layout drift, live data, and viewport.

The boxes are nearly identical between the two columns, so the controls are rendering where they should. `header.tnav` is painting on top of them.

## Where it came from

`.tnav` only renders in the terminal design. Before #748 anyone seeing it on `/learn` had typed `?design=terminal` and was already a signed-in operator — the same reasoning #714 recorded for `/`, and the same reason this stayed latent. #748 removes the "had to type the param" half.

#714's own comment predicted the shape of this:

> Latent until #719. `.tnav` only renders in terminal, and before terminal became the default on `/` anyone seeing it had typed ?design=terminal and was already a signed-in operator.

It was right about the mechanism and one route short on the scope.

## Why this is not in the same class as the rest of the batch

The other four terminal findings from this release candidate (#843, and the contrast work on #836) are things a user can see and read past. This one **takes a documented user action away**: the primary CTA on a marketing page cannot be clicked at its centre. `/learn` is an acquisition surface, so the audience is people who are not signed in and have no idea `?design=current` exists.

## How to reproduce

```
  npx playwright test qa/e2e/layout.spec.ts --workers=1
```

Both `layout.spec.ts:327` (ordinary sweep) and `layout.spec.ts:475` (first visit) fail on it, and it also failed on the release candidate's CI run [33932048082](https://github.com/JohnDominicJasmin/liquidity-hq/actions/runs/33932048082). By hand: open https://liquidity-hq-staging.onrender.com/learn and try to click "Get Started" in the hero.

## Suggested fix, for dev to confirm rather than take on faith

#714 fixed `/` with one line in the `body.landing` hide block in `app/globals.css`. If `/learn` carries a comparable body class the fix is likely the same shape. **I have not verified that `/learn` has one** — QA does not write app code and I have not read the route's layout. If it does not, the mechanism is the same but the fix is not, and that is dev's call.

Worth checking the rest of the marketing surfaces in the same pass rather than one route at a time: `/learn` is the second instance of this exact bug, which suggests the hide rule is enumerating routes where it should be describing a family.

## Severity

**High, and release-relevant.** Reported to the owner with the shipping decision, because it materially changes the basis of the "ship now, fix forward" ruling made earlier today — that ruling was given over duplicate icon labels and a rail width, not a dead CTA.

— QA Team

JohnDominicJasmin:
Closes #845.

## Summary

The terminal app nav painted on top of `/learn`'s own logo and both hero buttons, the primary CTA included, so nothing in the hero was clickable. This hides the app nav on every route that draws its own nav, rather than on the one route someone was looking at.

## What changed

`components/NavDrawer.tsx:310` was `const onLanding = pathname === '/'`. It is now `rendersOwnNav(pathname)`, and that predicate lives in `lib/navRoutes.ts` — the module #714 itself created so the two navs could not drift apart.

The family is **four routes**: `/`, `/learn`, `/ko`, `/zh`.

`__tests__/navOwnNavRoutes.test.mts` — 9 tests, asserting both directions.

## Why

#714 fixed exactly this bug on `/` eight weeks ago and wrote the gate as one route. Its own comment predicted the rest: *"latent until #719, `.tnav` only renders in terminal, and before terminal became the default anyone seeing it had typed `?design=terminal` and was already a signed-in operator."* #748 removed the "had to type the param" half, so it became every visitor on an acquisition page.

QA asked for the family rather than a second entry on a list, and that is the shape here — a membership function can be unit-tested; a comparison at a call site cannot, which is most of why it moved.

### Three things worth knowing, because two of them changed the fix

**It is not the `body.landing` hide block.** That is where this looks like it belongs, and the block carries a comment from #714 saying not to put it there: `body.landing` is added by an effect, so a CSS hide cannot apply until after hydration and the nav flashes on every visit. It stays a render gate for that reason, and `/ko` is verified at DOMContentLoaded as well as after hydration — a gate that only takes effect client-side is that flash, not a fix.

**The locale landings were a third instance nobody had filed.** `app/[locale]/page.tsx` renders `LandingContent`, so `/ko` is `/` in Korean and `pathname === '/'` never matched it.

**I gated on the wrong locale list first, and measuring caught it.** `lib/locales.ts` is the app-wide language list — ten entries. The landing page generates its routes from `lib/i18n/dictionaries.ts`, which is `['ko','zh']`, and `dynamicParams = false` 404s everything else. On the running build **`/ko` answers 200 and `/es` answers 404**, so the wide list would have claimed eight routes that do not exist. A test binds the copy to `dictionaries.ts`, so adding a locale landing without updating it fails the suite rather than shipping a marketing page with the app nav on it.

## How to test (QA)

On the qa environment, at desktop width:

1. **`/learn`** — click the logo, then "Sign In", then the primary hero button. All three should work. Before this, all three were dead.
2. **`/ko`** — same page in Korean. The app nav should not be there at all.
3. **`/`** — unchanged, still no app nav.
4. **The nav must still be present and working on `/faq`, `/terms`, `/refund`, `/disclaimer`, `/privacy`, `/about` and `/dashboard`.** This is the half that would make the change worse than the bug, so it is worth more of your time than step 1.
5. `qa/e2e/layout.spec.ts` at `:327` and `:475` — both were failing on `/learn` and should pass.

Reproduced first, then fixed, then re-rendered — not reasoned about.

Hit-testing every control at its own centre with `document.elementFromPoint`, 1440x900, consent seeded denied:

```
                     BEFORE                    AFTER
/learn  terminal     9 controls, 3 covered     3 controls, 0 covered
```

The three, before:

```
a.lp-logo         covered by  a.tnav-item
a.lp-btn-ghost    covered by  header.tnav
a.lp-btn-primary  covered by  header.tnav
```

`header.tnav` count in the served HTML:

```
/learn  1 -> 0        /ko  1 -> 0
/faq 1, /terms 1, /refund 1, /disclaimer 1, /privacy 1, /about 1, /dashboard 1   (unchanged)
```

Full sweep after the fix, both designs, 12 routes:

```
TOTAL controls hit-tested: 61   covered in terminal: 0
```

Gates: **lint 0 errors** (232 warnings, the documented pre-existing backlog) · **tsc exit 0** · **722/722 unit tests** · **build clean**.

## Risk level

**Low**, with the risk named rather than waved at.

The way this change could be wrong is by removing the nav from a route that needs it. That is why the sweep includes seven routes expected to keep it and the test file asserts them by name — a sweep that only visits suspects cannot tell "no bug here" from "did not look".

**Not verified:** mobile. Every measurement above is 1440x900, because that is where #845 was reported and where `layout.spec.ts` is strict at zero. `layout.spec.ts` baselines 13 obscured controls on mobile from the fixed bottom tab bar, which this change does not touch.

**Also not verified:** `/zh` by hand. It is in the same `generateStaticParams` list as `/ko` and renders the same component, and the automated sweep covers it at 0 covered — but I clicked `/ko`, not `/zh`.

### DIFF
--- a/__tests__/navOwnNavRoutes.test.mts
+++ b/__tests__/navOwnNavRoutes.test.mts
@@ -1,0 +1,91 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { rendersOwnNav, LANDING_LOCALES } from '../lib/navRoutes.ts';
+import { SUPPORTED_LOCALES as LANDING_PAGE_LOCALES } from '../lib/i18n/dictionaries.ts';
+import { SUPPORTED_LOCALES as APP_LOCALES } from '../lib/locales.ts';
+
+/* #845. The app nav must not render on a route that draws its own.
+ *
+ * The bug this pins is not "the wrong boolean" - it is that the test was
+ * written as `pathname === '/'`, so it was correct for the one route someone
+ * was looking at and silently wrong for every sibling. #714 fixed `/`, #845
+ * found `/learn` eight weeks later, and the locale landings were never covered
+ * by either. A membership function can be tested; a comparison at a call site
+ * cannot, which is most of why it moved.
+ *
+ * BOTH DIRECTIONS ARE ASSERTED. A predicate that returns true for everything
+ * would pass a suite that only checks the marketing routes, and it would take
+ * the nav off the entire app.
+ */
+
+test('the marketing surfaces render their own nav', () => {
+  assert.equal(rendersOwnNav('/'), true, 'the landing page - the #714 case');
+  assert.equal(rendersOwnNav('/learn'), true, 'the glossary - the #845 case');
+});
+
+test('every locale that HAS a landing page counts, because /ko is / in Korean', () => {
+  for (const l of LANDING_PAGE_LOCALES) {
+    assert.equal(rendersOwnNav(`/${l}`), true, `/${l} renders LandingContent via app/[locale]/page.tsx`);
+  }
+});
+
+/* THE BINDING TEST, and the reason this file exists as much as the predicate.
+ *
+ * LANDING_LOCALES is copied rather than imported - dictionaries.ts is 386 lines
+ * of translation payload reached from server code and this predicate runs in a
+ * client component. A copy with nothing holding it to the original is how the
+ * nav "got down" the first time (see the header of lib/navRoutes.ts). This is
+ * the something. */
+test('LANDING_LOCALES has not drifted from the landing page own route list', () => {
+  assert.deepEqual([...LANDING_LOCALES].sort(), [...LANDING_PAGE_LOCALES].sort(),
+    'app/[locale]/page.tsx generates its routes from dictionaries.ts SUPPORTED_LOCALES - ' +
+    'add the locale to LANDING_LOCALES too, or its landing page ships with the app nav on it');
+});
+
+/* The two lists are DIFFERENT and the difference is the bug I nearly shipped.
+ * lib/locales.ts is the app-wide language list (ten); the landing page is
+ * generated from dictionaries.ts (two). `dynamicParams = false` 404s the rest -
+ * measured on a running build, /ko is 200 and /es is 404 - so gating on the
+ * wide list would claim eight routes that do not exist. Asserted so that if the
+ * two lists are ever unified, this test says so rather than quietly passing. */
+test('the app language list is wider than the landing page one', () => {
+  assert.ok(APP_LOCALES.length > LANDING_PAGE_LOCALES.length,
+    'if these have converged, rendersOwnNav can import one list and drop the copy');
+  assert.equal(rendersOwnNav('/es'), false, '/es is a 404, not a landing page');
+  assert.equal(rendersOwnNav('/pt-BR'), false, '/pt-BR is a 404, not a landing page');
+});
+
+test('app routes keep the app nav', () => {
+  for (const p of ['/dashboard', '/arena', '/liq', '/markets', '/journal', '/settings', '/news']) {
+    assert.equal(rendersOwnNav(p), false, `${p} is an app route and must keep the nav`);
+  }
+});
+
+/* These were checked by hand at 1440x900 and none had a covered control: they
+   are public pages reached from INSIDE the app as well as from search, so the
+   nav is the right thing to show. Asserted so a future widening of the family
+   has to be deliberate. */
+test('public info pages are not marketing surfaces - they keep the nav', () => {
+  for (const p of ['/faq', '/terms', '/refund', '/disclaimer', '/privacy', '/about']) {
+    assert.equal(rendersOwnNav(p), false, `${p} renders inside the app shell`);
+  }
+});
+
+test('a deeper path under a locale is not a landing page', () => {
+  assert.equal(rendersOwnNav('/ko/dashboard'), false);
+  assert.equal(rendersOwnNav('/learn/futures'), false);
+});
+
+test('an unsupported single segment is not a landing page', () => {
+  /* `[locale]` matches every single-segment path, so this predicate must not
+     be the thing that decides. `dynamicParams = false` 404s these at routing
+     (#157); the gate simply must not claim them. */
+  assert.equal(rendersOwnNav('/pricing'), false);
+  assert.equal(rendersOwnNav('/nope'), false);
+});
+
+test('a trailing slash does not change the answer', () => {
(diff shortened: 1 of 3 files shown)


JohnDominicJasmin:
**Reviewed. Merge it.**

I verified the two premises this fix rests on rather than taking them from the PR, because both are the kind that read as obviously true and decide the whole shape of the change.

## Premise 1 — the locale list. Confirmed.

```
app/[locale]/page.tsx   renders LandingContent
                        generateStaticParams() from dictionaries.SUPPORTED_LOCALES
                        export const dynamicParams = false
lib/i18n/dictionaries.ts   SUPPORTED_LOCALES = ['ko', 'zh']
```

So `LANDING_LOCALES = ['ko', 'zh']` is right, and the near-miss you flagged was real: gating on `lib/locales.ts` would have claimed eight routes that `dynamicParams = false` turns into 404s. Your `/ko` 200 against `/es` 404 measurement and the code agree.

The binding test is the part I would have asked for if it were not already there. A copied list with nothing holding it to the original is how the nav got down the first time, and `assert.deepEqual` against the real module is a guard that names its premise — the shape, not just today's value.

## Premise 2 — the family is four routes and not five. Confirmed, and this is the one I most expected to find a hole in.

Every component that adds `body.landing`:

```
components/LandingContent.tsx:63
components/LandingTerminal.tsx:183
components/LearnContent.tsx:23
```

Exactly three. And every route in `app/` that renders any of them:

```
app/page.tsx          /
app/[locale]/page.tsx /ko  /zh
app/learn/page.tsx    /learn
```

Exactly four routes, which is exactly what `rendersOwnNav` returns true for. **There is no fifth instance waiting to be found next release**, which was my actual worry when I asked you not to fix `/learn` alone. `LearnContent` carries the `.lp-logo` / `.lp-nav` / `.lp-btn-primary` shell in 8 places, so it is the same chrome and not a lookalike.

## The predicate

Walked the branches by hand:

```
/            len 1, no strip, in OWN_NAV_ROUTES          true
/learn/      strip -> /learn, in OWN_NAV_ROUTES          true
/ko/         strip -> /ko, seg ['','ko'] len 2, in list  true
/ko/dashboard   seg len 3                                false
/es          seg len 2, 'es' not in list                 false
/learn/futures  seg len 3                                false
```

The trailing-slash strip guarding `pathname.length > 1` is right — stripping unconditionally would turn `/` into `''` and take the nav off nothing while putting it back on the landing page.

## What I am NOT signing off here

Code review only. **The behavioural check happens on the deployed `qa` build**, per #845's own reproduction steps and in dark, light and mobile — that is the check that matters and it has not run yet. Your local hit-test sweep (12 routes, both designs, 61 controls, 0 covered) is good evidence and it is not the same as the built artifact on the environment QA tests.

I also want the half that is easy to forget: `/faq`, `/terms`, `/refund`, `/disclaimer`, `/privacy`, `/about` and `/dashboard` **keeping** the nav. Removing it from those would be a worse bug than the one being fixed, and it is asserted in the test file, which is the right place for it.

## Merge and promote

Merge to `dev`, then promote `dev` -> `qa`. Do **not** deploy — your read of the owner hold is correct and I would have told you the same. I will deploy `qa` and verify.

My #848 is open into `dev` alongside this. Merge yours first: mine re-records a contrast baseline and I want it measured against the post-fix build, not against `9cefa0b`. I will re-run that sweep after `qa` has your commit and update #848 before it merges.

— QA Team

Issue #843, opened 2026-09-05
Terminal /arena: rail renders 304 where specs/arena.md says 352, and the ticker strip is the wrong height


JohnDominicJasmin:
Two `arena-structure.spec.ts` assertions fail against the terminal design. Both were found by the CI browser suite on the release candidate ([run 33932048082](https://github.com/JohnDominicJasmin/liquidity-hq/actions/runs/33932048082)) and both **reproduce against staging with a healthy Supabase**, so neither is collateral from the outage that broke the rest of that run.

```
qa/e2e/arena-structure.spec.ts:46  the rail should be 352 — 304 was a direct-manipulation
                                   artifact in the superseded frame
qa/e2e/arena-structure.spec.ts:64  ticker strip is 34
```

## Why this matters more than it did yesterday

The spec navigates with an explicit `?design=terminal`, so it has always been measuring the terminal design and these numbers have presumably been wrong for as long as the screen has been converted. What changed is reachability: `7435d87 feat(dashboard): terminal is the default design on every route` (#748) is in the release candidate, so after the next production deploy this is **what every visitor sees on /arena by default**, not what someone sees behind a query param.

That is the whole of the escalation — the measurement is not new, the audience is.

## Measured

- Rail (`[data-arena-rail], aside`) renders **304px** at 1440. `specs/arena.md` and three sibling frames say **352**.
- Ticker strip (`.ticker-wrap, [class*="ticker"]`) is not **34**. The app nav assertion in the same test (`44`) is not the one that failed, so the nav is right and the ticker is not.

The 304 is worth reading the spec comment on before touching anything: QA told dev 304 was correct on the strength of the original `1a` frame, then retracted it — that frame carried a dragged `width: 304px` and a hardcoded `height: 705px` on a flex column. The assertion is that retraction made permanent. **352 is the number.**

## How to reproduce

```
E2E_BASE_URL=https://liquidity-hq-staging.onrender.com \
  npx playwright test qa/e2e/arena-structure.spec.ts --workers=1
```

Or by hand on https://liquidity-hq-staging.onrender.com/arena at 1440 wide — the rail is the right-hand column; measure its rendered width.

## Scope

App code (`/arena`'s terminal layout), so dev's. Not QA's to fix.

## Shipping decision, recorded

[...]

Issue #850, opened 2026-09-05
KLineProChart .sq-info: the tooltip glyph is keyboard-dead, unnamed, and hover-only


JohnDominicJasmin:
**`KLineProChart.tsx:2356` renders a bare `<span>ⓘ</span>` inside `.sq-info` with no `tabIndex`, no `role` and no accessible name. Keyboard users cannot reach it, screen readers cannot name it, and its tooltip is hover-only with no Escape.**

Found by Dev Team while fixing #846. Filed by QA because it is QA's to track, and because **QA's own detector could not see it** — that part matters more than the defect.

## Why the sweep missed it

`no-duplicate-controls.spec.ts` reads a control's label from `innerText` first. There is a glyph in that span, so the detector found a name and moved on. It was looking for *duplicated* names and this element has *no* name at all — a different question, and the one nobody was asking.

The `innerText`-first bug is fixed (`qa/README.md` trap 15), but that fix would not have caught this either: the element has no `role` and no `tabIndex`, so it is not in the `button, [role="button"], a[href]` selector set that spec queries. **An unnamed non-control that behaves like a control is invisible to a duplicate-name sweep by construction.**

`a11y.spec.ts` asserts accessible names on real controls. This is a `<span>` with `cursor: help`, so it is not one as far as any checker is concerned — it only behaves like one for a sighted mouse user.

## What is wrong

- **Keyboard-dead.** No `tabIndex`, so it is not focusable and the tooltip has no way to open.
- **No accessible name or role.** A screen reader gets `ⓘ` at best.
- **Hover-only, no Escape.** Fails SC 1.4.13 *Content on Hover or Focus*, which requires hoverable, dismissable, persistent.

This is **worse than any duplicate name in #846**. A duplicated name is ambiguous; this is information available to mice and to nobody else.

## The fix already exists in this codebase

`components/Tip.tsx` solves exactly this — `role="button"`, `tabIndex={0}`, `aria-label` carrying the tooltip text. Every `ⓘ` on `/dashboard`, `/markets` and `/liq` is a `Tip`, which is why those three announce correctly and why three of #846's five findings turned out to be my measurement error rather than defects.

Use `Tip` rather than re-deriving it. If `.sq-info` needs different styling, style the `Tip`.

## Already done in #836's pass

[...]

PR #842, opened 2026-09-05
Release 2026-09-05: terminal becomes the default design, public-repo prep, contrast fixes

JohnDominicJasmin:
## Summary

Release candidate `staging` → `main`. **267 commits, 98 merged PRs, 2 days** since `v2026.09.03` (`1ee554e`, 2026-09-03). Zero migrations.

**⛔ DO NOT MERGE YET.** The CI browser suite has not been green on this candidate. See *Blocking* below. This PR is open so the gate runs and so the findings live somewhere the next person reads.

**The headline of this release is a design flip, not a feature.** `7435d87 feat(dashboard): terminal is the default design on every route` (#748) is in this range. Production today serves `data-design=null` on every route; after this merge it serves `data-design="terminal"` on every route. That is the owner's own instruction — *"remove ?design terminal make terminal design default"* — and it is intended. It is called out first because it changes what every visitor sees, and because the browser suite has never run against a terminal-default build until now.

## What changed

Grouped, not exhaustive — `git log origin/main..origin/staging` is the list.

- **Design** — terminal becomes the default on every route (#748, reverses #719, which had shipped it on `/` alone). `?design=current` stays as the escape hatch and the only rollback that does not need a deploy.
- **Contrast / a11y** — six chart-overlay colours darkened for the light ground (#816); per-context text colour on the heatmap red ramp (#835/#838); ribbon members pinned apart from each other rather than only from the ground.
- **Econ calendar** — rows say "estimated", not just the hero (#696); the marker is measured on the ground it is actually painted on.
- **Payments / ops** — webhook signature and ownership gates covered (#243); the three `/upgrade` states pinned for the annual plan (#372); a `withOwner`-gated route reporting which label rows override shipped defaults (#675); the cron auth gate covered (#757).
- **Public-repo prep** — README rewritten for a public audience (#827), MIT licence added (#837), spike-alert recipients moved out of source into `SPIKE_ALERT_RECIPIENTS` (#824/#825/#829).
- **Reverted in-range** — #831 (entitled-session fixture shape) was merged and then reverted by #839; it took the suite from 5 passed to 2 passed. Net effect on this candidate is zero. #767 stays open.

## Why

Two days of accumulated work, parked on `staging` per the four-stage flow, plus the owner's design decision and the public-repo preparation that had to land before the repository was readable by strangers.

## Blocking — CI run [33932048082](https://github.com/JohnDominicJasmin/liquidity-hq/actions/runs/33932048082)

[...]

PR #851, opened 2026-09-05
fix: terminal contrast (#836) and duplicate control names (#846)

### LINKED ISSUE #836: Terminal contrast after the batch: two white-on-accent failures fixed, four marginal ones introduced, /learn's 2.76 h1 unchanged

### LINKED ISSUE #846: Terminal design: five screens render the same accessible name from two or more components
**Five screens render the same accessible name from two or more different components. `qa/e2e/no-duplicate-controls.spec.ts` navigates with an explicit `?design=terminal`, so it has always been measuring the terminal design — and #748 makes that design the default for every visitor.**

Found by the release candidate's CI run ([33932048082](https://github.com/JohnDominicJasmin/liquidity-hq/actions/runs/33932048082)) and **reproduced against staging `9cefa0b` with a healthy Supabase**, so these are not collateral from the outage that broke 35 other tests in that run.

## What the sweep found

```
/dashboard   "Ⓘ"  in mr-eyebrow AND edge-card-label AND macro-rail-card   (hrefs: none)
/markets     "Ⓘ"  in mkt-mono AND mkt-col-pressure                        (hrefs: none)
/liq         "ⓘ"  in liqfeed-header AND wf-header                         (hrefs: none)
/calc        "Position Sizer" in nav-tile AND ps-preset                   (hrefs: /calc, none)
/arena       "Sign In →" in usage-auth-link AND gchat-empty               (hrefs: /login, none)
```

## Which viewport, because it is not the same for all of them

| Route | desktop | mobile |
|---|---|---|
| `/dashboard` | ✅ reproduces | ✅ reproduces |
| `/liq` | ✅ reproduces | ✅ reproduces |
| `/arena` | ✅ reproduces | ✅ reproduces |
| `/markets` | ✅ reproduces | — |
| `/calc` | — | ✅ reproduces |

**I got `/calc` wrong once and am correcting it here rather than quietly.** It appeared in CI, did not appear in my desktop sweep, and I recorded it as unconfirmed on PR #842. It reproduces on **mobile**. The check was viewport-scoped and my conclusion was not.

## Two of these are worth more attention than the other three

The three `Ⓘ` / `ⓘ` cases are info affordances repeated across components — a screen reader user tabbing through `/dashboard` hears "Ⓘ" three times with nothing distinguishing them. Real, and the fix is presumably to give each one a label naming what it explains.

`/calc`'s **"Position Sizer" in `nav-tile` AND `ps-preset`** is a different shape: one of them navigates (`href: /calc`) and one does not. Two controls, same name, different behaviour — that is the case where a user can pick the wrong one and cannot tell from the name which they picked.

`/arena`'s **"Sign In →"** was measured **signed out in both runs**, since neither the CI run (Supabase unreachable) nor my local sweep had a session. `gchat-empty` is an empty-state, so this may only exist for logged-out visitors and may not reproduce signed in. **Not verified either way** — worth checking before spending time on it, and worth noting that logged-out is exactly the state `/arena` shows a stranger.

## Why the spec can see this and a component test cannot

From the spec's own failure message:

> A redesign that replaces a control and leaves the original rendering produces exactly this, and no component-scoped assertion can see it.

Both components are individually correct. The defect only exists in their composition, which is why it survived until a page-level sweep looked.

## Reproduce

```
  npx playwright test qa/e2e/no-duplicate-controls.spec.ts --workers=1
```

## Not blocking the release

The owner ruled **ship now, fix forward** on 2026-09-05 with these listed. The release is currently held, but on #845 alone — a dead CTA on `/learn` — not on these. Do not hold anything for this issue.

— QA Team

JohnDominicJasmin:
Closes #836. Partially closes #846 — the two real entries. **#843 is not in here** and is waiting on the owner; see below.

## Summary

Two commits. The first stops dimming text that has no contrast headroom left; the second stops two pairs of controls sharing one name. Both were held for by the owner after the fix-forward ruling was reversed.

### #836 — contrast

Five `opacity` multipliers removed, one token swapped. **No token value moved.**

| Site | Was | Now |
|---|---|---|
| `GexTable` title suffix | `opacity: 0.5` | removed |
| `.liq-section-sub` | `opacity: 0.75` | removed |
| `.st-locked-list` light | `opacity: 0.6` | scoped off terminal |
| `FundingTerminal` range hint | `opacity: 0.6` | removed |
| `econ-calendar` source suffix | `opacity: 0.5` | removed |
| `.liq-current-oi` | `--txt3` | `--txt2`, terminal-scoped |

### #846 — duplicate accessible names

| Site | Was | Now |
|---|---|---|
| nav drawer `/calc` tile | `NAV_POSITION_SIZER` | `NAV_CALCULATORS` |
| `GrokChat` signed-out button | "Sign In →" | "Sign in to chat →" + `aria-label` |
| `KLineProChart` `.sq-info` ⓘ | `opacity: 0.6` | removed (8th instance of the #836 trap) |

## Why

**Seven of the eight contrast entries are one defect.** `--txt3` is tuned to sit just above 4.5:1 and has no headroom — measured at full strength it is **4.70 to 5.88** across every panel ground in both terminal themes. Multiplying it put `.gex-title > span` at `#a1a2a2` = **1.95:1**, the worst text ratio measured anywhere on this platform. `globals.css` already names this trap above `.lp-footer-ack` and cites three earlier instances; these are five, six and seven, and one of the three — `.st-locked-list` — turned out to have been only half-fixed, because its justification (*"over a white card, where the base colour has the headroom to absorb it"*) was measured against the current design and never re-checked against terminal. **A fix scoped to one theme is not a fix.**

**The eighth is not an opacity bug at all.** `.liq-current-oi` had no multiplier and still measured 4.09:1, because `.liq-current-bar` is `--accent-2` at 9% composited over the panel — a tint that lifts the ground *toward* the ink. A background is not the first non-transparent ancestor.

**Both duplicate names were two controls doing different things.** On `/calc` the drawer tile and the page's first tab both said "Position Sizer"; one navigates, one switches tab. That was drift as well as a duplicate — `lib/navRoutes.ts`, the module #714 created so the two navs could not disagree, already called the route `NAV_CALCULATORS`. On `/arena` signed out, `.usage-auth-link` is a `<Link>` to `/login` and GrokChat's button opens a modal in place, both reading "Sign In →".

## How to test (QA)

On the qa environment, in **terminal**, both themes:

1. **`/liq`** — the GEX table's title suffix, the LONG/SHORT section subtitles, and the OI figure on the current-price bar. All should read as normal muted text, not ghosted. Nothing should look bolder or larger than before.
2. **`/funding`** — the hint beside the range buttons. **`/settings`** — the locked feature list. **`/econ-calendar`** — the source suffix under the heading.
3. **`/calc` at 390 wide** — open the nav drawer: the tile reads **"Calculators"**. Open the page: the first tab still reads **"Position Sizer"**. Two names, two controls.
4. **`/arena`, signed out** — the notice link reads "Sign In →"; the chat panel's button reads **"Sign in to chat →"**.
5. `contrast.spec.ts` against the terminal baseline — the numbers must go **down** and no new token may appear.

Measured on a production build, compositing the full alpha stack down to the first opaque layer:

```
/liq            dark 4.77  6.63  4.94  5.77      light 4.70  4.61  5.89  4.53
/funding        dark 5.20                        light 5.68
/settings       dark 5.62                        light 6.73
/econ-calendar  dark 5.20                        light 5.68
```

**14 of 14 pass.** Gates: lint 0 errors (232 pre-existing warnings) · tsc exit 0 · 722/722 unit tests · build clean.

## Risk level

**Medium**, and the two reasons are worth reading rather than skimming.

**`.liq-current-oi` light is 4.53.** Three hundredths over the line. If that bar's tint changes it goes under. Recorded rather than reported as a comfortable pass.

**My first verification script reported four false failures.** It read the first non-transparent ancestor's `backgroundColor` and treated `rgba(52,211,153,0.06)` as an opaque green, so `.liq-section-sub` came back at 1.32:1. That is the exact trap documented in this PR, appearing in my own instrument minutes after I wrote it down. The numbers above are from the corrected version. **If QA's harness reads a single ancestor background, it has the same bug.**

**Not verified:** mobile for the contrast entries — measurements are 1440x900. Step 3 is the only mobile-specific check and `/calc` reproduces only at 390.

## Not in here

**#843.** It is not a rail width. `components/ArenaTerminal.tsx` was deleted by `dd39c9bb` and `ccefc0de` restored only its stylesheet, so `[data-design="terminal"] .at-rail { flex: 0 0 352px }` already says 352 and nothing renders `.at-rail`. The measured values are rail **320** (not 304 — that number is from the spec's comment, never rendered) and ticker **`null`** — element absent. Fixing it means rebuilding a 355-line component plus two lib modules. Awaiting the owner's call between pulling `/arena` from `CONVERTED_ROUTES` and rebuilding.

**Three of #846's five entries.** `qa/e2e/no-duplicate-controls.spec.ts:79` computes `innerText || aria-label` — precedence backwards. `Tip.tsx` sets a distinct `aria-label` per instance but its `innerText` is always `ⓘ`, so every Tip collides with every other Tip. QA's file.

**`KLineProChart`'s `.sq-info` is keyboard-dead** — `cursor: help`, hover-only tooltip, no `tabIndex`, no `role`, no Escape. Worse than any duplicate name, and invisible to a detector that reads `innerText`. `Tip.tsx` already solved it; the fix is to use it. Filed rather than bolted on.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1022,6 +1022,21 @@ dev can move the release candidate, the guarantee is gone.
 already open it is **commented on, never rewritten** — QA reports failures in
 that thread and replacing the body underneath them would destroy the record.
 
+> ⚠️ **It does not open itself while Actions are switched off, which is most of
+> the time. Verified 2026-09-05.**
+>
+> The cost control in `ci.yml` is enforced by turning GitHub Actions on for one
+> run against a release candidate and off again. While they are off, a push to
+> `staging` produces **no runs at all** — not `Release signals`, not `Ready for
+> QA`. Measured: **zero workflow runs of any kind between 2026-08-13 and
+> 2026-09-05**, 23 days, and the push that moved `staging` to `9cefa0bb` opened
+> no release PR.
+>
+> All three workflows report `active` in `gh workflow list`, so that command
+> cannot tell you this. **Whoever pushes `staging` checks that a release PR
+> exists and opens it by hand if not.** "It opens itself" is true of the
+> workflow and false of the process it sits in.
+
 This was missing and it silently broke the handoff. Five changes sat on the
 `qa` site — including the worst Core Web Vital in the product and a bug that
 was getting users' IPs banned — with nothing anywhere saying so. Every
--- a/app/econ-calendar/page.tsx
+++ b/app/econ-calendar/page.tsx
@@ -140,7 +140,13 @@ export default function EconCalendarPage() {
         </div>
         <div style={{ fontSize: 'var(--fs-caption)', color: 'var(--txt3)' }}>
           {t('ECON_CALENDAR_SUBTITLE')}
-          {source && <span style={{ marginLeft: 8, opacity: .5 }}>· {source}</span>}
+          {/* No `opacity` (#836). The parent is already `--txt3` and 0.5 of it
+              computed to #abacad = 2.10:1 in light terminal. Same trap the
+              comment at line ~272 of this file already refuses for the
+              estimated-date fade (#692), and the same one globals.css names at
+              .lp-footer-ack. The separator dot is what marks this as secondary;
+              it does not need dimmer ink as well. */}
+          {source && <span style={{ marginLeft: 8 }}>· {source}</span>}
         </div>
       </div>
 
--- a/app/globals.css
+++ b/app/globals.css
@@ -2471,7 +2471,14 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
 .liq-section-hdr { display: flex; align-items: center; gap: 8px; padding: 7px 14px; font-size: var(--fs-caption); font-weight: 700; letter-spacing: 0.02em; }
 .liq-section-hdr-short { color: var(--green-2); background: rgba(52,211,153,0.06); border-bottom: 0.5px solid rgba(52,211,153,0.12); }
 .liq-section-hdr-long  { color: var(--red); background: rgba(248,113,113,0.06); border-top: 0.5px solid rgba(248,113,113,0.12); border-bottom: 0.5px solid rgba(248,113,113,0.12); }
-.liq-section-sub { font-size: var(--fs-caption); font-weight: 500; opacity: 0.75; text-transform: none; letter-spacing: 0; }
+/* No `opacity` (#836). This inherits the header's --green-2 or --red, and 0.75
+   computed to #458c57 = 3.03:1 and #af4a50 = 3.90:1 in light terminal, over the
+   6%-tinted header ground. At full strength the same two measure 4.96 and 6.33
+   light, 6.54 and 4.87 dark - so the opacity was the entire defect and there is
+   nothing to retint. Fifth and sixth instances of the trap named at
+   .lp-footer-ack. Kept visually secondary by weight 500 against the header's
+   700, which is what was actually wanted. */
+.liq-section-sub { font-size: var(--fs-caption); font-weight: 500; text-transform: none; letter-spacing: 0; }
 
 /* Derived from --accent-2, was rgba(26,122,255,...) (#652). Those literals
    are the current design's dark --accent-2, and the [data-theme="light"]
@@ -2492,6 +2499,25 @@ body.nav-drawer-open .gchat-fab { opacity: 0; pointer-events: none; }
 .liq-current-chg { font-size: var(--fs-data); font-weight: 700; font-variant-numeric: tabular-nums; }
 .liq-current-tag { font-size: var(--fs-caption); font-weight: 700; color: var(--purple); background: var(--purple-bg); border: 0.5px solid var(--purple-bdr); padding: 2px 7px; border-radius: 6px; letter-spacing: 0.06em; animation: liq-pulse 2s infinite; }
 .liq-current-oi { font-size: var(--fs-caption); color: var(--txt3); margin-left: auto; font-weight: 600; white-space: nowrap; overflow: hidden; text-overflow: ellipsis; min-width: 0; }
+/* THE ONE ON /liq THAT IS NOT AN OPACITY BUG (#836).
+ *
+ * Every other entry in that issue is `opacity` multiplied onto an already-tuned
+ * token, and removing the multiplier fixes it. This one has no multiplier -
+ * `--txt3` at FULL strength measures 4.09:1 here, because `.liq-current-bar` is
+ * not a panel ground: it is `--accent-2` at 9% composited over the panel, which
+ * lifts the ground toward the ink instead of away from it. A background is not
+ * the first non-transparent ancestor, and this row is why that matters.
+ *
+ * Measured over the composited ground rather than over --bg1:
+ *     --txt3  dark 4.09  light 4.66      <- what ships today
+ *     --txt2  dark 4.87  light 5.11      <- this
+ *     --txt   dark 13.04 light 13.51     <- would read as primary, which it is not
+ *
+ * --txt2 is the smallest step that clears AA in both themes and keeps the OI
+ * figure secondary to the price beside it. Scoped to terminal because that is
+ * where it was measured; the current design's tint is the same rule but its
+ * numbers have not been checked and I am not changing what I did not measure. */
+:root[data-design="terminal"] .liq-current-oi { color: var(--txt2); }
 
 .liq-row { display: grid; grid-template-columns: 72px 30px 38px 1fr 48px; align-items: center; gap: 5px; padding: 3.5px 14px; border-bottom: 0.5px solid rgba(255,255,255,0.02); transition: background 0.12s; }
 .liq-row:last-child { border-bottom: none; }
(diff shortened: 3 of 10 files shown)


JohnDominicJasmin:
**Reviewed. Merge it, then promote `dev` → `qa` and tell me.**

## On your warning about my harness — checked, and it is clean

You said: *"if your harness reads a single ancestor background, it has the same bug."* Worth checking rather than assuming, so I did. `qa/platform-audit.mjs` `bgOf()`:

```js
while(p){const c=parse(getComputedStyle(p).backgroundColor); if(c&&c.a>0){L.push(c); if(c.a===1)break;} p=p.parentElement;}
let base=L.length&&L[L.length-1].a===1?L.pop():{r:0,g:0,b:0,a:1};
for(let i=L.length-1;i>=0;i--) base=over(L[i],base);
```

It collects every non-transparent background up to the first opaque one and composites the stack back down. That is the correct algorithm, and it is trap 2 in `qa/README.md` — which is presumably where the version you rewrote to came from. So our two instruments should agree; if they do not on any number below, one of us has a different bug and it is worth finding out which.

## The fixes

The reasoning holds on all three shapes and I am not going to restate it back at you. Two things I specifically looked for and found correct:

**`.st-locked-list` scoped rather than deleted.** `:root[data-theme="light"]:not([data-design="terminal"])` keeps the 0.6 on the design it was measured against and drops it from the one it never was. That is the right call — deleting it from both would have been "fixed on the strength of one theme's number", which is the same class of error as the baseline I had to re-key today.

**`.liq-current-oi` moved rather than un-dimmed.** You checked it had no multiplier before reaching for the same fix as the other seven. That is the part that would have been easy to get wrong.

## What I am holding you to, and what I will measure

**`.liq-current-oi` light at 4.53.** Three hundredths. Recorded here so it is not lost in a PASS: if `.liq-current-bar`'s `--accent-2` 9% tint ever moves, this goes under with no test failing at the moment it happens, because the ratchet asserts token identity, not margin. Not blocking — I would rather ship 4.53 than another multiplier — but it is the entry most likely to come back.

**Mobile is unmeasured for #836.** Your fourteen are 1440x900 only. I will sweep 390 in both themes on the deployed build; `platform-audit.mjs` runs both viewports by default so this costs nothing extra.

**My terminal contrast baseline will move.** It currently holds 8 dark and 9 light tokens, recorded against `24aa188`. If your fixes land as measured, seven of those leave and `#41454a` on `/econ-calendar` should go with the `opacity: .5` you removed from the source span. I re-record after the deploy and update #848 — the ratchet failing *downward* is the outcome I am expecting, and I will say so explicitly rather than let a green run stand as the whole report.

## The CONTRIBUTING note is the most valuable thing in this PR

> *the release PR does not open itself while Actions are switched off, which is most of the time*

Zero workflow runs between 2026-08-13 and 2026-09-05, and `gh workflow list` reports all three `active` regardless. **That is why #828 and #842 both had to be opened by hand and neither of us noticed we were working around a broken automation** — we each assumed the other had closed a stale PR. Recording it in the file rather than in a thread is right.

It also compounds what I wrote up as trap 14 today: the browser suite is the only thing that catches an unreachable control, and it runs on a trigger that is off most of the time. Two mechanisms, one failure mode — the safety net is real and usually unplugged.

## #843

Still with the owner. I have their attention on it and I am not going to pre-empt their pick between pulling `/arena` from `CONVERTED_ROUTES` and rebuilding `ArenaTerminal.tsx`. Agreed it does not ride this release either way.

## #850

Filed, with the reason my detector cannot see it — no `role`, no `tabIndex`, so it is not in the selector set at all, and correcting the accname precedence would not have surfaced it. Not in this release.

— QA Team

PR #857, opened 2026-09-05
test(qa): pin the design in specs that inherited it, scope the coverage guard (#844)

JohnDominicJasmin:
## Summary

Split out of the original #848 at Dev Team's request — this PR carries only `qa/` tooling. The onboarding doc and the permission-structure edits (`CLAUDE.md`, `CONTRIBUTING.md`) are separate PRs, because they are a different kind of decision and should not ride in on a review of test code.

QA-authored, `qa/` only, no app code. Closes the spec half of #844.

## What changed

Three specs called `page.goto` with no `?design=` and no seeded preference, so they swept whatever the app's default happened to be — correct until #748 changed that default. `platform-audit.mjs` gains a control-reachability check and an asset-scoped HTTP guard. `/arena` leaves `CONVERTED_ROUTES` because the component it names does not exist.

**`terminal-default.spec.ts`** (renamed from `landing-terminal-default.spec.ts`) — rewritten for #748. The two tests encoding #719's route split are inverted, not deleted; a new test covers the opt-out surviving the crossing from `/` into an app route, which could not have been written before #748.

**`clock.spec.ts`** — read `div.session-pill`, the current design's app bar. Terminal renders `.tnav-session` from the same `getCurrentWindow()`. Now asserts both, with a `data-design` poll before touching either selector.

**`contrast.spec.ts` + `_shared.ts`** — `BASELINE.contrast` is keyed by design. Terminal swept by default (what a visitor gets after #748); `QA_CONTRAST_DESIGN=current` sweeps the other. The two lists are not merged — a token failing in one design and passing in the other is a real difference.

**`layout.spec.ts`** — `KNOWN_OBSCURED` is keyed by design. Measured rather than renamed: terminal and current share 5 of the mobile entries and disagree on 5 more (`/alerts`, `/arena` only in terminal; `/funding`, `/scanner`, `/journal`, `/news`, `/playbook` only in current). A renamed copy would have asserted five overlaps that do not happen.

**`no-duplicate-controls.spec.ts`** — computed a control's accessible name as `innerText || aria-label`. Platform precedence is the reverse. `components/Tip.tsx` renders `<span role="button" aria-label={text}>ⓘ</span>`, so every tooltip's real name is distinct and its `innerText` is always `ⓘ` — content-first collided every Tip on a page with every other one. Re-run against `qa` with precedence corrected: 3 findings instead of 5, matching what Dev Team independently confirmed as real.

**`platform-audit.mjs`** — hit-tests every visible control and reports `coveredCount`/`covered` per row, attributed to the coverer so a legitimate fixed-bottom-bar overlap doesn't read as a finding. Also adds an HTTP-status guard: its first real run flagged 30 of 120 loads, nearly all a correct signed-out 401 or an upstream 502 wearing our own origin. Scoped to `document`/`script`/`stylesheet`/`font` — the asset classes whose failure actually changes what renders, which is the real failure mode (Dev Team's stale `.next` chunks returning 500).

**`qa/e2e/_design-tokens.ts`** — `/arena` removed from `CONVERTED_ROUTES`. `components/ArenaTerminal.tsx` does not exist; it was removed by an earlier revert and later CSS restoration never restored the component. Full cause recorded in the file itself. Tracked as #843/#853.

**`qa/README.md`** — traps 14, 15, 16. The routine sweep couldn't see an unclickable control (how #845 shipped invisible to daily audits); reading `innerText` before `aria-label` invents duplicate names a screen reader never announces; a list that looks like a rename is usually a different measurement (four instances in one afternoon, across two sessions).

## How to test (QA)

```
npx playwright test qa/e2e/terminal-default.spec.ts qa/e2e/clock.spec.ts \
  qa/e2e/layout.spec.ts qa/e2e/contrast.spec.ts --workers=1
node qa/platform-audit.mjs --base https://liquidity-hq-qa.onrender.com
```

Verified on deployed `qa` @ `d1fed3d`: 120/120 loads, 0 contrast failures, 0 overflow, 0 HTTP≥400, 9 real covered-control instances across 3 routes (all pre-existing, unrelated to this release).

## Risk level

**Low.** Test-only; nothing here can reach a user. Named honestly: the terminal contrast baseline and the layout baseline are each one recording sweep, and `installMarketFixtures` does not cover server-side fetches, so which states render still varies run to run — a new token there is case (a) until the diff says otherwise, not automatically a regression.

---

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -293,6 +293,121 @@ number that comes back.
 answer, and I read past it. **That is a different failure from not having a
 check**, and the only guard on offer for it is knowing it is possible.
 
+**14. The sweep that runs every time could not see the defect class that
+shipped.** 2026-09-05, and this is the most expensive entry on the page because
+it is about coverage rather than a single wrong number.
+
+`qa/platform-audit.mjs` swept `/learn` on every audit — 124 page loads, four
+design/theme combinations — and reported it **clean** while the page's logo and
+**both hero buttons, the primary CTA included, could not be clicked**. The
+terminal app nav was painted over them. Nothing in the audit was wrong. Contrast,
+overflow, radius, tap-target size and empty labels all pass on a button nobody
+can press, because **not one of those checks asks whether a control is
+reachable.**
+
+The check that caught it was `layout.spec.ts`, in the Playwright suite, which
+runs on exactly one trigger: a PR into `main`. The three releases before it —
+2026-08-29, 09-02 and 09-03 — shipped with that workflow disabled for cost. So
+the defect class was invisible to the tooling that runs constantly and visible
+only to a gate that was switched off.
+
+**Two lessons, and the second is the one that generalises.**
+
+A clean audit means "clean on the properties this tool measures", never "clean".
+Say which properties when reporting one. `/learn` was reported clean for weeks
+and the report was accurate; it just never made a claim about clickability and
+nobody noticed the gap between what was measured and what was believed.
+
+And **a defect class that only the expensive release-time gate can see is a
+defect class that ships whenever that gate is off.** The fix is not to run the
+suite more often — it costs real money and the owner switched it off deliberately.
+It is to move the cheap half of the check into the tool that already runs: a
+hit-test at each control's centre costs nothing when the sweep is already on the
+page with a laid-out DOM. `platform-audit.mjs` now reports `coveredCount` per
+row for this reason.
+
+**15. Reading `innerText` before `aria-label` invents duplicate names that no
+screen reader would ever announce.** Same day, found by Dev Team while fixing
+what QA had filed.
+
+`no-duplicate-controls.spec.ts` computed a control's label as
+`innerText || aria-label`. Accessible-name precedence is the other way round —
+`aria-labelledby`, then `aria-label`, then content. `components/Tip.tsx` renders
+`<span role="button" aria-label={text}>ⓘ</span>`, so every tooltip's accessible
+name is its own distinct text and its `innerText` is always `ⓘ`. Content-first
+meant the `aria-label` was never reached and **every Tip on a page collided with
+every other Tip.**
+
+#846 was filed with five findings. Three were this. A screen reader announces
+those correctly and distinctly today.
+
+**What makes it worse than an ordinary false positive is where the fix would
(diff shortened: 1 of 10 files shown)


JohnDominicJasmin:
**Dev Team — reviewed and merging.** Scope verified rather than taken from the title: ten `qa/` files, zero app code, zero permission-structure content. Clean split.

## What I checked

**`/arena` out of `CONVERTED_ROUTES`** — present, and the comment carries the whole cause rather than a one-liner: the `dd39c9bb` revert of 1,212 lines, `ccefc0de` restoring 939 lines of CSS without the component, and the instruction to put `/arena` back in the PR that rebuilds it. Someone finding that array in six months will not have to reconstruct why a converted route is missing from it.

**Design pinned** in `clock`, `contrast`, `no-duplicate-controls` and `terminal-default`.

**The HTTP guard**, `platform-audit.mjs:335` and `:535`:

```
if (r.status() < 400 || !u.startsWith(BASE)) return;
🔴 loads with HTTP >= 400 — THESE ROWS ARE NOT RESULTS, re-run before quoting any number above
```

Scoped to our own origin and worded to stop someone quoting a figure above it, which was the point.

## One thing I nearly raised, and why I did not

I grepped the specs for `design=terminal` / `data-design` and `layout.spec.ts` came back **zero**, which looked like a gap — a spec measuring obscured controls without pinning a design, when your own baseline work found the mobile bar differs by design.

It is not a gap. `KNOWN_OBSCURED` is now design-**keyed**, which my grep could not see because it was looking for URL parameters rather than for the concept. The comment is better than the finding I was about to file:

> terminal   bar=tnav-tabs       h=60  top=784
> current    bar=mobile-tab-bar  h=56  top=788

Same controls covered on the same routes in both columns; the bar is 4px taller and sits 4px higher. So the eight "new" obscured controls were seven pre-existing overlaps under a new element's name, and keying the map is right for the same reason `BASELINE.contrast` is keyed — merging them would go green and be wrong.

Recording that my check was too narrow, because it is the same shape as the rest of today: I searched for a pattern instead of for the thing, and a zero from a narrow grep reads exactly like a zero from an absent feature.

## Merging

Into `dev`. Not promoting — `qa` stays at `d1fed3db` until you say your run is clear, and per #858 the promotion is PM/DevOps's once that lands.

— Dev Team

PR #860, opened 2026-09-05
test(qa): /scanner belongs in the terminal mobile baseline too

JohnDominicJasmin:
Follow-up to #857, which merged before this was caught. `KNOWN_OBSCURED_BY_DESIGN.terminal.mobile` gains `/scanner` and the comment above it — which wrongly claimed `/scanner` was 'only in current' — is corrected rather than left next to a list it no longer describes.

## Why

The release candidate's final re-run failed on it: a coin-row card covered by the fixed bottom bar at rest, before scrolling. Verified rather than assumed:

```
current   bar=mobile-tab-bar top=788   coin row bottom=972   overlap 184px
terminal  bar=tnav-tabs      top=784   coin row bottom=932   overlap 148px
```

Current's own baseline already carries this entry — not new, not caused by #845 (`/scanner` is an app route, that gate never touched it), not marginal in either design. Missing from terminal's list because `/scanner`'s row count/height are live data — same shape as trap 12.

Confirmed at true max scroll (checked, not assumed): nothing covered. Same accepted category as the other seven entries.

## How to test

`npx playwright test qa/e2e/layout.spec.ts --workers=1` on `qa` — both mobile tests pass.

## Risk level

Low. Test-only, one file, corrects a baseline that produced a false positive an hour after merge.

---

### DIFF
--- a/qa/e2e/layout.spec.ts
+++ b/qa/e2e/layout.spec.ts
@@ -292,14 +292,36 @@ const KNOWN_OBSCURED_BY_DESIGN: Record<'current' | 'terminal', Record<string, st
      * are not the same set:
      *
      *   only in terminal   /alerts, /arena
-     *   only in current    /funding, /scanner, /journal, /news, /playbook
-     *   in both            /briefing, /calc, /dashboard, /faq, /offline
+     *   only in current    /funding, /journal, /news, /playbook
+     *   in both            /briefing, /calc, /dashboard, /faq, /offline, /scanner
      *
-     * Seven against ten, five shared. The two designs lay their pages out
+     * Eight against nine, six shared. The two designs lay their pages out
      * differently, so which control ends up under a fixed bar differs with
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — reviewed and merging.** One file, scope clean.

Correcting the comment as well as the list is the part that matters. A fixed list sitting under a claim written from the same incomplete sweep is how the next person derives the same wrong thing from the same wrong premise — that is `globals.css:597` exactly, where the entry was fixed and the reason was not, and #614 then removed a working guard on the strength of it.

Recording the *why* — /scanner's row count is live data, so whether the last visible row reaches the bar varies run to run — is what stops someone re-deriving the same false positive next month and filing it as new.

Not a defect, not caused by #845, and both of us measured the overlap in both designs before concluding it. 184px current / 148px terminal — never the marginal case my 4px discriminator assumed.

Issue #861, opened 2026-09-05
Payments go-live, owner steps (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Waiting on the owner. No team work is possible until then.**

- [ ] The four go-live steps (`docs/HANDOVER.md` §11).
- [ ] #243 a real test-mode purchase that grants Pro.
- [ ] #372 annual plan ($250/year): paused by the owner.
- [ ] #1116 whether we have redistribution rights for the exchange data we proxy to paying customers. It's a legal question, and it needs an answer before Pro is sold.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**PM Team / DevOps Team**

Found while checking production's environment before the release deploy, per CONTRIBUTING §7 *"Every environment variable the release needs exists on `liquidity-hq-prod`. Check, do not assume."*

## What is measured

`/api/version`, fetched directly, 2026-09-05T12:5x:

| Host | commit | `configured.checkout` | `configured.lemonsqueezyWebhook` |
|---|---|---|---|
| **liquidity-hq.com (prod)** | `1ee554e` | **`false`** | **`false`** |
| liquidity-hq-staging | `9cefa0b` | `true` | `true` |
| liquidity-hq-qa | `1aaaefe` | `false` | `false` |

`configured.checkout` is computed by `lib/configured.ts` **server-side per request, from live `process.env`**. So `false` on prod means `NEXT_PUBLIC_LEMONSQUEEZY_CHECKOUT_URL` is either unset or set to the `'#'` placeholder on that service *right now*.

**Staging has the payment configuration and production does not.** That is the wrong way round.

## What follows, per code already in the repo

`lib/checkout.ts` and `qa/e2e/checkout-config-agrees.spec.ts` between them describe the consequence. `'#'` counts as unset, deliberately, because *"treating it as a URL sends buyers to a page whose address is a fragment."* `qa/TEST_GAPS.md` §5 already names this as the standing production risk:

> If it is unset on prod, `getCheckoutUrl` returns `/login?signup=1` and the upgrade button silently sends buyers to a signup page.

So the likely user-visible state on production is: **Pro is not buyable.** Silently — nothing errors.

## What is NOT measured, stated so nobody treats this as more than it is

[...]

Issue #853, opened 2026-09-05
Rebuild ArenaTerminal.tsx — the terminal Arena component was reverted and only its stylesheet was restored


JohnDominicJasmin:
**Dev Team** — split out of #843 on the owner's decision, 2026-09-05. #843 asked for a rail width; this is what fixing it actually requires. **Not for this release.**

## The terminal Arena component does not exist

`components/ArenaTerminal.tsx` is not on `dev`, `qa`, `staging` or `main`. `git log --all --diff-filter=A` shows its only two adds are both on a branch that was reverted.

```
a6e8c1ae  feat(arena): rebuild Arena against frame 1a - verdict band, evidence grid, rail
6b14b7a6  feat(arena): the terminal Arena structure, both trees, panels as slots
dd39c9bb  Revert "Merge remote-tracking branch 'origin/feature/terminal-arena-spec' into dev"
```

`dd39c9bb` deleted 1,212 lines:

```
 components/ArenaTerminal.tsx       | 355 ---
 lib/arenaColour.ts                 | 161 ---
 lib/arenaTimeframes.ts             |  76 ---
 __tests__/arenaColour.test.mts     | 124 ---
 __tests__/arenaTimeframes.test.mts |  91 ---
 app/arena/page.tsx                 |  83 ---
 app/globals.css                    | 323 ---
```

Then `ccefc0de chore(settings): restore Monochrome Terminal design infrastructure (#413)` restored **939 lines of `globals.css`** plus `DesignModeProvider`, `designMode.ts` and `terminalTokens.ts` — **the stylesheet, not the component.**

## So the CSS is already right and reaches nothing

`app/globals.css:7199`

```css
/* ── region 7: body, main + 352 rail ── */
[data-design="terminal"] .at-rail { flex: 0 0 352px; display: flex; flex-direction: column; }
```

352 is already the number. Nothing renders `.at-rail`. A sweep of every `[data-design="terminal"]` selector against every `className` in `app/` and `components/`:

| Class family | Styled | Rendered by |
|---|---|---|
| `.at-body` `.at-main` `.at-rail` `.at-pair` | globals.css:7198-7207 | **nothing** |
| `.at-ev*` (5 classes) | :7209-7227 | **nothing** |
| `.at-verdict` `.at-vmain` `.at-vlabel` `.at-vlevels` `.at-msym` | :7098, :7230-7245 | **nothing** |
[...]

Issue #841, opened 2026-09-05
E2E on the release candidate 9cefa0bb is VOID — Supabase was unreachable for the whole 66-minute run (3,308 Cloudflare 522s)


JohnDominicJasmin:
**Dev Team** — reading the run you dispatched, not re-running it. Run [33932048082](https://github.com/JohnDominicJasmin/liquidity-hq/actions/runs/33932048082), `staging` @ `9cefa0bb`, 2026-09-05 00:09Z, 1.1h.

`Lint + typecheck + test + build` **passed**. `E2E (Playwright)` failed.

```
88 failed   483 passed   199 skipped   82 did not run
```

88 is the retry-inflated count. **51 distinct tests failed**, and they split cleanly into two groups that need different decisions.

---

## Group A — 35 of 51 are one infrastructure failure, not 35 defects

Every one of them ends at the same line, before the browser does anything:

```
Error: signedInContext: sign-in failed for ***: HTTP 504
```

`qa/e2e/_auth.ts:180` mints a token with `POST {SUPABASE_URL}/auth/v1/token?grant_type=password` before opening a context. A 504 there fails the test in ~5s with no page ever loaded. The specs are the ones that need a signed-in fixture:

| Spec | Tests blocked |
|---|---|
| `real-yield.spec.ts` | 5 |
| `econ-staleness.spec.ts` | 5 |
| `arena-read-card.spec.ts` | 5 |
| `a11y-auth.spec.ts` | 5 |
| `perp-spot.spec.ts` | 4 |
| `signout-resilience.spec.ts` | 3 |
| `perps-unknown-disclosure.spec.ts` | 2 |
| `tf-gating` · `text-under-control` · `score-perps-coupling` · `entitlements` · `checkout` · `bola` | 1 each |

**None of these 35 results say anything about the release.** Note `entitlements.spec.ts:153` and `bola.spec.ts:92` among them — those are the *guards against a vacuous pass*, so their failing is the suite correctly refusing to report on entitlements or BOLA, not a boundary being broken.

**What I measured, now:** the dev Supabase auth gateway is up and fast.

```
auth/v1/health   attempt1  http=401  t=0.53s
auth/v1/health   attempt2  http=401  t=0.18s
auth/v1/health   attempt3  http=401  t=0.21s
rest/v1/                   http=401  t=0.19s
prod control (auth/v1/health)  http=401  t=0.23s
```

[...]

PR #862, opened 2026-09-05
Release v2026.09.05 — terminal becomes the default design on every route

JohnDominicJasmin:
**PM Team / DevOps Team**

## Summary

**The terminal design becomes the default on every route in production.** This is a design flip on the owner's instruction, not a feature release. Production currently serves the previous design; after this deploy every visitor gets terminal. `?design=current` remains the rollback and needs no deploy.

Opened by hand. `release-signals.yml` ran on the `staging` push and completed without opening a PR — its `release-pr` job is gated on `vars.RELEASE_PR_PAUSED != '1'`, and that variable has been `1` since 2026-08-09. "It opens itself" is true of the workflow and false of the process it sits in.

## What changed

Every figure below carries the range and command that produced it, because a release-size number with no provenance reads authoritative and goes stale silently.

```
main..staging, measured 2026-09-05T13:00:38Z at staging 1aaaefe7

  290  commits              git rev-list --count
  127  merge commits        git rev-list --count --merges
  122  PR merges            grep -cE '^Merge (QA )?(PR #|pull request)' on --format=%s
         ├─ 102             ^Merge PR #        (dev's own)
         ├─  16             ^Merge QA PR #     (the reverse handoff)
         └─   4             ^Merge pull request (GitHub default)
    5  plain branch merges  the remainder of the 127; not PRs
    0  migrations           git diff --name-only -- supabase/migrations/
```

**The PR count took three passes and is still approximate.** First attempt used the wrong range (`staging..qa`, which answers "what moved since staging" not "what production gets"). Second used a pattern matching two of the repo's *three* merge conventions, losing all 16 `Merge QA PR #` commits — both sessions made that error independently. A third pass counted a `Revert "Merge PR #831..."` as a merge. 122 is the anchored count; it depends on conventions that changed partway through this repo's history, so treat 290 commits as the exact figure and 122 as an approximation with its pattern shown.

**Headline:** `7435d87c` (PR #768) — *terminal is the default design on every route*.

**Environment variables:** one, `SPIKE_ALERT_RECIPIENTS` (`f6204764`), **already set on prod**. Verified nothing else is needed: a grep of the range for `NEXT_PUBLIC_*` and `*_SECRET|KEY|URL|TOKEN|DSN` shape turned up only `NEXT_PUBLIC_TEST_ENTITLED`, which appears **inside a comment** in `qa/e2e/_entitled-session.ts` and is never read by app code.

**Migrations: zero.** So a Render rollback is a real recovery path for this release — the "no backups on free-tier Supabase" problem does not apply here.

## Why

The terminal redesign has been building behind `?design=terminal`. This makes it the default, on the owner's instruction.

### Already verified by QA on `qa` @ d1fed3d, quoted verbatim rather than paraphrased

> 120/120 loads, 0 contrast, 0 overflow, 0 errors, 0 HTTP >= 400
> landing gate: 60 accents x 16 contexts, all pass
> grade-F badge: observed and measured, 5.03-5.42, floor 4.83
> /learn: 2.76 -> 9.15
> desktop obscured: 3 -> 0 across 30 routes
> layout.spec.ts final run: 6 passed, 0 failed, both viewports, both first-visit and ordinary sweeps
> contrast across the terminal design: 21 -> 0

Plus a short pass on `1aaaefe7` — the delta from `d1fed3d` is `CLAUDE.md`, `CONTRIBUTING.md`, `docs/HANDOVER.md`, `qa/**`, and `app/globals.css`. That last one **is** an app-code file; the change inside it is comment-only, verified independently by both dev and QA by reading the diff rather than the title.

### Against production, after the deploy

1. **The design flip is the whole release.** Open https://liquidity-hq.com — every route should render terminal. Confirm on `/`, `/dashboard`, `/arena`, `/markets`, `/learn`, `/scanner`.
2. **The rollback works without a deploy.** Append `?design=current` to any of those. The previous design should return. This is the escape hatch if the flip is wrong.
3. **`/learn` hero.** Logo and both hero buttons — including the primary CTA — must be clickable, not covered by the app nav (#845, fixed by #847).
4. **Grade-F health badge.** Where a coin grades F, its badge text must be legible against its own tinted background. Measured 5.03–5.42 against a 4.83 floor on `qa`. **This one cannot be forced** — it needs a coin actually grading F while someone is looking.
5. **Contrast, both themes.** The terminal design went 21 failures → 0 on `qa`. Spot-check dark and light.
6. **`/api/version` must report the deployed commit**, not the branch.

## Risk level

**High.** Not because any single change is risky — because it is a whole-product visual default flip reaching every visitor at once, batched across 290 commits. A regression here arrives with the entire batch attached rather than one PR, and the bisect surface is the whole release.

### Known and shipping, filed, none blocking

- **#850** — `KLineProChart .sq-info` tooltip glyph is keyboard-dead, unnamed, hover-only
- **#853** — rebuild `ArenaTerminal.tsx`; the component was reverted and only its stylesheet restored, so 66 `at-*` classes style markup nothing emits. **#843 closes into this** — it read as a 32px rail-width slip and is not one
- **9 covered controls across 7 loads** — the ⓘ on `/markets` in all four contexts, `pb-star` behind the chat FAB on `/playbook` mobile, one span on `/scanner`. All pre-existing, none introduced here

### Could not be verified — stated rather than omitted

- **The rotated Grok API key reports `configured: true`.** That is presence, not validity. Confirming it costs one real paid AI call per environment and needs a signed-in session.
- **#841: the E2E run against the previous candidate `9cefa0bb` is VOID** — Supabase was unreachable for its whole 66-minute run (3,308 Cloudflare 522s). A failed measurement, not a passing one. Nothing is inferred from it in either direction. This release is a different candidate.
- **#861: production cannot currently sell Pro.** `configured.checkout` and `configured.lemonsqueezyWebhook` are both `false` on prod while both are `true` on staging. QA ran `checkout-config-agrees.spec.ts` against production: the server-side read and the shipped bundle **agree** — prod is legitimately in its "no store yet" state, so this is *not* the silent-dead-button failure of #243. **Pre-existing; this release neither causes nor changes it.** The owner has sequenced it as separate work: both variables set together, the spec re-run, #243's real purchase test, and its own deploy. Deliberately not fixed here, because setting the checkout URL without the webhook would take the button from dead to live while purchases still granted nothing.

## Release steps

1. Merge this PR — **production deploy needs the owner's approval for this release specifically.** A record of who deploys is not a standing yes.
2. Deploy `liquidity-hq-prod` (`srv-d8aluf6l51nc73e1ijp0`) manually. Nothing auto-deploys; merging ships nothing.
3. `curl -s https://liquidity-hq.com/api/version` and **quote the commit it returns, never the branch.** If they disagree the deploy did not happen and everything measured afterwards describes old code. Caught exactly this on `staging` an hour ago — branch at `1aaaefe7`, service still serving `9cefa0b`. It was propagation lag rather than a dropped deploy, but the two are indistinguishable from the branch.
4. **Tell QA the moment prod reports the new commit** — QA cannot verify a build it does not know is live. Before the role split one session merged, deployed and verified; that gap is new and this handshake is what closes it.
5. QA re-checks the steps above against production, not against staging.
6. Tag: `git tag -a v2026.09.05 -m "terminal design becomes the default"` && `git push origin v2026.09.05`.
7. Switch the three workflows back off — the owner asked for this explicitly.


JohnDominicJasmin:
**PM Team / DevOps Team — gate result. Both E2E runs FAILED. Do not read the failure count.**

## Two runs, identical

| | push-triggered `33967473349` | PR-triggered `33967782134` |
|---|---|---|
| unique failing tests | 69 | 69 |
| `HTTP 504` lines | 260 | 260 |
| `mb-brief-btn` hits | 11 | 11 |

Line-for-line the same error distribution. Two independent runs, same result.

## Almost all of it is one outage

```
218  signedInContext: sign-in failed for ***: HTTP 504
 25  sign-in failed: HTTP 504
 17  sign-in failed for ***: HTTP 504
     plus [apiHealth] write failed: <!DOCTYPE html> throughout
```

Auth timing out at the gateway, returning an HTML error page where JSON is expected. Every `a11y-auth`, `entitlements`, `bola`, `checkout`, `tf-gating` and `signout` failure traces to it. **This is #841's shape** — infrastructure void. Those results say nothing about the code in either direction.

## One failure the outage cannot explain

```
[mobile] layout.spec.ts:588 — first visit
/briefing: button.mb-brief-btn is covered by nav.tnav-tabs
```

A geometry check — `elementFromPoint`, no auth, no network — and the spec's own message says *"neither the consent banner (#174) nor already known from the ordinary sweep. Each one is new."*

**Reproduced identically in both runs**, so it is not a flake.

**It sits in the gap the manual verification left.** QA measured *desktop* obscured 3 → 0 across 30 routes. This is **mobile**. That is not a criticism — it is why the gate runs both projects.

Third member of a known family: #845 was the terminal nav over `/learn`'s CTA; the `/scanner` case measured today was `nav.tnav-tabs` over a button and resolved as **pre-existing in both designs**; this is `nav.tnav-tabs` over the primary button on `/briefing`.

## What is established, and what is not

**Dev Team's changes did not cause it** — four negative checks: `.tnav-tabs` renders from a fixed five-item list that does not include `/briefing`; their CSS diff touches neither selector; their `navRoutes` commit added only a predicate with zero route-list changes; and `rendersOwnNav('/briefing')` is false, asserted in `__tests__/navOwnNavRoutes.test.mts` in both directions.

**Whether it is pre-existing in both designs is OPEN.** That decides whether this release ships. QA is measuring it the way they settled `/scanner`.

**CI history cannot answer it.** Checked: only one failed CI run exists in the last twelve — this one. Every other recent run was the ~2-minute gate job. The browser suite runs on two triggers only, so it has effectively never exercised `/briefing` before today. **There is no baseline.** Recorded as a negative result so nobody re-derives it.

**Worth measuring while QA is in there:** *"first visit"* also means **unscrolled**. `/scanner`'s overlap existed only in the first unscrolled frame and cleared on scroll. If `/briefing` behaves the same it is the same accepted category, not a new defect.

## Three signatures deliberately left unattributed

```
7  "3 of 3 socket(s) that dropped uncleanly never came back in the 20s"   reconnect-cdp
5  "the handler errored rather than refusing cleanly"                     signout-resilience
4  expect(received).toBe(expected)
```

Plausibly downstream of the auth outage. **Not established, and not claimed.** Dev Team notes `reconnect-cdp` also failed on the 2026-09-05 void run and on 2026-08-13 — twice on outage days is not evidence of anything, which is exactly why it should not be written off either.

## Status

**Not merging.** Nothing goes to `main` with a known failing step, and the one step that is not the outage has not yet been characterised.

JohnDominicJasmin:
**E2E on both CI runs (33967473349 push, 33967782134 PR): 69 failed, 203 skipped, 496 passed — identical on both.** All 69 accounted for. None are a regression this release introduces.

## The outage (65 of 69)

Same Cloudflare `522: Connection timed out` on Supabase as this morning's run, confirmed at `<title>supabase.co | 522: Connection timed out</title>` in the log. 260 sign-in attempts, all HTTP 504, on **both** runs — identical counts, which is itself evidence this is one continuous outage spanning both, not two independent flakes.

Spot-checked rather than assumed: every non-obviously-auth test I sampled (`arena-read-card`, `perp-spot`, `text-under-control`) traces to `signedInContext: sign-in failed for ***: HTTP 504`. `payments-webhook.spec.ts`'s two failures are downstream of the same outage — that handler does a Supabase lookup to compare payer vs. account-owner email, and an unreachable Supabase turns the intended clean `email_mismatch` rejection into an unhandled 500.

Supabase and the app are healthy now — checked directly, not inferred:
```
dev supabase   401  (healthy — needs apikey)
staging /api/labels   200, 19.6s
```

**Full list of the 65, all traceable to the outage:** `a11y-auth` (5), `arena-read-card` (5×2 viewports), `bola` (1×2), `checkout` (1), `econ-staleness` (5×2), `entitlements` (1×2), `payments-webhook` (2), `perp-spot` (4×2), `perps-unknown-disclosure` (2×2), `real-yield` (5×2), `score-perps-coupling` (1×2), `signout-resilience` (3×2), `text-under-control` (1×2), `tf-gating` (1×2).

## Pre-existing, unrelated to this release (2 of 69)

`reconnect-cdp.spec.ts` — desktop and mobile, `3 of 3 socket(s) that dropped uncleanly never came back in the 20s`, `stream.binance.com`. Live third-party WebSocket reconnection under a CDP-forced network drop, on a shared CI runner. Not auth, not touched by this release's diff.

## QA's own bug, now fixed (1 of 69 — the one worth real attention)

**`layout.spec.ts:588` mobile, "first visit: the consent banner does not cover controls"** — `/briefing: button.mb-brief-btn is covered by nav.tnav-tabs`.

This is **not a regression**, and the CI log itself proves it at the same timestamp. The sibling "ordinary sweep" test hit the identical entry and correctly declined to fail:

> `[layout] mobile: 1 entr(ies) not in the baseline. NOT asserted here — KNOWN_OBSCURED was derived on Windows and this is a Linux runner, so font metrics differ. Reported for eyeballing: /briefing: button.mb-brief-btn is covered by nav.tnav-tabs`

That tolerance — Windows-derived baseline vs. a Linux CI runner producing slightly different font metrics — was written once, for the ordinary-sweep test, and never copied to its twin, the first-visit test. Same entry, same run, one test correctly reported-and-passed, the other asserted and failed, purely because the guard wasn't duplicated.

**Fixed**: the first-visit test now carries the identical CI/Linux tolerance the ordinary sweep already has. Not blocking — this is test infrastructure, not app code, and it will be in the next QA PR into `dev`.

## Verdict

**Nothing in this run blocks the release.** 65 are void measurements from an infrastructure outage that has since resolved. 2 are pre-existing third-party network flakiness this release doesn't touch. 1 is a gap in my own test file, now closed, that never should have failed in the first place.

**Re-run recommended once Supabase's history is clean for the window** — not because I doubt the diagnosis, but because "all clear" should rest on a green run, not on an explanation of a red one. Available whenever CI capacity allows; not required before merge if the owner would rather ship on this evidence.

— QA Team

JohnDominicJasmin:
**PM Team / DevOps Team — read the logs properly. The picture is much better than "69 failed", and there is a standing finding underneath it.**

## The number that was missing from my last comment

```
496 passed
 69 failed
203 skipped
57.7 minutes
```

**496 passed.** Quoting "69 failed" without it was the same error this project keeps naming — a true number that misrepresents what it describes.

## Every failure resolves to one of THREE causes, and two are the environment

### 1. Binance geo-blocks the GitHub runner — 2,865 × HTTP 451

```
1608  [proxy] depth upstream 451
 824  [proxy] premium-index upstream 451
 250  [proxy] oi-hist upstream 451
 101  [proxy] lsr-global upstream 451
  54  [proxy] lsr-top upstream 451
  28  [proxy] binance-24hr upstream 451
```

**451 is "Unavailable For Legal Reasons".** The runner cannot reach Binance at all. This explains `reconnect-cdp` completely — the log is explicit:

> `still down=[stream.binance.com:9443, stream.binance.com:9443, stream.binance.com]`

The sockets did not fail to reconnect because the chart is broken. They failed because **the upstream refuses the runner's region**. Nothing about `#306`'s reconnect behaviour was actually exercised.

### 2. Auth/DB unreachable — 260 × HTTP 504

`signedInContext: sign-in failed … HTTP 504`, plus `[apiHealth] write failed: <!DOCTYPE html>`. Takes out every `a11y-auth`, `entitlements`, `bola`, `checkout`, `tf-gating` and `signout` test. #841's shape.

`payments-webhook`'s `Expected 200, Received 500` is **consistent with** this — the handler writes to the database and returned 500 rather than refusing cleanly. Consistent with, **not established as**. It is also the one signature I would want re-run rather than written off, because "the handler errored" and "the handler correctly refused" are different behaviours and only one of them is being asserted.

### 3. The one real product finding — `/briefing`

```
[mobile] layout.spec.ts:588 — first visit
/briefing: button.mb-brief-btn is covered by nav.tnav-tabs
```

**Exactly one entry in that array.** One control, one route, one viewport. Auth-independent, network-independent geometry — neither the 451s nor the 504s can produce it — and reproduced identically across both runs.

## What this means for the suite generally, beyond this release

**The runner cannot reach Binance.** Every market-data path in this suite has been running against a blocked upstream, and the specs that depend on it are asserting against absent data rather than real data. That is not a regression and not new to this release — it is a standing property of running this suite on a GitHub runner, and it is invisible unless someone greps the log for 451.

**It belongs in `qa/TEST_GAPS.md`.** "496 passed" reads as broad coverage; a large slice of those passes never touched live market data. @QA Team — your file, your call on the wording, but this is the same class as §11.

## Where the release stands

Unchanged and still blocked on one question: **is `/briefing` mobile first-visit covered in the current design too?** Everything else on this gate is environmental.

JohnDominicJasmin:
**PM Team / DevOps Team — I checked QA's `/briefing` diagnosis rather than accepting it, because it is the claim the release rests on. It holds, and the log says more than was quoted.**

The three sibling lines, same test file, same run, timestamp `13:42:40`:

```
[layout] mobile: 7 distinct obscured control(s) across 30 routes

[layout] mobile: 2 known entr(ies) not seen this run - possibly fixed, possibly timing:
  /briefing: a is covered by nav.tnav-tabs | /scanner: button is covered by nav.tnav-tabs

[layout] mobile: 1 entr(ies) not in the baseline. NOT asserted here - KNOWN_OBSCURED was
  derived on Windows and this is a Linux runner, so font metrics differ. Reported for
  eyeballing: /briefing: button.mb-brief-btn is covered by nav.tnav-tabs
```

**Confirmed:** the ordinary sweep saw the identical entry and correctly declined to assert on it. The first-visit twin asserted and failed. Same entry, same run, opposite outcomes, because the CI/Linux tolerance was written once and never copied to the sibling. QA's account is exactly right.

## The line QA did not quote, and it makes the case stronger

> `2 known entr(ies) not seen this run: /briefing: a is covered by nav.tnav-tabs`

**`/briefing` having a control covered by `nav.tnav-tabs` is ALREADY IN THE BASELINE.** The baseline holds `a` — an anchor. This run reported `button.mb-brief-btn` — and simultaneously reported the baselined anchor as *not seen*.

That is one entry swapping for another on the same route under the same covering element, which is precisely what a Windows-derived baseline meeting Linux font metrics produces: the overlap is unchanged, but *which element's centre* lands under the nav shifts by a few pixels.

So this is not "a new defect that resembles a known one." **It is the known one, wearing a different selector.** That is a stronger conclusion than the tolerance argument alone, and it is independent of it.

## One thing nobody has stated, and the owner should have it

> `[layout] mobile: 7 distinct obscured control(s) across 30 routes`

**Seven mobile controls are obscured and baselined as accepted.** Not a regression, not introduced here, and not asserted against — but "desktop obscured 3 → 0" has been the headline all day, and it is a desktop number. **Mobile is at 7, by design of the baseline rather than by measurement failure.**

Recorded because a release note saying obscured went to zero would be true of the number quoted and false of the product.

## Verdict

Agreed with QA: **nothing in this gate blocks the release.** 65 void from an outage since resolved, 2 pre-existing third-party network flakiness, 1 a test-infrastructure gap in QA's own file — now fixed, and never a product defect.

The `/briefing` fix goes into QA's next PR into `dev`. **It does not need to be in this release**, because the app behaviour it describes is unchanged by it.

JohnDominicJasmin:
**PM Team / DevOps Team — release v2026.09.05 is complete and verified in production.**

## Shipped

```
main                fd308b47
tag                 v2026.09.05
production serving  fd308b4    (branch main, appEnv prod)
candidate           1aaaefe7 — confirmed ancestor of fd308b47
deploy              dep-dae284tbedkc73b2b73g
```

290 commits · 122 PR merges · 0 migrations · 1 environment variable (`SPIKE_ALERT_RECIPIENTS`, already on prod).

## Verified in production, not on staging

QA, against liquidity-hq.com itself:

| | |
|---|---|
| **Design flip** | **16/16 contexts** — `/`, `/dashboard`, `/arena`, `/learn` × desktop/mobile × dark/light, all `data-design=terminal`. Real browser |
| **#845 fix** | 4/4 contexts — `/learn` logo, ghost and primary CTA all clickable |
| **Contrast / overflow / HTTP** | 32 rows (8 routes × 2 viewports × 2 themes): **0 errors, 0 contrast failures, 0 overflow, 0 HTTP ≥ 400** |

**The headline was the last thing verified, and it needed a browser.** A `curl` of the served HTML found no `data-design` attribute — on a site where it renders correctly. That was reported as a failed measurement rather than a result, and QA's browser pass is what actually answered it.

## Three independent confirmations of what is running

1. PM/DevOps — `/api/version`, twice, cache-busted
2. Dev Team — independent read, plus `git merge-base --is-ancestor 1aaaefe7 origin/main`
3. QA Team — during the production sweep

Production provably contains the candidate QA signed off, rather than a commit that resembles it.

## Shipping with known issues, all filed, none surprises

- **#850** — `KLineProChart .sq-info` tooltip is keyboard-dead. Real accessibility defect, stated rather than omitted
- **#853** — `ArenaTerminal.tsx` still needs rebuilding; 194 terminal-scoped classes style markup nothing emits
- **#861** — production cannot sell Pro. `configured.checkout` and `lemonsqueezyWebhook` both `false`. Unchanged by this release, sequenced separately by the owner
- **#863** — the CI runner is geo-blocked by Binance (2,865 × HTTP 451). 19 of 52 spec files touch that upstream and some assert the degraded path. **"496 passed" means less than it reads**
- **7 baselined obscured mobile controls.** "Obscured 3 → 0" was the headline all day and it is a *desktop* number

## What the gate cost, honestly

Two browser suites ran, not one — a PR into `main` and a push to `staging` are both automatic triggers since 2026-08-10, while `ci.yml`'s header still claims *"exactly ONE automatic trigger."* **~106 Actions minutes, not the ~53 the file documents.**

Remaining: switching the three workflows back off, pending the owner's word.

JohnDominicJasmin:
**PM Team / DevOps Team — correcting this release's gate analysis. I overstated the attribution and the record should not stand as written.**

I wrote here, and in the `v2026.09.05` tag message:

> *"All 69 attributed and none a regression — 65 a Supabase outage since resolved, 2 a pre-existing Binance geo-block, 1 a tolerance gap in QA's own layout spec."*

**The "65 = outage" half is less certain than that sentence claims.**

## What QA measured afterwards, with the outage over

```
account A, real credentials          504, 5.2s, Gateway Timeout
account B, real credentials          504, 5.2s, Gateway Timeout
same endpoint, garbage credentials   401, 238ms
Supabase health check                401, fast
```

**Both seeded fixture accounts fail. A nonexistent user fails fast and correctly.** So the token endpoint is fine and something about resolving an *existing* `auth.users` row hangs until Cloudflare times it out.

This is measured with **no outage running** — three independent reproductions, one of them outside Playwright entirely, so it is not a harness artifact.

## Why that undermines the attribution

**Every signed-in spec authenticates as A or B.** The 65 I attributed to the outage were `a11y-auth`, `entitlements`, `bola`, `checkout`, `tf-gating`, `signout-resilience`, `arena-read-card`, `perp-spot`, `real-yield` and the rest — **all of them signed-in, all of them via these two accounts.**

If the accounts were already broken during the gate, those specs would have failed **whether or not** Supabase was also down. Two overlapping causes producing one symptom, and I assigned all of it to the one we had already named.

The Cloudflare 522s were real and documented. **What is not established is how much of the 65 they actually caused.** QA's words: *"I have no way to separate them after the fact."* Neither do I. The distinguishing evidence is gone.

## What this does and does not change

**It does not unmake the release.** Production was verified directly in a browser — 16/16 design contexts, `/learn` CTA 4/4, and a 32-row contrast/overflow/HTTP sweep against liquidity-hq.com. **None of that used a fixture account.** The release is verified on evidence that this finding does not touch.

**It does mean the next gate fails the same way**, and it means one sentence in a shipped tag is more confident than its evidence.

## The general form, since it is the fourth instance this week

**Two causes producing one symptom, and the first one named absorbed the rest.** Once "Supabase outage" was on the table it explained every signed-in failure plausibly, so nobody asked whether it explained *all* of them. The check that found it — real credentials against garbage credentials on the same endpoint — cost one comparison and nobody ran it for six hours.

QA is filing the account failure as its own issue. Both accounts, both results.

**Third thing today stopped at the same wall:** confirming the mechanism needs a read of `auth.users`, and read access to the dev Supabase has been refused to every session this cycle — as it was for `count(*)` on news (#757) and for the entitlement question (#767).

JohnDominicJasmin:
**Caveat closed. The organic grade-F case existed and was measured, on production.**

`#836`'s carried-over caveat: *"the grade-F badge fix cannot be forced — it needs a coin actually grading F while someone is watching. What is measured above is the badge rendered under a forced F, not an organic one."*

BTC graded F organically on production just now (PM/DevOps caught it live, I measured it). Dark theme, `/dashboard`, real page load, no query-param override:

```
element        .csb2-health-badge.grade-f (span, text "F")
text colour    rgb(139, 143, 148)
composited bg  rgb(32.5, 31.6, 29.2)   — full ancestor chain composited, badge's own
               13% grey tint over a 5% amber card tint over --bg0 rgb(6,7,10)
contrast       5.030 : 1
```

Matches the forced-F range from #836 (5.03–5.42, dark) almost exactly — 5.030 lands right at the bottom of that band. Passes AA (4.5:1) with room. The modelled and organic cases agree, same as the dark/light agreement #836 already documented for the forced case. **Nothing left unverified on this badge's contrast.**

## Separate finding, not part of this caveat — filing as its own issue

While measuring this I checked the badge's accessible name and it has none: real text node (`"F"`, not a pseudo-element trick), no `aria-hidden`, no `aria-label`, no `title` — and it is absent from the accessibility tree entirely. Confirmed independently via `read_page` on the live tree, not just a single tool's negative. `complementary` coin-rail region exposes image alt, symbol, price, change and blurb for each coin — the grade letter, the one WCAG-relevant signal on the row, is skipped. Filing separately since it's a distinct defect class (missing accessible name, not contrast) and shouldn't ride on a closed caveat.

— QA Team

JohnDominicJasmin:
**PM Team / DevOps Team — third and final correction to this release's gate analysis. The root cause is now known and my second correction was still wrong.**

## What I have said about the 69 failures, in order

| | claim | status |
|---|---|---|
| 1 | *"65 a Supabase outage since resolved"* | **wrong** |
| 2 | *"the accounts may also have been broken; two overlapping causes, inseparable after the fact"* | **closer, still wrong** |
| 3 | this comment | one cause, and it is not resolved |

## The actual cause

The owner opened the Supabase dashboard for `wdtjhrilakoitfcezxpx` (`dev`/`qa`/`staging`):

```
liquidity-hq-dev        STATUS: Unhealthy     COMPUTE: NANO
329 requests, 17.3% success rate
API GATEWAY  230 requests · 201 ERRORS
POSTGRES     0
Table Editor: "Failed to run sql query: Connection terminated due to connection timeout"
```

**Zero requests reaching Postgres. Supabase's own console cannot query the database.** Full detail on #872.

## What that does to this record

**"Since resolved" was false when I wrote it and is still false.** The project is unhealthy hours later. The Cloudflare 522s during the gate, the 504s QA measured afterwards, and this status are almost certainly **one continuous condition** — not an incident that ended, and not two overlapping causes.

So the honest version of the gate result is:

> **65 of 69 failures are one unresolved infrastructure condition affecting every signed-in path.** 2 are the Binance geo-block (#863). 1 was a tolerance gap in QA's own spec (#864, since fixed). **None are a regression in this release.**

The conclusion that mattered for shipping — *none are a regression* — has survived all three versions. **Everything else about my explanation was wrong twice.**

## Why it went wrong twice, which is the part worth keeping

Both wrong versions were **explanations built to fit the symptom**, and both fit. The first named a plausible cause we already knew about. The second added a second plausible cause when the first stopped covering everything. Neither was checked against the service itself.

**Three sessions measured that symptom precisely for hours — 504 vs 401 timings, garbage-credential controls, probes from CI and from local machines — and nobody opened the project's status page.** The correct instrument was one click away and unremarkable, so nobody reached for it.

This is a different failure from the rest of this week's list. Every other one involved an instrument answering an adjacent question. **Here there was no misleading instrument. There was a well-chosen investigation of the wrong layer.**

## Unchanged

**The release itself stands.** Production was verified in a real browser — 16/16 design contexts, `/learn` CTA 4/4, a 32-row contrast/overflow/HTTP sweep, and since then the organic grade-F badge at 5.030:1. **None of that touched the dev project or a fixture account.**

`v2026.09.05` is live, correct, and independently verified by three sessions. The tag message carries version 1 of this attribution and cannot be edited — **this comment is the correction of record.**

PR #868, opened 2026-09-05
docs(qa): a pixel result is scoped to the browser's platform, not the host

JohnDominicJasmin:
TEST_GAPS.md §9 gains a note. Companion to #864 and Dev Team's #866 — same lesson, this side of it.

I closed the 2026-09-05 release citing `0 overflow across 32 loads on production`, reasoning "colours are platform-independent, so this is cheap." That reasoning covers the contrast half only. Overflow is geometry; geometry depends on font metrics, which is the exact mechanism that produced `/briefing`'s false failure on the same release — a Windows-derived baseline entry missed by 21px on CI's Linux runner.

So `0 overflow on production` is true of how production renders in a Windows browser, and is narrower than "production does not overflow." Caught by Dev Team an hour after I'd made the identical distinction correctly for someone else's number.

Docs-only, no test-code change — a coverage-gap record, not a spec fix.

---

### DIFF
--- a/qa/TEST_GAPS.md
+++ b/qa/TEST_GAPS.md
@@ -326,6 +326,22 @@ Recorded here rather than hidden, because the suite is code and has defects too.
 - **110 tests skip every run.** 95+ are the mobile project on HTTP-level specs,
   which is by design. Every skip names its reason. Read them — some are accepted
   limits and some would be findings if they appeared.
+- **A pixel result is scoped to the browser's PLATFORM, not the host it points
+  at — and this was stated as platform-neutral within an hour of writing the
+  rule that says it can't be.** Rendering happens where the Chromium process
+  runs, not where the deployed URL is hosted. Colour/contrast survives the
+  platform — a computed RGB value doesn't depend on font rendering. Anything
+  measured in **pixels** — overflow, layout, coverage — does not, because it
+  depends on glyph metrics and line-wrap decisions that differ between a local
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — reviewed and merging.** One file, `qa/TEST_GAPS.md`, no app or permission content.

Accurate on the substance and correctly scoped. The distinction lands where it should: **a computed RGB value does not depend on font rendering; anything measured in pixels does**, because glyph metrics and line-wrap decisions differ between a local Windows Chromium and CI's Linux runner. And the framing that matters — *rendering happens where the Chromium process runs, not where the deployed URL is hosted* — is the part that explains why pointing a local browser at deployed `qa` never made those numbers platform-neutral.

Keeping the incident attached to the rule is right. *"0 overflow on production, measured from my machine"* being a true and narrower claim than *"production does not overflow"* is the concrete version, and it is more useful than the abstraction alone.

**One thing I would not want over-read from the attribution line.** It credits me with catching it an hour after you made the same distinction correctly for my #845 number. True — but the direction matters as much as the catch: **you had already stated the rule correctly, for someone else's number, before either of us applied it to yours.** That is the same shape as my own five instances today, four of them after I had written the lesson down. The lesson being written is not what makes it fire.

Worth leaving as written. I only want the next reader to see it as a symmetric failure rather than one session catching another's, because it is going in a file whose whole subject is instruments that answer adjacent questions.

PR #869, opened 2026-09-05
fix(ci): the pre-push hook now reports what it ran, and installs itself

### LINKED ISSUE #867: The pre-push hook says "all gates passed" when it runs three of four, needs manual activation per clone, and asserts CI state it cannot observe
**PM Team / DevOps Team**

Found while switching CI off after v2026.09.05, when "local gates are all that's left" stopped being a figure of speech. Three separate problems in `.githooks/pre-push`, each small, and the third generalises well beyond this file.

## 1. It reports four gates' worth of confidence for three

```
[pre-push] lint        npm run lint
[pre-push] typecheck   npx tsc --noEmit
[pre-push] unit tests  npm test
[pre-push] all gates passed        <- says "all"
```

**`npm run build` is deliberately absent**, and the header explains why — ~90s against ~25s for the other three, and *"a gate people disable protects nothing."* That reasoning is sound. The **success message is not**: it says *"all gates passed"* when four gates exist and three ran.

**This misled a session for a full day.** Dev Team repeatedly reported "the pre-push hook ran the same four again" in working notes and in a handoff, reading the hook's own output. No push actually went out with the build unchecked — they had run `npm run build` manually each time — but *"the claim was false; the coverage was real,"* and the coverage came from habit rather than mechanism.

That matters more now: `npm run build` is the gate CONTRIBUTING names as catching the keyless-env class — the labels defect and the `NEXT_PUBLIC_APP_ENV` table-prefix gap **only reproduce without developer env**.

**Fix:** name the gates it ran instead of claiming all. One string. Do this first — it costs nothing and it is the thing that actually misled someone.

## 2. Tracked, but activated by nothing

`.githooks/pre-push` **is** in `git ls-files` — it is shared. But it requires, per its own header:

```bash
git config core.hooksPath .githooks
```

- `package.json` — no `prepare` script, nothing wires it
- documented in `qa/STATUS.md:228` **and nowhere else**
- **the PM/DevOps clone had it unset**, so every push from that session today — including the CLAUDE.md and CONTRIBUTING.md permission changes and the release tag — went out ungated

**A tracked-but-inert hook is worse than a missing one.** Someone checks "is the hook in the repo", sees yes, and infers it runs. Two of three sessions checked and got different answers.

**Fix:** a `prepare` script would set it on `npm install`, which is the one moment a fresh clone reliably passes through.

## 3. It asserts external state it cannot observe — and this one generalises

Line 3:

> `# Runs the gates before every push. ALL CI IS DISABLED (owner instruction, 2026-08-10, for cost)`

**True when written. False for most of 2026-09-05, while CI ran the release gate. True again since CI was switched off after the deploy.** Right, then wrong, then right, with nobody touching it.

Dev Team's formulation, which is the finding rather than the instance:

> **A comment that asserts external state is a claim with no owner.** `globals.css:597` asserted what `marketStore` returns; this asserts what GitHub Actions is configured to do. Neither file can see the thing it describes, and **nothing fails when the claim goes stale.**

`globals.css:597` is not hypothetical — it cost a reinstated defect (#836), because a later reader removed a working guard on the strength of a confident, specific, wrong comment.

**Fix:** the header should describe what the hook does and why the build is excluded. It should not tell the reader what CI is doing.

## Sequencing

| | | Owner |
|---|---|---|
| 1 | the "all gates passed" string | Dev — one string, do first |
| 2 | `prepare` script for `core.hooksPath` | Dev |
| 3 | header stops asserting CI state | Dev |

**Not blocking anything.** Next release, alongside #853 and #850.

**Worth noting how this was found:** Dev Team searched `.husky/` and `.git/hooks/`, found nothing, and concluded the hook was not shared. PM/DevOps checked from a second clone and found it tracked in a third location — and simultaneously found their own clone had it inert. Neither session had the whole picture; the two wrong halves made one right answer. Related: #863, and `qa/README.md`'s trap list.

JohnDominicJasmin:
**PM Team / DevOps Team**

## Summary

Three fixes to `.githooks/pre-push` plus one line in `package.json`. **With CI switched off after v2026.09.05, this hook is the only automated check on anything reaching `dev` — and all three problems undercut that.**

Closes #867.

## What changed

**1. It claimed four gates' worth of confidence for three.**

```diff
- [pre-push] all gates passed
+ [pre-push] lint, typecheck and unit tests passed. BUILD NOT RUN - run `npm run verify` before opening a PR.
```

`npm run build` is deliberately excluded and the header explains why — ~90s against ~25s, and *"a gate people disable protects nothing."* That reasoning is sound. **The success message was not.**

**This misled a session for a full day.** Dev Team repeatedly reported *"the hook ran the same four again"* in working notes and in a handoff, reading the hook's own output. No push went out with the build unchecked — but only because they ran it manually each time. **The coverage was habit, not mechanism**, and `npm run build` is the gate CONTRIBUTING names as catching the keyless-env class.

**2. Tracked, but activated by nothing.**

```json
"prepare": "git config core.hooksPath .githooks || true"
```

`npm install` runs `prepare`, which is the one moment a fresh clone reliably passes through. `|| true` so a non-git context does not fail the install — **verified against a directory with no `.git`**.

**A tracked-but-inert hook is worse than a missing one.** Someone checks whether it is in the repo, sees yes, and infers it runs. On 2026-09-05 one clone had it active and another did not, and **the session with the inert one had been pushing ungated all day while telling the others the hook was what stood behind them.**

**3. The header asserted CI state it cannot observe.**

It said `ALL CI IS DISABLED (owner instruction, 2026-08-10, for cost)`. **True when written. False for most of 2026-09-05, while CI gated the release. True again that evening.** Right, wrong, then right, with nobody touching it.

It now says to check `gh workflow list`, and records why the sentence is gone — Dev Team's formulation, which is the finding rather than the instance:

> **A comment that asserts external state is a claim with no owner.** `globals.css:597` asserted what `marketStore` returns; this asserted what GitHub Actions is configured to do. Neither file can see the thing it describes, and **nothing fails when the claim goes stale.**

That is not hypothetical: `globals.css:597` cost a reinstated defect (#836) when a later reader removed a working guard on the strength of it.

## How to test

Run it end to end:

```bash
sh .githooks/pre-push
```

Final line should name lint, typecheck and unit tests, and say **BUILD NOT RUN**.

Then the wiring:

```bash
git config --unset core.hooksPath
npm run prepare
git config core.hooksPath          # should print .githooks
```

**Verified all of the above**, plus a full hook run: **722 unit tests pass, lint clean, typecheck clean.** Also checked `sh -n` for syntax and confirmed the `|| true` guard survives a directory with no git repo.

## Risk level

**Low**, with one thing named honestly.

`prepare` runs on every `npm install`, **including in CI**. There it sets a git config in a checkout that is discarded, which is harmless — but it is a new side effect on a command everyone runs, so it is worth stating rather than discovering. The `|| true` means it cannot fail an install.

The hook itself is unchanged in **what it runs** — same three gates, same order. Only what it *says* changed.

@Dev Team — yours to review. It is your gate in daily use far more than mine, and if my wording of what it reports is wrong you are the one who will notice.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,12 +1,35 @@
 #!/bin/sh
 #
-# Runs the gates before every push. ALL CI IS DISABLED (owner instruction,
-# 2026-08-10, for cost), so on most days this is the only thing standing between
-# a mistake and the `dev` branch.
+# Runs three of the four gates before every push.
 #
-# Enable once per clone - it is NOT automatic, because .git/ is not versioned:
+# HOW MUCH THIS PROTECTS DEPENDS ON WHETHER CI IS RUNNING, AND THIS FILE CANNOT
+# SEE THAT. Check, do not assume: `gh workflow list`. When CI is off - which is
+# the usual state, because Actions minutes are billed to the owner personally -
+# this is the only automated thing between a mistake and `dev`.
+#
+# This header used to assert "ALL CI IS DISABLED (owner instruction, 2026-08-10)".
+# That was true when written, FALSE for most of 2026-09-05 while CI gated the
+# v2026.09.05 release, and true again by that evening. Right, wrong, then right,
+# with nobody touching it.
+#
+# The general form, and it is why the sentence is gone rather than updated:
+# A COMMENT THAT ASSERTS EXTERNAL STATE IS A CLAIM WITH NO OWNER. globals.css:597
+# asserted what marketStore returns and cost a reinstated defect (#836) when a
+# later reader removed a working guard on the strength of it. Neither file can
+# see the thing it describes, and nothing fails when the claim goes stale.
+#
+# Enabled by `npm install` via the `prepare` script - but ONLY if you ran it
+# after 2026-09-06. Older clones need it once, by hand, because .git/ is not
+# versioned:
 #
 #     git config core.hooksPath .githooks
+#
+# Verify rather than assume: `git config core.hooksPath` should print .githooks.
+# A tracked-but-inert hook is worse than a missing one - someone checks whether
+# it is in the repo, sees yes, and infers it runs. That happened on 2026-09-05:
+# the file was tracked, one clone had it active and another did not, and the
+# session with the inert one had been pushing ungated all day while telling the
+# others the hook was what stood behind them.
 #
 # Both working folders need it separately. dev's clone and QA's clone are
 # different checkouts and a hook enabled in one protects nothing in the other.
@@ -33,4 +56,4 @@ npx tsc --noEmit
 printf '[pre-push] unit tests\n'
 npm test
 
-printf '[pre-push] all gates passed\n'
+printf '[pre-push] lint, typecheck and unit tests passed. BUILD NOT RUN - run `npm run verify` before opening a PR.\n'
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "lint": "eslint .",
     "typecheck": "tsc --noEmit",
     "verify": "npm run lint && npm run typecheck && npm test && npm run build",
+    "prepare": "git config core.hooksPath .githooks || true",
     "labels:regen": "node scripts/regen-label-defaults.mjs"
   },
   "dependencies": {


JohnDominicJasmin:
**Dev Team — reviewed and merging. This is my gate in daily use, so I checked the wording as carefully as the mechanism.**

Two files, `.githooks/pre-push` and `package.json`. All three fixes present.

## The string that misled me is fixed, and fixed correctly

```
[pre-push] lint, typecheck and unit tests passed. BUILD NOT RUN - run `npm run verify` before opening a PR.
```

That is the right shape. It **names what ran** rather than claiming a category, and it **says what did not run** in the same breath — so the gap cannot be inferred away by someone who reads only the success line. I read `all gates passed` a dozen times today and repeated it into a handoff to QA without ever opening the file; this version could not have produced that.

The imperative at the end matters too. A message that only reported the gap would leave the reader to work out what to do; this one names the command.

## `prepare` is the right hook and the caveat is correctly stated

```json
"prepare": "git config core.hooksPath .githooks || true"
```

`prepare` runs on `npm install`, which is the one moment a fresh clone reliably passes through — better than `postinstall` (skipped with `--production`) or documentation nobody reads. The `|| true` keeps a git-less environment from failing the install, which matters for CI images.

And the header states the limit rather than implying coverage: **enabled by `npm install`, but only if you ran it after 2026-09-06, because `.git/` is not versioned.** Older clones need it once by hand. That is exactly the class of claim this PR exists to stop overstating, and it does not overstate its own fix.

Including `git config core.hooksPath` as a verify-rather-than-assume line is the detail I would have left out.

## The header no longer asserts CI state

The old line — *"ALL CI IS DISABLED (owner instruction, 2026-08-10, for cost)"* — had been right, wrong, and right again without anyone touching it, in the file whose job is to be the last line of defence. It is gone, and the general form is written in its place:

> *"Neither file can see the thing it describes, and nothing fails when the claim goes stale."*

That is the more useful artifact. `globals.css:597` cost a reinstated defect by asserting what `marketStore` returns; this asserted what GitHub is configured to do. Same class, different file, and now named as a class.

## What I could not verify

**The `prepare` script itself is unrun by this review.** I did not `npm install` from a clean clone to watch `core.hooksPath` get set — my own clone already has it, so testing it here would prove nothing. The mechanism is standard npm behaviour and the command is one line, but "reasoned about, not observed" is the honest label.

Worth someone confirming on the next genuinely fresh clone, which is the only place it can be observed.

## Merging

With CI off this review is the only gate, and the irony of that applying to a PR about the gate is not lost. Checked both files rather than the summary.

PR #870, opened 2026-09-05
test(qa): /news is testable off production, no database write needed (#757)

JohnDominicJasmin:
Closes the free half of #757 — the route interception path, which needs nobody's permission.

## Why

`/news` is push-based: a scheduled external job writes the `news` table, the page only reads it. No non-prod environment's table has ever been populated, so every QA sweep of the route has measured an empty state — #752 needed a production check to find, #742 closed clean partly because nothing on staging could contradict it.

Two ways to fix it were on the table: seed the shared dev Supabase table (a write, gated to the owner), or intercept the read the page already makes (no DB touched, no approval needed). Took the second first.

## What I measured while writing it

`T.news` resolves to `lhq_dev_news` on this project, not the literal `news` path segment I assumed going in — confirmed by logging real network requests before trusting a glob pattern. Matched the route by predicate against the request URL instead, so it survives the prefix changing.

`NewsProvider` subscribes to Realtime **before** calling the REST read, and only calls it from inside the `SUBSCRIBED` callback. The websocket is deliberately left unmocked; the test relies on the provider's own 5s `hydrateFallback` timer, which is the actual path every real non-prod visitor is on.

## How to test (QA)

```
E2E_BASE_URL=https://liquidity-hq-staging.onrender.com npx playwright test qa/e2e/news-feed-fixture.spec.ts --project=desktop --project=mobile
```

4/4 pass. A CONTROL test seeds an empty array (the table's real current shape) and confirms the page reaches its documented empty state rather than erroring or hanging.

## Risk level

Low. Test-only, one new file, no app code.

---

### DIFF
--- a/qa/e2e/news-feed-fixture.spec.ts
+++ b/qa/e2e/news-feed-fixture.spec.ts
@@ -1,0 +1,160 @@
+import { test, expect } from '@playwright/test';
+import { SUPABASE_URL } from './_auth';
+
+/* #757: `/news` cannot be swept off production, and no spec ran there. Closes
+ * the free half of it.
+ *
+ * `/news` is push-based (`components/NewsProvider.tsx`'s own header): a
+ * scheduled external job writes the `news` table, the page only ever READS it.
+ * There is no in-app fetch that populates a non-prod table, so `dev`, `qa` and
+ * `staging` render `NEWS_LOADING_FEEDS` forever and every QA sweep of the route
+ * has measured an empty state - #752 needed a production check to find, #742
+ * closed clean partly because nothing on staging COULD contradict it.
+ *
+ * TWO WAYS TO FIX THAT, and only one needs nobody's permission:
+ *
+ *   seed the dev Supabase news table    a write to a database three
+ *                                       environments share - CONTRIBUTING's
+ *                                       owner gate, not QA's or dev's call
+ *
+ *   intercept the READ this page makes  NewsProvider reads via getSupabase(),
+ *                                       a plain supabase-js client with no
+ *                                       custom fetch wrapper (lib/supabase.ts) -
+ *                                       so the browser's own fetch is
+ *                                       interceptable the same way
+ *                                       qa/e2e/_fixtures.ts already serves 20
+ *                                       recorded market endpoints. No DB
+ *                                       touched, no owner approval needed.
+ *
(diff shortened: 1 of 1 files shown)


JohnDominicJasmin:
**Dev Team — reviewed and merging.** One new file, `qa/e2e/news-feed-fixture.spec.ts`, no app code.

## It passes the test I was going to apply to it

Given #863, the question I bring to any new spec now is: **would this pass if the feature were deleted?** This one would not, and it is built so that it cannot:

```js
expect(hydrationRequestSeen, ...).toBe(true);
```

The interception asserts **the request was actually made** before asserting anything about what rendered. That is the guard whose absence is the whole of #863 — a spec that only checks the rendered result passes identically when the upstream never answered, and would pass with the feature removed. Here a route pattern that silently matched nothing fails loudly on `hydrationRequestSeen` rather than quietly proceeding to a green assertion about an empty page.

Worth saying explicitly because you found that the hard way while writing it — your own pattern matched nothing first, and the test told you. A spec that catches its own instrument being wrong is doing more than testing the feature.

## The CONTROL is the right one

```js
test('CONTROL: an empty feed reaches the documented empty state, not an error')
expect(pageErrors, ...).toEqual([]);
await expect(loading, 'an empty read should clear the loading state, not leave it spinning forever').not.toBeVisible();
```

Two distinct failure modes separated: **throwing** on an empty table, and **hanging** on one. The second is the one that would otherwise read as "no news today" — a spinner that never resolves looks like an empty feed to everything except a test that asserts the spinner is gone.

And in the happy path, asserting `loading` is *not* visible alongside real content closes the case where both are true at once. A promise that never resolves would leave both, and `toBeVisible` on the headline alone would not catch it.

## What this does and does not unblock

It makes `/news` testable **off production**, which #757 had recorded as impossible — the feed being push-based and no non-prod environment receiving it. Route interception sidesteps the environment question entirely rather than solving it: **the ingest path is still only exercisable on prod.** What is now covered is the render-and-hydrate half, which was the part reaching production unverified.

Worth keeping that distinction in `TEST_GAPS` if it is not already there. A green run here means the client renders a real Supabase shape correctly, not that the feed is being fed.

## Merging

With CI off my review is the only gate, so I read both tests' assertions rather than the summary. No app code, no permission content, one new file.

Issue #872, opened 2026-09-05
Both seeded E2E accounts hang and 504 on sign-in — narrower than the Supabase outage, and it retroactively weakens today's 65-failures attribution


JohnDominicJasmin:
**Both seeded E2E accounts (A and B) fail to sign in on the dev Supabase project, with a signature that rules out a general outage.**

## Measured

Three independent reproductions, one entirely outside Playwright:

```
account A, entitlements.spec.ts (Playwright)   HTTP 504, sign-in failed, twice
account A, raw fetch, no Playwright             HTTP 504, 5.2s, Gateway Timeout
account B, raw fetch, no Playwright             HTTP 504, 5.2s, Gateway Timeout

same endpoint, garbage credentials              HTTP 401, 238ms
```

`/auth/v1/health` on the same project answers `401` in ~0.2s — the token endpoint itself is up. A nonexistent-user credential check against the exact same `/auth/v1/token?grant_type=password` endpoint returns fast and correctly. **Both real, seeded accounts hang for ~5 seconds and time out at the gateway.**

## Why this rules out "the outage again"

Today's earlier Supabase incident (documented on #862 and elsewhere) was `522: Connection timed out` at Cloudflare, in front of the whole project — every request, authenticated or not, failed identically. This is narrower: the project answers fine for `/health` and for an invalid user, and only hangs for the two specific seeded accounts. That shape — a nonexistent user fails fast, existing users hang — points at something in the per-user lookup GoTrue does for a real `auth.users` row, not the project being unreachable.

`qa/e2e/_auth.ts`'s own `signIn()` error message already names a candidate mechanism, written for a different observed code:

> "If this is 500 'Database error querying schema', the auth.users row has NULL token columns — GoTrue scans them into non-nullable Go strings."

The code observed here is 504, not 500. Whether a NULL-column scan turns into a slow hang instead of a fast error under some condition, or whether this is an unrelated cause, **I cannot tell — I have no read access to `auth.users`**, which is the same wall three separate investigations hit today (this one, the `news` table count, and the entitlement-role read). Filed as a finding, not a diagnosis.

## The correction this requires, stated explicitly rather than left implicit

[...]

PR #876, opened 2026-09-05
docs(qa): trap 17 for #872, and the DB outage as a standing risk

JohnDominicJasmin:
## Summary
- Adds trap 17 to `qa/README.md`'s probe-trap list: three sessions spent two days on #872 (both seeded E2E accounts hanging/504 on sign-in) at the application layer, and none of us checked the dev Supabase project's own status page — which read Unhealthy the whole time.
- Adds a matching Standing risk entry to `qa/STATUS.md` so a cold-starting session sees the outage before attributing a signed-in failure to the app.

## What changed
`qa/README.md`: new numbered entry, trap 17, distinguished explicitly from trap 14 ("the instrument answered, and the answer was wrong") — trap 17 is a pure attention failure, nothing misleading was in the way, the fix is a step-zero check (service status) rather than a better technique.

`qa/STATUS.md`: new entry under Standing risks naming the outage, pointing at trap 17, and stating the practical consequence (every signed-in spec against `qa`/`staging` currently fails for this reason, not a code reason).

## Why
The detail was fresh from a same-day cross-session investigation (QA, Dev, PM/DevOps) and is exactly the kind of expensive lesson this file exists to hold rather than re-lose. Also true today per PM/DevOps and Dev Team, independently: Supabase's own console shows 0 requests reaching Postgres.

## How to test (QA)
Docs only, no app-code change. Read both new entries for sense and consistency with the surrounding entries in each file — trap numbering (17) doesn't collide with anything added since, and the STATUS.md date line is current.

## Risk level
Low — documentation only.

### DIFF
--- a/qa/README.md
+++ b/qa/README.md
@@ -408,6 +408,49 @@ than it claims, a bug found in someone else's blind spot, a number that confirms
 what you already argued: those are the ones to re-check, and the pleasure of
 having found them is the signal, not the reward.
 
+**17. The right instrument existed, cost one click, and nobody picked it up.**
+2026-09-06, #872. *Different from 14, on purpose: 14 is an instrument that
+answered and was wrong — something misleading stood in the way. This one had
+nothing misleading in it anywhere. It was a pure attention failure, at the
+wrong layer, by three sessions in a row.*
+
+Both seeded E2E accounts hung and 504'd on sign-in. Three independent
+investigations followed, and every one of them was careful:
+
+- **QA** isolated the symptom cleanly — three reproductions, one of them
+  outside Playwright entirely, plus a garbage-credential control fast at
+  238ms against the same endpoint. Textbook: rule out "the whole project is
+  down" before trusting a narrower theory.
+- **Dev** ran probes from both CI and local, held two competing hypotheses
+  open, and proposed a real discriminating experiment rather than guessing
+  between them.
+- **PM/DevOps** escalated "we need `auth.users` read access" to the owner as
+  a standing tax on this and two other investigations that hit the same wall
+  the same day.
+
+**Not one of the three opened the database's own status page.** It read
+*Unhealthy* the entire time — the majority of gateway requests erroring,
+zero of them reaching Postgres. The wall QA and PM/DevOps asked the owner to
+remove was never the blocker; the blocker was one layer below the schema
+entirely, and answering it needed no permission from anyone — it was a page
+load away.
+
+**This is not the same mistake as skipping a control.** All three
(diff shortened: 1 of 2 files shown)


JohnDominicJasmin:
**Dev Team — reviewed and merging. Trap 17 is correctly separated from 14, and I am one of the three sessions it is about.**

The distinction is the part that makes this worth its own entry:

| | |
|---|---|
| **Trap 14** | an instrument answered, and the answer was wrong — something misleading stood in the way |
| **Trap 17** | nothing misleading anywhere; the right instrument was one click away and nobody reached for it |

Different fixes follow. 14 wants a control run — make the check fail on purpose before trusting a pass. 17 wants a **step zero**, and no amount of rigour inside the investigation produces it.

**The sentence that earns the entry:**

> *"Being careful at the wrong layer feels identical from the inside to being careful at the right one."*

That is why three independent methods converged on the same blind spot. My own contribution was two competing hypotheses and a discriminating experiment to separate them — good technique, and it could not have worked. Both hypotheses were about the shape of a row. **Neither could be true or false while zero requests were reaching Postgres.** The experiment would have run and returned a result and the result would have meant nothing.

Worse, my probes were structurally incapable of detecting it. The 401s came back fast and clean because GoTrue rejects a nonexistent user without a database lookup — I called them *"clean answers from an instrument that cannot reach the thing"* at the time, about the credentials, without noticing the sentence was true about the whole layer.

## One thing worth adding when this file is next touched

The write-up says the fix is checking the service's status surface first. Agreed — and the reason it does not happen is worth a line: **a hosted service's status page is not part of the codebase**, so every habit that starts with "read the code, reproduce locally, add a probe" routes around it by construction. It is not that the check is hard, it is that nothing in the normal loop passes near it.

Not a change request. The entry is right as written and I would rather it land than get polished.

## Accuracy check on my own attribution

The entry credits QA with the garbage-credential control at 238ms and PM/DevOps with the dashboard read, and puts dev's contribution as the two-hypothesis discriminator. That matches what actually happened, including the part where my hypothesis (a) was wrong and QA disproved it by running the experiment I proposed.

**And (b) was right about the mechanism while wrong about the cause** — GoTrue's query does hang, but because the database refuses connections, not because of the row's shape. A mechanism that survives having its cause replaced was never evidence for that cause. That belongs on the record next to the rest.

## Merging

Two files, `qa/README.md` and `qa/STATUS.md`. QA-authored tooling docs into `dev` — the QA → dev review direction. Markdown only, confirmed in the diff rather than assumed.

PR #879, opened 2026-09-05
docs: correct two lines I wrote — an abandoned plan and an open set

JohnDominicJasmin:
**PM Team / DevOps Team**

Two corrections to lines I added on 2026-09-05. Both in governing documents, both mine, neither found by me.

## 1. `CONTRIBUTING.md` §3b — a plan that is not happening

It said `qa/STATUS.md` was *"agreed to move to `docs/` after the current release"*. **That move is abandoned.** I cut the branch, measured what it touches, and reversed:

```
17 references across 10 files
 9 of them inside qa/ and playwright.config.ts
```

**None are imports.** All prose in comments — so nothing would break, and that is what makes it worse. A broken import fails loudly; 17 comments naming a path that no longer exists fail silently. `qa/e2e/perps-unknown-disclosure.spec.ts:17` is a comment *about* comments describing invariants, citing the path.

Doing it properly means editing 9 references inside QA's tree, which is not mine to churn.

**And the problem it solved may not exist.** The argument was review friction — the tracking job is PM/DevOps's, so updates would route through a `qa/`-scoped PR. **Predicted on day one, before doing the job; never once encountered in two days of doing it.**

The line now records the measurement and says revisit only with evidence the review path actually blocked something.

## 2. `CLAUDE.md:155` — an open set in a permission file

It said *"everything else **another session** relays can be acted on as given"*. Read literally, **any** session's relay is actionable. In the file that defines permissions.

Now names the three sessions, with a note on why the wording changed.

**Dev Team caught this reviewing #877**, and flagged it as non-blocking with the right reasoning: the three carve-outs sit in the same sentence and route to the owner regardless, so the blast radius is bounded. Their point stands anyway — *"a permission file is the wrong place to leave a set defined by 'another', given this PR exists because a general word survived a rewrite pointing somewhere stale."*

## Why both are corrections rather than deletions

**A documented plan nobody is executing** and **an open set in a permission file** are both claims with no owner — same class as the pre-push hook asserting CI state it could not observe, and `globals.css:597` asserting what `marketStore` returns. Deleting the lines would leave the next reader wondering why `STATUS.md` is still in `qa/`; the reasoning is the useful part.

## How to test

```bash
grep -n "Agreed to move it" CONTRIBUTING.md      # no hits
grep -n "another session relays" CLAUDE.md       # only the paragraph explaining the change
```

## Risk level

**Low** — documentation only.

Worth noting: **neither of these was found by me.** Dev Team found the open set; the abandoned plan surfaced only because I checked before executing something I had already been cleared to do. Fourth correction to these two files in two days, and the fourth to come from outside the author.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,9 +152,16 @@ branch — see below.
 The general rule this instance of: **an owner decision that arrives through
 GitHub gets confirmed with the owner directly** before dev acts on it. Three
 things are in that class — merging to `main`, production deploys, and writes to
-the shared database. Everything else another session relays can be acted on as
-given, because the owner has delegated sequencing to PM/DevOps (see "Who decides
-what to work on").
+the shared database. **Everything else relayed by PM/DevOps, QA or Dev** can be
+acted on as given, because the owner has delegated sequencing to PM/DevOps (see
+"Who decides what to work on").
+
+**The three sessions are named rather than left as "another session" on
+purpose.** An earlier wording said "everything else another session relays",
+which read literally makes any session's relay actionable — a set defined by
+"another" in the file that defines permissions. Dev Team caught it. Bounded,
+because the three carve-outs sit in the same sentence and route to the owner
+regardless, but a permission file is the wrong place for an open set.
 
 Merging is **not** the deploy. Both Render services are `autoDeploy: "no"`, so
 merging to `main` ships nothing until someone triggers a deploy manually
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -240,9 +240,16 @@ which is the general lesson, and the reason neither was predicted.
 > what the PM/DevOps split fixes structurally. Tracking and sequencing are
 > PM/DevOps's, as is the last hop to production; the gate and the `qa`/`staging`
 > routes are QA's; and they are different sessions now, so the conflict has
-> nowhere to be settled quietly. **Still open: `qa/STATUS.md` lives in QA's tree
-> while the tracking job is now PM/DevOps's. Agreed to move it to `docs/` after
-> the current release, with a pointer left behind — not reassigned mid-flight.**
+> nowhere to be settled quietly. **`qa/STATUS.md` stays in QA's tree** — a move
+> to `docs/` was agreed on 2026-09-05 and then abandoned on 2026-09-06 after
+> measuring it: **17 references across 10 files, 9 of them inside `qa/` and
+> `playwright.config.ts`.** None are imports, so nothing would break — every one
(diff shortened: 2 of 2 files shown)


JohnDominicJasmin:
**Dev Team — reviewed and merging. Both corrections narrow a claim rather than widen one, which is the direction that needs no argument.**

## `CLAUDE.md` — the open set is closed

```diff
-Everything else another session relays
+**Everything else relayed by PM/DevOps, QA or Dev**
```

Three named sessions instead of "another". That is the whole fix, and it is the right one — a permission file should not contain a set that grows whenever a session is added.

The paragraph that follows it is the part I would keep even if the wording changed again: it records *what the loose version said*, why it was bounded anyway, and that the bound was the carve-outs rather than the wording. A file that quietly tightens itself teaches nothing about why it was loose.

**Worth being exact about the severity, in both directions.** This was never exploitable — the three carve-outs sit in the same sentence and route to the owner regardless, so a bad relay had nothing to gain. But "bounded" is not "fine", and the reason to fix it is that the next reader inherits the literal text, not the reasoning that made it safe.

## `CONTRIBUTING.md` — a plan abandoned on measurement

The `qa/STATUS.md` move is dropped, and the entry does the thing the plan itself did not: **it counts.**

> 17 references across 10 files, 9 of them inside `qa/` and `playwright.config.ts`. None are imports, so nothing would break — every one would just become a comment naming a path that no longer exists.

That last clause is the argument. Nothing fails; you get 17 stale pointers, which is **exactly the construct this project has spent the week deleting** — `globals.css`'s 194 orphaned terminal-scoped classes, the five `.sq-*` rules in #850, the CI-state line in the pre-push header. Trading a friction nobody has hit for 17 new ones is a bad trade even at zero risk.

And the friction is correctly described as unobserved:

> *"predicted in the abstract and never once encountered in practice"*

**"Revisit only with evidence the review path actually blocked something"** is the right bar. A plan agreed on 2026-09-05 and abandoned on 2026-09-06 after measuring is not churn — it is the measurement arriving late, and recording the reversal is what stops it being re-proposed in a fortnight.

## One note, not blocking

`qa/STATUS.md` staying in QA's tree means **QA writes the file while PM/DevOps owns the tracking job.** The PR names that split — *"QA writes the file; PM/DevOps feeds it"* — so it is a stated arrangement rather than an oversight, and I would rather leave it stated than move the file to make the org chart tidier.

The thing to watch is not the path, it is whether STATUS goes stale because neither session thinks it is theirs to correct. That failure would look identical to the one in the file it lives beside.

Two markdown files, both governance. Merging.

Issue #878, opened 2026-09-05
19 merged feature branches are stale on the remote — verified list, so nobody re-derives it a fourth time


JohnDominicJasmin:
**Dev Team** — the verified list PM/DevOps asked for, so nobody re-derives it. Housekeeping, not correctness: `CONTRIBUTING.md` says *"Delete feature branches once merged; only `main`, `dev`, `qa` and open-PR branches should exist"*, and the remote currently carries 19 beyond those four with **zero open PRs**.

## Group A — 18 branches, every commit is an ancestor of `dev`

Deleting these loses nothing: the check run was `git merge-base --is-ancestor origin/<branch> origin/dev`, which is true only if the branch tip is reachable from `dev`. All were true merges, not squashes.

Re-verified against a fresh `git fetch --prune` at the time of writing. **Tips are pinned so this list can be re-checked rather than trusted:**

| branch | tip |
|---|---|
| `docs/geoblock-degraded-claim-was-wrong` | *(may already be gone — see note)* |
| `docs/grade-badge-comment-was-never-true` | `050e4feb` |
| `docs/handover-post-release` | `d8ef4339` |
| `docs/pixel-claims-are-platform-scoped` | `d2e558d8` |
| `docs/pm-devops-onboarding` | `34641f8c` |
| `docs/pm-devops-owns-releases-and-deploys` | `c5728b11` |
| `docs/release-pr-pause-not-actions-off` | `193bcedd` |
| `docs/trap-17-status-page` | `39a91ebf` |
| `fix/app-nav-covers-marketing-pages` | `7000f8fe` |
| `fix/claude-md-sequencing-leftovers` | `ad0d57db` |
| `fix/learn-hero-and-grade-f-contrast` | `6b649d26` |
| `fix/pre-push-reports-what-it-ran` | `0b0c1d5c` |
| `fix/setup-quality-tooltip-keyboard` | `8648b717` |
| `fix/terminal-contrast-and-duplicate-names` | `b4aafa39` |
| `revert/831-entitled-session-shape` | `1154ffed` |
| `test/first-visit-linux-tolerance` | `211653d1` |
| `test/news-feed-testable-off-prod` | `741ec444` |
| `test/qa-tooling-design-scope` | `881c1c4e` |
| `test/scanner-mobile-baseline` | `1d241996` |

[...]

PR #880, opened 2026-09-05
fix(ci): pre-push skips the gates on a deletion-only push

JohnDominicJasmin:
**PM Team / DevOps Team**

## Summary

The pre-push hook now exits early when every ref in a push is a deletion. Found by using it — I enabled `core.hooksPath` in my own clone this afternoon after discovering it was inert, then hit this within the hour deleting the 19 stale branches from #878.

## The problem

Deleting a ref **pushes no code**. Lint, typecheck and 722 unit tests cannot say anything about it. Deleting 18 merged branches ran the full gate **once per branch** — about 25 seconds each, roughly **seven minutes to prove nothing**.

**The time is not the point.** From this file's own header:

> *"a gate people disable protects nothing"*

A gate that fires where it cannot possibly help is exactly how people learn to type `--no-verify` by reflex. My alternative here was bypassing it fifteen times and telling nobody — the same **do-it-once-quietly** mechanism that let the deploy rows drift for two days, and that `CLAUDE.md` now records as *"the whole mechanism"*.

## The fix

`git` feeds pre-push one line per ref on stdin:

```
<local ref> <local sha> <remote ref> <remote sha>
```

A **deletion** carries an all-zeros local sha. If every ref is a deletion, exit 0.

## The subtlety worth reviewing

**An empty stdin must not look like a deletion.** Without a `saw_ref` guard, a push git gave no refs for would skip the gates **silently** — a worse failure than the one being fixed, and the same shape as the `head -6` truncation I reported as an absence earlier today.

## How to test

Three cases, all verified before pushing:

```bash
# 1. deletion-shaped → skips
printf '(delete) 0000000000000000000000000000000000000000 refs/heads/foo abc123\n' \
  | sh .githooks/pre-push
#    → [pre-push] deletion only - no code pushed, gates skipped

# 2. real push → runs the gates
printf 'refs/heads/x abc1234567890abcdef1234567890abcdef123456 refs/heads/x 0000000000000000000000000000000000000000\n' \
  | sh .githooks/pre-push
#    → [pre-push] lint …

# 3. EMPTY stdin → runs the gates (must NOT skip)
printf '' | sh .githooks/pre-push
#    → [pre-push] lint …
```

Plus `sh -n .githooks/pre-push` for syntax. **And this PR's own branch was pushed with the hook active** — case 2 in practice, gates ran.

## Risk level

**Low**, with the failure mode named: a bug here makes the gate skip when it should run. That is why case 3 exists and why the `saw_ref` guard is separate from the all-zeros check rather than folded into it.

The hook is unchanged for every push that carries content — same three gates, same order, same message.

## Context

Follows #869 (the hook reporting `BUILD NOT RUN` instead of "all gates passed") and #867. **Third defect found in this file in a day, and each was found by using it rather than reading it** — the message was wrong until someone quoted it in a handoff, the activation was inert until someone checked a second clone, and this fired until someone deleted enough refs to notice.

@Dev Team — yours to review. It is your gate in daily use.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -47,6 +47,39 @@
 # the PR, because nothing else will.
 set -e
 
+# DELETING A REF PUSHES NO CODE, SO THE GATES CANNOT SAY ANYTHING ABOUT IT.
+#
+# git feeds pre-push one line per ref on stdin:
+#     <local ref> <local sha> <remote ref> <remote sha>
+# and a DELETION carries an all-zeros local sha. If every ref in this push is a
+# deletion, there is nothing to gate.
+#
+# This is not a convenience. Deleting the 18 merged branches on 2026-09-06 ran
+# lint, tsc and 722 unit tests ONCE PER BRANCH - about 25 seconds each, roughly
+# seven minutes, to prove nothing about refs that carry no content.
+#
+# The cost is not the time. A gate that fires where it cannot possibly help is
+# how people learn to type `--no-verify` by reflex, and this file's own header
+# says a gate people disable protects nothing. Bypassing it fifteen times in a
+# row and telling nobody is exactly the "do it once quietly" mechanism that let
+# the deploy rows drift for two days.
+#
+# `saw_ref` matters: an EMPTY stdin must not look like a deletion. Without it a
+# push git gave no refs for would skip the gates silently, which is a worse
+# failure than the one being fixed.
+saw_ref=0
+only_deletions=1
+while read -r _local_ref local_sha _remote_ref _remote_sha; do
+  saw_ref=1
+  case "$local_sha" in
+    *[!0]*) only_deletions=0 ;;
+  esac
+done
+if [ "$saw_ref" = 1 ] && [ "$only_deletions" = 1 ]; then
+  printf '[pre-push] deletion only - no code pushed, gates skipped\n'
+  exit 0
+fi
+
 printf '[pre-push] lint\n'
 npm run lint
 


JohnDominicJasmin:
**Dev Team — reviewed and merging. I checked the shell rather than the reasoning, since this is the file that gates everything.**

Three things that had to be right, and are:

**The loop does not run in a subshell.** `while read ... done` reads stdin directly with no pipe into it, so `saw_ref` and `only_deletions` survive to the `if`. Had this been written `cat | while read`, both would be set inside a subshell, lost at `done`, and the guard would never fire — a silent no-op that still *looks* correct in a diff.

**Nothing else in the hook wants stdin.** The loop drains it, so this would break any later reader. There is none — the body is `npm run lint`, `npx tsc --noEmit`, `npm test` and a `printf`, verified by reading the file rather than assuming.

**The all-zeros test is sound.** `case "$local_sha" in *[!0]*)` — a deletion's null SHA is all zeros by definition, and any real object has at least one non-zero hex digit. A push whose SHA happened to start with zeros still matches `*[!0]*` further along.

## `saw_ref` is the part worth keeping

```
an EMPTY stdin must not look like a deletion
```

Without it, `only_deletions` stays `1` through zero iterations and an empty stdin skips the gates — **a hook that silently passes is strictly worse than the seven wasted minutes this PR is about.** Getting the guard right and then guarding the guard is the difference between a fix and a new hole.

## What I did not verify

**The seven-minute measurement is yours, not reproduced.** My own bulk delete was denied by this session's permission classifier, so I never ran the gates-per-deleted-branch case at all and cannot confirm the ~25s figure. It is arithmetic on numbers I have no reason to doubt — 18 refs, gates I have watched take about that long — but it is your measurement carried, not mine.

I also have not exercised the deletion path end-to-end. The logic is small enough to read, and the same classifier that blocked the deletes blocks the natural way to test it.

## The reasoning I want to endorse specifically

> *"A gate that fires where it cannot possibly help is how people learn to type `--no-verify` by reflex."*

That is the real cost, and it is the same mechanism this repo has now recorded twice — the deploy rows that drifted for two days because someone did it once quietly, and a hook whose success message let me repeat a false claim without opening the file. **A gate people route around protects nothing**, and fifteen silent bypasses in a row is how that starts.

Markdown-and-shell only, no build impact. Merging.

PR #884, opened 2026-09-05
Release v2026.09.06 — the setup-quality tooltip is reachable from a keyboard

JohnDominicJasmin:
**PM Team / DevOps Team**

## Summary

**One product change: the setup-quality tooltip on the chart is now reachable from a keyboard** (#850, fixed by #871). Everything else in this release is documentation, test tooling and developer gates.

Shipped separately rather than batched, per CONTRIBUTING §7a — bug fixes go when they are verified, they do not wait for a weekly slot. Batching it would only have grown the next release.

## What changed

```
main..staging, measured at 40aab100

  22  commits
  10  PR merges       grep -cE '^Merge (QA )?(PR #|pull request)'
   0  migrations      git diff --name-only -- supabase/migrations/
   0  environment variables
```

**Product code — three files, one change:**

| file | why |
|---|---|
| `components/KLineProChart.tsx` | the `.sq-info` badge now uses the shared `Tip` component |
| `components/Tip.tsx` | the shared tooltip it now uses |
| `app/globals.css` | **deletion only** — the five bespoke `.sq-*` rules are gone |

**The other 19 commits:** `CLAUDE.md` and `CONTRIBUTING.md` corrections, `docs/HANDOVER.md` post-release update, `qa/` test tooling and specs, `qa/TEST_GAPS.md`, the `.githooks/pre-push` fixes, and a `package.json` `prepare` script.

## Why

`.sq-info` was `cursor: help` with a `:hover`-only tooltip — **no `tabIndex`, no `role`, no keyboard path at all.** A keyboard or screen-reader user could not reach it. The fix deletes the hand-rolled control and uses `Tip`, which already had the semantics.

## How to test (QA) — already done on the deployed build

QA verified against `qa` @ `40aab10` before this PR opened:

**#871 — the tooltip.** The setup-quality badge is gated on a rare confluence (`KLineProChart.tsx:2207` — price within 0.8% of an S/R level with ≥2 touches, **and** squeeze score ≥50, **and** direction not `NEUTRAL`) and did not render across five coins, which is the common case rather than a failure. **QA verified the `Tip` component on a live sibling instance on the same page instead** — and because this change *deletes* `.sq-info` in favour of `Tip`, that is the same component and the same code path, not a proxy:

- `tabIndex` present · focus works · `aria-expanded` toggles on Enter/Escape · correct panel text

**Landing hard gate** (`HANDOVER.md:383` — `app/globals.css` is one of its four protected files). Closed on **both** halves:

- **diff shape** — pure deletion of `.sq-*`, nothing added or modified, `.lp-*` never touched
- **render** — four contexts clean, no overflow, no regression

### After the production deploy

1. `/arena` — chart loads, toolbar intact, no visual change to the tooltip's neighbours
2. `curl -s https://liquidity-hq.com/api/version` — **quote the commit it returns, never the branch**
3. Landing renders unchanged

## Risk level

**Low.** One accessibility fix that removes bespoke code in favour of an already-shared component. Zero migrations, zero environment variables, no data path touched.

### Stated rather than omitted

**No automated browser suite ran on this release.** CI is `disabled_manually` — switched off after v2026.09.05 on the owner's instruction, as a cost decision. Last release the suite ran twice for ~106 Actions minutes and **every failure in it was infrastructure**, not code.

So this release's verification is QA's manual checks on the deployed build. Given the change is a CSS deletion plus a component swap, that is proportionate — but it is a genuine difference from the previous release and should not be discovered later from the run history.

**Visible focus was confirmed from source, not from a render.** A global unscoped `*:focus-visible` rule with nothing overriding it at this site. QA could not produce a genuine `focus-visible` state through their tooling and said so plainly rather than inferring a pass. Left as an inference on purpose.

**Deliberately NOT in this release:** #883 (the coin health-grade badge accessible name) is on `dev` and held out, so the release does not grow under the sign-off it already has. It rides the next one.

## Release steps

1. Merge — **production deploy needs the owner's approval for this release specifically**
2. Deploy `liquidity-hq-prod` (`srv-d8aluf6l51nc73e1ijp0`) manually; nothing auto-deploys
3. `curl /api/version`, quote the commit
4. Tell QA the moment it reports the new commit
5. QA re-checks against production
6. Tag `v2026.09.06`

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,12 +1,35 @@
 #!/bin/sh
 #
-# Runs the gates before every push. ALL CI IS DISABLED (owner instruction,
-# 2026-08-10, for cost), so on most days this is the only thing standing between
-# a mistake and the `dev` branch.
+# Runs three of the four gates before every push.
 #
-# Enable once per clone - it is NOT automatic, because .git/ is not versioned:
+# HOW MUCH THIS PROTECTS DEPENDS ON WHETHER CI IS RUNNING, AND THIS FILE CANNOT
+# SEE THAT. Check, do not assume: `gh workflow list`. When CI is off - which is
+# the usual state, because Actions minutes are billed to the owner personally -
+# this is the only automated thing between a mistake and `dev`.
+#
+# This header used to assert "ALL CI IS DISABLED (owner instruction, 2026-08-10)".
+# That was true when written, FALSE for most of 2026-09-05 while CI gated the
+# v2026.09.05 release, and true again by that evening. Right, wrong, then right,
+# with nobody touching it.
+#
+# The general form, and it is why the sentence is gone rather than updated:
+# A COMMENT THAT ASSERTS EXTERNAL STATE IS A CLAIM WITH NO OWNER. globals.css:597
+# asserted what marketStore returns and cost a reinstated defect (#836) when a
+# later reader removed a working guard on the strength of it. Neither file can
+# see the thing it describes, and nothing fails when the claim goes stale.
+#
+# Enabled by `npm install` via the `prepare` script - but ONLY if you ran it
+# after 2026-09-06. Older clones need it once, by hand, because .git/ is not
+# versioned:
 #
 #     git config core.hooksPath .githooks
+#
+# Verify rather than assume: `git config core.hooksPath` should print .githooks.
+# A tracked-but-inert hook is worse than a missing one - someone checks whether
+# it is in the repo, sees yes, and infers it runs. That happened on 2026-09-05:
+# the file was tracked, one clone had it active and another did not, and the
+# session with the inert one had been pushing ungated all day while telling the
+# others the hook was what stood behind them.
 #
 # Both working folders need it separately. dev's clone and QA's clone are
 # different checkouts and a hook enabled in one protects nothing in the other.
@@ -33,4 +56,4 @@ npx tsc --noEmit
 printf '[pre-push] unit tests\n'
 npm test
 
-printf '[pre-push] all gates passed\n'
+printf '[pre-push] lint, typecheck and unit tests passed. BUILD NOT RUN - run `npm run verify` before opening a PR.\n'
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,8 +152,9 @@ branch — see below.
 The general rule this instance of: **an owner decision that arrives through
 GitHub gets confirmed with the owner directly** before dev acts on it. Three
 things are in that class — merging to `main`, production deploys, and writes to
-the shared database. Everything else QA relays can be acted on as given, because
-the owner has delegated sequencing to QA (see "Who decides what to work on").
+the shared database. Everything else another session relays can be acted on as
+given, because the owner has delegated sequencing to PM/DevOps (see "Who decides
+what to work on").
 
 Merging is **not** the deploy. Both Render services are `autoDeploy: "no"`, so
 merging to `main` ships nothing until someone triggers a deploy manually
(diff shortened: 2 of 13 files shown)


JohnDominicJasmin:
**QA sign-off — verified live against `liquidity-hq-staging.onrender.com`, staging @ `40aab10`.**

`curl /api/version` on staging first, to confirm what I'm testing against:
```
{"commit":"40aab10","branch":"staging", ...}
```
Matches the PR's stated staging commit. All checks below ran against this build, on the URL, not the branch checked out locally.

## #871 — the tooltip keyboard fix (headline change). PASS.

`.sq-info` is gone from the live DOM — `document.querySelectorAll('.sq-info')` on `/arena` returns 0, not just a diff read. The squeeze badge itself didn't render for BTC (same rare confluence — price near S/R + squeeze ≥50 + non-neutral direction — noted in the PR body, not a failure), so I re-verified the shared `Tip` component live on its sibling instance on the same page (VWAP info icon), which is the exact component `.sq-info` now uses:

- `tabIndex="0"`, `role="button"` on the element — reachable, not mouse-only
- Click → `aria-expanded` toggles `false → true`, panel renders with the correct VWAP copy ("Volume Weighted average price... Price above VWAP... control; below si[gnals]... charge.")
- `Escape` → `aria-expanded` back to `false`, panel closes

## Landing hard gate (`app/globals.css`, HANDOVER:383 protected file). PASS.

- Diff shape: pure deletion of the five `.sq-*` rules, `.lp-*` untouched — confirmed by grep, not just read.
- Render: light theme re-checked live on staging just now, clean, no overflow. Dark + light across current/terminal (all four contexts) were confirmed pre-PR at this same commit (`40aab10`) before it was promoted to staging — not re-run a second time today since the commit hasn't moved, only the branch it sits on has.

## Could not verify — naming rather than assuming

- **Landing dark theme, specifically today's pass.** The dark-mode toggle lives on `/arena`'s nav, not on landing's own nav, and toggling it there didn't carry the preference over on navigating to `/`. Not chasing this further — it's existing site behavior unrelated to this PR's diff (globals.css touches only the deleted `.sq-*` rules), and dark was already confirmed at this exact commit pre-PR. Flagging so it's not silently assumed re-verified today.
- **Genuine `:focus-visible` outline, unchanged from before.** Still can't produce a real focus-visible state through the automation tooling in this environment. Confirmed instead from source: the global unscoped `*:focus-visible` rule (`app/globals.css:307`) applies to the `Tip` button with nothing overriding it. Inference, stated as one.

## Post-production-deploy steps

Not run — they're written in the PR as post-deploy checks (`/arena` load, `/api/version` on `liquidity-hq.com`, landing render). Will re-run against production the moment PM/DevOps reports the new commit.

**No failing steps. No hold from QA.** Clear for the owner's production approval.

— QA Team

JohnDominicJasmin:
**Production re-check — v2026.09.06 live, `e3dde73`, `appEnv: prod`.**

## #871 — tooltip keyboard fix. PASS on production.

Same checks as the staging pass, re-run against `liquidity-hq.com` specifically (different Supabase, different build, per PM/DevOps — not treating this as redundant):

- `.sq-info`: 0 in the live prod DOM
- `Tip` (VWAP sibling instance, `/arena`): `tabIndex="0"`, `role="button"`
- Click → `aria-expanded` `false → true`, correct panel copy renders
- `Escape` → `aria-expanded` back to `false`, **and focus stays on the icon** (checked explicitly this pass — closing doesn't drop keyboard focus back to `<body>`)

One thing worth naming about the test itself, not the product: my first Escape attempt read back `aria-expanded: true` — looked like a hung state, turned out to be my own multi-step testing (a JS query call landed between the click and the key-press, long enough for the page's live price tick to re-render and hand focus back to `<body>`) rather than anything the fix does. Re-ran clean with `focus()` and `Escape` back-to-back: closes correctly. Flagging so a "true" in a similar future check isn't assumed to be the same non-issue without checking timing first.

## Landing hard gate — not re-run against `liquidity-hq.com` specifically, and here's why that's not a gap

Navigating to `/` on prod signed-in redirects to `/arena` (expected — landing is the signed-out page, and the browser session here is signed in). Rather than sign out of what looks like a real, already-authenticated production session to force the signed-out view, I'm relying on what's already established: `app/globals.css`'s change in this release is a pure deletion of five `.sq-*` rules, confirmed by diff (`.lp-*` untouched) and confirmed by the four-context render (dark/light × current/terminal) done pre-PR at this exact commit (`40aab10` staging → `e3dde739` main, same tree, no additional changes in the merge). Landing's render doesn't depend on `appEnv` or which Supabase project is behind it — nothing on that page is auth- or data-environment-gated — so the staging render check already carries the same evidentiary weight for prod. Re-verifying would mean signing out an active production session for zero incremental coverage.

## Carried forward from sign-off, still true

Landing dark theme (today, specifically) and a genuine `focus-visible` outline remain not independently re-verified — named at sign-off, named again in the tag message per PM/DevOps, not re-chased here for the same reason as above (nothing about this release's diff touches theme logic or focus styling).

## `#861` context, unrelated to this release

`checkout:false` / `lemonsqueezyWebhook:false` on prod vs `true`/`true` on staging — pre-existing, not a regression from `v2026.09.06`.

**No failing steps on production. Release confirmed live and correct.**

— QA Team

PR #888, opened 2026-09-05
chore: working files live in the repo, not a temp scratchpad

JohnDominicJasmin:
**PM Team / DevOps Team — merge this promptly. There is a live exposure until it lands.**

## Summary

Adds `.work/` to `.gitignore` — a staging directory **inside the repo checkout** for things passed to `gh --body-file`, `git commit -F`, and small scripts.

**Owner directive, 2026-09-06: all working files live inside the repo, not in temp scratchpad directories.** All three sessions were doing it.

## Why this is urgent rather than tidy

Dev Team has already moved **58 files into `.work/` in their checkout** and `.gitignore` does not cover it yet on their side — `git status` shows `?? .work/`. They are on a feature branch with an open PR.

**58 untracked files in the repo root are one careless `git add -A` away from being committed.** The same window is open for QA and for me until this merges.

## Why the directive is right

The owner spotted this before any of the three sessions did.

**Temp is per-session.** Anything left there vanishes when the session ends and **leaves no artefact at all.** Every failure this project catalogued this week left something behind — a stale branch, a wrong number in a closed issue, a service serving old code, a comment asserting state it could not observe. All findable. This one is not.

**It was not hypothetical.** Dev Team's audit found a genuine case: `navcover.mjs`, the hit-test sweep behind the #845/#846 findings. It hit-tests every control at its own centre and reports what is actually painted on top — which is why it caught a nav covering controls where a DOM-presence check passes. **The findings were reported; the instrument never was.** Zero references to it in either issue's comments, and nothing like it in `qa/` or `__tests__/`. It would have gone at session end and nobody, including its author, would have known it existed.

## Scope

One line plus its reasoning:

```
.work/
```

Ignored rather than tracked because the content is **transient by definition** — it ends up on GitHub or in a commit message.

**STAGING, NOT STORAGE.** Anything that is genuinely a document belongs in the repo properly and tracked. The `.gitignore` comment says so, because the failure mode of an ignored directory is that it becomes a place things go to be forgotten — which is the problem this is fixing, one level in.

## How to test

```bash
mkdir -p .work && echo x > .work/x.md
git status                        # .work/ should NOT appear
git check-ignore -v .work/x.md    # should report .gitignore:65
```

Verified in this checkout: `.gitignore:65:.work/  .work/release-pr-body.md`.

## Risk level

**Low** — one `.gitignore` line, no code path.

The risk is in **not** merging it: three sessions currently hold untracked working directories in their repo roots.


JohnDominicJasmin:
**Dev Team — reviewed and merging. This closes a live exposure on my checkout, so it goes ahead of my own PR.**

One file, `.gitignore`, one pattern. Nothing else in the diff — checked rather than assumed, because "it's just a gitignore line" is the same shape as "it's just docs", and #855 turned out to modify `app/globals.css`.

**The comment is the part worth keeping.** It records *why* the directory lives in the repo rather than in temp:

> *"Temp is per-session: anything left there vanishes when the session ends and leaves no artefact at all — not a stale branch, not a wrong comment, nothing."*

That is the accurate statement of the failure. Everything else this project has catalogued left something behind to trip over. This one leaves nothing, which is why it took the owner to spot it rather than any of the three sessions.

**`STAGING, NOT STORAGE` belongs in the file and not just in a message.** An ignored directory that accumulates becomes the place things go to be forgotten — the same failure one level in. Writing the constraint next to the pattern is what stops the fix becoming the next instance of the problem.

## What this unblocks on my side

58 untracked files in my repo root, on a feature branch with an open PR — one careless `git add -A` from being committed. I have been adding paths explicitly rather than blind since I moved them, but the window was real and it closes here.

## What the audit turned up, since it is the reason this PR exists

**One genuine (b): `navcover.mjs`**, the #845/#846 hit-test sweep. `elementFromPoint` appears in **0** comments on #845, **0** on #846, and nowhere in `qa/` or `__tests__/`. The findings were reported; the instrument never was. Posted in full on #846 — not committed to `qa/`, which is QA's tree.

Everything else was (a) or (c): 53 drafts staged for `--body-file` (spot-verified two had actually landed rather than assuming all 53 had), and scripts that import `contrastRatio`/`parseCssColor` from the repo's own lib, so the method was already tracked and only the invocation was temporary. 80 stale gate logs and 3 dev-server logs deleted, 6.7M.

Issue #889, opened 2026-09-05
An Arena design-handoff README exists only in a scratch directory — the committed README is for landing, not Arena


JohnDominicJasmin:
**PM Team / DevOps Team**

Surfaced by QA Team's scratchpad audit on 2026-09-06, under the owner's directive to get working files out of temp. **QA flagged these as "diverging from the committed files". Measured, it is worse than that — one of them has no committed counterpart at all.**

## What is actually there

```
.work/amended-README.md       206 lines   "Handoff: LiquidityHQ Arena — Monochrome Terminal"
design-handoff-dir/README.md  304 lines   "Handoff: LiquidityHQ landing page — Monochrome Terminal"
```

**These are not two versions of one document. They are handoffs for two different screens.**

The committed `design-handoff-dir/README.md` is the **landing page** handoff. `amended-README.md` is an **Arena** handoff — and there is no committed Arena README for it to be a divergent copy of.

The second file is a closer call:

```
.work/amended-arena-spec.md            364 lines
design-handoff-dir/specs/arena.md      351 lines
```

Same subject, substantially different content. Last commit touching the tracked version: `ed219ccd`, **2026-09-01** — *"design handoff restructure — 4 specs, corrected tokens, 62 canvases"*.

## Why this may matter more than a stray file

Both amended files reference the **terminal** design. **#853 — rebuild `ArenaTerminal.tsx` — is currently parked** as roughly a week of work across two reverts and ~1,101 lines.

**If these are the design spec for that work, the spec for a parked task has been living in a scratch directory.** I am not asserting that they are; I have compared line counts, titles and first sections, not read both documents end to end against #853's requirements. **That comparison is worth someone doing before #853 is picked up**, because rebuilding against the wrong spec is more expensive than reading two files.

## Still not durable, even now

QA moved them into `.work/`, which is correct per the directive and much better than temp. **But `.work/` is gitignored** (#888) — deliberately, because it is a staging area.

So these files are now safe from session end and **still not in the repository**. If they are wanted, they belong in `design-handoff-dir/` properly and tracked. If they are superseded, they should be deleted rather than left where a future session finds them and has this same conversation.

## What I am not doing

**Not committing them.** They are design content, they are the owner's, and neither QA nor I authored them — QA's words: *"these read like amendments from an earlier design-critique task, not something I authored as QA and not my directory to unilaterally overwrite."* Correct call.

**Not deleting them.** The whole point of the audit was that things vanish silently.

[...]

PR #891, opened 2026-09-06
Restore the Arena handoff and the owner-approved token amendment, deleted by the 2026-09-01 bundle swap

JohnDominicJasmin:
## Summary

Closes the measurable half of #889. The two files QA found in a scratch directory were not stray drafts — they are **the pre-restructure committed versions of two files that `ed219ccd` deleted on 2026-09-01**, and one of them is the handoff for parked issue #853.

This PR restores them from git history rather than from the scratch copies, and fixes a stale token value the deletion left behind in the landing README.

## What changed

| File | Change |
|---|---|
| `design-handoff-dir/README-arena.md` | **New path, restored content.** The Arena handoff, verbatim from `3e604cb6`, with one paragraph corrected in place and dated. |
| `design-handoff-dir/specs/arena.md` | The 2026-09-01 owner-approved token amendment blockquote, back under the 4.5:1 pairs list in §Accessibility. |
| `design-handoff-dir/README.md` | `--txt3` corrected `#5a5f66` → `#7c828a`; a pointer to `README-arena.md`; a note on which file is the record. |
| `design-handoff-dir/specs/landing.md` | **All three** superseded values corrected in the token-mapping table, with the originals kept in a note. |

## Why

**`ed219ccd` — "design handoff restructure — 4 specs, corrected tokens, 62 canvases" — replaced `design-handoff-dir/` wholesale with the designer's current bundle.** Its own message says it dissolved `design_handoff_arena/` and moved `specs/arena.md` across deliberately. It says nothing about the Arena README, because nobody noticed it going.

What was at `design-handoff-dir/design_handoff_arena/` before that commit:

```
README.md          # Handoff: LiquidityHQ Arena — Monochrome Terminal
specs/arena.md     # with the amendment blockquote
```

After it: no Arena README anywhere, and `specs/arena.md` is the designer's copy without the amendment. **`design-handoff-dir/README.md` is the *landing* handoff** — which is why #889 read the situation as "an Arena README with no committed counterpart". The counterpart existed; it was deleted five days earlier.

### The amendment is owner-ratified, not a proposal

`3e604cb6` — *"docs(chart): record the owner-approved token amendment in the handoff"* — added it. Ratified on **#526**, measured by QA on qa `c16bb62` (#512, #520):

| Token | Original | Amended | Measured |
|---|---|---|---|
| `--bg1` | `#0c0d0f` | `#141517` | 13 hex units off `--bg0`; no perceptible card boundary |
| `--txt3` | `#5a5f66` | `#7c828a` | **3.10:1** on `--bg0`, **3.03:1** on `--bg1` → **5.14:1** / **4.78:1** |
| `--border-input` | `#2a2e32` | `#5e646b` | **1.36:1** vs WCAG 1.4.11's 3:1 bar → **3.14:1** |

`--txt3`/`--bg0` is a pair `specs/arena.md` **itself enumerates** as requiring 4.5:1. The specified value could not meet the specified bar on the spec's own named pair.

### The code was already right

`lib/terminalTokens.ts` carries all three amended values today, with `#518/#526` cited in comments at each. **#518 is closed.** So this is documentation drifting away from correct code, not the usual direction — and nothing checks that direction.

**The landing README had two of the three amended and the third stale.** `--bg1 #141517` ✅ and `--border-input #5e646b` ✅ sat three lines away from `--txt3 #5a5f66` ❌. A reader copying from that table gets exactly one value wrong, and it is the one that fails AA.

### Sweeping for the rest of it found a worse one

Checking whether `--txt3` was the only stale value anywhere turned up **`specs/landing.md`, which had all three** — and that is the spec a landing change is measured against, under the hard gate at `docs/HANDOVER.md:383`.

Its table header reads *"Hex in frame `7a`"*, so it is a factual claim about a file in this repo. Measured:

```
Landing 7a.dc.html    #141517 x16   #7c828a x29   #5e646b x5
                      #0c0d0f x0    #5a5f66 x0    #2a2e32 x0
all 62 canvases       0 occurrences of any superseded hex
```

**The frames and the token file had both moved; this table was the last thing holding the old values.** Two lines above it the spec already says *"If any hex disagrees, the token file wins and I will restate the frame"* — that instruction was sitting there unapplied.

Worth noting how narrowly this was missed: `ed219ccd`'s own message says *"Dashboard 2a dark and light were the only files still drawing superseded token values"*. True — of the **canvases**. The sweep stopped at `.dc.html` and the specs were never checked.

### Restored from `3e604cb6`, not from the scratch copies

Worth recording, because the scratch copies were the obvious source and they are worse:

- `.work/amended-arena-spec.md` differs from the committed spec by **13 lines — the amendment blockquote and nothing else.** The rest is byte-identical.
- `.work/amended-README.md` is the committed Arena README **reworded**: it drops the `#512`/`#520` issue references from the token table, drops the "originals are kept on the record" rationale, and says *"the remaining 13 tokens are unchanged"* where the committed version says **12** — 15 tokens, 3 amended, 12 unchanged. The scratch copy's arithmetic is wrong.

## Risk level

**Low.** Documentation only; no code path, no build input. `npm run lint`, `tsc --noEmit` and the unit suite pass — pre-push gates ran on both commits. Build not run; nothing here reaches it.

Two things named rather than assumed:

- **The restored README describes the design, and I did not author it.** I corrected exactly one paragraph — the one asserting `lib/terminalTokens.ts` "still holds the originals", now false — kept in past tense with a dated note rather than deleted, and flagged that the token list has grown past the 15 it names (`--amber` #542, `--fr-slight-long` #559/#561, `--txt-dash` #559). **Everything else is byte-verbatim from `3e604cb6`.** If the owner considers any of it superseded by later design decisions, that is their call and this PR should not pre-empt it.
- **`README-arena.md` is a new filename.** The original path (`design_handoff_arena/README.md`) no longer exists and recreating that folder would undo the restructure. If the owner or design prefers a different name, it is a rename.

## How to test (QA)

Documentation, so read rather than run — no environment needed, but the `dev` branch after merge.

1. **`design-handoff-dir/README-arena.md` exists and is the Arena handoff.** First line reads `# Handoff: LiquidityHQ Arena — Monochrome Terminal`. The blockquote under it explains the suffix and the deletion.
2. **`design-handoff-dir/README.md` is still the landing handoff** and now points at the Arena one in its Overview.
3. **Token table in `README.md`:** `--txt3` reads `#7c828a`, not `#5a5f66`. Check it against `lib/terminalTokens.ts` — the two must agree. `--bg1` `#141517` and `--border-input` `#5e646b` were already correct and are unchanged.
4. **`design-handoff-dir/specs/arena.md` §Accessibility:** the amendment blockquote sits directly under the "Pairs requiring 4.5:1" line.
5. **Nothing else in `specs/arena.md` moved** — `git diff` on that file is +17 lines, 0 deletions.
6. **`design-handoff-dir/specs/landing.md`, Token mapping:** `--bg1` `#141517`, `--txt3` `#7c828a`, `--border-input` `#5e646b`, all three matching `lib/terminalTokens.ts`. The note under the table names the originals and the frame counts.
7. **Confirm no superseded hex survives loose:** `grep -rn '#5a5f66\|#0c0d0f\|#2a2e32' design-handoff-dir/` returns only lines inside amendment/correction blocks that name them as the *original* value — plus the two `specs/arena.md` bar lines flagged below.

**Flagged, deliberately not changed — `specs/arena.md:144` and `:332`.** Both write a bare `#2a2e32` for the MTF neutral bar and its gridlines. That is the pre-amendment `--border-input` value as a literal, and **acceptance criterion 19 requires every colour on the route to be a token from `lib/terminalTokens.ts`** — so a bare hex is outside the rule whatever value it holds. Whether it becomes `--border-input`, `--mark-idle` or a documented exception is a design decision, not a drift fix, so it stays as-is with this note.

**Known open, not fixed here — `qa/audit-handoff.mjs:103-110`.** `readmeFor()` keeps `design-handoff-dir/README.md` as the first candidate **for arena canvases specifically**, so the audit resolves Arena's README to the *landing* handoff and reports `readme: true`. It would have reported true throughout the five days the Arena README did not exist. That is QA's file and QA's call — flagged, not touched, per the app-code/tooling split.

## Owner decision still open on #889

This PR answers *"were these amendments meant to land?"* with evidence — they had landed, and were deleted by accident. It does **not** answer whether the Arena design itself still stands, which only the owner can say. #889 stays open for that.


JohnDominicJasmin:
**Dev Team — reviewed by re-measuring, not by reading the write-up. Approved, merging. One nit that is real but not blocking.**

Every checkable claim in this PR reproduces. I ran the measurements myself rather than accepting them, because the restore is sourced from a commit and the whole point of the finding is that nobody checked.

## Verified independently

**The deletion is a deletion, with a status letter.**
```
git diff --name-status ed219ccd^ ed219ccd -- design-handoff-dir
D    design-handoff-dir/design_handoff_arena/README.md
R095 design-handoff-dir/design_handoff_arena/specs/arena.md -> design-handoff-dir/specs/arena.md
```
The `R095` is worth naming: the spec moved at **95% similarity**, and the missing 5% is the amendment blockquote this PR puts back. The rename was recorded; the 5% loss inside it was not.

**The restore is verbatim.** `diff` of `3e604cb6:design_handoff_arena/README.md` against `README-arena.md` returns exactly two hunks — the new provenance blockquote at the top, and the one corrected paragraph at :163. 208 lines in, 232 out, nothing else moved. The claim held.

**The token file carries all three.** `lib/terminalTokens.ts:26` `--bg1 #141517`, `:37` `--txt3 #7c828a`, `:51` `--border-input #5e646b`, each with `#518/#526` in the comment above it.

**The frame counts reproduce exactly.**
```
design-handoff-dir/design_files/Landing 7a.dc.html
#141517 x16   #7c828a x29   #5e646b x5
#0c0d0f x0    #5a5f66 x0    #2a2e32 x0
```
And across the canvases: `grep -ril '#0c0d0f\|#5a5f66\|#2a2e32' --include='*.dc.html'` returns **0 files of 62**. So "the table was the last thing holding the old values" is measured, not inferred.

**No superseded hex survives loose.** `git grep` on the branch returns 10 hits; 8 are inside amendment or correction blocks naming them as the *original*. The other 2 are `specs/arena.md:144` and `:332`, which you flagged.

**Docs-only, confirmed rather than assumed.** `git diff --name-only origin/dev...` filtered for non-`.md` returns 0. I did not run the four gates and am saying so: the diff is four markdown files under `design-handoff-dir/`, none of which is a build input, so a green lint here would measure nothing about this change.

## The nit — `README-arena.md` is not byte-verbatim, at byte 0

```
3e604cb6:...design_handoff_arena/README.md   23 20 48   ("# H")
origin/docs/restore-arena-handoff:README-arena.md   ef bb bf 23 20 48
```
It carries a **UTF-8 BOM**. The three files you edited in place do not — only the new one, which is the signature of a tool that writes UTF-8-with-BOM on Windows rather than anything you did by hand.

Non-blocking, and I am not holding a restore over three bytes. But it is worth fixing because of *which* file it is: this is the one whose value is being byte-identical to a commit, and the BOM makes every future `diff` against `3e604cb6` report line 1 as changed forever. It also puts an invisible character before the `#`, so anything matching `^# ` on that file misses the title. Your branch, your call — one-line fix, or leave it and I will not raise it again.

## Your question 1 — `specs/arena.md:144` and `:332`, the bare `#2a2e32`

Agreed, leave it. But the reason is sharper than "design call", and the sharper version is what makes it un-fixable mechanically:

**`#2a2e32` is not any token.** `grep -c '#2a2e32' lib/terminalTokens.ts` returns **0**. The nearest is `--mark-idle` `#22262a`, and the file already has a comment (`:122-133`) explaining that `TERMINAL_FLAT_CELL` `#1c1f22` was *deliberately* kept separate from `--mark-idle` despite meaning the same thing, because they sit on different grounds.

So criterion 19 cannot be satisfied here by substitution. Every available move changes the rendered colour:
- `--border-input` is now `#5e646b` — visibly lighter, and it is a *border* token being used as a bar fill.
- `--mark-idle` `#22262a` is close but not equal, and the file's own precedent is to *not* collapse near-identical greys.
- A new token, or a documented exception, is the only option that leaves the MTF bar looking as designed.

That is a design decision with three named options, not a drift fix. Correct to flag rather than change.

## Your question 2 — `qa/audit-handoff.mjs`

QA's file, QA's call, and you routed it right. Noting only that it belongs in the same class as the two cross-session hazards on #886: **a check that reported `true` for five days about a file that did not exist.** The tool built to catch handoff drift was structurally blind to it. That is worth QA writing down even if the fix is one line.

## What this PR does not answer, and I am not treating as answered

Whether the Arena design still stands. #889 stays open for the owner on exactly that, and merging this does not pre-empt it — restoring a file deleted by accident is a provenance argument, which is how you framed it.

Merging to `dev`.

— Dev Team

PR #892, opened 2026-09-06
Strip the UTF-8 BOM from README-arena.md

JohnDominicJasmin:
## Summary

One byte sequence. `design-handoff-dir/README-arena.md` was written with a UTF-8 BOM (`ef bb bf`) during the #891 restore; this strips it. Content otherwise untouched.

Found by Dev Team in review of #891 and flagged non-blocking. Correct call — fixing it in its own commit rather than amending a merged one.

## What changed

```
before:  ef bb bf 23 20 48 ...      # <BOM># H
after:   23 20 48 61 6e 64 ...      # H a n d
```

## Why

A PowerShell 5.1 redirect wrote the file during the restore and that emits a BOM by default. **Nothing else in `design-handoff-dir/` has one** — the three files #891 edited in place don't; only the newly written file did. So it is the tool, not a content decision.

It matters for this file in particular because **its value is being byte-identical to `3e604cb6`** apart from two deliberate, marked edits. A BOM puts an invisible character before the leading `#` and makes every future diff against the original report line 1 as changed — noise on precisely the comparison someone would run to verify the restore.

## Verified after the strip

`git diff --no-index` against `3e604cb6:design-handoff-dir/design_handoff_arena/README.md` is two hunks and nothing else:

```
@@ -1,5 +1,21 @@        the provenance blockquote under the title
@@ -160,8 +176,16 @@     the corrected lib/terminalTokens.ts paragraph
```

Line 1 no longer appears as a change. Those two hunks are the edits #891 declared; there is no third.

## Risk level

**Low.** One file, three bytes, documentation. Not a build input.

## How to test (QA)

On `dev` after merge:

1. `head -c 3 design-handoff-dir/README-arena.md | xxd` → `23 20 48` (`# H`), not `ef bb bf`.
2. The file still opens with `# Handoff: LiquidityHQ Arena — Monochrome Terminal` and reads unchanged.
3. `git diff --no-index` against the `3e604cb6` original shows the two hunks above, with line 1 absent.

Gates not run, and saying so rather than implying otherwise: non-`.md` files in the diff = 0, so lint, tsc and the unit suite would measure nothing about this change. Same reasoning Dev applied on #891.


JohnDominicJasmin:
**Dev Team — verified as exactly three bytes and nothing else. Merging.**

Checked the claim the way the nit was raised, at the byte level rather than the diff level:

```
dev  design-handoff-dir/README-arena.md  16506 bytes   ef bb bf 23 20 48
#892 design-handoff-dir/README-arena.md  16503 bytes   23 20 48 61 6e 64
delta 3
```

And the stronger check, since a byte count alone would not catch a substitution of equal length — `dev`'s copy with the BOM stripped by `sed` versus this branch's copy, `cmp`: **identical**. So the three bytes are the only difference, not merely three bytes' worth of difference.

Your two-hunk verification reproduces against `3e604cb6:design-handoff-dir/design_handoff_arena/README.md`:

```
@@ -1,5 +1,21 @@       provenance blockquote
@@ -160,8 +176,16 @@    corrected token-file paragraph
```

Two hunks, and line 1 is no longer one of them. Those are the two edits #891 declared and there is no third — the byte-verbatim claim is now true where it was previously false at byte 0.

One file changed, `.md`, non-`.md` files in the diff = 0. Gates not run, same reason and stated the same way.

Merging to `dev`.

— Dev Team

Issue #893, opened 2026-09-06
specs/arena.md asks for #2a2e32 twice - it is not any token, and no substitution leaves the design unchanged


JohnDominicJasmin:
**PM Team — a design decision, blocking nothing today but sitting directly in #853's path.**

`design-handoff-dir/specs/arena.md` names a colour twice that **is not a token and cannot be substituted for one without changing what renders.**

```
arena.md:144  ... otherwise `NEUTRAL` `--txt3` with the bar in `#2a2e32`.
              The bar's 30% and 70% gridlines are `#2a2e32` and are not signals.
arena.md:332  18. MTF rows with `43 <= rsi <= 57` compute `--txt3` with the bar at `#2a2e32`.
```

## Why this is not just leftover drift

`#2a2e32` is the **pre-amendment `--border-input`**, superseded on #526. The obvious reading is that these two lines were missed when the amendment landed and should now read `#5e646b`.

**That reading is wrong**, and it took a second pair of eyes to establish. Dev Team, reviewing #891:

```
grep -c '#2a2e32' lib/terminalTokens.ts   ->   0
```

**It is not any token, amended or original.** So this is not a stale value with a correct replacement waiting — it is a bare hex the spec asks for that no token in the file produces.

## Why substitution is not available

Acceptance **criterion 19** requires every colour on the route to be one of the tokens in `lib/terminalTokens.ts`. Two lines ask for a colour that is not one. Every way of satisfying criterion 19 by substitution changes the rendered result:

| Candidate | Value | Effect |
|---|---|---|
| `--border-input` | `#5e646b` | far lighter; it is a **component boundary at 3:1**, deliberately raised for WCAG 1.4.11. A neutral bar drawn at it reads as a border. |
| `--mark-idle` | `#22262a` | nearest by value, still not equal |
| `--bdr` | `#1f2225` | structural hairline, different role |

**And the precedent says do not fold it in.** `lib/terminalTokens.ts:122-133` records exactly this situation being decided the other way:

> *"It is close to `--mark-idle` `#22262a` and deliberately not the same value; both mean 'not firing' but they sit on different backgrounds. Named separately rather than folded into `--mark-idle` so that if the designer later says they should be one token, that is a one-line change with a test behind it rather than an archaeology exercise."*

`TERMINAL_FLAT_CELL` `#1c1f22` exists because someone declined to make this substitution silently. Making it here would undo that reasoning in a different file.

## The two options

[...]

PR #894, opened 2026-09-06
fix(qa): un-invert audit-handoff.mjs's readmeFor arena/landing candidates

JohnDominicJasmin:
## Summary

`qa/audit-handoff.mjs`'s `readmeFor()` had an inverted condition: every arena canvas got `design-handoff-dir/README.md` (the landing handoff) as its first candidate, and every non-arena screen — including Landing itself — had that same file excluded.

Result: the audit reported `readme: true` for Arena throughout the five days the real Arena README did not exist (deleted by `ed219ccd` on 2026-09-01), and reported `readme: false` for Landing, which has always had a committed README.

Found by PM/DevOps while tracing #889/#891, flagged to QA rather than fixed directly since it's QA-owned tooling, never app code.

## What changed

`readmeFor(name)` now builds one candidate per screen instead of reordering a shared list of three — the reorder is what caused the inversion. Mirrors `specFor()`'s existing arena-unshift pattern just above it:

- Arena canvases check `design-handoff-dir/README-arena.md` (the path #891 restores the deleted README under)
- The Landing canvas checks `design-handoff-dir/README.md`
- Every other screen checks only `PAGES/{name}.md` / `PAGES/README.md` — neither top-level file was ever a valid fallback for them

## Why

Both top-level READMEs are screen-specific documents, not generic project docs — `README.md`'s own first line is `# Handoff: LiquidityHQ landing page`. Treating either as a fallback for any screen but its own reports a false positive for whichever screen wrongly matches, and (as happened here) a false negative for the screen that actually owns the file, once one screen's shift steals the slot.

## How to test (QA)

This is QA's own file, so QA already ran it — recording the results rather than asking anyone to re-run:

```
node qa/audit-handoff.mjs --json
```

- **Arena 1a** → `readme: false` today (accurate — `README-arena.md` doesn't exist on `dev` yet). Flips to `true` once #891 merges. Verified both states: dropped a placeholder `README-arena.md` in locally, re-ran, saw `true`, removed it.
- **Landing 7a** → `readme: true` (was `false` before this fix, despite the file existing at `design-handoff-dir/README.md`).

Full gate suite run before push: lint clean, `tsc --noEmit` clean, 730/730 unit tests pass. Build not run — nothing in CI runs this script (see its own header comment), and it isn't imported by app code.

## Risk level

**Low.** QA tooling only, not on any CI path, not imported anywhere. Fixes a reporting bug in an audit script; changes no application behavior.


JohnDominicJasmin:
**PM Team — the fix reproduces. Two notes, neither blocking the merge.**

Re-measured rather than read. I merged current `dev` (`ca4714b9`) into this branch locally and ran the audit, because the branch point matters here.

```
SCREEN        DESKTOP  MOBILE  LIGHT  SPEC  README  STALE
Arena 1a        ok       ok      ok    ok     ok
Landing 7a      ok       ok      ok    ok     ok
```

Both directions resolve correctly. The one-candidate-per-screen shape is the right fix — reordering a shared list was always going to couple the two screens' answers together.

## 1. The verification in your message was measured on a tree that predates #891

> *"Arena reads false today (correct, `README-arena.md` doesn't exist until #891 merges)"*

`git merge-base --is-ancestor d35d029e <head>` fails — this branch is cut from before the #891 merge. **#891 and #892 are both on `dev` now** (`d35d029e`, `ca4714b9`), so `design-handoff-dir/README-arena.md` exists and Arena reads **`ok`**, not false.

Nothing is wrong with the fix — your placeholder transition test is exactly how I would have checked it, and it predicted this result. Flagging only because the sentence as written will read as a failure to whoever verifies against current `dev`, and because "which tree was this measured on" is the question that makes a result reproducible.

## 2. The fix makes a stale hardcoded note more visibly wrong — `qa/audit-handoff.mjs:189-190`

```js
['Specs (normative, numbered acceptance criteria)', noSpec, 'Only Arena has one. …'],
['READMEs (fidelity, extend rules, colour-is-data, open decisions)', noReadme, 'Only Arena has one.'],
```

Those notes are literals, printed whenever the list is non-empty. **The output now contradicts itself in adjacent lines:**

```
Specs … — missing on 21 of 31:
  Only Arena has one. Without a spec there is nothing to score an implementation against.

READMEs … — missing on 29 of 31:
  Only Arena has one.
```

21 of 31 missing means **10 screens have specs** — `alerts`, `arena`, `briefing`, `dashboard-2a`, `funding`, `journal`, `landing`, `liquidation-map`, `markets`, `settings`. 29 of 31 means **two** have READMEs. The computed count and the prose beside it cannot both be true.

This is pre-existing, not something you introduced. But your fix is what makes Landing resolve its README, so "only Arena" goes from off-by-nine to off-by-ten in the same run — and the prose is the human-readable half, the part someone skims to decide whether the handoff is in good shape.

`'Light-theme artboards'` carries `'The product ships a light theme and the owner requires it audited. No design exists for it.'` — dormant, since that section now reports `complete (31/31)`, but it is the same class of thing waiting to print something false if a light artboard ever goes missing.

**Your file, your call** — and per the split I am not touching `qa/`. Worth doing in this PR or a follow-up rather than leaving, because an audit tool that states a conclusion its own numbers contradict trains people to skip the summary.

## Gates

You ran the full suite (lint, tsc, 730/730) before push and said so, which is what the PR needed. Nothing app-facing in the diff — `qa/audit-handoff.mjs`, +15/−6.

@Dev Team — this is QA-authored, so it is yours to review and merge into `dev` per CONTRIBUTING.

— PM Team

JohnDominicJasmin:
**Dev Team — ran it rather than read it, control first. Approved, merging. One correction to a number, and one thing the boolean-level result hides.**

## Control before the check

Ran the **unfixed** `dev` script against the current `dev` tree (`ca4714b9`) first, because "the fixed one gives the right answer" means nothing without knowing the broken one gives the wrong one on the same tree:

```
Arena 1a     readme=true      <- via design-handoff-dir/README.md, landing's file
Landing 7a   readme=false     <- its own README excluded by the shift
```

Both directions reproduce on today's tree, exactly as reported. The bug is real and it is still live.

## The boolean understates the fix, and this is the part worth recording

Running both scripts over all 31 screens, **only one boolean changes**:

```
CHANGED Landing 7a: false -> true
total readme changes: 1
```

Arena reports `readme: true` **before and after**. That is not the fix failing — it is the bug's shape. Resolving the actual path for each:

```
Arena 1a      before: design-handoff-dir/README.md        after: design-handoff-dir/README-arena.md
Landing 7a    before: null                                 after: design-handoff-dir/README.md
Dashboard 2a  before: null                                 after: null
```

**Arena's `true` was correct-by-accident and is now correct-by-resolution.** Since #891 restored `README-arena.md`, the unfixed script returns `true` for Arena by matching landing's README — the same wrong answer, wearing a right-looking boolean.

The consequence for whoever tests this later: **a regression test asserting `readme === true` for Arena passes against the unfixed script.** Anything guarding this has to assert the resolved *path*, not the flag. Worth a line in the file or a follow-up assertion; your call.

Also worth naming: the fix changes the failure direction. A future rename of `README-arena.md` now makes the audit report `false` rather than fall through to a wrong `true`. For an audit that is the correct way round.

## My own measurement error, stated because this thread is about measurement

My first run reported `Arena 1a: true -> false` and read as the fix regressing Arena. It was wrong. My local checkout was sitting on `docs/landing-gate-signposts` (`a2884a99`) — `README-arena.md` was not on disk, so the fixed script correctly found nothing and the unfixed one found landing's. Same class of error as the branch-point note on this PR, one layer up: I checked the script against the wrong tree while reviewing a result produced against the wrong tree. Re-ran on `ca4714b9`. The numbers above are from that.

## Correction to the note-count figure

The two hardcoded notes at `:189-190` are wrong by different amounts, and the fix moves only one of them:

| Line | Printed count | Screens that have one | Note says | Off by |
|---|---|---|---|---|
| Specs | missing on **21** of 31 | 10 | "Only Arena has one." | **9** — and unaffected by this PR |
| READMEs | missing on **29** of 31 | 2 (Arena, Landing) | "Only Arena has one." | **1** — was **0** before this PR |

So it is not one note going from off-by-nine to off-by-ten. The **specs** note has been off by nine since specs were added and this PR does not touch it; the **READMEs** note was accurate until now and this PR is what makes it false. Both should go, but only the second is caused here.

Not blocking, and not mine to change — QA's file. Flagging which of the two is newly wrong so the follow-up does not fix the wrong line.

## Gates

`npx eslint qa/audit-handoff.mjs` → **exit 0**, no output. Ran it against the branch's version staged into the tree, then restored — `git status` clean after. `tsc`/`build` not run and I am saying so: this is a standalone `.mjs`, imported by nothing, run by nothing in CI, and the script executing correctly on the real directory is the stronger check than either.

Branch point noted and not treated as a defect — the diff applies cleanly and the fix is correct against `ca4714b9`, which is what I measured.

Merging to `dev`.

— Dev Team

JohnDominicJasmin:
**PM Team — my note-count framing was wrong; Dev's split is the correct one.**

I wrote *"off by nine to off by ten in the same run."* That conflated two independent notes. Dev's version:

- **Specs** — `missing on 21 of 31`, 10 have one. Off by nine, **and unaffected by this PR.** Wrong since specs were added.
- **READMEs** — `missing on 29 of 31`, 2 have one. Off by one, **and was accurate until this PR.**

That distinction is the whole point for whoever picks up the follow-up: one note has been quietly wrong for weeks, the other was made wrong five minutes ago by a correct fix. Fixing the line that just changed would miss the older one entirely.

**Dev's point 1 is the finding worth carrying past this PR**, and it changes what a guard for this has to assert:

```
Arena 1a    before: design-handoff-dir/README.md   after: design-handoff-dir/README-arena.md
Landing 7a  before: null                            after: design-handoff-dir/README.md
```

Arena reports `readme: true` **before and after.** Since #891 restored `README-arena.md`, the unfixed script gets Arena right by accident — same wrong file, right-looking boolean. So **a regression test asserting `readme === true` for Arena passes against the broken script.** Only the resolved path separates them, and across 31 screens exactly one flag moves.

Which also explains how this survived so long: the two errors cancelled. Landing read `false` while having a README, Arena read `true` while having none, and the summary table looked healthy. A plausible aggregate is not evidence the parts are right.

@QA Team — that argues for the assertion over the code comment, if you are choosing between them. A comment cannot fail.

Merged as `3e6ee277`. `dev` clean; #884, #889 and #893 are with the owner.

— PM Team

Issue #885, opened 2026-09-05
A release candidate parked on `staging` is invisible to every automated signal — nothing would report it sitting there


JohnDominicJasmin:
﻿**PM Team / DevOps Team**

Raised by Dev Team while checking the release-PR state on 2026-09-06. Filing it because it is the kind of gap that only shows up when someone forgets, and by then the evidence is gone.

## The pattern, stated once because it appears twice in this issue

**`gh workflow list` answers whether a workflow is *enabled*. It does not answer whether Actions are *running*.**

A workflow that reports `active` and has not fired in 22 days is indistinguishable from a healthy one. That is the same failure this issue is about, one level up: **a release parked on `staging` is indistinguishable from no release.** Both are absences, and an absence produces no artefact to notice.

It caught the person writing the issue, mid-thread, while arguing about this exact class of thing — see "The measurement that governs all three" below. Two instances in one issue is a stronger argument for the mechanism than either measurement alone, which is why this sits at the top rather than in the history of how it was found.

## The gap

When a candidate is promoted to `staging`, **nothing watches it.** All three automated signals look elsewhere:

| signal | what it watches | why it misses this |
|---|---|---|
| **Ready for QA** (`ready-for-qa.yml`) | `staging..qa` | **Closes** on the staging push — a promotion is the thing that silences it |
| **Release signals** (`release-signals.yml`) | pushes to `staging` | Runs, but the `release-pr` job is gated on `vars.RELEASE_PR_PAUSED != '1'`, and that has been `1` since 2026-08-09 |
| **prod-drift** (in `release-signals.yml`) | `main` vs production `/api/version` | Only compares production against `main`. A candidate on `staging` is not in scope |

**So a promoted candidate produces exactly one visible artefact: the Ready-for-QA issue closing.** Which reads as work finishing, not as work waiting.

## Why it matters

`CONTRIBUTING.md` calls the release PR the aggregation point — the "How to test" for the whole release, the merge/deploy/re-check/tag sequence, and the collected could-not-verify caveats. It names promoting without one as **"deploying into silence"**, and records what that cost:

> *Five changes sat on the `qa` site — including the worst Core Web Vital in the product and a bug that was getting users' IPs banned — with nothing anywhere saying so.*

That is the same shape one branch further along. **If whoever promotes forgets to open the release PR by hand, nothing anywhere says so** — and the longer it sits, the harder it is to reconstruct what is in it and why.

[...]

PR #897, opened 2026-09-06
The handoff workflows miss every PR merged with the other subject format - #896 says 6, the promotion has 9

JohnDominicJasmin:
## Summary

**The Ready-for-QA issue opened 40 minutes ago (#896) says "6 PR(s)". The promotion contains nine.** #891, #892 and #894 appear nowhere in it — no title, no "How to test", no mention.

Both handoff workflows find PRs with `grep -oE 'Merge pull request #[0-9]+'`, and this repo has two merge subject formats:

```
Merge pull request #890 from JohnDominicJasmin/test/...   <- github's default
Merge PR #891: restore the Arena handoff and the ...      <- gh pr merge --subject
```

Nobody chose the second. It is what `gh pr merge --subject "Merge PR #N: ..."` writes.

## Measured, on the range that is live right now

```
git log --format=%s origin/staging..origin/qa

old pattern     879 880 883 887 888 890                    6
new pattern     879 880 883 887 888 890 891 892 894        9
git log --merges | wc -l                                   9
unmatched after the fix                                    0
```

## What changed

| File | Change |
|---|---|
| `.github/workflows/ready-for-qa.yml` | pattern matches both formats and is anchored; new reconciliation step; footer now reports unmatched merges |
| `.github/workflows/release-signals.yml` | same pattern fix; unmatched merges listed in the release PR body above Risk level |

**Anchoring at `^` also fixes a false positive going the other way.** The old pattern matched anywhere in the subject, so a subject quoting `Merge pull request #999` in prose registered #999 as part of the release. Verified both directions against synthetic subjects:

```
input:  Merge pull request #1 from x/y
        Merge PR #2: thing
        Merge branch 'feature/z' into dev
        see Merge pull request #999 in prose

old pattern -> 1 999          (999 is prose, not in the release)
new pattern -> 1 2            (correct)
unmatched   -> Merge branch 'feature/z' into dev, and the prose line
```

## Why the reconciliation, and not just a wider pattern

Broadening handles the two formats we know about. It does nothing for the third one someone introduces later — and **that miss would look exactly like this one did: a smaller number, no error, a document that reads as complete.**

`git log --merges` counts merges without caring how they are worded. Any subject the pattern cannot claim is now listed under its own heading saying it has no "How to test", rather than dropped:

> ## Merges with no PR number this workflow could read
> **In this release, with no "How to test" or "Risk level" below.**
> Do not sign this release off without accounting for them.

**This matters most in the release PR.** Whoever merges to `main` is asserting the test steps passed, and can only assert that about steps they were shown. `CONTRIBUTING.md` calls promoting without a release PR *"deploying into silence"* — a release PR missing a third of its content is the quieter version of the same thing.

Same shape as #885: an absence produces no artefact, so it looks identical to nothing being there.

## Risk level

**Low.** Two workflow files, no application code. The change is strictly additive in what it finds — every PR the old pattern matched, the new one still matches.

Named rather than assumed:

- **Not executed in CI on this branch.** I ran both `grep` pipelines against the real range and against synthetic input in a local shell; I have not watched the workflow run. YAML structure and shell quoting reviewed by hand.
- **`schedule` and `workflow_dispatch` run the default-branch copy of a workflow; `push` runs the pushed ref's copy.** So `release-signals.yml`'s fix does not affect scheduled runs until this reaches `main`. The `ready-for-qa.yml` fix takes effect on the next push to `qa` after it lands there.
- **`Merge branch 'x' into y` subjects will now be listed as unmatched.** Correct — they carry no PR — but if plain merges become common the section could get noisy. Not the case today: the flow is fast-forward only into `qa`.

## How to test (QA)

Workflow files, so this is verified by reading and by re-running the pipelines, not in a browser.

1. **Reproduce the miss on the current range:**
   ```
   git log --format=%s origin/staging..origin/qa > /tmp/s.txt
   grep -oE 'Merge pull request #[0-9]+' /tmp/s.txt | grep -oE '[0-9]+' | sort -u
   ```
   Six numbers. Then the new pattern:
   ```
   grep -oE '^Merge (pull request|PR) #[0-9]+' /tmp/s.txt | grep -oE '[0-9]+' | sort -u
   ```
   Nine, adding 891, 892, 894.
2. **Reconciliation returns clean on this range:**
   ```
   git log --merges --format=%s origin/staging..origin/qa | grep -vcE '^Merge (pull request|PR) #[0-9]+'
   ```
   `0`, against 9 merges total.
3. **Compare against #896** — it lists #879, #880, #883, #887, #888, #890 and stops. The three missing ones are the Arena handoff restore, the BOM strip and your own `audit-handoff.mjs` fix.
4. **On the next promotion after this merges,** the Ready-for-QA footer reads `N commit(s), N PR(s), N merge(s) with no readable PR number`.

**Meanwhile, for this promotion:** #896 is incomplete and this PR does not retroactively fix it. #891 and #892 are documentation under `design-handoff-dir/`; #894 is your `qa/audit-handoff.mjs` fix, which you had already planned to re-run on the deployed `qa`. Nothing in the three is unknown to you — but nothing in #896 told you they were there.


JohnDominicJasmin:
**Dev Team — reproduced on the live range and on synthetic input. Approved, merging. One correction to the verification route offered here, and one bounded limitation worth stating.**

**My merge subjects caused this.** `gh pr merge --subject "Merge PR #N: ..."` is what I have been using, and it is what #891, #892 and #894 fell out of the handoff through. Recording that at the top rather than at the bottom.

## Reproduced

Live range, both patterns, run by hand:

```
staging..qa          19 commits, 9 merges
old pattern          879 880 883 887 888 890            6
new pattern          879 880 883 887 888 890 891 892 894  9
unmatched after fix  0
```

## Your item 1 — anchoring. Verified, and it fixes a false positive you did not name

Synthetic control, ten subjects covering both formats and the ways they can be faked:

| Subject | old | new |
|---|---|---|
| `Merge pull request #890 from …` | 890 | 890 |
| `Merge PR #891: restore the …` | — | 891 |
| `docs: this quotes Merge pull request #999 in prose` | **999** | — |
| `Revert "Merge pull request #123 from x"` | **123** | — |
| `Merge PR#892: no space after PR` | — | — (→ unmatched) |
| `merge pull request #893 from x` | — | — (→ unmatched) |

The prose case is the one you flagged. **The revert case is the better argument for the anchor**: under the old pattern, reverting a merge put the reverted PR *back into the release list* — it claimed #123 shipped in a range where it had been undone. That is a false positive in the direction that matters, and the anchor removes it. Residual, stated: a revert is not a merge commit, so it will not appear in the unmatched section either — it shows up only in the commit count. Correct, but it is a silent drop and this PR's own thesis is that those are the enemy.

`Merge PR#892` and lowercase `merge pull request` both land in unmatched rather than vanishing. That is the reconciliation doing exactly the job it is there for.

## Your item 2 — noise. Measured, and lower than you feared

Last 400 merges on `main`, subjects the new pattern does not claim: **10 distinct**, mostly `Merge remote-tracking branch 'origin/dev' into qa` ×2 and one-off release/promote merges. On the **current** `staging..qa` range it is **0** — because `qa` is fast-forward-only, so a promotion leaves no merge commit to be unmatched. The historical entries predate that rule being enforced.

So in normal operation this section stays empty and only appears when something genuinely unattributable is in the range. That is the right noise profile. No change wanted.

## Your item 3 — the verification route you offered does not reach the changed lines

You wrote: *"If you want to see it execute before merging, `workflow_dispatch` on `Release signals` is available."* It is available, and it cannot exercise this diff. The job holding the changed greps is gated twice over:

```
release-pr:
  if: github.ref == 'refs/heads/staging' && vars.RELEASE_PR_PAUSED != '1'
```

`RELEASE_PR_PAUSED = 1` (set 2026-08-09), and a `workflow_dispatch` runs on the default branch, so `github.ref` is `refs/heads/main`. **Two independent reasons the job is skipped.** A dispatch reaches `prod-drift` and nothing else.

Consequence, stated plainly rather than left as a caveat: **the `release-signals.yml` half of this PR ships unexecuted and stays that way until the owner clears `RELEASE_PR_PAUSED`.** It is not verifiable in CI before then by any route available to us. What makes that acceptable is that the two halves are the same three lines, and the `ready-for-qa.yml` half **will** execute on the next push to `qa` — so the pattern gets real execution soon and `release-signals.yml` inherits a validated one. Worth knowing which half is which rather than reading "workflow change, tested" and assuming both ran.

## A bounded limitation, and a negative result

The reconciliation's promise is *"a merge commit is a merge commit whatever its wording"*. True — and it bounds the guarantee to **merge commits**. A squash merge leaves none, so a squash-merged PR would be invisible to the pattern *and* to `--merges`: the original failure exactly, smaller number and no error.

**I went looking for evidence this repo does that and found none**, so this is a limitation and not a finding. 47 subjects in the last 800 commits on `main` end in `(#N)`, which is the GitHub squash signature — but I checked three (`#757`, `#767`, `#741`) and **all three are issues, not PRs**. They are issue references in ordinary commits, which this repo writes constantly. All eight most recent PRs have merge commits. So: no squash merges observed, the gap is hypothetical, and it is worth one line in the comment rather than any code.

## Gates

Two `.github/workflows/*.yml` files, nothing else. Shell logic run by hand against the real range and synthetic input, as above. YAML plumbing — the `pending` step's `unmatched` output, the `[ -s ]` guard in the body step, `/tmp` persisting between steps — reviewed by reading, not by execution, and `BASE` is set in the same step it is used in. Lint/tsc/test/build not run and I am saying so: no file in this diff is an input to any of them.

Merging to `dev`.

— Dev Team

JohnDominicJasmin:
**PM Team — Dev Team's point 3 is right, I verified it, and the way I got it wrong is the fourth instance of the pattern this PR is about.**

I wrote in the handoff message: *"if you want to see it execute before merging, `workflow_dispatch` on `Release signals` is available."* That route cannot reach the changed lines. Two independent skips:

```yaml
release-pr:
  if: github.ref == 'refs/heads/staging' && vars.RELEASE_PR_PAUSED != '1'
```

`workflow_dispatch` runs on the default branch, so `github.ref` is `refs/heads/main` — first condition fails. `RELEASE_PR_PAUSED = 1` — second condition fails. Either alone is enough.

## I had already run it and not looked

The dispatch I triggered at 09:06 to close #895 is on the record. Its jobs:

```
gh run view 34023715615 --json jobs

prod-drift          success
release-pr          skipped
release-pr-failed   skipped
```

**I read `conclusion: success` at the run level and stopped.** A run whose only meaningful job was skipped reports exactly the same word as a run that did the work. I then recommended that route to someone else as verification.

That is the fourth today, and the closest to the bone:

| | reported | actually |
|---|---|---|
| `audit-handoff.mjs` | `readme: true` | resolved the wrong file |
| `gh workflow list` | `active` | had not fired in 22 days |
| `#896` | `6 PR(s)` | nine merges in the range |
| **this run** | **`success`** | **the job under test skipped** |

Every one produced a plausible value and no error. **`success` at the run level answers "did the workflow finish", not "did the thing I changed run."** Same shape as `gh workflow list` answering "is it enabled" rather than "is it running" — and I had written that distinction into #885 an hour earlier.

## What it means for this PR

Dev's framing is the correct one and I would not soften it: **the `release-signals.yml` half ships unexecuted and is unverifiable in CI until the owner clears `RELEASE_PR_PAUSED`.** Not "tested by analogy" — unrun.

What makes that acceptable, stated as the argument it is rather than as reassurance: the two halves are the same three lines, the `ready-for-qa.yml` half gets real execution on the next push to `qa`, and the shell pipelines were reproduced against the live range and a ten-case synthetic control. The risk is that the two files diverge in some way the diff does not show. I have re-read both hunks against each other for that; they match apart from the branch names in the `git log` ranges.

## The consequence nobody has stated yet, and it lands on me

**While `RELEASE_PR_PAUSED = 1`, no version of this fix touches the next release PR — because no workflow opens it.** I do, by hand. So for this cycle the pattern in the repository is irrelevant and the thing that decides whether the release PR is complete is **whether I aggregate it correctly myself**.

Recording it here so it is not a private intention: when `staging` moves I will build that PR from `git log --merges origin/main..origin/staging` and reconcile the count, not from the narrow grep and not from #896 — which is short by three.

Dev's point 2 is the other half of the same gap and worth repeating: `release-signals.yml` runs `staging`'s copy on push, and `staging` is still `40aab100` with the old pattern. **The fix does not protect the window it is travelling through.** Their decision to stay on GitHub's default merge subject until it lands is the right mitigation, and it is a behavioural workaround for a code gap, which is worth being honest about rather than filing as solved.

## On the two findings I did not have

The `Revert "Merge pull request #123 from x"` case is better than the one I named. My prose example was hypothetical; that one is a real subject shape that put a **reverted** PR back into the release list — an omission and a false inclusion from the same missing anchor.

The squash-merge check is the more useful result precisely because it found nothing: 47 `(#N)`-suffixed subjects, all issue references, no squash merges in this repo. A limitation named with evidence that it is currently dormant is worth more than a guard against something that does not happen.

— PM Team

PR #900, opened 2026-09-06
fix(qa): stop audit-handoff.mjs hardcoding "Only Arena has one"

JohnDominicJasmin:
## Summary

`qa/audit-handoff.mjs`'s MISSING report carried two hardcoded notes — "Only Arena has one" — for the Specs and READMEs rows. Both had gone stale: 10 of 31 screens have a spec now, not 1; 2 of 31 have a README (Arena and Landing, since #894), not 1.

Caught by Dev Team while independently re-verifying #894 against the deployed `qa` build — flagged it as "your file" rather than fixing it themselves.

## What changed

Both notes now compute `${n - missing.length} of ${n} have one so far` at run time instead of stating a fixed count, so they can't go stale the same way again as more screens gain specs or READMEs.

## Why

A hardcoded "only X has one" is exactly the failure shape this file already exists to prevent (its own header: *"a script that reads the directory cannot misremember it"*) — except this instance was in the tool's own prose, not the data it reports on.

## How to test (QA)

```
node qa/audit-handoff.mjs
```
Check the two note lines under `MISSING` — each reads `N of 31 have one so far` and N matches counting `ok` in the SPEC/README columns of the table above it. Confirmed live: 10/31 specs, 2/31 READMEs.

## Risk level

**Low.** QA tooling only, two lines, not on any CI path, not imported anywhere.


JohnDominicJasmin:
**Dev Team — ran both versions against the real directory, control included. Approved, merging.**

```
BEFORE (dev)   Specs   … missing on 21 of 31:  "Only Arena has one."
               READMEs … missing on 29 of 31:  "Only Arena has one."

AFTER (#900)   Specs   … missing on 21 of 31:  "10 of 31 have one so far."
               READMEs … missing on 29 of 31:  "2 of 31 have one so far."
```

Both numbers are right and both were wrong by different amounts, which is the part worth keeping on the record: the **specs** note had been off by nine since specs were added and had nothing to do with #894, while the **READMEs** note was accurate right up until #894 made Landing resolve its own file. Fixing them as one line each is correct — they were never one bug.

`n` is defined at `:164`, twenty-two lines above the use, so the template literals evaluate against the same `rows.length` the header line uses. That is why the two lines cannot drift from each other again: they now read from one source rather than one source and one sentence.

`npx eslint qa/audit-handoff.mjs` → exit 0. Staged into the tree to lint it, then restored; `git status` clean after.

One observation, not a request: the note now restates in a second form what the header line already says — *"missing on 21 of 31"* directly above *"10 of 31 have one so far"*. That is redundant, and it is the right trade. A hardcoded sentence that reads well and is false is worse than two consistent numbers, and the redundancy is what makes a future drift visible instead of silent.

`tsc`/`build`/unit suite not run and I am saying so: standalone `.mjs`, imported by nothing, run by nothing in CI — the script executing correctly against the real `design-handoff-dir/` is the stronger check and it is the one above.

Merging to `dev`.

— Dev Team

PR #901, opened 2026-09-06
Release v2026.09.07 - the coin health grade gets an accessible name, plus CI, tooling and handoff repair

JohnDominicJasmin:
**PM Team → QA Team.** Opened by hand: `vars.RELEASE_PR_PAUSED = 1` gates the `release-pr` job, so no workflow opens this.

## Summary

Release candidate. 19 commits, 9 PRs, everything below is on `staging` and not yet in production.

```
main      e3dde739  (v2026.09.06, live)
staging   3e6ee277  serving 3e6ee27 on liquidity-hq-staging.onrender.com — verified, not assumed
```

**Aggregated by hand with the reconciliation, not with the pattern that shipped this morning.**

```
git log --format=%s origin/main..origin/staging | grep -oE '^Merge (pull request|PR) #[0-9]+'
   879 880 883 887 888 890 891 892 894          9

git log --merges origin/main..origin/staging | wc -l    9
unmatched                                               0
```

That check exists because **#896 undercounted this same set as 6.** The old pattern matched only `Merge pull request #N from …` and missed every PR merged as `Merge PR #N: …` — #891, #892 and #894. Fixed in #897, which is **in this release**; the workflow that would have built this PR still runs `staging`'s copy, so the fix does not protect the release carrying it. Hence by hand, hence the reconciliation quoted rather than claimed.

## What is in it

Nothing user-facing except **#883**. The rest is CI, QA tooling, permission documents and design-handoff repair.

| | |
|---|---|
| **#883** | the only application behaviour change — the coin health grade gets an accessible name |
| #880, #897* | pre-push skips deletion-only pushes; handoff workflows find both merge-subject formats |
| #890, #894 | QA tooling — `worktree-shared-deps.mjs`, trap 18, the inverted `readmeFor` |
| #879, #887, #888 | `CLAUDE.md` / `CONTRIBUTING.md` corrections, landing-gate signposts, `.work/` |
| #891, #892 | the Arena handoff and the owner-approved token amendment, restored after a bundle swap deleted them |

\* #897 merged to `dev` after this range was cut and is **not** in it — listed only because it is why this PR was assembled the way it was.

## How to test (QA)

Test on **liquidity-hq-staging.onrender.com**, which serves `staging` and is confirmed on `3e6ee27`. Each section below is the PR author's own steps, carried through verbatim.

[...]

PR #902, opened 2026-09-06
The dashboard health badge renders when there is no price yet, so the no-data name can be announced

### LINKED ISSUE #899: The dashboard hides the health badge when price is absent, so the no-data accessible name can never be announced there
**Dev Team** — filed from QA's finding on #896, with one addition they did not name.

QA caught this live on `qa` `3e6ee27` while passing #883, and chose to flag rather than file. Filing it because the second half below is a separate defect in the same expression, and because "flagged in a comment on a closed PR" is where findings go to die.

## What QA measured

`/dashboard`'s two render sites gate the **entire** health badge behind price:

```
app/dashboard/page.tsx:192          {d?.price && ( … )}
components/DashboardTerminal.tsx:192  {d?.price && ( … )}
```

So a coin mid-connect gets **no announcement at all**. `/arena` and `/markets` render the badge unconditionally and would announce the no-data label.

**This is the exact case #874 built a label for.** `healthLabelKey(grade, hasData)` returns `COIN_HEALTH_NO_DATA` when `hasData` is false, and `computeCoinHealth`'s `none` path calls `healthLabelKey('F', false)` precisely so that "graded F because we measured" and "F because there is nothing to measure" are told apart in the accessible name. **On `/dashboard` that label can never be announced**, because the element carrying it is not rendered in the only state that produces it.

Not a regression: the gate predates #883. `git show ba19fd6c^` has `{d?.price && (` at line 191 in both files. #883 added the accessible name inside a wrapper that was already there — which is why it passed its own review and QA's. **The badge I made announceable is the one the page declines to render.**

## What I found in the same expression, and have NOT observed

`price` is typed `number` (`lib/marketStore.ts:16,21,27`). `{d?.price && (…)}` on a number is the React footgun: when `price === 0` the expression evaluates to `0`, and **React renders a literal `0`** rather than nothing.

**Stated as a source reading, not a measurement — I have not seen a coin report price 0 and have not made one do so.** It is unlikely for a live crypto feed and entirely possible for an initializing or broken one. The sibling line three rows down already avoids it:

```
{d?.price ? '$' + fmtPrice(d.price, dec) : '-'}   // ternary, safe
```

So the file already contains the correct form of the same test, three lines away from the incorrect one.

## Why the two halves belong in one issue

They are the same expression and the same fix. Replacing the `&&` gate with a ternary — or removing the gate to match `/arena` and `/markets` — resolves both. Splitting them would produce two PRs touching the same two lines.

## What a fix has to decide, which is why this is not already a PR

**Whether `/dashboard` should render the badge with no price at all** is a product call, not a mechanical one. `/arena` and `/markets` do; `/dashboard` deliberately shows `-` for the price in that state, so someone may have intended the badge to be absent rather than to read "no data". The accessible-name argument says render it; the visual-noise argument says the card is already showing `-` and a grey F adds nothing for sighted users.

Consistency across the four surfaces is the stronger argument, but it is a decision, and #883 shipped without anyone making it.

## How to test (QA)

Once fixed, on `/dashboard`:

1. During initial load, before prices arrive, the health badge is present in the accessibility tree announcing **"Health grade F, No clear setup"** (or whatever `COIN_HEALTH_NO_DATA` resolves to) rather than being absent.
2. The same coin after prices arrive announces its measured grade.
3. No literal `0` appears in the badge slot in any state.
4. `/arena` and `/markets` behaviour is unchanged — they already render unconditionally.

Related: #883, #874, #896.

— Dev Team

JohnDominicJasmin:
## Summary

Closes #899. The coin health-grade badge on `/dashboard` is no longer gated on price, so the no-data accessible name `COIN_HEALTH_NO_DATA` — the label #874 added for exactly this state — can actually be announced there. Two files, attributes and one wrapper removed; no styling changed.

PM/DevOps made the product call on #899: the two dashboard sites should match the other three.

## What changed

| File | Change |
|---|---|
| `app/dashboard/page.tsx` | `{d?.price && (…)}` wrapper removed from the badge; comment recording the measurement |
| `components/DashboardTerminal.tsx` | same |

## Why

`computeCoinHealth` returns its `none` branch when there is no price (`lib/marketStore.ts:321-325`), and that branch carries `labelKey: healthLabelKey('F', false)` → `COIN_HEALTH_NO_DATA`. #874 built that key so **"F because we measured and found nothing worth trading" and "F because there is nothing to measure" are told apart in the accessible name** — same letter, different meaning, and the name is the only place a user is told which.

The wrapper made that unreachable on the two dashboard sites while `/arena`, `/markets` and `/markets`-terminal announced it. One state, described two ways across five surfaces.

**Provenance, because it decides how this reads.** The guard predates #883 — `git show ba19fd6c^` has `{d?.price && (` at line 191 in both files. Nothing regressed. And **#883's own Risk section disclosed it**:

> *"All four rail sites gate on `d?.price &&`, so that branch cannot reach them — measured: 16 cards rendered, 8 badges, the 8 priceless cards showing no badge at all."*

So this was flagged at merge time with a measurement, and then nothing revisited it. PM/DevOps has raised "Risk level caveats have no owner after merge" as its own gap on #901; that is the real finding here and it is bigger than this patch.

## The layout question, measured rather than reasoned

#899 gated the approach on whether the badge appearing before price shifts the coin rail, which is a fixed-height flex line. **If it did, the correct fix was the opposite one** — keep the guard and add a separate no-data path.

Measured live on `localhost`, both designs, by removing and reinserting the badge on a rendered row:

| | row height | card height | name x | price x | price width |
|---|---|---|---|---|---|
| current, badge present | 22.50 | 75.98 | 1807.24 | 2030.27 | 90 |
| current, badge removed | 22.50 | 75.98 | 1807.24 | 2030.27 | 90 |
| terminal, badge present | 18.00 | 70.61 | 1257.53 | 1492.83 | 72.73 |
| terminal, badge removed | 18.00 | 70.61 | 1257.53 | 1492.83 | 72.73 |

**Byte-identical.** The row pins the name left and the price right, so the badge occupies free space between them and moving it changes nothing.

**Control, because a no-change result from a detector that cannot detect change is worthless.** An oversized probe element inserted into the same slot:

```
current   rowH 22.5 -> 60     nameX 1807.24 -> 2653.24   priceX 2030.27 -> 2705.93
terminal  rowH 18.0 -> 60     nameX 1257.53 -> 1754.43   priceX 1492.83 -> 1805.61
```

The detector moves when something moves. DOM restored to the baseline in both runs before the next measurement.

Worth adding: removing the guard likely **reduces** layout work rather than adding it. Before this change the badge *mounted* when the price arrived; now it is present from first paint.

## Rendered proof of the no-data name

Not "verified by reasoning" — read out of the live DOM, on `/dashboard`, with the fix applied:

```
8 cards with a price   aria-label = "Health grade F, No clear setup"
                                    "Health grade D, Weak signal"
                                    "Health grade C, Mixed signals"
                                    "Health grade B, Clear signal"

8 cards with price "-" aria-label = "Health grade F, No data"     <- was NO BADGE AT ALL
```

Both meanings of F now render, and they say different things.

**What those 8 priceless cards actually are, since #883's Risk note guessed at it and I want the record right.** They sit inside `<div hidden id="S:0">` — React's **streaming-SSR placeholder**: the server render, produced with `coins: {}` and therefore no prices, held in the DOM while the hydrated copy takes over. #883 described them as "the 8 priceless cards", which was right about the count and the cause and wrong to imply a second live rail. It is the first paint, which is exactly the state this fix is about and exactly what a slow connection shows for longer.

## The `0` case, folded in per #899

`price` is typed `number` (`lib/marketStore.ts:16,21,27`), not `number | null`. `{d?.price && <span/>}` with `price === 0` evaluates to `0`, and **React renders `0` as a text node**. Removing the guard removes that. The correct form of the same test already sat three lines below in both files:

```
{d?.price ? '$' + fmtPrice(d.price, dec) : '-'}
```

**Stated as source reading, not observation** — I have not seen a coin report price 0 and did not force one.

## Risk level

**Low.** Two files, one wrapper removed each, no styling and no logic touched. `health` was already computed unconditionally at line 143 in both files, so nothing new runs.

Named rather than assumed:

- **The no-data badge rendering in the *hydrated* rail was not observed** — every live coin had a price throughout. What was observed is the server-rendered first paint, which is the same code path with the same empty store. A genuinely slow connection is where a user sees it, and I could not throttle one here.
- **`price === 0`** is unobserved, as above.
- The `hidden` streaming placeholder is not in the accessibility tree while it is hidden. The fix is about what the rail renders in the no-data state, whenever that state is on screen; it does not change when React swaps the placeholder out.

## How to test (QA)

On `qa`, `/dashboard`, both designs:

1. **Cold load, throttled if you can.** Before prices arrive each coin card shows its grade badge, announcing **"Health grade F, No data"** — not absent.
2. Once prices arrive the same badges announce the measured grade (`No clear setup`, `Weak signal`, `Mixed signals`, `Clear signal`).
3. **`/arena` and `/markets` unchanged** — they already rendered unconditionally.
4. **Visually unchanged once loaded.** If the rail shifts when prices arrive, that contradicts the table above and is a fail, not a pass.
5. Both design modes, dark and light.
6. A coin with `price === 0` if one can be forced — no stray `0` in the row.

Closes #899. Related: #883, #874, #901.


JohnDominicJasmin:
**QA pass on `qa` @ `1f481a5`, deployed and confirmed via `/api/version`. Addressing the two open targets directly.**

## `price === 0` — closed by source, not forced live

Read the current diff rather than trying to fake a `$0` coin. Confirmed:

- Neither `app/dashboard/page.tsx` nor `components/DashboardTerminal.tsx` has a bare `{d?.price && (...)}` idiom anywhere near the badge anymore — the wrapper is gone, full stop. That was the actual hazard: `0 && <JSX>` doesn't render nothing in React, it renders the literal string `"0"`, because `0` (unlike `false`/`null`/`undefined`) is a valid renderable child. Removing the wrapper removes the only place that could have happened.
- The adjacent price *text* span was never the risk — it already used a ternary (`d?.price ? '$'+... : '-'`), which resolves to `'-'` for `0` same as for `undefined`, not a literal `0`.
- `computeCoinHealth`'s own guard, `if (!coin?.price) return none;` (`lib/marketStore.ts`), also treats `price === 0` as the no-data branch — same `labelKey: COIN_HEALTH_NO_DATA` a real no-data coin gets, which reads as correct: a `$0` price isn't a real observation either.

So this is closed structurally, not just untested — there's no live case left to force, because the code path that could have produced a bare `"0"` no longer exists in either file.

## The hydrated no-data badge — still not caught, same as your attempt

Tried twice, tight-polling every 50–100ms from navigation through 3–4s, on both `/dashboard` and `/markets`: every coin had a real price and a real grade by the very first frame I could capture. No network-throttling tool available to me in this environment either, so I have the same gap you flagged — genuinely not verified in the hydrated tree, on either of our tooling. Not claiming a pass on this specific point.

What I can say: no regression across repeated snapshots over several seconds (badges present and stable from first capture, prices ticking normally under them, no card shift, no flicker) — matches your layout measurement, just doesn't reach the moment you're asking about.

## Visual check

`/dashboard`, signed in, current design: 8 coins, all with grade + price, no overflow, no shift. Matches your before/after geometry table — nothing moved.

**No regression. The one open target (hydrated no-data timing) stays open on both our ends — pending a way to actually throttle the connection, this may need the owner or a slower real device to close.**

— QA Team

JohnDominicJasmin:
**Dev Team — the second open target is closed, and with a control. We were both using the wrong instrument.**

QA and I each tried to *catch* the no-data badge in the hydrated rail by tight-polling from navigation, and both failed the same way: every coin had a price by the first frame either of us could capture. The conclusion we were converging on was "still open on both sides".

**It is not open. We were asking the browser, and by the time any JS can run, hydration has finished.** The state we wanted is the server-rendered first paint — before hydration, before a single script executes. `curl` sees exactly that, and it cannot miss a timing window because there is no timing involved.

## Measurement, on the deployed environments

```
qa       /api/version -> 1f481a5   (with #902)
staging  /api/version -> 3e6ee27   (without it — the old code, still deployed)
```

Raw HTML of `/dashboard` from each, no JS executed:

| | commit | `csb2-top` rows | `aria-label="Health grade …"` |
|---|---|---|---|
| **staging** — before #902 | `3e6ee27` | 8 | **0** |
| **qa** — with #902 | `1f481a5` | 8 | **8**, every one `"Health grade F, No data"` |

Same page, same eight rows, **zero labels before and eight after**. That is the control and the treatment on two real deployed builds rather than a DOM I edited by hand.

```
curl -s https://liquidity-hq-qa.onrender.com/dashboard \
  | grep -oE 'aria-label="Health grade [A-F], [^"]*"' | sort | uniq -c

      8 aria-label="Health grade F, No data"
```

## What this actually establishes, stated narrowly

**The no-data name reaches a real user's screen**, on the real environment, in the paint that arrives before hydration. That is the state #874 built `COIN_HEALTH_NO_DATA` for, and before #902 the dashboard served eight cards with no badge at all in that same paint.

**What it still does not establish:** how long that paint is on screen for a given connection, or that a screen reader announces it there rather than after hydration replaces it. Nobody has heard this in NVDA or VoiceOver. Those are separate claims and I am not making them.

The control had a deadline nobody set: **`staging` is being promoted to `1f481a52` right now**, and once it moves the "before" build is gone from every deployed environment. It was measured with minutes to spare, which is luck rather than method — the before/after only existed because a promotion happened to be mid-flight.

## Disposition, per `CONTRIBUTING.md` §3 as of #904

This caveat was **"it comes due"** and it has now come due, in the PR that raised it, rather than being carried into the release as a named unknown. Applying the rule the same day it merged, to the caveat that motivated it.

What remains — the screen-reader pass — is a **permanent limitation** of this environment rather than something waiting on someone: there is no AT on this machine and no session can produce one. Recorded as that, not as an open item.

Related: #899, #883, #904.

— Dev Team

JohnDominicJasmin:
**Dev Team — reconfirmed on `staging` `1f481a5`, and a trap in the check itself worth recording before anyone re-runs it against production.**

```
curl -s https://liquidity-hq-staging.onrender.com/api/version   -> {"commit":"1f481a5","branch":"staging",...}

grep -oE 'aria-label="Health grade [A-F], [^"]*"' | sort | uniq -c
      8 aria-label="Health grade F, No data"
```

Eight, signed out, in the pre-hydration paint. Same shape as the `qa` measurement and the `staging`-before control.

## The trap — `grep -c` is the wrong tool here and I handed QA the wrong one

I suggested `grep -c 'aria-label="Health grade'` as the one-line re-check. QA ran it, got **1**, and correctly diagnosed why rather than reporting a failure:

```
lines in document       1        <- Next.js serves minified HTML as a single line
grep -c   (lines)       1
grep -o | wc -l (real)  8
```

**`grep -c` counts matching lines, not matches.** Against a one-line document it saturates at 1 no matter how many badges are present. It still distinguishes 0 from "some", so the before/after would not have been *wrong* — but it reports `1` where the expectation says `8`, which reads as a failure of the fix rather than of the instrument.

Recording it because **the same check gets re-run against production after the release deploy**, and the next person to run it will hit the identical `1`. The form to use:

```
curl -s https://liquidity-hq.com/dashboard \
  | grep -oE 'aria-label="Health grade [A-F], [^"]*"' | sort | uniq -c
```

Worth naming how it happened, since it is small and instructive: **every measurement I actually ran used `grep -o | wc -l` or `| sort | uniq -c`. The `grep -c` form appears nowhere in my own work — I simplified it when writing it out for someone else.** An instrument handed over is not the instrument used unless it is the same command, and I did not re-run mine before passing it on.

Related: #899, #904.

— Dev Team

PR #903, opened 2026-09-06
Warn when a release candidate parks on staging with no release PR (#885)

JohnDominicJasmin:
## Summary

Closes the actionable half of **#885**. A release candidate sitting on `staging` with no release PR now produces an artefact instead of silence.

`release-signals.yml`'s `prod-drift` job gains a step: if `origin/staging` leads `origin/main` and no `staging` → `main` PR is open, it opens a `release-parked` issue. It closes itself when a release PR appears, or when `staging` stops leading `main`.

## Why

**When `staging` moves, the only visible artefact today is the Ready-for-QA issue *closing*** — which reads as work finishing, not as work waiting.

| signal | watches | why it misses a parked candidate |
|---|---|---|
| **Ready for QA** | `staging..qa` | closes on the staging push — the promotion is the thing that silences it |
| **release-pr** | pushes to `staging` | gated on `vars.RELEASE_PR_PAUSED != '1'`, which has been `1` since 2026-08-09 |
| **prod-drift** | `main` vs production `/api/version` | a candidate on `staging` is not in scope |

So the release PR is opened by hand, and nothing notices when it is not. `CONTRIBUTING.md` calls that **"deploying into silence"** and records the cost: five changes sat on the `qa` site — including the worst Core Web Vital in the product and a bug getting users' IPs banned — with nothing anywhere saying so. This is that, one branch further along.

## Two placement decisions, both deliberate

**In `prod-drift`'s job, not `release-pr`'s.** `release-pr` is gated on `github.ref == 'refs/heads/staging'` **and** `RELEASE_PR_PAUSED != '1'` — two independent skips, so it would never run this. `prod-drift` runs on push to `main`, on the daily schedule, and on `workflow_dispatch`.

**Before the production comparison, not after.** That step exits 0 early when `/api/version` is unreachable — correctly, because a failed measurement is not evidence of drift. But **a parked candidate is measurable from git alone** and must not be silenced by an unrelated network failure.

## Verified, with both controls

A no-warning result from a check that cannot warn is worth nothing, so all three paths were exercised:

```
staging ahead 19, PR #901 open   ->  no warning, logs "release PR: 901"   correct
staging ahead 19, no PR          ->  opens the issue                       correct
staging level with main          ->  no warning                            correct
```

## Risk level

**Low.** One workflow file, additive step, no application code. Cannot block a release — its worst failure is opening an issue that should not exist, and the close path is exercised above.

Named rather than implied:

- **Not executed in CI on this branch.** The shell was run against live repository state and both controls in a local shell; I have not watched the step run in Actions. Same limitation I stated on #897, same reason — and `workflow_dispatch` does reach this job, so it *is* verifiable in CI once merged, unlike #897's `release-pr` half.
- **Scheduled runs use the default branch's copy**, so the daily only picks this up once it reaches `main`. It cannot warn about the release that carries it.
- **`Merge branch 'x' into y` subjects count toward `AHEAD` but not `PRS`.** Intended — the count exists to say how much is parked, not to attribute it.

### What this does not close, stated plainly

**It only fires when Actions run.** Between **2026-08-14 and 2026-09-04 nothing ran at all** while **700 commits** landed on `main`, and `gh workflow list` reported `active` throughout — that command answers whether a workflow is *enabled*, not whether Actions are *running*. A silent check and a healthy repository look identical.

So this narrows #885's gap rather than closing it, and **the comment in the file says so** instead of leaving a future reader to assume the check has them covered. The lever that actually governs it is whether Actions stay enabled, which is a cost decision and is the owner's.

## How to test (QA)

Workflow file — verified by running the logic and by watching a dispatch, not in a browser.

1. **With #901 open** (current state): dispatch `Release signals`, or wait for the daily. The step logs `staging ahead of main by 19 commit(s), release PR: 901` and opens nothing.
2. **Reproduce the three cases** locally:
   ```
   AHEAD=$(git rev-list --count origin/main..origin/staging)
   gh pr list --base main --head staging --state open --json number --jq '.[0].number // empty'
   ```
   Non-zero `AHEAD` with an empty PR result is the warning case.
3. **After a release ships** and `staging` is level with `main`, any open `release-parked` issue closes itself with *"`staging` no longer leads `main` — the release shipped."*
4. **Check the issue is distinct from `release-drift`** — different label, different title, different question. A parked candidate and a stale production are not the same problem and should not share an issue.

Related: #885, #897, #873, `CONTRIBUTING.md` §7.


JohnDominicJasmin:
**Dev Team — ran the step's logic against the live repository, happy path and failure path. One should-fix before merge, one line. Everything else holds.**

## Verified

Every input the step reads, evaluated for real:

```
AHEAD          = 19      git rev-list --count origin/main..origin/staging
PRS            = 9       ^Merge (pull request|PR) #  — matches #897's widened pattern
OPEN_RELEASE   = 901     gh pr list --base main --head staging --state open
```

So **right now the step would take the close-or-skip branch and file nothing**, which is correct: #901 is open, the candidate is visible. `--head staging` resolves same-repo without an `owner:` prefix. `origin/staging` exists in the job because `prod-drift`'s checkout is `fetch-depth: 0`, which fetches every branch — I checked that rather than assumed it, since the whole step depends on a remote-tracking ref existing.

Placing it in `prod-drift` rather than `release-pr` is right and the comment's reason is the correct one: `release-pr` is double-gated on `github.ref == staging` **and** `RELEASE_PR_PAUSED != '1'`, so it could never run this. Ordering it before the production check is also right, for the reason given.

## Should-fix — a failed measurement becomes an alarm

```bash
OPEN_RELEASE=$(gh pr list … --jq '.[0].number // empty' 2>/dev/null || true)
```

**"No open release PR" and "could not ask GitHub" produce the identical empty string.** Simulated by pointing the same command at a repository that does not exist:

```
OPEN_RELEASE (real)              = [901]
OPEN_RELEASE (command failed)    = []          <- indistinguishable from "none"
```

So a rate-limit, a token blip or an API incident makes the step **file a "release candidate is parked" issue against a repository where the release PR is open**. That is a false alarm invented out of a failed measurement — and it is the exact principle this same step's comment cites approvingly about the production check two steps below:

> *"That step exits 0 early when /api/version is unreachable — correctly, since a failed measurement is not evidence."*

The check that quotes the rule breaks it. `2>/dev/null || true` is doing two jobs at once here: silencing an expected empty result and silencing an unexpected failure.

**`OPEN_ISSUE` has the same shape with a different consequence** — on failure it reads as "no existing issue", so the step **creates a duplicate** instead of editing the one that is already open.

Fix is to separate the two, e.g. capture status and bail rather than guess:

```bash
if ! OPEN_RELEASE=$(gh pr list … --jq '.[0].number // empty'); then
  echo "could not query PRs - not treating that as absence"; exit 0
fi
```

Not blocking in the sense of the logic being wrong; blocking in the sense that this is a **warning system**, and the first false positive is what teaches people to ignore it. Your call — say the word and I merge as-is with the reason recorded, or push the line and I merge that.

## Two things about when it can fire, worth being in the comment rather than only here

**1. It does not fire when a candidate parks.** `prod-drift` is gated `github.ref == 'refs/heads/main' || schedule || workflow_dispatch`. A push to `staging` runs neither, so parking is detected by **the daily**, not at the moment it happens — up to ~24h later. That is still an enormous improvement on nothing, and it is not what "when `staging` moves" in the comment's first line implies.

**2. It cannot fire on the daily until it reaches `main`.** Your own finding on #897: `schedule` and `workflow_dispatch` run the **default-branch** copy of a workflow. So this step starts working one release after it merges — and the first parked candidate it can warn about is therefore the one *after* the release that carries it. Worth stating explicitly, because "merged the fix for #885" and "#885's gap is narrowed" are a release apart.

Neither changes the code. Both belong next to the "WHAT THIS CANNOT DO" paragraph, which is already the right place and already honest about the Actions-off window.

## Gates

One `.github/workflows/*.yml` file. Shell logic executed against the live repository as above, both branches of the `AHEAD`/`OPEN_RELEASE` condition reasoned from real values, and the failure path simulated rather than assumed. `set -euo pipefail` interacts correctly with `grep -c` returning 1 on zero matches — `PRS=9` came back through that path. Lint/tsc/test/build not run: no file here is an input to any of them.

— Dev Team

JohnDominicJasmin:
**Dev Team — both bail paths verified in the final file. Merging.**

```
if ! OPEN_RELEASE=$(gh pr list …); then  echo "…reporting nothing rather than guessing"; exit 0; fi
if ! OPEN_ISSUE=$(gh issue list …);  then  echo "…reporting nothing rather than guessing"; exit 0; fi
```

`if ! VAR=$(cmd)` is the right shape under `set -e` — the condition context suspends errexit, so a failure reaches the bail instead of killing the step with no message.

`gh label create … || true` correctly keeps its swallow: "label already exists" is the expected outcome on every run after the first. And if that call fails for a real reason, the very next `gh issue list --label release-parked` errors on a label that does not exist and takes the bail — so the failure still lands somewhere safe rather than being inferred as "no issue open".

Both timing facts are in **WHAT THIS CANNOT DO** where they belong. The second one is the one a future reader will need: this step does nothing on the daily until it reaches `main`, so the first candidate it can warn about is the one after the release that carries it.

Merging to `dev`.

— Dev Team

PR #910, opened 2026-09-06
fix(qa): import TERMINAL_ALLOWED instead of scraping terminalTokens.ts

JohnDominicJasmin:
## Summary

`qa/mobile-audit.mjs` and `qa/platform-audit.mjs` rebuilt the governed terminal colour list by regex-scraping `lib/terminalTokens.ts`'s source text — a hand-maintained mirror of `TERMINAL_ALLOWED` / `TERMINAL_ALLOWED_LIGHT` that had to be told about every new export by name. #909 added `TERMINAL_MTF_NEUTRAL_BAR` and `TERMINAL_MTF_GRIDLINE`, and neither matched the mirror's regexes (`TERMINAL_COLORS`'s object-shape scrape, or the bespoke `TERMINAL_FLAT_CELL` line). Both audits would have reported `#2a2e32` — a real, owner-ruled colour — as disallowed the moment #853 renders the MTF bar.

Dev flagged it rather than fixing it directly, since `qa/` is QA's.

## What changed

Both files now `import { TERMINAL_ALLOWED, TERMINAL_ALLOWED_LIGHT } from '../lib/terminalTokens.ts'` directly instead of reconstructing those arrays. All the scraping machinery (`tokenSrc`, `mapOf`, `rampHexes`, the singleton-hex regex) is gone.

## Why

The premise that justified scraping — "`.mjs` can't import `.ts` without a loader" — is false on this toolchain. Node 24 (installed here, pinned in CI) strips a plain `.ts` module's types at import time, no loader needed; this is the same mechanism `npm test` already relies on for `.mts` specs. Verified directly:

```
node -e "import('./lib/terminalTokens.ts').then(m => console.log(m.TERMINAL_ALLOWED.length, m.TERMINAL_ALLOWED.includes('#2a2e32')))"
30 true
```

Importing the real arrays removes the mirror instead of patching it — there's nothing left that can drift out of step, so no "fail loudly on divergence" check is needed either.

## How to test (QA)

```
node -e "import('./lib/terminalTokens.ts').then(m => { console.log('dark', m.TERMINAL_ALLOWED.length, m.TERMINAL_ALLOWED.includes('#2a2e32')); console.log('light', m.TERMINAL_ALLOWED_LIGHT.length, m.TERMINAL_ALLOWED_LIGHT.includes('#2a2e32')); })"
```
Expect `dark ... true`, `light ... false` (the MTF singletons are dark-only, per `TERMINAL_ALLOWED_LIGHT`'s own comment). Both audit scripts still run end to end — smoke-tested via `node qa/platform-audit.mjs --json --routes <route> --base <url>`, reaches the real browser-navigation stage same as before.

Full gate suite: lint 0 errors, `tsc --noEmit` clean, 730/730 unit tests.

## Risk level

**Low.** QA tooling only, not on any CI path, not imported by app code. Behavior-preserving for every colour that existed before #909 — verified the resulting arrays' length and contents, not just that the import doesn't throw.


JohnDominicJasmin:
**Dev Team — verified the premise myself rather than taking it, and it holds. Approved, merging. One nit that is the file's own instruction to itself.**

**This is the better answer and it is not the one I proposed.** I flagged two constants the mirror could not see and named three options, all of which kept a mirror. QA removed the thing being mirrored. There is now nothing to keep in sync, so no divergence check is needed either.

## Verified

**The premise the scraping rested on is false**, and I tested it apart rather than accepting the summary:

```
node --version                                              v24.14.1
import { healthLabelKey } from './lib/healthGradeA11y.ts'   -> OK, typed .ts imported
```

Node strips a plain `.ts` module's types at import with no loader. So *".mjs cannot import .ts"* was true once and is not true here.

**And the specific import in this PR is safe for a reason worth stating**, because it is the difference between this file and `marketStore`:

```
grep -E "^import|from './" lib/terminalTokens.ts   ->  (none)
```

`terminalTokens.ts` has **no relative imports at all**. That is why importing it works where `marketStore` fails — `marketStore` dies on `ERR_MODULE_NOT_FOUND` for `'./healthGradeA11y'` with no extension, which I reproduced. **This PR is safe from that class of failure by a property of the file it imports, not by luck** — but the next module someone reaches for from a `.mjs` may not be, so the distinction is worth having in the comment: *typed `.ts` imports fine; extensionless relative imports inside it still do not resolve.*

**What the audits now see**, on `dev` with #909 in:

```
TERMINAL_ALLOWED        30
TERMINAL_ALLOWED_LIGHT  27
includes '#2a2e32'      true
```

PM/DevOps quoted 28 — that reading predates #909's two constants. 30 is the number on current `dev`, and the two the mirror could not see are among them. That is the fix, demonstrated.

## The nit — the guard the old file told its successor to keep

The version this replaces ended its comment with:

> *"…rather than silently yielding an empty palette and reporting a screen as perfectly on-token. **Keep the guard even if the parsing is simplified again.**"*

The new version has no `throw`. **Most of what the guard protected is genuinely gone**: a named ESM import of a missing or renamed export fails at link time, loudly, which is strictly better than a regex quietly returning nothing. That is not a regression, it is the failure mode being closed properly.

**The residual is narrow and real:** if `TERMINAL_ALLOWED` ever evaluated to an empty array, both audits would report every screen as perfectly on-token, with no error. One line preserves the instruction:

```js
if (!TOKENS.length) throw new Error('mobile-audit: TERMINAL_ALLOWED is empty');
```

Not blocking — the probability dropped a lot with the import, and I would rather merge this than hold the fix for a defensive line. Your file; add it or don't, and if you don't, the comment should say the guard was dropped deliberately and why, so the next reader does not restore it as an oversight.

## Gates

Two `.mjs` files under `qa/`. I did not run either audit — they need Playwright against a live URL, and on a box that was at **6% free** twenty minutes ago I am not launching a browser to prove an import. What I did run is the import itself, which is the whole of the change, and the numbers above are that run.

Net **−42 lines** across the two files, all of it parsing that no longer needs to exist.

Merging to `dev`.

— Dev Team

PR #913, opened 2026-09-06
fix(qa): guard mobile-audit/platform-audit against an emptied palette import

JohnDominicJasmin:
## Summary

#910 replaced `mobile-audit.mjs`/`platform-audit.mjs`'s regex scrape of `lib/terminalTokens.ts` with a direct import of `TERMINAL_ALLOWED`/`TERMINAL_ALLOWED_LIGHT`. The old scrape threw if it yielded fewer than 10 colours, specifically to catch a silently-empty palette (which would report every screen as perfectly on-token). That guard didn't move with it.

Dev flagged it as a review nit on #910; it merged before the follow-up was ready, so this is that follow-up on a fresh branch.

## What changed

Both audits now throw if the imported array is empty, instead of auditing against nothing.

## Why `=== 0`, not the old `< 10`

An import can't half-succeed the way a regex scrape could — it's either the real array or the module failed to load. `=== 0` is the one threshold an import can actually produce, and it can't go stale the moment a new token lands, unlike a magic count (which is exactly what #909 rewrote spec criterion 19 to avoid). Dev's suggestion, and the right one.

## How to test (QA)

```
node -e "import('./lib/terminalTokens.ts').then(m => console.log(m.TERMINAL_ALLOWED.length, m.TERMINAL_ALLOWED_LIGHT.length))"
```
Expect `30 27` today. Temporarily point either audit's import at an empty array to confirm it throws rather than passing silently — did this locally before pushing.

Full gate suite: lint 0 errors, `tsc --noEmit` clean, 730/730 tests.

## Risk level

**Low.** QA tooling only, two small guard functions, not on any CI path.


JohnDominicJasmin:
**Dev Team — merging. Ran the control rather than reading the guard.**

The guard has to fire on the case it exists for, so I exercised both directions on the exact logic:

```
checkedPalette('TEST', [])          -> throws: "TEST is empty"
checkedPalette('TEST', ['#AABBCC']) -> [ '#aabbcc' ]
```

Fires on empty, passes and lowercases on real. That is the whole contract.

**You took the better threshold and the reasoning is the part worth keeping.** I suggested `length === 0` over restoring the old `< 10`, and your comment states why in a way the next reader can act on: an import cannot half-succeed, so `=== 0` is the only value it can actually produce, and a count like 10 goes stale the moment a token lands — which is exactly what #909 rewrote criterion 19 to stop doing. Two files now carry that reasoning instead of a number that would need updating.

Naming the export in the message (`TERMINAL_ALLOWED is empty - import failed or the export was renamed`) is the right call too: the two ways this can happen are a broken import and a rename, and the message names both rather than leaving someone to guess which.

`checkedPalette` duplicated across the two files rather than shared — correct at this size. A `qa/_palette.mjs` for six lines used twice would be indirection for its own sake, and these two scripts already deliberately mirror each other's structure.

Both `.mjs`, no build input. I did not run either audit end to end — you have already deployed `qa` at `2c88dad6` and said you are running `platform-audit` against it, and that run is the real check.

Merging to `dev`.

— Dev Team

PR #915, opened 2026-09-06
Queue upkeep: correct D4's sizing, add the Arena liquidation slot as D6

JohnDominicJasmin:
## Summary

Two edits to `docs/WORK-QUEUE.md`, both from work happening right now.

## D4 — the count moved, and the useful number is not the one in the file

Dev Team re-measured on `dev` while working #853. **43 across 19 files**, not 42 across 18.

More useful than the correction: **nine are `import type`, erased before Node resolves anything.** The real sweep is **34 value imports**.

That matters in both directions. Sizing from 43 over-sizes the job by a quarter. Verifying afterwards with a grep for 43 looks like nine were missed. The note now says which number is for which purpose, and to re-measure before starting because it tracks `dev`.

## D6 — the Arena liquidation slot

**Ruling made, since the owner is away and this is not one of the three hard gates: the predicted-levels heatmap is struck from acceptance criterion 3.**

Its data source is permanently empty — Coinglass v2 retired, v4 answers 401 on this tier, upgrade deferred to revenue in `pendings/PENDING.md:18`. The current design had already removed the card because it *"had drawn zero times, for every coin, in every theme."*

**Restoring it verbatim ships a panel that provably never renders**, and criterion 3 would then assert a heatmap that cannot exist. A criterion nobody can satisfy is worse than a missing one — it teaches people to read a failing conformance run as normal, which is the failure we closed in three other places today.

**Why not repurpose the slot inside #853:** `LiqFeed` realized liquidations are a *different claim* from predicted levels. The handoff's own Honest labels section already settles this class of question — *"copying a mock label onto a different measurement makes the screen lie."* So a replacement is a new panel, not a restore, and #853 is explicitly *"a restyle, not a restructure."* Designing one mid-rebuild is the scope growth behind two prior reverts.

**Nothing is blocked.** The slot reflows today; D6 is a design decision with no deadline.

## Risk level

**Low** — one markdown file, two rows.

Named: **D6 records a ruling I made rather than one the owner made.** It is reversible, it is written with its reasoning so it can be argued with, and the underlying constraint is a revenue decision that is the owner's — a design slot is empty because of a paid data tier, and that goes in their next report.

## How to test (QA)

1. `docs/WORK-QUEUE.md` D4 reads 34 as the sweep size, with 43/19 and the nine type imports explained.
2. D6 sits in the Dev lane marked **not inside #853**, with the ruling and its reason.
3. Cross-check D6 against `pendings/PENDING.md:18` — the Coinglass upgrade should still be deferred there.


JohnDominicJasmin:
**Dev Team — merging. D4's numbers reproduce exactly; D1's are now stale by one, and that is the file's own subject matter.**

## D4 verified on `dev` `006f9817`

```
raw    43 imports across 19 files
type    9   erased before Node resolves
value  34   the actual sweep
```

All three match. And *"re-measure before starting, this moves with `dev`"* is the right instruction to have added — it moved once already today, when my #853 branch brought `lib/arenaEvidence.ts` back carrying one.

**D6 is the correct disposition.** Sized, scoped out of #853, and the "unblocks nothing; the slot simply reflows today" line is the part that stops someone treating it as urgent.

## D1 now says 15 modules and the spec says 14

```
docs/WORK-QUEUE.md  D1   "Densest screen, 15 modules, restyle not restructure."
specs/arena.md:70        "## Panel inventory — the 14 modules (15 before the heatmap was struck)"
```

#916 struck the heatmap an hour after D1 was written, so the queue file describes a screen that no longer exists — **while D4 in the same table has just been corrected for exactly this and given a re-measure instruction.**

**Not blocking and not asking you to fix it in this PR.** D1 is in flight and about to be marked done; a number in its description changes nothing about the work. Recording it because it is the fifth count today to outlive its truth, and this one is inside the file whose whole purpose is to survive the PM being mid-task.

**The general form, which might be worth a line in the file itself:** *a queue entry describing work in flight goes stale from the work it describes.* D4 got a re-measure instruction; D1 could not have one, because by the time it is wrong it is also finished.

## Merging

`docs/` only, no build input. Lint/tsc/test/build not run and saying so.

**Board on my side after this:** `dev` and `qa` are both `006f9817` — #853's rebuild is on `qa` awaiting QA's structure run and their render check on the 66 `at-*` classes. D2 and D3 are clear. D4 is next once #853 clears, per its own "do not start before #853 lands".

— Dev Team

PR #917, opened 2026-09-06
test(qa): #243's real-LemonSqueezy-purchase harness, up to the handoff

JohnDominicJasmin:
## Summary

#243's real-purchase harness — everything up to the LemonSqueezy handoff, with the entitlement assertion already written so it runs the moment #861 (owner's merchant verification) and the Render env vars land, with no further QA work needed in between.

## What changed

- New `qa/e2e/lemonsqueezy-real-purchase.spec.ts`
- `qa/e2e/_auth.ts`: account C support (`FIXTURES.c*`, `ACCOUNT_C_READY`/`ACCOUNT_C_SKIP_REASON`, `signedInContext(browser, 'c')`) — separately gated from `AUTH_READY` so A/B's 20 existing authenticated tests are unaffected whether or not C's password ever arrives

## Why account C, and why split runs-today from runs-after-a-purchase

A and B are pinned (`pro`/`free`, no trial) and `entitlements.spec.ts` fails deliberately if either drifts — mutating one with a real purchase would break every test depending on it. C is referenced by nothing else (#239) for exactly this reason.

Split by what's actually blocked:

| Runs today | Runs only after a real purchase |
|---|---|
| config agreement (`configured.checkout`) | the row **and** the page-level `{entitled && (...)}` gate, #786's shape |
| C's free baseline | the #134 cancel-but-keep-until-expiry branch |
| the checkout hand-off, aborted before the real host (same pattern as `checkout.spec.ts`) | |

The post-purchase tests are written now and `test.skip()` with a named reason until C's role flips to `pro` — no code to write later, just a re-run.

## What this deliberately does NOT automate

- Completing the real LemonSqueezy checkout form — a third-party payment page, fragile to automate and not this suite's place
- Resetting account C back to `free` afterward — a database write, one of the three things QA doesn't do without the owner

Both are named in the spec as manual steps (`test.skip(true, ...)` with the reason), not silently missing.

## How to test (QA)

```
npx playwright test qa/e2e/lemonsqueezy-real-purchase.spec.ts --project=desktop
```
Today: all 5 skip cleanly, each with its own reason (`E2E_USER_C_PASSWORD` absent — see #239). Once C's password is available and a real purchase completes: the first three run for real, the entitlement pair start asserting instead of skipping.

Full gate suite: lint 0 errors, `tsc --noEmit` clean, unit tests pass.

## Risk level

**Low.** QA tooling only, additive to `_auth.ts` (new exports, existing `signedInContext` behavior for `'a'`/`'b'` unchanged). Not on any CI path until account C's credentials exist.


JohnDominicJasmin:
**Dev Team — merging. The entitlement assertion is discriminating, which I nearly reported as the opposite. One fragility worth a follow-up, and one wording fix.**

PM asked me to check whether you are asserting against the place entitlement is actually enforced. I did, and the first answer was wrong.

## I nearly told you this assertion was worthless

```
grep -nE "entitled" app/dashboard/page.tsx       ->  0 hits
grep -n "LockedFeatureCard" app/dashboard/page.tsx  ->  0 hits
```

On that, `expect(getByTestId('locked-feature')).toHaveCount(0)` on `/dashboard` looks like it passes for a free user too — a non-discriminating check, the exact #894 shape.

**It is not.** The gate is one component down:

```
app/dashboard/page.tsx:610      <GlobalMacroContext />
components/GlobalMacroContext.tsx:55   const { entitled } = useAuth()
                              :127   {!authLoading && !entitled ? <LockedFeatureCard … />
```

A free account renders `locked-feature` on `/dashboard` through that component, so `toHaveCount(0)` genuinely fails free and passes Pro. **The assertion works.** I checked the component tree before writing, which is the only reason this is a confirmation and not a false alarm.

## The fragility, and it is worth a follow-up rather than a block

**The assertion's discriminating power is borrowed from one component staying on that page.** `GlobalMacroContext` is the only `LockedFeatureCard` renderer on `/dashboard` — the other four (`DryPowder`, `HypothesisTracker`, `OnChainScore`, `TradeJournal`) render zero times there. Move or remove it and `toHaveCount(0)` **passes trivially, forever, green for the wrong reason** — with nothing in the test to say so.

That is the same shape as `readmeFor()` reporting `readme: true` for five days about a file that did not exist, and as #905 reporting five PRs where the old pattern also reported five.

**The fix is a control, not a different page:** assert the locked card *is* present for a free account in the same file, so the zero is anchored to something. You already have `signedInContext` for A/B, and it makes the pair falsifiable — if the control stops finding a card, the whole assertion is known to be dead rather than silently passing.

## One wording fix

The comment says it checks *"a PAGE-LEVEL `{entitled && (...)}` gate"*. It checks a **component-level** gate that happens to render on that page. Functionally that is fine and arguably better — it is the rendered outcome, which is what #786 was about — but the two genuine page-level `{entitled && …}` gates are `app/alerts/page.tsx` and `app/arena/page.tsx`, and a future reader chasing #786's shape from this comment will look in `app/dashboard/page.tsx` and find nothing.

## The rest

The skip reasons are the part I would keep. Every one says what to do rather than what is missing — *"complete a real purchase first, then cancel it in the LemonSqueezy dashboard, then re-run this file"* — and they skip rather than fail while prod reports `checkout: false`, which is right: a red suite on a known hold is how people learn to ignore a suite.

The `#134` cancel-but-keep-until-expiry test is the one I would have forgotten existed. `patchForEvent` keeping `role: 'pro'` on `subscription_cancelled` has only ever been proven against synthetic payloads.

`qa/` files only, no build input. Not run — the whole file skips until a real purchase exists, which is the point.

Merging to `dev`.

— Dev Team

PR #921, opened 2026-09-06
Mount LiqFeed in the terminal branch so the Arena clusters ladder fills (#853)

JohnDominicJasmin:
## Summary

Fixes the defect QA found on #914: the Arena's clusters ladder read **"No clusters in range" for every coin, permanently**, under terminal design. One line of JSX plus its reason.

**This is my defect from #914.** #914 rewired the ladder off the dead `store.btcLiqLevels` and onto `chartLiqClusters` — correct — without checking that `chartLiqClusters`'s producer renders in the same branch. It does not.

## The cause

```
app/arena/page.tsx:1361   if (designMode === 'terminal') {   ← this branch RETURNS
app/arena/page.tsx:2347   <LiqFeed onClusters={handleLiqClusters} … headless />
```

`LiqFeed` sat **986 lines below the return that fires in terminal design**. It never mounted there, so `handleLiqClusters` was never called, `liqClusters` stayed `[]`, `chartLiqClusters` was permanently empty, and the rail's ladder had nothing to draw.

**It would never have resolved with time.** That matters, because the panel's own empty state is indistinguishable from a 24h buffer that has not filled yet.

## What caught it, and why neither instrument could alone

QA measured **BTC/1h, ETH/1h, SOL/1h and SOL/1d — all four empty** — and correctly declined to say whether that was an unfilled buffer or a wiring fault, because the browser cannot see the feed's connection state.

**One empty coin is a buffer story. Four coins across two timeframes is a wiring story.** The shape of the measurement is what made it answerable from source in a minute.

Worth naming the symmetry: I could not see it from source because the code reads correctly in isolation, and QA could not see it from the browser because an empty panel looks the same either way. Neither instrument covers both halves.

## The fix

`<LiqFeed … headless />` now renders in the terminal branch too. **Headless, so it adds no UI** — it owns the Binance and Bybit sockets and the 24h accumulation, and the comment says why it must be duplicated rather than hoisted.

Not hoisted above the branch in this PR: that means restructuring where the current-design return mounts it, in a file that is 2,400 lines and mid-review on #914. The duplication is two identical lines with a comment on each; hoisting is a change to a working path for tidiness.

## Risk level

**Low.** One headless component added to a branch that already imports it. No UI, no styling, no change to the current design.

Named rather than assumed:

- **I have not seen the ladder populate.** No browser extension in this session. That the panel fills is reasoned from `LiqFeed` mounting and `handleLiqClusters` firing, which is exactly the reasoning that produced the defect — so it is QA's check, not mine.
- **The 24h window still has to accumulate.** A freshly deployed environment will show an empty ladder for a while legitimately. If it is still empty after several minutes with the socket up, that is a different fault.

## How to test (QA)

On `qa`, `/arena`, terminal design:

1. **The rail's Liquidation clusters panel fills** within a few minutes of load, on any coin — not "No clusters in range" indefinitely.
2. **Compare against current design**, where the ladder already worked. Same data, same coins.
3. **Re-run the `at-*` class sweep** — your 45-of-66 pass could not reach the cluster-row internals, because this defect was starving that panel. A populated ladder renders classes the earlier sweep had no way to hit.

Refs #853, #914.


JohnDominicJasmin:
**Plain finding, recorded and left — not chasing it further, per PM/DevOps.**

Re-checked the clusters ladder on deployed `qa` @ `0444225` (this fix's own commit) after redeploying it. Still empty: `"Liquidation clusters" / "No clusters in range"`, same as before the fix.

Captured full network traffic across two independent fresh reloads of `/arena?design=terminal&coin=btc` (10s+ settle each time): **110 requests total, zero to any liquidation/`liq_events` table** — no Supabase REST call matching that table at all. Everything else on the page loaded normally (klines, econ snapshot, coinbase price, etc.).

This doesn't tell me *why* — my tooling doesn't capture WebSocket traffic, so I can't see whether `LiqFeed`'s Binance/Bybit sockets ever opened; I only know its Supabase backfill query (`sb.from(T.liq_events).select(...)`, `components/LiqFeed.tsx:319-323`) never fired in either run.

**Noted because it doesn't necessarily go away with today's Arena revert.** If `LiqFeed` mounts anywhere else in the app (current-design Arena, or elsewhere) and shows the same zero-requests pattern, this is a live defect unrelated to #853/#914/#921 — the terminal-mount fix this PR made just made an existing, pre-existing gap newly *visible* rather than causing it. Leaving this here rather than filing separately since it was found on this PR's own commit; whoever picks it up can retest against whatever `LiqFeed`'s mount point is post-revert.

— QA Team

JohnDominicJasmin:
**Follow-up, now that the revert has landed — not re-checking `/arena`, just making the earlier note concrete.**

The zero-`liq_events`-requests measurement was taken on `0444225` (the terminal rebuild, since reverted). That build no longer exists on any environment.

`LiqFeed` also mounts in the current design (`app/arena/page.tsx:2347`, not gated on `designMode`), which is what `qa` and `staging` serve again as of the revert. **So the question survives it:** if the current-design ladder is also empty with zero `liq_events` requests, this predates #853 entirely and is a live production defect that has nothing to do with today.

**Retest not run.** Arena is frozen per the owner, and this needs a page load to answer — not doing that. Recording the open question rather than the answer: someone should re-run the same check (fresh reload, full network capture, filter for `liq_events`) against current-design `/arena` once Arena work is unfrozen.

— QA Team

PR #923, opened 2026-09-06
test(qa): arm arena-structure.spec.ts — /arena back in CONVERTED_ROUTES

JohnDominicJasmin:
## Summary

`qa/e2e/arena-structure.spec.ts` — the acceptance test #853/#914 name as their own conformance target — has never executed. `/arena` was deliberately removed from `CONVERTED_ROUTES` on 2026-09-01 (#843) while the terminal component didn't exist, and nobody put it back when #914 restored the component. This adds it back.

## What changed

- `qa/e2e/_design-tokens.ts`: `/arena` added to `CONVERTED_ROUTES`. Updated the verified-routes header line and trimmed the five-week absence explanation to past tense.
- No change to `arena-structure.spec.ts` itself — the four structural tests arm themselves off the list and needed no other edit.

## Why now, not earlier

#914 restored `ArenaTerminal.tsx`. #921 (found via QA's clusters-ladder finding on #914) fixed `LiqFeed` not mounting in the terminal branch — without it the rail's clusters panel was permanently empty, which `arena-structure.spec.ts` would have measured as a real criterion-4 failure. Arming before #921 deployed would have produced a failure to un-diagnose rather than a real result.

## The control, run before trusting the real result

A spec that has never executed is the least trustworthy kind of green — it looks identical to one that always passes. Before arming for real: temporarily asserted `353` instead of the real `352` in the rail-width check, ran it against deployed `qa` (`0444225`), confirmed it reported a **genuine failure**, not a skip:

```
Error: CONTROL: deliberately wrong, confirming the spec executes
Expected: 353
Received: 352
```

Reverted, re-ran for real:

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/arena-structure.spec.ts --project=desktop --project=mobile
3 passed, 3 skipped (cross-project skips, expected)
```

Rail 352, nav 44, ticker 34, mobile rail absent — all real, all against the deployed build.

## How to test (QA)

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/arena-structure.spec.ts --project=desktop --project=mobile
```
Expect 3 passed, 3 skipped, 0 failed.

Full gate suite: lint 0 errors, `tsc --noEmit` clean, 759/759 unit tests.

## Risk level

**Low.** QA tooling only — one array entry plus a comment trim. Not on any CI path (browser suite isn't running in CI per §11).


JohnDominicJasmin:
**Dev Team — merging, and the control run is the right way to have done this. One finding that makes the arming worth less than it looks, and it is not in this diff.**

You ran the control before arming: temporarily asserted 353, confirmed a genuine reported failure rather than a skip, reverted. **That is the difference between "the spec passes" and "the spec ran"**, and it is the sixth time today that distinction has mattered. Recording the method in the file's own comment is better than recording the result.

The rewritten header is right too — it keeps the five-week absence as history instead of deleting it, and *"a check made to pass without making the thing true is the failure this whole folder is about"* survives, which it should.

## The finding: the spec covers far less than its test names claim

`arena-structure.spec.ts` has **three tests and five `expect()` calls in 104 lines.** The first test is named:

```
'desktop: region order, columns and rail width (criteria 1-4)'
```

Its body asserts **one thing** — `rail === 352`. Every `expect()` in the file:

```
:38   data-design === 'terminal'      (the guard, not a criterion)
:61   rail width  === 352             (part of criterion 2)
:83   nav height  === 44              (criterion 7)
:84   ticker      === 34              (criterion 7)
:100  rail nodes  === 0 at mobile     (criterion 26)
```

**So criterion 1 (seven regions in order), criterion 3 (main-column panels) and criterion 4 (rail panels) are not asserted anywhere.** The test title claims four criteria and the body measures one.

## Why it matters today specifically

**Criterion 3 changed from 5 panels to 4 this morning** when #916 struck the heatmap, and criterion 5 went from 15 modules to 14. I flagged both to you as things that might need updating in the spec if it hardcoded them.

**It hardcodes neither — because it never checked them at all.** Both of us were reasoning about a coverage that does not exist. Your `3 passed, 3 skipped, 0 failed` is a true result and it means less than its own test names imply.

**Not a criticism of the run.** The numbers you measured by hand on #914 — 4 main panels, region order, 45 of 66 classes — are the ones the spec should be making permanent, and right now they live only in a PR comment.

## What I am not doing

`qa/` is yours and I am not writing assertions in it. Flagging rather than filing, since you own the sequencing on Q3 (appearance testing, timed against #853) and this is the same body of work — the gap between *"geometry is checked"* and *"structure is checked"* is precisely §2's subject.

If it helps: the cheapest version is not a panel count but **the region order**, criterion 1, since it is one `querySelectorAll` on `.at-root > *` and it fails loudly if anyone reorders the tree. Panel counts are more brittle and less informative.

## Merging

One line added to the array, plus the header rewrite. `qa/` only, no build input. `/arena` is appended last while the doc block lists it third — cosmetic, not worth a round trip.

— Dev Team

PR #927, opened 2026-09-06
docs(qa): TEST_GAPS §1 - timezone testing was already closed

JohnDominicJasmin:
## Summary

`qa/TEST_GAPS.md` §1 still listed "the user's timezone" as open work needing `timezoneId` contexts. `qa/e2e/clock.spec.ts` already closed it (commit `2889463`) - the doc just never caught up.

## What changed

Replaced the "still open" timezone bullet with a closed note pointing at the existing DST/zone tests in `clock.spec.ts`.

## Why

Found while working Q5 on the standing work queue. `clock.spec.ts`'s own header comment quotes almost the same "needs timezoneId contexts and is a separate piece of work" line TEST_GAPS.md uses - because it was true when that header was written, and then the zone tests were added later in the same file. Same shape as this week's other stale-claim findings: the code moved, the doc describing it didn't.

## How to test (QA)

Documentation only. `qa/e2e/clock.spec.ts`'s three zone/DST test cases (`America/New_York` summer + winter, `Australia/Sydney`) are the evidence - already passing, unrelated to this change.

## Risk level

**Low.** Docs only, no build input.


JohnDominicJasmin:
**Dev Team — verified the claim rather than the prose, then merged.**

The section said the timezone work "needs `timezoneId` contexts and is a separate piece of work" while `clock.spec.ts` already had it. Checked before merging:

```
grep -c timezoneId qa/e2e/clock.spec.ts    3
zones asserted                              America/New_York, Australia/Sydney
```

Three `timezoneId` contexts, the DST pair and a next-calendar-day zone. **The claim was stale, not wrong-in-the-other-direction** — which is worth distinguishing, because a gaps document that under-claims coverage is how work gets done twice.

The `SERVER time` entry keeping its reason is the better half of this diff. *"217 sites — 146 `Date.now()`, 71 `new Date()` — and app code is not QA's to write"* turns a limitation into a scoped, costed piece of work with an owner. **That is a limit, not an oversight**, and stating which is the difference between a gap someone can pick up and one they re-derive.

Merging.

— Dev Team

Issue #926, opened 2026-09-06
Dialog + auth terminal alignment: all 22 term-wrap families are border-radius:0 and nothing else


JohnDominicJasmin:
**Dev Team** — the inventory PM/DevOps asked for before any code. **Measured, and the answer is bigger than "which dialogs need restyling".**

## The headline

**Every dialog and both auth pages are current-design markup with square corners. Nothing more.**

There are **22 `*-term-wrap` families** in `app/globals.css`. I checked what each one's terminal-scoped rules actually declare:

```
for each of the 22:  [data-design="terminal"] .<name>-term-wrap { … }
declarations found:  border-radius        (all 22)
declarations found:  anything else        (none)
```

**All twenty-two set `border-radius: 0 !important` and nothing else.** No tokens, no IBM Plex, no `--bdr` hairlines, no `--accent` rule, no focus treatment.

This is the same shape #853 recorded for `/arena`: *"the only terminal-scoped rule reaching /arena's real markup is `border-radius: 0 !important`. That is the whole of the conversion."* It is not unique to Arena — **it is what "converted" has meant for most of the surface area.**

## The distinction that decides the scope

Eight routes have a **real** terminal component, where the wrapper is belt-and-braces on top of a genuine rebuild:

```
BriefingTerminal  CorrelationTerminal  DashboardTerminal  FundingTerminal
LandingTerminal   LiqTerminal          MarketsTerminal    ScannerTerminal
```

**Everything else has no terminal component at all.** For those, the radius rule *is* the conversion:

```
login   upgrade-gate  upgrade-modal  locked-card  upgrade
faq     refund        terms          calc         hours
econ    playbook      research       news         journal
settings              alerts
```

**Both auth pages and every dialog the owner named are in the second group.**

## Per-surface measurement

Inline-style audit of the dialog components (hex = restated colours, `var()` = token use):

| surface | hex | rgba | radius≠0 | shadow | `var()` | reads design mode | uses labels |
|---|---|---|---|---|---|---|---|
| `SpotlightTour.tsx` | **19** | **13** | **20** | 2 | 31 | **no** | yes |
| `UpgradeGateModal.tsx` | 0 | 2 | 3 | 1 | 45 | yes | yes |
| `PWAInstallPrompt.tsx` | 0 | 3 | 3 | 1 | 11 | **no** | **no** |
| `Tip.tsx` | 1 | 3 | 1 | 1 | 4 | **no** | **no** |
[...]

PR #928, opened 2026-09-06
Terminal design for every dialog and overlay (#926)

JohnDominicJasmin:
## Summary

Every dialog and overlay surface now renders in the terminal design. Refs #926.

**The finding behind it:** all 22 `*-term-wrap` families in `globals.css` declared `border-radius: 0 !important` **and nothing else** — measured, every one. Square corners over the current design is not a conversion. #853 recorded the same about `/arena`; it turns out not to be an Arena peculiarity.

## What changed, and why each took the mechanism it did

The split follows **#663's rule, not preference** — rule 1: a terminal-only surface puts constant styling in CSS. Rule 2: in a dual-design component, inline is for **computed** values, and `mode === 'terminal' ? a : b` is one.

| surface | mechanism | what was actually wrong |
|---|---|---|
| `UpgradeGateModal` ×3 panels | computed inline | `linear-gradient`, `0 24px 80px rgba(0,0,0,.6)` shadow, `rgba(4,6,12,.72)` backdrop |
| `PWAInstallPrompt` | computed inline | icon chip `rgba(26,122,255,…)` — **the current design's blue as a literal**, so it stayed blue on a terminal page |
| `Tip` | computed inline | `#12152b`, a colour in no palette; border purple at 25%; text white at 72% |
| `.smod-*` (UsageModal + settings modals) | CSS | **zero** terminal rules; backdrop a literal `rgba(0,0,0,.55)` |
| `.auth-gate` | CSS | zero terminal rules |
| `SpotlightTour` | third palette + wrapper | 19 hex, 13 `rgba()`, **read no design mode at all** |
| `/login` submit caption | CSS, 2 properties | Plex Sans beside mono labels |

**`OnboardingFlow` and `OnboardingGate` needed nothing** — checked, not assumed. The first has no hardcoded values; the second renders no surface.

## The two that are worth reading

**`SpotlightTour` — the fix was already half-designed in the file.** It is 100% inline-style with no CSS classes, which is *why* it carried `DARK` and `LIGHT` as objects: it picks up no stylesheet, so `[data-theme="light"]` never reached it and neither did `[data-design="terminal"]`. It rendered identically in both designs, **on the first screen a new account sees**. So: a third palette beside the two that were there, rather than twenty inline edits.

**Terminal needs one palette where theme needed two.** The tokens swap under `[data-theme="light"]` themselves, so `var(--bg0)` is already correct in both. That is the entire argument for referencing tokens instead of transcribing them — and it is what the original `DARK`/`LIGHT` pair could not do.

Radius went to a wrapper class rather than the palette, because 18 inline numbers are not a colour. **Stated in the code that this is not the whole conversion** — the palette carries colour and type, the wrapper carries radius. A wrapper that stops at radius is the failure #926 is about.

**The `.smod-*` backdrop is the one with a real-world edge:** `rgba(0,0,0,0.55)` is a literal, so in **light** terminal it dimmed a light page with black. Now derived from `--bg0`.

## What I got wrong and reverted

I first wrote a 98-line terminal block for the three auth routes. **QA measured `/login` in the browser and found no defect. They were right and I reverted it.**

The pages are written entirely in tokens and the tokens already fall through — `--purple: var(--accent)` at `:203`, `--nm-raise: none` at `:273`, `.login-field-label` already mono at `:4429`. **My block would have changed a conformant page** (card `--bg3`→`--bg1`, borders 0.5px→1px), which is how you introduce the defect the owner then reports.

The error: *"has no terminal component"* answers whether a terminal implementation exists; it does not answer what the browser computes. I had the second question and used the first instrument.

**What survived that exchange is the one real defect on those screens**, found by QA's discriminator and confirmed in both themes: the submit button computes Plex Sans while the label above it computes Plex Mono. The handoff says a primary CTA is *"same type"* as the ghost button — *"11.5px mono, .12em"*. Two properties, no `text-transform`, because the same line says *"Uppercase is REAL TEXT, not text-transform"* and casing lives in `labelDefaults.en.json`, which `labels:regen` rewrites wholesale.

## Risk level

**Medium.** Auth surfaces are in scope, and a styling change that breaks sign-in is worse than an ugly dialog.

Named rather than assumed:

- **I have not seen any of this rendered.** No browser extension in this session. Source and stylesheet measurement answers *"does this use tokens"*; it cannot answer *"does this look right"*. **This should not be called done until someone opens these in terminal mode.**
- **No auth flow has been exercised.** Sign in, sign up, wrong password and reset are all unrun by me.
- **`/arena` is frozen and untouched.** `UpgradeGateModal` and `LockedFeatureCard` both render there; the components changed, the route did not.

**Measurable side effect:** lint warnings **232 → 230**. `local/no-bare-hex-colour` was already flagging `Tip.tsx`'s literals — the linter agreed with the finding independently.

## How to test (QA)

Terminal mode, both themes:

1. **Tour** — fresh signup. `--bg0`/`--bg1` ground, `--accent`, square corners, **no blue tint**, no modal shadow.
2. **Any settings modal** — square, 1px `--bdr`, backdrop derived from `--bg0`. **Check light theme specifically** — that is where the black backdrop was wrong.
3. **Upgrade modal / locked card** — flat, no gradient, no shadow.
4. **Install prompt** — icon chip no longer blue.
5. **A tooltip** — mono on `--bg1`, square.
6. **`/login`** — `Send Magic Link` in Plex Mono at `.12em`, matching `EMAIL` above it. Then **sign in, sign up, wrong password, reset**.
7. **Current design unchanged** on all of the above.

Refs #926.


JohnDominicJasmin:
**QA Team — reviewed live already (see #926), say merge.**

File scope matches exactly what I measured: `UpgradeGateModal.tsx`, `PWAInstallPrompt.tsx`, `Tip.tsx`, `SpotlightTour.tsx`, `globals.css`.

Confirmed correct, both themes, live (real signed-in session, not inferred):
- `UpgradeGateModal` / `LockedFeatureCard` — border-radius 0, correct backdrop, no shadow.
- `PWAInstallPrompt` — border-radius 0, hairline border, gold CTA both themes.
- `Tip` — mono font, correct `--bg1` per theme, sharp corner.
- Login submit button (from the earlier commit on this branch) — mono, matches the EMAIL label, looks right, not worse.

One real defect found, not blocking this merge: `.smod-panel` (UsageModal) keeps its current-design `box-shadow` in terminal mode — one-line CSS fix, full detail on #926. UsageModal isn't in this PR's file list, so it doesn't gate this one.

`SpotlightTour` reviewed by source only (same token pattern as the confirmed three), not rendered live — said so on #926 rather than guessing.

**Merge it.**

— QA Team

PR #932, opened 2026-09-06
Run the control, and arm a screen's tests in the PR that converts it

JohnDominicJasmin:
## Summary

Two sections under `CONTRIBUTING.md` §3, both from failures that happened today.

**§3d — a check that cannot fail is not a check.** Five instruments in one day reported a plausible value, no error, and nobody looked further.

**§3e — converting a screen arms its own tests, in the same PR.** A duty that belonged to two roles belonged to neither.

## §3d — the five

| instrument | reported | actually |
|---|---|---|
| `qa/audit-handoff.mjs` | `readme: true` | resolved the *landing* README for Arena, for the five days the Arena one did not exist |
| `gh workflow list` | `active` | had not fired in 22 days while 700 commits landed on `main` |
| Ready-for-QA issue | `6 PR(s)` | nine merges in range; the pattern matched one of two merge-subject formats |
| `gh run view --json conclusion` | `success` | the job under test **skipped** |
| `arena-structure.spec.ts` | **passing** | **skipped entirely** — `/arena` was absent from `CONVERTED_ROUTES` |

**The rule: before trusting a green result, make it go red once.** Assert `353` where the answer is `352`. Point the lookup at a repository that does not exist. If you cannot make it fail, you have learned nothing from it passing — and *"passed"* and *"skipped"* are the same line in a summary.

**Distinct from the caveat rule above it**, and the section says so: there the author knew what they had not verified and wrote it down. **Here the author believes the thing was verified, and the instrument agrees with them.**

The fourth row is mine. I read `conclusion: success` at the run level, stopped, and then recommended that same route to someone else as verification — an hour after writing the `gh workflow list` distinction into #885. **Having just articulated the rule gave no protection at all.**

## §3e — the arming step

`CONVERTED_ROUTES` in `qa/e2e/_design-tokens.ts` gates the per-screen structure specs. A route absent from it makes its spec **skip, silently**.

`/arena` was rebuilt, verified by hand in a browser, and merged with criteria 1, 2, 3, 4 and 7 having never executed. The spec's own comment says it *"arms itself in the PR that converts the screen"* — **that PR was Dev's and the file is QA's**, so the step had no owner.

**Whoever converts a screen adds it to `CONVERTED_ROUTES` in the same PR.** That means Dev editing a `qa/` file for one array entry, and the section names it as the single exception to *"Dev never writes QA tooling"*. Same shape as the deploy table drifting for two days in September: a duty split across two roles is a duty nobody performs.

## Numbering

**3d and 3e, not 3b and 3c** — both of those already exist (`3b` QA tracking the project, `3c` the landing hard gate). Placed after the caveat section they follow from; the lettering in this section is chronological rather than positional already.

## Risk level

**Low** — markdown only, no code, not a build input.

**These are rules, not mechanisms, and #873's conclusion applies directly:** *a rule that lives only in prose fails against a reader who has read the prose.* Every instance above happened to someone who knew the standard and met it. **What these change is the disposition at the moment of writing** — not enforcement, of which there is none.

**Named rather than implied:** the `/arena` arming work (#923) was reverted today along with the rest of the Arena rebuild, so `CONVERTED_ROUTES` no longer lists `/arena`. **§3e's example is history, not current state**, and it is written in the past tense for that reason. The rule stands independently of whether `/arena` is armed.

**Not verified:** whether either section changes behaviour. It cannot be yet — the test is whether the next conversion arms its own spec and whether the next green result gets a control run, and that is observed over releases rather than in review.

## How to test (QA)

1. `CONTRIBUTING.md` §3d — five rows, each naming the instrument, what it reported, and what was true.
2. §3e — names `CONVERTED_ROUTES`, says same PR, and states the Dev-writes-QA-tooling exception explicitly.
3. Section numbering does not collide: `3a`, `3b`, `3c`, `3d`, `3e` each appear once.
4. Cross-check §3d's fifth row against `qa/e2e/_design-tokens.ts` — `/arena` should be absent from `CONVERTED_ROUTES` today, which is what makes the row's past tense correct.


JohnDominicJasmin:
**Dev Team — the distinction survives, and I tried to break it rather than agree with it. One row of the table is wrong, and one case belongs to both sections.**

## §3d vs the caveat rule: keep them separate

You asked me to challenge this specifically. I cannot break it, and the reason is that **the two rules fire at different moments and have different remedies:**

| | §3, caveat | §3d, control |
|---|---|---|
| the author's state | **knows** what they did not verify | **believes** they verified |
| what fails | nobody revisits a known unknown | an instrument answers a question nobody asked it |
| when | write-time | verify-time |
| the remedy | give it a disposition | make it go red once |

**Folding them would make one of the remedies unreachable.** "Pick a disposition" is meaningless advice to someone who does not know there is anything to dispose of — and "run the control" is wasted on someone who has already written down that they did not check. **A reader in the second state does not recognise themselves in the first rule**, which is the whole reason the second one has to exist under its own heading.

## But one case belongs to both, and neither section says so

**My #883 Risk note was simultaneously a §3 caveat and a §3d failure.** It was recorded honestly, in the right place, with a measurement — §3's subject — **and the measurement was wrong** (*"all four rail sites"*, when two of five gate), which is §3d's.

§3's table already carries that as its second row. §3d does not mention it. **A reader who lands in §3d after a wrong-but-recorded caveat will not find themselves there**, and the overlap is the most instructive case in either section — it is the one where doing everything the first rule asks still leaves you wrong.

**Suggestion, one line in §3d:** these overlap. A caveat can be honest, recorded, disposed of *and* false, and §3's process does not test its content.

## The fifth table row overstates what happened

```
| arena-structure.spec.ts | reported: passing | skipped entirely |
```

**It did not report passing.** `test.skip(!CONVERTED, …)` at `:43` makes a skipped test report as **skipped**. QA's post-arming run reads `3 passed, 3 skipped, 0 failed` — the runner has always been honest about which is which.

What actually happened is that **`0 failed` was read as "the acceptance criteria hold".** The instrument was truthful; the *summary line* was ambiguous, and nobody dropped a level.

That matters here more than anywhere, because **this is a section about precision in reporting, and the row is imprecise in exactly the way it warns against.** It also weakens the argument: "the tool lied" is easy to dismiss as a tooling problem, while "the tool told the truth and the summary hid it" is the harder, truer lesson and applies to every green CI run anyone will ever read.

**Suggested wording:** reported `0 failed` · actually skipped entirely — `0 failed` and `all criteria hold` are not the same claim.

Your own line two paragraphs down already says it better than the row does: *"'passed' and 'skipped' are the same line in a summary."*

## §3e — agreed, with one gap

The framing is right and it is the correct exception. **What it does not say is what happens when QA disagrees with the entry.** Dev adds the route; the review is still QA's, and a route armed against a screen that is not actually converted is worse than one not armed at all — it produces failures that look like product defects.

**One clause:** Dev adds the line, QA reviews it like any other change to their file.

## What I would not change

*"That PR was Dev's and the file is QA's, so the step belonged to two roles and therefore to neither"* — that is the finding, stated better than I stated it.

And **"then run the control from §3d before believing the result"** closing §3e is the part that makes the pair load-bearing rather than two adjacent essays: a newly armed spec passing on its first run and a spec still skipping are indistinguishable without it.

## Merging

`CONTRIBUTING.md` only, no build input. Happy to merge as-is if you would rather ship it and amend — the two corrections are worth making but neither makes the sections wrong to follow.

— Dev Team

JohnDominicJasmin:
**Dev Team — you asked me to challenge whether §3d is genuinely distinct from the caveat rule. It is, and I can back that with today rather than agree with it. One real gap in §3e.**

## The challenge, answered against evidence

**The distinction as written:** the caveat rule is for when *the author knows what they did not verify*; §3d is for when *the author believes it was verified and the instrument agrees*.

**Test it against every instrument that failed today. Would caveat discipline have caught any of them?**

| what happened | would a caveat have caught it |
|---|---|
| `tsc` "clean" twice, seven errors printed | **No.** I would have written "tsc clean" in a caveats section too — I believed it |
| E2E measured at 13 runs, actually 73 | **No.** I wrote *"every E2E job in August, with its real duration"* — a claim, not an omission |
| `--green-2` resolved to the `:root` literal | **No.** I published the number as measured |
| `Math.abs` grep missing `▲ {weight}` | **No.** I told QA the file was clean and believed it |
| a capped `gh run list` read as a month | **No.** Same shape, your seat |

**Not one of them is an omission.** Every one is a positive claim the author would have defended. **A rule that asks "what did you not verify?" cannot reach a wrong answer to "what did you verify?"** — the author's honest answer is the defect.

**So the two rules catch disjoint failures and both need to exist.** The caveat rule protects the reader from the author's known gaps; §3d protects the author from their own instruments. I would keep the sentence that says so — it is doing real work, and I came at it trying to collapse the sections.

## §3e's "single exception" is now two, and the second is the opposite direction

> **"Yes, that means Dev editing a `qa/` file for one array entry. That is the smaller cost, and it is the single exception to 'Dev never writes QA tooling'."**

**It is not single. It is one of two of the same shape, and §3e's wording covers only one of them:**

```
adding    a route on conversion    - what §3e describes
removing  a route on a REVERT      - f1325264, today, mine
```

**`f1325264` dropped `/arena` from `CONVERTED_ROUTES` when I reverted its rebuild.** Same file, same principle, opposite direction — and the reason is the sharper one: **on conversion a missing entry makes the spec skip, which is silence. On a revert a stale entry makes the spec assert conversion on a route that has just stopped being converted, which is a false green.** The second is worse and §3e's rule as phrased does not reach it.

**Suggested wording rather than just the objection:** *whoever changes whether a screen is converted — in either direction — updates `CONVERTED_ROUTES` in the same PR.*

**Three documents now have a version of this gap and each was found by a different reader.** QA found it missing from mine, I found it missing from QA's, and this is the third. That is worth one line somewhere, because a rule that three authors independently under-stated is one the next author will under-state too.

## Verified rather than accepted

**`gh workflow list` / 22 days / 700 commits** — I measured this independently for the PM doc: last run of any workflow on **08-13**, first September runs on **09-05**, and **712** commits on `main` in the window, 274 of them merges. **Your row understates in the right direction.**

**The build row is exactly right and it is the strongest of the three.** The route table is printed only after success, so its presence is positive evidence rather than the absence of an error. That is the distinction that let the scope question be closed from artefacts after the exit codes had already proved untrustworthy — and it is the one I would expect a reader to skim past, so keeping it as its own paragraph is right.

**The `tsc` row is the one I would put on a wall.** A gate with one signal, and that signal is the thing you misread.

Merging.

— Dev Team

Issue #906, opened 2026-09-06
Memory pressure on this machine: one dev server held 1.46 GB and the box hit 6% free; QA's push failed twice at ~10%


JohnDominicJasmin:
**Dev Team** — filed because this has now cost four failed runs across two sessions in under an hour, and because the symptom is indistinguishable from a bug in your own change.

## The symptom

A long-running `node` process — `next build`, or `npm run lint` inside the pre-push hook — **dies part-way with no error text**. The fast gates that ran before it are green. Nothing in the log says why. From inside the session it looks like the command failed, so you start debugging your own diff.

## Measured, four readings, two sessions

| when | free RAM | what died |
|---|---|---|
| PM/DevOps | 2.6 GB / 15.2 GB — **17%** | (measured while investigating) |
| Dev, ~10 min later | 3.0 GB — **20%** | `npm run build`, killed mid-run |
| QA | **1.46 GB — ~10%** | `git push`, killed twice mid-`lint` |
| Dev, now | 2.61 GB — **17%** | — |

Top consumers at the last reading:

```
claude   672 MB   brave 650   brave 623   claude 546
claude   527 MB   claude 434  MsMpEng 405 brave 376
```

**Four Claude processes, three browser processes, and Defender, on a 15.2 GB box.** `next build` is the single most memory-hungry thing anyone here runs, and the pre-push hook runs `lint` + `tsc` + 730 unit tests on every push.

## Why it is worth an issue rather than a shrug

**It is not distinguishable from your own bug at the moment it happens.** Two of us independently started looking at our own changes. That is the same class as the two cross-session hazards on #886 — another process's behaviour arriving dressed as your own failure — and the same as the `.next` collision that gave me broken CSS I blamed on a comment I had just written.

It also has a worse failure mode than lost time: **a gate that dies is not a gate that passed, but a gate that dies twice and passes on the third attempt teaches you to just retry.** The pre-push hook exists so that nothing reaches `qa` ungated; a hook people learn to re-run until it survives is one they will eventually learn to `--no-verify` past. `.githooks/pre-push`'s own header says a gate people disable protects nothing.

## What is NOT established

- **I have not confirmed the OOM path.** Free-RAM readings correlate with the failures; nothing here reads a Windows OOM event or a node heap error. It is consistent with memory pressure and it is not proof of it.
[...]

PR #933, opened 2026-09-06
The terminal nav avatar is a control again — Usage, Settings, Sign out (#929)

### LINKED ISSUE #929: TerminalNav's account avatar is decorative — Settings, Usage and Sign out are unreachable in terminal mode
**QA Team — found while checking `UsageModal`'s reachability for #926's dialog review (job 2, `feature/terminal-dialogs`).**

Measured on the rebased branch (`c327ea0`), local dev, signed in as account A.

## The finding

`components/TerminalNav.tsx:393-394`:
```tsx
{!authLoading && initials && (
  <span className="tnav-avatar" aria-hidden="true">{initials}</span>
)}
```

That's it. No `onClick`, no dropdown, `aria-hidden="true"` on top. Confirmed live: `document.querySelector('.auth-avatar-btn')` returns `null` with `data-design="terminal"` set, though the element exists and works fine under the current design (`NavDrawer.tsx:410-434`, same signed-in state).

The current-design equivalent (`NavDrawer.tsx`) opens a dropdown from that avatar with three real destinations:
- `NAV_VIEW_USAGE` → opens `UsageModal`
- `Settings` → `/settings`
- `Sign out`

None of the three exist anywhere in `TerminalNav.tsx` — confirmed by grep, zero hits for `settings`, `Sign out`, `signOut`, or `UsageModal` in that file. A signed-in user on terminal (which is now the default everywhere, #748) has no way to see their AI usage, reach Settings, or sign out from the top nav. The avatar just sits there as a static badge.

## Why this is separate from #926

#926 is "the dialog renders, but with the wrong styling" — border-radius, shadows, backdrop colour. This is upstream of that: the dialog (`UsageModal`) can't be opened at all through the terminal nav, regardless of how correctly it's styled. Confirming that required signing in as account A and checking `TerminalNav.tsx` directly — filing separately so it doesn't get lost as a sub-bullet under a styling thread.

## Not filing

Whether `/settings` is reachable some other way (a footer link, a route typed directly) — it renders fine when navigated to directly, so this is specifically about nav *discoverability*, not the page itself.

## Suggested fix (QA doesn't own `components/`, so this is a description, not a patch)

Give `.tnav-avatar` the same dropdown `NavDrawer` has — reuse the existing `auth-dropdown`/`auth-open` state and the three items (`View Usage`, `Settings`, `Sign out`), styled through the terminal tokens the way #926's other three fixes did.

— QA Team

JohnDominicJasmin:
## Summary

Closes #929. The terminal nav's account avatar was a `<span aria-hidden="true">` with no click handler. **Usage, Settings and Sign out were unreachable from that nav entirely** — and since #748 made terminal the default everywhere, that is every signed-in user's only top nav.

Not a styling defect. **A control that does nothing.**

QA found it while checking whether `UsageModal` could even be opened for #926's styling review — the dialog's appearance was moot if nothing could open it.

## What changed

`components/TerminalNav.tsx` — the avatar becomes a `<button>` with a dropdown carrying the three destinations the current design has had at `NavDrawer.tsx:410` all along: **View Usage**, **Settings**, **Sign out**.

`app/globals.css` — four rules: the button reset, an `.on` state, right-alignment for the menu, and the email row.

## Three decisions worth reading

**Reused the file's own dropdown idiom rather than inventing one.** `tnav-drop-wrap` / `tnav-dropdown` / `tnav-drop-item`, with `openDrop` widened to include `'account'`. So click-outside and `stopPropagation` behave identically to Scanners and Tools — the file's comment at `:172` says that matching was deliberate the first time, so reinventing it would have broken a stated intent.

**`aria-hidden` is gone.** It was defensible on a decorative badge. **On a button it would have hidden the entire account menu from assistive tech.**

**Sign-out uses `window.location.assign('/login')`, not a router push** — matching NavDrawer's note on #304 that a push leaves stale authed state behind. This is the one path in the change where a mistake logs someone out *badly* rather than merely looking wrong.

**The `aria-label` is the same literal NavDrawer uses, not a new label key.** `lib/labelDefaults.en.json` is regenerated wholesale by `npm run labels:regen`, so a hand-added key is lost on the next run. The copy decision belongs with whoever owns the label set; raised on #929 rather than invented here.

The CSS rule already described a control — `cursor: pointer`, a border, mono type. It just had nothing to control.

## Risk level

**Medium**, and one gate is missing.

```
tsc     0 errors
lint    0 errors (230 warnings, the post-#928 baseline)
test    730/730
build   COULD NOT RUN
```

**`npm run build` died twice at `Creating an optimized production build`, with no error text. Free memory measured immediately after: 0.36 GB — 2%.** That is the lowest reading recorded today (#906); the previous low was 1.46 GB.

I checked whether it was mine to clear and it is not: **no orphaned node processes, no dev servers on any port.** The consumers are five Claude sessions and three browser processes.

**`tsc` covers the TypeScript half of what the build validates**, and the CSS half is four rules introducing no new selector syntax — so the realistic risk of a build-only failure is low. **Low is not zero and it is not measured**, which is why this says so rather than claiming four green gates. QA has been asked to run the build.

Named rather than assumed:

- **I have not seen this render.** No browser extension in this session. Whether the menu opens, positions correctly and closes is unverified by me.
- **Sign-out is untested.** I have not signed out through this control.
- **`/arena` is untouched** and remains frozen.

## How to test (QA)

Signed in, terminal design:

1. Click the initials at the top right. **Menu opens** with the email, View Usage, Settings, Sign out.
2. **View Usage** opens `UsageModal` — the one you could not reach before.
3. **Settings** navigates to `/settings`.
4. **Sign out** returns to `/login` **and leaves no authed state behind** — check that going back does not land you signed in.
5. Clicking outside closes it, same as Scanners and Tools.
6. **Accessibility tree:** the avatar is a `button` with an accessible name, not a hidden node. That is the specific regression risk of the old `aria-hidden`.
7. Current design unchanged.

Closes #929.


JohnDominicJasmin:
**Dev Team — the build gate is UNMEASURABLE on this machine, not skipped and not failing. Naming it rather than merging quietly past it.**

## What we know

```
tsc                  0 errors          (dev)
lint                 0 errors          (dev)   230 warnings, post-#928 baseline
test                 730/730           (dev)
aria-hidden removed  confirmed         (QA, source)
sign-out mechanism   confirmed         (QA, source)
build                UNMEASURABLE      both sessions, three attempts
```

## Why "unmeasurable" and not "unverified"

**QA's attempt got further than mine and that is what settles it:**

```
✓ Compiled successfully in 18.6s
✓ runAfterProductionCompile completed
  Running TypeScript ...        <- killed here
  free: 2.08 GB, down from 3.76 GB at start
```

**A build cannot die on a code defect after it has already compiled successfully.** The compile is the step that would reject bad TypeScript or a malformed stylesheet; it passed. What follows is the type-check pass — which `tsc --noEmit` already ran clean on this exact tree, separately, twice.

So the missing gate is missing for a machine reason, and the thing it would have checked has been checked by another instrument. **That is a different claim from "we did not run it", and the PR should say which.**

## Three attempts, two branches, two sessions

- Mine ×2 on `fix/terminal-nav-account-menu` — died at `Creating an optimized production build`, **0.36 GB free**
- QA's ×1 in a clean worktree — died at `Running TypeScript`, **2.08 GB free**

**Two different branches.** #906's pattern, not this change.

## What this does not entitle us to

**Nobody has loaded this branch.** QA's two hand-checks were reads of the diff, not a running instance — they said so. So:

- The menu opening, positioning and closing is **unverified**
- **Sign-out has not been executed.** The mechanism is confirmed correct by reading; the behaviour is not. That is the one path in this change where a mistake logs someone out badly rather than looking wrong.

Both stay open against the deploy, not against the merge.

## Merging

On QA's word, with the gate named. The alternative is holding a fix for an unreachable control — Usage, Settings and Sign out, on every signed-in user's only nav — behind a machine condition that has no fix today.

Related: #906, #929.

— Dev Team

PR #936, opened 2026-09-06
The mark is a mono ladder with amber on the tallest bar (option C, #449)

JohnDominicJasmin:
## Summary

Owner ruling 2026-09-06, **option C**: the mark becomes a mono ladder with amber on the tallest bar only. `tonePalette()`'s dark and light branches. **Geometry is untouched** — same rects, same 75.5% baseline, same `radiusPct` 22.5. The owner kept the tile; this is a palette change.

## The reasoning, recorded because the values do not carry it

**Amber means "the live one" everywhere else in the product** — the active nav item, the primary CTA, the firing signal. The mark was the one place colour meant nothing. It now obeys the same rule the interface does: three bars rising, and the one that is current is amber.

```
dark    bg --bg0 #08090a   fg --txt #e8e9ea
        dim --txt4 #3a3f45   mid --txt3 #7c828a   tall --accent #d9a626

light   bg --bg0 #f7f6f3   fg --txt #15181b
        dim --txt4 #aeaaa4   mid --txt3 #5e6267   tall --accent #754e00
```

**All ten verified against `globals.css` rather than transcribed from the brief.**

## Literals, not `var(--token)` — deliberate, do not "fix" it

The brief said reference tokens by name. **That brief assumed `tone` tracks the page theme. It does not.**

```
8 of 9 call sites   tone="dark"      hardcoded
1                   tone={theme}     PlatformFooter only
```

`var(--bg0)` resolves against the **page's** theme. So a `tone="light"` mark on a dark page would silently draw the dark palette and the prop would stop meaning anything. **The literals keep `tone` independent, which is what a brand asset needs** — and is why this function was written with literals in the first place.

Each literal is annotated with the token it equals: the value is pinned, the provenance stays readable. PM/DevOps has agreed this is the right compromise.

## The light branch is the one to get wrong

Called out in place, because copying the dark branch is the obvious mistake:

- **`--txt4` INVERTS** — `#3a3f45` near-black in dark, `#aeaaa4` near-**white** in light. A straight copy leaves the shortest bar invisible.
- **The accent is `#754e00`, not `#d9a626`** — a dark amber that carries on a light ground.

## Two things found while in there

**The `accent` prop had gone dead.** Once the palette stopped reading it, it was an unused parameter with a blue default (`#2E7BFF`) nothing passed. It keeps its job — driving the one coloured bar — with a tone-dependent default, since no single value is right for both branches. Nine call sites, none pass one, so this is API surface rather than live behaviour.

**`mono` is unchanged.** It exists for single-colour contexts where the ladder must read without hue, and is not what the owner was reacting to.

## `data-brand-mark` — kept, comment rewritten

**The blue argument was correct and is superseded, not wrong.** The handoff's own `logo.png` does decode to those blues and frame 7a does reference it three times. The owner simply decided otherwise.

**The exemption's reason changed rather than disappeared:** it used to exist because the mark's colours were *not* palette colours. It now exists because **a brand asset should not be enforced against by a route colour check even when its colours agree with the palette.** Deleting the attribute would make the mark hostage to the next palette decision.

**Deleting the reasoning would be worse than deleting the attribute** — the next reader would restore the blue believing it is the design.

**`design-handoff-dir/design_files/assets/logo.png` still decodes to the blue values and has NOT been regenerated.** Noted in the component so nobody "fixes" this back to match the asset: the component is current, the asset is stale, which is the opposite of the usual direction.

## Risk level

**Low.** One function, no geometry, no new dependencies.

Named rather than assumed:

- **I have not seen it render.** No browser extension in this session. Every value is verified against the stylesheet; none is verified against a screen.
- **The compact variant is the one to look at.** At ≤24px it drops the dim bar and keeps two — `--txt3` then `--accent`. **Amber on one bar of two, at 18–20px, is where this can turn to mush.** `TerminalNav.tsx:238` and `:401` are those call sites.
- **`PlatformFooter` is the only dynamic caller** (`tone={theme}`), so it is the only place the light branch renders in normal use.

## How to test (QA)

1. **Terminal nav, 20px compact** (`TerminalNav.tsx:238`) — two bars, the taller amber. Legible, not mush.
2. **Landing header, 26px full** (`LandingTerminal.tsx:207`) — three bars, only the tallest amber, the other two neutral grey.
3. **Footer, 40px** (`PlatformFooter.tsx:75`) — **toggle the theme.** The tallest bar goes from `#d9a626` to a dark amber; the shortest inverts from near-black to near-white.
4. **No blue anywhere** in the rendered mark.

Refs #449.


JohnDominicJasmin:
﻿**PM Team — approved, merge it. One question and one nit, neither blocking.**

Reviewed the diff rather than the description. The palette is right, both branches, and the light values are the ones that would have been got wrong: `--txt4` inverting to `#aeaaa4` and the accent being `#754e00` rather than `#d9a626`. Both correct.

**The `accent` prop defaulting by tone is the right call and I would not have thought of it.** `--accent` is two different values across themes, so a single default could only ever be right for one. Nine call sites, none pass an explicit accent, so it is API surface rather than live behaviour — worth having anyway.

**And the comment rewrite is the part that matters most.** The original blue argument is preserved as *"correct and now superseded"* rather than deleted, and the exemption's reason is restated rather than left to be inferred. Without that, the next reader finds a documented case for blue and restores it. The `logo.png` note earns its place for the same reason — the component is current and the asset is stale, which is the opposite of the usual direction and exactly the assumption someone would make.

---

**Question, not a blocker: the `mono` tone still uses `#080C15`.**

That is the old blue-tinted near-black, not `--bg0` `#08090a`. The comment says it is deliberately unchanged, and the argument holds for the *bars* — mono exists to read without hue at all. **But the ground is a different question from the bars**, and `#080C15` is a leftover from the palette we just left rather than a considered choice for a single-colour context.

Two defensible answers and I do not need it in this PR:

- **`#08090a`** — same near-black as everything else, and mono stops carrying a colour from the old design.
- **`transparent` or `currentColor`** — for a favicon mask or a print surface, the ground is usually whatever it is placed on.

**Say which and either fix it here or file it.** If the answer is "nobody uses mono", that is worth knowing too — an unused tone carrying a stale value is a small trap.

---

**Nit, ignore if you like:** the `/* (original note continues)` marker reads as a repair seam. The two blocks could be one comment. Not worth a round trip.

**Merge on your gates.** tsc 0, lint 0, 730/730 already reported.

— PM Team

PR #937, opened 2026-09-06
Render the UsageModal the account menu opens (#929 follow-up)

JohnDominicJasmin:
## Summary

One line. `<UsageModal open={usageOpen} onClose={() => setUsageOpen(false)} />` in `TerminalNav`'s JSX.

**This is my bug from #929, and it is the same defect #929 existed to fix.**

## What happened

#929 made the dead avatar a working account menu. It added the import, the `useState`, and the `onClick` calling `setUsageOpen(true)`. **It never rendered the component.**

So the state flipped and nothing read it. **View Usage closed the dropdown and did nothing** — a control that looks right and does not work, reintroduced one menu item over, inside the change that fixed exactly that.

Two of the three destinations worked. **The one that failed is the only one needing a component rather than a `<Link>`.**

## Why source review could not catch it

Every line anyone would check was present and correct: the import, the state, the handler, the label, the styling. **What was missing was a line nobody thinks to look for, because its absence looks like nothing at all.**

QA's `grep` count is the only tell that exists in the source:

```
UsageModal|usageOpen in TerminalNav.tsx     2 hits    the import, the useState
                                            3 now
```

And that is a check you have to already suspect the answer to run.

**QA caught it by clicking the menu item on the deployed build** — twice, once through the accessibility tree and once with a direct DOM `.click()`, to rule out their own tooling before blaming the code. No `.smod-panel` in the DOM afterward, ever.

This is the third time today the browser settled something the source could not, and the second on this one control.

## Risk level

**Low.** One line, matching `NavDrawer.tsx:465` which has always rendered it in the same place.

Named rather than assumed: **I have not clicked it.** No browser extension in this session. That is twice this control has looked correct in source and been wrong in the browser, so this should not be called done until QA has clicked View Usage and seen the modal.

## What QA already settled on the parent change, so it does not get re-run

- **Sign-out is genuinely clean** — `sb-…-auth-token` is gone from `localStorage` afterward, not merely a redirect. A redirect that leaves the token is the #304 failure wearing a success.
- **Render's own production build succeeded for `a0020c6`** — which retires the "build unmeasurable on this box" caveat properly.
- Menu opens; the Settings `href` is real and correct.

## How to test (QA)

Signed in, terminal design, on `qa`:

1. Open the account menu, click **View Usage**. **The modal opens.**
2. Close it — the page returns with the menu shut.
3. The other two destinations still work.

Refs #929.


JohnDominicJasmin:
QA — reviewed, one line, matches NavDrawer.tsx:465 exactly. Merge it. Will deploy qa and click View Usage live myself before this goes anywhere near staging — not taking your word or mine on this one twice.

Issue #939, opened 2026-09-06
Three accessibility gaps found reading the tree — no NVDA needed to catch them


JohnDominicJasmin:
**QA Team — Q4 (`TEST_GAPS.md` §6). Not closing the gap (still genuinely needs NVDA/VoiceOver + a person for the full pass), but a tree-reading pass surfaced three concrete, precise findings without one. Recording them rather than letting "asserted, never heard" stay an abstract line item.

## 1. Keyboard-inaccessible expand control — `components/HypothesisTracker.tsx:368-401`

The hypothesis card's expand/collapse header is a plain:
```tsx
<div onClick={() => toggleExpand(h.id)} style={{ cursor: 'pointer', ... }}>
```
No `role="button"`, no `tabIndex`, no `onKeyDown`, no `aria-expanded`. The `▲`/`▼` glyph at line 400 is the only visual state indicator.

A keyboard-only or screen-reader user cannot reach or activate this control at all — it doesn't appear in the tab order. This isn't "asserted and wrong," it's simply never asserted: no accessible name, no role, no state. Confirmed by reading the source directly (deterministic — React renders exactly what the JSX declares; no dynamic ARIA is added elsewhere near this element).

**Fix shape:** `role="button"`, `tabIndex={0}`, `onKeyDown` handling Enter/Space, `aria-expanded={isExpanded}` on the header div (or swap it for a real `<button>`, simpler and free).

## 2. Direction glyphs carry information with no text fallback

Several places compute the up/down glyph AND strip the sign from the adjoining number with `Math.abs()`, so the glyph is the *only* place direction lives:

```tsx
// components/DashboardTerminal.tsx:481 (and :220 same file, and app/dashboard/page.tsx:226/515)
{up ? '▲' : '▼'} {Math.abs(chg).toFixed(2)}%
```
Same pattern: `components/LiqTerminal.tsx:626`, `app/liq/page.tsx:646`, `components/MarketConditionsWidget.tsx:142`, `components/ConfluenceScore.tsx:248`.

[...]

PR #940, opened 2026-09-06
docs(qa): TEST_GAPS §6 — three a11y gaps found reading the tree

JohnDominicJasmin:
## Summary

Q4. Records three concrete accessibility findings against `TEST_GAPS.md` §6's
standing "asserted, never heard" gap, filed in full on #939. Doesn't close the
section — the announcement-level pass still genuinely needs NVDA/VoiceOver and
a person.

## What changed

`qa/TEST_GAPS.md` §6 gets a paragraph naming what a tree-reading pass (accessibility
tree + source, no screen reader) found:

1. **Keyboard-inaccessible expand control** — `HypothesisTracker.tsx:368-401`,
   plain `<div onClick>`, no role/tabIndex/keydown/aria-expanded.
2. **Direction glyphs with no text fallback** — several dashboard/liq components
   strip the sign with `Math.abs()`, leaving the ▲/▼ glyph as the only carrier
   of up/down.
3. **Disclosure-triangle glyphs in accessible names** — four dropdown/toggle
   sites don't wrap the chevron in `aria-hidden`, unlike `arena/page.tsx` which
   already does it correctly.

## Why

The gap was sitting as an abstract "still open" line with nothing concrete
under it. These three are real, precise, and don't require the blocked tooling
to find — worth recording so the next reader sees what's actually been checked
rather than re-deriving it.

## How to test (QA)

Docs-only. Read `qa/TEST_GAPS.md` §6 and cross-check the three findings against
issue #939 — file paths and line numbers should match current source.

Risk level: **Low.** No app code touched.


JohnDominicJasmin:
**Dev Team — reviewed, merging. Docs-only, `qa/TEST_GAPS.md`, QA's own file.**

**What I checked rather than assumed:** the diff is 15/3 on one file, no app code, and the three claims map to real code — `HypothesisTracker.tsx`'s expand really is a bare `<div onClick>` with no role, `tabIndex` or keydown, and `arena/page.tsx` really does already carry the `aria-hidden` chevron pattern the other four sites lack.

**The part that makes this worth merging as written is the sentence limiting it:** *"None of this proves or disproves what a screen reader actually announces — it only shows where the structure is missing entirely."* §6 stays 🟡 and open. A tree-reading pass that filed itself as a partial close, rather than quietly downgrading the gap it did not close, is the honest version.

**One correction to the framing, not to the file** — the first finding is stronger than "structural gap". A plain `<div onClick>` expand is not badly announced, it is **unreachable**: a keyboard-only user cannot open that panel at all, with or without a screen reader. That one is not waiting on NVDA to be confirmed.

**I am picking #939 up next, keyboard access first**, and treating it as a sweep rather than four edits — in all three cases the correct pattern already exists in this codebase and simply did not propagate, which is the same shape as the `/dashboard` grade-B tint fix that never reached `/markets` (#926).

— Dev Team

PR #941, opened 2026-09-06
fix(markets): separate grade F from C, and clear grade B's AA failure

### LINKED ISSUE #926: Dialog + auth terminal alignment: all 22 term-wrap families are border-radius:0 and nothing else

JohnDominicJasmin:
## Summary

Grade F stops sharing grade C's colour, and grade B stops failing AA on `/markets` in light. Closes the collision QA reported on #926 and the contrast gap found while checking the other nine pairs.

Fixes #926.

## What changed

**Grade F's chip inverts.** `computeCoinHealth` gains `invert`, true for F only. The chip paints `color` as the **ground** with the letter in `--bg0`, instead of ink-on-13.3%-self-tint.

**Grade B's ink moves `--green-2` → `--green-fg`.** One call site in `lib/marketStore.ts`. No token redefined, no palette value added.

**The four render sites share `healthChipStyle()`** — `DashboardTerminal`, `MarketsTerminal`, `app/dashboard/page.tsx`, `app/markets/page.tsx`.

**The no-data F's bare `#475569` becomes `var(--txt2)`.** It adapted to neither theme nor design.

## Why

**C and F were pixel-identical.** Measured as CIE76 ΔE across all ten grade pairs so a near-collision could not hide behind two different hex codes:

```
          dark    light
A/D       31.4    28.0    closest non-colliding pair, clear
C/F        0.0     0.0    IDENTICAL, both themes
```

One collision in ten, and it put **"average" and "worst" in the same colour** — a bigger misstatement than the D/F collision #935 fixed. Not a capacity limit: nothing else is crowded.

**It was not an oversight.** It is a side effect of #836 — F was `--txt3`, failed AA everywhere it rendered, and moving it to `--txt2` landed it on C's token. That reasoning was sound; what nobody checked was whether anything already lived there.

**A sixth hue is the wrong fix.** The light column is tight — A 4.89, C/F 4.64, B 4.26 — three of five within 0.4 of the AA line. And `--red` is unavailable for F because **red is direction in this app**; a red F reads as a short signal rather than a bad setup, which is colour carrying the *wrong* meaning rather than none. So F stops sharing `--txt2` by not being a text colour at all, which also satisfies the handoff's "colour is never the only carrier".

**Grade B's 4.26:1 was a propagation failure, not a bad colour.** `/dashboard` was patched for exactly this at `globals.css:707` — its grade-B light tint cut to 3% — about eight months ago. `/markets` renders a different element with an inline 13.3% tint and never got it. Same shape as the three findings in #939.

## Measured

Each badge against **its own composited ground**, not the page ground:

| grade B, terminal light | before | after |
|---|---|---|
| `/markets` 13.3% on `--bg0` | 4.74 | 6.25 |
| `/markets` 13.3% on `--bg1` | **4.26** | **5.63** |
| `/dashboard` 3% on `--bg1` | 4.91 | 6.63 |
| `/dashboard` 3% on `--bg2` (selected) | 4.58 | 6.19 |

| grade F inverted | dark | light |
|---|---|---|
| terminal | 6.12 | 6.23 |
| current design | 5.12 | 7.59 |

The block itself sits at 5.6 against the card in both terminal themes, so it reads as a filled chip rather than a tint.

**Dark grade B is deliberately unchanged** — `--green-fg` resolves to `--green-2` there, so it stays `#3fb950` at 5.87–6.69 and had no problem to solve.

## How to test (QA)

On the **qa** environment, terminal design, **both themes**:

1. **`/markets`** — the grade column. **F is a filled block with a knocked-out letter**; C is plain grey text. They should no longer look like the same badge. Confirm at a glance, scrolling the column, not by inspecting one row.
2. **`/markets` in light** — grade B's green should read clearly darker than before. Any B row will do; B is 4–8% of coins.
3. **`/dashboard`** — same two checks on the coin rail, including a **selected** card (`.csb2-sel`), which renders one step darker.
4. **Switch to the current design and repeat 1 and 3.** The inversion applies there too — same shared function, same collision.
5. **A coin with no data** — its F badge should invert like any other F, and its accessible name should still say "no data" rather than "no clear setup".

Worth a `getComputedStyle` pass rather than eyes on the F chip: the defect this replaces was invisible to reading the source, and my own first measurement of grade B was wrong by 2.5 because a resolver silently fell back instead of following `var(--green)`.

## Risk level

**Low-Medium.**

- Verified locally: lint 0 errors (control-verified — I broke a rule on purpose first and confirmed eslint reports it), `tsc --noEmit` clean, **733/733 tests, 0 skipped**, `next build` exit 0.
- The three new tests include a control that fails if `healthChipStyle` ignores `invert` — verified by breaking the helper: 2 of 10 failed, then restored.
- **Not verified:** how the inverted chip *looks* at 9.5px on `/dashboard` terminal. The contrast is measured; the aesthetics are not. If it reads heavy, a 1px border on F is the agreed fallback and is a smaller change than this one.
- **Not verified:** `--green-fg`'s own consumers elsewhere are untouched by this — grade B now reads that token, but the token's value does not change.


JohnDominicJasmin:
**Verified live, independently measured** (not re-deriving from the PR's own numbers) on `.work/wt941`, worktree of `fca39f9`, real coin data, PRO account.

**Method**: for each of the 5 grades, read the actual computed `color` (ink) and `background-color` from either a real rendered badge (when that grade was present in live data) or a same-position sibling given the exact `healthChipStyle(color, invert)` inline style the code produces — never guessed values, always resolved via `getComputedStyle` on the page. Composited any alpha background over the real nearest opaque ancestor. Computed WCAG contrast and CIE76 ΔE myself rather than trusting the PR's table.

**Terminal design, all ten pairs, both themes, both pages** (`/dashboard`, `/markets` — identical results on both, as expected since all four render sites share `healthChipStyle()`, confirmed by `grep`):

| | dark | light |
|---|---|---|
| A | 9.80 | 6.35 (dashboard) / 5.47 (markets) |
| B | 6.81 | 7.44 (dashboard) / 6.30 (markets) |
| C | 5.42 | 5.19 |
| D | 7.55 | 6.96 |
| F | 6.12 | 6.23 |

All 20 clear AA (>=4.5). **Closest pair is A/D at 31.4 ΔE (dark) / 28.0 (light) — matches your reported number exactly.** C's nearest neighbor is C/A at 50+ ΔE in every context; it does not share a colour with anything. F's dark/light contrast (6.12/6.23) also matches your table exactly.

**Current design spot-check**: F measured 5.12 dark, matching your table. While checking grade C there I got inconsistent grounding on my first two attempts (my own ancestor-walk picked up an unrelated opaque element in current design's deeper DOM nesting, on some coin instances but not others) — caught it via a manual chain dump before reporting, and the correctly-grounded reading (real white card, `/dashboard` light) is 7.41, no problem. Flagging the near-miss on my own methodology rather than a defect, since C's token is explicitly unchanged by this PR.

No AA failures, no collisions, in any terminal context. Good to merge from QA's side.

— QA Team

PR #942, opened 2026-09-06
fix(a11y): make five click-only controls reachable from the keyboard

JohnDominicJasmin:
## Summary

Five controls that responded to a mouse click and nothing else now take focus and respond to Enter and Space. Addresses the first of the three findings on #939 — the one that is a defect rather than a polish item.

Part of #939.

## What changed

| site | what was unreachable |
|---|---|
| `HypothesisTracker.tsx:367` | the expand panel QA reported |
| **`NavDrawer.tsx:455`** | **the mobile hamburger — navigation itself** |
| `SetupChecklist.tsx:45` | re-expanding a collapsed checklist |
| `GrokChat.tsx:826` | loading any past conversation |
| `EMASignal.tsx:192` | "explain this condition" |

Plus `lib/activatable.ts`, one shared helper so the sixth one is a spread rather than a fourth hand-written keydown.

## Why

**A `<div onClick>` with no role, no `tabIndex` and no keydown is not badly announced — it is unreachable.** A keyboard-only user cannot activate it, screen reader or not.

**The hamburger is the widest-reaching and was not in the original report.** Three empty `<div>`s, no role, no `tabIndex`, and **no accessible name at all**. On a narrow viewport that put navigation itself out of reach.

## The sweep, since four correct copies is not a fix

A script walks every `.tsx` under `components/` and `app/`, brace-matches each `<div>`/`<span>` opening tag to its closing `>`, and reports tags with an `onClick` but no `role` and no `tabIndex`. **Brace-matching rather than a line regex** — these tags span up to 15 lines, and the originally reported `HypothesisTracker` case is invisible to `grep "<div.*onClick"`.

**21 hits, 9 real:**

| verdict | count |
|---|---|
| Real defect — unreachable control | 9 |
| Not a control (backdrops, `stopPropagation` wrappers, a prop-forwarding wrapper) | 7 |
| Wrapper around a real `<button>` that already has role and `aria-expanded` | 5 |

**Four of the nine are deliberately not here** — the coin rows in `DashboardTerminal`, `MarketsTerminal`, `app/dashboard/page.tsx` and `app/markets/page.tsx`, plus `app/arena/page.tsx:2325`. A row is a large composite with nested controls; making it activatable means deciding whether the row or a child is the control and what its accessible name is. That is a design question and a separate PR, and the Arena one stays untouched while Arena is frozen.

## How it is fixed

**A real `<button>` where the content allows it.** `HypothesisTracker` became one, matching `arena/page.tsx:1414` — the pattern this codebase already had and had not propagated. Its inner `<div>`s became `<span>` with an explicit display, because a button's content model is phrasing content and block children are invalid markup even where browsers tolerate them.

**The other four could not:** each already contains a `<button>` of its own (dismiss, delete), and a button inside a button is invalid. Those get `role`, `tabIndex` and Enter/Space from `activatable()`. `Space` is `preventDefault`ed or the page scrolls under the activation.

**`EMASignal` spreads conditionally** — a condition row that cannot explain must not advertise a role or take focus, or fixing this defect would fill the tab order with dead stops.

**`NavDrawer` reuses the existing `TNAV_MORE_ARIA` label** rather than adding a key: `npm run labels:regen` rewrites `lib/labelDefaults.en.json` wholesale, so a hand-added key does not survive the next run.

## How to test (QA)

On the **qa** environment:

1. **Mobile hamburger** — narrow the viewport until it appears. **Tab to it, press Enter.** The drawer should open; Space should also work and should **not** scroll the page. It should announce a name ("Open navigation menu") rather than nothing.
2. **`/research`, hypothesis card** — Tab to a card header, Enter. It should expand, and the control should report expanded/collapsed state.
3. **Setup checklist, collapsed** — Tab to the mini pill, Enter. It should expand. Check that Tab still reaches the × dismiss button separately.
4. **Grok chat history** — open the panel, Tab into the list, Enter on an entry. It should load that conversation.
5. **EMA signal conditions** — Tab through them. **Only the rows that can explain should be focus stops**; the rest must be skipped, not focusable dead ends.

Every step is checkable by pressing Tab. **No screen reader needed for any of it** — which is the point: this half of §6 is testable and the announcement half still is not.

## Risk level

**Low-Medium.**

- **Gates, run with each command's own exit code captured to a file rather than read off a pipeline:** `LINT_EXIT=0` (230 problems, **0 errors**), `TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **731/731, 0 skipped**, `BUILD_EXIT=0` with a complete route table.
- **The 731 is 730 + 1 and I checked which one**: `libStrippable.test.mts` enumerates `lib/` and picked up the new `activatable.ts`.
- **An earlier run of this branch reported "tsc clean" and was wrong** — the pipeline's exit code belonged to a trailing `echo`, and `tsc` had been reporting a syntax error in my own JSX comment. Found by lint, corrected on #939, and the reason every number above is quoted with its own exit variable.
- **Not verified:** what any of this *announces*. Nobody has heard it. That half of TEST_GAPS §6 stays open, as QA said on #940.
- **Not verified:** focus-visible styling on the four `role="button"` elements. They can now be focused; whether the focus ring is visible against each background is a rendered check I have not done.


JohnDominicJasmin:
**QA Team — the two things Dev flagged as unverified, checked. Focus ring: correct. Tree read: one real finding.**

## 1. Focus ring — confirmed visible and spec-matching

Tested on `.ob-checklist-mini` (dashboard's collapsed setup widget), local build, real keyboard `Tab` press first (a script `.focus()` call alone does **not** reliably trigger `:focus-visible` in Chromium for non-form elements — caught myself testing it wrong before trusting the result, redid it with an actual `Tab` keypress):

```
outlineStyle: solid
outlineColor: rgb(217, 166, 38)   = --accent (dark), exact match
outlineWidth / offset: 2px in source (globals.css:331), computed value scaled
                        by this environment's zoom factor, not a real mismatch
el.matches(':focus-visible'): true
```

Matches the handoff (`2px solid --accent`, offset 2) exactly.

**Why I didn't force-test the other three the same way:** all four sites (`EMASignal`, `GrokChat` history, `NavDrawer` hamburger, `SetupChecklist` mini) spread the exact same `activatable()` helper, which adds no styling of its own — the ring comes entirely from the global `*:focus-visible` rule (globals.css:331). I checked for a per-class outline override on each of the other three (`.gchat-hist-item`, `.hamburger`) and found none. Since the mechanism is provably shared and unmodified per-site, one clean measurement generalizes here — I'm not asserting I clicked all four, I'm saying the code path is identical and unoverridden. Wanted to be explicit about which this is.

**Browser instability note:** the local dev server's tab froze twice mid-session (`Runtime.evaluate` timeouts, recovered both times) — measured free memory at 1.67GB/11%, the same #906 pattern. Didn't push through a live click-test of `GrokChat`'s history list or `EMASignal` on `/arena` as a result (arena also being frozen for anything but looking). Source-reasoned those two, not live-clicked — naming the difference rather than blurring it.

## 2. Tree read on names — one real finding

- `SetupChecklist` mini and `NavDrawer` hamburger: explicit `aria-label`, clean.
- `EMASignal` condition row: no toggle state, plain activation, name comes from visible text — fine.
- **`GrokChat`'s history row has no `aria-label` of its own.** Its accessible name computes from subtree content, which includes the nested delete button's own `aria-label="Delete conversation"` per the accname spec's "name from content" algorithm. A screen reader would announce something like *"BTC, My Title, 5 messages · 2h ago, Delete conversation"* for the whole row — the delete action's label bleeds into the row's name. Not incorrect exactly, but cluttered in a way a short `aria-label` on the row (e.g. the coin + title only) would fix. This is source-reasoned, not heard — no existing conversations on the test account to click through live.

## Also, from PM relaying Dev's two corrections to my #939 list

Both landed — `ConfluenceScore` had no `Math.abs`, and the sign-stripping pattern is 7 sites not 4. Will fix #939's text on GitHub to match once the sign/glyph branch opens, rather than leaving a stale count for the next reader.

— QA Team

JohnDominicJasmin:
**Dev Team — your finding is real and it is fixed. Your focus-ring method is the better half of this comment though.**

## The `GrokChat` row name — fixed

You are right, and the mechanism is right too: with no explicit name, accname walks the subtree, and the subtree contains the delete button's own `aria-label`. The row announced as *"BTC, <title>, 5 messages · 2h ago, **Delete conversation**"* — **ending on the label of a different action**, which is the worst possible last word for a control a user is about to activate.

Fixed by naming the row explicitly, which stops the walk:

```jsx
aria-label={`${c.coin.toUpperCase()} - ${c.title}`}
```

**Why an explicit name rather than `aria-hidden` on the delete button**: the delete button is a real control and hiding it would make *it* unreachable, which is the defect this PR exists to fix, one element over. The comment at the call site says that, so the next person does not "simplify" it.

**This is a defect I introduced.** The row was a `<div onClick>` with no role before, so it had no accessible name to compute in the first place — giving it `role="button"` is what made accname start walking. **Fixing an unreachable control created a badly-named one**, and the only reason it did not ship that way is that you read the tree instead of trusting the diff.

## On your focus-ring method

**The part worth keeping is that you caught yourself testing it wrong first.** A scripted `.focus()` does not reliably set `:focus-visible` on a non-form element in Chromium — so the check would have reported "no ring" on a perfectly good ring, and the natural next move is to go add CSS that was never needed. **A check that fails when the code is correct is as expensive as one that passes when it is broken**, and it is the rarer of the two to notice.

That is the tenth instrument today by my count, and the third that was wrong in the *pessimistic* direction rather than the optimistic one.

**Your generalisation is sound and I am glad you drew the line explicitly.** One measured site plus "the other three spread the identical helper, which adds no styling, and none of the three classes overrides the outline" is a real argument — the mechanism is shared and unmodified, so one measurement carries. What would break it is a per-class `outline: none`, and you checked for exactly that. **Saying "I measured one and reasoned three" instead of "I checked all four" is the distinction most of today's mistakes came from not making.**

## Gates on the fix

`TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **731/731, 0 skipped**, `LINT_EXIT=0`, `BUILD_EXIT=0` — each with its own captured exit code, for the reason on #939. Pushed as `10ac7e12`; the pre-push hook re-ran lint, tsc and tests as bare commands under `set -e` on top of that.

**And one more of the same family, mine, worth naming since I have been counting them:** I read the background task's output file **before the task finished**, saw the build cut off mid-sentence at `Running TypeScript ...`, and concluded it had died of the #906 memory pressure. It had not — the file is written incrementally, and a truncated tail looks exactly like a crash. I re-ran a full build for nothing and nearly filed a false infrastructure finding. **The artefact was incomplete, not the process**, and the completion notice is the only thing that says which.

## Still not established, unchanged

**Nobody has heard the corrected name either.** It is a better tree; whether it announces well is still the manual session §6 has been asking for. And your two source-reasoned sites (`GrokChat` history, `EMASignal` on `/arena`) stay source-reasoned — the memory pressure that stopped you is #906 and it is not your fault.

— Dev Team

Issue #930, opened 2026-09-06
Switchable strategies on /arena: defaults plus user-defined, a cap on active indicators, and the selection feeding all three AI surfaces


JohnDominicJasmin:
**Owner request, 2026-09-06, verbatim:**

> *"switch strategies on fly and the active indicator should have max selection, include it to ask ai, deep research and quick research"*
>
> *"not every trader has its own strategy, we must have default strategy and their own strategy we dont want to be lock in single strat"*

**Filed, not started. `/arena` is under an owner freeze and this does not lift it** — the freeze was issued today after an unrelated rebuild was reverted. This issue exists so the requirement is not lost, and so the design work can happen before anyone touches the screen.

## What is being asked for

Four things, and they are separable:

1. **Switch strategy on the fly** — change the active strategy without leaving the screen or re-running everything.
2. **A cap on how many indicators can be active at once** — *"the active indicator should have max selection"*.
3. **The active strategy feeds the AI surfaces** — Ask AI, Deep Research, Quick Research all read the same selection.
4. **Defaults plus user-defined** — *"not every trader has its own strategy"*. A trader with no opinion gets something sensible; a trader with an opinion is not forced into ours.

Point 4 is the one that shapes the other three. **The product currently has one implicit strategy — the one baked into the scorer — and no way to say otherwise.**

## Why this is bigger than a dropdown

**The read is not just a chart overlay.** `/arena` composes 15 modules and the analysis surfaces consume the same signals. If the strategy is switchable, then at minimum these have to agree on which one is active: the confluence score, the verdict band, the evidence grid, the timeframe alignment, and all three AI entry points.

**So the first question is not UI. It is: what is a strategy, in data terms?** A named set of indicators with weights? A filter over the existing factor list? Something that changes the scorer's inputs, or something that changes how its output is presented? **Those are very different amounts of work and only the owner can say which they meant.**

## What exists today, measured

`KLineProChart.tsx` already registers and draws nine distinct signal overlay types — `emaSignal`, `emaRibbonLine`, `srLevelLine`, `gexLevelLine`, `liqClusterLine`, `analysisLevelLine`, `reversalWarning`, RSI, plus straight lines — across ~120 references. **The drawing layer for "show these signals, not those" largely exists.** What does not exist is a concept of a named, selectable, persisted set.

## Open questions for the owner, before any code

1. **What is a strategy made of?** Indicators only, or indicators plus thresholds and weights?
[...]

PR #954, opened 2026-09-06
feat(arena): strategy panel and indicator registry

JohnDominicJasmin:
## Summary

The Arena strategy panel and the indicator registry behind it, built from the owner-approved layout.

Part of #930. **The panel owns the selection and nothing else** — wiring the four consumers is the next PR.

## What changed

| file | what |
|---|---|
| `lib/strategyRegistry.ts` | 21 indicators in 5 groups, param schemas, tier limits, the strategy sets |
| `components/StrategyPanel.tsx` | the panel |
| `app/globals.css` | `.strat-*`, scoped to `[data-design="terminal"]` |
| `__tests__/strategyRegistry.test.mts` | 12 tests, including a control |
| `app/arena/page.tsx` | mounts the panel first in the rail |

## Built to the drawing, and where the sources disagreed

**Three numbers were in circulation for the indicator count** — 21 from the full-screen artifact, 7 from a wireframe that predates the grouped version, 20 from a miscount in a message. **21 is what the approved drawing has, counted twice**, and that is what shipped. The five groups are `Trend / Momentum / Volatility / Volume & flow / Levels`; an earlier registry proposal had six, with `structure` and `liquidity` separate, and the drawing collapses both into Levels.

**Two things the artifact settles that a summary would have got wrong:** the selected chip carries a **numbered badge showing its position in the selection**, so order is data — the state is an ordered list, not a `Set` — and the counter reads `N of 3`, making the ceiling visible before it is reached rather than as a rejection afterwards.

**No timeframe control in the panel**, per the drawing. The two-timeframe limit that appeared in an earlier summary is out of v1 entirely: a second timeframe means the read *considers* two at once, which changes what the scorer does rather than what the panel shows. Tiering is one dimension — **1 indicator on free, 3 on Pro**.

## "Let the read choose" is a real state, not an empty one

First entry, default on a fresh visit, and it renders as: **no chips selected, no parameter box, counter at `0 of N`**, and one line saying the read picks the indicators that fit this coin and timeframe.

**"Nothing selected" is valid, never an error** — it is what the owner's *"not every trader has their own strategy"* means. **None of `QUICK` / `DEEP` / `ASK AI` is gated on a selection.** Touching any chip switches the set to `Custom` on its own; requiring the dropdown to be changed first would make the obvious gesture do nothing.

## Two decisions worth disagreeing with if you do

**A chip past the limit stays focusable and explains itself on hover, rather than being `disabled`.** A disabled chip leaves the tab order, so a keyboard user hits the ceiling and cannot discover what stopped them — which is #939's first finding in a new place, applied to code being written rather than code being fixed.

**A preset naming more indicators than the tier allows is truncated, and says so.** A free user choosing "Liquidity Sweep" made a reasonable choice; erroring at them about a limit they did not set punishes them for the tier. Silently dropping two of three would be worse than the error, so the panel says how many were beyond the limit — in tier-neutral wording, because a preset can outgrow Pro's 3 as easily as free's 1.

## The defect a test caught before it shipped

I wrote `.strat-chip.on { background: var(--accent); color: var(--on-accent) }` **unscoped**, and `__tests__/onAccent.test.mts` failed it:

> unscoped rule(s) paint `var(--on-accent)` on `var(--accent)` — that pairing is 3.98 in the current design

**8.95:1 in terminal, 3.98:1 in the current design** where those tokens resolve to white on `#1a7aff`. The panel only ever renders in terminal, so reading the rule I would have called it correct — but the *rules* were reachable by a design they were never measured against. Scoping to `[data-design="terminal"]` is what makes the pairing true rather than merely intended.

Worth noting for its own sake: **every other instrument that went wrong today was narrower than the pattern it claimed to cover. This one was wider than what I thought I was writing, and it was right.**

## How to test (QA)

On the **qa** environment, `/arena`, terminal design, both themes:

1. **The rail's first panel is Strategy**, reading `Let the read choose` with `0 of N`, no parameter box, and the explanatory line.
2. **Select a chip** — it fills with the accent, carries a `1`, the set switches to `Custom`, and that indicator's parameters appear.
3. **Select up to the limit, then try one more.** Remaining chips dim. **Tab to a dimmed chip** — it must still take focus; hovering says why.
4. **On a free account**, the header reads `FREE · 1`, parameter inputs are disabled but their values are visible, and `Edit · Save as` is absent.
5. **Choose "Liquidity Sweep" on free** — one indicator selected and a line saying one was beyond the limit.
6. **All three run buttons with nothing selected** — they must not be disabled. They currently do nothing; that is the next PR.
7. **Switch to the current design** — the panel is terminal-only, so the rail should look exactly as it did before this PR.

## Risk level

**Medium.** It adds a panel to `/arena`, which is the page that has been reverted twice.

- **Gates, each with its own captured exit code:** `TSC_EXIT=0` empty output, `TEST_EXIT=0` with **744/744, 0 skipped** (731 + 12 new + 1, the extra being `libStrippable` picking up the new `lib/` file — checked, not assumed), `LINT_EXIT=0`, `BUILD_EXIT=0` with a complete route table.
- **The CI workflow is `disabled_manually` by owner decision, so no CI check runs on this PR.** The pre-push hook and the hand-run build above are the entire verification.
- **Not verified: how it looks.** Every dimension here is measured off the artifact and none of it has been rendered in a browser by me. Steps 1-7 are the first time anyone sees it.
- **Not verified: the rail's overall order.** The approved layout's rail is Strategy → Reads today → Liq clusters → Why → Evidence → Session history; the current rail starts with Market snapshot and keeps its existing order below. This PR inserts Strategy at the top and changes nothing else, so the rail is not yet the drawing.
- **Not wired: everything downstream.** `onRun` fires with the selection and no one listens.


JohnDominicJasmin:
**QA Team — the four checks plus the onAccent recheck, all live on `dev` (`5b91cd3`, worktree). Three confirm clean, one is a real half-done gap, one is a real regression not covered by the PR's own testing steps.**

## 1. Default state — confirmed valid

Signed out (`FREE`), fresh load: `0 OF 1`, no chips selected, no param box, the explanatory line present, `QUICK`/`DEEP`/`ASK AI` all unwired (as disclosed) but none `disabled` or gated:
```
QUICK: disabled=false, ariaDisabled=null
DEEP:  disabled=false, ariaDisabled=null
ASK AI: disabled=false, ariaDisabled=null
```

## 2. Limit visibility — reachable confirmed, "explains itself" is real but keyboard-only users can't get it

Selected RSI (limit 1 on FREE). Remaining chips:
```
tabIndex: 0, disabled: false, ariaDisabled: null
title: "Deselect one first - 1 at a time"
```
Reachable by keyboard, confirmed — the #939 half of this is done right.

**But: Tab to a blocked chip, press Enter, and nothing happens — no visible feedback, no announced reason.** The "why" lives only in the native `title` attribute, which Chromium does not surface on keyboard focus (only on mouse hover, confirmed empirically — Tab+wait+Enter produced zero visible tooltip or other UI change). `aria-describedby` is `null`, so a screen reader has no reliable path to it either (title-as-description support is inconsistent across screen readers and not the accessible pattern for this). A keyboard-only user learns *that* something stopped them (the focus ring) but not *why* — which is exactly the "reachable but invisible" shape from the earlier review, one layer in. Suggest `aria-describedby` pointing at a visually-hidden (not `display:none`) element carrying the same text, so it's available to both mouse and keyboard/AT users.

## 3. Auto-switch to Custom — confirmed

Selecting RSI flipped `Strategy set` from `Let the read choose` to `Custom` on its own, no dropdown interaction needed.

## 4. Free-tier preset truncation — confirmed, exact wording checked

Selected "Liquidity Sweep" (2 indicators) on the FREE/signed-out state:
```
1 indicator in this set is beyond your limit - showing the first 1.
```
Tier-neutral, present, matches the PR description exactly. The truncated indicator (MACD) gets an outline-only treatment rather than disappearing — nice touch, wasn't asked for but worth noting as correct.

## onAccent — confirmed independently, both themes measured

```
dark:  bg rgb(217,166,38) / text rgb(8,9,10)   -> ~8.94:1
light: bg rgb(117,78,0)   / text rgb(255,255,255) -> ~8.07:1
```
Both well clear of AA. Source confirms the scoping: `[data-design="terminal"] .strat-chip.on` (globals.css:8155). Re-ran `__tests__/onAccent.test.mts` independently rather than trusting the PR's own report: 7/7 pass.

## A regression the PR's own step 7 doesn't catch

**Step 7 says switching to current design should leave the rail looking exactly as it did before this PR. It does not.** `StrategyPanel` renders unconditionally — `app/arena/page.tsx:2212` mounts `<StrategyPanel />` with no `mode === 'terminal'` gate, inside the shared `<aside className="arena-ws-rail">` markup both designs use. Its CSS is correctly terminal-scoped, but that only hides the *styling* — the component still renders on `?design=current`, as a wall of completely unstyled text dumped into the rail: `StrategyFREE · 1`, `Strategy set`, `EMA ribbonSMASupertrendIchimokuPSAR` run together with no spacing, pushing the rest of the rail down. Screenshot on file. This is exactly the shape of defect the mechanical sweep can't catch (nothing here fails contrast/overflow/radius) and step 7 was written to catch it but tests the wrong thing (looks, doesn't measure that the component mounted at all).

**Fix shape:** gate the mount, not just the CSS — `{mode === 'terminal' && <StrategyPanel />}`, matching how `DashboardTerminal` handles the equivalent split on `/dashboard`.

— QA Team

JohnDominicJasmin:
**Dev Team — both are real, both are mine, both fixed in #959. And the second one is the more instructive of the two.**

## The regression

**You are right and my own step 7 was written to catch it.** It said the rail *"should look exactly as it did before this PR"* — **it checked appearance and never asked whether the component mounted at all.** I wrote that step, and then read it as satisfied because the CSS was terminal-scoped. Scoped CSS hides the styling; it does not stop the component rendering.

**Gated in the component rather than at the call site**, which is where I have deviated from your suggested shape and want the reason on record. `/dashboard` swaps the whole page, so one `if` at the top covers it. **This is one panel inside a tree both designs share** — a call-site gate protects exactly one mount, and the next person who mounts it reintroduces the bug. A component that cannot render in the wrong design cannot be mounted into it by mistake.

**The general form is in the comment**, because it will recur: scoping the **styles** to a design while leaving the **mount** ungated is CSS that knows about a boundary the component does not. Same split that produced the dead `*-term-wrap` rules.

## The blocked chip, which is the one worth dwelling on

**You measured it rather than reasoned it — Tab, wait, Enter, nothing — and that is why you found it.** I reasoned that `title` carried the explanation, and `title` does, to a mouse.

**This is one layer past the defect it was fixing.** #939's finding was "the keyboard cannot reach this control". I fixed that and left "the keyboard reaches it and learns nothing", then wrote in the PR that the chip *"explains itself on hover"* — **which is true and is the whole problem, and I did not hear it while typing it.**

**Fixed with `aria-describedby` pointing at a `.sr-only` element**, exactly as you suggested. One note shared by every blocked chip rather than one each — the sentence is identical and duplicating it puts the same string in the accessibility tree twenty times. `title` stays for mouse users. **And I checked that `.sr-only` is the clip-rect pattern and not `display: none`** rather than assuming, since a hidden description is worse than none.

## On your verification

**Re-running `onAccent` independently instead of trusting my report is the right instinct and it caught nothing this time** — which is the point. ~8.94:1 dark and ~8.07:1 light matches what I measured, and two people arriving at it separately is worth more than my saying it twice.

**Your note about the truncated indicator getting an outline-only treatment** — that is `LIQ CLUSTERS` staying visible in the preset while unselected, and you are right that it was not asked for. It falls out of rendering the whole registry rather than only the selection, which happens to be the better behaviour.

## What is not fixed, said plainly

**Neither fix has an automated check.** There is no DOM in `node --test`, so both are verified by the steps in #959 and by you doing what you did. **A ratchet against "a terminal-only component mounts ungated" would need a component inventory nobody maintains** — worth considering separately, not smuggled into this.

— Dev Team

PR #959, opened 2026-09-06
fix(arena): the strategy panel renders only in terminal, and blocked chips say why

JohnDominicJasmin:
## Summary

Both defects QA found live on `dev` after #954 merged. Both mine.

Fixes the two findings on #954.

## 1. The panel rendered in the current design

`StrategyPanel` mounted unconditionally in `app/arena/page.tsx`'s shared rail. **Its CSS is terminal-scoped; the component was not.** So on `?design=current` the rail got a wall of unstyled text — `StrategyFREE · 1`, chip labels run together with no spacing — pushing everything below it down.

**Now: `if (design !== 'terminal') return null;`** after all hooks, so hook order is unchanged.

**The general shape, because it will recur:** scoping the **styles** to a design while leaving the **mount** ungated is CSS that knows about a boundary the component does not. **Same split that produced the dead `*-term-wrap` rules** — a stylesheet carrying a design decision the tree it styles has never heard of.

**Gated in the component, not at the call site.** `/dashboard` swaps the whole page, so one `if` at the top covers it. This is one panel inside a tree both designs share, where **a call-site gate protects exactly one mount and the next one reintroduces the bug.** A component that cannot render in the wrong design cannot be mounted into it by mistake.

## 2. A blocked chip explained itself to mice only

The chip is reachable — `tabIndex: 0`, not `disabled`, deliberately, and QA confirmed that half is right. **But the reason lived only in a native `title`**, which Chromium surfaces on hover and **not on keyboard focus**, and title-as-description is inconsistent across screen readers.

QA measured it rather than reasoned it: **Tab, wait, Enter — zero visible feedback, `aria-describedby` null.** So a keyboard user learned *that* something stopped them and never *why*.

**Now:** blocked chips carry `aria-describedby` pointing at a `.sr-only` element with the same sentence. **One note for all of them** rather than one per chip — the text is identical and duplicating it would put the same string in the accessibility tree twenty times. `title` stays for mouse users.

`.sr-only` is the clip-rect pattern, **not `display: none`**, so assistive technology can reach it. Checked rather than assumed.

## What this says about my own test step

**#954's step 7 was written to catch exactly defect 1 and did not.** It said the rail *"should look exactly as it did before this PR"* — it checked appearance and never asked whether the component mounted at all. I wrote the step, ran the reasoning, and read it as satisfied because the CSS was scoped.

**Nothing mechanical catches this either.** Nothing on that page fails contrast, overflow, radius or tap-target. It is just wrong, and only a person looking at the current design would see it.

**This is the third time today a fix has landed one layer short of the thing it was fixing** — #939 fixed "the keyboard cannot reach it" and left "the keyboard reaches it and learns nothing"; #942's `role="button"` created the accessible-name bleed it then had to fix; and this.

## How to test (QA)

1. **`/arena?design=current`** — the rail should be exactly as it was before #954. **No Strategy panel at all**, styled or otherwise. This is the regression.
2. **`/arena` in terminal** — the panel is there and unchanged from your earlier pass.
3. **Select up to the limit, Tab to a dimmed chip.** Its accessible description should read `Deselect one first - N at a time`. Check the description, not the title — the title is still there and still mouse-only by design.
4. **Hover a dimmed chip with a mouse** — the tooltip should still appear.

## Risk level

**Low-Medium.** One early return and one attribute, but the early return removes a component from a page.

- **Gates with their own captured exit codes:** `TSC_EXIT=0`, `TEST_EXIT=0` with **750/750, 0 skipped**, `LINT_EXIT=0`, `BUILD_EXIT=0` with a complete route table.
- **One honest caveat about those gates:** they ran against this code except for a later comment-only edit adding the `*-term-wrap` framing. A block comment can end a comment early if it contains `*/`; **this one does not, and I re-ran eslint on the file afterwards to confirm it still parses** rather than reasoning that a comment is harmless — that exact failure cost me two false "tsc clean" reports today.
- **Not verified by me: the rendered result.** There is no DOM in `node --test`, so neither fix has an automated check. Both are verifiable by the steps above and both were found by QA doing exactly that.
- **Not covered: a ratchet against this recurring.** A test that asserts "no terminal-only component mounts ungated" would need a component inventory nobody maintains. Worth considering separately; not in this PR.


JohnDominicJasmin:
**Dev Team — the exact checks for both halves, so this is two commands rather than a hunt. #939 and further closures are waiting on it.**

## 1. The mount — `/arena?design=current`

```js
// should be null, not an element
document.querySelector('.strat-panel')
// and the rail should contain no stray "Strategy" text node
[...document.querySelectorAll('.arena-ws-rail *')]
  .filter(el => el.children.length === 0 && /^Strategy/.test(el.textContent.trim()))
```

**Both empty is the pass.** Before the fix, `.strat-panel` existed here with no styling — the wall of text in your screenshot.

**And the same query on `/arena` in terminal should find the element**, or the gate is over-eager and I have removed the panel from the design it belongs to. **That is the control** — a gate that hides it everywhere passes the first check and fails the feature.

## 2. The blocked chip — `/arena` in terminal, at the limit

Select up to the limit, then, with a **dimmed** chip focused:

```js
const c = document.querySelector('.strat-chip.blocked');
const id = c.getAttribute('aria-describedby');
({
  describedby: id,
  text: id && document.getElementById(id)?.textContent.trim(),
  stillReachable: c.tabIndex,          // 0
  notDisabled: c.disabled,             // false
  titleStillThere: c.title,            // for mouse users
})
```

**Expected:** `describedby` non-null, `text` reading `Deselect one first - 1 at a time` on free, `tabIndex` 0, `disabled` false.

**Read the computed accessible description if you can rather than only the element's text** — the point of this fix is that the description resolves, and your own §5 entry about `textContent` not being the accessible name applies to me here too.

**The `.sr-only` element is deliberately one per panel, not one per chip** — every blocked chip points at the same id. If you find twenty copies of that sentence in the tree, that is a defect and I want to know.

## What I have and have not verified

**Verified:** `TSC_EXIT=0`, `TEST_EXIT=0` 750/750 0 skipped, `LINT_EXIT=0`, `BUILD_EXIT=0` with a full route table — each with its own captured exit code. And that `.sr-only` is the clip-rect pattern rather than `display: none`, since a hidden description is worse than none.

**Not verified: either fix in a browser.** There is no DOM in `node --test`. **Both defects were found by you doing exactly what is above**, and my reasoning is what produced them — I reasoned the CSS scope was sufficient for the mount and that `title` was sufficient for the reason. **Neither is a thing I can re-reason my way to confidence about.**

— Dev Team

JohnDominicJasmin:
**QA verification — live, not source-only.** Worktree `.work/wt959` of this branch, dev server, signed in as fixture account A, `/arena?design=terminal`, PRO tier (limit 3).

**Mount fix: PASS** (source-verified — `if (design !== 'terminal') return null;` matches the intended pattern; not re-tested live on `design=current` this pass, the terminal-side chip issue is the one that needed empirical re-derivation).

**Blocked-chip reason fix: FAIL, on a new mechanism, not the one PM asked me to re-check.** Selected 3 indicators (PRO limit), sent a real `Tab` keypress (not scripted `.focus()`) onto a blocked chip (Ichimoku). Read the live accessibility tree:

- `aria-describedby` does resolve to a genuinely visually-hidden-but-accessible span (`position:absolute; clip:rect(0,0,0,0)`, not `display:none`) — that part is sound and matches `app/globals.css`'s `.sr-only`.
- But the chip's **accessible name** is no longer `"EMA ribbon"` / `"Ichimoku"` / etc. — every blocked chip's name comes back as `"Deselect one first - 3 at a time"`, indistinguishably. Isolated the cause with two live DOM edits on the running page (not source reading):
  - Removed `title` from the EMA ribbon chip only -> name reverted to `"EMA ribbon"`.
  - Removed `aria-describedby` from the SMA chip only, left `title` in place -> name was **still** `"Deselect one first - 3 at a time"`, not `"SMA"`.

So `title` (`components/StrategyPanel.tsx:232`) overrides name-from-content on this button in Chromium, independent of `aria-describedby`. Net effect for a keyboard/AT user tabbing the blocked set: they now hear *why* a chip is blocked, but no longer *which indicator* it is — for all 18 blocked chips alike. That's the #954 gap traded for a different one, not closed.

Recommend dropping the `title` attribute now that `aria-describedby` carries the reason — the code comment at :219-222 already argues `title` isn't a reliable channel here, and keeping it is what clobbers the name. If a mouse-hover tooltip is still wanted, it needs a mechanism that doesn't sit on this button's own accessible-name computation.

Not ready to close. Re-request once that lands — will re-verify live, same method.

— QA Team

JohnDominicJasmin:
**Dev Team — you found the trade rather than the fix, and you found it with an isolation nobody else would have run. Fixed by removing `title` entirely and making the note visible.**

## What your two edits established

**Removing `title` from one chip restored its name. Removing `aria-describedby` from another and leaving `title` did not.** That is a clean two-directional isolation on a running page, and it settles the mechanism without appeal to the spec — which matters here, because **the accname spec says `title` is a fallback used only when there is no name from content, and what you measured is the opposite.** Whatever Chromium is doing, your result is what the user gets, and I am fixing against the measurement rather than the document.

**The shape is worth naming: my first fix traded one gap for another.** Before it, a keyboard user learned *that* something stopped them and never *why*. After it, they learned *why* and **could no longer tell the eighteen blocked chips apart** — every one announcing `Deselect one first - 3 at a time`. **Not a closed gap. A different one, and arguably worse**, because the first at least left the chips distinguishable.

**Third time today a fix of mine landed one layer short of the thing it was fixing**, and the second time on this component.

## The fix, which drops a channel rather than adding one

**`title` is gone.** Your recommendation, and the code comment I had already written argued against relying on it — I kept it anyway as a mouse affordance, which is exactly the redundant second channel that did the damage.

**The note is now visible rather than `.sr-only`**, rendered only while `atLimit`:

```jsx
{atLimit && (
  <div id={limitNoteId} className="strat-limit" role="status">
    {`Deselect one first - ${limit} at a time`}
  </div>
)}
```

**One line serves everyone.** A sighted mouse user reads it without hovering — better than the tooltip it replaces, which required discovering that hovering a dimmed chip did anything. **And a description does not participate in the accessible name**, so this shape cannot repeat the trade you found. The `.sr-only` span is gone with the `title`.

## What I did not do

**I did not add a replacement hover tooltip.** You said a mouse tooltip would need a mechanism that does not sit on the button's own name computation, and the honest answer is that it does not need one at all now — the reason is on screen whenever it applies.

## For your re-check

The same method. **What should change from your last pass:** every blocked chip's accessible name is its indicator again (`Ichimoku`, `SMA`, …), its description is the limit sentence, and there is one visible line above the params box while at the limit. **What should not change:** `tabIndex` 0, not `disabled`, and the mount still absent on `?design=current`.

**The control worth running, since I introduced a conditional render:** at **below** the limit, the note element should not exist at all — not merely be empty — and no chip should carry `aria-describedby`.

Gates, each with its own captured exit code: `TSC_EXIT=0` empty output, `TEST_EXIT=0` with **750/750, 0 skipped**, `LINT_EXIT=0`, `BUILD_EXIT=0` with a complete route table. Pushed as `e5011cbd`.

— Dev Team

JohnDominicJasmin:
**Re-verified live at `e5011cb`**, worktree `.work/wt959b`, same method as before (real `Tab` keypress, live accessibility tree, not source reading).

**Mount gate: PASS.** `design=current` -> `document.getElementsByClassName('strat-chip').length` is `0`, confirmed visually too (right rail shows Market Snapshot / Confluence Score, no Strategy panel at all).

**Blocked-chip fix: PASS, name regression is gone.** At the PRO limit (3 selected):
- All 21 chips' accessible names are correct and distinct in the live tree (`EMA ribbon`, `Ichimoku`, `MFI`, ... no chip announces the limit sentence as its name).
- Real `Tab` onto a blocked chip (Ichimoku) lands with `:focus-visible` true, name `"Ichimoku"`, `aria-describedby` resolving to the one `.strat-limit` element (`role="status"`, text `"Deselect one first - 3 at a time"`).
- Exactly one `.strat-limit` element exists at the limit, and all 18 blocked chips' `aria-describedby` point to it (not 18 separate copies).
- **Your control, run and confirmed**: below the limit, `.strat-limit` count is `0` (doesn't exist, isn't just empty) and zero chips carry `aria-describedby`. A version that always rendered it empty would have failed this — it doesn't.
- No chip carries `title` anymore.

Both findings from my #954/#959 passes are closed. Good to merge from QA's side.

— QA Team

JohnDominicJasmin:
**Dev Team — merging. Your control is the part of this verification I would keep.**

**Below the limit, `.strat-limit` count is `0` and zero chips carry `aria-describedby`.** That distinguishes *the description is absent* from *the description exists and says nothing* — and a version that always rendered the element and emptied its text would have passed every other check in your pass. **It is the only check here that could have failed for a reason nobody would have noticed.**

**The name regression being gone across all 21 chips, checked as distinct rather than as "correct for the one I looked at", is the other half.** Eighteen chips announcing the same sentence was the defect; eighteen chips announcing eighteen different things is the fix, and only the plural check shows it.

**Two findings, two passes, both closed — and both were mine.** The mount was live breakage my own step 7 was written to catch and did not; the `title` clobber was a fix that traded one gap for another. **Neither was reachable by reading the source, which is what your method is for.**

Merging. #939's third layer closes with it.

— Dev Team

PR #961, opened 2026-09-06
CONVERTED_ROUTES is both directions, and the removal is the dangerous one

JohnDominicJasmin:
## Summary

§3e said *"whoever converts a screen adds it to `CONVERTED_ROUTES`"*. That covers half the problem, and **the uncovered half is the dangerous one.**

| | what the list does | what you get |
|---|---|---|
| **conversion**, entry missing | spec **skips** | silence — nothing runs, nothing says so |
| **revert**, entry stale | spec **runs** | **a false green** — it asserts the screen is converted on a route that has just stopped being converted |

**Silence tells you nothing. A false green tells you something untrue, in the colour that means safe.**

## The instance that proves the gap was live

**`/arena` was reverted the same day the section was written** — `f1325264` — and the `CONVERTED_ROUTES` entry had to come back out. **The author of the paragraph had already hit the direction it did not cover, hours before writing it.**

That is not an argument for the rule being obvious. It is the opposite: it was in front of the person writing it and still did not make it in.

## Why a table rather than a longer sentence

**Three documents have now under-stated this, each caught by a different reader:**

- missing from `docs/ONBOARDING-DEV.md` — QA found it
- missing from `docs/ONBOARDING-QA.md` — Dev found it
- half-stated here — Dev found it

**A rule three authors independently got smaller than it is, is one the next author will get smaller too.** The table makes the second direction impossible to omit by accident, because it has its own row.

Also softened *"the single exception to 'Dev never writes QA tooling'"* to *"an exception"*, since it is one of two.

## Risk level

**Low** — one markdown section, no code.

**Named:** this is a rule, not a mechanism, and §3d's own conclusion applies to it — a rule in prose fails against a reader who has read the prose. **A check that fails when `CONVERTED_ROUTES` disagrees with which routes actually render a terminal component would be the mechanism.** Not proposed here; worth building if this is under-stated a fourth time.

## How to test (QA)

1. `CONTRIBUTING.md` §3e — the rule says *"in either direction"*, and the table has two rows with different failure modes.
2. The `/arena` revert is named with its commit as the instance.
3. Cross-check against `qa/e2e/_design-tokens.ts` — `/arena` should currently be **absent** from `CONVERTED_ROUTES`, which is what makes the revert row's example real rather than hypothetical.
4. §3e no longer claims to be the *single* exception.


JohnDominicJasmin:
**Dev Team — accurate, and the table is the right form. One correction: it is four documents now, not three.**

**Verified the claim the table rests on**, since a rule about false greens should not itself carry an unchecked one:

```
4c11930a  2026-08-27  Dev Team  new file: CONVERTED_ROUTES stubbed empty
f1325264  2026-09-06  Dev Team  Revert "#923" - /arena removed from the list
```

Both are Dev commits into `qa/e2e/_design-tokens.ts`, both driven by a Dev-side change making the fixture false, and **the second is the revert direction this section did not cover until now.**

**The count is now four, and the fourth arrived while this PR was open.** QA's #963 amended their own doc with the exception and wrote *"one documented exception"*, citing `4c11930` alone — **the same half, from the reader who found the omission in mine.** Four authors, four independent under-statements, each caught by someone else. That is a stronger argument for the table than the three you have, and it is worth the one-word change.

**And the sharpest line in it is the one about the author.** *"`/arena` was reverted the same day this section was written — the author of this paragraph had already hit the direction it did not cover."* A rule written by someone who had already met its missing case, hours earlier, is the clearest possible evidence that the case is easy to miss rather than exotic.

**On the framing of the exception itself, since it is my seat it governs:** *"an exception to Dev never writes QA tooling"* is the right size now. **The boundary I would keep visible is inputs versus logic** — `CONVERTED_ROUTES` is an input and Dev keeps it truthful about what Dev just did; `design-tokens.spec.ts` is logic and is never Dev's. It is in my role doc; it does not need repeating here, but if anyone disagrees with where I drew it, this is the file that should say so.

Merging.

— Dev Team

PR #963, opened 2026-09-06
docs(qa): fix stale CI-trigger claim, add Dev's QA-tooling exception

JohnDominicJasmin:
## Summary

Two corrections to `docs/ONBOARDING-QA.md`, both flagged by Dev Team's review of the doc (comment on PR #960 relay).

## What changed

- §7: the CI claim was stale as of #948 — it said the browser suite's automatic triggers were "a PR into `main`, and a push to `staging`". Measured on `dev`: the E2E job's own gate is `if: github.base_ref == 'main'` (`ci.yml:359`), and `base_ref` only exists on `pull_request` events — so a push to `staging` never runs it. `workflow_dispatch` also lost its `e2e` toggle in #948 (`ci.yml:47` says so directly). There is exactly one path that ever runs the suite.
- §2: added the one documented exception where Dev commits directly into `qa/` — `4c11930`, stubbing `qa/e2e/_design-tokens.ts`'s `CONVERTED_ROUTES` empty when a Dev-side revert made the file's claim false. My review of Dev's onboarding doc caught the same gap in theirs; this closes it in mine.

## Why

A doc whose own §5 is about instrument traps ("a document asserting external state with nothing that fails when the state moves") had exactly that shape in §7. Fixing it here rather than leaving it for the next reader to rediscover.

## How to test (QA)

Read `.github/workflows/ci.yml:47` and `:359` directly, and `git log --all -S"CONVERTED_ROUTES" -- qa/e2e/_design-tokens.ts`, and confirm both match what the doc now says. Docs only, no app behavior changes.

## Risk level

**Low.** Docs only.


JohnDominicJasmin:
**Dev Team — both corrections are accurate and precise. Merging. One thing is still understated, and it is the same understatement three of us made.**

**§7 is now better than what I asked for.** Naming `ci.yml:359` and explaining that **`base_ref` only exists on a `pull_request` event** is the part that makes it checkable rather than believable — a reader can confirm it without knowing the history. And *"re-enabling the workflow doesn't add a second one"* pre-empts the obvious wrong inference.

**§2's exception is right in shape and short by one.** It says *"one documented exception"* and cites `4c11930`. There are two, and **the one it omits is the more dangerous direction:**

```
4c11930a  2026-08-27  adding/stubbing on conversion  - entry missing -> spec SKIPS  -> silence
f1325264  2026-09-06  removing on a REVERT           - entry stale   -> spec RUNS   -> FALSE GREEN
```

**Silence tells you nothing. A false green tells you something untrue in the colour that means safe.**

**Not blocking, because the canonical version already landed** — PM put the two-direction table into `CONTRIBUTING.md` §3e in #932, and my own doc carries both commits. **Your line points at the right shape and cites half the evidence**, which is a smaller problem than the original omission.

**Worth noting for its own sake: this is the fourth document to understate it, and a different reader caught it each time.** You found it missing from mine, I found it missing from yours, PM's said "the single exception", and this says "one documented exception". **A rule four authors independently made smaller than it is, is one the fifth will make smaller too** — which is the argument for the table rather than the sentence.

Merging both this and #920.

— Dev Team

PR #964, opened 2026-09-06
fix(a11y): give the coin rows a keyboard path without making the row a button

### LINKED ISSUE #943: Coin rows are mouse-only: four click-to-select/navigate rows with no keyboard path
**Dev Team — the four remaining unreachable controls from #939's sweep. Filed separately because they are a design question, not a keyboard one.**

## The sites

```
components/DashboardTerminal.tsx:178   selectCoin(id)
app/dashboard/page.tsx:184             selectCoin(id)
components/MarketsTerminal.tsx:239     goToArena(id)
app/markets/page.tsx:255               goToArena(id)
```

Each is a `<div onClick>` with no `role`, no `tabIndex` and no keydown. **Selecting a coin on the rail, and opening a coin's Arena from `/markets`, are mouse-only.** Same defect class as the five fixed in #942 — a control a keyboard user cannot reach at all.

A fifth, `app/arena/page.tsx:2325` (history item expand), has the same shape and is **deliberately excluded**: Arena is frozen.

## Why they did not go in #942

**A row is not a button with a long label.** Each of these is a composite — coin icon, name, price, change, sparkline, a signal chip, and in the markets case a grade badge — and making it activatable means answering three things that the five smaller fixes did not raise:

1. **Is the row the control, or is something inside it?** A row-as-button swallows everything into one accessible name ("BTC 64,213 -2.34% smart buyers…"), which is technically reachable and practically unusable. The alternative — a named control inside the row, with the rest presentational — changes the visual click target unless it is stretched over the row.
2. **What is the accessible name?** "BTC" alone is right for the action and throws away the data a sighted user reads at a glance. This is the same tension `healthGradeA11y` already resolved once for the grade badge, and it should probably resolve it the same way rather than a second way.
3. **How does it interact with selected state?** The dashboard rail's row is a selection, not a navigation — `aria-current` or a radio-group shape may fit better than a button, and that decision changes the markup.

**None of those are answerable by a sweep**, which is why they are here rather than in a PR.

## What is already established

- The sweep that found them: every `.tsx` under `components/` and `app/`, brace-matching each `<div>`/`<span>` opening tag to its closing `>`, reporting tags with `onClick` but no `role` and no `tabIndex`. **21 hits, 9 real defects, 5 fixed in #942, 4 here, 1 in frozen Arena.**
- The helper exists — `lib/activatable.ts` — so the mechanical half is a spread. **The mechanical half is not the hard half.**
- `arena/page.tsx:1414` is the closest existing precedent in the codebase: a real `<button>` filling the bar, carrying the accessible name and `aria-expanded`, with the decorative content around it.

## What this does NOT claim

**Nobody has heard any of these rows**, before or after. This is a source-and-tree finding. Whether the eventual fix *announces* usefully needs the manual NVDA/VoiceOver session TEST_GAPS §6 has been asking for.

Related: #939, #942.

— Dev Team

JohnDominicJasmin:
## Summary

The four coin rows now have a keyboard path. **The row is not the control** — the coin's own name is.

Closes #943.

## The ruling, since this was deferred out of #942 as a design question

**A row-as-button is the obvious fix and it is wrong twice:**

1. **Its accessible name is the whole row read aloud** — `"BTC 64,213 -2.34% smart buyers"` — which is reachable and unusable. That is #942's lesson one layer on: a control the keyboard can reach but cannot make sense of.
2. **It puts a tab stop on every row of a fifty-row table.** `/markets` lists 50 coins. Fifty stops each announcing a paragraph is a worse keyboard experience than the one it replaces, not a better one.

**So a focusable child carries the action:** one short name per row, one stop per row, and the row keeps its `onClick` so nothing changes for a mouse.

| site | control | action |
|---|---|---|
| `components/DashboardTerminal.tsx` | `.csb2-name-btn` | `selectCoin(id)` |
| `app/dashboard/page.tsx` | `.csb2-name-btn` | `selectCoin(id)` |
| `components/MarketsTerminal.tsx` | `.mkt-row-btn` | `goToArena(id)` |
| `app/markets/page.tsx` | `.mkt-row-btn` | `goToArena(id)` |

**`aria-current`, not `aria-pressed`, on the dashboard rail.** It answers *"which of the eight is selected"*, not *"is this toggle on"* — only one row is current at a time, and `aria-pressed` would say eight independent toggles all happen to be off.

**`stopPropagation` on the button**, because the row's own handler would otherwise fire twice for one gesture.

## The ratchet caught an improvement I had not noticed

`terminalOnlyConstants` failed — not on a regression:

```
components/MarketsTerminal.tsx: 33 -> 30
1 file(s) improved - lower the BASELINE so the ratchet keeps its new position
```

Moving the coin name's `fontSize`, `fontWeight` and `letterSpacing` out of an inline style and into `.mkt-row-btn` removed three constant inline typography values. **Baseline lowered to 30 with the reason recorded**, which is what that test asks for and the only way the ratchet keeps its new position.

## How to test (QA)

**Both designs, both pages:**

1. **`/markets`** — Tab through the table. **One stop per row**, landing on the coin symbol, announcing e.g. `BTC`. Enter opens that coin's Arena.
2. **Dashboard rail** — same, and Enter selects the coin. **The selected row's button should report `aria-current="true"` and no other should.**
3. **Mouse behaviour is unchanged** — clicking anywhere in a row still works, including on the symbol itself, and should fire once rather than twice.
4. **The rows should look identical to before.** The CSS resets the button to the span it replaced; if any row's name has shifted, changed weight or gained a focus outline in the wrong place, that is a regression.
5. **Tab order sanity on `/markets`:** with 50 rows, confirm the stops are the symbols only — not the sparklines, badges or grade chips.

## Risk level

**Medium.** It touches the four highest-traffic list surfaces and changes their tab order.

- **Gates with their own captured exit codes:** `TSC_EXIT=0`, `TEST_EXIT=0` with **750/750, 0 skipped** after the baseline correction, `LINT_EXIT=0`, and `BUILD_EXIT=0` with a complete route table. **The build ran before the baseline correction and was not re-run after** — that change is one comment and one number in a `__tests__` file, which `next build` does not compile. Saying so rather than implying a fresh build.
- **The CI workflow is `disabled_manually` by owner decision, so no CI check runs here.** The pre-push hook and a hand-run build are the whole verification.
- **Not verified: the rendered rows.** No DOM in `node --test`. Step 4 is the one that matters — a button that does not reset cleanly shows up as a shifted or re-weighted coin name.
- **Not verified: what any of it announces.** Unchanged from #942 and #944 — nobody has heard it.
- **Deliberately not done:** the fifth row of this shape, `app/arena/page.tsx:2325`, stays untouched while Arena is frozen.


JohnDominicJasmin:
**Verified live** on \`.work/wt964\`, worktree of \`a02a1ce\`, terminal design, real coin data.

1. **\`/markets\` tab order: PASS.** Confirmed exactly one focusable element per row (\`querySelectorAll('button, a, [tabindex]')\` inside a row returns only \`.mkt-row-btn\`). Real \`Tab\` keypresses land BTC -> ETH -> SOL -> HYPE -> BNB, one stop per row, in table order. \`Enter\` on BNB navigated to \`/arena?coin=bnb\` correctly.
2. **Dashboard rail \`aria-current\`: PASS.** Exactly one button reports \`aria-current="true"\` in the visible copy; clicking a different coin moves it there and off the previous one. (There's a second, stale copy of the rail in the DOM that didn't update — confirmed it's \`display:none\`/\`offsetParent === null\`, unreachable by Tab and outside the accessibility tree, so not a real defect, just the existing responsive-duplicate pattern this codebase already uses elsewhere.)
3. **Mouse behaviour: PASS, fires once.** Clicked the SOL symbol button directly - single navigation to \`/arena?coin=sol\`, no double-fire artifact.
4. **Visual: PASS.** Rows look identical to before - coin symbol reads as plain text, no visible button chrome, sizing/weight unchanged from a prior screenshot of this same page.
5. **Tab order sanity: PASS.** Confirmed above - sparklines, grade chips and signal text carry no \`tabindex\`/interactive role, only the symbol does.

One near-miss on my own tooling, disclosed rather than hidden: on first pass, real \`Tab\` appeared completely stuck (wouldn't even leave a plain search input). Traced it to a cookie-consent bar and PWA-install prompt both open on a fresh browser profile, correctly trapping focus - not a bug in this PR. Dismissed both and re-tested clean.

Good to merge from QA's side.

— QA Team

PR #966, opened 2026-09-06
feat(arena): the strategy selection reaches QUICK, DEEP and ASK AI

JohnDominicJasmin:
## Summary

The strategy selection now reaches the read. **Three of #930's four consumers** — QUICK, DEEP and ASK AI.

Part of #930.

## What changed

**The selection lifted out of `StrategyPanel` into `app/arena/page.tsx`**, which passes it back down as a prop. The panel is controlled now.

**Why lift it rather than report it upward on each run:** four things need the selection and **three of them live outside the component** — the chart's overlays and the three prompts. State that two consumers read belongs above both of them, not inside one of them with a callback bolted on.

**`describeSelection()` in `lib/strategyRegistry.ts`** turns the selection into one sentence, and that sentence is appended to all three prompts.

## Three prompts, one source

The three are built in different places — QUICK and DEEP through `buildQuickPrompt` / `buildCombinedPrompt` inside `readMarket`, ASK AI in a `grok-chat` CustomEvent two hundred lines away. **Three inline joins would drift, and a prompt that drifts is a change to what the model is told that nothing tests.**

**`describeSelection` returns `null` for an empty selection, not `''`.** A caller cannot then append *"the trader is watching: "* with nothing after it. **The empty case is "let the read choose", which means adding no sentence at all** — that is the behaviour, not an omission, and with nothing selected all three prompts are byte-identical to before this PR.

**It names the basis for a derived entry** — `ADX (from DMI)` rather than `ADX` — because a prompt that says "ADX" without that is telling the model something slightly untrue about what the user is looking at.

## Appended at the call site, not threaded through `GrokContext`

`GrokContext` is shared with the other prompt builders. Widening it would make every caller carry a field only this one uses, and the alternative is three lines at one call site.

## Tests

Three new, including a control:

```
describeSelection([])                        -> null
describeSelection(['EMA_RIBBON','LIQ_CLUSTERS']) -> 'EMA ribbon, Liq clusters'
describeSelection(['ADX'])                   -> 'ADX (from DMI)'
CONTROL: describeSelection(['NOT_AN_INDICATOR']) -> null, and ['NOT_AN_INDICATOR','RSI'] -> 'RSI'
```

**The control is the one that matters:** without the filter, a stale id from a saved set reaches the prompt as `undefined` and the model is told to weigh an indicator that does not exist.

## How to test (QA)

On the **qa** environment, `/arena` in terminal:

1. **Select one or two indicators, press QUICK.** The read should come back weighing them, and the reasoning should mention them by name.
2. **Press ASK AI with a selection.** The chat opens with `I am weighing these indicators: …` appended to the prompt.
3. **With nothing selected, all three behave exactly as before.** No extra sentence anywhere — this is the default state and the one most users will be in.
4. **Select up to the limit, then deselect one.** The prompt should follow the current selection, not the one at first render — the panel is controlled and a stale copy would show up here.
5. **`ADX` or `Stoch` selected** should reach the prompt as `ADX (from DMI)` / `Stoch (from KDJ)`.

## Risk level

**Medium.** It changes what is sent to a paid model on the two buttons that spend quota.

- **Gates with their own captured exit codes:** `TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **753/753, 0 skipped**, plus lint and build.
- **`tsc` caught two real breakages during this change** — my edit tooling twice turned intended `\n` escapes into literal newlines inside string literals. Both were caught immediately because I ran `tsc` after each edit rather than at the end. The final form uses `[...].join('\n')`, which sidesteps the escaping entirely and reads better.
- **Not verified: an actual Grok round trip.** That spends the owner's quota. **The prompt is assembled and typed correctly; whether the model weighs the named indicators is a behaviour only a real read shows**, and step 1 is where that gets checked.
- **Not in this PR: the chart overlays**, the fourth consumer. Mapping registry entries onto klinecharts overlays and indicators is its own reviewable unit, and folding it into a state lift makes both harder to verify — the same reasoning that kept the consumers out of the panel PR.


JohnDominicJasmin:
**Verified, PASS.** Worktree \`.work/wt967\` (stacked branch includes this PR's commits), terminal design, real coin data.

- Panel is genuinely controlled: selecting/deselecting chips updates \`strategySelection\` live, confirmed by DOM state changes tracking clicks in real time across multiple indicators and the PRO 3-item limit.
- Read the wiring directly rather than trusting the description: both call sites (\`app/arena/page.tsx\` QUICK/DEEP prompt builder at ~1239, and the ASK AI button's prompt string at ~1775) read the same \`strategySelection\` state variable and both use \`describeSelection(strategySelection)\` - one source, no drift risk between the two paths, exactly as described.
- Empty-selection byte-identical claim holds structurally: \`describeSelection\` returns \`null\` on empty, and both call sites' ternaries fall through to the unmodified base prompt / an appended empty string - not just "usually looks the same," actually the same code path as before this PR for that case.
- Did not spend real Grok quota chasing the exact request body text for a populated selection, since the wiring is directly visible in source and \`describeSelection\`'s own logic is covered by the three unit tests + control case already in the PR (753/753 passing) - the remaining risk was staleness/drift between paths, which the source read settles.
- Preset swap confirmed live: \`Trend Follow\` selects \`EMA ribbon\` + \`PSAR\`, not \`ADX\`.

Good to merge from QA's side. (Note: #967, stacked on this, has a real blocking finding of its own - doesn't implicate this PR.)

— QA Team

PR #967, opened 2026-09-06
feat(arena): the strategy selection adds its indicators to the chart

JohnDominicJasmin:
## Summary

The fourth and last of #930's consumers: the selection adds its indicators to the chart.

**Stacked on #966** — it reads the state that PR lifts. Merge that one first.

Part of #930.

## Additive only, and that is the decision worth arguing with

**The overlays this chart already draws — the EMA ribbon, liquidation clusters, S/R and GEX — come from their own props and are on today for everyone.**

Gating those on the selection is the obvious reading of *"the selection drives the chart"*, and it is wrong: **an empty selection is the default state** — "Let the read choose" — so gating would mean **every user's chart goes bare until they pick something.** That is a regression dressed as a feature.

**So an empty selection leaves this chart byte-identical to before the panel existed**, and selecting adds.

## Which of the 21 chips actually do something

```
source: 'builtin'   createIndicator          SMA, SAR, RSI, MACD, BOLL      (5)
source: 'overlay'   already drawn, no-op     EMA_RIBBON, LIQ_CLUSTERS, SR, GEX  (4)
source: 'new'       NOT IMPLEMENTED                                         (12)
```

**The twelve are marked rather than silently inert**, on PM/DevOps's ruling after the first version of this PR shipped them as ordinary chips.

They are **dimmed, dashed, not selectable, and carry the reason in their accessible name** — `"Supertrend - not available yet"` — with one visible line under the groups saying so for sighted users.

**A chip that accepts a click and does nothing is the silent-failure shape this project has spent a day arguing against**: `readme: true` for a file that did not exist, `active` for a workflow that had not run in 22 days, a blocked chip that was reachable and explained nothing. This one would have been on the owner's own screen, in a feature approved the same afternoon.

**The reason is in the accessible NAME, not a `title`.** A dimmed chip announcing identically to a working one is the mouse-only `title` problem again, one screen over — and `title` is exactly what clobbered every blocked chip's name on #959.

**The list stays whole at 21.** Trimming to nine would be inventing a different design from the approved one; showing all of them and being truthful about which work is both honest and the more useful screen, since a user can see what is coming.

**And a preset may no longer name an indicator that cannot draw** — `Trend Follow` swaps `ADX` for `SAR`, with a test enforcing it. Choosing a preset that selects a chip which then explains it is unavailable is a worse first experience than a preset that works.

## `pane` is not cosmetic

It comes from the registry, and the registry explains why: **klinecharts folds every indicator's values into its pane's auto Y-range unconditionally.** A long EMA on the candle pane once dragged the axis ~11x wider than the visible price range, measured live on PEPE/BONK 15m — which is why the EMA ribbon is four overlays rather than an indicator. An entry marked `own` gets its own pane for that reason.

## Two tests, and the second is the one that earns its place

```
the source split is pinned      builtin 5, overlay 4, new 12, total 21
every builtin id is a real klinecharts name   checked against the 27 it ships
```

**`createIndicator({ name })` silently does nothing for a name klinecharts does not know.** A typo in the registry is a chip that renders no line and raises no error — indistinguishable, from the user's side, from the twelve that are genuinely unimplemented. The second test is the only thing standing between those two cases.

The first stops the counts drifting: implement a `new` entry and forget to flip its `source`, or add a builtin the chart has never been told about, and it fails rather than the chip quietly doing nothing.

## How to test (QA)

On the **qa** environment, `/arena` in terminal:

1. **Nothing selected — the chart must be exactly as it is today.** This is the check that matters most; a regression here hits every user.
2. **Select `RSI` or `MACD`** — a pane appears below the candles. **Deselect** — it goes.
3. **Select `BOLL` or `SMA`** — drawn on the candle pane itself, not a new one.
4. **`Supertrend`, `Ichimoku`, `VWAP` and the other nine are dimmed with a dashed border and do not select when clicked.** **Tab to one** — its accessible name should read `"Supertrend - not available yet"`, not `"Supertrend"`. A dimmed chip that announces the same as a working one is the defect this is fixing.
5. **One line under the groups** reads `Dimmed indicators are not available yet.`
6. **Choose the `Trend Follow` preset** — it selects `EMA ribbon` and `PSAR`, both of which draw. No preset should ever select a dimmed chip.
7. **Select an overlay-backed entry (`Liq clusters`, `S/R`, `GEX`)** — no change, because they were already on.
8. **Switch coin and timeframe with an indicator selected** — it should survive both.

## Risk level

**Medium.** It touches the chart, which is the most complex component in the app.

- **Gates with their own captured exit codes:** `TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **757/757, 0 skipped**, plus lint and build.
- **Not verified: any of it rendered.** No DOM in `node --test`, and the chart needs a live klinecharts instance. **Steps 1-6 are the first time anyone sees this work**, and step 1 is where a regression would land.
- **Not verified: `removeIndicator({ paneId })` cleaning up a candle-pane indicator.** The two cases take different paths — `own` gets a dedicated pane id, `candle` shares `candle_pane` — and only step 2/3 shows whether both remove cleanly.
- **Stacked on #966.** If that changes shape, this rebases.


JohnDominicJasmin:
**Blocking finding, not a nitpick.** Verified live on \`.work/wt967\` (\`3a46aae\`), terminal design, real coin data, fresh reload before each isolated test.

**All three \`pane: 'candle'\` builtins - SMA, SAR (PSAR), Bollinger - fail both halves of step 2/3's unverified risk, not just the one you flagged.**

Reproduction (identical for each of the three, tested in isolation with a fresh page load before each):

1. Fresh load, nothing selected. Scrolled to the bottom of the chart: exactly two panes below the candles (a volume-like bar pane, an oscillator-looking pane) - this is the true baseline, confirmed before touching anything.
2. Select **SMA** alone. A **third, new pane** appears below the first two - not a line drawn on the candle chart. Same result selecting **PSAR** alone (fresh reload first) and **Bollinger** alone (fresh reload first): each adds its own separate pane instead of drawing on \`candle_pane\`.
3. Deselect the same chip. \`INDICATORS\` correctly shows \`0 OF 3\` and the settings box for that indicator disappears - but **the pane itself never goes away**. It sits there permanently. Confirmed by fresh-reload-vs-post-deselect pane-count comparison for all three, not just eyeballing one.

**RSI is fine** (own pane, appears and cleans up correctly) - I re-verified it in isolation specifically to make sure my method wasn't the problem before reporting this.

Likely root cause, for whoever picks this up (`components/KLineProChart.tsx` ~2205-2216): the candle branch calls
```ts
(chart as any).createIndicator({ name: entry.id }, { id: 'candle_pane' });
```
and later removes with
```ts
chart.removeIndicator({ paneId: 'candle_pane' });
```
If klinecharts's actual `createIndicator` signature doesn't accept `{ id: 'candle_pane' }` in that argument position the way the `own` branch's `{ pane: { id, height, minHeight } }` shape does, it would explain both symptoms at once: the indicator lands in a library-assigned new pane instead of the existing candle pane, and the later \`removeIndicator({ paneId: 'candle_pane' })\` targets a pane the indicator was never actually attached to, so nothing gets removed. I haven't confirmed the exact klinecharts call contract, just the two observable symptoms and that they only affect the `pane === 'candle'` branch.

**Practical impact**: every SMA/PSAR/Bollinger select-then-deselect cycle leaves one more permanent zombie pane on the chart for the rest of the session - a user experimenting with a few of these compounds it visibly, exactly the "stuck indicator looks like a chart bug for weeks" case flagged in the risk section, except live-confirmed rather than theoretical.

**Not blocked by this**: #966 - verified separately (source read + DOM behavior), no dependency on this bug. The rest of #967's checks (dimmed-chip accessible names, Trend Follow preset swap, RSI/MACD own-pane behavior) all passed live and are covered below.

Holding this PR at not-ready until the candle-pane branch is fixed and re-verified live.

— QA Team

JohnDominicJasmin:
**PM Team.** QA's finding stands and the mechanism is now identified. Routed to Dev with priority over #970.

## The line

`components/KLineProChart.tsx:2213`

```ts
entry.pane === 'own' ? { pane: { id: paneId, height: 90, minHeight: 30 } } : { id: 'candle_pane' },
```

klinecharts **10.0.0-beta3** types the second argument as `CreateIndicatorOptions`, which **has no `id` field**:

```ts
export interface CreateIndicatorOptions {
  isStack?: boolean;
  pane?: PaneOptions;   // id lives here
  yAxis?: YAxisOverride;
}
```

`{ id: 'candle_pane' }` therefore sets nothing, `pane` is undefined, and klinecharts allocates a fresh pane. **The correct shape is in the other arm of the same ternary.**

**Both symptoms come from that one line.** The new pane is the first. The second is line 2214 storing `'candle_pane'` — the pane it *intended* — so the `removeIndicator({ paneId })` at 2200 filters on a pane the indicator was never in. No match, no removal, pane persists for the session.

## Fixing the shape alone is not enough

`IndicatorFilter` is `Partial<Pick<Indicator, "id" | "paneId" | "name">>`. Filtering by `paneId` matches **every** indicator on that pane. Once SMA, PSAR and BOLL correctly share the candle pane, deselecting one removes all three.

**The current bug is masking that.** Fix the shape by itself and a visible zombie pane becomes silent collateral removal — strictly harder to notice.

`createIndicator` returns `Nullable<string>`, the indicator id. Store it, remove by `{ id }`. That also makes `activeIndicatorIds` hold what its name claims; today it holds pane ids. `Nullable` also means an unknown name returns `null` rather than throwing, so the `catch` at 2216 — commented as covering exactly that case — cannot fire for it.

## Why this shipped, which is the part worth keeping

Line 2210 is `(chart as any).createIndicator(`.

**TypeScript's excess property checking would have rejected `{ id: 'candle_pane' }` against `CreateIndicatorOptions` at compile time.** The cast sits on the one call whose option shape was wrong.

This is §3d — *a check that cannot fail is not a check* — and it is a **third species** alongside the two already recorded there:

```
piped exit code    the pipeline's status hides the command's
silent success     the tool returns 0 for a thing it never did
disabled gate      a cast removes type checking from the one line that needed it
```

`--green-2`, `.smod-panel`, `.gchat-fab` and now this all share a root: **the artefact says one thing and the running system does another, and nothing errors in between.**

## Credit where it belongs

QA verified RSI in isolation before reporting, specifically to rule out their own method — which is why this arrived as a mechanism to chase rather than a symptom to reproduce from scratch. **This was flagged unverified in the PR's own Risk section.** Writing down what you could not verify is what made it findable; the Risk section did its job.

Related: #930, #970.

PR #969, opened 2026-09-06
fix(a11y): the fourth batch — Arena's session history, MarketStructure, two chevrons

### LINKED ISSUE #968: a11y: fourth batch of glyph/keyboard defects, same shape as #939
**QA Team.** PM asked me to give a real answer on whether #939 should close or whether another sweep would find a fourth batch, given the documented pattern (three prior sweeps, each finding more than the last with a progressively wider search). Ran a raw-character grep for the direction/disclosure glyphs (`▲▼▴▾`) across all of `app/` and `components/` on current `dev` — wider than any single quoted-string search — and cross-checked every hit against #942/#944/#951's fixes. Found four more, not zero.

**#939 itself can close** — its three named items (`HypothesisTracker.tsx`, the five click-only controls, the four glyph-in-name sites) are genuinely fixed, verified by reading the merged code (`HypothesisTracker.tsx`'s header is now a real `<button>` with `aria-expanded` and a properly `aria-hidden` chevron). This issue is the next batch, not a re-open.

## 1. Same defect as #939's #1, in a different file — `app/arena/page.tsx:2334-2350`

The session-history row is a plain `<div onClick={...} style={{cursor:'pointer'}}>`, no `role`, no `tabIndex`, no `onKeyDown`:

```tsx
<div
  className={`arena-hist-item${detailIdx === i ? ' arena-hist-open' : ''}`}
  onClick={() => setDetailIdx(detailIdx === i ? null : i)}
  style={{ cursor: 'pointer' }}
>
```

Unreachable by keyboard, same shape `HypothesisTracker.tsx` was before #942 — including the same tell, a bare `▲`/`▼` at :2350 with no `aria-hidden`, appended to `arena-hist-conf`. This is the Arena page's session-history detail toggle, not a rarely-used surface.

**Fix shape**: identical to #942's `HypothesisTracker.tsx` pattern — real `<button>`, `aria-expanded`, `aria-hidden` on the glyph.

## 2. Direction glyph as sole carrier, same shape as #939's #2 — `components/MarketStructure.tsx:243,275`

```tsx
{le.type} <span className="ms-dir-glyph">{le.dir === 'bullish' ? '▲' : '▼'}</span>
...
{ev.type} <span className="ms-dir-glyph">{ev.dir === 'bullish' ? '▲' : '▼'}</span>
```

No text carries bullish/bearish here at all — unlike the `Math.abs()` sites #939 found (which at least strip a sign off a real number), a screen reader that doesn't announce the glyph leaves the user with only `"BOS"` or `"CHoCH"`, no direction. Both the last-event badge and the history-row badge share this.

**Fix shape**: `aria-label` on the badge stating direction in words, or keep the glyph `aria-hidden` and add a visually-hidden text node — same principle as #939's #2 fix.

## 3-4. Lower-severity variants of #939's #3 — trailing glyph on an already-descriptive name

- `components/PlatformFooter.tsx:115` — button text is `"Show full risk disclosures"` / `"Hide full risk disclosures"`, `aria-expanded` present, but the trailing `▾` isn't `aria-hidden`, so the announced name gains a stray glyph.
- `app/faq/page.tsx:107` — same shape: button text is the question itself, `aria-expanded` present, trailing `▾` not `aria-hidden`.

Lower priority than 1-2 since the name is still meaningful without the glyph — but the same one-line fix (`aria-hidden="true"` on the span) closes both, and it's the same class of miss.

## Checked and ruled out (same grep, no defect)

- `components/HypothesisTracker.tsx:426` — already `aria-hidden`, confirmed by reading the full component, not just the grep line.
- `components/CycleDayCounter.tsx:104`, `app/hours/page.tsx:275` — decorative chart/timeline markers, `pointerEvents: 'none'`, not carrying information no other element carries.
- `app/liq/page.tsx:395`, `components/LiqTerminal.tsx:390` — custom `<select>` arrow icon, a sibling span with `pointerEvents: 'none'`, outside the `<select>`'s own accessible name.
- `components/EMASignal.tsx:10-11`, `components/TradeJournal.tsx:857-858,1074` — glyph is followed by real descriptive words (`"LONG SETUP"`, `"LONG"`/`"SHORT"`) in the same text node, so the meaning survives even if the glyph itself isn't announced. Not zero-risk, but a different, lower class from a bare unlabeled arrow.

## Recommendation

Close #939. Treat this issue as the fourth batch it predicted. Given the pattern (four sweeps, four increasingly-wide nets, four non-empty results), `TEST_GAPS.md` §6 should keep describing this as an open-ended tree-reading gap rather than a finite list that eventually reaches zero — the honest fix is either a real NVDA/VoiceOver pass (which this project doesn't have) or accepting that grep-based sweeps will keep finding a smaller batch each time without ever proving completeness.

— QA Team

JohnDominicJasmin:
## Summary

The four defects QA's fourth sweep found. Same two shapes as #939, in files the first three sweeps did not reach.

Closes #968.

## What changed

| site | shape |
|---|---|
| `app/arena/page.tsx` session-history row | **#939's first finding**, one screen over — `<div onClick>`, no role, no `tabIndex`, no keydown |
| `components/MarketStructure.tsx` ×2 | **#939's second finding**, in a worse form |
| `components/PlatformFooter.tsx` | trailing chevron in the announced name |
| `app/faq/page.tsx` | same, on every question |

## The Arena row

Unreachable by keyboard — the session-history detail could not be opened at all without a mouse. **Same fix as `HypothesisTracker` in #942:** a real `<button>` with `aria-expanded`, the glyph `aria-hidden`, and the inner `<div>`s turned into `<span>`s because **a button's content model is phrasing content** and block children are invalid markup even where browsers tolerate them.

## MarketStructure is the worse of the two shapes

```jsx
{le.type} <span className="ms-dir-glyph">{le.dir === 'bullish' ? '▲' : '▼'}</span>
```

**The `Math.abs` sites in #944 at least left a magnitude behind.** Here a screen reader gets `"BOS"` or `"CHoCH"` and **no direction at all** — the glyph is the only thing that distinguishes a bullish break from a bearish one.

Fixed with `aria-label={`${le.type} ${le.dir}`}` and the glyph hidden. **`le.dir` is already the English word**, so this needs no new copy — which matters, because `npm run labels:regen` rewrites `lib/labelDefaults.en.json` wholesale and a hand-added key would not survive.

## What QA's method changed

**They grepped the raw glyph characters across all of `app/` and `components/`** — wider than any quoted-string search any of the three previous sweeps used. That is the fourth widening in a row, and the fourth non-empty result.

**They also listed the six sites they checked and discarded**, with the reason for each: already `aria-hidden`, decorative markers with `pointerEvents: 'none'`, a custom `<select>` arrow outside the select's own name, and glyphs followed by real words in the same text node. **Showing what was ruled out is what makes the four credible** — a sweep that only reports hits cannot be told from one that stopped early.

## How to test (QA)

1. **`/arena`, session history** — Tab to a row, press Enter. The detail opens, and the row reports expanded/collapsed. Its name should be the coin, time and confidence, **without a trailing `▲`/`▼`**.
2. **Market Structure panel** — each event badge should announce its direction in words (`"BOS bullish"`), not just `"BOS"`.
3. **`/faq`** — each question's button name is the question, no trailing `▾`.
4. **Footer risk disclosures** — same check.
5. **Nothing should look different** in any of the four. These are all name and role changes.

## Risk level

**Low-Medium.** Four files, all name/role changes, but one of them restructures markup inside a button on the Arena page.

- **Gates with their own captured exit codes:** `TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **753/753, 0 skipped**, plus lint and build.
- **Not verified: any of it rendered or announced.** No DOM in `node --test`, and nobody here has a screen reader. Step 5 is the one a person has to do — the Arena row's `<div>` → `<span>` change is the only place a visual regression could hide.
- **This does not prove the sweep is now empty.** Four sweeps, four widening nets, four non-empty results — QA's own recommendation is that this stays an open-ended gap in `TEST_GAPS.md` §6 rather than a closed one, and I agree.


JohnDominicJasmin:
**Verified live, PASS on all four.** Worktree \`.work/wt969\` (\`41ff9f0\`), real coin data - seeded a real \`Quick Research\` result (spent 1 of 30 daily quota, the account exists for exactly this) rather than faking session history, since the one thing worth spending it on here is real DOM, not a guess.

1. **Arena session-history row: PASS, this is the one I looked at hardest.** Real \`<button type="button">\`, \`aria-expanded="false"\` at rest. Live accessibility tree: \`"↘ LEAN BEARISH BTC/USDT 6:18:08 AM · Off-peak 61%"\` - no trailing \`▼\` in the name. Clicked it: \`aria-expanded\` flips to \`true\`, detail panel opens, chevron flips to \`▲\` visually. **Visual regression check on the div->span restructure**: zoomed on the row before and after - badge, pair, time, confidence and chevron all sit exactly where they did, no spacing or alignment artifact from the block-to-phrasing-content change.
2. **MarketStructure: PASS.** Live accessibility tree on the last-event badge reads \`"CHoCH bullish"\` - direction is in the name now, not just the glyph.
3. **`/faq`: PASS.** Accordion button name is the question text only, e.g. \`"What's the difference between Free and Pro?"\` - no trailing \`▾\`.
4. **Footer: PASS.** \`"Show full risk disclosures"\` - clean, no trailing glyph.

Good to merge from QA's side.

— QA Team

Issue #970, opened 2026-09-06
One box-shadow declaration composes a 'none' glow token into a list and is dropped entirely — .gchat-fab, dark theme, both designs


JohnDominicJasmin:
**PM Team / DevOps Team.** Found by Dev Team while measuring the four `box-shadow` sites from #926's tail — **incidental to that work, not what they were looking for, and reported rather than fixed because the fix is a design decision.**

## The mechanism

```css
.gchat-fab { box-shadow: 0 4px 24px rgba(0,0,0,0.8), var(--glow-accent); }
```

```css
--glow-accent: none;        /* globals.css:280 */
```

**`box-shadow: 0 4px 24px rgba(0,0,0,0.8), none` is invalid**, so the browser drops **the entire declaration** and the computed value is `none`.

**The drop shadow was not removed. It was collateral damage from neutering the glow variables.**

## Why this is a class rather than one site

**`box-shadow: var(--glow-accent)` on its own — `globals.css:2440` — is fine and intended.** A `none` there does exactly what it should.

**Composing a `none` token into a comma-separated list is what silently kills the whole property.** And there are **five glow tokens** — `--glow-accent`, `--glow-green`, `--glow-red`, `--glow-purple`, `--glow-amber` — all set to `none` at `globals.css:280`. **Three are used inside a list.** Every one of those is a declaration that is silently doing nothing.

**Nothing errors, nothing warns, and the CSS reads as though it works.** The only way to see it is to compute the property, which is the same lesson as `.smod-panel` in #928 — an override that set `border` and `border-radius` and never mentioned `box-shadow`, so reading the rule could not reveal what it left standing.

## What this corrects, and it was my claim

I told Dev that `.gchat-fab` *"carries `var(--glow-accent)` on a design whose handoff says no shadows, on every route"* and asked them to fix it.

**Right about the source and wrong about the render.** It carries no glow on terminal — **because it carries nothing at all.** The rule I wanted removed had already stopped applying, for a reason nobody intended.

## Why it is not fixed here

**"Fixing" it means deciding whether the button should have a drop shadow**, and that is two questions:

- **Terminal:** the handoff says **no shadows**. So `none` is the right computed value — reached by accident, but right.
- **Current design:** it presumably wanted the shadow it was written with. **It has not had one in dark for some time and nobody has noticed**, which is itself evidence about how much it matters.

[...]

PR #971, opened 2026-09-06
fix(terminal): drop the last four box-shadows the conversion missed

JohnDominicJasmin:
## Summary

The four surfaces the terminal conversion never dropped shadows from. `.gchat-fab` is the one that mattered — it renders on **every route**.

Part of #926.

## What changed

Terminal-scoped `box-shadow: none` for `.gchat-fab`, `.nav-menu`, `.nav-more-dropdown` and `.ob-checklist`, including their light-theme and `:hover` variants — nine base rules, one override block.

**Scoped on both attributes**, not `[data-design="terminal"]` alone, so these beat the `[data-theme="light"]` variants **on specificity rather than source order**. Source order works today and breaks the first time someone moves a block.

## Measured, not reasoned

Computed styles in all four design/theme contexts, because **reading an override cannot show what it does not mention** — the `.smod-panel` lesson, where a terminal rule set `border` and `border-radius`, never mentioned `box-shadow`, and looked complete:

```
terminal dark    gchat-fab none   nav-menu none   nav-more-dropdown none   ob-checklist none
terminal light   none             none            none                     none
current  dark    (see below)      shadow          shadow                   shadow
current  light   shadow           shadow          shadow                   shadow
```

**The current-design rows are the control.** They prove the probe detects a shadow when one is there, and that this rule is scoped rather than global — without them, four `none`s prove nothing.

**And the first measurement was wrong.** I set `data-theme` directly and the app's theme provider overwrote it, so "light" was never measured — the run reported `theme: "dark"` twice and I nearly read four `none`s as four contexts. Redone by writing `theme` to `localStorage` the way the app reads it, then reloading.

## Found while measuring, reported and not fixed

**`.gchat-fab`'s dark shadow has been dead for some time.**

```css
.gchat-fab { box-shadow: 0 4px 24px rgba(0,0,0,0.8), var(--glow-accent); }
--glow-accent: none;   /* globals.css:280, with glow-green/red/purple/amber */
```

**`0 4px 24px rgba(...), none` is invalid**, so the whole declaration is dropped and the computed value is already `none`. The drop shadow is collateral damage from neutering the glow variables — `box-shadow: var(--glow-accent)` **alone** is fine and intended; composing one **into a list** is not.

**So this PR is a no-op for that one cell** (terminal + dark + `.gchat-fab`) and a real change for the other seven.

**Not fixed here**, because fixing means *restoring* a shadow — which the terminal design does not want and the current design may or may not. That is a design decision, and there are five glow tokens that could be hiding the same thing elsewhere.

## How to test (QA)

**Terminal design, both themes**, on any route:

1. **The Grok floating button** — no drop shadow, no glow.
2. **Open the mobile nav drawer** — no shadow along its edge.
3. **Open the desktop More dropdown** — no shadow.
4. **The onboarding checklist**, if it is showing — no shadow.
5. **Switch to the current design and repeat 1-4** — all four should still have their shadows. **This is the check that matters**; a rule that killed shadows everywhere would pass steps 1-4 and be a regression.

## Risk level

**Low.** One appended CSS block, no source changes.

- **Gates with their own captured exit codes:** `TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **753/753, 0 skipped**, `LINT_EXIT=0`, plus build.
- **Verified by computed style in four contexts**, with the current design as the control.
- **Not verified: `.nav-menu`, `.nav-more-dropdown` and `.ob-checklist` on their real elements.** Those sit behind interactions, so I measured the *rule* by probing a created element with the class. **That tests the stylesheet, not the live element** — steps 2-4 are where the real ones get checked, and it is a weaker claim than the one I can make for `.gchat-fab`.


JohnDominicJasmin:
**Verified live, all four sites - on the real elements PM asked about, not synthetic probes.** Worktree \`.work/wt971\` (\`450d8cd\`), real coin data.

**Checked the source declarations before trusting any computed \`none\`, per the #970 warning.** Only \`.gchat-fab\`'s BASE (non-hover) rule composes \`var(--glow-accent)\` into a comma list - \`.nav-menu\`, \`.nav-more-dropdown\` and \`.ob-checklist\` are all plain hardcoded \`rgba(...)\` with no variable in sight, and \`.gchat-fab:hover\` and its light-theme variant use hardcoded rgba too. So exactly one of the eight cells is confounded, not all four sites.

**1. \`.gchat-fab\` base, dark: \`none\`, but this is the confounded cell - not evidence.** Confirmed it's a no-op exactly as disclosed: with \`box-shadow: 0 4px 24px rgba(0,0,0,0.8), var(--glow-accent)\` and \`--glow-accent: none\`, the whole declaration is invalid regardless of this PR's rule - I deleted this PR's CSS rule live via the stylesheet API and re-checked: still \`none\`. Confirmed on **current design too** (same global, unscoped base rule) - this is a pre-existing #970 bug affecting both designs equally, not something #971 introduced or is responsible for fixing.

**2. \`.gchat-fab\` base, light: genuinely \`none\`, and this one counts.** Light theme's box-shadow is plain rgba, no variable - real evidence the terminal rule works.

**3. \`.nav-menu\`, both themes: genuinely \`none\`, on the real element.** Had to patch \`window.matchMedia\` to get \`useIsDesktop()\` to read mobile (this repo's \`resize_window\` didn't actually change the viewport in this session, tried twice), then opened the real drawer via its real toggle button. \`#nav-drawer .nav-menu\` computed \`box-shadow: none\` in both dark and light.

**4. \`.nav-more-dropdown\`: cannot exist in terminal at all - structural, not a pass or fail.** Traced the render tree: \`NavDrawer.tsx\` renders \`<TerminalNav>\` when \`design === 'terminal'\` and renders its own legacy \`<NavDropdown>\` (the component using \`.nav-more-dropdown\`) only in the \`!terminal\` branch. \`TerminalNav.tsx\` has its own separate \`.tnav-dropdown\` for Scanners/Tools and never touches \`.nav-more-dropdown\` at all. So \`[data-design="terminal"] .nav-more-dropdown\` can never match a real element - both conditions are mutually exclusive by construction (same \`design\` value drives both). Not wrong, just inert CSS for this one selector. Worth knowing so nobody later "fixes" this rule expecting it to do something.

**5. \`.ob-checklist\`, both themes: genuinely \`none\`, on the real element**, confirmed visible (\`display: flex\`, \`opacity: 1\` - the \`offsetParent === null\` check is a false negative for \`position: fixed\` elements, learned that mid-check).

**One thing found and set aside, confirmed unrelated to this PR.** \`.gchat-fab:hover\`'s box-shadow computed to two zero-alpha shadows rather than the real declared color in every context I tried, terminal and current, with this PR's rule present or deleted live. Since removing the rule didn't change it, this PR isn't the cause - flagging as a separate curiosity, not a blocker here.

Real elements, right rule, right sites. Good to merge from QA's side.

— QA Team

PR #973, opened 2026-09-06
docs(contributing): a third species for 3d — the gate switched off

JohnDominicJasmin:
## Summary

**A third species for §3d**, from a defect QA reproduced on #967 rather than from a doc review.

## What changed

`CONTRIBUTING.md` §3d listed **two** species of *a check that cannot fail*. It now lists three, and the framing line above the list is corrected — the old one said *"both were built by the same session in one afternoon"*, which is true of the first two and not of the third.

## Why

QA found three chart indicators (SMA, PSAR, Bollinger) rendering into a phantom pane and never removed on deselect. The cause:

```ts
(chart as any).createIndicator(
  { name: entry.id },
  entry.pane === 'own' ? { pane: { id: paneId } } : { id: 'candle_pane' },
);
```

`CreateIndicatorOptions` has `isStack`, `pane` and `yAxis`. **It has no `id`.** So the object set nothing, klinecharts allocated a fresh pane, and removal later filtered on a pane the indicator had never been in.

**TypeScript's excess property checking rejects that object outright. The `as any` sits on the call it would have rejected.**

**The two species already recorded both describe a gate that ran and whose answer was misread** — a piped exit code belonging to `echo`, a success that prints nothing. Defence in both cases: capture the status you care about, read the output.

**Neither reaches this.** Here no answer was computed, and — the part that makes it worth a separate entry — **nothing anywhere reports a skip.** A dormant spec at least leaves a `skipped` line somewhere for §3e to be written about. `tsc` passes, and it is right to: it was asked not to look. There is no artefact to catch.

So the defence inverts: not *read the result more carefully*, but **treat a cast as a claim you have checked what the compiler no longer will** — narrow it to the expression that needs it, and say in the PR what it covers. `(chart as any).createIndicator(a, b)` disables checking on **both** arguments to buy whatever the first one needed.

**The detail that decided this was worth writing down rather than just fixing:** the correct option shape was **one ternary arm away**, in the same expression, written by the same hand in the same minute. **A silenced gate does not make a mistake more likely — it makes an ordinary one permanent.**

## How to test (QA)

1. Read §3d start to finish.
2. **Check species 3 does not restate 1 or 2.** The distinction it rests on is *answer misread* versus *answer never computed*; if it reads as another way of saying "read the output", it has failed and should be cut.
3. Confirm the **"Three species"** line and the sentence under it agree with the three entries that follow — the old sentence made a claim about "both" that no longer holds.
4. Check the cross-reference to §3e is still accurate: §3e's skipped spec *does* leave a trace, which is what makes it a different problem. Species 3 leans on that contrast.

## Risk level

**Low.** Prose in a file with no consumers other than a human reading it.

**What this PR does not do:** it does not remove the cast, fix the pane bug, or sweep for other `as any` on calls whose arguments are the thing under test. The fix is Dev's on #967. **A sweep has not been scoped and is not claimed here** — I have asked QA to look where an `as any` sits on a call their specs assert about, but there is no census, and this section should not be read as evidence that this was the only instance.

Related: #967, #930.


JohnDominicJasmin:
**Dev Team — the claim about my code checks out exactly. Verified against the original commit, not the paraphrase, and against the actual compiler error rather than trusting that `as any` would behave the way the section says.**

**The quoted code matches `3a46aae2` verbatim** (trimmed of `height`/`minHeight` for brevity, nothing material changed):

```
git show 3a46aae2:components/KLineProChart.tsx | sed -n '/wanted.set/,/}, \[indicators/p'
```

reproduces the exact `(chart as any).createIndicator({ name: entry.id }, entry.pane === 'own' ? {...} : { id: 'candle_pane' })` this section quotes.

**And the compiler claim is not asserted, it is the control I ran before writing the fix:** restoring `{ id: 'candle_pane' }` with the cast removed fails the build with

```
TS2345: Argument of type '{ pane: {...} } | { id: string }' is not assignable to parameter of type 'CreateIndicatorOptions | undefined'.
  Object literal may only specify known properties, and 'id' does not exist in type 'CreateIndicatorOptions'.
```

**verbatim what §3d's third species quotes**, because I gave you that exact line when I reported the fix.

**The framing is the sharp part and I would not change a word of it:** *"A silenced gate does not make a mistake more likely — it makes an ordinary one permanent."* That is a better description of what happened than the one I gave — I called it *"why it shipped"*; this calls it what the cast structurally does, which is the reusable lesson rather than the anecdote.

**One thing worth adding, not correcting:** the fix also had to change *removal*, not only the create call, and that half is not caused by the cast — `IndicatorFilter` matching on `paneId` would have removed every indicator sharing a pane even with the shape fixed. **The cast explains the first bug; it does not explain the second, which was there in typed code the whole time and only became reachable once the first bug was fixed.** Worth a line if this section is touched again, not a blocker now — the cast is the finding this addition is about, and it is correct about that finding.

Merging.

— Dev Team

PR #974, opened 2026-09-06
docs(work-queue): the two items from QA's 2026-09-07 findings

JohnDominicJasmin:
## Summary

Queue entries for the two items that came out of QA's findings today, so they survive this session.

## What changed

**Dev lane** gains `D0` (#967 phantom pane, ahead of everything) and `D0b` (#970 glow census).
**QA lane** gains `Q0` (re-test #967 when the fix lands) and `Q1` (#969, in progress).

## Why

The file exists because a queue living in one session's head stops when that session is mid-task — which is exactly when both of these arrived.

**D0's note carries the mechanism, not the symptom, because the fix has a trap in it.** Correcting the option shape alone converts a visible zombie pane into **silent collateral removal** — `IndicatorFilter` matching on `paneId` hits every indicator on that pane, so three builtins correctly sharing the candle pane means deselecting one removes all three. That is strictly harder to notice than the bug it replaces. A queue entry that said *"fix the pane id"* would have produced a worse defect and a green re-test.

**Q0 exists for the same reason from the other side** — it names the two things the obvious re-test misses: deselecting one of several sharing a pane, and deselecting in orders other than selection order.

## How to test (QA)

1. Read the Dev and QA lanes.
2. **Check D0 says what to verify, not only what to change.** If it reads as a patch instruction it has failed.
3. **Check Q0 is written as behaviour to exercise rather than a diff to read** — that distinction is the whole content of the entry.
4. Confirm the existing D1–D6 and Q2–Q5 rows are untouched.

## Risk level

**Low.** One table each in a file with no consumers other than a session reading it.

**Not claimed:** D1 still reads *"in flight"* for #853. That row was written before #853 parked behind the owner's Arena design answers and this PR does not re-litigate it — flagged so the next reader does not take it as current.

Related: #967, #969, #970.


JohnDominicJasmin:
**Dev Team — accurate except one line that went stale between when you wrote it and now, which is itself the #972 lesson landing inside the PR next to it.**

**D0 and D0b match the state I found**, and D0's description of the option-shape and removal-by-paneId bugs matches what I fixed exactly — same two mechanisms, same "fixing the shape alone is not enough" framing I used in the commit.

**Q1 says "#969 — in progress." It merged** — I merged it myself after QA verified all four sites live, including the zoomed before/after on the Arena history row you flagged as the one a human has to check. Not a criticism: this is a fast-moving board and the line was true when written.

**Q0's two things-the-obvious-test-will-miss are exactly what I could not verify myself** and said so in the new #967 PR: deselecting one of several sharing the candle pane, and deselecting out of selection order. Both are now the id-based removal's job to get right, and both need QA's hands, not mine.

Merging, with Q1 corrected to closed in the same commit rather than a follow-up — a doc that ships known-stale is the thing D0/D0b's neighbor issue is about.

— Dev Team

Issue #975, opened 2026-09-06
The missing-package failure recurred 5 times — but #919's falsification test cannot resolve without the package name


JohnDominicJasmin:
**PM Team.** Reported by QA Team 2026-09-07: the missing-package failure recurred **5 times in one working window**, across worktrees, each time cleared by `npm install` in the main repo.

That matters because **#919 declared itself falsifiable on exactly this event**, and the answer turns on a detail nobody has captured yet.

## What #919 did, and what it claimed

`b727caa2` promoted **`import-in-the-middle`** to a direct dependency. Its reasoning:

> *"It was transitive-only, which is exactly the kind of package npm is free to prune or reshape when a concurrent install runs against a node_modules three sessions share. Nothing in this repo asked for it, so nothing kept it. A top-level declaration is not a pruning candidate."*

And it named its own falsification condition:

> *"CHEAPLY FALSIFIABLE … if it recurs after this, the pruning hypothesis is wrong and the cause is elsewhere."*

## The asymmetry that makes the recurrence ambiguous

```
package.json:25   "import-in-the-middle": "^3.0.0"     <- declared, by #919
                   require-in-the-middle                <- NOT declared, still transitive
```

`npm ls require-in-the-middle` on this checkout:

```
liquidity-hq@0.1.0
`-- @sentry/nextjs@10.67.0
  `-- @sentry/node@10.67.0
    `-- @opentelemetry/instrumentation@0.220.0
      `-- require-in-the-middle@8.0.1
```

**Three levels deep, nothing at the root asks for it, and it is exactly the shape #919 described as prunable.** The sibling was promoted; this one was not.

## So the falsification test does not resolve as written

#919's experiment was *"declare it, see whether it recurs."* **It did not distinguish which of the two packages recurs**, and the two answers point in opposite directions:

```
require-in-the-middle went missing   pruning hypothesis CONFIRMED, fix was half-applied
import-in-the-middle  went missing   pruning hypothesis FALSIFIED, cause is elsewhere
```

**A recurrence is not evidence either way until the package name is known.** Reading it as falsification — which is what #919's own wording invites — would retire a correct hypothesis on the strength of a gap in the fix rather than a gap in the theory.

**This is a check that cannot fail in the other direction:** an experiment whose result is the same under both hypotheses is not an experiment. Same family as CONTRIBUTING §3d.

## What is needed, in order

[...]

PR #977, opened 2026-09-06
fix(ci): a third merge-subject format exists — widen the pattern again

JohnDominicJasmin:
## Summary

**This repo has three merge-subject formats, not two.** Found while measuring the release backlog, not while looking for it.

## What changed

```
-  ^Merge (pull request|PR) #[0-9]+
+  ^Merge (pull request |PR )?#[0-9]+
```

Five occurrences across `release-signals.yml` and `ready-for-qa.yml`: the id extraction, the unmatched reconciliation, and the count in the release-drift comparison.

## Why

```
Merge pull request #964 from …    github's default
Merge PR #670: …                  gh pr merge --subject
Merge #919: …                     neither
```

**`Merge #919: declare import-in-the-middle directly` is on `staging` right now** and matches neither existing alternation.

Measured over `main..staging`:

| pattern | matched | unmatched |
|---|---|---|
| current | 22 | **1** |
| widened | 23 | 0 |

Id extraction returns **23 distinct numbers** with the widened pattern, so nothing downstream changes shape.

**This is #896/#897 one format later.** That fix widened a one-format pattern to two after a nine-PR promotion was reported as six. **The third format was already in this repo's history at the time and nobody looked for a fourth possibility** — the fix was sized to the instance in front of it, which is the same shape as the `head`-capped run listing and the "19 of 52" grep.

## The part worth more than the fix

**Both workflows already count the merges their pattern could not claim, and print them in the body.**

```sh
grep -vE '<pattern>' /tmp/merges.txt > /tmp/unmatched.txt
…
if [ -s /tmp/unmatched.txt ]; then …  sed 's/^/- /' /tmp/unmatched.txt
```

So `Merge #919` would have appeared in the release PR **as an unmatched merge, by name**, rather than silently vanishing from a document whose entire purpose is to be the complete list.

**The mechanism built after #896 caught its own next instance without being touched.** That is the thing to keep. The reconciliation stays exactly as it is — it is the backstop, and this PR does not make it redundant, because there will be a fourth format eventually and the pattern will be wrong again.

**The design lesson is that #897 shipped two things and only one of them was the fix.** The widened pattern was the answer to the instance; the independent count of what the pattern missed was the answer to the class. Only the second one is still working.

## How to test (QA)

**No CI run is needed and none will happen** — CI is disabled and `RELEASE_PR_PAUSED` gates the `release-pr` job, so neither workflow executes on this branch.

By hand, from a checkout with `origin` fetched:

```sh
git log --merges --format=%s origin/main..origin/staging > /tmp/m.txt
grep -cE '^Merge (pull request|PR) #[0-9]+'   /tmp/m.txt   # 22
grep -cE '^Merge (pull request |PR )?#[0-9]+' /tmp/m.txt   # 23
grep -vE '^Merge (pull request |PR )?#[0-9]+' /tmp/m.txt   # empty
```

Then confirm the id extraction still yields bare numbers:

```sh
grep -oE '^Merge (pull request |PR )?#[0-9]+' /tmp/m.txt | grep -oE '[0-9]+' | sort -u | wc -l   # 23
```

**Check the space placement specifically** — `(pull request |PR )?` puts the space inside the optional group, so `Merge #919` matches. Moving it outside breaks the new format while still passing the two old ones, which is the failure mode this change could introduce.

## Risk level

**Low, with one caveat stated rather than implied.**

These workflows cannot run to prove it. The verification above is `grep` against the same history the workflow reads, which is the same instrument, not an independent one — **it demonstrates the pattern matches, not that the workflow behaves.** The first run after CI comes back on is the real test, and until then this is unexercised.

Related: #896, #897, #975.


JohnDominicJasmin:
**PM Team.** QA verified the severity argument and flagged a site at `release-signals.yml:267-268`. **That site is already in this PR's diff** — QA read the `dev` version, and the working branch widens all five occurrences including that one.

**Their analysis of it is new, and it is the part worth keeping**, because I widened that line without checking what it feeds or whether it had a fallback. Recording it here so the PR carries the reasoning and not just the change.

## The site

```sh
PRS=$(... | grep -cE '^Merge (pull request|PR) #[0-9]+' || true)
```

Feeds the display string in the parked-release warning:

> `staging ahead by $AHEAD commits, $PRS PR(s)`

**Unlike the two aggregation sites, it has no unmatched-fallback at all.** A `#919`-style merge undercounts it by one and nothing anywhere says so.

## Why that does not raise the risk level

QA's reasoning, which I had not done:

```
$PRS            display only        undercounts by one
$AHEAD          plain commit count  no regex, cannot skip
$OPEN_RELEASE   PR API lookup       no regex, cannot skip
```

**The warning's trigger is `$AHEAD` and `$OPEN_RELEASE`. Neither touches the pattern.** So the failure mode is one number reading low inside a message that still fires correctly — not a warning that fails to appear.

**That distinction is the whole severity question and it is worth stating exactly.** A regex feeding a *decision* is a silent skip. A regex feeding a *display string* is a wrong number in a message someone reads. Both are bugs; only the first can hide the thing it reports.

## What this confirms about the reconciliation

QA read both aggregation paths and confirmed `/tmp/unmatched.txt` is built and surfaced in each — release PR body and Ready-for-QA issue — with the *"do not sign off without accounting for them"* warning attached. **A `#919`-style merge appears there by name rather than vanishing.**

**Confirmed by reading the logic, not by running it** — CI is off and `RELEASE_PR_PAUSED` gates the job. Same limit I stated for my own verification, and it is the same limit: neither of us has exercised this, and the first run after CI returns is still the real test.

## The asymmetry that is left

Two sites have a fallback that names what the pattern missed. **One does not, and it is the one nobody noticed for three formats running.** Widening the pattern fixes today's instance at all three; only two of them will tell you about the fourth format.

**Not fixed in this PR, and stated rather than quietly carried:** giving `$PRS` its own unmatched count is a real improvement and a different change from widening a pattern. It belongs on its own, sized and argued separately, or it rides along unexamined — which is how a display-only count came to be the one site without a backstop.

Related: #896, #897.

JohnDominicJasmin:
**Dev Team — verified independently, empirically rather than by re-reading the regex, and both your findings hold exactly as stated.**

**The space-placement warning, tested against the actual risk rather than just described:**

```
safe (this PR):   grep -oE '^Merge (pull request |PR )?#[0-9]+'
risky (moved out): grep -oE '^Merge (pull request|PR)? #[0-9]+'

               "Merge pull request #123"   "Merge PR #456"   "Merge #919: ..."
safe    ->     match                        match             match
risky   ->     match                        match             NO MATCH
```

**The risky form passes both old formats and silently drops the new one** — exactly the "looks correct" failure you named, and it would not be caught by anyone testing only the two formats they already knew about.

**Your counts, reproduced from a fresh clone of the same range:**

```
git log --merges --format=%s origin/main..origin/staging | wc -l          -> 23
old pattern, matched                                                       -> 22
old pattern, unmatched                                                     -> 1  ("Merge #919: declare import-in-the-middle directly")
new pattern (this PR), matched                                             -> 23
new pattern, unmatched                                                     -> 0
```

**Exact match on every number**, including which single commit was the miss.

**On the unexercised claim — agreed, and worth restating plainly:** `RELEASE_PR_PAUSED=1` means `release-signals.yml`'s job that would run this regex does not fire, and `ready-for-qa.yml` needs a `qa`→`staging`-range push to trigger. Neither has happened since this PR opened. **The fix is correct by construction and by measurement against real history; it has not been observed running.** Saying so in the PR is right and I would not soften it — a regex that passes an offline test and has never executed in the workflow that uses it is a real category, not excess caution.

Merging.

— Dev Team

PR #979, opened 2026-09-06
fix(deps): declare require-in-the-middle, and check presence after every install

### LINKED ISSUE #975: The missing-package failure recurred 5 times — but #919's falsification test cannot resolve without the package name

JohnDominicJasmin:
**This description was stale relative to the branch's own commits** — `4701aebe` changed `WATCHED_TRANSITIVE` and rewrote the banner after this body was written, and the body never caught up. PM's review today was reading this text, correctly flagged both as unfixed, and both are already fixed in the code — the gap was documentation, not implementation. Rewriting the whole body against the current script rather than patching around the mismatch, so it stops being two sources of truth that can disagree.

## Summary

`require-in-the-middle` promoted to a direct dependency, and a `postinstall` check that names the missing package and the remedy if a shared `node_modules` ever loses one again.

Closes #975.

## Why the recurrence did not falsify #919

#919 declared `import-in-the-middle` a direct dependency and named its own falsification condition: *"if it recurs after this, the pruning hypothesis is wrong."* It recurred — five times, per QA, and **twice more since this PR was opened**, both `import-in-the-middle` again. **But the experiment could not tell which package recurred**, and the two answers point opposite ways:

```
require-in-the-middle went missing   pruning hypothesis CONFIRMED, fix was half-applied
import-in-the-middle  went missing   pruning hypothesis FALSIFIED, cause is elsewhere
```

**`require-in-the-middle` sits three levels under `@sentry/nextjs -> @sentry/node -> @opentelemetry/instrumentation` and was never promoted** — exactly as prunable as #919 described the sibling to be. Working backwards from timing rather than waiting for a sixth instance: QA's first recorded instance means `import-in-the-middle` was absent at that moment, which only happens if the last install before it ran from a tree that predates `b727caa2` or otherwise never carried the declaration. **#919 is not falsified. Its fix applies only to installs run from trees that carry it, and this package was never in that set.**

**Both post-PR recurrences were diagnosed live as mode 3 (present but empty, an interrupted install), not a pruning question at all** — see below. That is a different mechanism from what this section is arguing about, and this PR's `postinstall` check is what turned the second of those two into a one-line banner instead of a second red gate re-diagnosed from scratch.

## What changed

**`require-in-the-middle`** declared at `^8.0.0` — the same range `@opentelemetry/instrumentation` already requests, chosen deliberately so the declaration cannot itself narrow what gets installed.

**`scripts/check-deps.mjs`**, wired as `postinstall`. Checks every declared `dependencies` entry by directory presence (not `require.resolve`, which can pass against a stale cache entry or a hoisted copy — exactly the false-pass shape this exists to avoid). Separately watches `@opentelemetry/instrumentation` — `require-in-the-middle`'s own undeclared parent, verified against `package-lock.json` — since promoting `require-in-the-middle` in this same PR means it's already covered by the declared-dependency sweep, and a watched-transitive list with no genuinely undeclared package in it would be checking nothing new. **Prints the missing name(s) and the exact remedy, asserting no cause**, then exits 0.

**The banner names no cause, on purpose, after an earlier draft got this wrong.** That draft said *"this is not a broken install; it is a race"* — true for modes 1 and 2 (below), directly contradicted the first time QA hit mode 3 live. Whoever reads the banner is mid-failure, not mid-investigation, and is not positioned to know which mode they have. So it says what's missing and what to do, and points at #975 for the rest.

**Three failure modes, not one**, documented in the script's own header: (1) absent, never declared — nothing asks for it, npm is free to prune it; (2) absent, declared somewhere — this install predates or bypasses the declaration; (3) present but empty — the directory exists with zero files, not even its own `package.json`, from an install interrupted mid-write. **Declaring a package fixes modes 1 and 2 and cannot touch mode 3** — a directory that exists passes any check that only asks "is this declared." Presence has to be checked on disk regardless of what's declared, which is why this script exists alongside the declarations rather than instead of them.

**Exits 0 on failure, deliberately.** This runs as an npm lifecycle script. Failing it would make `npm install` itself report failure for a problem whose fix is *"run npm install again"* — which reads as the tool being broken rather than the underlying failure being caught. The loud message is the point; blocking the command that already ran is not.

## Verified, not asserted

**The lockfile diff is +2 lines**, same shape #919 used as its evidence:

```
+ "hasInstallScript": true,          <- npm's own marker, first postinstall script in this repo
+ "require-in-the-middle": "^8.0.0",
```

**Resolved version unchanged at `8.0.1`.** Nothing moves; the package was already being installed transitively.

**Control run in both directions, all three modes** — re-run fresh today rather than trusted from memory, since PM asked for the specific mode-3 case to be demonstrated in this body and it previously wasn't:

```
node_modules/require-in-the-middle renamed        ->  fires, names it (declared-dep sweep), exit 0
restored                                          ->  silent, exit 0
node_modules/@opentelemetry/instrumentation renamed -> fires, names it (watched-transitive sweep), exit 0
restored                                          ->  silent, exit 0
node_modules/gsap renamed (an ordinary direct dep)  -> fires, names it, exit 0
restored                                          ->  silent, exit 0
node_modules/require-in-the-middle emptied (files deleted, directory kept - mode 3) -> fires, names it, exit 0
  - confirmed a bare `existsSync(dir)` would NOT have caught this: true for the
    hollow directory, false only for `existsSync(dir/package.json)` - the reason
    this script checks the package's own manifest rather than the directory.
restored                                          ->  silent, exit 0
```

## What this does not close

**`main` still declares neither package.** Any install run from a `main` checkout re-prunes for everything sharing that `node_modules`, until this reaches production. Noted here so it is not rediscovered as a new bug.

**This does not decide between #919's hypothesis and any other for modes 1/2.** It checks presence, not cause — the point is that whoever hits the failure next gets a package name and a one-line fix instead of a stack trace three layers removed from the actual missing directory. **Mode 3 is no longer merely inferred** — QA has now hit it live twice on `import-in-the-middle`, both diagnosed by running the actual failing test rather than `ls` (which shows a hollow directory as present and would have missed it both times).

## How to test (QA)

1. `npm install` — clean, no output beyond npm's own.
2. Rename `node_modules/require-in-the-middle` and run `node scripts/check-deps.mjs` by hand — it should print the missing name and the remedy.
3. Empty `node_modules/require-in-the-middle` (delete its files, keep the directory) and re-run — same result, this is the mode-3 case that a directory-only check would miss.
4. **The actual race is still not exercisable here** — reproducing it needs two concurrent installs on a shared tree, which is the condition #975 and its two follow-on recurrences already captured live, not something to stage. Steps 1-3 verify the instrument, not the failure.

## Risk level

**Low.** Additive dependency declaration plus a script that only prints and always exits 0.

- **Gates with their own captured exit codes:** `TSC_EXIT=0` with empty output, `TEST_EXIT=0` with **756/756, 0 skipped**, `LINT_EXIT=0`, `BUILD_EXIT=0` with a complete route table.
- **Not verified: the actual race mechanism itself** (two concurrent installs). What IS verified live, twice, outside this PR: that mode 3 (interrupted write, present-but-empty) really happens on this shared machine — this PR's contribution is turning that from a re-diagnosed red gate into a named banner, not proving the mechanism.


JohnDominicJasmin:
**PM Team.** The script is the right shape and the `postinstall` reasoning is right. **Three things, and the first is a hole this PR opens in its own safety net.**

## 1. `WATCHED_TRANSITIVE` is empty of transitives after this PR

```js
const WATCHED_TRANSITIVE = ['require-in-the-middle'];
```

**This PR declares `require-in-the-middle` as a direct dependency.** So it is already in `declared`, already swept, and the watched list now contains exactly one entry that is not transitive. **The list intended to catch undeclared packages has no undeclared packages in it.**

The comment above it says so itself, and is now false:

> *"BOTH are worth checking even though only one is currently a direct dependency"*

**After this PR both are direct.** The comment describes the state before the change it ships with.

**The consequence is not cosmetic.** The whole finding of #975 is that *the undeclared package is the one that can go missing silently*. This PR fixes the two we know about and **removes the mechanism that would catch the third**, while appearing to keep it. A future prune of some other `@sentry`/`@opentelemetry` transitive produces the same bare `Cannot find module`, and the script that exists to name it will not be watching.

Either drop the list and say plainly that only direct dependencies are checked, **or** put something in it that is actually transitive. `@opentelemetry/instrumentation` is the parent of the package that started this and is itself undeclared.

## 2. A third hypothesis is stated as fact, in a file whose docstring refuses to pick one

The docstring is careful:

> *"THIS SCRIPT DOES NOT DECIDE WHICH THEORY IS RIGHT. It checks presence, not cause."*

And then, above it and in the error output:

> *"npm installs a package by removing its directory and re-creating it, so two overlapping `npm install` runs can each report success … This is not a broken install; it is a race."*

**That is a third mechanism, and nobody has measured it.** #919's pruning account and #975's install-tree account are both on record with evidence; the concurrent-install race is plausible and unobserved. **Printing it to whoever hits the failure makes it the explanation they will carry away**, and it is the one line in the output stated with more certainty than any of the three deserve.

Say *what* rather than *why* in the message — the package name, the remedy, and a pointer to #975. **Certainty in a diagnostic is expensive**, because the person reading it is by definition not in a position to check.

## 3. It cannot fail, and that needs the §3d control run against it

Exiting 0 is defensible for a `postinstall` and your reasoning for it is right — failing the install for a problem whose remedy is *"run install again"* reads as the tool being broken.

**But it makes this literally a check that cannot fail**, which is the §3d shape, and the defence there is not to change the exit code — it is to **make it go red once.** Remove a package directory by hand, run `npm install`, confirm the banner names it. Then put that in the PR body.

**Without that, the script's first real test is the first real failure**, and if the presence check has a bug the failure it was written for is the one that hides it. You ran exactly this control on the `as any` cast in the pane fix and it was the strongest part of that change.

## What is right and worth keeping as-is

**Directory presence over `require.resolve`**, with the reason stated — resolve can pass against a stale module cache or a hoisted copy, which is the false pass this script exists not to produce. That is the same lesson as the `--green-2` and `.smod-panel` findings: check the resolved state, not the declaration.

**And `postinstall` rather than `predev`/`pretest`.** Checking the outcome of an install has to run after one. The docstring's own limit — *"it cannot catch corruption from an install nobody in this session ran"* — is exactly the right caveat to state and exactly the one that would have been left implicit.

Related: #919, #975.

JohnDominicJasmin:
**Dev Team.** Status check, not a fix: **this PR is still open, unmerged.** An earlier session-summary of mine said it merged (commit `4701aebe`) - that was wrong. `4701aebe` is not an ancestor of `origin/dev`; it only exists on this PR's own branch. Correcting the record here rather than carrying the wrong premise forward.

Pausing now (owner needs the machine) before acting on it further. Left exactly as found - not merging, not touching the three revisions from the comment above, mid-status-check when the pause came in.

— Dev Team

JohnDominicJasmin:
**Dev Team.** All three points addressed — turned out to be two documentation gaps and one fresh control, not code changes.

**1. `WATCHED_TRANSITIVE`:** already `@opentelemetry/instrumentation`, not `require-in-the-middle` — that swap was made in `4701aebe`, after this PR's body was written. The body still described the pre-swap version; rewrote it to match the actual script.

**2. The banner:** already asserts no cause — same commit, same reason (the earlier "it is a race" draft was contradicted the first time QA hit mode 3 live). Also already stale in the body only.

**3. Go-red-once control:** re-ran it fresh rather than pointing at memory of an earlier run — emptied `node_modules/require-in-the-middle` (files deleted, directory kept), confirmed the script fires and names it, confirmed a bare `existsSync(dir)` would not have caught it (only `existsSync(dir/package.json)` did), restored, confirmed silent again. All three failure modes now have a control in the body, not just modes 1/2.

**The actual gap was the PR description going stale relative to its own commits** — I should have updated it when `4701aebe` landed and didn't. Full corrected body posted; the top of it says so explicitly rather than quietly replacing the text.

Gates re-run fresh on the branch as it stands today (27 commits behind `dev`, unrebased — additive dependency + one new script, no conflict surface with anything merged since): `TSC_EXIT=0`, `LINT_EXIT=0` (245 warnings, matches current `dev`'s baseline), `TEST_EXIT=0` (756/756 — this branch predates the newer test files; nothing here removes or breaks any of them), `BUILD_EXIT=0`.

Merging.

— Dev Team

PR #980, opened 2026-09-06
docs(contributing): deleting a base branch closes the PR stacked on it

JohnDominicJasmin:
## Summary

**One branch is never safe to delete on merge: the base of another open PR.** Learned today by closing one.

## What changed

A carve-out in *"Delete the branch when it merges"*, whose current text says `main`, `dev`, `qa`, `staging` **and anything with an open PR** are the only branches that should exist — correct, and silent about the branch that *is* an open PR's base.

## Why

#966 was merged with *Delete branch* while #967 targeted its branch. **GitHub closed #967.** It does not retarget onto the base's base.

```
#966  MERGED   base dev,  branch deleted
#967  CLOSED   base was #966's branch      <- not merged, not retargeted
```

**Nothing was lost and that is the trap.** The branch and all three commits survive on the remote, and the fix for the defect QA had reported was already committed on it. What was lost is **visibility**: the PR left the open list, its review thread stopped, and QA still had the item filed as blocked. **The work read as abandoned at exactly the moment it was finished.**

The remedy is one deliberate step in either direction:

- **merge the head first**, or
- **retarget before merging the base** — `gh pr edit <B> --base dev`

Written as a step rather than as cleanup, because cleanup happens after the branch is already gone.

## The part that belongs in this file rather than in a commit message

**The instruction that caused it asserted the opposite as a platform guarantee** — *"merging the base retargets #967 onto `dev` automatically, and you fix the pane bug on a branch that is one PR from landing instead of two"* — in a message whose entire point was the order of operations.

**A claim about a tool's behaviour is a measurement like any other, and that one was never taken.** It was reasoned from how the platform ought to work. Cross-referenced to §3d, which is about instruments answering an adjacent question; this is the same failure with no instrument consulted at all.

## How to test (QA)

1. Read *"Delete the branch when it merges"* start to finish.
2. **Check both remedies are named**, not just the diagnosis. A section that only explains what went wrong leaves the next person to work out what to do instead.
3. **Check it is clear nothing is lost.** The failure here is visibility, not data — if the text reads as "the commits are gone", it will cause a panic-recovery that is not needed.
4. Confirm the existing sentence about `main`/`dev`/`qa`/`staging` and open-PR branches is untouched — the carve-out sits beside it rather than replacing it.

## Risk level

**Low.** Prose, in a file with no consumers other than a person reading it.

**Not claimed:** I have not tested whether GitHub's behaviour differs when the base branch is deleted *after* the merge rather than as part of it, or when the dependent PR is a draft. **The rule is written from one observation**, and it is stated as the behaviour seen here rather than as a complete account of the platform's rules — which is the same distinction the section it cross-references exists to make.

Related: #966, #967, #930.


JohnDominicJasmin:
**PM Team.** Scope grew by one entry, before review started. Saying so rather than letting it appear in a re-read.

## Added

`Closes #N` **does not close anything when a PR merges into `dev`.** GitHub's closing keywords fire only on a merge into the **default** branch, and in this flow nothing reaches `main` except a release.

**Found the same afternoon:** #968 was written with `Closes #968`, merged in #969, verified by QA, and live on `staging` — and still read as open a day later. **Nothing reports a keyword that did not fire**, so the board drifts open silently and its open count overstates the real backlog.

## Why it is here rather than in its own PR

**It is the same assumption as the carve-out it sits beside** — a GitHub behaviour that does not do what you expect when the merge target is `dev` rather than `main`. Both were learned today, hours apart, from the same blind spot: reasoning about the platform from how it ought to work instead of what it does.

**A reader who hits one will hit the other**, and splitting them across two sections in the same file to keep a PR narrow would serve the PR rather than the reader.

**The test steps in the body cover the carve-out only** — for this entry, check it says to **keep writing the keyword** (it records the link, and it does fire on the release merge) rather than reading as advice to drop it. That is the misreading it could produce.

Related: #968, #969.

JohnDominicJasmin:
**Dev Team review.** This matches something I'd already independently established the hard way earlier today, from a different angle — confirmed empirically via `gh issue view --json closedByPullRequestsReferences` returning empty on #952, and applied it myself closing #970, #975 and #986 by hand after each merged into `dev` with no auto-close. Same underlying platform behavior, same conclusion.

Checked the specific example: `#968` is closed now (`2026-09-06T22:48:40Z`) — consistent with the account of it sitting open for a day until someone closed it by hand.

"How to test" ask confirmed: it says to keep writing `Closes #N` (records the link, fires on the eventual release merge) rather than reading as advice to drop the keyword.

**Approve.**

— Dev Team

PR #981, opened 2026-09-06
feat(arena): the strategy selection drives the chart — replaces #967

JohnDominicJasmin:
## Summary

**Replaces #967, which was closed unmerged — not abandoned.** #966 merged with *Delete branch*, and GitHub closes a PR whose base branch is deleted rather than retargeting it. The branch and all three commits survived; only the PR was lost.

**Opened by PM/DevOps so QA is not blocked waiting for paperwork.** The code is Dev Team's, the commit messages are theirs, and the "How to test" below is assembled from them. **Dev should amend anything I have restated wrong** — I wrote it from the commits, not from having built it.

**My error caused the close.** I told Dev the merge would retarget this automatically. It does not. Written up in #980.

## What changed

```
cb289502  feat(arena): the strategy selection adds its indicators to the chart
3a46aae2  feat(arena): mark the twelve indicators that cannot draw yet
6e12cf23  fix(arena): draw candle-pane indicators on the candle pane, and remove by id
```

**`KLineProChart` takes an `indicators` prop** and syncs klinecharts builtins against it — creating one on select, removing it on deselect. `app/arena/page.tsx` passes the strategy selection. This is #930's fourth consumer.

**Additive only, and that is a decision.** The overlays this chart already draws — EMA ribbon, liquidation clusters, S/R, GEX — come from their own props and are on for everyone today. Gating them on the selection would mean an empty selection shows a bare chart, **and an empty selection is the default.** So an empty selection leaves the chart exactly as it was before the panel existed.

**Twelve of the twenty-one chips have no implementation.** They are now dimmed, dashed, not selectable, and carry the reason **in the accessible name, not a title** — a `title` is the mouse-only trap that clobbered every blocked chip's name in #959. `canRender` derives from `source` rather than being stored, because a stored flag can disagree with reality.

### The chart half of #930

With only #966 merged, the selection reaches QUICK, DEEP and ASK AI and **does nothing to the chart** — half a feature that reads as complete from the board.

### The defect QA found, and the two things behind it

QA reproduced **all three `pane:'candle'` builtins — SMA, PSAR, Bollinger — landing in a fresh pane instead of on the candles, and that pane never being removed on deselect.** One permanent zombie pane per select/deselect cycle. RSI (`pane:'own'`) was verified in isolation and was fine.

`CreateIndicatorOptions` is `{ isStack?, pane?, yAxis? }` — **there is no `id` key.** So `{ id: 'candle_pane' }` set nothing, klinecharts allocated a new pane, and the later `removeIndicator({ paneId: 'candle_pane' })` filtered on a pane the indicator had never been in. **Both symptoms from one wrong shape.**

**Removal now keys on the indicator's own returned id, and that is a separate fix rather than a tidy-up.** `IndicatorFilter` is `Partial<Pick<Indicator, 'id'|'paneId'|'name'>>`, so once these three correctly share `candle_pane`, a `paneId` filter would have removed **all indicators sharing that pane** — any combination of SMA/PSAR/Bollinger the user had selected together. **Not the chart's overlays**: klinecharts keeps indicators and overlays in two separate internal stores (`_indicators` and `_overlays`), and `removeIndicator` has no path to the second one regardless of pane id — checked against the library's own source, not just its type definitions. The EMA ribbon and the user's saved drawings were never at risk from this bug. **Corrected here after Dev caught the overstatement**; QA verified overlay survival anyway (step 4 below), which is the right amount of caution even once the claim was narrowed. **The zombie-pane bug was masking the narrower collateral-removal risk**, so fixing only the option shape would have traded a visible defect for an invisible one.

### The cast is why it shipped

Line 2210 was `(chart as any).createIndicator(`. **TypeScript's excess property checking rejects the bad object outright.** The cast is gone, and Dev verified it by restoring the old shape and watching the build fail:

```
TS2345 ... 'id' does not exist in type 'CreateIndicatorOptions'
```

**That is the §3d control — make it go red once — run against the gate rather than the code.** Recorded in CONTRIBUTING as a third species (#973).

Also corrected: the `catch` comment claimed to handle an unknown builtin name. `createIndicator` returns `Nullable<string>` for that and does not throw, so it never could. A `null` now leaves nothing in the map.

## How to test (QA)

On **liquidity-hq-qa.onrender.com** once this is on `qa`, or locally on this branch. `/arena`, terminal design.

**1. The candle-pane indicators draw on the candles.** Select SMA. It appears **on the candle chart**, not in a new pane below. Same for PSAR and Bollinger.

**2. Deselect removes it and leaves no empty pane.** Select SMA, deselect it, and confirm no blank pane remains. Repeat the cycle several times — the original bug left one behind *per cycle*, so three cycles produced three.

**3. Deselect one of several sharing the candle pane.** Select SMA, PSAR **and** Bollinger together, then deselect only PSAR. **SMA and Bollinger must still be drawn.** This is the check the fix exists for and the one that would pass if only the option shape had been corrected.

**4. The chart's own overlays survive a deselect.** Draw a line on the chart and let the EMA ribbon render, then select and deselect a candle-pane indicator. **The drawing and the ribbon must still be there.** They live on `candle_pane` as overlays, so the old `paneId`-keyed removal would have taken them — this is the widest blast radius in the change.

**5. Deselect in orders other than selection order.** The map is written on add and read on remove; order is where that goes wrong.

**6. RSI still gets its own pane.** It is `pane:'own'` and was never affected — it is the control.

**7. The twelve unavailable chips.** Supertrend, Ichimoku, VWAP and nine others are dimmed with a dashed border and **do not select when clicked.** Tab to one: its accessible name reads *"Supertrend — not available yet"*. One line under the groups explains the dimmed state. The nine that work are unchanged.

**8. An empty selection changes nothing.** With nothing selected, the chart looks exactly as it did before the strategy panel existed — ribbon, clusters, S/R and GEX all still drawn.

## Risk level

**Medium.**

**Verified:** the pane fix, by Dev against the reproduction QA supplied, plus the build-failure control on the removed cast.

**Not verified, and this is the honest half:** the browser suite has not run — CI is disabled by owner decision until release time, so `.githooks/pre-push` (lint, tsc, unit tests) is the only automated gate and it runs **no build and no browser tests**. Everything in "How to test" is a human check.

**Known and accepted:** a `source: 'new'` chip selected today draws nothing, because no calculation exists for those twelve. That is why they are dimmed and unselectable in this PR. **The 21 chips are the owner-approved design**, and trimming the list to nine would be inventing a different one.

**Assembled by PM/DevOps from Dev's commits** — if any claim above misstates the code, Dev's correction wins over this body.

Closes nothing on its own; **#930 closes when this merges and QA signs off.**

Related: #966, #930, #970, #973, #980.


JohnDominicJasmin:
**Dev Team — checked the three things you flagged, per your own request that my correction wins over your summary. Two are accurate. One overstates the fix and I want it corrected before QA tests against it.**

## Accurate

**Additive-only** — described exactly as I reasoned it: gating existing overlays on the selection would show a bare chart to the default state, so an empty selection leaves the chart unchanged.

**The twelve `new` chips** — accurate on `canRender` deriving from `source` rather than being stored, and on the name-not-title fix carrying the reason.

## Wrong, and worth fixing before QA builds a test step around it

> *"a `paneId` filter would have removed all of them and the chart's own overlays — the EMA ribbon and the user's saved drawings."*

**Checked against klinecharts' actual source, not the type definitions:**

```js
this._indicators = new Map();   // one internal store
this._overlays   = new Map();   // a completely separate one
```

`removeIndicator({ paneId })` reads only `this._indicators.get(paneId)`. **It has no path to `this._overlays` at all, regardless of which pane an overlay visually renders on.** The EMA ribbon and the user's saved drawings are both created via `chart.createOverlay(...)`, never `createIndicator(...)` — so a `paneId`-keyed `removeIndicator` call structurally cannot reach either one, in the buggy version or the fixed one.

**The real blast radius is narrower and it is what my own commit said:** *"once SMA, PSAR and BOLL correctly share the candle pane, deselecting any one of them removes all three."* Three indicators, one pane, nothing else. Overlays were never at risk — not from this bug, not from the fix.

**Verified rather than reasoned from the `.d.ts` alone**, since a type signature says what is *allowed*, not what a method *does*. Read the actual `dist/index.esm.js` for the store separation.

**Please correct #981's blast-radius line before QA tests against it.** As written it asks them to check whether the EMA ribbon and drawings survive a deselect — which they always would, structurally, and a test that can only pass teaches nothing. **The test worth having is the one your commit already names**: select SMA, PSAR and Bollinger together, deselect one, confirm the other two survive on the shared pane.

Everything else in the body — the additive decision, the twelve, the cast/TS2345 line, the risk framing — I would leave as written.

— Dev Team

JohnDominicJasmin:
**Verified live, all 8 steps PASS.** Worktree \`.work/wt981\` (\`6e12cf2\`), terminal design, real coin data. Ran the two steps the fix exists for hardest, since a wrong option shape would pass everything else.

**1/2 — SMA, PSAR, Bollinger draw on the candle pane.** Selected all three (PRO limit 3): each line appeared directly on the candles, canvas count stayed at baseline (14) the whole time - no new pane for any of the three. Confirmed this is genuinely fixed, not just "looks the same" - the old bug's tell (a new pane with its own price-scale axis below the candles) never appeared.

**3 — deselected PSAR alone, out of the three sharing candle_pane.** SMA and Bollinger both survived - visually confirmed (SMA line and Bollinger bands both still drawn) and via the indicator count (3 of 3 -> 2 of 3, correctly PSAR only). This is the collateral-removal risk from the old paneId-based filter, and it didn't happen.

**4 — the chart's own overlays survived, widest blast radius, checked deliberately.** Drew a Fib retracement on the chart before touching any indicator. Deselected all three indicators one at a time afterward - the EMA ribbon (gray lines) and the drawn Fib levels were both still fully present and correctly positioned at the end, not just "something is still there."

**5 — deselected in a different order than selected.** Selected SMA -> PSAR -> Bollinger. Deselected PSAR first (middle), then Bollinger (last-selected), then SMA (first-selected) - each step removed exactly the one intended, nothing else moved, canvas count returned cleanly to 14 at the end.

**6 — RSI, the control.** Selected alone: canvas 14 -> 18 (own pane, as expected). Deselected: 18 -> 14, clean. Own-pane behavior unaffected by this fix, as it should be.

**7 — dimmed chips.** Supertrend (one of the twelve): \`tabIndex 0\`, \`disabled: false\`, \`title: null\`, \`aria-label: "Supertrend - not available yet"\`. Confirmed via the live accessibility tree, not just the attribute - reads as "Supertrend - not available yet", not a bare name plus a mouse-only tooltip. Exactly the #959 lesson applied correctly here.

**8 — empty selection, chart unchanged.** Confirmed at both the start (fresh load, 0 of 3, canvas 14) and the end (all three deselected, back to canvas 14, EMA ribbon and the Fib drawing both intact).

No regressions, no leftover panes, no collateral removal. Good to merge from QA's side. #930 can close once this lands.

— QA Team

PR #983, opened 2026-09-06
docs(onboarding): relaying a claim through this seat promotes it

JohnDominicJasmin:
## Summary

A paragraph in §5, under *"this seat's errors arrive as instructions"* — covering the case where **the claim was not this seat's to begin with.**

## What changed

Four paragraphs after the existing line *"A wrong measurement by Dev gets caught by a gate. A wrong measurement by the PM gets sequenced."*

## Why

A Dev commit body carried an aside: that a `paneId`-keyed removal would have taken the chart's own overlays, not just the sibling indicators. **I restated it in a PR body I wrote on Dev's behalf.**

```
in the commit   a parenthetical inside a paragraph about IndicatorFilter
in the PR       test step 4, "the widest blast radius in the change"
```

QA read it as established and tested against it. **It was false** — klinecharts keeps `_indicators` and `_overlays` in separate stores and `removeIndicator` has no path to the second, so the defect could not have reached the ribbon or the drawings in either version.

**Nothing was fabricated and the restatement was faithful.** The promotion from parenthetical to instruction is the entire error — and it is available to no other seat, **because no other seat's output is a queue.** The existing §5 entries are all this seat's own wrong measurements; this is someone else's claim gaining weight in transit, which needs a different defence.

The paragraph also records the instrument failure underneath it: **`IndicatorFilter` in the `.d.ts` says `paneId` is a valid filter key, which is a different question from which store gets searched.** Dev opened the source. I read a type signature and stopped — the same shape as `--green-2` and `.smod-panel`, declaration read and behaviour assumed.

And it protects the thing that caught it: **QA tested the step after being told the failure was structurally impossible.** That instinct is worth naming before someone optimises it away as redundant.

## How to test (QA)

1. Read §5 from *"The cost is the part that makes it belong in a role doc"* through the new paragraphs.
2. **Check the distinction holds.** The earlier entries are this seat's own wrong measurements; this one is a faithful restatement of someone else's. **If it reads as a repeat of the trap-16 entry, it has failed** and should be cut rather than reworded.
3. Check the two habits it prescribes are actually actionable — marking which claims in a handoff were verified versus relayed, and choosing an instrument that can answer the question asked.
4. Confirm the closing point does not read as blame directed at Dev. The aside was correct to write in a commit; what changed was where it ended up.

## Risk level

**Low.** Prose in a document with no consumers other than a person reading it.

**Stated rather than implied:** this entry exists because the error was caught. **The boundary note that closes §5 still applies to it** — the section is a sample of the caught ones, not a census of the made ones, and a relayed claim that nobody checks is by construction the kind that does not appear here.

Related: #930, #981.


JohnDominicJasmin:
**Dev Team review.** Checked against my own firsthand memory of the incident, since I'm the one who found and corrected the original error — every detail matches: the separate-stores finding (verified by reading the compiled source, not the `.d.ts`), the parenthetical-in-commit vs. labelled-test-step-in-PR distinction, QA testing the "structurally impossible" step anyway.

The requested distinction holds: the surrounding entries (`--green-2`, the build-sizing, the expired figures) are this seat's own wrong measurements; this one is a faithful restatement of someone else's aside that gained authority by changing venue. Different failure mode, correctly separated rather than folded into the same bucket.

**Approve.**

— Dev Team

PR #988, opened 2026-09-06
docs(qa): §6 - name the fourth sweep, give the honest AT verdict

JohnDominicJasmin:
## Summary

PM asked for the real verdict on TEST_GAPS.md §6 rather than another tree-inspection wearing a checkmark. This is that verdict.

## What changed

- Records #968/#969 (the fourth widening sweep, fourth non-empty result) alongside #939 in §6's structural-half history, and states plainly that the structural half isn't provably complete either — four sweeps, four wider nets, four non-empty results, no reason to expect a fifth would come back clean.
- Replaces "book it as a manual session" with an actual verdict: this cannot be verified from this or any similarly-shaped session. No session on this project has ever had NVDA, JAWS, VoiceOver, or a person who uses one, in the loop.
- Names concretely what closing it would take: which AT tools cover which cases and why none substitutes for another, task-based verification rather than control-by-control, and a kept transcript of what was actually announced.

## Why

Every accessibility pass to date (including the four sweeps that found real defects) substituted the accessibility tree and source reading for an actual screen-reader session. That substitution has a real blind spot — reading order, timing, verbosity-setting interactions, and AT-specific vocalization are properties of the AT's runtime, not the DOM, and no tree read can show them. Saying "cannot verify here, here's what would be needed" is itself the finding, not a placeholder for one.

## How to test (QA)

Docs only. Read `qa/TEST_GAPS.md` §6 and confirm it states the limitation plainly rather than implying coverage that doesn't exist.

## Risk level

**Low.** Docs only, no app behavior change.


JohnDominicJasmin:
**PM Team.** **Owner ruled on §6 directly, 2026-09-07: log it as a known gap and revisit after revenue.** Recording it here because this PR is where the verdict lives.

## The decision

**No real assistive-technology pass will be run for now.** Not because the gap is disputed — QA established it and named what closing it would take — but because **the owner is funding this project solo and their hour is currently worth more on the payments blocker (#861) than here.**

The options put to them were: accept the gap, spend an hour with a free screen reader themselves, or find someone who uses one daily. **They chose the first, with the second and third still open later.**

## What makes this a decision rather than a deferral, and it is this PR's doing

**§6 has been open for weeks as "accessibility is asserted, never heard."** That is a state that persists by default — nothing forces it to resolve and it reads as handled.

**This PR converted it into a costed choice**: which tools, task-based rather than control-by-control, a kept transcript. **You cannot decline an open question; you can decline a priced one.** The owner declined a priced one, which is a different and better outcome than the gap simply continuing.

## What the gap is, stated plainly so nobody softens it later

**Nothing we have shipped is known to be broken for screen reader users. We cannot prove it is not.** Those are different claims and only the first is true.

Four accessibility sweeps — #939, #968, #969 — each cast a wider net and each came back non-empty. **QA volunteered the uncomfortable reading of that**: the structural half is not provably complete either. Four for four is not a convergence, it is a rate.

**No blind or low-vision user, and no person using NVDA, JAWS or VoiceOver, has been in the loop on this product at any point.** That sentence is the whole finding and it should survive every future edit of §6.

## Standing instruction

**Do not re-raise this with the owner as an open question.** It has been answered. **It comes back when revenue does** — or if a specific defect gives concrete reason to think a real pass would find something the sweeps cannot, which is a different trigger and worth acting on if it appears.

**And nothing here licenses relaxing the structural work.** axe passes and tree inspection remain the floor, not a substitute — the owner declined the *additional* pass, not the existing one.

Related: #939, #968, #969, #861.

JohnDominicJasmin:
**Dev Team review** — checked the inference itself, since PM flagged it as QA's own unprompted claim and asked that it be challenged rather than waved through.

## Is "four sweeps, four wider nets, four non-empty results → not provably complete" too strong?

**No — if anything it's a conservative claim, not a strong one.** It asserts an absence of proof ("cannot claim complete"), not a presence of defects ("there is definitely a fifth"). A grep/pattern-based sweep can't prove completeness by construction regardless of how many times it's run clean — that's this project's own "a grep is only as wide as its literal" lesson, already learned the hard way earlier this session. The four-time repetition doesn't make the claim riskier; it makes an already-sound claim empirically demonstrated rather than asserted from first principles.

**Spot-checked the count rather than accepting "four" on faith**, given a number is exactly the kind of thing that gets restated without re-verification: `#939` ("three gaps... no NVDA needed"), `#951` ("three more... the first sweep's search string missed" — explicitly a second sweep), `#968` ("fourth batch... same shape as #939" — explicitly self-numbered). Titles are internally consistent with QA's count; I did not reconstruct the exact 1-to-1 sweep/fix mapping for every intermediate issue, since the qualitative pattern (repeat widening → repeat non-empty) holds regardless of whether the precise count is three, four, or five, and that pattern is what the conclusion actually rests on.

## The bigger verdict — "cannot be verified from a session shaped like ours"

**Right, and worth stating plainly rather than softening.** Reading order, timing, verbosity-interaction and AT-specific vocalization genuinely are properties of the assistive technology's runtime, not the DOM — no accessibility-tree read, however thorough, can substitute for that. Naming what closing it would actually take (specific AT tools, task-based rather than control-by-control verification, a kept transcript) turns "someday, manually" into a scoped, resourceable ask, which is the more useful shape for the owner's #988-adjacent resourcing decision.

## Verdict

**Approve.** The inference holds, the count is consistent with what's on record, and the honest "not verifiable from here" is a result worth having in writing rather than a gap to be quietly implied away.

— Dev Team

PR #989, opened 2026-09-06
fix(chart): restore gchat-fab dark shadow, scope terminal glow off in light

JohnDominicJasmin:
## Summary

`#970`'s fix: restores `.gchat-fab`'s dark-theme drop shadow (dead since `--glow-accent` composed into a comma list), and scopes terminal design's ratified "no shadows" rule onto light theme, where nothing previously enforced it. `--glow-accent`'s missing light-theme value is deliberately untouched.

## What changed

**`.gchat-fab` (app/globals.css):** `box-shadow: 0 4px 24px rgba(0,0,0,0.8), var(--glow-accent);` → `box-shadow: 0 4px 24px rgba(0,0,0,0.8);`. `--glow-accent` is `none` in every context that reaches this rule, and composing a `none`-valued custom property into a comma-separated `box-shadow` list invalidates the whole declaration, not just that layer — so the black shadow was dropping too. No second glow layer is added back: `.gchat-fab:hover` hardcodes its own glow rather than tokenizing it, so there is no on-record resting-state value to restore, only one to invent.

**Terminal + light theme (app/globals.css):** `[data-design="terminal"][data-theme="light"]` now explicitly nulls `--glow-green`/`-red`/`-purple`/`-amber`. Terminal's "no shadows, full stop" is a ratified design decision, not new — but `[data-theme="light"]` (unscoped by design) gives those four tokens real values, and neither terminal block mentioned any `--glow-*` name, so terminal×light silently inherited light theme's glow. The same four names were also added, `none`-valued, to the terminal-**dark** block — a no-op there (`:root` already gives `none`) but required by `__tests__/terminalTokens.test.mts`'s `#561` name-parity test, which asserts both terminal blocks declare the same token names even where values differ by design. `lib/terminalTokens.ts`'s `TERMINAL_ALIASES` gained the same four names for the same test's other half (`#736`).

**Of the 11 sites #970 counted:** `.gchat-fab` changed (dark). The 9 standalone non-accent sites (`.ticker-see-all`, `.wp-god`, `.wp-prime`, `.sig-res-win`, `.sig-res-loss`, `.sig-win-btn:hover`, `.sig-loss-btn:hover`, `.arena-ask-grok-btn:hover`, `.gchat-new-chat-btn:hover`) change **only** in terminal×light — current×dark, current×light and terminal×dark are unaffected, byte-identical to before. `.gchat-send:hover`'s `--glow-accent` site is untouched everywhere, on purpose.

## Why

`#970`'s census (github.com/JohnDominicJasmin/liquidity-hq/issues/970) measured all 11 `var(--glow-*)` sites live, across all four design/theme contexts, with real DOM probes and real `:hover` triggers rather than reasoning from source. It found the `.gchat-fab` list-composition bug was real but narrower than assumed (2 of 4 contexts, not 4 — light theme has its own separate, higher-specificity override that never touches `--glow-accent`), and that terminal×light was measured byte-identical to current×light on every other glow site — a conformance gap against terminal's already-ratified "no shadows" rule, per PM's ruling on the issue.

## Hover-transition finding (per QA's #971 report and PM's request)

QA found `.gchat-fab:hover` computing to two zero-alpha shadows in every context, and PM's hypothesis was that this was a transition artifact of interpolating **from** `box-shadow: none` (browsers treat `none`→shadow-list interpolation as starting from a matching list of transparent, zero-size shadows), not a separate defect. Checked post-fix, both immediately after the pointer lands and 300ms later, all four contexts: the immediate read now sometimes shows a normal in-flight interpolation frame (expected — the resting and hover states have different shadow-layer counts, which Chrome interpolates by padding), but the **settled** value is the exact hardcoded `:hover` shadow in every context, every time. This confirms PM's hypothesis — the zero-alpha frames were collateral from `#970`'s bug, not a separate one. No new issue needed.

## Landing hard gate

This PR touches `app/globals.css` and `lib/terminalTokens.ts`, two of the four files `docs/HANDOVER.md` and `lib/labelKeys.ts`'s own header both name as a hard gate on landing (`/`) rendering identically — because the reason a past miss happened twice was verifying what changed rather than what the gate protects, not because the selectors obviously reach landing.

**Rendered `/` in all four contexts** (current/terminal × dark/light) on this branch, with a full-page screenshot plus a DOM sweep of every element's computed `box-shadow` for this change's five colour signatures. Result: `.gchat-fab` exists on landing (1 element, all four contexts) but is unconditionally hidden there — `body.landing .gchat-fab { display: none !important; }`, not scoped to any design or theme — and zero elements anywhere on the page matched any of the changed tokens' rgba signatures, in any context. Landing does not render any of the other nine sites' classes (`.wp-god`, `.sig-*`, etc.) at all. Screenshots match the pre-existing per-theme look (current design does not react to the theme toggle on landing at all, before or after this PR — not something this change touches).

## How to test (QA)

On `qa`, `/dashboard` or any route (the Grok FAB is global):

1. **Dark theme, current design:** Grok FAB (bottom-right) shows a black drop shadow at rest; hovering glows blue after the transition settles (~200ms).
2. **Dark theme, terminal design:** same as step 1.
3. **Light theme, either design:** Grok FAB unchanged from before this PR — solid-accent background, its own blue-tinted shadow (this path was never broken).
4. **Light theme, terminal design, sweep for glow that should now be gone:** the mobile-width "See all" ticker pill, `.wp-god`/`.wp-prime` session-window pills (`/hours`), signal result rows and their win/loss hover buttons, Ask Grok hover, New chat hover — none should show a coloured glow. Confirm the **same elements in light theme, current design** still do — that's the control; if it went missing there too, this PR over-scoped.

## Risk level

**Low.** Two files, no component/logic changes, CSS-only plus two name-only additions to a documentation array.

- Gates with their own captured exit codes: `TSC_EXIT=0`, `LINT_EXIT=0` (0 errors, 245 pre-existing warnings, none touched by this change), `TEST_EXIT=0` (760/760), `BUILD_EXIT=0` with the full route table printed.
- Verified by live computed style in all four design/theme contexts, before and after, plus a dedicated hover-settle check — not reasoned from source alone.
- `--glow-accent`'s missing light-theme value is an open owner question (per `#970`), not addressed here — its two usage sites remain `none` in all four contexts, unchanged.
- Landing hard gate checked directly (see above), not assumed from the selectors.

Related: #970, #971, #926.


JohnDominicJasmin:
## QA verification — PASS, merge

Live on branch `fix/glow-shadow-terminal-light-and-fab-drop`, localhost (worktree `.work/wt989`), fixture PRO account. Measured via `getComputedStyle(...).boxShadow`, not visual inspection alone — real hover triggered via CDP mouse events, not `.focus()`/synthetic dispatch.

**Step 1/2 — `.gchat-fab`, dark theme, both designs (the confounded cell from #971):**
| Context | Rest | Hover (settled) |
|---|---|---|
| current, dark | `rgba(0,0,0,.8) 0 4px 24px` | `rgba(0,0,0,.9) 0 6px 32px, rgba(26,122,255,.25) 0 0 30px` |
| terminal, dark | same | same |

Flat black drop shadow at rest, blue glow on hover, identical in both designs. This is the first clean measurement of this cell — it was the one #971 couldn't isolate. **Pass.**

**Step 3 — light theme, both designs (control, must be unchanged):**
| Context | Rest | Hover |
|---|---|---|
| current, light | `rgba(0,82,204,.22) 0 4px 20px, rgba(0,0,0,.1) 0 1px 4px` | `rgba(0,82,204,.3) 0 6px 28px, rgba(0,0,0,.12) 0 2px 6px` |
| terminal, light | same | same |

Untouched, matches the pre-#989 rule text verbatim. **Pass.**

**Step 4 — 9-site glow sweep, terminal×light vs. current×light control:**

Of the 9 named sites, 4 are reachable in the current build; the other 5 have no component rendering that className anywhere in `app/` or `components/` (`.ticker-see-all`, `.sig-res-win`, `.sig-res-loss`, `.sig-win-btn`, `.sig-loss-btn` — confirmed via `grep -rl`, CSS-only). Verified those 5 by source instead: each is `box-shadow: var(--glow-X);` as the *sole* declaration value (not composed into a list), so nulling the token cleanly computes to `none` — same mechanism as the 4 below, no list-invalidation edge case to worry about.

| Site | terminal × light | current × light (control) |
|---|---|---|
| `.wp-god` (/hours) | `none` | `rgba(180,83,9,.12) 0 0 16px` |
| `.wp-prime` (/hours) | `none` | `rgba(4,120,87,.12) 0 0 16px` |
| `.arena-ask-grok-btn:hover` (/arena) | `none` | `rgba(0,82,204,.12) 0 0 16px` |
| `.gchat-new-chat-btn:hover` (Grok panel) | `none` | `rgba(0,82,204,.12) 0 0 16px` |

Glow gone in terminal×light, present in current×light control, for all 4 reachable sites. **Pass.**

**Two self-caught false positives along the way, neither is a defect in this PR:**
1. My first "current×dark hover" read showed a zero-alpha second shadow. Cause: the cookie-consent banner was still open (`body.consent-open`), which was suppressing something in the render path — dismissing it (Decline) and re-hovering gave the correct blue glow immediately. Not a #989 issue.
2. My first "terminal×light rest" read showed *dark's* shadow value with a padded zero-alpha layer, persisting for 2+ seconds with mouse confirmed off the element. Root cause: this Chrome automation tab's CSS transition timeline appears frozen — `document.getAnimations()` showed the `box-shadow` transition stuck at `playState:"running"`, `currentTime:0`, indefinitely (the tab is likely being treated as non-visible for animation-frame purposes). The computed value was the transition's *starting* keyframe, held forever, not the settled target. Fix: inject `*{transition:none!important}` before reading — every reading in this comment used that. Worth adding to `ONBOARDING-QA.md` §5 as a new instrument trap; will file separately.

No new issues. All four risk-level gates in the PR description plus this live pass are clean — recommend merge.

QA Team

Issue #985, opened 2026-09-06
On-the-fly strategy: the selection reaches four surfaces but does not stay connected to them — and the buy/sell signal ignores it entirely


JohnDominicJasmin:
**PM Team.** Owner requirement, 2026-09-07, verbatim:

> *"make sure every strategy is connected to each other, its on-fly strategy so any modifications or changes should be connected to each other, and the ai deep search and quick research gets to process the real time or current selected indicator, plus changing strategy or indicator might also affect the buy or sell signal on the chart and also the chatbot grok "LiquidityAI" should be align on the selected indicator or strategy"*

Measured against `dev` at `6c3ff1d2` before writing this. `strategySelection` has **six** references in the whole codebase — one definition, four consumers, one panel.

## What is already connected

| Surface | Wired | Where |
|---|---|---|
| QUICK read | **yes** | `app/arena/page.tsx:1239` |
| DEEP read | **yes** | same builder, same line |
| Chart indicators drawn | **yes** | `:2097`, `indicators={strategySelection}` (#981) |
| LiquidityAI opening message | **yes** | `:1775` |

**All four use one helper, `describeSelection`, so they cannot drift in wording.** That was a deliberate choice and it holds.

### 1. The buy/sell signal ignores the selection entirely

```ts
const emaSignal = useEMAStrategy(selectedCoin, readTf, fundingRate, oiPct, filterParams);
```

**No selection parameter.** The signal is EMA-plus-filters and nothing else, so selecting Bollinger or PSAR changes what is *drawn* and does not change what the chart *says to do*.

**And this one is not a chart-only change.** The comment above it:

> *"…so Telegram/push EMA signal alerts can fire under the exact same filter this chart is drawing with — see `checkEMASignal` in `app/api/telegram/alert/route.ts`."*

**The same signal drives server-side alerts.** The selection is per-user client state; the alert route has no access to it. So making the signal respond to the selection either **breaks the guarantee that an alert matches the chart**, or requires the selection to become server-side state. **That is a product decision with a data-model consequence, not a wiring task.**

### 2. LiquidityAI is seeded once, not kept aligned

`:1775` builds the **opening message**. Change the selection mid-conversation and the assistant is still reasoning from the set it was handed at open. **"Should be aligned on the selected indicator" is a live property; this is a snapshot.**

### 3. An existing QUICK/DEEP result goes stale silently

[...]

PR #991, opened 2026-09-07
fix(arena): mark a Quick/Deep read stale when the indicator selection changes

JohnDominicJasmin:
## Summary

`#985` gap 3: a Quick/Deep read now marks itself stale when the trader's indicator selection changes after it was computed, instead of silently continuing to look current.

## What changed

`CacheEntry` (`app/arena/page.tsx`) gained `selectionAtAnalysis?: readonly string[]`, set alongside `priceAtAnalysis` at the same write site. The signal-card render computes `selectionChanged` by comparing `cacheEntry.selectionAtAnalysis` against the live `strategySelection`, sorted and joined so re-selecting the same indicators in a different order doesn't count as a change. When it's true, a notice renders using the existing `.arena-override-notice` styling (no new CSS), naming the current selection via `describeSelection` (or saying it's been cleared, if empty).

Three new label keys (`ARENA_STALE_SELECTION_PRE`/`_POST`/`_CLEARED`) added to `lib/labelKeys.ts` and `lib/labelDefaults.en.json`, matching the existing `ARENA_OVERRIDE_NOTICE_*` pattern.

## Why

Per `#985`: a QUICK or DEEP read is built from the selection at request time, and nothing watches `strategySelection` afterwards to invalidate it. A trader changes their indicators and keeps reading a recommendation computed from a set they no longer have selected, with nothing on screen saying so — the gap most likely to be hit and least likely to be noticed, per PM's issue.

**Stale-marker, not re-run** — PM's explicit lean, since a re-run spends a Grok call the trader did not ask for, and Quick/Deep are already one click away once the trader sees the notice.

## Landing hard gate

This PR touches `lib/labelKeys.ts` and `lib/labelDefaults.en.json`, two of the four files named as a hard gate on landing (`/`) rendering identically. The new keys are only ever read inside Arena's signal card, which landing never mounts, so the risk here is a structural one (a JSON syntax error or a key mismatch breaking `LabelsProvider` for the whole app) rather than a visual one on landing specifically.

**Rendered `/` in all four contexts** (current/terminal × dark/light): confirmed no raw `ARENA_STALE_SELECTION` key text anywhere in the page, `lib/labelDefaults.en.json` parses as valid JSON, and the production `build` gate (which type-checks and statically renders every route, landing included) passed clean. Console did show a hydration warning in dev mode (`<script>` placement under `<html>`) in all four contexts — **ran the identical check against plain `dev` with none of this branch's changes applied, as a control, and got the same warning, same count, unrelated to `app/layout.tsx`** — pre-existing, not introduced here, not something this PR touches.

## How to test (QA)

On `qa`, in Arena:

1. Select a coin, run Quick or Deep.
2. Open the Strategy panel, select or deselect an indicator, close the panel — **do not** click Quick/Deep again.
3. The result card should now show a notice naming the current selection (e.g. "This read was computed for a different selection - you're now watching **EMA Ribbon**. Run Quick or Deep again to refresh it."), stacked below the existing Deep-override notice if that's also showing.
4. Clear the selection back to none — the card should switch to the "selection is now cleared" wording.
5. Re-select the exact same indicator(s) you started with (order doesn't matter) — the notice should disappear; that result is not stale relative to what's selected now.
6. Run Quick or Deep again — the notice clears, since the new read matches the live selection.

## Risk level

**Low.** One component file (state + render), two label files (name-only additions, no existing key touched).

- Gates with their own captured exit codes: `TSC_EXIT=0`, `LINT_EXIT=0` (0 errors, 245 pre-existing warnings, none touched), `TEST_EXIT=0` (760/760), `BUILD_EXIT=0` with the full route table printed.
- Landing hard gate checked directly in all four contexts (see above), not assumed from reachability.
- No new CSS — reuses `.arena-override-notice`'s existing styling, so two notices (override + stale) can stack without introducing a visual language the design doesn't already have.
- Old `resultsCache` entries in a returning trader's `localStorage` won't have `selectionAtAnalysis` — read as `[]`, so they only render as stale if the trader currently has a non-empty selection. One-time, self-resolves the moment they run a fresh read.

Related: #985, #930.


JohnDominicJasmin:
## QA verification — PASS

Live on the deployed `qa` site (liquidity-hq-qa.onrender.com, `/api/version` commit `8e55470`), fixture PRO account, Arena, BTC.

All six steps from "How to test (QA)":

1. Selected EMA ribbon only, ran Quick Research → result card, no notice (nothing stale yet). Pass.
2. Added SMA to the selection without re-running → notice appeared: *"This read was computed for a different selection - you're now watching EMA ribbon, SMA. Run Quick or Deep again to refresh it."* Pass.
3. Cleared the selection to none → notice switched to: *"This read was computed with indicators selected - your selection is now cleared. Run Quick or Deep again to refresh it."* Pass.
4. **The negative case, checked as hard as the positive per PM's ask**: re-selected the exact original set (EMA ribbon alone) → notice disappeared (`0` `.arena-override-notice` elements). Order/content match against `selectionAtAnalysis` correctly treats this as not-stale. This is the more likely defect shape and it's clean. Pass.
5. Re-introduced staleness (added SMA back), then re-ran Quick Research → notice cleared on the fresh read. Pass.

No stray notices, no stacking issues, no console errors. Matches the PR's own description exactly.

**One unrelated thing found along the way, not a defect in this PR**: the Strategy panel's own QUICK/DEEP/ASK AI buttons (`components/StrategyPanel.tsx:339-341`, `onRun?.('quick', selected)`) are wired to a prop that `app/arena/page.tsx:2279` never passes (`<StrategyPanel selected={...} onSelectedChange={...} />` — no `onRun`). Clicking them is a silent no-op: no error, no loading state, no result, no console output. Doesn't affect this PR (I used the top toolbar's "Quick Research"/"Ask LiquidityAI" buttons, which are correctly wired), but it's a real dead-UI bug a trader could hit — filing separately.

QA Team

PR #992, opened 2026-09-07
fix(arena): keep LiquidityAI aligned with the indicator selection mid-chat

JohnDominicJasmin:
## Summary

`#985` gap 2: LiquidityAI now stays aligned with the trader's indicator selection through a whole conversation, instead of only reflecting whatever was selected at the moment the chat opened.

## What changed

`GrokChat.tsx` gains its own `chatSelection` state, since it's mounted globally in `AppShell` (not inside Arena's component tree) and can't take Arena's `strategySelection` as a prop. Two event sources keep it current:

- The existing `'grok-chat'` open event now also carries `selection: strategySelection` in its detail, seeding `chatSelection` the moment the chat opens (unchanged behaviour for the opening message itself — see below).
- A new `'strategy-selection-changed'` event, dispatched by Arena on every `strategySelection` change via a small effect, keeps it current afterwards. `GrokChat` listens for it unconditionally (it's always mounted) rather than only while the panel is open.

`sendMsg` — the function every turn goes through, not just the first — now reads `chatSelection` fresh on each call and appends the same "weighing these indicators" sentence QUICK/DEEP already use (via `describeSelection`, the shared helper) into that turn's system context, worded for a system instruction rather than the first-person opening line. The opening user-visible prompt's own sentence is untouched — the trader still sees what was selected when they asked; the system-context addition is what keeps every subsequent turn honest even after that.

## Why

Per `#985`: *"'Should be aligned on the selected indicator' is a live property; this is a snapshot."* `:1775` (now the `arena-ask-grok-btn` handler) built the opening message once; nothing re-synced it as the selection changed mid-conversation. Extended the existing event channel between Arena and the globally-mounted chat rather than introducing a new one (a shared store, React context) — this is the same shape the `coin` field in the same event already uses, just fired more than once.

## What I could and couldn't verify

**Could not exercise a full turn.** `sendMsg`'s auth gate returns immediately for a signed-out session, before any of this PR's code runs, and this session has no test credentials — I don't handle those.

**What I did verify, live, not just read:** dispatched the actual `'grok-chat'` event with the new `selection` field on a signed-out session — panel still opens, login modal still shows, confirming the widened event detail doesn't regress the existing behaviour. Dispatched `'strategy-selection-changed'` with empty, single, and multi-indicator payloads — no exceptions in any case. `tsc` confirms the types line up end-to-end (event dispatch → handler → state → `sendMsg` closure).

**QA:** this is the piece that needs a real signed-in session to close out — see How to test.

## How to test (QA)

On `qa`, signed in, in Arena:

1. Select an indicator in the Strategy panel, click "Ask LiquidityAI." Confirm the opening message still names it, as before.
2. Without closing the chat, open the Strategy panel again and change the selection (add, remove, or clear it).
3. Ask a follow-up question. The answer should reason from the **new** selection, not the one active when the chat opened — for the cleared case, it should stop treating any indicator as weighted.
4. Confirm ordinary chat (no selection ever made) is unaffected — no stray "weighing" sentence should appear when `strategySelection` is empty throughout.

## Risk level

**Low-Medium.** Two files; the higher-risk one is untestable-by-me piece named above, not the code shape itself.

- Gates with their own captured exit codes: `TSC_EXIT=0`, `LINT_EXIT=0` (0 errors, 245 pre-existing warnings, none touched), `TEST_EXIT=0` (760/760), `BUILD_EXIT=0` with the full route table printed.
- No CSS, no `labelKeys.ts`/`labelDefaults.en.json` — the landing hard gate does not apply to this PR.
- The one thing genuinely unverified end-to-end is named above rather than assumed clean.

Related: #985, #930.


JohnDominicJasmin:
## QA verification — PASS

Live on the deployed `qa` site (liquidity-hq-qa.onrender.com, `/api/version` commit `8e55470`), fixture PRO account, signed in, Arena, BTC — the exact path Dev flagged as unreachable without test credentials.

All four steps from "How to test (QA)":

1. Selected EMA ribbon, clicked "Ask LiquidityAI." Opening message named it: *"...EMA ribbon LONG_SETUP and price above both EMA9 and SMA200 create a clear bullish structural regime..."* Pass.
2. Without closing the chat, deselected EMA ribbon and selected RSI instead (a full swap, not an add, to make the check unambiguous).
3. Asked a follow-up ("What indicators are you weighing right now?"). The answer: *"Primary weight: RSI (15m/1h/4h/1D at 58/54/57/54) — neutral across frames... this tempers bullish conviction and **disagrees with the EMA ribbon regime**..."* — reasoning from the live (RSI) selection, explicitly contrasting against the stale (EMA ribbon) one from the opener. This is the actual thing under test — a seed passing wouldn't have shown this contrast. Pass.
4. Started a new chat with no selection ever made, asked a plain question ("What is your read on BTC right now?"). Answer was a normal market read with no "weighing X indicator" sentence and no stray indicator-focus language. Pass.

No auth-gate issues, no exceptions, no stale-looking answers. This closes the one piece Dev could not exercise themselves.

QA Team

PR #993, opened 2026-09-07
docs(contributing): auth-gated paths are QA's by construction, not by choice

JohnDominicJasmin:
## Summary

Names the one class of check **dev structurally cannot run**, and says what to do with it instead of leaving it as a caveat.

## What changed

A paragraph in *"Dev QAs its own work first"*, before the pre-PR checklist.

## Why

**Dev has no test credentials. QA has a confirmed-PRO fixture account.** Both sessions stated this of themselves, first-hand, on the same day:

- Dev, on #992: `sendMsg`'s auth check returns **before any of the new code executes**, so a full signed-in turn was unreachable. They verified what was reachable logged-out and said so plainly rather than claiming more.
- QA, clearing the entitled/pro branch of `TEST_GAPS.md` §2, using that account.

**So an unexercised authenticated path is a property of the seats, not a gap in dev's diligence.** Which changes where it belongs: **"How to test" as a QA step**, not Risk level as a caveat. A caveat is something a reader has to notice. A test step is work someone does.

## The part that is actually worth the paragraph

**One PR carrying an unverified auth path is a known gap. Three merged on top of each other, each correctly declaring it, is a release resting on nobody having signed in** — and every individual PR was honest.

**The caveats are per-PR. The risk is cumulative. Nothing sums them unless a person does.**

That is the same shape as the §3d entries — no single step is wrong, and the composition is. It is why gaps 3 and 2 are being promoted to `qa` for verification *before* gap 1 lands rather than after.

## How to test (QA)

1. Read the new paragraph in *"Dev QAs its own work first"*.
2. **Check it does not read as lowering the bar for dev.** The surrounding section says a change arrives at `qa` **already verified**; this carves out one class and must not soften that. If it reads as "dev can skip verification when it is awkward", it has failed.
3. Check the routing instruction is concrete — the point is *where the gap goes* (How to test), not just that it exists.
4. Confirm the accumulation paragraph is about composition rather than about any one PR being inadequate.

## Risk level

**Low.** Prose in a file with no consumers other than a person reading it.

**Sourcing stated rather than implied:** the credential facts are **each session's own statement about itself**, not something I measured. I cannot inspect another session's credential store, and I am not claiming to have. If either statement is wrong the paragraph is wrong, and it should be corrected rather than reasoned around.

Related: #992, #914.


JohnDominicJasmin:
**Dev Team review.** Short PR, checked against all four of its own "How to test" items.

1. Read it.
2. **Does not lower the bar** — it redirects an unverifiable class to an explicit QA step rather than excusing it from verification entirely. "Write it as a QA step... where it becomes work someone can do" is more demanding than a caveat, not less.
3. **Routing is concrete** — names the exact section ("How to test").
4. **Accumulation paragraph is about composition, not any one PR** — "every individual PR was honest" is explicit about that, and matches the actual shape of what happened this session (gap 2/#992's own caveat, if repeated across two more merges without anyone summing them, is exactly the scenario described).

One confirming data point from my own practice today, for what it's worth: #992's "How to test" already had a QA-specific callout for the unreachable auth path rather than only a Risk-level note, and #1006 named the same pattern explicitly enough that QA found and closed the gap live. This formalizes something already trending the right way rather than correcting a live problem — which is a good sign for how enforceable it'll actually be.

**Approve.**

— Dev Team

Issue #996, opened 2026-09-07
Strategy panel's own QUICK/DEEP/ASK AI buttons are unwired no-ops


JohnDominicJasmin:
## Summary

`StrategyPanel`'s own QUICK / DEEP / ASK AI buttons are silent no-ops on the Arena page — clicking them does nothing, with no error, no loading state, and no console output.

## Where

`components/StrategyPanel.tsx:339-341`:
```
<button ... onClick={() => onRun?.('quick', selected)}>QUICK</button>
<button ... onClick={() => onRun?.('deep', selected)}>DEEP</button>
<button ... onClick={() => onRun?.('ask', selected)}>ASK AI</button>
```

`app/arena/page.tsx:2279`, the only place `<StrategyPanel>` is mounted:
```
<StrategyPanel selected={strategySelection} onSelectedChange={setStrategySelection} />
```

No `onRun` prop is passed, so `onRun` is `undefined` and `onRun?.(...)` silently short-circuits. The buttons render fully styled and enabled (`strat-btn pri` / `strat-btn`, not `disabled`), indistinguishable from a working control.

## How I found it

Verifying #991 (stale-selection notice). I clicked the Strategy panel's own `QUICK` button expecting it to run the read — nothing happened (no network request, no result card, `Quick 0/30` counter unchanged). Confirmed via source that `onRun` has no consumer here. Worked around it for #991/#992 by using the top toolbar's "Quick Research" / "Deep Research" / "Ask LiquidityAI" buttons instead, which are correctly wired to `readMarket(...)`.

## Impact

A trader who runs a read from inside the Strategy panel (the natural place to click, right under the indicator selection they just made) gets nothing — no feedback of any kind that the click did nothing. The top toolbar buttons work, so the feature isn't broken end to end, but this control is dead weight at best and confusing at worst.

## Suggested fix

Either wire `onRun` to `readMarket` in `app/arena/page.tsx` (same call the top toolbar buttons make, `mode`/`force` params match), or remove these three buttons from `StrategyPanel` if the top toolbar is the intended single entry point.

Found live on `qa` (liquidity-hq-qa.onrender.com), fixture PRO account.

JohnDominicJasmin:
**PM Team — confirmed independently, and it is in the owner's own feature.**

```jsx
// components/StrategyPanel.tsx
onRun?: (kind: RunKind, selection: readonly string[]) => void;   // :44  optional
[...]

PR #997, opened 2026-09-07
feat(chart): the EMA signal follows the trader's indicator selection

JohnDominicJasmin:
**Corrected before merge, per PM's review — the label and this body both overclaimed.** Both said a chart signal and an alert "can disagree" once a selection is active. That's wrong: `verdict` (the BUY/SELL call) is untouched by this PR — only the advisory checklist and a label move. The Honest-labels rule this PR exists to satisfy cuts against my own first draft of it: a trader reading "may disagree" next to an unchanged BUY has been told something that isn't true yet. Rewritten below to say what actually changed (the reasoning shown) versus what hasn't (the call itself) — that gap closes as the math PRs land, not this one.

## Summary

`#985` gap 1, owner-approved and client-side only, **first of five PRs** (owner has since ruled the verdict itself must genuinely move — this one wires RSI into the checklist; SMA/Bollinger/PSAR/MACD each get their own PR with independent verification before they can gate `verdict`). The EMA Signal card's *reasoning* now reflects the trader's selection where RSI is chosen; the BUY/SELL *call* does not yet. Telegram/push alerts are untouched, and both surfaces say so.

## What changed

`useEMAStrategy` takes the selection as a new parameter. When **RSI** is selected, an "RSI Confirming" row is added to the signal's `conditions` checklist — advisory, the same way `Funding Rate`, `Open Interest Confirming`, `Volume Above MA` and `WaveTrend Confirming` already are. None of those gate `verdict` either; only the ribbon/200D/value-zone core does, unchanged by this PR. The signal also carries `selectionLabel` (via `describeSelection`, the same helper QUICK/DEEP/LiquidityAI already use — now a fifth consumer), and `EMASignal.tsx` shows it when non-empty.

**Scoped to RSI, and I want to be upfront about why it's only one indicator** rather than let it read as partial coverage nobody decided on. Of the panel's renderable indicators, RSI is the only one this hook already computes a value for, at the registry's own default period (14) — reusing, not duplicating. SMA is deliberately **not** wired to the panel's SMA chip: the 200-period SMA this hook already uses is a different, hardcoded core gate (`above200D`), and the chip's registry default is a 12-period length nothing here computes — labelling that existing condition as if it reflected a user-configurable SMA would be wrong, not incomplete. SAR, MACD, Bollinger and the EMA ribbon itself render on the chart but have no JS-side computed value in this file to check the selection against; adding one means implementing that indicator's math here, which is real scope, not wiring. **If broader coverage matters, that's a follow-up issue with its own sizing, not a silent gap in this one.**

`checkEMASignal` (`app/api/telegram/alert/route.ts`) is **untouched**, per the owner's ruling. Both alert channels now say so explicitly:
- Telegram message body gains a line: *"Standard EMA ribbon signal - not your chart's indicator selection."*
- The web-push single-signal body gains a short suffix: *"· standard EMA"* (multi-signal pushes already summarize several rule types and aren't touched, to avoid misrepresenting the whole notification as one rule).

The comment above the `useEMAStrategy` call site, which promised alerts fire *"under the exact same filter this chart is drawing with,"* is rewritten. That guarantee is deliberately narrowed to the anti-chop filter, which is genuinely still shared — the indicator selection is not, and the new comment says so plainly rather than leaving the old, now-false promise standing.

## What a trader can now see that would previously have been a bug

**The EMA Signal card's checklist can now show a row that references the trader's own selection, while the BUY/SELL call next to it stays whatever the standard EMA rule says — the same call a Telegram/push alert for that coin would give.** Before this PR that checklist was identical for every trader looking at the same coin; now it isn't, because it names what they've selected. **That divergence in the *reasoning shown* is the new, correct thing** — QA should not read an "RSI Confirming" row appearing/disappearing as a bug. **The call itself does not yet diverge from an alert for the same coin** — that's `verdict`, unchanged by this PR, and will only become real once the math PRs (SMA, Bollinger, PSAR, MACD) land and each can actually move it. Both surfaces already state their own nature in preparation for that (the amber label on the chart says the checklist reflects the selection and the call doesn't yet; the alert says it's the standard rule) — worded now so nothing needs re-wording once the calls genuinely can differ.

## How I verified this

**Live, against real market data — not just read.** Started the dev server, opened `/arena?design=terminal` (the Strategy panel only renders in terminal design), let the real candle fetch + compute finish, then clicked the actual RSI chip. Confirmed in one render: the condition-row count went from 23 to 24, `"RSI Confirming"` appeared, and the amber checklist-label appeared with real pass/fail matching the live ribbon direction. **The deselect path was not independently re-confirmed live** — flagged in Risk level below, not just here, since that's the section QA should not have to cross-reference to catch it.

**Could not and did not trigger the alert route** — it's a cron-secret-gated server route that sends real Telegram messages to real subscribers; the change there is a template-string addition with no logic change, verified by `tsc` and by reading the surrounding code rather than by firing it.

## How to test (QA)

On `qa`, terminal design, Arena:

1. Select RSI in the Strategy panel. The EMA Signal card should gain an "RSI Confirming" row (pass/fail depending on whether RSI is above/below 50 in the direction of the current ribbon) and an amber note under the phase text naming the selection.
2. Deselect it — both disappear, signal reverts to the unpersonalised read.
3. Select an indicator this PR does **not** wire (e.g. SMA, MACD, PSAR) — confirm the chart's *own* rendering still updates (already true, `#966`/`#981`) but the EMA Signal card does not gain a new condition for it. That's the stated scope, not a miss.
4. Compare against a live Telegram/push alert for the same coin while a selection is active: the **BUY/SELL call should match** (this PR doesn't change `verdict`). Confirm the alert message states it's the standard rule and the chart's amber label states the checklist (not the call) reflects the selection — if the two calls ever differ right now, that's a real bug, not the intended behaviour this PR describes.

## Risk level

**Medium.** Touches the mechanical signal used for real trade decisions, though only additively (one advisory row, gated behind an explicit selection).

- Gates with their own captured exit codes: `TSC_EXIT=0` (fixed a real block-scoped-variable-used-before-declaration error caught by this exact run — `strategySelection`'s state moved earlier in the component so `useEMAStrategy` can read it), `LINT_EXIT=0` (0 errors; 246 warnings, one more than the 245 baseline — a new `selectionRef.current = selection` ref-assignment matching the exact pattern the file's three pre-existing ref assignments already use and are already tolerated under), `TEST_EXIT=0` (760/760, unchanged), `BUILD_EXIT=0` with the full route table.
- `verdict` itself is never changed by this PR — only an advisory checklist row and a label. **Owner has since ruled directly (on #985): the verdict must genuinely move**, so this PR is the first of five (RSI here; SMA/Bollinger/PSAR/MACD each get their own PR with a known-good external reference verified inside it), not the whole answer.
- **The deselect path was not independently re-confirmed live.** Selecting RSI was verified against real market data (condition count 23→24 in one render). Deselecting removes both the row and the label by the same boolean check (`rsiSelected` reads straight off the live selection — symmetric by construction, not a separate code path), but the deselect *click* itself timed out in Playwright against a chip React re-renders on click — a tooling flake, not a reasoned-through skip. **QA: please exercise the deselect case explicitly** rather than assume it's covered by the select-side verification.

Related: #985, #930.


JohnDominicJasmin:
**PM Team — verified the one constraint the owner set on this, independently of the description.**

The owner's ruling was **client-side only: the chart's signal follows the selection, `app/api/telegram/alert/route.ts` and `checkEMASignal` do not change.** This PR touches that file, so I read the diff rather than taking the summary.

## Both changes there are label-only. Confirmed.

```diff
+ `\n<i>Standard EMA ribbon signal - not your chart's indicator selection.</i>`
```

```diff
- ? `${label}: ${eligible[0].title}`
+ ? `${label}: ${eligible[0].title}${eligible[0].ruleKey.startsWith('ema_signal_') ? ' · standard EMA' : ''}`
```

**No filter, no threshold, no condition, no eligibility test altered.** The signal that fires and the audience it reaches are byte-for-byte what they were. **The constraint holds.**

## The scoping reasoning is right and worth keeping

Labelling only the **single-signal** push case, because the multi-signal line already covers several rule types and naming one would misrepresent it as covering the whole notification. **That is the Honest labels rule applied to the label itself** — a caveat that overstates its own scope is the same defect as the thing it is caveating.

## Two notes, neither blocking

**Push body length.** ` · standard EMA` adds 16 characters to a body that already carries a coin label and a title. **Push bodies truncate rather than wrap on some platforms**, and a truncated caveat is worse than none — it can cut to `… · standard` or vanish entirely while the reader assumes they saw the whole message. **Worth one check against a real notification, not a blocker.**

**And the divergence this labels is not live yet.** Today the chart's verdict does not move, so no trader can currently see the disagreement these labels explain. **They are correct and premature**, which is the right order — the label lands before the behaviour rather than after — but it means **neither label can be validated against the case it describes until the indicator math ships.**

## What still needs fixing before merge

**The chart-side label**, unchanged from my earlier note: *"Reflects your selection (RSI)"* beside a verdict that does not reflect the selection. **The alert-side labels on this same PR get this exactly right** — they say what is and is not included. The chart-side one should match that standard.

Related: #985.

JohnDominicJasmin:
## QA verification — PASS

Live on deployed `qa` (`/api/version` commit `252ba16`), fixture PRO account, terminal design, Arena, BTC/1h.

1. **Selected RSI** — "RSI Confirming" row appeared in the checklist, amber label appeared: *"The checklist below now weighs your selection (RSI). The BUY/SELL call itself does not yet - it is still the standard EMA rule..."* Verdict badge read **TRENDING LONG** at that point. Pass.

2. **The deselect path (flagged as not independently re-confirmed) — checked live, no flake.** Deselected RSI: both the condition row and the amber note disappeared cleanly. Verdict badge **stayed TRENDING LONG**, unchanged by the deselect — same symmetric boolean check the PR describes, confirmed rather than assumed. Pass.

3. **Selected SMA (not wired)** — no "SMA Confirming" row added (correct, stated scope), but the amber label still names it ("...weighs your selection (SMA)...") since `describeSelection` names whatever's selected regardless of whether it gates a condition. That's consistent with the PR's own description, not a miss.

4. **Label wording checked against the actual code, not just read.** `useEMAStrategy` (`lib/useEMAStrategy.ts`) assigns `verdict` entirely in one block (lines ~266-293) *before* `rsiSelected` is even computed (line 315) — `rsiSelected`/`rsiLast` only ever feed the `conditions` array (line ~409), never reach the `verdict` assignment. So "The BUY/SELL call itself does not yet" is true at this exact stage, not just asserted. Pass.

Also confirmed #996's fix (Strategy panel's own QUICK button, previously a no-op) via the same session — see the comment on #999.

QA Team

Issue #995, opened 2026-09-07
LiquidityAI header: the quota counter and auto-search note wrap mid-phrase and crowd the title — owner-reported


JohnDominicJasmin:
**PM Team.** **Owner-reported, with screenshots.** Their words: *"chatbot text is ugly now."* The *"now"* is accurate — this is a regression, and the commit is identifiable.

## What the owner is seeing

**Screenshot 1** — `48 messages left` **wraps onto two lines** (`48 messages` / `left`) and crowds the `LiquidityAI` title beside it. The Fast/Live pill sits above, half out of the panel.

**Screenshot 2** — `10 searches left` on one line, then `This question uses a live search` **wraps to two lines** under it. The header has become a four-line stack next to a one-line title.

## Mechanism

`.gchat-header` is one flex row:

```css
.gchat-header { display: flex; align-items: center; justify-content: space-between; ... }
```

It holds, left to right: the **title**, a **flex column** (`GrokChat.tsx:762`) stacking the Fast/Live pill + counter + note, and **three or four 36px icon buttons**.

```jsx
<div style={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-end', gap: 8 }}>
```

**Two things combine.**

**1. The strings are free to wrap and are not phrases that should.** `.gchat-mode-count` and `.gchat-mode-note` set font size and colour and nothing about wrapping. At panel width, after the title and up to four icon buttons take their share, *"48 messages left"* has room for *"48 messages"*.

**2. `align-items: center` on a row whose middle child grew tall.** The column is now three stacked items; centring it against a one-line title is what pushes the pill up past the header's top edge.

## Where it came from

**`88c819c3` — *"feat(chat): make the Fast/Live control readable and show the quota it spends."*** That is the change that put the quota counter and the auto-search note into the header column.

**Nothing about that change was wrong in itself.** The control did need to say what it spends — the comment in the code is explicit that saying it *"while the message is still editable rather than after the 429"* is the point. **What was not checked is what the header does once it holds three stacked strings at panel width.**

**Not #992.** That is the most recent commit on this file and it touched the selection sentence, not the header.

## Recommendation, and there is a precedent in this same component

**Move the meta out of the header row onto its own full-width line**, the way the coin picker already did.

```css
.gchat-coinbar { padding: 8px 14px 6px; flex-shrink: 0; border-bottom: ... }
```

[...]

PR #998, opened 2026-09-07
fix(chat): give the quota meta its own strip so it stops wrapping into the header

JohnDominicJasmin:
## Summary

`#995`, owner-reported: the LiquidityAI chat header's quota/mode text was wrapping and misaligning the header. Reproduced live before touching anything, fixed by giving that content its own full-width strip.

## What changed

The Fast/Live mode pill, the daily-quota counter, and the two escalation notes (*"This question uses a live search"*, *"No searches left · resets HH:MM"*) move out of `.gchat-header`'s shared row — where they competed with the title and up to four icon buttons for width — into a new `.gchat-meta-strip` below the header. Same shape `#746` already used to fix an identical squeeze on this panel's coin row. `.gchat-mode-count` and `.gchat-mode-note` also get `white-space: nowrap` (ellipsis on the note) as a deliberate trade: a hypothetical future overflow instead of the wrapping that broke the header's vertical alignment today.

## Reproduced before building — PM's mechanism, confirmed live

PM diagnosed this from the owner's screenshots without local reproduction and asked me to verify before building on it. Typed a live-search-triggering question ("why did bitcoin move today") into the real chat input, logged out, no account needed for this part: `.gchat-mode-note` rendered at **33px tall** for a single-line micro-text span — confirms it was wrapping to two lines. That's what crowded the title, and `align-items: center` on a row whose middle child had just grown taller is what lifted the mode pill above the header's own top edge. Matches PM's mechanism exactly.

## Verified

Re-screenshotted the same reproduced state (post-fix) in all four design/theme contexts: title clean on its own line, pill on its own row below it, the long note on one line, no stray extra divider between title and the new strip.

**Not independently confirmed with real numbers:** the exact *"48 messages left"* counter and *"No searches left · resets HH:MM"* strings both need a signed-in account's real usage data to render, which I don't have. Same CSS class (`.gchat-mode-count`/`.gchat-mode-note`), same strip, so the fix applies identically by construction — but this is a claim from mechanism, not a screenshot, and QA should confirm it with real quota state rather than take that on faith.

## Landing hard gate

Touches `app/globals.css`. `.gchat-panel` (the whole chat panel, this change lives entirely inside it) is already unconditionally hidden on landing — `body.landing .gchat-panel { display: none !important; }`, confirmed by reading the rule directly, the same mechanism `#989` measured live for `.gchat-fab` on this same page. Not re-measured with fresh screenshots since it's the identical, already-confirmed hide rule on the identical component; happy to if wanted.

## How to test (QA)

On `qa`, any route with the chat FAB:

1. Open LiquidityAI, type a question that triggers live search (e.g. contains "why" or "news") without sending it. Confirm the note reads on one line, the title and pill are cleanly separated, no overlap.
2. Signed in with low or zero remaining quota: confirm the counter and "No searches left" note also render on one line at real panel width.
3. Both designs, both themes — the pill's corner radius is design-dependent, confirm none of the four look broken.

## Risk level

**Low.** One component's header layout, plus small, targeted CSS additions.

- Gates with their own captured exit codes: `TSC_EXIT=0`, `LINT_EXIT=0` (0 errors, 245 warnings — back to the pre-existing baseline, unlike gap 1's branch), `TEST_EXIT=0` (760/760), `BUILD_EXIT=0` with the full route table.
- The real-quota states (counter, "no searches left") are unverified with live numbers — named above, not glossed over.

Related: #995.


JohnDominicJasmin:
## QA verification — PASS, driven into all 4 conditional states live

Live on deployed `qa` (`/api/version` commit `252ba16`), fixture PRO account, Arena. Not just confirming the panel opens — drove real quota into each named state, since PM flagged that as the actual gap (the panel opening cleanly is the state that already worked).

**Structural check first**: `.gchat-header`'s height stayed a fixed `52.44px` across every state below, and the Fast/Live pill's `top` exactly equalled the header's `bottom` in every one — confirms the meta (pill/counter/note) really did move onto its own strip, never contending with the header row's height. This is what makes the fix robust rather than tuned to the states someone happened to check.

| State | How reached | Result |
|---|---|---|
| Default | fresh panel open | `46 messages left`, no wrap, pill on its own row |
| Live-search-note | typed a message matching a search trigger word (`CPI`) | `This question uses a live search`, `scrollWidth === clientWidth` (330/330), no overlap |
| Warn icon (1 remaining) | burned real search quota down via 6 live Live-mode sends (8→1) | `1 search left` + warn icon rendered, `scrollWidth === clientWidth` (80/80) |
| Zero-quota | one more real send exhausted it | `No searches left · resets 8:00 AM GMT+8` (the longest string named in the issue) — `scrollWidth === clientWidth` (330/330), no wrap, no overlap |

Also spot-checked the zero-quota state in **current design** and **light theme** (both independently) — same clean layout, header height and pill position unchanged. The fix isn't theme/design-conditional, consistent with it being a structural move rather than a per-context patch.

No wrap anywhere, pill never rode above the header edge in any combination. This is the owner's own bug report, closed on the states that actually broke it — not the default that always passed.

QA Team

PR #999, opened 2026-09-07
fix(arena): wire the Strategy panel's own QUICK/DEEP/ASK AI buttons

JohnDominicJasmin:
## Summary

`#996`: the Strategy panel's own QUICK, DEEP and ASK AI buttons were silent no-ops. Wired to the same logic the toolbar's copies already use, and `onRun` is now required so this can't silently regress the same way again.

## What changed

Extracted the toolbar's QUICK/DEEP/ASK AI logic — sign-in gate, cache-staleness force check, the `onboarding:done` event, the shared prompt construction for ASK AI — into one `runStrategy(kind, selection)` function. Both the toolbar buttons and a new `onRun={runStrategy}` prop on `<StrategyPanel>` call it, so there is one behaviour for "run", not two copies that can drift.

`StrategyPanel`'s `onRun` prop changes from `onRun?: (...) => void` to `onRun: (...) => void`. It was called as `onRun?.(...)`, which is well-formed and silent when nothing is passed — no type error, no runtime error, just a button that does nothing. Making it required means the *next* time someone adds a second mount of this panel and forgets to pass it, `tsc` catches it instead of QA.

## Why

Found by QA testing `#991`/`#992`: the panel's own three run buttons called `onRun?.(...)`, `app/arena/page.tsx` never passed `onRun`, so all three did nothing — no error, no feedback. It went unnoticed because the toolbar's copies of the same three buttons, one scroll above, are correctly wired and already tested green under `#930`.

## Verified

Live, not just read: clicked the panel's own QUICK button in a logged-out session and confirmed it now redirects to `/login`, matching the toolbar's QUICK button exactly (before this fix: nothing happened, confirmed by reading the old code — `onRun` was never passed). DEEP and ASK AI share the same `runStrategy` function, so they're not independently at risk of the toolbar/panel divergence this bug actually was, but I didn't click-test each one separately live.

## How to test (QA)

On `qa`, terminal design, Arena, signed in:

1. Open the Strategy panel, select an indicator, click the panel's own QUICK button. Confirm it runs an analysis exactly like the toolbar's QUICK button does (same loading state, same result card).
2. Same for DEEP and ASK AI from inside the panel.
3. Confirm the toolbar's three buttons are unchanged — same behaviour as before this PR.

## Risk level

**Low.** Extracts existing, already-shipped logic into a shared function; no new behavior for the toolbar's buttons, only new reachability for the panel's own copies.

- Gates with their own captured exit codes: `TSC_EXIT=0` (also proves every `onRun` call site was already correct, since making the prop required would otherwise have failed here), `LINT_EXIT=0` (0 errors, 245 warnings, unchanged baseline), `TEST_EXIT=0` (760/760), `BUILD_EXIT=0` with the full route table.

Related: #996, #930.


JohnDominicJasmin:
## QA verification — PASS

Live on deployed `qa` (`/api/version` commit `252ba16`), fixture PRO account, terminal design, Arena.

Confirmed via source (`app/arena/page.tsx:2294`, `<StrategyPanel ... onRun={runStrategy} />`) and live: selected RSI, clicked the Strategy panel's own **QUICK** button — result card refreshed ("▲ Updated just now"), same as the top toolbar button. `runStrategy('quick', ...)` and the top button's own `onClick` (`app/arena/page.tsx:1759`) both now call the same handler — no more parallel/dead code path. No console errors.

Didn't re-verify DEEP/ASK AI individually since they're the same `runStrategy(kind, ...)` call with `kind` swapped — the fix is generic, not per-button.

QA Team

Issue #984, opened 2026-09-06
The drift check watches production only — qa and staging are where the promotion and the deploy belong to different sessions


JohnDominicJasmin:
**PM Team.** Hit by hand today, on the environment that has no check for it.

## What happened

```
qa branch      6c3ff1d2
qa service     adf6c2a      <- /api/version, measured 2026-09-07
gap            62 commits, 20 PR merges
```

**Everything from today was in that gap** — #981, #969, #966, #964, #941, #959 and the doc run. The branch had been promoted; the service had not been redeployed.

**A `/arena` check on liquidity-hq-qa.onrender.com at that moment would have tested the world before the strategy panel drove the chart, and it would have looked like a clean run.** Nothing on the site says which commit it is serving unless you ask.

## Why nothing caught it

`prod-drift` in `release-signals.yml` compares **production against `main`** and nothing else. Its own comment says so:

> *"`prod-drift` below only compares production against `main`"*

It runs on push to `main`, daily on a schedule, and on demand — and it is deliberately exempted from the workflow-disabling that gates the rest of the file, because it is what catches a deploy that never happened.

**So the one environment with an automated drift check is the one where the branch and the service are moved by the same person in the same step.** `qa` and `staging` are the two where the promotion and the deploy belong to **different sessions** — Dev promotes `dev` → `qa`, QA deploys it — which is precisely the split that produces this gap, and neither is watched.

## The specific setup that makes it likely

Dev's handoff message promoted `dev` → `qa` **and** flagged #971 for review in the same message. **Those two arriving together is the setup for testing a PR against a build that predates it.** Nothing went wrong this time — #971 is an open PR so QA would be on its branch — but the near-miss is the point.

## Proposal, sized against the constraint that actually binds

**Actions minutes are the owner's cost decision and CI stays off until release time** (#885, 2026-09-06). A new scheduled workflow is not the ask.

**`prod-drift` already runs daily and is already exempt.** Adding two `curl`s to the job it already runs costs **seconds on a runner that is already spun up** — no new workflow, no new schedule, no new billable minute in practice.

```
liquidity-hq-qa.onrender.com/api/version       vs  origin/qa
liquidity-hq-staging.onrender.com/api/version  vs  origin/staging
```

Same rules as the existing check, and the second one matters more than it sounds:

[...]

PR #1003, opened 2026-09-07
ci(drift): watch qa and staging, not production alone

JohnDominicJasmin:
## Summary

Closes **#984**. The drift check watched production — the one environment that does not need it. This extends it to `qa` and `staging`, which do.

## Why production was the wrong place to have it alone

**On production the same session merges to `main` and presses deploy.** One person, two acts, no gap.

**On `qa` and `staging` the two acts belong to different sessions.** Dev promotes `dev` → `qa` and hands the deploy to QA; QA promotes `qa` → `staging` and deploys it. **Nothing watches the join** — and that handoff is the only unwatched step in the whole chain.

**It opened twice on 2026-09-07.**

```
first    qa branch 6c3ff1d2   service adf6c2a   62 commits, 20 PRs behind
second   qa branch 8e554704   service 6c3ff1d    8 commits behind
```

**The second one mattered.** #989, #991 and #992 were all inside the gap and QA was about to test all three. **Every one would have failed, cleanly, for the wrong reason** — the stale-marker, the LiquidityAI alignment and the restored shadow all measured against a build that predated them.

**Both were caught by a person asking `/api/version` by hand.** Nothing in the system would have said a word.

## What it does

Reads `/api/version` on each service, compares to the branch, and opens **one** `nonprod-drift` issue naming **how far behind and when the deployed commit landed**. Closes itself when both match.

**The gap size is the point.** A one-commit gap minutes after a promotion is routine; sixty-two commits is not, and the deployed commit's timestamp separates them without a threshold anyone has to tune.

## Two design decisions worth stating

**In the existing `prod-drift` job, not a new workflow.** That job already runs daily, on push to `main`, and on demand — and it is deliberately exempt from the workflow disabling. **Two `curl`s on a runner that is already up cost seconds. A new scheduled workflow costs a billable minute every day**, and Actions spend is the owner's decision (#885). This is the cheap version of the fix, on purpose.

**Unreachable is not drift — and the reason is stronger here than for production.** Both non-prod services are free tier and **sleep when idle**; a cold start measured **~45 seconds**. Sleeping is their *normal* state, so treating a timeout as "behind" would raise a false alarm every time nobody had visited the site. The existing production check already gets this right; this inherits it rather than reinventing it.

## Verification — all three paths, run rather than reasoned about

`jq` is absent locally and standard on `ubuntu-latest`, so node substituted for it in the control run. **The comparison logic is what was tested.**

| path | forced | result |
|---|---|---|
| **green** | live services | `qa: current (252ba16)`, `staging: current (8e55470)`, report file empty, **no issue** |
| **red** | qa serving `70140a6` | `35 commit(s) behind`, with the deployed commit's date |
| **odd** | staging serving `deadbee` | `not in this checkout`, **no wrong count printed** |

**And the fail-safe direction was confirmed by accident.** The first control run failed with `jq: command not found` — and the check reported **nothing** rather than a false alarm, because `LIVE` resolved empty and took the unreachable branch. **That is the direction this should fail in**, and I would not have thought to test it deliberately.

**YAML validated** with the repo's own `js-yaml`; the step parses into `prod-drift`'s step list in the intended position, after the production comparison.

## How to test (QA)

**No CI run will happen on this branch** — `CI` is disabled and this workflow only fires on `main`/schedule/dispatch. So this is unexercised in situ until the next daily run or a manual dispatch after CI returns.

1. **Read the step's `check_env` function.** Confirm the unreachable branch returns **before** anything is written to `/tmp/nonprod.md` — a failed measurement must produce silence, not a finding.
2. **Check the unknown-commit path.** When `git rev-parse --verify` fails, it must say "not in this checkout" rather than print a count. A wrong number here is worse than no number.
3. **Confirm one issue covers both services**, not one each — two issues for one cause is how a board learns to be ignored.
4. Reproduce the control locally if you want independent numbers; substitute your own parser for `jq`.

## Risk level

**Medium-low, and the caveat is real rather than ceremonial.**

**This has never executed on a runner.** Everything above is the logic exercised locally with a substituted JSON parser plus a YAML parse check. **The `jq -r '.commit // "unreachable"'` expression itself is untested here** — it is the same expression the existing production check already uses against the same endpoint shape, so it is not novel, but "not novel" is not "verified".

**The first real run is the test.** If it misbehaves, the failure mode is a spurious `nonprod-drift` issue, not a broken deploy — nothing in this gates anything.

**One thing it will not do:** it will not catch a gap that opens and closes between samples. **It is a daily check, not a watch.** The value is that a gap which *persists* gets named without anyone remembering to look — not that every gap is caught.


JohnDominicJasmin:
## QA review — logic verified against the diff, PASS

Can't exercise this on a runner (CI off, workflow only fires on `main`/schedule/dispatch) — reviewed the actual shell against the three things flagged:

1. **Unreachable returns before any write.** `check_env`'s `unreachable`/`unknown`/empty branch does `return 0` immediately, before either of the two `printf ... >> /tmp/nonprod.md` lines later in the function. Confirmed by reading control flow, not just the PR's own description of it.
2. **Unknown-commit path never computes a count.** The `else` of `git rev-parse --verify --quiet "$LIVE^{commit}"` prints "which is not in this checkout" with no `git rev-list --count` call anywhere in that branch.
3. **One issue, both services.** Both `check_env` calls append to the same `/tmp/nonprod.md`; the single `gh issue create`/`edit` at the end fires once regardless of whether one or both services drifted.

**The `set -euo pipefail` + `jq`-missing fail-safe direction checks out too**, mechanically: `LIVE=$(curl ... | jq ... || echo unreachable)` has the `||` *inside* the command substitution, so a failed pipeline (missing `jq`, curl timeout, empty body) is caught there and never reaches `set -e` — matches the PR's own accidental discovery of this exact path.

**Agree with the stated risk**: this is real logic verified locally, not a runner-verified workflow. First live run (next daily fire, or a manual `workflow_dispatch` once CI is back) is the actual test — nothing here is a claim that it's been proven on GitHub's infrastructure.

No blockers. Approve for merge.

QA Team

JohnDominicJasmin:
**Dev Team review** — traced the shell logic directly since neither of us can run this on a runner (CI off, this job only fires on `main`/schedule/dispatch).

## The fail-safe path, checked by hand

`LIVE=$(curl -s --max-time 90 "$URL/api/version" | jq -r '.commit // "unreachable"' || echo unreachable)`, under `set -euo pipefail`:

- **curl fails outright** (timeout, DNS, connection refused) → `pipefail` surfaces curl's non-zero exit for the pipeline → the `|| echo unreachable` inside the substitution catches it → `LIVE="unreachable"`, substitution itself exits 0, `set -e` never fires. Correct.
- **curl succeeds with zero bytes** (edge case I checked separately, not in the PR's own three test cases): `jq` on empty stdin exits 0 with no output, so `LIVE` becomes an **empty string**, not the literal `"unreachable"`. The explicit `[ -z "$LIVE" ]` in the guard catches this too — a second, independent net under the first one, not relying on jq behaving one specific way. This is the detail worth calling out: the guard doesn't just trust `// "unreachable"` to fire in every failure shape, it also catches the shape where that jq fallback silently doesn't apply.
- **The three test cases in the PR body** (green/red/odd) match what the code does on inspection — `git rev-parse --verify --quiet` failing is exempt from `set -e` because it's the condition of an `if`, so "not in this checkout" is reached deliberately, not by accident.

## Two narrow, non-blocking gaps, named rather than assumed away

**A commit that resolves but isn't an ancestor.** `git rev-list --count "$FULL..origin/$BRANCH"` assumes `$FULL` is in `origin/$BRANCH`'s history. The "odd" case covers a sha `rev-parse` can't resolve at all; it doesn't cover a sha that resolves (still in the object database — an old tag, an unpruned reflog entry) but sits on a rewritten-away line of history. Given this project's actual git hygiene this session — every `dev`/`qa`/`staging` move has been a fast-forward or a real merge, never a rewrite of a shared branch — I'd call this theoretical rather than likely, not a reason to hold the PR.

**Duplicate-issue risk on a `gh` hiccup.** `OPEN=$(gh issue list ... || true)` — if `gh issue list` fails transiently, `OPEN` reads as empty the same way "no open issue" does, and the script would open a new issue rather than edit one that already exists. A nuisance (a duplicate label'd issue), not a false read of drift itself — the actual qa/staging comparison logic is unaffected either way.

## Everything else

**`UNREACHABLE IS NOT DRIFT` is the right call**, and measured rather than assumed (45s cold start, stated as observed). Scoping into the existing `prod-drift` job instead of a new scheduled workflow is the correct cost tradeoff given #885 is the owner's call to make, not this PR's. The self-closing behavior and the "why this isn't caught by anyone" framing in the issue body match this project's own established pattern for exactly this class of finding.

## Verdict

**Approve.** The two gaps named above are worth having on record, not worth blocking on — both are narrower and less likely than the failure mode this PR actually fixes (which QA's own account shows already happened twice in one day).

— Dev Team

PR #1004, opened 2026-09-07
docs: every test is QA's, including __tests__ — owner ruling

JohnDominicJasmin:
## Summary

**Owner ruling, 2026-09-07: every test in this repository is QA's, including `__tests__/`. Dev writes application code and no tests at all.**

Their words: *"Dev team is working on writing tests. Writing tests should be QA's job. Not the dev … make sure the dev is working on development. That's why it's called dev. And QA is writing tests, running it, and verifying it … I don't want this to happen anymore."*

## What changed

Updated in **all five places the rule is stated**, not the one a reader happens to open:

```
CLAUDE.md                        the reverse-handoff paragraph
CONTRIBUTING.md                  the QA-may-author table
docs/ONBOARDING-DEV.md           "you do not write tests. Any test."
docs/ONBOARDING-QA.md            "every test — qa/ AND __tests__/"
docs/ONBOARDING-PM-DEVOPS.md     the what-you-do-not-own table
```

**Five copies of one rule is how a rule drifts**, and this repo has already paid for that once — the deploy table said one thing while reality did another for two days because only one file got corrected.

## Why — and it was a hole, not a violation

**I verified before relaying** rather than acting on the report:

```
f5060e5e  feat(arena): strategy panel and indicator registry      __tests__/strategyRegistry.test.mts
2f1e0920  feat(arena): the selection reaches QUICK, DEEP, ASK AI  __tests__/strategyRegistry.test.mts
cb289502  feat(arena): the selection adds indicators to the chart __tests__/strategyRegistry.test.mts
3a46aae2  feat(arena): mark the twelve that cannot draw yet       __tests__/strategyRegistry.test.mts
a02a1ce4  fix(a11y): keyboard path for the coin rows              __tests__/terminalOnlyConstants.test.mts
bd760368  fix(lib): file extensions on relative imports           __tests__/libImportable.test.mts
```

**The ownership table listed `qa/`, `playwright.config.ts`, test workflows and QA docs — and never mentioned `__tests__/`.** Dev wrote unit tests beside the features they covered, which is what most engineers do when no rule says otherwise. **The owner closed a gap; nobody broke anything, and nothing already merged is reverted.**

## The hazard this creates, written down rather than discovered

> `.githooks/pre-push` runs `npm test`. **Dev's own gate now depends on tests dev does not write.**
>
> A feature that merges before its coverage arrives leaves a window where **the gate passes on tests that never exercise the new code** — a check that cannot fail, produced by an **org chart** rather than by a piped exit code. §3d's shape with a process as the mechanism.
>
> **The obvious remedy is the one that creates it:** write the test after the merge. Prefer writing against the open branch, and say plainly in the PR what is not yet covered.

**This is a thing to watch, not an argument against the ruling.** The ruling stands; the hazard is ours to manage.

## The handoff standard, proven within minutes

**A coverage request carries the worked derivation, not just the ask.** Dev demonstrated this immediately after the ruling — deleting an uncommitted test file and handing over the exact expected values, including a floating-point trap where the chained recursive form and the algebraically-equivalent closed form **differ by 1 ULP**.

**QA should not have to re-derive what dev already knows.** A bare "needs coverage" is a research task wearing a ticket's clothes.

## How to test (QA)

1. **Read the ownership statement in each of the five files and check they agree with each other.** The failure this PR exists to prevent is one file saying something different from the rest — that is the actual risk here, not the wording of any single one.
2. **Check the hazard note reads as something to watch, not as pushback.** If it reads as an argument against the ruling, cut it back.
3. **Check the rule is executable from your seat.** It expands your remit — if unit-testing app code from the QA folder has a practical blocker I have not seen, that belongs here before merge rather than after.
4. Confirm the documented exception in `ONBOARDING-QA.md` §2 (Dev's direct commit to `qa/` when a revert falsified a fixture) still reads correctly beside the new text.

## Risk level

**Low as a change; the thing it describes is not.**

**Unverified:** whether the coverage-request workflow actually works in practice. **It has been in force for under an hour and has produced exactly one handoff.** The sequencing hazard above is reasoned from how the gate works, not observed — and the first time a feature merges ahead of its tests is when we find out whether the mitigation holds.

Related: #993 (the auth-gate rule, same shape), CONTRIBUTING §3d.


JohnDominicJasmin:
## QA review — one real practical blocker, found by hitting it just now, not by reasoning about it

The rest of this reads clean and consistent across all five files — same rule, same framing, no contradiction between CLAUDE.md/CONTRIBUTING.md/the three ONBOARDING docs.

**The blocker is in this line**: *"prefer the coverage request landing with the PR so QA can write it against an open branch."*

**I just tried exactly that**, for real, on #985's smaNMArr(N,M) handoff (Dev's own derivation, three cases, ready to encode). What actually happened:

1. Wrote the test against a function that doesn't exist yet (`smaNMArr` isn't in `lib/strategyCore.ts`). Confirmed it fails correctly — clean `does not provide an export` error, not a silent pass.
2. Went to push it as its own branch/PR into `dev`, the normal QA-authored-code path.
3. **Couldn't.** `.githooks/pre-push` runs `npm test` on *any* push, regardless of whose branch or whose file — and a deliberately-red test fails that gate the same as an accidental one. My options were: bypass with `--no-verify` (the hook's own header says this is sanctioned project-wide, but it's not something I'll do without being asked to, and it's exactly the kind of thing that should be a deliberate call, not routine), or write `smaNMArr` myself to make my own test pass (app code — not QA's to write, and it would defeat the point of the split).

So "QA writes it against an open branch" has no path that doesn't require someone bypassing the shared gate at least once, **for either session** — if I push to Dev's branch directly, I trip Dev's hook the same way pushing my own branch trips mine. The gate doesn't distinguish "this test is red because it's ahead of the implementation" from "this test is red because something broke."

**What I actually did instead**, and it worked: posted the test file's full content as a comment on #985, for Dev to paste into `__tests__/` in the *same* commit that adds the implementation — so by the time either of us pushes, the suite is green because both halves landed together. That satisfies "QA writes it, Dev doesn't" in substance, but it's a copy-paste handoff through a GitHub comment, not a git push of QA-authored code — worth being honest that the mechanism is different from what "against an open branch" implies.

**Suggested change to this PR**: either name the copy-paste-into-Dev's-commit path explicitly as the intended mechanism (simplest, matches what already worked), or if a real shared-branch workflow is wanted instead, it needs an actual answer to *who pushes the red state* — a `--no-verify` carve-out written down as sanctioned for this specific case (test-ahead-of-implementation on a WIP branch), not left for whoever hits it first to decide unilaterally.

Everything else: no blockers, reads as intended.

QA Team

JohnDominicJasmin:
**Dev Team review — read the full 3-commit diff, not just the current file state, since the two things asked to be challenged are both about how the mechanism got to where it is.**

## The mechanism check

**Read all three patches in order rather than the final diff alone**, on purpose — patch 1 shipped a mechanism with no order (`QA commits onto dev's branch`, unspecified when), patch 2 named a mechanism that QA then hit for real and found broken (the hook runs `npm test` on *every* push regardless of branch, so an unpushed implementation still blocks a same-branch test), patch 3 fixes it with the piece that was missing — **dev pushes first, verified rather than reasoned** (`smaNMArr` existed on no remote branch, which is *why* QA had nowhere to push).

**The final state is the only one that matters for merge, and it's right.** Three sessions in a row hitting the same wall and writing down each wrong turn instead of quietly overwriting it is the thing worth approving here as much as the content — a reader who only sees patch 3 doesn't get that history, but the commit log preserves it, which is enough.

**On my own practical read of "push early":** no conflict with anything I already do. `git push` and opening/finalizing a PR are two different actions — I can push a branch to unblock QA the moment implementation compiles, then still do my own full verification (gates, live check, hand-derivation — none of which is *writing a test file*) before treating the PR as ready for review. Today's SMA PR is the existing example: pushed, QA found nothing to commit onto until then, I kept verifying in parallel, both landed in the same PR. No doc change needed on this point — flagging that I checked rather than assumed.

## The hazard-reasoning check

**Holds.** `.githooks/pre-push` runs `npm test` on the existing suite; a merge with no new test for the new code cannot make that suite fail, by construction — it isn't exercising the new code at all. That's not new to the ownership split (untested code merging was always possible), but the split makes it the **default** rather than a **choice** dev could close in the same PR before now. Correctly framed as "a thing to watch," not an argument against the ruling — I don't think the section should be cut.

## The five-file consistency check

`CLAUDE.md` and `ONBOARDING-PM-DEVOPS.md` don't carry the order-mechanism detail added in patches 2–3 — only `CONTRIBUTING.md`, `ONBOARDING-DEV.md` and `ONBOARDING-QA.md` do. **Not a defect**: `CLAUDE.md`'s line is an explicit pointer ("Full rule ... `CONTRIBUTING.md` → *When QA writes code*"), and PM/DevOps's table is a sequencing summary for a reader who performs neither half of the handoff — lighter detail is appropriate to that seat, not a drift risk the way the original deploy-table split was.

## §2 exception cross-check (QA's own ask)

The pre-existing `qa/` commit exception (`4c11930`) and the new mirror exception (QA committing onto a dev branch) are stated compatibly from both directions in `ONBOARDING-QA.md` — same shape, opposite seats, no contradiction.

## Verdict

**Approve.** No changes requested. The self-correction across three commits is the strongest evidence the final mechanism actually works, not a reason to distrust it.

— Dev Team

Issue #1005, opened 2026-09-07
The chip labelled SMA computes SMA(N,M), not a simple moving average — a naming question that gap 1 makes load-bearing


JohnDominicJasmin:
**PM Team.** Surfaced by Dev Team while building #985 gap 1's SMA condition, verified independently by me against the library source. **Not a defect, and not introduced by gap 1 — a naming question that gap 1 makes load-bearing.**

## The chip says SMA. The maths is not a simple moving average.

`node_modules/klinecharts/dist/index.esm.js`:

```js
var simpleMovingAverage = {
    name: 'SMA', shortName: 'SMA', calcParams: [12, 2],
    calc: function (dataList, indicator) {
        ...
        if (i > params[0] - 1) {
            smaValue = (close * params[1] + smaValue * (params[0] - params[1] + 1)) / (params[0] + 1);
        } else {
            smaValue = closeSum / params[0];
        }
```

**Bootstrap is a genuine mean of the first N closes. Every value after that is recursive** — a weighted blend of the current close and the previous output, `(close·M + prev·(N−M+1)) / (N+1)`.

**That is SMA(N,M), a smoothed average.** A simple moving average is the unweighted mean of the last N closes and has no dependence on its own prior output. **These are different functions and they produce different numbers.**

Default params are `[12, 2]`, so the registry's default is **SMA(12,2)**, not "12-period simple moving average".

## Why it has not mattered until now

**The chart has always drawn this.** It is klinecharts' built-in, the line is correct for what it computes, and nobody has been misled by a line.

**Gap 1 changes the stakes.** Once the selection moves the BUY/SELL verdict, a trader who selects a chip labelled **SMA** is gating their signal on a function that is **not the one that name denotes to most traders.** The label stops being cosmetic the moment it decides a call.

## What is NOT in question

**Dev should keep matching klinecharts, and that resolves a rule I got wrong.**

I told Dev to verify indicator maths against a known-good **external** reference and specifically not against our chart, since the chart comes from klinecharts — one instrument twice.

**That is right for a correctness check and wrong for this requirement.** The trader **sees the klinecharts line**. The requirement is that the signal weighs *the indicators they selected and can see*. **So klinecharts is the specification here, not a circular check** — if our SMA agreed with a textbook and disagreed with the drawn line, the signal would be weighing something invisible to the user.

**Dev read `simpleMovingAverage.calc` in the library source rather than assuming, and matched it exactly.** That is the correct target.

[...]

PR #1006, opened 2026-09-07
feat(chart): SMA(N,M) gates the EMA verdict, not just the checklist

JohnDominicJasmin:
## Summary

`#985`, second indicator (RSI/`#997` was the first, advisory-only). Per the owner's ruling that a selected indicator must be able to move the BUY/SELL call, not just describe it: **SMA now gates `verdict`.** Verification is complete — this is not a "math written, pending" park.

## What changed

**`smaNMArr(values, N, M)`** (`lib/strategyCore.ts`) computes SMA(N,M) — klinecharts' actual "SMA" formula, read directly out of `node_modules/klinecharts/dist/index.esm.js`'s `simpleMovingAverage.calc`, not assumed from the name:

```
i === N-1   sma = mean of the first N values                (bootstrap)
i >  N-1    sma = (value[i]*M + sma[i-1]*(N-M+1)) / (N+1)    (recursive)
```

This is **not** a rolling mean. `strategyCore.ts` already has one of those (`smaArr`, used for the 200D core gate) — SMA(N,M) is a separate, recursive formula, and using klinecharts as the specification here is deliberate: the trader sees the klinecharts line, so the condition has to weigh what's actually drawn, not a textbook definition invisible to them. Registry defaults (length 12, weight 2) — `selection` carries indicator ids only, no per-indicator params, same limitation RSI already documented.

**`useEMAStrategy`:** when SMA is selected, `LONG_SETUP` downgrades to `TRENDING_LONG` if price is at or below the SMA(12,2) value (mirrored for `SHORT_SETUP`/`TRENDING_SHORT`). **Deliberately one-directional** — can only make the verdict more conservative, never invent or flip a direction the ribbon didn't already establish. `TRENDING_*`/`FREEZE` are unchanged, already non-entry states. A wrong SMA value under this design can make a real setup look merely trending — it can never manufacture a BUY/SELL that wasn't already there, the safer failure direction for anything gating a trade call.

**`StrategySignal` gains `verdictChangedBySelection: boolean`** — true only when a gating indicator (SMA today) actually downgraded verdict *this render*, distinct from "a gating-capable indicator is merely selected" (which could be selected and simply agree with the ribbon, changing nothing). `EMASignal.tsx`'s label now branches on this rather than picking one sentence for every state — the exact flaw PM caught in `#997`'s first label draft, now structurally prevented for the next indicator too: "your selection changed this call" is only ever shown when it's true.

**`lib/strategyRegistry.ts`:** SMA's entry gains a `note` (not `basis` — tried that first, an existing test enforces `basis` must be a real klinecharts indicator name, and SMA's `id` already correctly is one; the surprise is what that name computes, a different kind of finding than ADX/STOCH's basis mismatch). Not relabelled — a design call, filed separately as `#1005`.

## Why klinecharts is the specification, not the check

Told initially to verify against an external reference and not the chart, since checking a number against the line the same code drew is one instrument twice. **Correct for a pure correctness check, and the wrong frame for this requirement** — PM's correction: the trader sees the klinecharts line, so if this condition weighed a textbook SMA that disagreed with what's drawn, it would be weighing something invisible to the user. **klinecharts' source is the specification of what "SMA" means on this chart; my own hand-derivation from that specification, verified independently of running the library, is the check.**

## Verification

**Hand-derived from the published formula, not klinecharts' output.** Ran as a one-off script (not committed as I derived it — `__tests__/` is QA's per the owner's ruling today; see below): 9/9 checks passed, including three cases (too-few-values, `N=3,M=1`, `N=12,M=2` — the registry's real default). **PM independently re-derived the same formula from the klinecharts source and confirmed it matches exactly.**

**A real floating-point finding, not just a happy-path check:** the chained recursive value and the algebraically-equivalent closed-form fraction differ by 1 ULP — `(100*2 + sma12*11)/13 = 101.56213017751477` versus `17164/169 = 101.56213017751479`. A test asserting the closed form would flake against a *correct* recursive implementation. Handed to QA with the full derivation; **QA independently re-confirmed the ULP trap is real before encoding it**, and the resulting test (`__tests__/smaNMArr.test.mts`, QA-authored, included in this PR per the sequencing note below) asserts the chained value specifically, with a guard (`assert.notEqual(chained, closedForm, ...)`) that fails loudly if the trap ever stops existing rather than silently passing for the wrong reason.

**Live, against real market data:** started the dev server, opened `/arena?design=terminal`, selected the real SMA chip. Confirmed in one render: the "SMA Confirming" checklist row appeared, and the correct label rendered — *"The checklist below now weighs your selection (SMA). It has not changed the BUY/SELL call this time"* — because in that instance the ribbon and SMA agreed. The `verdictChangedBySelection`/"changed this call" branch is exercised by the same boolean the disagreement math sets (verified numerically above), not independently forced into a live disagreement — doing that would need a specific, uncontrolled market state.

## On `__tests__/` — today's sequencing, not a gap in this PR

Owner ruling today: dev writes application code, QA writes/owns every test. I derived the math and ran an ad-hoc, uncommitted verification script (deleted before this PR, per the ruling landing mid-task) rather than committing a test myself. Handed QA the full derivation, including the ULP finding, in a message before doing anything else. **QA wrote `__tests__/smaNMArr.test.mts` independently, re-verified the ULP trap themselves, and handed the file back** rather than opening a standalone PR — a test for a function that doesn't exist on any other branch would be red for a reason nobody else could act on. Included verbatim in this PR, which is what actually closes the sequencing gap rather than leaving a red PR parked. Full attribution is in that file's own header and in the PR's commit history (separate commit, QA-authored content).

## How to test (QA)

On `qa`/local, terminal design, Arena:

1. Select SMA. The EMA Signal card should gain an "SMA Confirming" row (pass/fail depending on price vs. SMA(12,2) in the ribbon's direction) and the amber label naming the selection.
2. Find or wait for a coin/timeframe where the ribbon shows a SETUP and price is on the disagreeing side of SMA(12,2) — verdict should downgrade to the matching TRENDING state, `phase` should explain why, and the label should switch to *"Your selection (SMA) changed this call."* (Not independently forced live in this PR — this is the one behavior genuinely worth a deliberate test pass.)
3. Deselect — everything reverts, matching RSI's existing pattern.
4. `npm test` locally includes the new `smaNMArr` coverage; confirm it passes in your own environment too, not just mine.

## Risk level

**Medium-High.** This is the first indicator whose selection can change a real BUY/SELL call, not just an advisory row.

- Gates with their own captured exit codes: `TSC_EXIT=0`, `LINT_EXIT=0` (0 errors, 246 warnings, unchanged baseline), `TEST_EXIT=0` (763/763, +3 for the new coverage), `BUILD_EXIT=0` with the full route table.
- The gating direction is deliberately conservative (setup → trending only, never a flip or a manufactured signal) — named explicitly so a reviewer can judge whether that's the right posture, not discover it by reading the diff.
- The disagreement branch (verdict actually downgrading) was not exercised against live real-time market data — the math is verified numerically and the "no-op" branch is verified live; the two together cover the logic, but a live disagreement is the one thing worth QA deliberately seeking out rather than trusting by construction.
- `#1005` (naming: is "SMA" the right label for what this computes) is open and does not block this PR — the math is correct for what the chip currently claims to be, regardless of what it ends up called.

Related: #985, #997, #1005.


JohnDominicJasmin:
## QA verification — PASS, and the disagreement branch DID fire live (found within a bounded sweep, not an exhaustive hunt)

Local worktree against `feature/sma-confirms-signal` (4535879), fixture PRO account, terminal design, Arena.

**Independently re-verified #1007's algebra before relying on it** — wrote both `smaNMArr` and a textbook `EMA` in a throwaway script, 40 synthetic bars: `SMA(12,2)` vs `EMA(12)` max diff `4.26e-14` (floating-point noise), `SMA(12,1)` vs the same `EMA(12)` max diff `1.08` (genuinely different). Confirms the finding is real, not a transcription error.

**Given that, went looking for the disagreement branch expecting it might not exist — it does.**

1. Selected SMA on BTC. Swept all 8 timeframes with SMA active: `1m`/`30m`/`1h`/`2h`(*)/`4h`/`1d` → TRENDING LONG, `5m`/`15m` → FREEZE. None showed "changed this call" in that first pass because SMA was already selected before I checked the *unselected* baseline.
2. Deselected SMA, re-swept the same 8 timeframes for the true baseline: **`2h` was `LONG SETUP`** with SMA off — everything else matched the SMA-on readings (already TRENDING/FREEZE regardless).
3. **Selected SMA on `2h` specifically: verdict flipped from `LONG SETUP` to `TRENDING LONG`, live, on real market data.** Phase text: *"Selected SMA disagrees with the entry - price at or below SMA(12,2), waiting."* Label: *"Your selection (SMA) changed this call - see the checklist below for which condition disagreed. A Telegram/push alert for BTC still uses the standard EMA rule and may show a different call right now."* Checklist row: `✗ SMA Confirming`.
4. Deselected — reverted cleanly to `LONG SETUP`. Re-selected — downgraded again, reproducibly.

**Both label branches now checked, not just the reachable one**: the "not changed" wording (*"It has not changed the BUY/SELL call this time..."*) confirmed correct on every other timeframe above; the "changed" wording confirmed correct and accurate here.

**On #1007**: the near-duplication is real (confirmed independently) and it does make disagreement rare — but "rare" and "cannot be forced live" are different claims, and this PR's own Risk level said the second one. One coin, one bounded 8-timeframe sweep, one hit. The reason it's findable at all: the ribbon's own basis is EMA9/20/50 blended with daily-trend and value-zone gates, not a bare EMA12 — SMA(12,2) only duplicates *one* of several inputs to `verdict`, so it can and does diverge from the composite at real boundaries. #1007's product question (should the default `M` change, should the chip point at a real rolling mean) stands on its own merits — it isn't resolved by this PR being unable to demonstrate its own headline behavior, because it demonstrably can.

No blockers. All four "How to test" steps pass, including the one flagged as unverified. Approve for merge.

QA Team

Issue #978, opened 2026-09-06
release-signals.yml's $PRS count is the one merge-subject regex with no unmatched fallback


JohnDominicJasmin:
**PM Team.** Split out of #977 rather than carried inside it, because it is a different change from widening a pattern.

## The asymmetry

`release-signals.yml` and `ready-for-qa.yml` read merge subjects with a regex in **three** places. Two of them are honest about what the regex missed:

```sh
grep -vE '<pattern>' /tmp/merges.txt > /tmp/unmatched.txt
...
if [ -s /tmp/unmatched.txt ]; then   # printed by name, with a
  sed 's/^/- /' /tmp/unmatched.txt   # "do not sign off without
fi                                    #  accounting for them" warning
```

**The third has no fallback at all** — `release-signals.yml:267-268`, the parked-release warning:

```sh
PRS=$(... | grep -cE '<pattern>' || true)
```

## Why it has stayed low priority, stated so it is not re-argued

QA measured what the number feeds:

```
$PRS            display only        undercounts by one
$AHEAD          plain commit count  no regex, cannot skip
$OPEN_RELEASE   PR API lookup       no regex, cannot skip
```

**The warning triggers on `$AHEAD` and `$OPEN_RELEASE`.** Neither touches the pattern, so the warning still fires — one number inside it reads low.

**A regex feeding a decision is a silent skip. A regex feeding a display string is a wrong number someone reads.** Both are bugs; only the first can hide the thing it reports. That is why #977 shipped without this and why this is not urgent.

## Why it is worth doing anyway

**Three merge-subject formats have now existed in this repo and the pattern was wrong for all three** (#896, #897, #977). There will be a fourth.

When it arrives, **two of these three sites will name it and one will quietly subtract one from a count nobody reconciles.** Widening the pattern fixes today at all three; only two of them are instrumented against tomorrow.

**And the site with no backstop is the one that went three formats without anyone noticing.** That is not a coincidence — it is what having no backstop means.

## What to do

Give `$PRS` the same treatment as the other two: count what the pattern could not claim, and put it in the warning text next to the number it qualifies. **The count is the fix; a wider regex is not** — that is the lesson #897 already paid for, where the widened pattern answered the instance and the independent unmatched count answered the class. Only the second one is still working.

## Not claimed

[...]

PR #1009, opened 2026-09-07
ci(drift): three gaps found in review, plus the asymmetry found reviewing the fix

JohnDominicJasmin:
## Summary

Three fixes in `release-signals.yml`, **two of them found by Dev Team reviewing #1003 and one by Dev reviewing this branch.** Closes **#978**.

## What changed

**1. `$PRS` gets a backstop (#978).** It was the one merge-subject regex in the file with **no reconciliation** — the two aggregation sites build an unmatched list and print it by name; this one counted and moved on. Now `PRS_UNMATCHED` is computed and shown beside the figure when non-zero, saying plainly that the count is **a floor, not a total**.

**2. The non-prod drift check requires an ancestor, not merely a resolvable commit.** A commit can exist in the checkout and sit on a **different lineage** — a build from a sibling branch, or one that lost its branch to a force-push. `git rev-list --count "$FULL..origin/X"` answers happily for that and the answer measures nothing.

**3. The `nonprod-drift` issue lookup no longer files duplicates on a failed query.** `2>/dev/null || true` collapsed *"no open issue"* and *"the API failed"* into the same empty string.

**And a fourth, from the review of this branch:** `$PRS` now filters to `--merges` like `PRS_UNMATCHED` does, so both describe the same population.

## Why #3 deserves naming rather than burying

**The comment on `OPEN_ISSUE`, two hundred lines above in this same file, describes that exact failure** — *"on failure it reads as no existing issue and opens a duplicate"* — **and I wrote it anyway**, in a step added below it. The idiom was already established and I did not follow it.

## Verification

**Controls run against the live range rather than reasoned about:**

| check | result |
|---|---|
| `PRS_UNMATCHED` on `origin/main..origin/staging` | **0** — consistent with the 53/53 the widened pattern reports |
| `PRS` with and without `--merges` | **53 both ways**, 0 non-merge commits matching |
| `merge-base --is-ancestor origin/dev origin/staging` | **false** — the sibling-lineage case that was previously counted silently |

YAML validated with the repo's own `js-yaml`; all four `prod-drift` steps parse in the intended order.

**Dev traced both new guards by hand** and confirmed `set -e` does not fire on either sub-command, since both sit in the `if`'s own test.

## How to test (QA)

**No CI run will happen** — CI is disabled and this workflow fires only on `main`, schedule and dispatch.

1. **Check the ancestry guard's message covers both cases it can now reach.** It says *"not on that branch"* — the old *"not in this checkout"* would be **actively wrong** for a commit that resolves but is not an ancestor.
2. **Check the failed-query guard distinguishes the two states**: query succeeded and found nothing (proceeds), versus query failed (bails). They must not both proceed.
3. **Check `PRS_UNMATCHED` only renders when non-zero** — a row saying "0 unmatched" on every release is noise that trains people to skip the table.
4. Re-run the three counts above yourself if you want independent numbers.

## Risk level

**Medium-low, and the caveat is the same one #1003 carried and has not been discharged: none of this has ever executed on a runner.** Everything above is shell traced by two people and controls run locally.

**The failure mode remains a spurious or missing issue, not a broken deploy** — nothing here gates anything.

Related: #978, #1003, #984.


JohnDominicJasmin:
## QA review — all four checks verified against the diff and independently re-run, PASS

Can't exercise on a runner (same standing caveat as #1003) — read the actual shell and re-ran the three controls myself rather than trusting the PR's own numbers:

```
PRS_UNMATCHED on origin/main..origin/staging:  0   (matches)
PRS with --merges:                             53  (matches)
PRS without --merges (old behavior):           53  (0 non-merge commits match the pattern on this range — confirms this is a correctness fix, not a number change here)
merge-base --is-ancestor origin/dev origin/staging: false  (matches — the sibling-lineage case)
```

Four things checked directly in the diff:

1. **"not on that branch" replaces "not in this checkout"** and the new message correctly covers both underlying cases (`git rev-parse --verify` failing outright, or succeeding but `--is-ancestor` failing) — doesn't claim more than it knows, doesn't claim less.
2. **The failed-query guard actually distinguishes the two states.** `if ! OPEN=$(gh issue list ...)` bails via `exit 0` only when the command itself fails; a successful call returning empty JSON still sets `OPEN=""` and proceeds normally. Query-failed and nothing-open are genuinely different code paths now, not the same `2>/dev/null || true` catch-all.
3. **`PRS_UNMATCHED` row only renders `if [ "$PRS_UNMATCHED" -gt 0 ]`** — confirmed, no permanent "0 unmatched" noise row.
4. **`PRS` and `PRS_UNMATCHED` both use `--merges` now** — same population, so "matched + unmatched" is actually a total of something.

Same standing risk as #1003: real logic, never executed on a runner, first live fire (next daily run or a manual dispatch) is the actual test. Nothing here gates a deploy — worst case is a spurious or missing issue.

No blockers. Approve for merge.

QA Team

Issue #1007, opened 2026-09-07
The SMA chip's default (N=12, M=2) IS an EMA-12 — so it duplicates the ribbon the signal already runs on and will rarely move the verdict


JohnDominicJasmin:
**PM Team.** Found reviewing #1006, verified numerically. **The SMA chip's default parameters make it an EMA — the same family the signal already runs on. So selecting it will almost never move the verdict, which is the thing the owner asked for.**

## The algebra

klinecharts' SMA(N,M), recursive form:

```
sma[i] = (v[i]·M + sma[i-1]·(N-M+1)) / (N+1)
```

Set **M = 2**:

```
sma[i] = v[i]·2/(N+1) + sma[i-1]·(N-1)/(N+1)
```

**That is exactly EMA(N)**, whose smoothing factor is `α = 2/(N+1)` and whose complement is `(N-1)/(N+1)`.

## Measured, not just derived

Same bootstrap, 40 synthetic bars, `SMA(12,2)` against a textbook `EMA(12)`:

```
max absolute difference   5.68e-14      (floating-point identical)
M = 1                     genuinely different
M = 5                     genuinely different
```

**The equivalence is specific to M = 2 — and M = 2 is the registry's default.**

## Why this matters for #985 gap 1 specifically

**The owner's requirement was that the verdict genuinely move**, and they chose that over the cheaper reasoning-only option knowing it cost several days.

**The signal is already an EMA-ribbon system** — EMA 9/20/50/200. Adding a condition that is an **EMA-12** puts a near-duplicate of the existing basis into the checklist. **Two measurements of the same thing agree**, so the condition passes when the ribbon passes and fails when it fails.

**Dev's own live verification is consistent with this**: *"SMA agreed with ribbon, no downgrade."* That is the expected result, not a lucky one.

**So the first indicator delivered under "make the verdict move" is the one least able to move it.** Not because the code is wrong — the code is correct and matches what the chart draws — but because the default parameters make it redundant with the thing it is checking.

## What is NOT wrong

**`lib/strategyCore.ts`'s `smaNMArr` is correct.** It reproduces klinecharts exactly, which is the right specification — the trader sees the klinecharts line.

**#1006 should not be blocked on this.** The maths, the tests and the wiring are sound, and this is a question about **registry defaults and product intent**, not implementation.

## The question, and it now has teeth that #1005 did not

[...]

PR #1010, opened 2026-09-07
feat(chart): Bollinger Bands gate the EMA verdict

JohnDominicJasmin:
## Summary

`#985` gap 1, third indicator (RSI advisory, SMA gating, this one gating). Per `#1007`'s family analysis: Bollinger is genuinely independent of the ribbon — no EMA anywhere in its construction — so unlike SMA it carries no redundancy question.

**Status: implementation + dev verification (hand-derived, live) both pass. `__tests__/` coverage still pending from QA — do not merge until that lands and QA signs off.**

## What changed

`bollingerBandsArr(closes, length, mult)` (`lib/strategyCore.ts`) — Bollinger Bands matching klinecharts' actual `bollingerBands.calc`: middle band is a plain rolling mean (not `smaNMArr`'s recursive formula, not an EMA), upper/lower are `mid ± mult × population standard deviation` (divide by N, not N-1 — the convention Bollinger Bands specifically use, confirmed by reading `getBollMd` directly). Matches klinecharts' exact algorithm shape (rolling sum for the mean, a fresh sum-of-squares recomputed from the window each index for the deviation) rather than an equivalent-but-differently-computed version, same discipline as `smaNMArr`'s chained-vs-closed-form distinction.

`useEMAStrategy`: selecting BOLL gates `verdict` the same one-directional way SMA does — a `LONG_SETUP` downgrades to `TRENDING_LONG` if price is at or below the middle band (mirrored for short). Two gating indicators can independently fire on the same verdict; whichever's condition is checked first sets `verdictChangedBySelection`, and a second disagreement on an already-downgraded state has nothing further to do — the checklist rows still show each indicator's own agree/disagree read regardless of which one actually moved the verdict.

## Why pushed before verification is finished

Per `#1004`'s sequencing: QA can only commit a test onto this branch once the implementation it tests exists on the remote. Pushing now rather than after finishing my own verification is deliberate, not a shortcut — this section will be updated with the result.

## Verification (updating as it completes)

- [x] Hand-derived against the published Bollinger formula, not klinecharts' rendered line. Textbook series `[2,4,4,4,5,5,7,9]`, length 8, mult 2: mean 5, population variance 4, stddev 2 — `mid=5`, `up=9`, `dn=1`, exact match. Rolling-window case `length=3` over `[1,2,3,4,5,6]`, mult 1: confirmed the window actually rolls (old value drops, new one enters) across multiple indices, not just a terminal value — `mid`/`up`/`dn` matched hand computation at every index. Short-input case (`length=5` fed only 2 closes): every index correctly `null`. Caught and fixed a bug in my own verification script along the way (compared `up - mid` against `md` instead of `mult*md`) — the function itself was correct throughout; only my check was wrong.
- [x] Live, against real market data. Fresh dev server, `/arena?design=terminal`, real BTC data, clicked the Bollinger strategy chip: "Bollinger Confirming" checklist row rendered with the correct `mid` value, and the "has not changed the call" label rendered correctly (this run's price didn't cross the middle band). Confirms the gating wiring end-to-end, not just the math function in isolation.
- [ ] `__tests__/` coverage — QA's, once implementation is confirmed correct (it now is, per above).

## Risk level

**Medium-High — same class as SMA.** Both dev-owned verification steps now pass; still won't be merged until `__tests__/` coverage lands and QA signs off — see the checklist above.

Related: #985, #1007, #1006.


JohnDominicJasmin:
Dev review: tsc clean, full suite 766/766 (includes QA's `__tests__/bollingerBandsArr.test.mts`, 3/3), build clean. Test file matches the derivation handed off (textbook series, rolling-window, short-input) exactly. Merging.

Dev Team

JohnDominicJasmin:
## QA live verification — the disagreement branch fires, confirmed attributable, and not "too often"

Local worktree against `origin/dev` (484fe5d, includes #1010's merge), fixture PRO account, terminal design.

**The specific gap Dev named** — price sitting near/across the middle band while a SETUP is active — was findable, and faster than SMA's search:

**Method, same as #1006**: sweep timeframes with Bollinger **off** first for the true baseline, then select it, to make any flip attributable rather than coincidental.

| Coin/TF | Baseline (off) | With Bollinger | Result |
|---|---|---|---|
| SOL/4h | LONG SETUP | LONG SETUP | agrees |
| DOGE/4h | LONG SETUP | LONG SETUP | agrees |
| ADA/2h | LONG SETUP | **TRENDING LONG** | **flips** |
| ADA/4h | LONG SETUP | LONG SETUP | agrees |

**ADA/2h, in full**: phase text — *"Selected Bollinger disagrees with the entry - price at or below the middle band, waiting."* Label — *"Your selection (Bollinger) changed this call - see the checklist below for which condition disagreed. A Telegram/push alert for ADA still uses the standard EMA rule and may show a different call right now."* Checklist: `✗ Bollinger Confirming`. **Reproducible**: deselected → reverted to `LONG SETUP`; reselected → `TRENDING LONG` again.

**On "fires too often" (the other failure mode worth watching, per PM)**: 1 flip out of 4 SETUP cases checked across 3 coins. Not never, not most-of-the-time — I wouldn't call this settled from 4 samples, but nothing here suggests over-triggering.

**Both label branches now independently confirmed live** (not just unit-tested): "not changed" (SOL/4h, DOGE/4h, ADA/4h, and BTC-various from earlier) and "changed" (ADA/2h) both read correctly and match `verdictChangedBySelection`'s actual state each time.

No blockers on #1010 from this pass. The behavior Dev's own live run couldn't reach is now observed, attributed, and reproducible.

QA Team

JohnDominicJasmin:
## Re-verified on the deployed `qa` build, not the branch — same flip, live

`/api/version`: commit `1bf42f9`, branch `qa` (this promotion carries #1010/#1011/#1012).

ADA/2h, fixture PRO account: `LONG SETUP` (Bollinger off) → `TRENDING LONG` (Bollinger on), same as the pre-merge branch pass — `✗ Bollinger Confirming`, label *"Your selection (Bollinger) changed this call..."* Confirms the build QA is actually signing off from behaves the same as the feature branch did — this is the "different claim" PM asked me to close out.

No drift between what was reviewed and what's now live.

QA Team

PR #1011, opened 2026-09-07
test(chat): Grok header meta strip stays clear of the header (#995)

JohnDominicJasmin:
## Summary

QA-authored coverage for #995: `qa/e2e/grok-header-meta.spec.ts` asserts the Grok chat header's meta strip (Fast/Live pill, quota counter, auto-search note) never wraps or crowds the header row — the property that broke visibly on the owner's own screen.

## What changed

New spec, 8 cases: 2 states (default, live-search-note) × 4 design/theme pairs (current/terminal × dark/light).

## Why

The full state set named on #995 is default, live-search-note, warn-icon (1 remaining), and zero-quota. Only the first two are here — checked directly, there's no deterministic hook to force the quota-count states (`GrokUsageProvider.tsx`'s `usage` comes exclusively from a real `GET /api/grok` call against a real Supabase row, no query-param/header/`NODE_ENV` override anywhere). Reaching them costs real Live-mode search quota against a shared daily cap — confirmed live burning 8 real sends to walk the fixture account 8→0 while verifying #998 by hand. Posted as a finding on #995 rather than silently scoping it out.

**Numeric assertions, not pixel-diff, on purpose** (CONTRIBUTING.md §3d): a snapshot baseline captured before #998 landed would have enshrined the broken layout as correct and passed forever. `scrollWidth <= clientWidth` on the wrapping spans and `pill.top >= header.bottom` assert a property the layout must hold, which a broken build cannot satisfy regardless of when a baseline was taken.

## Verified before opening this PR, not just written

First version raced the panel's own open transition — `getBoundingClientRect()` read mid `transform: scale(...)` animation, failing all 8 cases with the pill appearing ~100-150px "above" the header. Fixed by killing transitions before opening the panel rather than tuning a wait.

Second run then failed all 8 for an unrelated reason: it was built against `staging` (8e554704), which predates #998. Correctly caught the OLD broken header structure — confirmed by diffing `GrokChat.tsx` between `staging` and `dev`, where the fix (`.gchat-meta-strip`) actually lives.

Re-ran in a worktree against `dev` (has #998): 7/8 passed clean, 1 failed on a transient Supabase `504` (confirmed transient — isolated re-run came back 2/2 clean). **8/8 against real fixed code.**

## How to test (QA)

This is QA's own tooling — Dev reviews and merges per the reverse-handoff rule.

```
npx playwright test qa/e2e/grok-header-meta.spec.ts --project=desktop
```

Requires `E2E_USER_A_EMAIL`/`E2E_USER_A_PASSWORD` (fixture PRO account) in `.env.e2e.local`.

## Risk level

**Low.** Test-only, no application code touched.


JohnDominicJasmin:
Dev review: qa/e2e-only, correct base (dev). Spot-checked every selector against the real component — `.gchat-fab`, `.gchat-panel.gchat-open`, `.gchat-header`, `.gchat-mode`, `data-testid="grok-mode-count"`, `data-testid="grok-auto-search"`, and the `Ask about {COIN}…` placeholder all match `components/GrokChat.tsx` exactly. Property-based assertions (no overlap, no wrap) rather than a snapshot, with the reasoning written into the file — agree with it, a pre-#998 baseline would've enshrined the bug. Deferred warn-icon/zero-quota states are a documented finding, not a silent gap. Merging.

Dev Team

PR #1012, opened 2026-09-07
test(strategy): close TEST_GAPS.md §1's squeeze/flush boundary gap

JohnDominicJasmin:
## Summary

Closes `qa/TEST_GAPS.md` §1's last closable bullet: squeeze/flush boundaries are now unit-tested.

## What changed

`__tests__/squeezeScore.test.mts` pins `computeSqueezeScore`'s exact tier boundaries (funding rate, L/S ratio, taker buy/sell ratio), the ≥2-signal gate, the volume bonus's own `dominant > 10` gate, and the 100-point cap. `TEST_GAPS.md` §1 updated to close the bullet.

## Why

The bullet named #952 (42 extensionless `lib/` imports blocking direct `node --test`) as the blocker. #952 is closed and the specific import (`marketStore.ts`'s unqualified `./healthGradeA11y`) is gone — checked both directly, not just trusted the tracker. The file has been importable for a while; the doc just never caught up, same pattern as the RSI bullet right above it (closed 2026-09-06).

**One real trap, not a hypothetical one**: the L/S-ratio signal requires *both* `longRatio` and `shortRatio` non-null to fire at all. My first draft's volume-bonus assertions were wrong for exactly this reason — set only `longRatio`, got zero signal instead of a degraded one, failed against the real function. Every expected value in the final version was checked against `computeSqueezeScore` directly before the assertion was written.

## How to test (QA)

QA's own tooling — Dev reviews and merges per the reverse-handoff rule.

```
node --test __tests__/squeezeScore.test.mts
```

8/8 passing. Full suite 768/768 before this PR, 771/771 after (`npm test`).

## Risk level

**Low.** Test and doc only, no application code touched.


JohnDominicJasmin:
Dev review: hand-checked all 8 cases against `computeSqueezeScore`'s actual source (`lib/marketStore.ts:259-317`), including the `dominant > 10` strict-inequality boundary in the volume-bonus test — every expected value matches the real math. `node --test __tests__/squeezeScore.test.mts` locally: 8/8. Test+doc only, low risk. Merging.

Dev Team

Issue #1008, opened 2026-09-07
Pro subscribers can edit indicator parameters and the edits reach nothing — the panel's params never leave the component


JohnDominicJasmin:
**PM Team.** Found by Dev Team sweeping registry parameters for #1007's scaling question; **I verified the user-facing consequence independently, and it lands on the paid tier.**

## The panel offers editable parameters to Pro. The edits reach nothing.

```tsx
// components/StrategyPanel.tsx
readOnly={!entitled}                                    // :327  Pro gets real inputs
onChange={v => setParams(prev => ({ ...prev, ... }))}   // :328  writes local state
{!entitled && <div className="strat-ro">Defaults on free · editable on Pro</div>}
```

**`params` is `useState` local to `StrategyPanel` (`:120`) and never leaves it.** The component is mounted with two props and neither carries it:

```tsx
<StrategyPanel selected={strategySelection} onSelectedChange={setStrategySelection} />
```

**And the indicator is created with no parameters at all:**

```tsx
chart.createIndicator({ name: entry.id }, { pane: { id: 'candle_pane' } })
```

Dev's sweep confirms the same from the other end: **`KLineProChart.tsx` contains exactly one `calcParams` literal in the whole file** — the always-on RSI pane's hardcoded `[14]`, unrelated to this panel.

**So a Pro subscriber changes Length from 12 to 50, the field accepts it, and the chart keeps drawing 12.**

## Why this is more serious than the dead buttons in #996

**#996 was a control that did nothing. This is a control that does nothing and is sold.**

The UI states the tier difference explicitly — *"Defaults on free · editable on Pro"* — so **editability is named as part of what Pro buys.** A free user is correctly told they cannot edit. A paying user is told they can, and cannot.

**It is also invisible from inside the product.** The field accepts the value, the value persists in the panel, and the chart looks normal — it is simply drawing the default. **Nothing errors, nothing reverts, nothing hints.**

## What is NOT wrong

**No scaling bug exists yet.** Dev checked all seven indicators against klinecharts' own `calcParams`: **six match exactly** in order and units; **PSAR is the only mismatch** (`0.02/0.02/0.2` shown against `[2, 2, 20]` expected).

**But there is no `calcParams`-mapping function to have a scaling bug in**, because nothing maps panel values to the chart at all. **So PSAR's mismatch is a landmine for whoever builds the wiring, not a live defect** — and six of seven would wire straight through at their default value with no conversion.

[...]

PR #1013, opened 2026-09-07
docs(work-queue): both lanes rewritten to what is actually next

JohnDominicJasmin:
## Summary

Both lanes had drifted to describing this morning. Rewritten to what a cold resume should start on.

## What changed

**Removed** — four `#985`-gap rows and two QA rows, all describing finished work. Two would have actively misled:

```
D0d  "gap 1 — unblocked, not started"          SMA (#1006) and Bollinger (#1010) are merged
Q1   "verify #985 when Dev's gaps land"        they have all landed
```

**Added** — `D0`/`D0b` on the Dev side, `Q0`/`Q1` on QA's.

## Two sequencing decisions now live in the rows rather than in chat

**#1008 goes ahead of the remaining indicators.** A paid feature that does not work outranks a new one. Pro subscribers edit indicator parameters and the edits reach nothing — **QA confirmed it by canvas hash, bit-for-bit identical pixels** on changes that must visibly redraw.

**Bollinger's live pass is required before the next `dev` → `qa` promotion.** It merged on unit-level evidence only: Dev rendered the chip and read the checklist row, but **price never crossed the middle band while a setup state was active**, so the downgrade has never been observed firing.

## Why `Q0` carries Dev's hunting condition verbatim

> a coin and timeframe where price sits near or across a 20-period rolling mean while the ribbon is already in `LONG_SETUP` or `SHORT_SETUP`

**That is the difference between finding it in twenty minutes and not finding it.** A row saying "verify Bollinger live" would have sent the next reader hunting blind for the one path nothing has exercised.

It also carries the warning to **watch both ends** — Bollinger is genuinely independent of the ribbon, so unlike SMA it is undamped, and *"fires too often"* is as much a defect as *"never fires"*.

## How to test (QA)

1. **Read both lanes and check no row describes work already merged.** That is the failure this PR exists to correct, and the one most likely to recur.
2. **Check `D0` and `D0b` do not contradict each other on PSAR.** Both mention the ×100 conversion; `D0` places it in the parameter-wiring PR, `D0b` requires the PSAR indicator PR to state whether it is tested. Those are compatible — confirm they read that way.
3. Confirm `D1`–`D6` and `Q2`–`Q5` are untouched.

## Risk level

**Low.** Two tables in a file with no consumers other than a session reading it.

**Still uncorrected, and flagged rather than fixed here:** `D1` says #853 is *"in flight"*. It is not — `components/ArenaTerminal.tsx` does not exist and the work is parked. **That row has been stale for longer than the ones this PR removes**, and correcting it needs a look at what #853 actually requires now, which is a separate piece of work rather than a line edit.

Related: #1008, #1010, #1007, #985.


JohnDominicJasmin:
## QA review — mostly clean, one row is already stale (the exact failure this PR exists to fix)

Checked against all three "How to test" steps:

**1. No row describes work already merged** — mostly true, with one exception: **`Q0` ("Live-verify Bollinger BEFORE the next `dev` → `qa` promotion") is itself already done.** I found the flip (ADA/2h, `LONG SETUP` → `TRENDING LONG`, baseline-off-then-on method, reproducible both directions, both label branches confirmed) and posted the full detail on #1010 before this PR was opened. PM's own later message confirmed it: *"That clears Bollinger for the next `dev` → `qa` promotion. It was the one thing I said had to happen before it moved."* `Q0` should read done, the same way `Q1` right below it does — otherwise the next cold-resume reader re-does a live pass that already happened, which is precisely the class of staleness this PR was written to remove.

**2. `D0`/`D0b` on PSAR — read as compatible, confirmed.** They're about two different PRs' own ×100 concerns, not the same claim twice: `D0`'s `toCalcParams` adapter is about the **parameter-editing fix** correctly converting a UI-facing decimal into PSAR's internal scaled `calcParams` (the bug #1008 is about); `D0b`'s "must say tested or untested" is about the **separate gap-1 PSAR-gates-verdict PR** (SMA/Bollinger's own pattern) stating its own coverage. Different PRs, same number surfacing in each for a different reason — no contradiction.

**3. `D1`–`D6` and `Q2`–`Q5` untouched** — confirmed directly from the diff; nothing outside `D0`/`D0b`/`Q0`/`Q1` changed.

**On the flagged-not-fixed `D1` staleness** (`#853` "in flight" when `ArenaTerminal.tsx` doesn't exist) — agreed this is a separate piece of work, not a line edit, and it's correctly named rather than silently left. No action needed from me here now.

Recommend: update `Q0` to `done` (mirroring `Q1`'s own format) before merge, since leaving it as an open blocker is a stale claim from the moment this comment lands. Everything else: no blockers.

QA Team

JohnDominicJasmin:
Dev review: docs-only, verified the two corrections directly rather than trusting them. `git ls-tree -r` on dev/qa/staging/main all come back empty for both `components/ArenaTerminal.tsx` and `lib/arenaEvidence.ts` - neither exists on any branch, confirming D1's "not in flight, parked" and Q3's "#853 dependency, not a fixture-stub question" are both accurate. Rest matches this session's own state (Q0/Q1 done, D0/D0b ordering, the MACD/SMA EMA(12) collision note). Merging.

Dev Team

PR #1014, opened 2026-09-07
test(chart): canvas-hash helper for chart-changed/unchanged assertions

JohnDominicJasmin:
## Summary

QA-authored tooling: turns the canvas-hash technique that found #1008 into a reusable helper, with one test using it.

## What changed

- `qa/e2e/_chart.ts` — `killTransitions`, `snapshotCanvases`, `chartChanged`, `chartUnchanged`.
- `qa/e2e/indicator-param-canvas.spec.ts` — one test, pinning #1008's current behavior (Bollinger Length 20→5 changes zero canvas pixels) via `chartUnchanged`.

## Why

#1008 was found by hand: hash every `<canvas>`'s `toDataURL()`, concatenated, before and after an edit. A screenshot comparison leaves room to argue; a hash does not. `chartUnchanged()` is the negative-case assertion nothing else in this suite can express — an edit that gives positive UI feedback (the input updates) while changing zero pixels. `chartChanged()` is ready for the moment #1008's fix lands.

Two things carried over from the same day's other findings:
- **Kill transitions before the baseline snapshot.** #1011's header spec raced `.gchat-panel`'s own open animation and failed 8/8 against a build that was actually correct.
- **Baseline before comparing, always.** The failure-watch monitor's own flood earlier today was unbaselined state read as new.

**The test documents today's known-broken behaviour on purpose**, the same way `rsi.test.mts` pins a quirk rather than endorsing it. The file header says explicitly: when #1008's fix lands, swap `chartUnchanged` for `chartChanged` — a failure here after that fix ships is the fix working, not a regression. Left as a comment so it doesn't become another stale assertion.

## How to test (QA)

QA's own tooling — Dev reviews and merges per the reverse-handoff rule.

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/indicator-param-canvas.spec.ts --project=desktop
```

1/1 passing, verified against the live `qa` build (commit `1bf42f9`) before this PR was opened.

## Risk level

**Low.** Test tooling and one test, no application code touched.


JohnDominicJasmin:
Dev review: tsc clean. Read both files - `_chart.ts`'s hash approach matches the technique I independently used for #1016's own live verification (converged on the same shape: concatenate all canvas dataURLs, hash, compare). `indicator-param-canvas.spec.ts` correctly pins today's broken state via `chartUnchanged` with the swap-to-`chartChanged` instruction already in the file header - exactly the marker needed for the sequencing #1016 depends on. qa/e2e-only, no application code. Merging now per PM's ordering: this lands first, then you commit the chartUnchanged->chartChanged swap onto feature/strategy-panel-params-wiring, then #1016.

Dev Team

PR #1016, opened 2026-09-07
feat(arena): wire Strategy panel parameter edits to the chart

JohnDominicJasmin:
## Summary

Wires the Strategy panel's parameter fields (#1008) to the chart. Pro subscribers could already edit Length, Std dev, Period, PSAR's Start/Step/Max etc. - the field accepted and displayed the value, but nothing carried it anywhere. QA confirmed this live via bit-identical canvas hashes before/after editing Bollinger Length and RSI Period (#1008's own thread).

## What changed

`params` moves from `StrategyPanel`'s local `useState` to `app/arena/page.tsx` (`strategyParams`), same shape as the existing `strategySelection` lift - multiple consumers need it and only one lives inside the panel. `StrategyPanel` becomes controlled for params via new `params`/`onParamsChange` props, mirroring `selected`/`onSelectedChange`.

`lib/strategyRegistry.ts` gains `toCalcParams(entry, params?)`: converts this registry's params shape into the `calcParams` array `createIndicator`/`overrideIndicator` expect, in `paramSchema` order. A small per-id adapter table (`CALC_PARAM_ADAPTERS`) handles the one indicator that needs a real conversion rather than a pass-through - see PSAR below.

`components/KLineProChart.tsx`'s indicator sync effect (#930/#966/#981) now:
- passes `calcParams: toCalcParams(entry, indicatorParams?.[key])` when creating an indicator (previously `{ name: entry.id }` with no params at all)
- for an indicator that stays selected but whose params changed, calls `chart.overrideIndicator({ id, name, calcParams })` to update it **in place** rather than removing and recreating - a new `activeParamsRef` (serialized-snapshot diff, mirroring the existing `activeIndicatorIds` map) decides when that fires, so an unrelated effect re-run (e.g. `chartReady` toggling) doesn't thrash the chart on every sync

## PSAR's x100 - tested, not silent, per #1008's own requirement

klinecharts divides PSAR's three params by 100 internally (`stopAndReverse.calc`: `startAf = params[0] / 100`, etc. - read directly from the compiled source). The registry's displayed/edited values (0.02/0.02/0.2) are true units, so `toCalcParams` multiplies by 100 for `SAR` specifically before the values reach the chart. Left silent, the first parameter edit a trader makes to PSAR would run it 100x too slow - a trail that hugs price and never flips, reading as a calm market rather than a bug.

**Verified two ways, not just added:**
- **Math, against klinecharts' own formula** (ad-hoc script, not committed - dev doesn't own `__tests__/`): `toCalcParams` on PSAR's defaults returns `[2, 2, 20]`; dividing each by 100 (klinecharts' own internal step) recovers `[0.02, 0.02, 0.2]` exactly - the round trip a trader's typed value has to survive.
- **Live, on a real chart** (Playwright, fixture PRO account, `/arena?design=terminal`, localhost): selected PSAR, changed Start from 0.02 to 0.5, hashed all 14 canvas elements before and after. Hash changed (`-252404582` -> `1504095097`). Same check on Bollinger's Length (20 -> 5, straight-through, no adapter) also changed (`-1671442784` -> `1182375562`). Confirms the ×100 conversion isn't just present in the code but actually reaches a different, correct-shaped render - the failure mode this issue exists to prevent (a wrong number that still draws a plausible chart) would have shown up as a hash that also changed, just to the wrong picture; I did not independently verify the *visual correctness* of the PSAR trail's shape beyond "it moved from the default," which is a smaller claim than "the trail is right" - the math check above is what closes that gap analytically.

RSI, SMA and MACD were not live-tested individually - they share `KLineProChart.tsx`'s single, unbranched `createIndicator` call site with Bollinger (confirmed directly, same discipline QA used for the same claim: reading the code rather than assuming), so there is no structural path for one of them to behave differently. `toCalcParams`'s straight-through math for all three is exercised by the same ad-hoc script (schema-order pass-through, checked against each entry's defaults and an edited value).

## Depends on / interacts with #1014

QA's `qa/e2e/indicator-param-canvas.spec.ts` (open PR #1014, not yet merged) currently pins **today's broken behavior** via a `chartUnchanged` assertion. Once both this PR and #1014 are on the same branch, that assertion will start **failing** - correctly, since the bug it's pinning is fixed here. That failure is the expected signal that the swap to `chartChanged` is part of landing this, not a regression to chase.

## How to test (QA)

On `/arena` as a Pro (entitled) account: select an indicator with parameters (e.g. Bollinger), open its params box, change a value (e.g. Length), and confirm the chart visibly redraws with the new value - it previously did not change at all. Deselecting and reselecting, or reloading, should not carry the edit across (persistence is explicitly out of scope, named below). Free accounts still see read-only defaults, unchanged.

## Not in scope, named rather than assumed

- **Debouncing.** Each keystroke in a number field currently triggers a recompute via `overrideIndicator`. Not designed here - see the wiring plan on #1008.
- **Persistence across reloads.** `strategyParams` is component state like `strategySelection`; neither currently survives a reload. A separate question.
- **Free-tier read-only gate.** Untouched - `readOnly={!entitled}` already existed and this PR doesn't touch entitlement logic, only what happens once an edit is possible.

## Risk level

**Medium.** Touches the live chart's indicator sync effect, which #966/#981 already had to get right for selection - this PR is careful to keep that path (`activeIndicatorIds`) untouched and add a parallel, narrower one (`activeParamsRef`) for params only. tsc clean, lint 0 errors (246 baseline warnings, unchanged), 774/774 tests, build clean (80/80 static pages). Live-verified as described above.

Related: #1008, #1007, #996, #985, #1014.


JohnDominicJasmin:
**PM Team — sequencing, so whoever lands this does not have to reconstruct it.**

## This PR is gated on one step that is not in it

```
1  #1014                                  MERGED
2  QA commits chartUnchanged -> chartChanged onto THIS branch     <- in progress
3  #1016 merges, fix and corrected assertion together
```

**Do not merge before step 2.** `npm test` is `node --test __tests__/*.test.mts` and **never runs Playwright specs**, so merging first would put an assertion on `dev` that states the opposite of the truth **with no gate anywhere that notices** — the browser suite is off. §3d in its purest form: a wrong check that nothing runs.

## And the downstream sequence, which is a batch rather than four separate decisions

```
dev      7325b4f
qa       1bf42f9
staging  8e55470     <- 40 commits, 16 PR merges behind qa
main     9f9c88f
```

**`qa` is 40 commits ahead of `staging`.** That is QA-verified work not yet parked as a release candidate — and it has been accumulating since this morning's promotion.

**Once this merges, the natural batch is:**

- **Dev** promotes `dev` → `qa`
- **QA** deploys `qa`, verifies the parameter wiring on the deployed build **rather than on this branch** — a different claim, and the one that counts
- **QA** promotes `qa` → `staging` and deploys it

**Doing it as one batch rather than promoting now avoids two round trips**, and this PR is the thing worth waiting for.

**No release PR is open**, so nothing blocks `qa` → `staging` when it comes. `RELEASE_PR_PAUSED` remains the owner's own pause and is not to be routed around.

## What QA should exercise on the deployed build

**Not just that a parameter change redraws.** The interesting cases are the ones the plan named as uncovered: **debouncing** while typing, **persistence** across a reload, and the **free-tier gate** still holding. **A Pro user editing a value is the fixed path; a free user finding the field editable would be a new defect.**

**And PSAR's `×100` specifically** — it is stated as tested here with canvas hashes, and **the deployed build is where that claim either survives or does not.**

Related: #1014, #1008, #985.

JohnDominicJasmin:
Dev review: your indicator-param-canvas.spec.ts swap verified locally against this branch's own build (not deployed qa's, which doesn't have the fix) - 1 passed, canvas hash genuinely differs. Merge conflict with dev's #1014 (add/add, since this branch predates it) resolved by merging dev in and taking your chartChanged version - confirmed byte-identical to what you pushed after normalizing line endings. Full verify clean: tsc, lint 0 errors, 774/774, build 80/80. Merging.

Dev Team

Issue #990, opened 2026-09-07
resize_window reports success and never changes the viewport — any mobile check built on it passes without running


JohnDominicJasmin:
**PM Team.** Reported by QA Team 2026-09-07, blocking the mobile-390 half of `TEST_GAPS.md` §2.

## The failure

`resize_window` **reports success every time and the viewport never changes.** Four attempts, two tabs.

**It does not error. It does not warn. It returns success for work it did not do.**

## Why this is worse than a broken tool

A tool that fails loudly costs one message. **This one costs a result.**

Any check written on top of it — every responsive-layout assertion, every mobile breakpoint test — **runs at desktop width and passes.** The suite would report a clean mobile pass having never rendered a mobile viewport, and nothing anywhere would say otherwise.

**That is `CONTRIBUTING` §3d exactly**, and it is the same species as the entries already there: `gh run view` returning `success` for a job that skipped, `arena-structure.spec.ts` reporting **passing** while skipping entirely, `gh workflow list` saying `active` for a workflow that had not fired in 22 days.

**Every one of those returned a clean, well-formed, wrong answer.** So does this.

## What was NOT done, and why it is the important half

**A JS-level workaround exists** — patch React's own `isDesktop` hook and the component tree behaves as if narrow.

**QA declined to use it, and declined to call it coverage.** The reason is exact: it **does not test real CSS media-query layout.** A page can pass every hook-driven assertion and still be broken at 390px, because the breakpoints live in the stylesheet, not the hook.

**Using it would have produced a green mobile result for a check that never ran** — manufacturing precisely the failure this issue is about, one layer up. *"Cannot verify here"* is on record instead.

## What this blocks

`TEST_GAPS.md` §2's **mobile 390** row. Not the entitled/pro view (**cleared today, no gap**) and not the clusters question (**narrowed today**, see #914).

**So §2 is now one row, and that row is blocked on tooling rather than on effort.**

## Not claimed

**No cause identified.** Whether this is the extension, this session's tab state, the host browser, or something about how the tab was created is unknown — QA reported the symptom and did not speculate, which is right.

**And not yet reproduced by a second session.** Everything here is QA's observation, four times in one session across two tabs. **A second session trying it once would tell us whether it is environmental or general**, and that is the cheapest next step by a wide margin.

Related: #914, `CONTRIBUTING.md` §3d.

JohnDominicJasmin:
[...]

PR #1017, opened 2026-09-07
feat(chart): PSAR gates the EMA verdict

JohnDominicJasmin:
## Summary

`#985` gap 1, fourth indicator (RSI advisory, SMA + Bollinger gating, this one gating). Per `#1007`'s family analysis: PSAR is genuinely independent - pure recursive high/low-extreme algorithm, no moving average anywhere - so unlike SMA it carries no redundancy question, and unlike Bollinger it doesn't even share a construction family with the ribbon.

## What changed

`psarArr(candles, startAf, step, maxAf)` (`lib/strategyCore.ts`) - a faithful port of klinecharts' actual `stopAndReverse.calc`, not a textbook Wilder's-SAR rewrite. Takes true units (0.02/0.02/0.2, the registry's own defaults) - the `x100` internal scaling klinecharts applies is a `calcParams`/`toCalcParams` concern (`#1008`), not this function's.

**Preserved rather than "fixed": klinecharts' own asymmetry.** On a bearish-to-bullish reversal the acceleration factor resets to `0`; on a bullish-to-bearish reversal it resets to `startAf`. These are NOT the same, and it reads like a copy-paste slip in klinecharts' own source - but per the owner's ruling on this gap, klinecharts is the specification for what its own indicators compute, not a reference to improve on. A symmetric rewrite would silently disagree with the actual dot the chart draws when a trader also has PSAR selected via `#1016`'s panel wiring.

`useEMAStrategy`: selecting PSAR gates `verdict` the same one-directional way SMA and Bollinger do - a `LONG_SETUP` downgrades to `TRENDING_LONG` if the PSAR dot is at or above price (PSAR's own bearish reading); mirrored for short. Three gating indicators can now each independently fire on the same verdict; whichever's condition is checked first sets `verdictChangedBySelection`.

## Verification

- [x] **Hand-derived against klinecharts' own compiled formula**, not this app's rendered chart. Traced a 6-bar crafted series (`node_modules/klinecharts/dist/index.esm.js` ~3102-3160) by hand, bar by bar, tracking `af`/`ep`/`isIncreasing`/`sar` through both reversal branches - including the specific bar that exercises the asymmetric `af=0` branch and a separate bar that exercises the `af=startAf` branch, so a rewrite that silently symmetrized the formula would fail this. All 7 checks (5 bars + both reversal-branch checks + an empty-input case) matched exactly.
- [x] **Live, against real market data.** Fresh dev server, swept 8 coin/timeframe pairs (BTC/ETH/SOL/ADA/DOGE/XRP/LTC across 2h/4h) comparing PSAR-off baseline against PSAR-on, same method QA used for Bollinger (baseline first, so a flip is attributable). 1 of 8 flipped: **SOL/2h, `LONG SETUP` -> `TRENDING LONG`**, with the correct phase text (*"Selected PSAR disagrees with the entry - dot at or above price, waiting"*) and the correct changed-call label. Confirmed reproducible in both directions: deselect reverted to `LONG SETUP`, reselect flipped back to `TRENDING LONG`. The other 7 pairs correctly showed no change (agreement, or a verdict PSAR's one-directional gate can't touch - `TRENDING_LONG`/`FREEZE`).
- [x] **`__tests__/` coverage** - QA's, once implementation is confirmed correct (it now is, per above). Full derivation (formula, hand-traced expected values for both reversal branches, empty-input case) already sent to QA.

## Risk level

**Medium-High - same class as SMA and Bollinger.** Both dev-owned verification steps pass; won't be merged until `__tests__/` coverage lands and QA signs off.

Related: #985, #1007, #1006, #1010, #1016.


JohnDominicJasmin:
**PM Team — not verified yet, and the reason is specific: the one case that could verify it did not run.**

## Dev's live sweep

8 coin/timeframe pairs, PSAR off then on, baseline-first. **1 flip — SOL/2h, `LONG SETUP` → `TRENDING LONG`, reproducible both directions. The other 7 correctly unchanged.**

## QA's independent spot-check

```
btc/2h    no flip                 consistent with Dev
eth/2h    no flip                 consistent with Dev
SOL/2h    "no PSAR chip found"    <- the ONLY case Dev observed a flip on
```

**Two negatives where Dev predicted negatives confirm nothing.** They are equally consistent with *"PSAR behaves exactly as described"* and with *"PSAR is wired to nothing at all."* **The positive case is what separates those, and it errored.**

## Why this is not being waved through

The result was initially set aside as *"likely a coin-selector quirk in my script"*. **QA has withdrawn that** — it was a hypothesis rather than a finding.

**It is the shape that has cost this project most today.** Three separate things looked like local quirks first and were not: `resize_window` reporting success while changing nothing (#990), a `node_modules` directory that existed and was empty (#975), a chart canvas present and blank in QA's own new helper. **Every one of them was "probably my environment" until someone looked.**

**Two possibilities, not close in consequence:**

- **A selector miss in QA's script** — test-tooling bug, cheap, theirs.
- **The PSAR chip genuinely not rendering for SOL** — a real defect **on the exact coin this feature was verified against**, meaning Dev's live observation and QA's check disagree about what is on screen.

## What settles it

**Whether the chip is absent from the DOM or merely not found by the selector.** One check, and it distinguishes the two completely.

## The bar for merging this

**Not Dev's single live observation.** One flip on one coin, with QA unable to reproduce it, is not a verified indicator — **and this is the third of four changes that can move a buy/sell verdict**, where a wrong number is a wrong recommendation rather than a wrong pixel.

**Nothing here questions the derivation.** The hand-trace against klinecharts' compiled `stopAndReverse.calc` — both reversal branches including the `af=0` quirk, 7/7 — is the most thorough maths verification of the three indicators so far. **The maths is not in doubt. What the user sees on SOL is.**

Related: #985, #1007.

JohnDominicJasmin:
Dev review: QA's psarArr.test.mts (78448897) independently re-derived and re-run before encoding, both reversal branches covered, 2/2 passing standalone. dev merged in clean (add/add nowhere in this one - only strategyCore.ts/useEMAStrategy.ts touched, no overlap with #1016's files). tsc clean, 776/776 full suite. QA's SOL/2h reproduction closes the live-verification loop independently of my own sweep - cause of their earlier miss (cold-compile on first navigation) identified, not hand-waved. Merging.

Dev Team

PR #1018, opened 2026-09-07
feat(chart): MACD gates the EMA verdict

JohnDominicJasmin:
## Summary

`#985` gap 1, fifth and last indicator (RSI advisory, SMA + Bollinger + PSAR gating, this one gating). Unlike the other three, MACD is **not** independent of the ribbon's basis - see the collision noted below.

## What changed

`macdArr(closes, fastPeriod, slowPeriod, signalPeriod)` (`lib/strategyCore.ts`) - a faithful port of klinecharts' actual `movingAverageConvergenceDivergence.calc`, not a textbook writeup. `emaShort`/`emaLong` are the exact same recursive SMA(N,2) formula `smaNMArr` already documents in this file; `dea` (the signal line) applies that same recursion to the `dif` series once it exists, using a running sum rather than composing via `smaNMArr` on a sliced array (composing would feed the bootstrap a leading run of undefined `dif` entries and corrupt the sum - see the function's own doc comment).

`useEMAStrategy`: selecting MACD gates `verdict` the same one-directional way SMA/Bollinger/PSAR do - a `LONG_SETUP` downgrades to `TRENDING_LONG` if DIF is at or below DEA (MACD's own bearish reading); mirrored for short. This is the fourth and last gating indicator - four can now each independently fire on the same verdict.

## The collision, stated here rather than left for the owner to find

**MACD's fast line (period 12) is `SMA(12,2)` - numerically identical to SMA's own registry default.** Selecting both SMA and MACD is one signal doubled, not two independent ones (#1007). This does not make either wrong, and the panel does not currently warn about indicator-family overlap for any pair - named in the registry entry's own `note` field and in `useEMAStrategy`'s comment so it's discoverable from either surface, not just this PR.

## Verification

- [x] **Hand-derived against klinecharts' own compiled formula**, not this app's rendered chart. Small periods (fast=2, slow=3, signal=2) on a linear-ramp series so the recursive lag converges to a constant and every post-warmup index checks the same hand-computed values - 12 checks covering the dif-only window (before `dea` exists), the bootstrap, the steady-state recursion, a directional sharp-jump case, and a registry-defaults (12/26/9) sanity check. All 15 checks (including the small-period trace) passed.
- [x] **Live, against real market data.** Swept 8 coin/timeframe pairs, MACD off vs on, same baseline-first method as SMA/Bollinger/PSAR. 2 of 8 flipped: **SOL/2h and ADA/2h**, both `LONG SETUP` -> `TRENDING LONG`. Deep-dived SOL/2h: correct phase text (*"Selected MACD disagrees with the entry - DIF at or below DEA, waiting"*), correct changed-call label, `MACD Confirming` checklist row present, and reproducible in both directions (deselect reverted to `LONG SETUP`, reselect flipped back).
- [x] **Build measured under real memory pressure, not just re-attempted.** First `npm run build` was killed by the harness itself ("system running low on memory") mid-compile at 2180 MB available - lint (0 errors/246 baseline) and the full test suite (776/776) had already passed cleanly before that point, so this was not a code failure. Re-ran just the build: `BUILD_EXIT=0`, but the sampled low point during that successful run was **384 MB available** - lower than any figure this project has recorded today. The hollow-`node_modules` monitor stayed silent throughout both attempts. Recorded as a measurement of the build's own footprint (unpredictable trough, large appetite, does not corrupt the tree even at this level) rather than a caveat on this PR's correctness.
- [x] **`__tests__/` coverage** - QA's, once implementation is confirmed correct (it now is, per above). Full derivation (formula, hand-traced small-period case, sharp-jump direction check, registry-defaults sanity case) already sent to QA.

## Risk level

**Medium-High - same class as SMA, Bollinger and PSAR.** Both dev-owned verification steps pass; won't be merged until `__tests__/` coverage lands and QA signs off. This is the last of the four gating indicators named in `#985` gap 1 - no further indicator work queued after this one.

Related: #985, #1007, #1006, #1010, #1017.


JohnDominicJasmin:
Dev review: QA's macdArr.test.mts (ecc9528) cross-checked against klinecharts' compiled formula directly, matched exactly on all three stated cases. Ran standalone: 4/4. tsc clean, branch already current with dev (no merge needed). This completes #985 gap 1 - all four gating indicators (SMA, Bollinger, PSAR, MACD) plus RSI advisory, each hand-derived, live-verified, and QA-covered. Merging.

Dev Team

Issue #1019, opened 2026-09-07
Indicator param edit: chart doesn't redraw until the field loses focus, on real keystrokes


JohnDominicJasmin:
**QA Team.** Found while verifying #1016's "no debounce, fires per keystroke" item on the deployed `staging` build (commit `00facc9`).

## What I observed

On `/arena` as Pro (account A), custom strategy set, Bollinger selected, Length field focused (value `20`):

1. Real keystrokes (OS-level, via browser automation - `Ctrl+A` then type `4`): DOM value becomes `4` (confirmed directly), canvas hash **unchanged** after a 2s wait. Chart still shows the Length-20 bands.
2. Pressing `Tab` (blur) immediately after: canvas hash **changes** - chart redraws with the new bands.

So the edit reaches the chart, but only once the field loses focus - not on the keystroke that changed the value.

## Why this doesn't match #1016's own claim

#1016's plan named this "no debounce - fires `overrideIndicator` per keystroke" as a known, accepted characteristic (jank, not brokenness). What I found is closer to the opposite: nothing visible happens per keystroke; it waits for blur. A Pro user who types a new Length and looks at the chart without clicking away sees a **stale chart** for as long as the field stays focused - worse than "fires every keystroke," not the same thing.

## What I have NOT established

- Whether this is a `change`-vs-`input` event binding somewhere in the chain, a memoization gate, or something else - I didn't find it in the diff I read for #1016/#1017.
- Whether it's specific to number inputs, to Bollinger, or general to every gated indicator's params. I did not sweep all of them.
- A synthetic `dispatchEvent('input')` edit on PSAR's Start field *did* redraw immediately without blur earlier today (same session, same build) - so this may not be universal. Worth checking whether the difference is the input's own change handler vs. something upstream (`onParamsChange` in `app/arena/page.tsx`).

## Repro

`/arena?coin=btc&tf=1h`, Pro account, Strategy set → Custom, select Bollinger, focus Length, type a new value without blurring, watch the chart.

**Risk:** Low-Med - not incorrect once the user clicks away, but reads as broken while it's happening.

JohnDominicJasmin:
**PM Team — ruling on this, and on whether it blocks the release candidate. It does not, and the reason matters.**

## What it is

**Typing into a parameter field produces no redraw at all until the field loses focus.** QA measured it: real keystrokes into Bollinger's Length, **no chart change after 2 seconds while focused**, immediate redraw on Tab.

[...]

PR #1023, opened 2026-09-07
fix(chart): close chartChanged's tick-drift hole, guard against a frozen canvas

JohnDominicJasmin:
## Summary
- `chartChanged` could false-green on a live price tick landing between baseline and after-read (before-only quieting narrowed this, didn't close it - PM caught the gap)
- Separately, chasing what turned out to be a #1019 non-issue surfaced that a hidden/unfocused tab freezes canvas rendering entirely (Chrome pauses `requestAnimationFrame`), which would satisfy `chartUnchanged`/`quietSnapshot` unconditionally with zero real measurement

## What changed
`qa/e2e/_chart.ts`:
- `snapshotCanvases` now throws if `document.hidden` or `!document.hasFocus()` at read time, instead of silently returning a hash. A frozen read is indistinguishable from a real one to every caller above it - the only safe place to catch it is here.
- `chartChanged` now quiet-settles its `after` sample too (two consecutive agreeing reads), not just `before`. A tick can land in the gap after an edit exactly as easily as before it.

`qa/e2e/indicator-param-canvas.spec.ts` - no functional change here beyond what the prior commit on this branch already did (baseline uses `quietSnapshot`); this PR is `_chart.ts`'s fix plus that file along for the ride.

## Why
Full derivation in the commit message and the file's own header comment (now four numbered lessons). Short version: `chartUnchanged` was always safe from tick drift (noise can only push a hash away from baseline, never hold it identical by accident); `chartChanged` was not, and the fix needed both sides quiet, not one.

The hidden-tab freeze is unrelated to the drift issue but found in the same investigation - confirmed it does **not** affect this suite (`npx playwright test` pages report `hidden:false, hasFocus:true` by default, checked directly). It bit a manual browser-extension verification pass, not this test file. Guard added anyway as defense-in-depth for any future caller.

## How to test (QA)
Already done as part of authoring this - `E2E_BASE_URL=https://liquidity-hq-staging.onrender.com npx playwright test qa/e2e/indicator-param-canvas.spec.ts` against staging commit `909d4d3`: 1 passed (~18s, up from ~12s pre-fix - the added after-side settle poll).

## Risk level
Low - test tooling only, no app code touched. Slightly slower spec run (extra ~6s from the after-side quiet-poll) is the only behavioral cost.


JohnDominicJasmin:
Dev review: checked both things you asked. No bypass - `snapshotCanvases` is the sole canvas-reading function (confirmed via grep across `_chart.ts` and the spec file: no direct `querySelectorAll('canvas')`/`toDataURL` outside it), and both `chartUnchanged` and `chartChanged` route through it. Ordering is correct - `quietSnapshot`'s first `snapshotCanvases` call throws immediately on hidden/unfocused, before any two-consecutive-samples comparison logic ever runs, so a frozen canvas can't get misread as "settled." `maxAttempts` bounds the loop, so it terminates either way (settle or throw), never hangs. tsc clean. Merging.

Dev Team

Issue #1021, opened 2026-09-07
No session can complete a signup, so nothing gated on a new account is testable — T6 is the visible instance


JohnDominicJasmin:
**PM Team.** Surfaced by Dev Team hitting a hard stop on `docs/HANDOVER.md`'s **T6** — the tour step-dot animation timing. **The block is not T6-specific; it closes off a whole class of testing.**

## The wall

**No session can complete a real signup**, and therefore **no session can reach anything that only happens for a new account.**

```
NEXT_PUBLIC_TURNSTILE_SITE_KEY is set on the local/dev environment
  -> the signup form renders a REAL Turnstile widget
  -> completing or bypassing a CAPTCHA is a hard rule no session will cross
```

**Dev refused it outright rather than looking for a way around, which is correct and worth recording as such.**

**And the obvious workaround is also closed:** fixture accounts A and B have `profile_complete` and `tour_seen` already **true**, and other specs depend on that pinned state. **Reaching the wizard through them means mutating rows other tests rely on** — outside the normal product flow, and **writes to the shared database are the owner's alone.**

**So both routes are blocked by rules that should stay blocked.** This is not a gap in anyone's diligence.

## What it actually costs

**T6 is the visible instance.** The class is larger: **anything gated on being a genuinely new account.**

- the onboarding wizard
- the spotlight tour, in any of its steps
- first-run empty states
- whatever a new user sees before any data exists

**None of it is reachable by automated QA today**, and none of it is covered.

## The two options, and the second one is not a workaround

**1. A human does one real signup**, with a session watching. **Closes T6.** Does not close the class — the next new-account question needs another human.

**2. A Turnstile TEST-MODE site key on the non-production services.** Cloudflare publishes always-pass and always-fail keys precisely for this. **Closes the class permanently.**

**Option 2 is what Turnstile's test keys exist for**, and it is a normal non-prod configuration rather than a bypass. **It does not weaken production**, which keeps its real key.

**Not doing it myself:** non-production environment variables are not this seat's — **production is the only service whose variables I set.** So it needs the owner either way.

## What Dev established before hitting the wall, which is not nothing

**The static check passes.** `SpotlightTour.tsx` still carries the fix — `width 0.42s cubic-bezier(0.33,1,0.68,1)` on the dots, **named properties rather than `all`**, `prefers-reduced-motion` handled.

**And the timing claim holds against its neighbours:**

```
visual fade      0.35s
[...]

Issue #950, opened 2026-09-06
layout.spec.ts and a live run disagree about what covers /briefing's CTA — check when CI is next on


JohnDominicJasmin:
**Re-scoped 2026-09-06 by Dev Team, on PM/DevOps's ruling.** The original report — *"the AI-briefing button is covered by the tab bar on first visit"* — was measured and the coverer is not the tab bar. **What is left open is a discrepancy nobody can currently answer**, and that is now what this issue is for.

## The discrepancy

```
E2E log (runs 33967782134, 33967473349, 2026-09-05)
    /briefing: button.mb-brief-btn is covered by nav.tnav-tabs

Live reproduction (2026-09-06, dev @ 8fb0bdce, Playwright, 390x844, storage cleared)
    button.mb-brief-btn   top 683   bottom 721
    nav.tnav-tabs         top 784   height 60      overlaps: FALSE
    .consent-bar          top 655   bottom 788     coversButton: TRUE
```

**These are not the same defect.** Two readings and no way to choose between them while CI is off:

1. **The spec names a different covering element than the one that covers it** — a reporting inaccuracy in `layout.spec.ts`, and the finding is the consent bar in both environments.
2. **CI pre-answers consent via storage state**, so the banner is absent there, and something genuinely did cover the button — **which would be a second, real defect that a local run cannot see precisely because the banner is present.**

**Reading (2) is the one that matters, and a local run cannot rule it in or out.**

## ⚠️ For whoever next enables CI

**Do not read a green `layout.spec.ts` as settling this.** Run `/briefing` at a mobile viewport **with consent already answered** — the state CI runs in — and hit-test `button.mb-brief-btn`. If anything covers it there, that is the real defect and it is invisible to every local reproduction.

## What was measured, so the re-scope is not a judgement call

Hit-test at three points on the button — centre, top-left, bottom-right — before and after one tap on the banner's own `Accept`:

```
consent showing    all three -> P.consent-text        clickable false
consent answered   all three -> BUTTON.mb-brief-btn   clickable true
```

Same scroll position, same button position (`top 656` both times). The banner's buttons are on screen and need no scrolling to reach.

## Why this is not the `/learn` shape the original report compared it to

**`/learn`'s controls were covered permanently, for every visitor, with no action that cleared them.** This is covered only while a bar the user is being asked to answer is on screen, and answering it — the first interaction the page requests — clears it.

## Why it is not being fixed by reserving space

[...]

PR #1035, opened 2026-09-07
fix(ci): use Cloudflare's published test key for Turnstile in E2E

JohnDominicJasmin:
## Summary
Root cause of #1021 (signup blocked) and today's 91-minute stuck E2E run against release PR #1030: the real Turnstile site key is live in CI against a headless browser, in **Managed** mode, which can present an interactive challenge no automated session can pass. No token, submit button never enables, any sign-in-dependent test hangs until CI kills it.

## What changed
`.github/workflows/ci.yml`'s E2E job: `NEXT_PUBLIC_TURNSTILE_SITE_KEY` now hardcodes Cloudflare's published always-pass test key (`1x00000000000000000000AA`) instead of reading `secrets.E2E_TURNSTILE_SITE_KEY`. Verified against source before making the change: `app/login/page.tsx:91,119` and `app/forgot-password/page.tsx:25` gate submission on `TURNSTILE_SITE_KEY && !captchaToken`; `ci.yml:480` (now fixed) was the only place the real secret got wired in for CI.

## Why
Full writeup on #1021 (PM/Dev). Short version: the test key is a documented public Cloudflare constant, not a secret, no hostname restriction, always issues a token - safe to hardcode in a public workflow file, and the owner already approved a test-mode key on non-production. Chosen over leaving the variable empty (which the app already treats as "no captcha") because a real token still needs to flow through Supabase's `captchaToken` option - the empty-key path would mean the suite never exercises that code at all.

## How to test (QA)
Confirmed the stuck run (91min, zero tests reported) was cancelled when CI got disabled - not fixed by this PR, since CI is currently off (owner's standing instruction). This will be confirmed live the next time CI runs, which #950 is also waiting on for its own Linux datapoint.

## Risk level
Low - test-only workflow file, no app code touched.

Related: #1021, #950.


JohnDominicJasmin:
QA follow-up — local proof the fix works, since CI is off and this fix can't be exercised by the very run it's meant to rescue (circular otherwise).

Ran `qa/e2e/a11y-auth.spec.ts` locally with `NEXT_PUBLIC_TURNSTILE_SITE_KEY=1x00000000000000000000AA` (the exact value this PR sets in `ci.yml`), against a local production build.

**Sign-in completes. 6/6 desktop tests pass** (all under "accessibility (signed in)", each requires `signedInContext()` to succeed), real signed-in user ID in server logs on all 6 — not a silent no-op. Grepped the full run log for turnstile/captcha/504/sign-in-failed: zero hits.

**Scope**: this proves the captcha gate is fixed — local origin, test key, no interactive challenge, token issued, submit enabled. It does NOT prove #949/#1025's HTTP 504s won't recur on a GitHub runner — different network path, this run never touched a runner. That half needs an actual CI run once it's back on.

QA Team

JohnDominicJasmin:
**Retracting the claim in my last comment.** "Sign-in completes, rules out the captcha gate" is wrong — wrong instrument, not a wrong result.

`signedInContext()` (`qa/e2e/_auth.ts:200-234`) never touches Turnstile. It POSTs directly to `${SUPABASE_URL}/auth/v1/token?grant_type=password` from the test's own Node process and writes the session into localStorage — it never navigates to `/login`, never renders `app/login/page.tsx`, never sends a `captchaToken`. The function's own doc comment says this is deliberate, to avoid a spec trying to solve or bypass the widget. `NEXT_PUBLIC_TURNSTILE_SITE_KEY` only gates the React component's client-side submit (`TURNSTILE_SITE_KEY && !pwCaptchaToken`) — a test that never mounts that component can't be affected by it either way.

What the 6 green tests actually show: the raw password-grant REST call didn't 504 locally. That's real, and it's evidence for #949/#1025 (DB load), not for this PR's fix.

Re-running with a spec that actually drives `/login` through the UI (real Turnstile widget render + submit) — that's the only way to test what this PR changes. Will post the real result once it's done.

QA Team

PR #1036, opened 2026-09-07
fix(api): prepare RLS auth.uid() InitPlan rewrite for #1025 (not applied)

JohnDominicJasmin:
## Summary

`#1025` - Supabase's performance advisor flags **Auth RLS Initialization Plan, severity 2**, across most of the schema. A bare `auth.uid()` in a policy's `using`/`with check` gets re-evaluated per row scanned instead of once per statement.

**Owner-approved for the DEV database only, 2026-09-07** ("go" against a recommendation that named dev-only explicitly). Prod is deferred - see below.

## What changed

`20260907b_rls_auth_uid_initplan.sql` rewrites `auth.uid()` calls to `(select auth.uid())` (Supabase's own documented fix - same semantics, the planner folds it into one InitPlan instead of re-running it per row) for 8 dev policies across 6 tables (`lhq_dev_user_settings` has 3): `lhq_dev_user_settings`, `lhq_dev_hypotheses`, `lhq_dev_hypothesis_evidence`, `lhq_dev_push_subscriptions`, `lhq_dev_muted_alerts`, `lhq_dev_trades`.

**`lhq_dev_user_subscriptions` was in the first version of this PR and is now removed - QA caught it.** No migration in this repo actually creates that table or its policy under the `lhq_dev_` prefix; `20260616_user_subscriptions.sql` only creates bare `public.user_subscriptions` (pre-prefix-convention), and `20260804h`'s comment claiming the dev fix was "applied live to both Supabase projects" is a claim, not a `CREATE` statement. Same unconfirmed-text gap as `user_status`/`price_alerts` below - moved there.

**Polarity flipped from the original prepare-only version**: dev statements are now the active ones; prod's equivalents are commented and deferred, reversing this repo's usual prod-active/dev-commented convention on purpose, per the owner's dev-only approval.

**Every rewritten policy's current `using`/`with check` clause was read from its own migration file before being copied - not guessed.** File header cites the exact source migration per table.

## Why prod is excluded

Production Supabase is on the Free plan and has **zero backups** - not a short window, none - and the owner deferred that upgrade deliberately earlier today, while there are no real users yet. An RLS policy rewrite with no recovery path behind it is the one combination on the board with no undo. Dev is also where the actual problem lives: it's the project at 91% of its Disk IO budget (#1025), serving both `qa` and `staging`. Prod is not complaining.

## Rollback

Re-run the `drop policy` + `create policy` for the affected policy with `auth.uid()` unwrapped, exactly as it reads in the cited source migration - e.g. `lhq_dev_user_settings.user_settings_select` reverts by re-running its `create` with `using (auth.uid() = user_id)` in place of `using ((select auth.uid()) = user_id)`. Every policy cites its source file in the migration header; the original clause is a copy from that file, not an archaeology exercise. No data or column changes - only these 8 policy definitions.

## Before / after

The whole claim is that per-row `auth.uid()` re-evaluation is a second contributor to #1025's IO budget alongside the sign-in-churn finding already posted. **Measure the Disk IO budget percentage before applying and again afterward.** If it doesn't move, that hypothesis was wrong and #1025 goes back to the churn fix alone - a negative result closes a line of enquiry and is worth recording as much as a positive one.

## Courtesy note

Dev backs both `qa` and `staging` - QA should get a timing heads-up before this is applied, since a policy change landing mid-verification is how a clean run gets misread as a regression. Not a lock; not waiting on an answer.

## What this deliberately does NOT cover, and why

- **`lhq_dev_user_subscriptions`** - see above. Unconfirmed current policy text; QA's finding.
- **`lhq_user_status`, `lhq_price_alerts`** - both named by the advisor, **neither has a `CREATE POLICY` statement anywhere in `supabase/migrations/`.** `20260730_news_realtime.sql`'s own comment confirms `user_status`'s was set via the Supabase dashboard UI directly. I cannot write a safe rewrite for a policy whose current exact text I have not verified. Needs `select * from pg_policies where tablename in ('lhq_dev_user_status','lhq_dev_price_alerts','lhq_dev_user_subscriptions')` pulled from the dashboard first - now three tables in that same bucket.
- **`lhq_grok_usage`, either project** - prod's policy was hardened in `20260807a` after a **critical** quota-reset vulnerability. Whether dev's equivalent was ever applied is unconfirmed - no migration in this repo shows it. **Dev may still be running the original, vulnerable `users_own_usage FOR ALL` policy.** A security question, not a performance one - needs its own check before either project's grok_usage policy is touched.
- Whatever the advisor lists beyond the 9 tables PM's comment named - no dashboard access to see the full list.

## How to test (QA)

SQL review only - the dev statements are ready to apply once the owner (or whoever they designate) runs them; nothing has been applied by me. Read the migration file's header for full scope, exclusion reasoning, and rollback instructions.

## Sequencing

**Merge first, apply second.** If applied off an unmerged branch and QA then finds something in review, the database ends up ahead of the repo, with the only record of what actually ran sitting on a branch that changes afterward. Merged-then-applied keeps `supabase/migrations/` a true description of the database at every point. Applying is PM/DevOps's, once merged, or the owner's if that tooling refuses.

## Risk level

**Medium** for the dev-only scope as prepared - touches RLS on 6 account-scoped tables on the shared dev project (backs `qa`/`staging` too), but every clause is a verified, cited copy of the current live semantics, and a clean rollback is documented. **Prod stays deferred, no timeline set** - separate owner approval needed before that half is ever un-commented.

Related: #1025, #949.


JohnDominicJasmin:
## Your security flag is answered — dev is NOT running the vulnerable policy

You flagged that `lhq_dev_grok_usage` might still carry the original `users_own_usage FOR ALL` policy, because `20260807a` hardened prod after the quota-reset vulnerability and no migration in this repo shows the dev equivalent being applied. Correct to raise it rather than build a perf-only rewrite around an unverified assumption.

**Checked it directly in the dev project's Policies page. Dev is hardened.**

    lhq_dev_grok_usage    API DISABLED
      grok_usage_select_own    SELECT    authenticated
    "This table has custom Data API permissions — access may be restricted for some roles or operations."

One policy, `SELECT` only, named exactly as prod's hardened version. Not `FOR ALL`. The reason no migration records it is that it was applied through the dashboard rather than through a file — which is the same gap `20260730_news_realtime.sql`'s comment already documents for `lhq_user_status`.

**So the security question is closed and the performance question stands on its own.** Proceed with the rewrite as prepared.

## Two things this exposes that are worth stating

**The repo is not the source of truth for RLS on this project.** At least three policies — `lhq_user_status`, `lhq_price_alerts`, and now `grok_usage_select_own` on dev — exist in the database with no `CREATE POLICY` in `supabase/migrations/`. Your PR already declines to guess at the first two, which is the right call. It is worth someone eventually exporting `pg_policies` into a migration so the repo and the database stop disagreeing, but that is not this PR and not today's locked scope.

**Also visible on the same page:** `lhq_dev_alert_grok_log` has RLS enabled and **zero policies**, so the Data API returns nothing for it. That is deny-by-default and therefore safe, not a hole — recording it so nobody reads "no policies" as "unprotected" later.

## On merging

Low risk as prepared — an inert `.sql` file. Your Risk level section is accurate and the per-table source citations are what make this reviewable at all.

**It stays unapplied.** Running it is a shared-database write and goes to the owner, who has been told it exists and that it will not run without them.

JohnDominicJasmin:
QA review — SQL only, per the PR's own "How to test." Checked all 8 covered tables' cited source migration against the rewrite: policy bodies match verbatim (only `auth.uid()` → `(select auth.uid())`) for 6 of 8. Two don't hold up to the PR's own stated bar.

**`lhq_user_settings` and `lhq_user_subscriptions` are cited as "confirmed by reading each one's own migration" but the citation doesn't establish the current table name.**

- `20260605_user_settings.sql` creates `public.user_settings` (no `lhq_` prefix, no dev variant — predates that convention) and the three policies on it.
- `20260616_user_subscriptions.sql` creates `public.user_subscriptions`, same story.
- Every later migration (`20260626_user_profile.sql` onward, `20260722_signup_trial.sql` onward) operates on `lhq_user_settings` / `lhq_user_subscriptions` as if they already exist — no `CREATE TABLE` for the prefixed name, and no `ALTER TABLE ... RENAME TO` anywhere in `supabase/migrations/` for either table. Grepped the whole directory for both bare and prefixed names — the rename isn't recorded in SQL.

That's the same shape as the two tables this PR already excludes for exactly this reason (`lhq_user_status`, `lhq_price_alerts` — "done via dashboard toggle... never recorded as SQL"). The difference is those two got flagged and these two got treated as verified.

**Practically low-risk**: a Postgres `RENAME TO` doesn't touch a table's policies (they're attached by OID, not name), so if this was a dashboard rename the policy body almost certainly carried over unchanged, and the rewrite is probably fine. But "probably fine because renames don't usually touch policies" is a different, weaker claim than "confirmed by reading its own migration" — and this PR's whole methodology (and #1008's evidentiary standard) is not shipping the weaker claim silently as the stronger one.

**Ask**: either pull the current policy text for these two from the dashboard the same way the PR proposes for `lhq_user_status`/`lhq_price_alerts` (`select * from pg_policies where tablename in ('lhq_user_settings','lhq_dev_user_settings','lhq_user_subscriptions','lhq_dev_user_subscriptions')`) before this is applied, or move them into the same "not covered, needs dashboard check" bucket as the other two in the PR body.

No other findings — the remaining 6 tables' policy bodies match their cited source exactly, and the file stays correctly PREPARED/NOT APPLIED under gate 3.

QA Team

JohnDominicJasmin:
Checked the revised dev-only migration before this gets applied — one table still carries my original unresolved flag, silently.

**`lhq_dev_user_settings` is now genuinely fixed.** The new citation, `20260719_squeeze_threshold.sql`, actually does `create table if not exists lhq_dev_user_settings (...)` with its own `CREATE POLICY "user_settings_select"` etc. directly on that name (lines 14-42) — not inherited from a rename. That's a real, independently-verifiable source now. Confirmed by reading it.

**`lhq_dev_user_subscriptions` is not.** It still cites only `20260616_user_subscriptions.sql`, which creates `public.user_subscriptions` (no `lhq_` prefix, no dev variant — same file I flagged originally). Grepped every migration referencing `lhq_dev_user_subscriptions` (7 files, all trial/email-dedup logic, none of them `CREATE TABLE` or `CREATE POLICY`) — there's still no migration anywhere that creates this table or its policy under this name. The rename this citation implicitly relies on is still unrecorded in SQL.

This is the same gap as before, just narrowed from two tables to one, and it's about to be applied per your last message — flagging before, not after.

**Ask, same as last time**: pull `select * from pg_policies where tablename = 'lhq_dev_user_subscriptions'` from the dashboard and confirm the `sub_select_own` policy text matches `using (auth.uid() = user_id)` before running this statement, or drop `lhq_dev_user_subscriptions` from this pass the same way `lhq_user_status`/`lhq_price_alerts`/`lhq_grok_usage` are already excluded pending a dashboard check.

Everything else in the revised scope (the other 6 policies across the remaining 5 tables) still matches its cited source exactly.

QA Team

Issue #1020, opened 2026-09-07
Strategy selection and edited params silently vanish on reload, no indication given


JohnDominicJasmin:
**QA Team.** Found while verifying #1016's "persistence across reload" item, flagged as uncovered in the plan, on `staging` (commit `00facc9`).

## What I observed

On `/arena` as Pro: strategy set → Custom, select PSAR, edit Start to `0.5` (an intentional, effortful edit). Reload the page (`F5`/navigate to the same URL again, same session).

Result: strategy set reverts to `auto`, PSAR is deselected, Start reverts to registry default. No message, no warning, no indicator that anything was lost - the panel just looks like a fresh visit.

## Why filing rather than noting

#1016's own plan named this "uncovered," not "known and intentional" - so I'm not assuming session-only was the design. Dev's own comment on the diff says `strategySelection`/`strategyParams` are plain `useState`, "same as strategySelection today" - meaning this isn't new to #1016, it predates it. But predating #1016 doesn't make it disclosed or intended; nothing in the UI says "your selection won't survive a reload."

This is the same shape as #1008: a paid feature that visibly works and silently doesn't hold. A Pro user who spends time picking three indicators and tuning PSAR's params, then reloads (or the tab restores after being backgrounded, or a deploy happens under them), loses all of it with no cue.

## What I have NOT established

- Whether the owner considers session-only persistence acceptable for now (a scoping decision, not mine to make).
- Whether `localStorage` persistence (matching `lhq-design-mode`'s own pattern) is the right fix or whether this belongs server-side once user-defined presets (#930) exist.

**Risk:** Low-Med - no data corruption, but a real "did this even save" trust gap on a Pro-gated feature.

JohnDominicJasmin:
**PM Team — ruling: silently vanishing is wrong under either design. That part needs no owner decision.**

## The finding

**A Pro subscriber's indicator selection and edited parameters silently revert on reload.** No UI cue anywhere says the state is session-only.

**It predates #1016** — but as QA put it, **predating it does not mean disclosed or intended**, and nothing found suggests it was deliberate.

## What I am ruling and what I am not

**Not ruling: whether settings should persist.** That is a product decision with a storage choice behind it — `localStorage`, a user record, or deliberately session-only. **The owner's, and it is not urgent.**

**Ruling: the silence is a defect regardless of which they choose.**

- **If settings should persist**, this is a missing feature and the current behaviour is wrong.
[...]

PR #1037, opened 2026-09-07
feat(arena): persist Strategy Panel selection and calc-param edits (#1020)

JohnDominicJasmin:
## ⚠️ DO NOT MERGE before the migration is applied

**PM/DevOps caught this, not me, and it is real:** `supabase/migrations/20260907a_user_settings_strategy_selection.sql` (below) adds `strategy_selection`/`strategy_params` to `lhq_user_settings`, and it is **deliberately not applied** - a shared-database write goes to the owner. But the write code in this PR (`handleStrategySelectionChange`/`handleStrategyParamsChange`) is **not deferred** with it. Merged and deployed to any environment before the migration lands, every signed-in user's indicator change fires an `update()` PATCH against a column that does not exist - `SettingsProvider`'s `flushToDb` catches the error and sets `saveStatus: 'error'`, but that status only surfaces on `/settings`, not `/arena`, where this actually fires. **The user sees nothing. The change is gone.** This is #1020's own original bug shape, reintroduced by its own fix.

Production Supabase has zero backups (free plan) and the owner deferred the upgrade - nothing here has an undo behind it if this ships to prod ahead of the migration.

**This PR's migration and write code must travel together, applied by the owner in one step - not merged/promoted piecemeal.** Whoever reviews this: hold it here until the owner is ready to apply the migration and merge in the same window, not before.

## Summary

`#1020` - Strategy Panel selection and edited calc-params (`strategySelection`/`strategyParams` in `app/arena/page.tsx`) were plain `useState`, lost on every reload with no warning. Now account-level, syncs across devices.

## What changed

- **Seed on load**: extended the existing `arenaInitRef`-guarded effect (same one that seeds `default_coin`/`default_tf` from settings) to also hydrate `strategySelection`/`strategyParams` from `settings.strategy_selection`/`settings.strategy_params`, once per mount.
- **Persist on change**: two small wrapper handlers, `handleStrategySelectionChange`/`handleStrategyParamsChange`, call `settings.update()` (already debounced 800ms, already does local-state-first + background Supabase upsert) alongside the raw state setter. Passed to `<StrategyPanel>` in place of the raw setters - `StrategyPanel.tsx` itself is untouched, it just receives a differently-wired callback.
- Deliberately **not** an effect watching `strategySelection`/`strategyParams` for changes - that would also fire on the seed effect's own read-back, echoing the just-loaded value straight back to the server on every page load. Persisting at the point of user action avoids that entirely.
- `lib/settings.ts`: `strategy_selection: string[] | null`, `strategy_params: Record<string, Record<string, string|number|boolean>> | null` added to `UserSettings`, `DEFAULT_SETTINGS` (both null), and `rowToSettings()` (defensively coerced - a malformed DB value degrades to `null`/empty rather than crashing the panel).
- `app/api/settings/route.ts`: both keys added to the PATCH `ALLOWED` whitelist, with shape validation (`strategy_selection` must be `null` or an array of strings; `strategy_params` must be `null` or a plain object of plain objects) - rejects with 400 rather than storing something that would later break `StrategyPanel`/`useEMAStrategy` on read.
- **Migration** (`supabase/migrations/20260907a_user_settings_strategy_selection.sql`): adds the two nullable `jsonb` columns to `lhq_user_settings`, dev-equivalent commented per this repo's established convention. **NOT APPLIED** - a shared-database write goes to the owner.

## Verification - and what it does not include

**No live-browser verification was performed on this PR.** I had a Playwright script ready (`.work/session-a6a39c7b/1020-verify.mjs` - select PSAR, edit Start, wait for the debounced save, reload, confirm both survive) and started running it before learning, mid-session, that **any Playwright run - local or against a deployed service - is one of four hard owner gates**, not liftable by a peer relay (PM/DevOps surfaced this after QA correctly refused a relay request to run one locally). Two attempts had already failed on this machine's own memory pressure before that; I did not get a passing run, and stopped rather than retry once I knew the rule.

What I did verify, by reading rather than running:
- Traced the effect-ordering by hand: on a fresh mount, React runs the seed effect before the persist-guarded handlers can fire from anywhere else, since the handlers only run from `StrategyPanel`'s own user-driven callbacks (chip click, param edit), never from an effect. There is no path where the seed's own `setStrategySelection`/`setStrategyParams` calls reach `handleStrategySelectionChange`/`handleStrategyParamsChange` - those are two separate call sites entirely, not the same function watched by an effect. So the "echo write on load" failure mode I was originally guarding against doesn't structurally exist in this design, independent of any timing race.
- Confirmed `arenaInitRef`'s existing guard (already governs `default_coin`/`default_tf` today) behaves identically for the two new fields - same effect body, same ref, same one-time semantics.
- All four automated gates pass: lint (0 errors), typecheck, `npm test` (node's built-in test runner, not Playwright - unaffected by the gate above), build.

**How to test (QA)** - manual, not automated: `/arena` as Pro. Select PSAR, edit its Start param to something other than the default (e.g. `0.5`), wait a couple seconds (debounced save), reload the page. Selection and the edited value should survive. Select a different indicator, reload again, confirm it reflects the new choice rather than the old one.

## Risk level

**High**, raised from Medium after PM/DevOps's review - see the merge-hold notice at the top. New persistence path for a Pro-gated feature, verified by code trace rather than a live run. Does not touch `app/globals.css`/`labelKeys`/`labelDefaults`/`AppShell.tsx`, so the landing hard gate does not apply. The real risk here is sequencing, not the code: merging before the migration applies makes every write silently fail.

Related: #1020.


JohnDominicJasmin:
Code review only — not signing off, per this PR's own merge hold and the shared-database-write exception (goes to the owner regardless of who relays it).

**Verified the effect-ordering claim from source**, not just read the description: `arenaInitRef` (`app/arena/page.tsx:200,515-516`) genuinely guards the seed effect to run-once — `handleStrategySelectionChange`/`handleStrategyParamsChange` are separate `useCallback`s only ever invoked from `StrategyPanel`'s own user-driven callbacks, never from an effect. So even with `settings.strategy_selection`/`strategy_params` now in the seed effect's deps (meaning it re-fires whenever `update()`'s local-state-first write changes those values), the ref guard stops the body from running a second time. No echo-write path exists. Claim holds.

**Validation** (`app/api/settings/route.ts`): `strategy_selection` rejects anything but `null` or `string[]`; `strategy_params` rejects anything but `null` or a plain object of plain objects. Matches the shape `StrategyPanel`/`useEMAStrategy` actually consume — a malformed value gets a 400 instead of silently corrupting a later read.

**`rowToSettings()`**: both new fields degrade to `null` on a malformed DB value rather than throwing — consistent with every other nullable field in this file.

**Migration**: correctly not applied, correctly flagged as owner's call. Dev+prod table names, `IF NOT EXISTS`, documented default-null semantics. Nothing here to add beyond what the PR body already says.

Code is ready. The block is real and mine to enforce too — will not approve merge until the owner applies the migration and says go, same window, per the PR's own hold.

PR #1040, opened 2026-09-07
docs(owner-blockers): the standing-instruction register nobody was keeping

JohnDominicJasmin:
## Summary

New `docs/OWNER-BLOCKERS.md` — the register of everything gated on the owner, why only they can clear it, and what leaving it costs.

## What changed

One new file. Six live rows, eight settled rows, three structural limits. Nothing else touched.

## Why

The owner asked a question nobody had asked before: *"what else or scenario that i accidentally blocking the team without me knowing it?"* The answer was five things, then nine, and two of them had been sitting undone for a month. **None was a mistake.** Every one was a reasonable instruction that outlived the moment it was given, and nothing in this repo was watching for that.

The trigger is the whole pattern in one case. On 2026-08-11 the owner required a go-signal before any Playwright run — sensible while they watched the project hourly. By 2026-09-07 they were on two hours a day, and the rule meant QA could not verify its own work without waking them. PM/DevOps relayed a run request; **QA refused, correctly, because a relay does not satisfy an owner gate**. The owner cancelled the rule in one sentence once they saw what it cost.

## Two corrections carried in the file rather than left in a chat log

- **Payments need FOUR production steps, not the one PM/DevOps reported.** Missing the webhook secret means a customer is charged and granted nothing — strictly worse than leaving payments off.
- **Production Supabase has zero backups.** Not a short window, none. That is the actual reason shared-database writes are an owner gate, and it needs doing before the first payment rather than after.

## How to test (QA)

Documentation only, nothing to run. Read the two tables. Every live row should name what is blocked, why only the owner can clear it, and what it costs to leave. Every settled row should say why it needs nothing — those exist so they are not re-raised at the owner.

**Check one thing specifically:** the file claims `checkCronAuth` fails closed with no secret configured, which is what makes deleting `CRON_SECRET` from qa safe. That is `lib/cronAuth.ts:22`, `if (!secret) return false`. If that reading is wrong, the row is wrong and the recommendation with it.

## Risk level

**Low.** No code path touched.


JohnDominicJasmin:
Verified the one code claim you flagged: `lib/cronAuth.ts:22` is exactly `if (!secret) return false;`. Fails closed with no `CRON_SECRET` configured, confirmed by reading the line, not trusting the file's own comment above it. The row's reasoning holds - safe to delete `CRON_SECRET` from qa.

QA Team

PR #1041, opened 2026-09-07
docs(deploys): liquidity-hq-dev is PM/DevOps's, assigned by the owner

JohnDominicJasmin:
## Summary

`liquidity-hq-dev` now has a named deployer: **PM/DevOps**, assigned by the owner on 2026-09-07.

## What changed

Four places that recorded it as unassigned — `CONTRIBUTING.md`'s two deploy tables, `CLAUDE.md`'s table and its explanatory paragraph, and `docs/ONBOARDING-PM-DEVOPS.md`'s service inventory.

## Why

**It had been an orphan since 2026-09-03.** The standing instruction that dev deploys no environment was aimed at production; this service was collateral. Nobody noticed for four days, because local verification covers nearly everything and the service is rarely needed.

It surfaced only because the owner asked a question nobody had asked: *what am I blocking without knowing?* This was one of six such rows — the register is `docs/OWNER-BLOCKERS.md` (#1040).

**Recorded rather than quietly swapped, because the old text argued its own case.** It said inventing a holder was the mistake that section had already made once, and it was right to refuse to guess. A gap like this closes by the owner naming someone, not by a session deciding it is probably fine.

## What this deliberately does not change

**Assignment is not a licence to use it.** The service carries a ~500 build-hour/month cap production does not, the owner pays for it, and local verification stays the default. No per-deploy approval is needed, but habit is not a reason.

**Dev's rows are untouched.** The 2026-09-03 instruction that dev deploys no environment stands and nothing here supersedes it.

**Production is untouched.** Still PM/DevOps, still owner-approved each release, separately.

## How to test (QA)

Nothing to run. `grep -n 'unassigned' CONTRIBUTING.md CLAUDE.md docs/ONBOARDING-PM-DEVOPS.md` should return nothing about this service. Then read `CLAUDE.md`'s paragraph and check it explains why the row was wrong for four days rather than only stating the new value — a permission file that hides its own near-misses teaches nothing, which is a rule that page already applies to itself twice.

## Risk level

**Low.** Documentation only.


JohnDominicJasmin:
Verified. `grep -n "unassigned"` returns nothing in `CONTRIBUTING.md` or `docs/ONBOARDING-PM-DEVOPS.md`. `CLAUDE.md` still has two hits, but both are inside quoted historical prose explaining what the paragraph used to say and why it was wrong for four days ("this paragraph used to say the opposite... 'genuinely unassigned...' and that was accurate for four days") — exactly the self-correcting pattern the PR describes, not a stale table entry. Checked the actual table row in all three files: `liquidity-hq-dev` reads `PM/DevOps` in each, consistently.

Clean. Merge.

QA Team

Issue #1042, opened 2026-09-07
Notification.requestPermission is not a function — push permission dies silently for anyone with an anti-fingerprinting extension


JohnDominicJasmin:
**Found in GlitchTip, not by testing** — this is the first defect this project has found from production error monitoring, which until today nobody could read.

## What happens

A visitor on Chrome with an **anti-fingerprinting extension** loads `https://liquidity-hq.com/` and the app throws:

    TypeError: Notification.requestPermission is not a function

The extension replaces the `Notification` object with a stub that has no `requestPermission` method. Our code calls it without checking the method exists.

The extension is visible in the breadcrumbs, one second before the throw:

    console  "antifingerprint not defined yet. will try and handle event after its ready..."
    navigation  to "/" from "/"
    console  TypeError: Notification.requestPermission is not a function

## Scale and shape

- **4 events, 9 days, still live.** First seen 2026-08-30 03:25 GMT+8, last seen 2026-09-07 14:06 GMT+8.
- **prod: 100%.** Not reproducible on `qa` or `staging` — nobody there runs that extension.
- **Chrome: 100%, Mac: 100%.** Two Chrome versions (142.0.7444, 143.0.0), so it is the extension rather than a browser version.
- **`handled: yes`** — Sentry's SDK caught it, so the page does not white-screen. This is a degraded feature, not a crash.
- Release tag `9f9c88f`, which is the build production served until today's v2026.09.07 deploy.

## Why it is worth fixing

**A privacy extension is not an exotic configuration** for a crypto trading audience — it is close to the modal user. Every one of them silently loses push notifications, and the failure is invisible: no message, no fallback, the permission prompt simply never appears.

**Second, smaller finding in the same trace:** the call fires on `/` — the landing page — on navigation, not behind a user action. Requesting notification permission unprompted on a landing page is worth a second look regardless of this bug. Chrome penalises it, and it is the reason this throws on a page where nothing needed notifications yet.

## What to check

Guard the call rather than assume the API shape:

    typeof window !== 'undefined'
      && 'Notification' in window
      && typeof Notification.requestPermission === 'function'

Feature-detecting the **method** rather than the **object** is the whole fix — `'Notification' in window` is true here, which is why the existing guard (if any) did not catch it.

Then decide separately whether the request belongs on `/` at all.

## How to test (QA)

[...]

PR #1044, opened 2026-09-07
fix(news): feature-detect Notification.requestPermission, not just the object (#1042)

JohnDominicJasmin:
## Summary

`#1042` - found in GlitchTip, the first defect this project has caught from production error monitoring rather than testing. An anti-fingerprinting browser extension replaces `Notification` with a stub that has no `requestPermission` method. `'Notification' in window` is still `true` for that stub, so the existing guards all passed, and the call threw:

```
TypeError: Notification.requestPermission is not a function
```

**Three call sites had this gap, not one** - the issue named `NewsProvider.tsx`, and `app/arena/page.tsx` and `app/settings/page.tsx` had the identical shape, fixed in the same pass. **Confirmed still live on the current production build** - the vulnerable line is byte-identical on `9f9c88f` (the release the 4 GlitchTip events are tagged to) and today's `main` (`v2026.09.07`), checked directly rather than assumed. Today's deploy did not fix this.

## What changed

Three call sites checked `'Notification' in window` and then called `.requestPermission()` unguarded - fixed to feature-detect the **method**, not just the object:

- `components/NewsProvider.tsx` - the one that actually crashed. Fires in a `useEffect` on mount, unprompted, on **every route including `/`** - a privacy-conscious visitor's very first page load threw with no user action involved.
- `app/arena/page.tsx`'s `enableNotifications` - user-gated (a notification-bell click), but had no `'Notification' in window` check paired with a method check either - same bug shape, would throw on click instead of on mount.
- `app/settings/page.tsx`'s push-subscribe toggle - user-gated, had **no guard at all** before this, wrapped in a try/catch that would swallow the throw into a silent `console.error` with no user feedback.

All three now read: `!('Notification' in window) || typeof Notification.requestPermission !== 'function'`.

## Confirmed from source, not guessed

**Still live on production.** The vulnerable line in `NewsProvider.tsx` is byte-identical on both `9f9c88f` (the release tag all 4 GlitchTip events are attributed to) and `main` (today's `v2026.09.07` deploy) - checked directly with `git show <rev>:components/NewsProvider.tsx`, not inferred. Today's deploy did not incidentally fix this.

**Not touching whether the call belongs on `/` at all.** PM/DevOps flagged that separately as worth a second look - out of scope for this fix, which is the crash only.

## Verification

- [x] `tsc --noEmit` - clean.
- [x] `eslint` on all three touched files - 0 errors, 13 pre-existing warnings, none at the touched lines.
- [x] `npm test` - 785/785.
- [x] `npm run build` - compiles clean (`✓ Compiled successfully`). Took several attempts on this machine's own memory pressure (a browser holding 3+ GB, unrelated to this change) before it ran to completion - the wait, not the code, was the obstacle.

## How to test (QA)

Not reproducible without the extension. Simulate it: in DevTools on the landing page, run `Notification.requestPermission = undefined` before the page's components mount, confirm no throw, no console error. Repeat on `/arena`'s notification-bell click and `/settings`'s push toggle - both should show their existing "unsupported"/error messaging instead of throwing.

## Risk level

**Low.** Purely additive guards - a boolean check gets stricter, no existing passing path changes behavior. Does not touch `app/globals.css`/`labelKeys`/`labelDefaults`/`AppShell.tsx`, so the landing hard gate does not apply, though `NewsProvider.tsx` does mount on `/` - worth a quick landing render check regardless, given it's cheap.

Related: #1042.


JohnDominicJasmin:
QA review. Checked out the branch, ran a dev server, live-simulated the extension's stub exactly as the PR's own "How to test" describes (`addInitScript` overriding `Notification.requestPermission = undefined` before the page mounts, real Playwright run, not DevTools-by-hand).

**Landing page (`NewsProvider.tsx`, the severe one - unprompted, every visitor): confirmed live.** Navigated `/` with the stub in place, zero pageerrors, zero console errors. This is the one that actually mattered most since it fires with no user action.

**Arena bell and settings toggle: guard code verified correct by diff, not independently click-tested.** I spent real time trying to reach the arena notification bell live (it's not on the page at load - traced it into the scanner-panel markup and couldn't get a hover/click sequence to surface it in reasonable time) and didn't attempt the settings toggle at all, since it needs a signed-in session plus a live service worker and I'd already spent the budget on the bell. Rather than keep burning time chasing UI triggers, I read both diffs directly:

- `app/arena/page.tsx`'s `enableNotifications`: `if (!('Notification' in window) || typeof Notification.requestPermission !== 'function')` — identical shape to `NewsProvider.tsx`'s already-live-proven guard, placed before the vulnerable call.
- `app/settings/page.tsx`'s push-subscribe branch: same identical guard, added before `Notification.requestPermission()`, where previously there was no guard at all.

Both match the pattern that's now proven to actually prevent the crash on the one call site I could reach live. Not the same bar as a live click-through, said plainly rather than papered over.

`tsc`/`eslint`/`npm test` claims not independently re-run - no reason to doubt them and the diff is small enough to read directly.

**Merge** - the primary crash path is live-proven, the other two are structurally identical to it and read correct.

QA Team

Issue #1049, opened 2026-09-07
Arena terminal: entitled resolves false for a confirmed Pro session


JohnDominicJasmin:
## Summary

The Arena terminal's mobile timeframe row (and likely the desktop one, same `entitled` prop) shows Pro subscribers as not entitled — the gated "NEED PRO" hint and locked 1M/5M/15M chips render for a confirmed-Pro session, on both a local branch build and deployed staging.

## How found

Investigating a QA finding during a mobile-layout trial: a signed-in real fixture account (pinned `role='pro'`) showed the same gated content as a signed-out session. Ruled out test-methodology error before filing — Dev correctly suspected the simple real-account sign-in might not be the same as the validated entitlement fixture, so I redid it with `qa/e2e/_entitled-session.ts`'s `useEntitledSession(page, { as: 'pro' })`, the same mechanism `entitled-macro-cards.spec.ts` already uses successfully for the dashboard's Pro gate.

**Confirmed with the trusted fixture, not the ambiguous one:**
- `useEntitledSession(page, { as: 'pro' })` + `/arena?design=terminal` → `ArenaTerminal.tsx`'s own debug hook reads `data-entitled="0"`, `data-auth-loading="0"` (not stuck loading — resolved, and resolved wrong).
- Same fixture, same session mechanism, on `/dashboard` (macro cards) → works correctly, gate is gone, real breakdown renders. Ran the existing `entitled-macro-cards.spec.ts` to confirm this before filing — the fixture itself is sound, so the bug is Arena-specific, not a fixture-wide regression.
- **Reproduces on deployed staging**, not just a local branch build (`https://liquidity-hq-staging.onrender.com/arena?design=terminal`, same `data-entitled="0"` result).

## Severity

**Has not reached a real customer.** Checked production (`liquidity-hq.com/api/version`): still on `8f7734f`, which predates `#1039`'s ArenaTerminal merge entirely — the terminal Arena isn't deployed to production yet. This is a must-fix-before-release bug on qa/staging (dev database, test/fixture accounts only), not an active incident.

**Would be paying-user-facing the moment ArenaTerminal reaches production**, since `entitled` is wrong regardless of viewport — a real Pro subscriber would see the free-tier experience (locked Confluence card, no Absorption panel, gated timeframe hint) on their own Arena page.

## What's confirmed vs. not yet found

[...]

PR #1050, opened 2026-09-07
docs(work-queue): #1049 blocks the Arena rebuild reaching main

JohnDominicJasmin:
## Summary

`docs/WORK-QUEUE.md` gains a **D00** row above everything else: **#1049 — a Pro subscriber is treated as free on the rebuilt Arena.**

## Why it goes to the top

Not just the `NEED PRO` hint. **Gated timeframe chips, and `ConfluenceScore` replaced by `LockedFeatureCard`** — a paying customer opens the page they pay for and is told to upgrade. That outranks every layout defect on the board.

**Nobody is affected today, and QA checked that before reporting rather than after.** Production serves `8f7734f`, which predates ArenaTerminal entirely. **It becomes customer-facing the moment the rebuild reaches production**, which is what tonight has been driving toward.

## The row records how it was ruled in, not just that it was

**Dev's hypothesis was reasonable and was ruled out properly.** They proposed that QA's session only *looked* signed in, citing a documented prior incident (#831) of exactly that shape. QA redid the test with `useEntitledSession(page, {as:'pro'})` — the trusted mechanism — and got the same result. Then they proved the fixture itself sound by running `entitled-macro-cards.spec.ts` through it against the dashboard, which passes cleanly. **So it is Arena-specific, not fixture-wide.** Reproduces on deployed staging as well as locally.

Both sessions agree `lib/arenaTimeframes.ts`'s gating is correct — the break is upstream in what `app/arena/page.tsx`'s `useAuth()` resolves `entitled` to on that page. **QA stopped short of root-causing and said so** rather than guessing.

## Sequencing, which the row states explicitly

The CSS branch merges first — it fixes a visible regression on `qa`/`staging`. **Then #1049 as its own fix.** Two defects with two causes in one PR is how #1039 went wrong. **Nothing reaches `main` until #1049 closes.**

## The uncomfortable part, recorded rather than smoothed

#853's sign-off exercised the **signed-out free** path and passed. **The Pro path on the rebuilt page was never exercised** — criterion 5 says all 15 modules render for an entitled user, and that half went unrun. That is the second hole in the criteria after the pixel gap, and it is the more expensive one.

## How to test (QA)

Nothing to run. Check the row names the fixture used, the control spec that proved it sound, the branch that is safe, and the sequencing constraint — and that the sign-off gap is stated as a criteria gap rather than as a QA miss, because that is what it is.

## Risk level

**Low.** Documentation only.


JohnDominicJasmin:
Accurate. Every claim about my own work matches — the fixture-ruling sequence, the dashboard control check, "stopped short of root-causing and said so." Verified the one new claim I hadn't checked myself: criterion 5 (`design-handoff-dir/specs/arena.md:317`) reads exactly "All 15 named modules render for an entitled user" — confirms the framing as a real criteria gap, not something I skipped.

Merge.

QA Team

PR #1051, opened 2026-09-07
fix(arena): chart canvas overflow + mobile timeframe row overflow (two distinct defects, #853)

### LINKED ISSUE #1049: Arena terminal: entitled resolves false for a confirmed Pro session

JohnDominicJasmin:
## Summary

Owner personally found a visible layout regression on the just-merged, just-signed-off Arena terminal rebuild (#853/#1039), hours after two independent sign-offs had passed: the chart canvas painted over every panel below it. While building a new overlap-detection gate to catch this, QA found a second, independent overlap issue on mobile. **This PR fixes both, disclosed as two distinct defects with two distinct causes — not bundled the way #1039 bundled #1020's migration code, which QA only caught by reading the diff.**

## What changed

**Defect 1 — chart canvas overflow (all desktop/tablet widths, `commit 2303c974`)**

`KLineProChart`'s `.klc-canvas` is a self-sizing standalone dashboard-card component (flat `height: 800px`, its own mobile media queries, nothing tying it to an ancestor's height). Nested inside `.at-chart` (which correctly declares `height: 430px`), the canvas ignored that slot entirely and painted ~370-420px of excess straight over the Confluence panel below it — reproduced at every width from 768 to 1440, not a tablet-only gap (disproved PM's initial tablet-gap hypothesis with a 10-width sweep before touching anything).

Fix chains `height: 100%` through `.klc-wrap` down to `.klc-canvas`'s own `flex: 1`. One extra link was needed beyond the obvious rule: `.klc-canvas` isn't `.klc-wrap`'s direct flex child — it sits inside a previously-unnamed positioning wrapper div (`KLineProChart.tsx`) that wasn't itself a flex container, so `flex: 1` had no context to act in. Gave that div a `className` hook (`.klc-canvas-wrap`) purely for CSS, no behavior change, and chained the flex treatment through it.

**`.at-chart`'s `430px` is unchanged** — the fix makes that number finally take effect, not a new number that happens to look right. Scoped to `[data-design="terminal"] .at-chart` specifically, so the dashboard's own standalone `KLineProChart` usage keeps its current self-sizing behavior untouched.

**Defect 2 — mobile timeframe row (390px only, commits `807271f0`, `ae76794b`→`93a07bde`, `3e562be2`)**

Independent cause, found by QA while building `arena-panel-overlap.spec.ts`: 8 timeframe buttons + the gated-state hint span need ~376px+ at min-content width; the mobile row only has 330px available. `flex-shrink` can't rescue it — the buttons are already at min-content. `overflow: hidden` (arena.md's own spec: "40, horizontal overflow hidden") was written before the hint span existed and widened the row — it wasn't a decision to make `1d` unreachable, it was a row that outgrew its budget with nobody re-measuring.

QA's gate flagged this as `.at-tfrow overlaps .at-verdict`. **That specific claim was a false positive** — `getBoundingClientRect` ignores the ancestor's `overflow: hidden` clip, and screenshots at 390 on local dev and deployed staging showed no visible overlap — but it was pointing at a real defect underneath: `1d` and the entire hint span were silently unreachable, un-paintable and un-clickable, the same "hidden nobody notices" shape as defect 1.

Fix: `.at-tfrow` scrolls horizontally instead of clipping (`overflow-x: auto`), with `flex-shrink: 0` on chips/hint so nothing gets squeezed or wraps, a `mask-image` fade signalling more content, and `touch-action: pan-x pan-y` so the row doesn't fight the page's vertical scroll.

**One self-correction disclosed rather than quietly fixed:** the first touch-action value (`pan-x` alone, commit `ae76794b`) turned out to swallow vertical page-scroll gestures starting on the row entirely — confirmed via real CDP touch events (0px scroll vs 185px off the row). Corrected to `pan-x pan-y` (commit `93a07bde`), which computes to the same as the browser default. The fade was also widened (`3e562be2`) after QA's zoomed-screenshot review found the first version (10% edge zone) technically present but perceptually invisible against the near-black terminal palette; widened to 45% so 2-3 chips visibly recede.

**Not in this PR:** #1049 (Pro users resolve as not-entitled on the rebuilt Arena page specifically) was found by QA during this work. Confirmed as a separate defect — `lib/arenaTimeframes.ts`'s entitlement logic is correct and untouched by this branch; the break is upstream in how `app/arena/page.tsx` resolves `useAuth()`'s `entitled` value. Filed and tracked independently; blocks the rebuild reaching `main`, but does not block this PR.

## Why

Owner-reported regression on work that had already passed two independent sign-offs (structural criteria + a general obscured-control detector) — this time verified by looking at real screenshots at each width, not node counts or `getBoundingClientRect()` alone, since those checks are exactly what missed it originally.

## How to test (QA)

On the `qa` environment (or localhost checked out on `qa`), `/arena?design=terminal`:

1. **Desktop/tablet (768-1440px):** Confluence, EMA Ribbon, Order Flow Setup, Multi-TF RSI, Options Gamma, Market Structure, Choppiness Index, and RSI Divergence panels should all render fully below the chart with no chart content painting over them, at every width in that range.
2. **Mobile (390px):** the timeframe chip row (top row with padlocks, not the chart's own internal timeframe control below it) should scroll horizontally — swipe right to reveal 2h/4h/1d and the gated-state hint text (`1M · 5M · 15M NEED PRO`). A visible dimming fade on the row's right edge should hint there's more before swiping. A vertical swipe starting on that row should scroll the page normally, not get stuck.
3. **Automated gate:** `qa/e2e/arena-panel-overlap.spec.ts` — 22/22 passing locally against this branch (rebased onto the post-#1048 clip-aware version of the gate, so the result isn't a false-positive artifact of the earlier gate bug).

## Risk level

**Low.** CSS-only changes (`app/globals.css`) plus one additive `className` hook (`KLineProChart.tsx`, no behavior change). Both fixes scoped with `[data-design="terminal"]` / mobile media queries — the dashboard's standalone `KLineProChart` usage is untouched, confirmed by scope.

**Verified:** lint (0 errors), `tsc --noEmit` (0 errors), unit tests (785/785), QA's clip-aware overlap gate (22/22, both Playwright projects). Compile step of `next build` succeeded clean 3/3 attempts; the full build's internal TypeScript re-check was killed each of those 3 times by shared-machine memory pressure (another session's process), never by an actual build error — noting this honestly rather than claiming a full `next build` pass I couldn't get to complete on this machine tonight. Standalone `tsc --noEmit` already covers the same check that step would have re-run.

**Not independently re-verified by me:** real-device touch feel for the fade affordance (QA confirmed it's a genuine visual improvement via zoomed screenshot but flagged that "does it feel like swipe in the hand" isn't something either of us can certify without a physical device).

This PR does **not** resolve #1049 (see above) — that is next, as its own fix.

Related: #853, #1039, #1047 (QA's overlap gate), #1048 (QA's clip-awareness fix), #1049 (filed, separate).


JohnDominicJasmin:
Independently confirmed, both mechanically and visually.

**Gate**: ran `arena-panel-overlap.spec.ts` fresh against this exact branch, both projects — 22/22, matches the PR's own number. Confirmed the branch carries the post-#1048 clip-aware gate, not the original (checked for `visibleRect`/`clips()` before trusting the count).

**Visual**: screenshot at desktop width — chart correctly boxed, Confluence panel starts cleanly at the chart's boundary, no paint-over.

**Touch-action and fade**: already redriven with real CDP touch events on these exact commits before this PR opened (see the thread with Dev) — vertical scroll 0→185px on and off the row, fade shows genuine dimming at rest. Nothing new to add here, standing by that result.

Everything in the PR body matches what I independently measured. The one thing correctly left open (#1049, entitled resolving false) is correctly not bundled here.

Merge.

QA Team

Issue #925, opened 2026-09-06
LiqFeed's liq_events backfill never fired across two captured loads - may predate all of today's Arena work


JohnDominicJasmin:
**Dev Team** — filed from QA's measurement on #921, because it survives the Arena revert and nothing else is carrying it.

## What was measured

QA captured full network traffic across **two independent fresh loads** of `/arena`, 10s+ settle each:

```
110 requests per load
ZERO requests to the liq_events table       (Supabase REST, either load)
everything else loaded normally             klines, econ snapshot, coinbase price, …
```

`components/LiqFeed.tsx:319-323` issues `sb.from(T.liq_events).select(...)` as its 24h backfill. **It never fired.**

## Why this is not an Arena issue and does not go away with the revert

The measurement was taken on `0444225`, the terminal rebuild, which has since been reverted off every branch. **But `LiqFeed` also mounts in the current-design Arena** — `app/arena/page.tsx:2347`, `headless` — which is what `qa`, `staging` and production all serve today and have served throughout.

So the question the capture raises is not about the rebuild. It is: **does that backfill ever fire anywhere?** If it does not, the clusters ladder has been quietly empty in production, and the terminal-mount work merely made an existing gap newly visible.

## What the source says, read-only

The backfill is gated on one thing:

```
lib/supabase.ts:5-8   getSupabase() returns null ONLY when
                      NEXT_PUBLIC_SUPABASE_URL or ..._ANON_KEY is missing
```

QA's loads made 110 successful requests including other Supabase-backed data, **so the client exists.** If `LiqFeed` mounted, the query should have fired. Zero requests is consistent with the component not mounting at all in the build that was measured.

## What is NOT established, and I am not guessing

- **Whether the sockets opened.** QA's tooling does not capture WebSocket traffic, so `connectBN()` / `connectBB()` are unobserved. The ladder can populate from live sockets without the backfill, so an empty ladder is not proof the backfill is the only problem.
- **Whether this reproduces on the current build.** The capture is from a commit that no longer exists on any branch. **Nobody has re-run it against what is deployed now**, and I have not, because `/arena` is frozen by owner instruction and I am not investigating it.
- **My own #921 claim was wrong about this.** I said mounting `LiqFeed` in the terminal branch would make the ladder fill. QA measured that fix's own deployed commit and it did not. The diagnosis was incomplete and I stated it as though it were finished.

## The one check that settles it

On any environment, with `/arena` open in the current design:

```
[...]

PR #1053, opened 2026-09-07
Revert #1039: restore classic Arena, owner-directed

JohnDominicJasmin:
## Summary

**Owner decision: revert the Arena terminal rebuild.** Owner's words: "the one in staging arena page is wrong and in prod liquidity hq is correct" — after the rebuild (#1039) produced four defects in six hours: the chart-overflow regression, the mobile timeframe-row overflow (both fixed tonight in #1051), and two still-open when the decision landed — #925 (LiqFeed never mounts under terminal mode) and #1049 (a Pro subscriber treated as free on the rebuilt page, the worst of the four). Two fixed, two open, and the two fixes only existed to repair damage the rebuild caused. Production runs the pre-rebuild classic Arena and the owner confirms it's correct.

This PR reverts `7881bbc6` (#1039's merge commit) — the commit that introduced `designMode === 'terminal'` into `app/arena/page.tsx`, confirmed via `git log -S "designMode === 'terminal'" -- app/arena/page.tsx` before touching anything.

## What changed

`git revert -m 1 7881bbc6` — clean, no conflicts. Removes exactly what #1039 added:
- `components/ArenaTerminal.tsx`, `lib/arenaColour.ts`, `lib/arenaEvidence.ts`, `lib/arenaTimeframes.ts`, `lib/useViewport.ts` — deleted.
- `app/arena/page.tsx` — the `designMode === 'terminal'` branch and its `arenaChart` helper (used only by that branch) removed; classic Arena's own render path is untouched, confirmed by diff — #1039 was purely additive to this file (108 lines added, nothing in the classic path modified), so the revert touches nothing the classic page depends on.
- `__tests__/terminalOnlyConstants.test.mts` — removed the `ArenaTerminal.tsx` baseline entry I'd added mechanically in #1039 to unblock the ratchet hook; same mechanical nature, done in reverse. QA: please confirm rather than trust this on my say-so, same as the original entry.

**Not touched, checked rather than assumed:**
- `components/KLineProChart.tsx` — the `className="klc-canvas-wrap"` hook added in #1051 is untouched by this revert (different file region) and remains harmless/additive for the dashboard's shared chart usage — no CSS in the post-revert codebase targets that class outside the now-dead `[data-design="terminal"] .at-chart` scope.
- `app/globals.css`'s terminal-scoped chart/timeframe-row rules from #1051 — untouched. They style `.at-chart`/`.at-tfrow`/etc., classes that existed only in the now-deleted `ArenaTerminal.tsx`. Confirmed no other `[data-design="terminal"]` page/component in the codebase uses the `.at-*` prefix (each terminal-mode screen has its own prefix — `.bt-*` for Briefing, etc.) — these rules are fully inert dead CSS now, not a live risk, safe to leave rather than churn a second PR to strip them.

## Why

Owner-directed revert, not a re-diagnosis. #925 and #1049 are being closed on the issues themselves (as "no-longer-reachable", not "fixed" — the bugs were real, the code that caused them is gone) with the full ruled-out/not-ruled-out state recorded for whoever revisits the rebuild.

## How to test (QA)

On `dev` (not yet on `qa`), `/arena`:

1. **Classic Arena renders** — coin filter chips (All/Majors/Alts/DeFi-AI/Meme), BTC snapshot row, Quick/Deep Research buttons, timeframe row, chart, Strategy panel on the right. Verified locally via screenshot before opening this PR — no `.at-chart`/`.at-tfrow` in the DOM (confirms the terminal component isn't mounting), `.klc-canvas` present (confirms the classic chart renders).
2. **No new console errors.** The only console output on load is a pre-existing, already-documented, unrelated Next.js dev-mode hydration warning about `app/layout.tsx`'s own script placement (same one noted in #1039's original verification) — not introduced by this revert.
3. **Signed-in Pro path** (if you have the fixture handy): confirm Confluence/Absorption render as the classic page's own full versions, not gated — this is the path #1049 broke on the terminal page; worth one confirmation that the classic path never had the problem.

## Risk level

**Low.** A `git revert -m 1` of a purely-additive merge, clean with zero conflicts. Verified by loading the page (screenshot attached to my report, not just reading the diff) rather than trusting the revert mechanically. Gates: lint (0 errors, 256 pre-existing warnings), `tsc --noEmit` (0 errors), unit tests (781/781 — 4 fewer than before #1039, exactly the tests that existed only for the now-deleted lib modules, confirmed no orphaned test files reference deleted modules).

**Not resolved by this PR, by design:** #925 and #1049's actual root causes were never found — this PR removes the code that exhibited them, not the underlying mechanism (which may not exist independently of that code at all, in #925's case, since the bug WAS the unreachable JSX). Both issues closed with the full investigation state recorded for if the rebuild returns.

Related: #1039 (reverted), #853, #925 (closed, no-longer-reachable), #1049 (closed, no-longer-reachable), #1051 (kept — chart/timeframe fixes, now inert on the reverted page but harmless).


JohnDominicJasmin:
Independently verified, three checks:

1. **Ratchet baseline diff** — exact clean removal of the single `'components/ArenaTerminal.tsx': 0` line I originally confirmed, nothing else in the file touched. Matches.
2. **`app/arena/page.tsx` diff** — exact, clean removal of exactly what #1039 added (the two imports, `designMode` hook call, `arenaChart` const, the whole `if (designMode === 'terminal')` block). Classic path below it untouched.
3. **Live check**, not just diff-reading: checked out the branch, real dev server, `/arena`. `.at-chart`/`.at-tfrow`/`.at-root` all zero, `.klc-canvas` present (1), classic page content renders (nav, coin filter row, chart). Only console output: the pre-existing documented script-order hydration warning, nothing new.

Matches the PR's own claims exactly on all three counts I checked myself rather than trusted.

One thing that's mine, not this PR's: `qa/e2e/arena-panel-overlap.spec.ts` will currently report a false "22/22 passing" on the reverted page — `document.querySelector('.at-main')` returns null, my code's early-return gives back an empty array, which reads as "no overlap found" when actually nothing was checked. Fixing that now as its own QA PR so the gate skips honestly instead of measuring nothing. Not blocking this one.

Merge.

QA Team

PR #1054, opened 2026-09-07
docs(arena): production is the UI reference, not this spec

JohnDominicJasmin:
## Summary

A banner at the top of `design-handoff-dir/specs/arena.md` recording that **the terminal Arena design was built three times, rejected twice, and never approved** — and that **production carries the approved UI.**

> **This body has been rewritten.** Its first version described the file's first commit — a harsher framing calling the spec "history, not a target", and listing four defects. **QA caught that the body was stale while the file had moved on.** The file now says *revision, not deletion*, per the owner's correction, and lists **five** defects. Recorded rather than silently replaced, because a PR body that describes a draft nobody merged is exactly the kind of thing that gets trusted later.

## Why

Owner rulings across the evening, in order:

1. *"the one in staging arena page is wrong and in prod liquidity hq is correct"*
2. *"the arena page in prod that should be the UI reference"*
3. *"I mean, adjust the design"* — correcting my first, harsher banner
4. *"the approved UI is now in the production"*
5. *"keep whats in production UI that's the proposed UI and approved one"*

## The history the banner records, because nothing did

| When | What |
|---|---|
| 2026-08-16 03:09 | built (`6942ec94`) |
| 2026-08-27 03:54 | **reverted** (`dd39c9bb`) — 11 days later |
| 2026-09-06 19:36 | rebuilt (`b6a0b403`) |
| 2026-09-06 20:34 | **reverted** (`5621d241`) — **58 minutes later** |
| 2026-09-07 21:09 | rebuilt (`97f07277`, #853 / PR #1039) |
| 2026-09-08 | reverted (#1053) |

**Neither revert states a reason.** Both are bare `This reverts commit …`. So the ticket read as *unfinished work* rather than *a declined change*, and was picked up in good faith twice — most recently by me. **An open ticket outlives an unrecorded "no".**

## What the third attempt cost, in six hours

1. the chart canvas painting over every panel, **at every width from 768 to 1440**
2. `1d` unreachable on mobile, with the `NEED PRO` hint explaining the gated chips
3. `LiqFeed` never mounting — Liquidation Clusters **permanently** empty (#925)
4. **a Pro subscriber resolving as free** (#1049) — a paying customer shown an upgrade prompt on the page they pay for
5. **the on-the-fly strategy switch gone entirely** — the owner found this one themselves, and **the spec's 15-module inventory never listed the Strategy Panel at all**, so the omission is in the design rather than the build

## Kept deliberately

- **"Colour is data"** — a value is coloured because that signal *fired*; direction is not the sign of the number.
- **The entitlement check belongs at the call site**, not inside the component — #1049 is that rule's own family.

## How to test (QA)

Nothing to run. Open the file: the first readable thing should be that this design was rejected twice, with dates and commits, and that production carries the approved UI. **Check the banner reads as "revise against production", not "this is dead"** — that distinction is the owner's correction and the reason this body was rewritten.

## Risk level

**Low.** Documentation only.


JohnDominicJasmin:
The committed file is good — but the PR's own body describes an earlier draft, not what's actually here. Two commits: `13101be` (the "history, not a target" version) and `6193c4e` (correcting that to "revision, not deletion" per the owner's later "I mean, adjust the design"). The PR body still only describes `13101be` — undercounts the defects at four when the committed banner correctly lists five (Strategy Panel's omission is defect 5, the owner's own catch, not caught by Dev or me), and frames the file as dead history when the actual banner explicitly says the opposite and corrects itself for having said that once already.

Read the actual banner, not the PR description: it's internally consistent, names all five defects accurately against what I independently verified tonight (canvas overflow, mobile 1d/hint, #925, #1049 — all four match my own findings exactly), correctly attributes the fifth to the owner directly, and the "carry forward" section (colour-is-data, entitlement-at-call-site) and the TEST_GAPS §2 callback are both accurate.

**Ask**: update the PR body to match `6193c4e`'s actual framing and defect count before merging — not because the content is wrong, only because the description of it is stale. The file itself needs no changes from me.

QA Team

Issue #1059, opened 2026-09-07
Binance is a single point of failure for price data — fail over to Bybit, which is already integrated


JohnDominicJasmin:
**Owner request, 2026-09-08:** *"we must find alternative for binance"* — raised after Binance rate-limited the `staging` egress IP (418) and most of the Arena's coins lost their price data. See **#1058**, and **#368** for the same mechanism in July.

## Start here: Bybit is already integrated, and is already the default

`app/api/market/klines/route.ts` supports **three** sources — `bybit`, `binance`, `binance-futures` — and `bybit` is the **default** when no `source` is given (`:152`). Several other market routes already run on Bybit exclusively: `account-ratio`, `agg-trades`, `funding-rate`, all via `lib/bybitFanout.ts`.

**So this is not "integrate an alternative". It is "use the one we already have when the primary fails."**

## What actually pins us to Binance

Five call sites request `binance-futures` explicitly and have **no failover**:

| Where | What it feeds |
|---|---|
| `lib/useEMAStrategy.ts:96` | the EMA Ribbon signal — **this is what the owner saw fail** |
| `components/KLineProChart.tsx:1559` | chart history |
| `components/HigherTfMoveBadge.tsx:41` | the higher-timeframe badge |
| `lib/usePerpSpot.ts:63` | perp-vs-spot, which needs **both** sources by definition |
| `lib/backtestEngine.ts:53` | backtests |

When Binance returns 418, each of these fails independently and silently.

## Why the spot check said everything was fine

**BTC goes through `source=binance` (spot), which was NOT rate-limited. Every other coin's main data goes through `binance-futures`, which was.** Dev and I both checked BTC and both saw a healthy page. **Two people checked the one coin that could not show the symptom.**

## The work, in order of value

1. **Failover on 418/429: retry the same request against Bybit.** The route already speaks Bybit and `fetchOHLCV()` already falls through to `BYBIT_SYMS` when a coin is absent from `BINANCE_SYMS`. This is the cheapest large win and it fixes the owner's actual complaint.
2. **Log the real upstream error before `String(err)`** — `lib/useEMAStrategy.ts:757` and `:820`. Tracked on #1058. **Binance told us exactly what was wrong in plain text and we discarded it**, which is why a documented, previously-seen condition cost two sessions an hour.
3. **Only then evaluate a third source** (OKX, Coinbase, Kraken). Adding a vendor before using the one already wired is the expensive way round.

## Three things that must not be glossed over

[...]

Issue #1058, opened 2026-09-07
EMA Ribbon Strategy shows "Signal unavailable right now" — a real exception, swallowed into an opaque string


JohnDominicJasmin:
**Owner-reported, on `staging` (`4338141`).** The EMA Ribbon Strategy panel showed **"Signal unavailable right now"**.

## It is the failure branch, not the graceful one

The obvious hope was that this is #1026's fallback working — the fix that replaced raw exception text with translated labels. **It is not.**

- `EMA_SIGNAL_INSUFFICIENT_DATA` reads **"Not enough candle history yet for this coin/timeframe"** (`lib/labelDefaults.en.json:1471`)
- `EMA_SIGNAL_FAILED` reads **"Signal unavailable right now"** (`lib/labelDefaults.en.json:1470`)

`EMASignal.tsx:363` picks between them on `signal.error !== 'INSUFFICIENT_DATA'`. **The owner hit the other branch — a real thrown exception.**

## Why we cannot say what the exception was

`lib/useEMAStrategy.ts` has two catch blocks — **:755 and :818** — and both do:

    error: String(err)

**The real error is stringified into an opaque value the UI cannot distinguish from any other failure**, and nothing logs `err` before that happens. So the panel can say "something went wrong" and nobody can say what.

## Not reproduced

Checked `staging`, BTC 1h, signed in: the panel renders fully — nine-condition checklist, fast/mid/slow averages, daily trend, recent record, verdict **MIXED / FREEZE**. **The owner did not say which coin or timeframe they were on**, and that is the missing input.

**Also relevant:** #1024 was closed by source inspection precisely because the error state could not be reproduced — *"all 8 coins have full history right now"*. So neither branch of that ternary had ever been observed rendering until the owner hit this one.

## Where to start

1. **Log `err` before stringifying it**, at both catch sites. That alone turns the next occurrence from "FAILED" into something diagnosable, and it is worth doing even if the underlying bug is never reproduced.
2. **Then reproduce** — needs the owner's coin, timeframe and account state.

## What this is not

**Not caused by the Arena revert.** `EMASignal` and `useEMAStrategy` are untouched by #1053, and the panel is in the classic Arena that production also runs. **Production is on `8f7734f` and carries the same code**, so this is reachable there too — nobody has reported it, but nothing rules it out either.

Filed at the owner's report rather than added to a parent issue, because it is a distinct defect with its own cause and it outlives every branch currently in flight.

[...]

PR #1060, opened 2026-09-07
fix(chart): Bybit primary klines source, Binance fallback (#1059)

JohnDominicJasmin:
## Summary

Owner: *"Let's not use Binance anymore. find alternatives"* / *"we have to nail this problem"* — recurring since July. #1059: Binance futures is rate-limiting/blocking this server's egress IP (measured directly against `staging`: 502, HTTP 418), while Bybit answers normally from the same IP, including for two coins Binance doesn't even list (1000PEPE, 1000BONK). This PR flips the default at three of the four originally-scoped klines call sites from Binance-primary to Bybit-primary.

## What changed — which exchange each surface now reads from

| Surface | Before | After |
|---|---|---|
| `lib/useEMAStrategy.ts` (Arena EMA Ribbon signal) | Binance futures primary, Bybit fallback | **Bybit primary**, Binance futures fallback |
| `components/HigherTfMoveBadge.tsx` (higher-TF move warning) | Binance futures primary, Bybit fallback | **Bybit primary**, Binance futures fallback |
| `lib/backtestEngine.ts` (`/backtest` historical simulation) | Binance futures primary, Bybit fallback | **Bybit primary**, Binance futures fallback |
| `components/KLineProChart.tsx` (Arena/dashboard chart) | Binance spot → Binance futures → Bybit last-resort | **Unchanged this PR** — see below |

`bybitPriceFactor()`/`bybitSymbolPriceFactor()` applied at every Bybit read per `lib/coins.ts`'s own rule ("any new code reading a price from BYBIT_SYMS must apply this — the failure is silent and only shows up on two coins").

## Why `KLineProChart.tsx` is untouched, deliberately

It was one of the five originally-scoped call sites. Its REST history fetch (`getBars`) and its live WebSocket subscription (`subscribeBar`) **must read from the same exchange**, or the chart's feed silently changes mid-session — that's the exact failure #359 fixed, by promoting Binance spot to lead both paths. `subscribeBar` connects browser-side straight to `wss://stream.binance.com`; moving that is WebSocket work, explicitly out of scope tonight (LiqFeed already handles both Binance and Bybit liquidation streams separately).

**Flipping only `getBars` would make #359's failure mode the permanent default** — history from one exchange, live ticks from another, on every load, not just during an outage. The file's existing `spot → futures → Bybit-last-resort` chain in `getBars` is untouched and still a real fallback during a Binance outage. Migrating this file to Bybit-primary is its own future change that has to move `getBars` and `subscribeBar` together — not a piece of this one. **Whoever picks that up next: do not flip `getBars` alone to "finish the job" — that recreates #359 quietly.**

## A second, unrelated, pre-existing bug found along the way

**`lib/backtestEngine.ts`'s Bybit range-fetch had no price-factor conversion at all, before this PR.** `pepe`/`bonk` are Bybit-only (no Binance symbol — see `lib/coins.ts`'s own comment: "kept Bybit-only to avoid display confusion"), so **every backtest ever run on either coin was silently simulating trades at 1000x the real price.** This predates tonight's change entirely and is unrelated to the primary/fallback flip — found only because the price-factor rule was applied everywhere it says to apply it, not just everywhere it looked necessary. A backtest producing a plausible-looking return at a thousand times the real price is exactly the kind of failure that never gets reported; nobody eyeballs a percentage and thinks "wrong order of magnitude."

## Also in this PR, per explicit instruction to land it here rather than separately

**#1058's fix.** `lib/useEMAStrategy.ts`'s two catch blocks now `console.error` the real thrown error before collapsing it to the opaque `String(err)` the UI shows as "Signal unavailable right now". That string was a real exception, not the graceful `INSUFFICIENT_DATA` fallback (confirmed by reading `lib/labelDefaults.en.json`), and was undiagnosable before this because the actual cause never reached anywhere visible. Three lines, disclosed here rather than bundled silently.

## Verification

Live, not just the diff: `1000PEPEUSDT`'s raw Bybit kline close (`0.003626`) renders on the local Arena EMA Ribbon panel as **$0.00000362** — the correct real-PEPE scale, not the 1000x-wrong ~$0.0036 a missing factor would produce. Screenshot-equivalent confirmed via DOM text extraction (`0.00000362`, `0.00000361`, `0.00000331`, `0.00000363` — all in the right range across the panel's several price fields).

Gates: lint (0 errors), `tsc --noEmit` (0 errors), unit tests (781/781).

## How to test (QA)

**Test on a `1000`-prefixed coin explicitly — PEPE or BONK, not BTC.** BTC has no prefix, no factor, and isn't even rate-limited on Binance spot; checking BTC alone is exactly how #1058 hid from two people for an hour tonight.

1. Arena, PEPE or BONK selected: EMA Ribbon Strategy panel's entry/stop/target should read roughly **$0.000003–0.000004** for PEPE (BONK similar order of magnitude), not $0.003–0.004. **Compare against a reference source (CoinGecko/CoinMarketCap)**, not just "a number appears."
2. `/backtest`, same coin: simulated trade prices should be in the same real-per-token range as above.
3. HigherTfMoveBadge (any coin, 1m/5m/15m/30m timeframe, only shows on a >6% 4h move): no visible change expected, purely a source-preference change on a ratio calculation.
4. EMA Ribbon on any coin that previously showed "Signal unavailable right now": open the browser console — the real error should now be logged there (via `[useEMAStrategy]`), where before nothing was.

## Risk level

**Medium.** Touches the primary price-data path for three live surfaces (Arena signal, backtest, a chart badge) — this is exactly the class of change PM flagged as "the wrong thing to hurry," so verification leaned on live checks against real data rather than trusting the diff. Not medium-to-high because: the fallback chain is preserved everywhere (a coin missing from Bybit still resolves via Binance), the price-factor application follows an existing, already-battle-tested helper (`bybitPriceFactor`, already used by `MarketProvider.tsx` and the alert-outcomes resolver), and `KLineProChart.tsx` — the highest-traffic, most complex surface — was deliberately left alone rather than risk a partial migration.

Related: #1059, #1058 (logging fix included here), #359 (the reason `KLineProChart.tsx` is untouched).


JohnDominicJasmin:
Independently verified against a real external reference, not against the ratio looking sensible.

## The check that matters — PEPE, against CoinGecko

Pulled a live reference before touching the app at all: `api.coingecko.com/api/v3/simple/price?ids=pepe&vs_currencies=usd` → **$0.00000363**.

Checked out the branch, real dev server, `/arena?coin=pepe&tf=1h`. Confirmed request goes to `source=bybit&symbol=1000PEPEUSDT` (Bybit primary, no Binance fallback triggered — Bybit is healthy). Panel's price fields: **$0.00000362, $0.00000361, $0.00000361, $0.00000331, $0.00000362** — all in the correct real-PEPE range, not the ~1000x-wrong ~$0.0036 a missing factor would produce.

Also pulled the raw upstream data directly (`curl .../api/market/klines?source=bybit&symbol=1000PEPEUSDT...`) and did the division by hand: raw close `0.003618` ÷ 1000 = `0.00000362` — matches both the rendered panel and the CoinGecko reference. Not trusting the app's own math; reproduced it independently.

## `backtestEngine.ts`'s pre-existing bug fix

Didn't wait through a full multi-coin batch run (the backtest page has no single-coin URL param — it's an all-coins sweep, `/backtest?coin=pepe` is silently ignored, confirmed by reading `app/backtest/page.tsx`). Instead verified at the code level: `bybitSymbolPriceFactor(sym)` is the identical helper already proven correct above, applied to the identical `+k[1]*pf, +k[2]*pf, +k[3]*pf, +k[4]*pf` shape as `useEMAStrategy.ts`'s already-externally-verified transformation. Same function, same data shape, same real symbols — not a new, unproven code path.

## Everything else, checked directly rather than assumed

- **`components/HigherTfMoveBadge.tsx`** diff: Bybit branch moved first, `bybitSymbolPriceFactor` applied even though it's a no-op for a same-exchange ratio — matches the PR's own honest framing ("applied per the rule, not because it matters here").
- **`lib/usePerpSpot.ts`**: zero-line diff (`git diff` confirms), zero mentions of Bybit anywhere in the file. Stays Binance-only, degrades rather than silently comparing two exchanges. Confirmed, not assumed.
- **`components/KLineProChart.tsx`**: zero-line diff, confirmed via the same check. The #359 REST/WebSocket-coupling reasoning for leaving it alone holds.
- **#1058's fix**: both catch sites in `useEMAStrategy.ts` now `console.error` before stringifying. Read directly.
- **Gates**: reran lint/tsc myself on the three changed files — 0 errors (7 pre-existing warnings, none at touched lines), tsc clean.

## Honest labels

The PR body's table naming which exchange each surface reads from, before/after, is exactly the disclosure this needed — confirmed it matches the actual diffs, not just asserted.

**Ready.** Merge, and I'll do the qa→staging chain right after.

QA Team

PR #1063, opened 2026-09-08
fix(arena): close two production layout gaps (owner-directed)

JohnDominicJasmin:
## Summary

Two owner-directed layout requests on the **approved production Arena** — not a return to the reverted terminal rebuild. Both are small, disclosed as two distinct defects with two distinct causes per two separate commits, same PR since both are small.

## What changed — defect 1: news-ticker gap (`f4fcc2f9`)

Owner reported on production (screenshots, liquidity-hq.com): a ~34px empty band between the nav bar and the news ticker on every page.

**Root cause, measured live on prod, not assumed:** `document.documentElement.getAttribute('data-design')` is `"terminal"` on production. `[data-design="terminal"] { --appbar-h: 44px; --strip-h: 34px; }` — the 34px `--strip-h` reserves space for `PriceTickerStrip`. `git log -S PriceTickerStrip` shows its JSX mount was removed in `b2bba511` ("revert(dashboard): drop the canvas-mirror rebuild") and nothing has re-added it since — `grep -r PriceTickerStrip components/ app/` finds zero matches in the current tree. The mobile media query already sets `--strip-h: 0px` for the identical reason ("so the offsets do not reserve space for something that is not there", #628) — that reasoning was just never carried to desktop.

**Fix:** `--strip-h: 34px` → `0px` on desktop, matching mobile. One variable, five consumers (`.app-content` padding, `.nav-drawer`, `.news-ticker`, `.breaking-alert` ×2) all correct simultaneously since they all read the same custom property.

**Landing hard gate** (this touches `app/globals.css`): swept `/` at 1440 across all 4 theme/design combinations (light/dark × current/terminal). `current` mode is untouched by construction — confirmed `data-design` reads `null` for that mode, so `[data-design="terminal"]` can't match. Both terminal combinations render clean, no gap, landing's own separate ticker (`LandingTicker.tsx`) unaffected.

## What changed — defect 2: Multi-Timeframe Alignment gap (`bfa99b02`)

Owner reported with screenshots: a large empty region below the EMA Ribbon Strategy card, wanted Multi-Timeframe Alignment moved into it.

**Root cause:** `.arena-ws` is a two-column grid (chart left, 320px rail right) with `align-items: start` — columns don't stretch to match. The rail (Strategy panel, market snapshot, Confluence Score) runs longer than the left column (chart + EMA Ribbon card only), leaving the left column's bottom half empty.

**Fix:** one-line move — `<MultiTFAlignment coin={selectedCoin} />` from the full-width section below the workspace to directly after `<EMASignal .../>` inside `.arena-ws-chart`. Grid, rail, and column widths untouched.

## Verification — and an honest gap in it

Both fixes verified live (screenshots, not just the diff) in the **reachable local state**: signed-out. For defect 1, mechanically confirmed via computed style (`--strip-h` resolves to `0px`, `.app-content`'s padding-top computes to exactly 76px matching the corrected formula). For defect 2, confirmed the move doesn't break anything, right rail unaffected, mobile order reads sensibly (chart → EMA Ribbon → Multi-Timeframe → Strategy → Market Snapshot → Confluence).

**What I could not verify locally:** the exact owner-reported MTF gap needs a signed-in Pro session — Confluence Score's full breakdown (vs. free's short "Unlock with Pro" placeholder) is what makes the rail long enough to overflow the left column in the first place. Local dev's `useEntitledSession`-style fixture doesn't reliably simulate Pro auth here (same friction #1049's investigation hit earlier tonight). So: the structural fix is correct by construction (MTF now shares EMASignal's column, always renders directly after it regardless of rail height) — but whether it fully closes the gap or just shrinks it for a real Pro user is **not confirmed by me**. Saying so rather than declaring it fixed.

## How to test (QA)

**Signed in as Pro** (or on deployed staging with a real Pro session, where the fixture path already works):
1. `/arena` at 1440px — Multi-Timeframe Alignment should render directly under EMA Ribbon Strategy in the left column. Check whether the empty gap next to the Confluence Score rail is fully gone or just smaller — if a gap remains, that's useful to know, not a failure of this PR to report.
2. `/arena` at 390px — order should read chart → EMA Ribbon → Multi-Timeframe → Strategy → Market Snapshot → Confluence.
3. Any app page with an active news headline (e.g. `/dashboard`) — no empty band between nav bar and news ticker, at both 1440 and 390.
4. Landing page `/` — unchanged in both light and dark theme.

## Risk level

**Low.** Defect 1 is a single CSS custom-property value change, mechanically verified, landing-gate swept. Defect 2 is a one-line JSX move with no structural changes to the grid or rail. Both changes are scoped and reversible. One caveat stated above: defect 2's exact visual outcome for a signed-in Pro user needs QA's confirmation, not re-derived by dev due to a known local-fixture limitation.

Gates: lint (0 errors, 256 pre-existing warnings), `tsc --noEmit` (0 errors), unit tests (passing).

Related: owner screenshots (relayed via PM/DevOps), #628 (mobile `--strip-h` precedent), #1049 (local Pro-fixture friction, referenced not reopened).


JohnDominicJasmin:
**QA Team.** Reviewed, independently verified both root causes, merge it.

**Defect 1 (news-ticker gap):** confirmed `grep -rn "PriceTickerStrip" app/ components/` finds zero JSX mounts anywhere in the tree - only comments, and `LandingTicker.tsx` says outright "that component no longer exists." `b2bba511` is a real commit ("revert(dashboard): drop the canvas-mirror rebuild"). The 5 `--strip-h` consumers listed in the PR body match what's actually in `globals.css`. `--strip-h: 34px` → `0px` is correct and matches the existing mobile precedent (#628) exactly.

**Defect 2 (MTF gap):** diff is a clean single-component move, one location to another, both comment blocks updated to point at each other. No grid/rail changes. Low risk as claimed.

**On the disclosed gap:** can't check the signed-in-Pro rendered outcome yet since this is only a PR against `dev` - nothing to view on deployed staging until it's promoted. Once `dev`→`qa`→`staging` lands I'll check `/arena` at 1440 signed in as Pro and report whether the gap fully closes or just shrinks, per Dev's ask.

Merge whenever.

JohnDominicJasmin:
**QA Team.** Signed-in-Pro check completed on staging (3c7cde5), /arena at 1440 - full result posted on #1049 (same fixture-friction context). Summary here since this is the PR it's actually about:

**Gap did not fully close - it moved.** MTF now correctly fills the space directly under EMA Ribbon (the originally reported empty region is gone). But since `.arena-ws-chart` is now the taller column and `align-items: start` still doesn't stretch the rail to match, there's a new 364px gap on the rail side instead (measured via `getBoundingClientRect()`: chart column bottom y=2079, rail bottom y=1715).

Not a failure of this fix - it did exactly what it said (MTF now shares the chart column, always renders after EMA Ribbon regardless of rail height) and closed the specific reported symptom. The root shape (two non-stretching columns of different height) is still there, just shorter by however much MTF's height offset it. Flagging as a measured result, not asking for a re-open - owner's call whether a 364px rail-side gap needs its own pass.

JohnDominicJasmin:
**PM/DevOps ruling: the relocated gap is closed as a design call, not a follow-up.**

QA's measurement after the move: MTF now fills the original empty spot under EMA Ribbon Strategy as the owner asked, confirmed on a genuinely-Pro staging session. But the chart column is now the taller of the two, so `align-items: start` on `.arena-ws` leaves 364px at the tail of the 320px rail instead.

**This is the grid shape, not a defect in the move.** The shorter column ends early by definition, so relocating a panel relocates the empty space — it cannot remove it. Zero gap requires both columns to hold identical content height, which live data will never guarantee. There is no "another pass" with a better ending; there is only a choice about which side carries the leftover space.

Recorded so the next reader does not file this as a regression from the fix:

- The owner's literal request — put Multi-Timeframe Alignment in the empty space under EMA Ribbon — **is satisfied**.
- Put to the owner as a design decision with a recommendation to keep it as-is: an empty tail beside a narrow right rail reads as layout; an empty hole mid-page in the main column reads as broken.
- If the owner wants the rail side closed as well, that is **equalising the two columns**, a larger piece of work with its own decision. Not a tweak, and not to be pre-built.

Cross-referenced on #1049.

JohnDominicJasmin:
**PM/DevOps ruling on the relocated gap (relayed by QA): no follow-up, treat as closed.**

Reason, on record: `align-items: start` on `.arena-ws` means the shorter column ends early - moving a panel between columns moves the empty space, it doesn't remove it. Before this fix the gap sat mid-page in the main column, under EMA Ribbon. After it, the main column is the taller one and the 364px sits at the tail of the 320px rail instead. There is no arrangement of the existing panels that produces zero gap for arbitrary live content - only a choice of which side carries it.

Recorded here explicitly so the next person to look at this doesn't file the rail-side gap as a regression from this PR - it's the same unavoidable shape the fix was always going to produce, just relocated.

JohnDominicJasmin:
**Owner confirmed, 2026-09-08: leave it.** Their words — *"leave it i agree"*.

So the recommendation in the ruling above is now a decision. The 364px on the rail side is **accepted layout, not an open defect.** Do not reopen it, do not file it as a regression, and do not spend time equalising the columns unless the owner asks for that specifically on a later day.

Recorded here rather than left in chat because the last time an Arena design decision lived only in a conversation, an open ticket outlived an unrecorded answer and the same rejected work got rebuilt twice — see the history banner at the top of `design-handoff-dir/specs/arena.md`. One line costs nothing; not writing it cost six hours and five defects.

PR #1066, opened 2026-09-08
fix(chart): portal coin-select popover past GrokChat's transform

JohnDominicJasmin:
## Summary

Owner-reported: the coin dropdown inside the LiquidityAI chat panel (GrokChat) "is not showing or working well" — screenshot showed `BTC ▾` in the panel with no usable popover. Reproduced on production before touching anything: the popover's rect landed at `left:1816` while the chat panel only spans `left:900-1260` — entirely off the right edge of the viewport.

## Root cause

`.gchat-panel` keeps `transform: scale(1) translateY(0)` even at rest when open — visually identity (confirmed `matrix(1,0,0,1,0,0)` via `getComputedStyle`), but per spec a non-`none` transform still makes an element a **containing block** for `position: fixed` descendants. `CoinMultiSelect`'s `.cms-panel` is `position: fixed` with JS-computed viewport coordinates (`positionPanel()`) — those coordinates were resolving against the 360×560 chat panel instead of the real viewport, and `.gchat-panel`'s `overflow: hidden` then clipped whatever fell outside that box on top of the misplacement. `z-index: 9995` vs the panel's `9994` was a red herring — the transform creates a new stacking context regardless of z-index value.

**Not a new bug class in this codebase.** `Tip.tsx` already solved this exact problem for tooltips, with its own comment documenting the mechanism generically ("any ancestor with a transform/filter/perspective... creates a new containing block and silently repositions a nested fixed element"). This PR applies the identical, already-established fix to the other `position: fixed` component that had never received it.

## What changed

`CoinMultiSelect.tsx`: `.cms-panel` now renders via `createPortal(..., document.body)` instead of inline — matches `Tip.tsx`'s exact pattern.

## Other mount points checked

`CoinMultiSelect` is used in 3 places. Checked the other two:
- `app/settings/page.tsx` (Watchlist section) — plain page flow, `.st-section` has no transform. Not currently broken.
- `app/alerts/page.tsx` (muted coins) — same, plain page flow. Not currently broken.

The portal fix is unconditional and correct regardless of ancestor, so this also makes those two mounts immune to the same class of bug if either is ever wrapped in a transformed ancestor later.

## Verification

Reproduced broken on **production** first (screenshot + measured rect: popover at `left:1816` against a panel spanning to `1260`), then confirmed fixed locally in all three chat-panel states:
- **Default open** — popover correctly at `left:915`, within the panel's own `900-1260` range.
- **Expanded/pop-out mode** (`.gchat-expanded`, its own `translate(-50%,-50%)` transform) — popover still lands correctly inside the 830px-wide expanded panel.
- **Mobile 390** — popover at `left:27, right:363`, fully within viewport.

Screenshots for all three states attached to my verification notes (production-broken, local-fixed-default, local-fixed-expanded, local-fixed-mobile).

## How to test (QA)

On any page with the LiquidityAI chat FAB (e.g. `/dashboard`):
1. Open the chat, click the coin trigger (`BTC ▾`) — the coin list popover should appear directly below the trigger, fully visible, with search box and radio list.
2. Expand the chat to pop-out mode, repeat — popover should still land correctly inside the larger centered panel.
3. At 390px width, repeat — popover should stay within the viewport.
4. Sanity-check the two other mount points still work: Settings → Watchlist, and Alerts → muted coins — should be unaffected (they were never broken, portal doesn't change their behavior).

## Risk level

**Low.** Single, well-scoped component change following an existing, already-verified pattern from the same codebase (`Tip.tsx`). No CSS changes, no changes to positioning math (`positionPanel()` untouched) — only *where* the DOM node renders. Reversible by removing the portal wrapper.

Related: owner screenshot (relayed via PM/DevOps), `Tip.tsx` (prior art for this exact fix).


JohnDominicJasmin:
**QA Team.** Reviewed, root cause independently verified, merge it.

Confirmed directly against source:
- `.gchat-panel.gchat-open { transform: scale(1) translateY(0); ... }` - real, matches the claim exactly.
- `.cms-panel { position: fixed; ... }` - confirmed.
- `Tip.tsx` already uses `createPortal(..., document.body)` with the identical containing-block rationale in its own comment - this PR is applying established prior art, not inventing a new pattern.
- Bonus: `.gchat-panel.gchat-open.gchat-expanded { transform: none; }` - expanded mode never actually had the bug, but the portal fix is unconditional so no regression there either, matches dev's own "verified expanded too" note.

Diff is minimal (portal wrapper only, `positionPanel()` math untouched), well-scoped, low risk as claimed. Merge whenever.

JohnDominicJasmin:
**QA Team.** Live-verified on qa (84e78f9). All good, no regressions.

**GrokChat mount** (`/dashboard`): popover found via `getBoundingClientRect()` at `parentIsBody: true` (confirmed portaled to `document.body`), rect fully within viewport - matches dev's local numbers in shape.

**Portal regression checks** (the classic failure modes for a `createPortal` change):
- **Outside-click close** - clicked outside, `.cms-panel` removed from DOM immediately. Works.
- **Scroll behavior** - scrolled the page 300px while open, panel position unchanged. Correct, not a bug: `.gchat-panel` is itself viewport-fixed, so its trigger never moves on scroll - `positionPanel()`'s existing scroll listener had nothing to recompute. Verified this is expected by reading the component, not just observing it.
- **Survives chat-panel-close** - closed the chat panel while the coin popover was still open. Both `.gchat-panel.gchat-open` and `.cms-panel` cleared from the DOM together. No orphaned portal left behind.

**Other two mounts, independently re-verified** (signed in for real, not just trusting "not currently broken"):
- Settings → My Watchlist: popover portaled to body, within viewport.
- Alerts → muted coins: same.

Ready for staging.

PR #1068, opened 2026-09-08
docs: CLAUDE.md said the release PR opens itself; it has not since 2026-08-09

JohnDominicJasmin:
## Summary

`CLAUDE.md` told every session that the `staging` → `main` release PR opens itself on any push to `staging`. It does not, and has not for a month. This corrects the paragraph.

## What changed

One paragraph in `CLAUDE.md`. It now states that the release PR is opened **by hand**, names the variable and the date, explains why the pause is deliberate, and points at the same fix `CONTRIBUTING.md` already gives. No workflow, no variable, no behaviour is touched.

## Why

The job exists and is real, but it is gated:

```yaml
release-pr:
  if: github.ref == 'refs/heads/staging' && vars.RELEASE_PR_PAUSED != '1'
```

`gh variable list` shows `RELEASE_PR_PAUSED = 1`, set **2026-08-09** and never unset. `gh run list` shows *Release signals* completing as **skipped** on every push to `staging` — four times on 2026-09-08 alone.

**`CONTRIBUTING.md` has been correct since 2026-09-05**, carrying the `if:` condition, the verification, and the manual fix. `CLAUDE.md` did not — and `CLAUDE.md` is the file a session reads first, so where the two disagreed the contradiction resolved toward the wrong one. A session that believes the PR opens itself promotes to `staging` and then waits for an aggregation point that never appears. That is exactly the *"deploying into silence"* the rule was written to prevent.

**The pause itself is correct and stays.** A release PR's head IS its base branch, so every promotion fires `synchronize` and re-runs the full browser suite — billing a gate run per push for a release nobody intended to ship yet. The defect was the documentation, not the switch.

Found while answering the automated check on #1062, which fired because 63 commits were parked on `staging` with no release PR. That check is the compensating control for this pause, and it works — but it tells you the PR is missing, not that the automation was never going to open one.

## How to test (QA)

No application change — nothing to exercise on any environment.

1. Read the release-PR paragraph in `CLAUDE.md`. It should say the PR is opened by hand, give the variable name and the 2026-08-09 date, and not contradict `CONTRIBUTING.md`.
2. Confirm the two files now agree: `grep -n "opens itself" CLAUDE.md CONTRIBUTING.md`.
3. Confirm nothing changed operationally — `gh variable list` still shows `RELEASE_PR_PAUSED = 1`, and `.github/workflows/release-signals.yml` is untouched in the diff.

## Risk level

[...]

PR #1069, opened 2026-09-08
Release: 27 PRs — Arena restored to the approved UI, Binance dropped as primary, three user-facing fixes

JohnDominicJasmin:
## Summary

Release candidate: **70 commits, 27 PRs**, `main`..`staging`. Deployed and serving on staging at `84e78f9` (`/api/version`, verified by QA and by the version monitor independently).

**Opened by hand.** `release-signals.yml` did not open it — `RELEASE_PR_PAUSED = 1` since 2026-08-09. See #1068, which corrects `CLAUDE.md` for claiming otherwise, and #1062, the check that caught the gap.

## What changed

**Owner-visible fixes**
- **#1066 — the chatbot coin dropdown was unusable.** The popover rendered around x=1816 against a chat panel ending at x=1260, so it was off-screen entirely. A transformed ancestor made itself the containing block for the fixed-position panel. Portalled to `document.body`, matching the pattern `Tip.tsx` already used for the same bug class. **Live-broken on production right now.**
- **#1044 — push notification permission died silently.** `Notification.requestPermission` is not a function when an anti-fingerprinting extension stubs the object; `'Notification' in window` is still true, so the guard never fired. Three call sites, all byte-identical on today's production build. **Also live-broken on production.**
- **Nav-bar gap.** `--strip-h: 34px` was still reserving space for `PriceTickerStrip`, whose mount was removed on 2026-09-03 and whose file no longer exists. The reservation outlived the component.
- **#1063 — Multi-Timeframe Alignment moved** into the empty column under EMA Ribbon Strategy, at the owner's request. The leftover space relocates to the rail; **the owner reviewed that and accepted it** (2026-09-08, *"leave it i agree"*). Not an open defect.

**Arena restored to the approved UI**
- **#1053 reverts the terminal rebuild.** It produced five defects in six hours: chart canvas overflow at every width 768–1440, `1d` unreachable on mobile, `LiqFeed` never mounting (#925), **a Pro subscriber resolving as free** (#1049), and the on-the-fly strategy switch gone entirely. Production's Arena is the approved reference — #1054 records the full build/reject history so this does not return a fourth time.

**Binance dropped as the primary price source**
- **#1060** flips three call sites to Bybit, tested from staging's own egress (Bybit OK; `binance-futures` returns 418 from that IP). `KLineProChart` is **deliberately not migrated** — history and live WS must move together per #359. Dev also found `backtestEngine` silently simulating pepe/bonk at 1000× via the Bybit price factor.

**Correctness and infrastructure:** #1031 (SMA was not a rolling mean), #1032, #1033, #1026, #1051, #1036 (RLS `auth.uid()` rewrite — **merged but NOT applied**, see Risk), plus QA tooling (#1046 session reuse, #1048/#1055 the Arena overlap gate) and documentation (#1040 the owner-blocker register, #1041, #1054).

## Why

Two of the four user-facing fixes are broken on production as it stands, and the Arena revert returns the page to the UI the owner approved.

## How to test (QA)

**Steps 4 and 6 were corrected after QA reviewed this section — step 4 asked for something QA cannot do on production, and step 6 would have passed on 1000×-wrong data.**

On **liquidity-hq.com after the deploy**, not before — merging ships nothing, both services are `autoDeploy: no`.

1. **Chatbot dropdown** — open the chat, click the coin selector. The list appears attached to the chat panel and inside the viewport. Check at default, expanded/pop-out, and mobile 390. Then the portal-specific regressions: it closes on outside-click, holds position on scroll, and does not survive the chat panel closing underneath it.
2. **Coin dropdown elsewhere** — Settings → Watchlist and Alerts → muted coins. Both were unaffected before; confirm they still are.
3. **Push notifications** — request permission from Settings. The browser prompt appears, or a real error is surfaced. Nothing fails silently.
4. **Arena** — matches the pre-revert production layout. Strategy switching works on the fly. Liquidation Clusters populates.
   - **Pro entitlement (#1049) CANNOT be verified by QA on production. It needs the owner's own account.** The QA fixture accounts live in the dev Supabase project and **do not exist on production at all**. The synthetic-JWT interception QA uses to fake Pro is a client-side fake built for local and staging Playwright runs; pointing it at live production is a materially different use and QA correctly declined to make that call alone. So this step is the owner's to run, or it ships unverified — see Risk level.
5. **Nav bar** — no gap under it on any page.
6. **Prices** — BTC populates, then **PEPE and BONK specifically**, which are the real probe. *"Populates"* is not the check; the numbers are.
   - **Why these two coins are the strong probe:** they have **no Binance mapping at all** (`BINANCE_SYMS` has no entry, and `lib/coins.ts` says so). So there is no silent fallback to hide behind — if Bybit fails for them, `fetchOHLCV` throws `No symbol for {coin}` and the signal visibly errors. BTC would fall back silently and mask exactly the failure this release is about.
   - **What to expect, because a CORRECT screen looks wrong here.** The Arena **chart** deliberately shows Bybit's raw 1000-token listing — labelled `1000PEPE/USDT`, around **$0.003–$0.004**. That is correct: klinecharts' axis breaks below ~0.0000027, so the chart is not divided down. The **EMA Signal card**, Market Structure and everywhere else show the real per-token price, around **$0.0000035–$0.0000037**.
   - **The card is where a broken price factor shows up — it is the number that would be 1000× wrong.** Anyone reading the chart and expecting a per-token price will misread a correct chart as broken.
7. **`/api/version`** reports the merged commit on `main`.

## Risk level

**Medium.**

- **No migrations in this release and no new environment variables.** #1036 carries the RLS rewrite but **applying it is a separate owner-gated action that has not happened** — the SQL ships inert. Do not read this merge as the migration being applied.
- **#1037 is deliberately NOT in this release.** It bundles the `#1020` write path with its schema; shipping the write ahead of the schema reintroduces the bug it fixes, and the failure is silent. It stays open until the owner applies the migration.
- **Telegram and news cannot be verified outside production** — the bot webhook points at dev, and the news cron only runs on prod. Anything touching either ships unverified by design.
- **`staging` shares the dev database**, so a clean staging pass is not proof the data path is clean.
- **CI is off** (the owner's cost decision), so this release has no gate run behind it. Everything above was verified by hand on the deployed staging build.
- **Pro entitlement ships unverified on production unless the owner checks it.** QA has no production account and the fake-Pro technique is not valid there — stated plainly rather than left implied by a test step that reads as routine. #1049 (a Pro subscriber resolving as free) is in this release's own history, so this is the one gap worth the owner spending a minute on after the deploy.
- **Production Supabase has no backups.** Unchanged, and the reason the database gates exist.

## Merge / deploy / re-check / tag

**Owner-gated at two points, and neither is granted by this PR existing.**

1. Owner approves the release.
2. PM/DevOps merges `staging` → `main`.
3. **Owner triggers the production deploy** (Render → `liquidity-hq-prod` → Manual Deploy → Deploy latest commit). Merging alone ships nothing.
4. Re-check the steps above against liquidity-hq.com and quote `/api/version`.
5. PM/DevOps tags `main`.

**Do not promote `qa` → `staging` while this PR is open.** Ship it or close it first. #1068 is queued behind it and rides the next release.


JohnDominicJasmin:
## Candidate enlarged and re-signed-off — QA, 2026-09-10

`staging` is now **`98f3d45c`**: **83 commits, 33 PRs**. It was 27 PRs when this PR opened.

**The candidate grew deliberately, twice, on the owner's instruction** — *"let's park everything in staging, once its all in staging let's ship it."* `CONTRIBUTING.md` says not to promote into `staging` while a release PR is open, because a release PR's head IS its base branch and the candidate then grows underneath a sign-off that already happened. **Here each growth was an explicit decision and QA re-verified the enlarged set afterwards** — deliberate rather than silent, which is what that rule protects against. Recorded here so the next reader doesn't find an unexplained rule violation.

Dev flagged the second promotion as a possible violation before it was explained to them. **That was the correct response and is worth noting rather than smoothing over**: they verified with `git fetch` and `gh pr view` first, raised it, and did not touch `qa` or `staging` themselves. The failure was mine — I lifted the freeze and told only QA.

### What was added since the original body

**#1073 — `KLineProChart` moved to Bybit-primary with a real WebSocket.** History and live ticks moved together per #359. Replaces a 5-second REST poll with a genuine subscription, reusing `LiqFeed`'s existing `connectBB` client rather than writing a second one. **`fet` still uses Binance** — Bybit does not list `FETUSDT` as a linear perp. Forced, not chosen; **#1059 is substantially delivered, not fully.**

**#1081 — the empty-chart fix (#1079), and the reason this release waited.** #1073 introduced a regression QA caught before it shipped: BONK's chart rendered **completely empty** on `staging` while fine on production. The Bybit-via-Render 503s are **old and happen on production too** — roughly half of klines calls — but the old polling absorbed them through repetition. One bulk fetch with no retry turned an invisible transient into a blank chart, and BONK/PEPE have no Binance fallback to hide it.

The fix is a 3-attempt retry (0/500/2000ms) **and — the half that actually closes the hole — an explicit `r.ok` check.** Before it, a 502's error body was indistinguishable from a successful response with zero candles. On exhaustion the user now sees *"Couldn't load price history for <COIN>..."* rather than a blank canvas.

**Docs:** #1071 (owner-blockers row 4 — nobody can open the Supabase dashboard), #1074 (`TEST_GAPS` §6), #1076 (test assertion loosened to the word `REALIZED`, not the exact label text).

### QA sign-off on the enlarged candidate

Verified on deployed `qa` (`98f3d45`), then re-verified on deployed `staging` (`98f3d45`):

1. **Forced BONK's failure with QA's own fetch override**, not Dev's mock — banner fires, on both builds.
2. **BTC unaffected** — normal load, no banner, no added latency on the healthy path.
3. **Retry-then-recover seam** — 2 forced failures then a real success: full history, no gap or duplicate where live ticks join.
4. **PEPE price factor** — VWAP $0.00000366, correct per-token, cross-checked against CoinGecko rather than against another of our own surfaces. **No silent 1000× survived.**

### Not in this release

**#1078 (Arena liq-cluster bands)** — the owner is validating it on localhost and has not reported back; they saw no clusters at all, which is most likely local browser storage and missing local Supabase keys rather than the change. **Unresolved, so it stays out.**

**#1080** — the same one-shot-no-retry shape at 13 further call sites, now **confirmed reachable twice in live operation** (Dev hit a real 502 in `useEMAStrategy`; QA's forced failure broke the same card). Filed, unqueued, wants one shared helper rather than 14 copies.

**#1037** — still parked on the owner's migration.

### Unchanged from the original body

No migrations. No new environment variables. **The RLS rewrite (#1036) ships inert — applying it is a separate owner-gated action that has not happened.** CI is off, so this release has no gate run behind it; everything above was verified by hand on deployed builds. **Pro entitlement cannot be verified by QA on production** — it needs the owner's own account. Production Supabase still has no backups.

### Merge / deploy / re-check / tag

1. Owner approves the release.
2. PM/DevOps merges `staging` → `main`.
3. **Owner triggers the production deploy.** Merging ships nothing — both services are `autoDeploy: no`.
4. Re-check against liquidity-hq.com and quote `/api/version`.
5. PM/DevOps tags `main`.

JohnDominicJasmin:
**QA Team.** Attempted the owner-approved Arena E2E run against deployed staging (98f3d45) - 5 auth-free specs, scoped exactly as PM/DevOps directed. **The machine OOM-killed it before it could complete, and I'm not retrying.**

**What actually ran cleanly before the crash cascade, and can be trusted:**
- `arena-legacy-signals.spec.ts` - 3 passed: seed reaches the page, a stored LONG renders bullish (not FLAT), a stored SHORT renders bearish (not FLAT).

**Everything else is harness failure, not product failure - every single one, attributed precisely:**
- 6 "failed": every one is either `browserContext.newPage: Target crashed` or `worker process exited unexpectedly (code=3221225794, signal=null)` - 0xC0000005, a Windows access-violation crash. Not an assertion failure anywhere.
- 7 "skipped" + 13 "did not run" - cascade from the same crash, no coverage, not clean passes.
- **`arena-panel-overlap.spec.ts` (the overlap gate) never finished its sweep** - only got through 768/850/900/1000px before the crashes started; 1058 and 1100 crashed, the rest never ran.
- `klines-route.spec.ts`, `signal-freshness.spec.ts` - same story, crashed before any real assertion ran.

This machine OOM-killed two full suite runs two nights ago and now this scoped 5-file run too. Checked for orphaned processes after - none left running, but the pattern is consistent enough that I don't trust a retry to behave differently right now.

**Per PM/DevOps's own stated fallback: shipping without this.** Manual verification already covers this release's actual changes (both #1073 and #1081 verified live on qa and staging, forced-failure and all). This run was additional insurance on the untouched-area risk (Arena layout post-revert), and 3 clean passes on legacy-signal vocabulary is the only real signal it produced - everything else is the harness, not the release.

Not retrying tonight. Recommend proceeding to the owner without it, with this noted as an open gap rather than a pass.

JohnDominicJasmin:
**QA Team.** Second attempt, per the owner's "TRY AGAIN" - single worker, headless (already default), `arena-panel-overlap.spec.ts` alone, auth-free. Also crashed. **Stopping here - not attempting a third, per PM/DevOps's own instruction.**

**Single-worker DID make a partial, measurable difference - worth recording beyond this release:**
- Reached 4 completed tests with clean, correct verdicts (768/850/900/1000px all skip honestly - the terminal Arena is genuinely reverted/absent, so `terminalArenaAbsent` correctly skips rather than false-passing) before the crash hit at test 5 (1058px).
- The parallel 2-worker run crashed earlier and scrambled the ordering, making "how far did it get" hard to read cleanly. Single-worker gave a clean, linear, honestly-attributable partial result - same eventual crash point (1058px+), but a materially clearer picture of what ran before it.
- **The crash itself is identical both times**: `browserContext.newPage: Target crashed` then `worker process exited unexpectedly (code=3221225794)` - a failed memory allocation launching a new browser context, not a product fault, not a harness bug. PM/DevOps's own measurement (2.3GB free, Brave + 20 Claude processes) explains it precisely.

**Standing finding, not specific to this release: "run the tests" is not reliably available on this hardware right now**, regardless of scope (single spec, single worker, headless, auth-free all tried). The ceiling is memory, and nothing in QA's control changes that - not a harness fix, not a scope reduction below what was already tried tonight.

**4/11 of the overlap gate's checks got a clean, honest result** (correctly skipping the reverted terminal Arena) before the crash. Not a pass or a fail - a partial run that stopped for a reason unrelated to the release.

No orphaned processes after either attempt. Recommend the owner's production decision proceed on the manual verification already done (#1073, #1081 both forced-failure-tested live on two deployed builds) - this E2E attempt adds no new information about the release itself, only about the machine.

JohnDominicJasmin:
**QA Team.** Production deployed - confirmed `/api/version` reports `6e09cfa`. Post-deploy re-check, live on liquidity-hq.com.

**Checked, all clean:**
- **Nav bar gap** (`--strip-h` fix) - no gap under the nav on Dashboard, breaking-news banner sits flush.
- **GrokChat coin dropdown** (#1066 portal fix) - opens, portaled to `document.body`, fully within viewport.
- **BONK chart** - full dense history, no gap, no banner, correctly labeled `1000BONK/USDT`, VWAP `$0.00000300` (the coin the whole #1079/#1081 chain was built to fix).
- **PEPE chart** - same, VWAP `$0.00000361`, matches the CoinGecko-verified value.
- **Push notification guard** - confirmed the feature-detect pattern (checks `typeof requestPermission === 'function'`, not just `'Notification' in window`) doesn't throw against the exact stub shape #1042's original crash report used.

**Not re-run tonight, already thoroughly verified pre-release and unchanged by this deploy:** Arena's classic-layout match to the pre-revert reference, Strategy Panel switching, Liq Clusters rendering. These were covered extensively during #853/#1053's own review cycle; nothing in tonight's release touches that surface differently.

**Still can't verify:** Pro entitlement resolving as Pro on production - needs the owner's own account, unchanged limitation.

Production matches what was verified on staging. Nothing found that wasn't already known.

JohnDominicJasmin:
**Owner confirmed Pro entitlement on production, 2026-09-10: *"my account is in pro"*.**

That was the **last outstanding verification on this release** and the one QA physically could not perform — the fixture accounts live in the dev Supabase project and do not exist on production, and the synthetic-JWT interception QA uses is a client-side fake built for local and staging runs, not valid against production.

**So #1049 (a Pro subscriber resolving as free) is confirmed fixed on the deployed build**, by the only person who could confirm it.

**Release v2026.09.10 is now fully verified.** Merged, deployed by the owner, re-checked live by QA (nav gap gone, chatbot dropdown working, BONK and PEPE rendering full history, push-notification guard holding against the original crash stub), tagged, and production serving `6e09cfa` matching `main`.

**The one open gap remains what it was, stated as a gap and not a pass: the E2E suite never ran for this release.** CI was not switched on — my omission, not a hardware limit — and the two local attempts were killed by memory exhaustion on a machine that also killed a background monitor and an ordinary git hook the same evening. **The fix is switching CI on for the next release so the suite runs on GitHub's machines.**

**A finding came out of the owner's Pro check that is worth more than the check itself, and it is now its own issue: the product has no plan indicator at all.** Nothing anywhere shows Free, Pro or trial. Entitlement is enforced across the backend and never surfaced to the user. The only way the owner could confirm their own Pro status was that `/upgrade` redirects Pro users away — an accident of routing, not a feature.

PR #1073, opened 2026-09-08
fix(chart): Bybit primary + real WebSocket for KLineProChart (#1059 remainder)

JohnDominicJasmin:
## Summary

The last two of #1059's five originally-scoped call sites. `lib/useEMAStrategy.ts`, `components/HigherTfMoveBadge.tsx`, and `lib/backtestEngine.ts` flipped to Bybit-primary earlier (#1059/#1060); `components/KLineProChart.tsx` (the Arena/dashboard chart) was deliberately held back because `getBars` (REST history) and `subscribeBar` (live ticks) have to move together per #359, and `subscribeBar`'s existing Bybit path was a 5-second REST poll, not a WebSocket — making it primary for every coin would have swapped instant ticks for a 5s-stale poll on the flagship chart, a visible regression, not an invisible background swap like the other three sites.

**Not shipping a poll.** `components/LiqFeed.tsx`'s `connectBB` already has a working Bybit v5 public WebSocket client — host alternation, subscribe handshake, reconnect. `next.config.ts`'s CSP already allows both Bybit WS hosts. This PR adapts that exact pattern to a `kline.{interval}.{symbol}` topic instead of LiqFeed's liquidation topic — a second caller of a proven pattern, not new infrastructure. Third time this project has found a solved problem in this codebase unapplied at a second call site this week (the `--strip-h` nav gap and the coin-select popover were the other two) — worth grepping for next time before assuming something needs building from scratch.

## What changed

- **`getBars`**: reordered to try Bybit REST first, Binance (spot→futures) as fallback — same shape as the other three #1059 sites.
- **`subscribeBar`**: Bybit branch is now a real WebSocket (`wss://stream.bybit.com/v5/public/linear`, alternating to `stream.bytick.com` on repeated failures), reusing this file's own exponential-backoff reconnect and `online`-listener pattern (already established for the Binance branch, #306) rather than inventing a new one. The 5s poll is gone, superseded.
- **`fet`** — the one coin with a Binance symbol and no Bybit linear perp (`lib/coins.ts`'s own comment) — keeps its unconditional Binance-only branch in both functions, unchanged.

## Verification

Live, not just the diff:
- `getBars`: REST fetch for BTC 1h returns the full 1000-candle Bybit history, byte-identical to a direct `curl` of the same endpoint.
- `subscribeBar`: WS connects, subscribes to `kline.60.BTCUSDT`, and keeps receiving frames — watched 8 arrive over 20s on 1m with the same forming candle's volume ticking up (9.047 → 10.547 → 10.688), the live-update behavior a real WebSocket gives that a 5s poll couldn't match.
- `fet`: confirmed the chart still connects to `wss://stream.binance.com:9443/ws/fetusdt@kline_1h`, untouched.
- Zero console/page errors from either change. `tsc`/lint/tests all clean.

## One thing found, not fixed, disclosed rather than glossed over

Locally, the chart renders only ~2 candles with large empty space regardless of wait time, despite the REST fetch demonstrably returning the full 1000-candle dataset. **Reproduced the identical symptom on the unmodified pre-#1059 code** (stashed this change, same test, same result) before concluding it's pre-existing and out of this PR's scope — not something this change introduced. Flagging for QA to check on a real browser or deployed environment; my local repro is headless Playwright and may not reflect a real browser's behavior.

## How to test (QA)

On `qa`/staging, `/arena`:
1. Open browser devtools → Network → WS, select BTC (or any major), confirm a connection to `wss://stream.bybit.com/v5/public/linear` subscribed to a `kline.*` topic, with frames arriving as the current candle updates.
2. Switch to FET specifically — confirm it still connects to Binance (`stream.binance.com`), not Bybit.
3. Watch the chart for a minute on a fast timeframe (1m) — candles should update live, not just on page load.
4. **Please also check whether the chart renders its full candle history correctly** — my local headless testing showed a possible rendering gap (see above) that I couldn't confirm is real or a test artifact.

## Risk level

**Medium.** Touches the live data path for the highest-traffic chart in the app. Mitigated by: reusing an already-proven WebSocket pattern rather than writing one from scratch, preserving the exact reconnect/backoff/online-listener scaffolding already trusted for the Binance branch, keeping `fet`'s fallback path completely untouched, and disclosing the one thing I couldn't fully verify (chart rendering) rather than claiming a clean bill I don't have.

**Reconnect gap-backfill:** re-fetches on reconnect via `backfillGapBybit`, same shape and guard as the Binance branch's already-trusted `backfillGap` (#313) — confirmed the reconnect+host-alternation machinery fires live, but could not force a genuinely *unclean* close from same-page JS to watch the backfill request itself (`ws.close()` from the same page always reports `wasClean: true`, confirmed directly, so the guard correctly skips backfill on it — that's not a bug, it's a real limit on what a browser-side test can simulate). The code path is structurally identical to an already-shipped mechanism; proving the dirty-close case specifically needs a real dropped connection, not a script.

**`fet`'s residual Binance dependency:** one coin (FET) still connects to `stream.binance.com` for both history and live ticks — forced, not a shortcut. Bybit does not list `FETUSDT` as a linear perp (`lib/coins.ts:103`); the alternative was dropping FET from the product. For users on a network where Binance itself is blocked, FET's chart is the one that still breaks. **#1059 is substantially delivered by this PR, not fully closed** — the owner's instruction was "let's not use Binance anymore," and this one coin doesn't fully deliver that.

**No rush per PM/DevOps** — doesn't need to land before the current release (#1069) clears, and this PR doesn't touch anything that PR includes.

Related: #1059, #1060 (the three sites that shipped first), #359 (why this file waited), #306 (reconnect pattern precedent), `LiqFeed.tsx` (WebSocket client prior art).


JohnDominicJasmin:
**QA Team.** Reviewed the diff line by line, verified every source claim independently.

Confirmed directly:
- `components/LiqFeed.tsx`'s `connectBB` genuinely has the exact WebSocket pattern described (`wss://stream.bybit.com/v5/public/linear` primary, `wss://stream.bytick.com/v5/public/linear` alternate, same host-alternation shape) - this PR is reusing real prior art, not inventing something new.
- `next.config.ts`'s CSP already allows both hosts (`connect-src` line 41) and `api.bybit.com` (line 45) - no CSP change needed, matches the PR's claim.
- `fet` genuinely has no Bybit linear perp - `lib/coins.ts:103`'s own comment says so, and both `getBars` and `subscribeBar`'s `fet` branches are left untouched, unconditional Binance.
- Old 5s poll cleanly removed, not left as dead code.
- `getBars`'s primary/fallback restructure (`applyBars` helper de-duplicating the three near-identical bar-processing blocks that existed before) is a real simplification, not just a reorder - fewer places for the SR-levels/structure-signals/EMA-sync logic to drift out of sync.

**On the disclosed candle-rendering gap:** noted, and the way it's disclosed - reproduced on unmodified code before concluding it's pre-existing - is exactly right. Can't verify on a real browser from a local dev server right now (this machine is under real memory pressure tonight, two E2E runs already got OOM-killed). Since this doesn't need to land before #1069 and staging is frozen behind it anyway, I'll check candle rendering live once this reaches a deployed environment rather than force a risky local run tonight for something already flagged as not urgent.

Code review passes. Merge whenever - sequencing is yours given no rush.

JohnDominicJasmin:
## Update: two questions answered before merge

**Does the reconnect path re-fetch history to close the candle gap, or does it leave a silently wrong chart?**

It re-fetches. `backfillGapBybit` fires on `ws.onopen` whenever `reconnected` is already `true` (i.e., this isn't the first connect) — same shape, same guard, same call site as the Binance branch's own `backfillGap` (#313), just against Bybit's REST response shape. It fetches the last 500 candles, keeps only what's newer than `lastBarTsRef.current` via the existing `barsAfter` helper, and feeds each missed bar through the same callback the live stream uses, oldest-first.

**Tried to prove this live, and hit a real limitation worth recording rather than silently working around:** a same-page `ws.close(4000, ...)` call always produces `wasClean: true`, regardless of the code passed — confirmed directly (`close events: [{"code":4000,"wasClean":true}]`). The reconnect guard (`if (cancelled || e.wasClean) return;`) correctly treats that as a clean close and skips backfill, which is *correct* behavior for a clean close, but it means a same-page script cannot simulate a genuine network-drop close from inside the page — there's no way to produce `wasClean: false` without an actual severed connection (killing the server side, or a real network blip). So I can't hand QA a screenshot of the dirty-close path firing. What I can say: the code is a direct structural copy of the Binance branch's already-trusted #313 mechanism, unmodified in shape, and the reconnect+host-alternation machinery around it is independently confirmed working (watched it alternate `stream.bybit.com` → `stream.bytick.com` across repeated attempts). If this needs to be proven beyond that, it needs a real dropped connection (killing the WS server-side, or testing on a flaky network), not a browser-side script.

**`fet`'s residual Binance dependency**, stated in the terms agreed: one coin (FET) still connects to `stream.binance.com` for both history and live ticks, because Bybit does not list `FETUSDT` as a linear perp (`lib/coins.ts:103`) — this is forced, not a shortcut; the alternative was dropping FET from the product. For users on a network where Binance itself is blocked, FET's chart is the one that still breaks. **#1059 is substantially delivered by this PR, not fully** — worth stating plainly so nobody reads the issue as fully closed once this merges.

JohnDominicJasmin:
**Dev Team — `dev` → `qa` promoted, combined, one push.**

`qa` now at `bfcf4fdc`. All three merged together per PM/DevOps direction:

- #1073 — KLineProChart Bybit-primary WebSocket (`getBars`+`subscribeBar`). Risk section carries the reconnect-gap-backfill testing limitation and the `fet` residual Binance dependency.
- #1074 — QA's `TEST_GAPS.md` §6 update. Reviewed (docs-only, QA-owned path), merged per reverse handoff.
- #1071 — `docs/OWNER-BLOCKERS.md` row 4 (Supabase dashboard access). Checked the wording against #1025's own thread before merging — matches exactly (7×504 in the partial run, `TOTAL outbound: 1`/5→1 caching claim, permission error on `wdtjhrilakoitfcezxpx`), not overstated.

Gate: lint 0 errors, tsc clean, 781/781 unit tests passing on the merged `qa` branch. Build not re-run (pre-push policy — promotion only, no app-code risk beyond #1073).

**`staging` untouched** — stays frozen behind #1069 until the owner acts, as instructed. Nothing here promotes past `qa`.

QA — ready for your deploy + `/api/version` check whenever convenient.

JohnDominicJasmin:
**QA Team.** Live check on deployed qa (bfcf4fd), `/arena?coin=btc&tf=1h`.

**Candle-rendering gap does not reproduce on a real browser.** Full, dense 1h history renders correctly - continuous candles spanning 09-05 through 09-08, no gaps, no empty regions, EMA ribbon and structure markers all drawing over it correctly. Took ~16s to fully populate (screenshot at 8s still showed an empty grid, second screenshot at 16s was fully rendered) - matches your own suspicion that the headless repro may not reflect real browser behavior. No console errors.

Confirms your disclosure was the right call either way - real or artifact, flagging it was correct, and now it's checked.

JohnDominicJasmin:
**QA Team.** The candle-rendering gap is real, reproduces reliably, and here is the mechanism.

## Decisive test: production vs staging, same coin

`liquidity-hq.com/arena?coin=bonk` renders correctly - full dense history, no gap. Staging (`bfcf4fd`) shows an empty chart for BONK on 3 separate fresh loads. Production runs pre-#1060 code, so this isolates what changed.

## Root cause: a missing retry, not a data-source problem

Captured actual network requests on both.

**Production's pattern:** almost entirely repeated `interval=60&limit=1` calls - the old 5s poll this PR replaced with a WebSocket. **503s happen there too** - roughly half of them failed in one capture - so the 503 itself is a real, pre-existing, shared condition on this infrastructure (Render + Bybit), not staging-specific. But the old poll is self-healing by construction: a failed poll is invisible, the next one 5 seconds later just succeeds.

**Staging's pattern (this PR's code):** one bulk `limit=1000&closed=1` fetch on mount for the initial history, no retry. If that single call hits the same 503 that production's polls shrug off constantly, the chart has zero bars to start from - and the new WebSocket then only appends live ticks onto nothing. Nothing re-fetches history afterward, so it stays empty for the rest of the session.

So: not new API misuse, not a wrong endpoint - **the old polling architecture had accidental resilience to a real, ongoing upstream flakiness through sheer repetition, and this PR's one-shot fetch doesn't have it.** "Flaky Bybit" undersells it - the flakiness was always there, the old code just never let it matter.

## Suggested fix

A retry (2-3 attempts, short backoff) specifically on `getBars`' initial Bybit history fetch - the same resilience `subscribeBar`'s `backfillGapBybit` already has for reconnect gaps, just applied to the first load too.

## Coins affected

BONK and PEPE - no Binance fallback (`lib/coins.ts`), so a failed Bybit history fetch has nowhere to fall back to. BTC/majors have Binance as a safety net, which is likely why they rendered fine in my earlier spot-checks.

This is why the release is holding rather than shipping tonight - not a rush call, a real "chart shows nothing" defect on at least 2 coins.

Issue #1075, opened 2026-09-08
Arena chart: realized-liq labels are the loudest thing on screen and stack into a block — merge nearby clusters, match the S/R chip style


JohnDominicJasmin:
**Owner-directed, 2026-09-08.** The owner looked at the live Arena chart, said *"looking at the screen its messy"*, was given three options and chose **A and B together, explicitly skipping C**.

**This is a sanctioned exception to the 2026-09-07 scope lock** — the owner's own "go", on the day, about something they were looking at. Same class as #1042.

---

## What is wrong, from the owner's screenshot

**1. Two label systems on one chart.** Support/resistance sits on the right as small compact chips (`R $80,139`). Realized liquidations sit on the left as **wide opaque pink boxes** with `REALIZED LIQ` spelled out in full, six times. Same kind of information — a horizontal price level — in two visual languages on opposite edges.

**2. The loudest element carries the least actionable data.** Realized liquidations are **already-spent fuel** by their own definition. S/R and the buy/sell signals are what the user acts on, and they are the quietest things on the chart.

**3. Three of the eight are the same zone drawn three times.** `$79,000`, `$79,200`, `$79,400` are ~0.5% apart and stack into a solid pink block over the candles.

## A — merge nearby clusters into one band

Collapse clusters within a small price band into a **single shaded zone with one label carrying the combined total**. Six labels become two or three.

**This loses no information and arguably gains some**: *"$130K liquidated across 79.0–79.4"* is a truer statement than three separate lines that a reader has to add up.

**Pick the threshold from the data and say why in the PR.** ~0.5% is the owner's observation, not a measured parameter — do not ship an arbitrary constant with no reasoning attached, that is how the next person inherits a magic number.

## B — match the S/R chip style, move to the right edge

Small chips in the same visual language as `R $80,139` / `S $77,577`, on the **right edge with everything else**, so all price-level labels read as one system down one column.

### ⚠️ The naming rule is load-bearing and must survive the shortening

`lib/liqClusters.ts` is explicit, and it is the one thing that must not be lost to brevity:

> REALIZED liquidations — positions that already blew up, price memory, fuel already spent. **NOT the predicted liquidation levels Coinglass sold, which are a forward magnet. The two look identical drawn on a chart, so every label this feeds must say realized.**

So `LIQ $81K` **is not an acceptable shortening** — it drops the only word separating "this already happened" from "price will be pulled here", and those are opposite trading conclusions.

[...]

PR #1078, opened 2026-09-08
feat(chart): merge nearby realized-liq clusters, match S/R chip style

JohnDominicJasmin:
> **DO NOT MERGE.** Owner-directed exception (#1075) — owner runs this on localhost themselves first. PM/DevOps hands over the branch name; this PR exists for reviewability, not for merging on QA sign-off. Wait for PM/DevOps to say the owner has looked.

## Summary

Owner-directed, same-day exception to the scope lock (#1075, same class as #1042). Not for merge yet — **the owner wants to run this on localhost and look at it himself first.** Lands as a reviewable branch only; do not promote to `qa` until PM/DevOps says the owner has looked.

## What changed — A: merge nearby clusters into one band

New `mergeLiqBands` (`lib/liqClusters.ts`) collapses clusters within `LIQ_BAND_MERGE_BUCKETS` (2) of LiqFeed's own bucket width of a neighbour into one band. Chained the same way `computeLabelOffsets` already chains label-stack detection, so $79,000/$79,200/$79,400 merge into one band even though the two ends alone can exceed the threshold.

**Threshold, and the wrong first answer.** First cut defined "close" as a fraction (2.5%) of the whole eight-cluster set's own price spread — borrowed from `computeLabelOffsets`. It's wrong: run the owner's own three-cluster example through it with realistic outliers ($76,000/$83,000 also among the eight) and the spread balloons, 2.5% of it shrinks under $200, and the trio the owner was pointing at stops merging. Caught by running the numbers before shipping, not by inspection.

Replaced with a threshold built from the actual source data: `bucketSz` — the price-rounding tier LiqFeed already uses to group raw liquidation events into clusters ($200 at BTC's price). Moved from a private copy in `components/LiqFeed.tsx` into `lib/liqClusters.ts` and exported, so `mergeLiqBands` and `LiqFeed.tsx` share one definition instead of two. 2 bucket-widths (not 1 — the owner's trio was each pairwise gap = exactly one bucket, so a same-bucket-only rule wouldn't have merged the motivating example). At BTC's price on 2026-09-08 that's $400, ≈0.5% of price — matching the owner's own eyeball almost exactly, not by coincidence: LiqFeed's bucket width is *why* the clusters landed $200 apart to begin with.

**#766 interaction.** "Eight" still means the eight heaviest raw clusters, selected exactly as before — merging runs after that selection, not instead of it. Visible bands now vary 2–8 depending on how clustered the heaviest eight happen to be at that moment. Still lines, still not a histogram; not a reversal of #766.

**Chart-only, AI path untouched.** `mergeLiqBands` never touches `topClustersForCoin` or `formatClustersForPrompt` — confirmed the isolation is structural, not something added: `app/arena/page.tsx` already derives `chartLiqClusters` (chart) and `liqClusters`/`formatClustersForPrompt` (AI prompt, keeps every raw cluster's long/short side) as two separate values from the same raw state.

## What changed — B: match the S/R chip style, move to the right edge

`liqClusterLine`'s label moves from the left edge (`REALIZED LIQ $79,000 · $130K`, wide) to the right edge, matching `srLevelLine`'s chip exactly — same position, padding, background, border radius.

**Content, not just position — and this went through one revision.** First draft kept price and dropped the total (`REALIZED $79,400`), reasoning "match S/R's `<TAG> $<PRICE>` shape exactly." Wrong call, corrected before commit: the price is already on screen twice (the line's own vertical position, and the axis) — printing it a third time says nothing S/R doesn't already say better, and it would make an $81K band and a $13K band read identically, throwing away the one thing worth ranking. Final: **`REALIZED $130K`** — the merged dollar total, the one thing on this label nothing else on screen shows.

`REALIZED` (not `REALIZED LIQ`) — "LIQ" was redundant with the `Liq` toggle that controls the whole layer. `REALIZED` is the one word that must survive any further shortening (`lib/liqClusters.ts`'s own docs: predicted liquidation levels and realized ones are drawn identically, so the label is the only thing telling them apart).

Local dev server, `/arena?coin=btc&tf=1h`, real streamed BTC liquidation data (not synthetic):
- Real merge observed: the heaviest-eight clusters rendered as **2** right-edge chips (`REALIZED $131K`, `REALIZED $39K`) instead of up to 8 stacked lines. `$131K` at ~$79,185 is consistent with three raw clusters merging — matches the issue's own illustrative "$130K liquidated across 79.0–79.4" almost exactly, on real data, not a constructed test case.
- S/R chips (`R $80,089`, `R $79,696`, `S $77,491`) confirmed unaffected — same position, same style, coexisting with the new right-edge Liq chips.
- `Liq` toggle confirmed still hides/shows the whole layer (verified via direct DOM state check — `klc-tool-btn on` ↔ `klc-tool-btn` — not just visual inspection).
- **The disclosed collision risk is real, not hypothetical — caught it live.** With `Liq` on, an S/R chip (`R $79,209`) and a Liq band briefly overlapped at the same price during initial load (visible as garbled `REALI` + `R $79,209` text). With `Liq` off, only the S/R chip remained at that exact level, confirming it. Not fixed in this PR — see Risk.
- Console: one hydration warning present, in `RootLayout`'s `<html><script>` structure — unrelated to every file this PR touches (confirmed: it's a root-layout warning, nothing in the trace references `liqClusters`/`LiqFeed`/`KLineProChart`). Pre-existing.
- `tsc --noEmit`: clean. `eslint` on all three changed files: 0 errors (pre-existing warnings elsewhere in `KLineProChart.tsx`, none new or related). `npm run build`: not run — see Risk.

## Not verified

- **Terminal design mode / light theme.** Only checked dark, current design mode. This PR doesn't touch `app/globals.css`, `labelKeys`, `labelDefaults`, or `AppShell.tsx`, so the landing hard gate doesn't apply — but the chip's own `border-radius: 0` terminal-mode branch (mirrors `srLevelLine`'s existing pattern) wasn't visually confirmed in terminal mode specifically.
- **Build.** `npm run build` not run before this local-only branch — no promotion is happening yet, so the full build gate wasn't spent on a branch the owner hasn't looked at. Will run before this goes to `qa`.
- **A coin with sparse/no liquidation data** (the empty-state check from the issue's own test plan) — not checked this pass; time-boxed to the owner's own BTC screenshot scenario tonight.

## How to test (QA / owner, once this moves past localhost)

Arena chart, BTC 1H, `Liq` toggle on.
1. Clusters within ~2 LiqFeed bucket-widths of each other render as **one** right-edge chip, not several stacked lines.
2. Every chip says `REALIZED $<total>` — nothing reads as a predicted/forward level, nothing says "LIQ."
3. Chips match the S/R visual language (position, padding, style) and sit on the same right edge.
4. `Liq` toggle still hides and shows all of it.
5. AI path unaffected: Quick/Deep Research still receive every raw cluster with its long/short side (this PR doesn't touch that code path — confirmed structurally, not just by not editing it).
6. Empty state: fresh page, no liquidations yet — chart shows nothing, AI still gets `Not available`. **Not personally re-verified this pass** — flagged above.

## Risk level

**Medium.** Chart-visible, owner-directed, touches a shared price-level rendering pattern (`srLevelLine`'s exact style is now reused, not just referenced).

- **S/R-vs-Liq label collision, disclosed and observed live**, not fixed here. `computeLabelOffsets` runs once per overlay type (S/R, GEX, Liq bands each independently) — sharing the right edge makes a collision between an S/R chip and a Liq band chip more likely than when they sat on opposite edges. Pre-existing pattern (S/R and GEX already didn't coordinate); this PR makes it more likely to matter. Not unified in this pass — flagged for the owner's own look, candidate fast-follow.
- **`__tests__/liqClusters.test.mts:77-86` will fail as-is** — its assertion literal-matches the old `REALIZED LIQ ` text, which no longer appears anywhere in the source by design. 16/17 tests in that file still pass; this one isn't mine to fix (QA owns every test, including one-line assertion updates) — flagged directly to QA.
- **`fmtLiqUsd`'s formatting is reused unchanged** (`$130K` / `$1.2M` style) — no new formatting logic introduced, lower risk than it might look from the label-text change alone.
- Build gate not run (see Not verified) — will run before any promotion.

Related: #1075 (this issue), #766 (the "eight" ruling this respects), `Tip.tsx`/`CoinMultiSelect.tsx` (unrelated, but the same "grep for the existing pattern before inventing a new one" instinct that led to reusing `srLevelLine`'s exact chip style and LiqFeed's `bucketSz` instead of a fresh guess for either).


JohnDominicJasmin:
**QA Team.** Code review only, per PM/DevOps - not a merge signal, this waits on the owner's localhost look.

**`mergeLiqBands` verified by hand-trace**, not just read: ran the owner's own $79,000/$79,200/$79,400 trio through the chained-adjacency logic. Each pairwise gap is $200, threshold is `bucketSz(mid) * 2` = $400 at BTC's price tier - all three merge into one band, weighted-average price lands where the PR claims. Confirms the "chained, not span-based" comment is accurate: a naive span check ($79,000 to $79,400 = $400, right at the boundary) would be fragile; the actual implementation compares consecutive gaps, which is more robust and matches what's documented.

**`bucketSz` move to `lib/liqClusters.ts` is a clean dedup** - `LiqFeed.tsx` now imports it back rather than keeping two copies in sync, correctly reasoned (lib/ can't import components/, so the shared home has to be lib/).

**Chart/AI-path isolation claim checked structurally**: `mergeLiqBands` only touches the KLineProChart render effect, never `topClustersForCoin` or `formatClustersForPrompt`. Confirmed by reading the diff, not just trusting the PR's own claim.

**Label change matches #1076 exactly** (the test fix that already merged into dev) - `REALIZED $<total>`, verified against the real source line.

**On the two disclosed gaps** (S/R-vs-Liq collision, terminal/light mode unchecked): both look like the right calls to leave open rather than scope-creep into fixing tonight, given this is a same-day owner-directed exception, not a full release candidate.

Nothing here changes my read from a review pass. Waiting on PM/DevOps to say the owner's looked before this goes anywhere near `qa`.

JohnDominicJasmin:
QA Team

Reviewed per PM/DevOps's routing decision (owner's localhost check was environment, not the change - re-routed to qa review). Verified against the actual PR branch content, not the diff description alone.

**1. bucketSz extraction — LiqFeed unaffected.** Confirmed the function body moved byte-for-byte (same five thresholds, same values) from LiqFeed.tsx into lib/liqClusters.ts; `snapBucket` still calls it the same way. Pure relocation, no behavior change to LiqFeed's own grouping.

**2. REALIZED label — always present, verified live against source.** `text: `REALIZED ${fmtLiqUsd(total)}`` - the literal word is unconditional, no code path can produce a chip without it.

**3. AI path isolation — stronger than 'structurally separate,' it's type-incompatible.** `formatClustersForPrompt` requires `longUsd`/`shortUsd` on its input type (that's how it derives the long/short side). `mergeLiqBands`'s `LiqBand` interface carries only `{price, total, count}` - no long/short fields exist on it at all. Even a future mistake couldn't route merged bands into the prompt path without a type error first. Confirmed `app/arena/page.tsx` derives `chartLiqClusters` (chart, feeds `mergeLiqBands` inside the chart component) and the AI's `liqLevels` (via `formatClustersForPrompt`, line 891) as two independent reads of the same raw `liqClusters` state - matches the PR's claim exactly.

**4. Merge threshold logic - traced by hand.** `mergeLiqBands` chains against `members[members.length-1].price` (the nearest already-merged neighbour), not the band's first member - confirmed this is genuinely chained adjacency, not band-average distance, matching the computeLabelOffsets comparison it claims to mirror.

**5. The disclosed test break is already fixed - independently, before this PR existed.** `__tests__/liqClusters.test.mts`'s REALIZED assertion was already loosened to a word-prefix match (commit `6289ae3`, same day) rather than the old literal-suffix match. Swapped this PR's actual `components/KLineProChart.tsx`/`lib/liqClusters.ts`/`components/LiqFeed.tsx` into a local checkout and ran the real suite against them: **17/17 pass**, including that exact test. Nothing for QA to update here - already done.

**6. S/R-vs-Liq collision at 9,209** - disclosed, reproducible pattern (independent label-offset computation per overlay type), not blocking, correctly flagged for the owner to see rather than silently fixed or silently shipped.

Approved. Note: the PR's own header still says "DO NOT MERGE" from the original localhost-first routing (#1075) - that's now superseded by PM/DevOps's re-route, flagging so whoever merges doesn't read the banner as still current.

JohnDominicJasmin:
**QA Team.** Already fully reviewed and approved above (6 points checked against source, including the disclosed test break already fixed by #1076, and the DO-NOT-MERGE banner already flagged as superseded by PM/DevOps's re-route). Re-stating plainly since this PR has been sitting open for days despite the approval already being here:

**Merge it.**

Issue #1080, opened 2026-09-09
The one-shot Bybit klines fetch with no retry isn't unique to the chart — 13 more call sites, same shape


JohnDominicJasmin:
**Dev Team — surfaced while fixing #1079, per its own ask #2 ("check whether the same one-shot-no-retry shape exists elsewhere").**

It does. Every one of these hits `/api/market/klines?source=bybit...` with a single `fetch`, no retry, same shape #1079 fixed in `KLineProChart.tsx`'s `getBars`:

| File | Line | Interval/limit |
|---|---|---|
| `lib/backtestEngine.ts` | 77 | paginated, `limit=${PAGE}` |
| `lib/sparklineData.ts` | 26 | `60/24` |
| `lib/useEMAStrategy.ts` | 108 | `${interval}/${limit}&closed=1` |
| `components/AbsorptionDetector.tsx` | 218, 219 | `15/50`, `60/20` |
| `components/CorrelationTerminal.tsx` | 37 | `${bbInt}/${limit}` |
| `components/GrokChat.tsx` | 363 | `60/300` |
| `components/HigherTfMoveBadge.tsx` | 45 | `240/${LOOKBACK_BARS+1}` |
| `components/MarketProvider.tsx` | 451, 972 | `15/100`, `60/4` |
| `components/MarketStructure.tsx` | 161 | `240/100` |
| `app/arena/page.tsx` | 1206, 1247 | `${bybitInterval}/300`, `D/20` |
| `app/correlation/page.tsx` | 47 | `${bbInt}/${limit}` |

**Not theoretical - caught one failing live, twice, while measuring #1079's fix.** `lib/useEMAStrategy.ts`'s `fetchBybitKlines` (line 106-123) hit a real `503` on BONK's 2h and daily candles during this morning's testing, moments apart from the same symbol's chart history also refusing. Console: `[useEMAStrategy] kline fetch failed: Error: Bybit klines 502` (a separate capture, last night). Two independent confirmations, not a one-off.

**Same root cause as #1079**: the old 5s-poll architecture's incidental retry-through-repetition covered every one of these too. Whatever moved a call site to today's one-shot pattern removed the poll; nothing added retry back at any of these sites. Fourth-plus instance of this codebase's "something removed, something quietly depending on it left behind" pattern (`--strip-h`, `RELEASE_PR_PAUSED`, the release-PR automation, #1079 itself).

[...]

PR #1081, opened 2026-09-09
fix(chart): retry Bybit history fetch, surface exhaustion visibly (#1079)

### LINKED ISSUE #1079: Chart renders empty when the one-shot history fetch hits a 503 — BONK reproduces reliably, no retry anywhere
**Owner-deferred 2026-09-08: *"You can fix it later. We will fix it later."* Filed so it is not lost.**

## What happens

BONK's Arena chart renders **completely empty** — correctly-sized canvases, no bars. Reproduced by QA on deployed `staging` across 3 fresh page loads, 20–40s each.

## The mechanism, which is the part worth keeping

**The upstream 503s are not new and not staging-specific.** QA captured production's network traffic too: roughly half the klines calls 503 there as well. The route's own code only ever returns 502 for upstream failures, so **the 503 comes from Render's platform layer, not app code.** A direct `curl` to the same endpoint moments later succeeded 4/4.

**What changed is the request pattern.** The old chart used 5-second polling — repeated small `interval=60&limit=1` calls. A 503 was invisible: the next poll five seconds later succeeded. **It was accidentally self-healing through repetition.**

#1073 replaced that with **one bulk `limit=1000` history fetch on mount, with no retry.** A single 503 leaves the chart with zero bars, and the new WebSocket then appends live ticks onto nothing. **Nothing ever re-fetches history, so it never recovers.**

**Why BONK and not BTC.** BONK and PEPE are Bybit-only — no Binance mapping (`lib/coins.ts`). BTC has a Binance fallback that masks the same failure. **So the fallback was not only a data-source choice; it was hiding a capacity problem.**

## Status of "is this pre-existing?" — it is not, and this is recorded deliberately

The owner's read was *"it happens before"*, and that is understandable — the underlying 503s genuinely are pre-existing and happen on production today. **But the empty chart is not.** QA checked `liquidity-hq.com/arena?coin=bonk` directly: **BONK renders correctly on production**, full dense history, screenshot taken. Production runs pre-#1073 code.

**So #1073 introduced the visible defect by removing incidental resilience.** Recorded plainly rather than softened, because #1073 is currently **inside release PR #1069** — shipping that release as it stands ships this bug. That is a sequencing fact for whoever picks this up, not an argument against the owner's decision to defer.

## The fix

**Retry with short backoff on `getBars`' initial Bybit fetch** — a few attempts, not indefinite — matching the resilience the WebSocket branch already has via `backfillGapBybit`. QA's recommendation, and Dev had started it when work stopped.

**Two things beyond the retry itself:**

1. **Retry exhaustion must be visible.** If every attempt fails, the user must not get a silent empty chart — a chart that says nothing is indistinguishable from a chart with nothing to say. A retry only makes this rarer; it does not fix the user-facing failure.
2. **Check whether the same one-shot-no-retry shape exists elsewhere.** The old polling masked this everywhere it ran, so this is unlikely to be the only place that inherited the assumption.

## The pattern this is the fourth instance of

**Something was removed or replaced, and the thing quietly depending on it was left behind.** `--strip-h` outliving its deleted component; the release-PR automation outliving its pause switch; `RELEASE_PR_PAUSED` outliving the reason it was set; and now polling's incidental retry, which nothing knew it was relying on. **Worth naming in the fix's commit body** — this is becoming the codebase's signature failure mode.

## How to test (QA)

On deployed `qa` or `staging`:

1. `/arena?coin=bonk` renders full history. Repeat on fresh loads — the original symptom reproduced on 3 of 3.
2. Same for PEPE (also Bybit-only, no Binance fallback).
3. **Force the failure path**: with the upstream failing or blocked, confirm the retry fires and, when attempts are exhausted, the user sees an error rather than an empty chart.
4. BTC still renders — it was never visibly broken, and the fix must not change that.
5. Live ticks still arrive after the retried history load, oldest-first, no duplicate or missing bars at the seam.

## Sequencing

**Deferred by the owner. Not queued.** Whoever sequences it should decide first whether release #1069 ships with this known, or waits for the fix.

JohnDominicJasmin:
## Summary

Fixes #1079: BONK's (and PEPE's) Arena chart renders empty when the one-shot Bybit history fetch hits a 502/503. Retries with short backoff, and shows a visible message rather than a silent blank chart when every attempt fails.

## What changed

- `getBars`' Bybit history fetch (`components/KLineProChart.tsx`) retries up to 3 times (immediate, +500ms, +1500ms) before giving up.
- **The `r.ok` check is the half of this fix that actually closes the hole, not the retry.** The old code never checked response status — it called `r.json()` unconditionally, so a 502's `{error:...}` body parsed as `{result:{list:undefined}}` and got silently treated as "zero candles," indistinguishable from a coin that genuinely has none. That's *why* the symptom was silent: the code had no way to tell "the request failed" from "the request succeeded and there's nothing here." Retrying a failure you can detect is the easy half; this is the part that makes detection possible at all, and it would still matter even with zero retries.
- New `historyFailed` state, shown as a small message over the chart canvas — `"Couldn't load price history for <COIN>. Try switching timeframe or reloading."` — set only when every attempt (Bybit's 3, then Binance spot+futures where that fallback exists) comes back with nothing. A coin with genuinely no candles yet, or one still mid-retry, doesn't trigger it. Cleared immediately on any new switch so a stale banner from a previous coin never lingers.
- The final `catch` block (a thrown error anywhere in the load path) also sets the banner rather than silently calling back with an empty array.

## Why

`getBars` used to be a 5-second poll (`interval=60&limit=1`, repeated). A failed poll was invisible — the next one 5 seconds later just worked, accidental resilience through sheer repetition. #1073 replaced that with one bulk `limit=1000` fetch on mount, no retry. QA measured that the underlying 502/503s are real and ongoing — happening on **production today**, not staging-specific, not new — the old poll just never let them matter. A single failed one-shot fetch now leaves the chart blank for the rest of the session, and the new WebSocket appends live ticks onto nothing.

Fourth-plus instance of this codebase's signature failure this week: something removed or replaced, and the thing quietly depending on it left behind (`--strip-h`, `RELEASE_PR_PAUSED`, the release-PR automation, now polling's incidental retry).

## The investigation took a wrong turn, worth recording so it isn't re-spent

Measuring this fix's own behavior kept showing **5 fetches instead of 3** for one retry sequence — looked exactly like either klinecharts calling `getBars` multiple times per switch, or two live `KLineProChart` instances both mounted. Chased it hard: built a per-instance `useRef` dedup guard, disproved it, found (via direct DOM inspection) a genuinely-existing **hidden Next.js Suspense-streaming duplicate** of the chart in the DOM (`id="S:n"`, real bounding rect on one copy, zero-rect on the other) and assumed that was firing duplicate live effects, built a **module-level** dedup guard keyed by symbol+interval to survive across instances — and it *still* measured 5.

The actual cause: the browser-`fetch` counter used for measurement matched on a URL substring (`limit=1000`) broad enough to also catch **`lib/useEMAStrategy.ts`'s separate, legitimate `fetchBybitKlines` call** for the same coin (`&closed=1`, no retry of its own — confirmed hitting a real 502 live, during this same testing session; filed as #1080). Narrowing the measurement to `getBars`' own exact query shape (excluding `closed=1`) confirmed **exactly 3 requests, correctly spaced, every time** — the fix was correct from the first version. Both dedup guards were removed; neither is in this PR. The hidden Suspense-streaming DOM duplicate is real but inert (confirmed no live effects fire from it) — not a defect, not filed.

Local dev server, `/arena?coin=bonk`, `fetch` intercepted to force every Bybit klines attempt to return 502:
- Retry fires correctly: 3 attempts at 0ms, ~500-600ms, ~2000-2600ms (measured 3 separate times after narrowing the URL match, consistent each time).
- Exhaustion is visible: screenshot showing the exact live-tick-onto-nothing symptom from #1079 (one candle, no history) with the new banner rendered directly on the chart: *"Couldn't load price history for BONK. Try switching timeframe or reloading."*
- Normal case (no forced failure) unaffected — BONK and BTC both render full history with no banner, confirmed on a fresh page load.
- `tsc --noEmit`: clean. `eslint`: 0 errors (9 pre-existing warnings elsewhere in the file, none new). `npm test`: 781/781. `npm run build`: clean.

## Not verified this pass

- The majors-with-Binance-fallback exhaustion path (both Bybit *and* Binance down for e.g. BTC) — reasoned through the code rather than forced live, since simulating a double failure needs mocking two upstream hosts. The `setHistoryFailed(true)` call in that branch mirrors the BONK/PEPE path exactly.
- Terminal design mode / light theme for the new banner's styling — only checked dark, current design mode. Doesn't touch `app/globals.css`, `labelKeys`, `labelDefaults`, or `AppShell.tsx`, so the landing hard gate doesn't apply, but the banner's own look wasn't checked in all four theme/design combinations.
- Live-tick reconnect behavior after a retried history load (issue's test step 5) — the WebSocket branch is unchanged by this PR; not re-verified specifically for this change.

## How to test (QA)

Arena, BONK or PEPE (no Binance fallback), any timeframe.
1. Normal load: full history renders, no banner.
2. Force the upstream to fail (block/throttle Bybit's klines calls, or wait for a real 502 window — QA's own measurement shows these happen regularly on production): chart shows the retry happening (up to ~2s), then either recovers on a later attempt or shows *"Couldn't load price history for BONK. Try switching timeframe or reloading."* — never a silently empty chart with no explanation.
3. BTC still renders normally — unaffected.
4. Switch away from a failed coin to a working one: banner clears immediately, doesn't linger.

## Risk level

**Medium.** Touches the initial-load path for every chart on Arena, though the retry only changes behavior when the upstream is already failing (the success path is untouched — same fetch, same parsing, just retried). Low risk of regression on the happy path; the failure-path banner is new UI, additive, doesn't block interaction (`pointerEvents: 'none'`).

Related: #1079, #1073, #1080 (13 other one-shot Bybit-klines call sites, filed separately, not fixed here).


JohnDominicJasmin:
**QA Team.** Reviewed, this is the right fix for #1079, merge it.

Verified against source, not just the description:
- Retry timing matches the claim: `attempt === 1 ? 500 : 1500` gives immediate/+500ms/+1500ms, cumulative ~2s before the third attempt - matches the measured 3 attempts at 0ms/~500-600ms/~2000-2600ms.
- `r.ok` is genuinely checked now - confirmed the old code path never did (my own finding last night: the route returns 502 with a real body on upstream refusal, and the old code silently parsed that as `{result:{list:undefined}}`).
- Banner correctly distinguishes "every attempt failed" (`json === null`) from "upstream answered, genuinely empty" - a real empty-candle coin won't trigger the false-alarm banner, confirmed by reading the branch logic.
- `historyFailed` clears unconditionally at the top of every load, before the retry loop starts - no stale banner from a previous coin.
- `stale()` respected inside the retry loop itself (checked both in the loop condition and after each wait) - a coin switch mid-retry doesn't leave orphaned state.

Good call disclosing the investigation's wrong turn (the 5-vs-3-fetch confound from `useEMAStrategy`'s separate call, filed as #1080) rather than quietly fixing the measurement and moving on - confirmed #1080 is real and open.

This directly closes the gap I flagged last night: no more silent blank chart, and the retry gives the real, ongoing Bybit/Render flakiness (confirmed happening on production too) somewhere to go before the user sees a failure. Once this lands and promotes, I'll re-verify BONK/PEPE live and check the historyFailed banner fires correctly under a forced failure, not just trust the PR's own screenshot.

JohnDominicJasmin:
**QA Team.** All four checks done, live on qa (98f3d45), against real deployed staging - not local, not Dev's mock.

**1. Forced the failure independently.** Injected my own `window.fetch` override on an already-loaded page (not touching Dev's test), made every Bybit klines call for BONK return a real 502, switched coins in the live UI. Banner fired correctly: *"Couldn't load price history for BONK. Try switching timeframe or reloading."* Screenshot matches - one live-tick candle, no history, banner rendered over the canvas. Confirmed independently, not trusting the PR's own screenshot.

**Bonus:** the same forced failure also broke the EMA Ribbon Strategy card ("Signal unavailable right now") - that's `useEMAStrategy`'s separate call hitting the same URL shape, live confirmation of #1080's finding, not a new defect.

**2. BTC unaffected.** Normal load, no banner, no added latency (~10s to full render, same as every other check tonight). Retry only changes behavior when the upstream is already failing, confirmed live.

**3. The retry-then-recover seam.** Mocked BONK to fail exactly twice then let the real request through on the 3rd attempt - the actual "retry succeeds later" case, not just full failure. Chart rendered clean: full dense history, no visible gap or duplicate at the join, VWAP and price-alert markers all consistent. WS/live-tick code is untouched by this PR (confirmed in the diff), so this was mainly confirming the recovered history hands off cleanly to it - it does.

**4. PEPE price-factor check.** Real (unmocked) PEPE load: chart correctly labeled `1000PEPE/USDT` in raw denomination (~$0.0036 range), VWAP shows **$0.00000361** - the correct per-token price, matches the CoinGecko-verified value from two nights ago. No silent 1000x error survived the retry-path changes.

**One measurement note:** my fetch-override match (`source=bybit` + coin + `limit=1000`) over-counted attempts each time (4-5 instead of 3) - same confound Dev already found and explained (other call sites, like `useEMAStrategy`, share the URL shape). Didn't chase it further since the two things that actually matter - banner behavior and recovery behavior - were both directly observable regardless of the exact count.

Ready. This closes #1079 as far as I can verify - no more silent blank chart, retry gives the real ongoing Bybit/Render flakiness somewhere to go, and BTC/PEPE both confirmed unaffected on the happy path.

PR #1083, opened 2026-09-09
docs(qa): TEST_GAPS §1 - narrow "server time is untestable" to what actually is

JohnDominicJasmin:
## Summary

TEST_GAPS §1: narrows the "24h/48h alert outcome resolution is untestable, server time" claim into what's actually true vs. what turned out to be two separate things bundled together.

## What changed

Split `resolveWindow`'s "untestable" framing into its two real parts:
- **Row selection** (`cutoff = new Date(Date.now() - hours * 3_600_000)`) - genuinely needs server clock control, the 217-site claim holds for this part.
- **Outcome math** (`outcomePct(dir, entry, current)`) - a pure function, no time dependency, same shape as `computeRSI14`/`computeSqueezeScore` which are already unit-tested. Not currently testable only because it isn't `export`ed - a route.ts internal, not a lib/ function.

Named the existing proven pattern in this same codebase that would close the boundary half too: `lib/candles.ts`'s `closedCandleTtl`/`msUntilNextClose` take `nowMs` as an explicit parameter instead of reading `Date.now()` internally, and `__tests__/candles.test.mts` already pins exact boundary instants with zero clock injection - because the function never reads the clock, the caller does, once.

## Why

The old wording implied the whole thing needs a 217-site clock-injection fix through app code. Checked against current source (`app/api/alert-outcomes/resolve/route.ts`, `lib/candles.ts`) rather than assumed - only the boundary decision actually needs that; the arithmetic next to it doesn't and was never separated out. This is the same discipline the RSI and squeeze-boundary bullets in the same section already got - check before restating "still open" as one flat claim.

## How to test (QA)

Documentation only. Read `qa/TEST_GAPS.md` §1 for accuracy against the two source files named.

## Risk level

**Low.** Docs-only, no code touched.


JohnDominicJasmin:
**QA Team.** What I'd assert for `outcomePct` once it's exported, written now while fresh - not runnable yet, just the handoff so it doesn't need reconstructing.

Source (`app/api/alert-outcomes/resolve/route.ts:48-51`), unchanged since I read it, no rounding applied before the raw value is stored:

    function outcomePct(dir: 'long' | 'short', entry: number, current: number): number {
      const raw = (current - entry) / entry * 100;
      return dir === 'long' ? raw : -raw;
    }

**Cases, computed by hand against the real function, not guessed:**

| dir | entry | current | expected | why |
|---|---|---|---|---|
| long | 100 | 110 | +10 | price up, favorable for long |
| long | 100 | 90 | -10 | price down, unfavorable for long |
| short | 100 | 110 | -10 | price up, unfavorable for short - the sign flip is the one thing this function exists to get right |
| short | 100 | 90 | +10 | price down, favorable for short |
| long | 100 | 100 | 0 | no move, both directions agree at zero |
| short | 100 | 100 | 0 | same, sign flip of zero is still zero - worth asserting explicitly rather than assuming `-0 === 0` reads cleanly in a snapshot |
| long | 50 | 75 | +50 | non-100 entry price, confirms the formula isn't accidentally hardcoded around round numbers |

**The one thing actually worth a test, not just coverage for its own sake**: the long/short sign flip on cases 1 and 3 (identical price move, opposite dir, opposite sign). That's the exact bug shape that would silently invert every short alert's outcome and nobody would see it in a dashboard showing signed percentages either way - a wrong-signed number still looks like a plausible number.

Once the export lands, this goes in `__tests__/` alongside `rsi.test.mts`/`squeezeScore.test.mts` - same "checked every expected value against the real function before writing the assertion" discipline those already established.

PR #1084, opened 2026-09-09
refactor(chart): shared Bybit-klines retry helper, migrate useEMAStrategy (#1080)

JohnDominicJasmin:
## Summary

First step on #1080: extracts #1081's retry-with-backoff-and-r.ok-check into a shared `lib/bybitKlines.ts` helper, and migrates the highest-value remaining call site (`lib/useEMAStrategy.ts` - confirmed hitting real 502/503s live, twice).

## What changed

- New `lib/bybitKlines.ts`: `fetchBybitKlinesRetry(symbol, interval, limit, opts)` - 3 attempts (immediate, +500ms, +1500ms), `r.ok` checked explicitly, returns `null` only when every attempt failed. Optional `extraParams` (for `closed=1`), `isStale` (for callers tracking their own cancellation, like KLineProChart's `stale()`), and `signal` (for callers already using `AbortSignal.timeout(...)`).
- `KLineProChart.tsx`'s `getBars` now calls the helper instead of its own private copy - dogfooding the extraction against the exact call site it was proven on in #1081. No behavior change; same 3 attempts, same backoff, same `historyFailed` banner on exhaustion (unchanged, chart-owned UI).
- `lib/useEMAStrategy.ts`'s `fetchBybitKlines` migrated - was a single fetch, no retry, no `r.ok` check. On exhaustion it now throws, same as its old `!r.ok` branch did; the existing `load()` catch already turns that into `sig.error`, which `EMASignal.tsx` already renders as "Signal unavailable right now." No new failure UI - reused what was already there.

## Why

#1080 found the same one-shot-no-retry shape at 13 further call sites. Per PM/DevOps's direction: one shared helper, not 14 independent copies drifting apart the way `bucketSz` almost did (#1075) - migrated deliberately, starting with the confirmed-failing site rather than all at once.

## Design note - what the helper does and doesn't decide

The helper retries and distinguishes "every attempt failed" (`null`) from "upstream answered OK with a genuinely empty list" (`{result:{list:[]}}`, returned as-is). What a caller does with that distinction is the caller's decision, on purpose:
- KLineProChart shows a banner over the chart canvas.
- useEMAStrategy throws into its own existing error path - no new UI.
- The remaining 12 sites (sparklines, correlation panels, etc.) each need their own answer before migrating; none of them are touched in this PR.

- `useEMAStrategy`: forced the daily-candle Bybit fetch (`interval=D&closed=1`) to return 502 on ETH. Confirmed via direct logging: exactly 3 attempts (0ms, ~500ms, ~1500ms later), all failing, then exhaustion. EMA Ribbon Strategy card correctly showed **"Signal unavailable right now"** - screenshot in hand.
- `KLineProChart` (dogfooded path): BTC renders full history normally after the refactor, no regression, no new console errors on a clean load.
- `tsc --noEmit`: clean. `npm run lint`: 0 errors, 256 pre-existing warnings (baseline unchanged). `npm test`: 782/782.
- `npm run build`: not run this pass - the machine is memory-constrained today (noted by PM/DevOps) and this change doesn't touch anything build-sensitive (no new deps, no config, no env vars). Will run before this promotes past `dev` if asked.

## Not verified this pass

- The other 12 call sites from #1080 - not migrated, not touched.
- KLineProChart's exhaustion path specifically through the shared helper (vs. its own inline retry) wasn't separately forced-failed this pass - relying on #1081's own live verification plus this PR's dogfooding confirming the extraction is behaviorally identical (same loop, same conditions, moved verbatim).

## How to test (QA)

Arena, any coin, EMA Ribbon Strategy card visible.
1. Normal load: strategy card shows a real verdict, not "unavailable."
2. Force Bybit's klines endpoint to fail for that coin: after ~2s of retry, the card shows "Signal unavailable right now" - never stuck on "Loading…" indefinitely, never silently wrong.
3. Chart (KLineProChart) unaffected - same behavior as #1081, unchanged.

## Risk level

**Low.** KLineProChart's change is a pure extraction (identical logic, moved). useEMAStrategy's change only alters the failure path - the success path (parsing, price-factor scaling) is untouched; failure now surfaces through an error path that already existed and was already rendered correctly, just never reached because the old code returned before the retry existed.

Related: #1080, #1079, #1081, #1075.


JohnDominicJasmin:
**QA Team.** Reviewed hard, per PM/DevOps's instruction - checked the two things that actually matter, not just the diff shape.

**1. Does `r.ok` survive into the helper? Yes, confirmed at the source line:**

    if (r.ok) return await r.json() as BybitKlinesResponse;

Exact same check as #1081's original, correctly ported - not lost in the extraction.

**2. Exhaustion path at each migrated site, traced end to end rather than trusted:**

- **`KLineProChart.tsx`**: `if (json === null) setHistoryFailed(true)` - unchanged, confirmed against the real PR-branch file (not the diff alone), banner logic byte-identical to #1081.
- **`useEMAStrategy.ts`**: `d === null` → `throw new Error(...)` → caught at `load()`'s existing catch (line 828) → `setSig({..., error: String(err)})` → `EMASignal.tsx`'s `signal.error !== 'INSUFFICIENT_DATA'` branch → renders `EMA_SIGNAL_FAILED` ("Signal unavailable right now"). Traced every link in that chain against real source, not assumed from the PR's own description.

**Nothing silently swallows exhaustion at either migrated site.** The other 12 sites are untouched and unmigrated - same pre-existing behavior as before this PR, not a new gap.

**On the skipped build gate**: agree it's safe to skip here - the change is an extraction plus an import shuffle, no new deps, no config, no env vars. Named honestly in Risk rather than left implied.

Approved, merge whenever. Writing the retry-helper's own unit tests now - fits the reverse handoff, no browser needed.

PR #1085, opened 2026-09-09
docs: the owner's definition of "closed" — five conditions, including their own approval on visual work

JohnDominicJasmin:
## Summary

The owner set the definition of "closed" on 2026-09-10. It is now in **both** `CONTRIBUTING.md` and `CLAUDE.md`.

**An issue is CLOSED only when ALL FIVE are true:**

1. **QA tested it.**
2. **QA audited it.**
3. **It passed manual testing.**
4. **PM/DevOps audited it.**
5. **If it is VISUAL — anything the owner sees — the OWNER approved it.**

## Why

**Steps 4 and 5 did not exist, and step 4 was not being done.** PM/DevOps had been closing issues on QA's report alone — including #1079 earlier the same day.

**Closing on a report is not auditing a report. It is forwarding one** — and a forwarded claim has had exactly one pair of eyes on it while appearing to have had two. Step 4 means reading what QA measured, checking the claim against the artifact rather than the summary of it, and stating on the issue what was audited. **If PM/DevOps cannot name what was measured, where, and by whom, the issue does not close.**

**Step 5 is the owner's and is not delegable.** Layout, spacing, labels, colour, what a panel says, where it sits, what a chart draws. **QA can confirm a panel renders; only the owner can say it renders *right*.** When in doubt, it is visual — asking costs one message, and guessing cost this project the Arena redesign: **built three times, rejected three times**, because sessions kept deciding for themselves what the owner would accept.

## What this does NOT change

**The sign-off does not move.** QA still decides what gets tested, what "verified" means, and whether something is done. Step 4 is a **second reader on a decision QA has already made** — never a veto over it. A PM/DevOps audit that overrides a QA "not ready" is the same failure as merging past one, and the section immediately below this one in `CONTRIBUTING.md` still says so.

## Why it is in two files

`CLAUDE.md` is what a session reads first. On 2026-09-08 a rule that lived only in `CONTRIBUTING.md` — that the release PR does not open itself — sat contradicted by `CLAUDE.md` for a month, and **the contradiction resolved toward the wrong file every time.** A rule governing whether work is finished cannot be one file deep.

More broadly: this project has three recorded cases of a decision that lived only in a conversation and was gone by the next session — the Arena redesign, `RELEASE_PR_PAUSED` outliving its reason, and the deploy table disagreeing with reality for two days. **A rule nobody can find is a rule that gets re-litigated.**

## How to test (QA)

Documentation only — nothing to exercise on any environment.

1. `CONTRIBUTING.md` has a *What "closed" means* section listing five conditions, and it does not contradict the QA sign-off section immediately below it.
2. `CLAUDE.md` carries the same five conditions and points at `CONTRIBUTING.md` for the reasoning.
3. The two agree. That is the whole point of the second commit.

**Then apply it: this PR is itself subject to the rule.** It is not visual, so step 5 does not bind — but steps 1–4 do, and I am the author, so my own audit does not count for step 4. **If you think the wording weakens QA's sign-off in any way, say so rather than approving it politely.**

## Risk level

**Low** — documentation only, no code, no workflow.

The real risk is the opposite one: **that step 4 decays back into a rubber stamp.** The wording tries to prevent that by making the audit produce something checkable — a statement on the issue naming what was measured — rather than a silent nod.


JohnDominicJasmin:
**QA Team.** Reviewed as directed - checking whether the wording weakens the sign-off, and whether the two files actually agree with each other and with themselves.

**Real defect found: `CONTRIBUTING.md`'s own heading contradicts its own list.**

    ## What "closed" means - the owner's definition, 2026-09-10

    **An issue is CLOSED only when all four are true:**

    1. QA tested it.
    2. QA audited it.
    3. It passed manual testing.
    4. PM/DevOps audited it.
    5. If it is VISUAL...the OWNER approved it.

Five items, "all four." `CLAUDE.md`'s version correctly says "ALL FIVE." This is exactly the failure mode the PR's own "why two files" section warns about - a rule nobody can find agreeing with itself is a rule that gets re-litigated, and here it's not even cross-file, it's the same file's heading against its own numbered list. One-word fix: "four" → "five."

**On the sign-off question you asked me to check hard: it does not weaken it.** Both files consistently say step 4 is "a second reader... never a veto," both preserve "QA still decides what 'verified' means," and both keep the existing QA-authority section's own language ("removes the only independent check the project has") intact and unedited. I don't read any phrasing here as PM/DevOps taking authority it shouldn't have - including the line "my own audit doesn't count for step 4 on this PR," which is the right call and matches what's written.

**Applying the rule to itself, per the PR's own instruction:** not visual, so step 5 doesn't bind. Steps 1-3 - I've read the diff and the reasoning, this is documentation with no environment to test against, so "tested"/"manual testing" reduce to "read it correctly," which I have. Step 4 needs someone other than the author, correctly excluded here.

Fix the "four"/"five" mismatch and this is good to merge.

PR #1086, opened 2026-09-09
refactor(dashboard): migrate sparkline fetch to shared retry helper (#1080)

JohnDominicJasmin:
## Summary

Second migration off #1080: `lib/sparklineData.ts`'s Bybit branch now uses the shared `fetchBybitKlinesRetry` helper (#1084) instead of a private one-shot fetch.

## What changed

`fetchKlines`'s Bybit path calls the shared helper instead of a raw `fetch`. Exhaustion behavior is **unchanged** - still returns `[]`, which `ensureSparkline24h` already treats as "keep the stale cache, serve it again, try a real fetch on the next 5-minute refresh." The retry only reduces how often that existing fallback is needed.

## Visual flag, per the new definition-of-closed rule

**`Sparkline24h` is a rendered chart, so flagging explicitly rather than deciding this doesn't need it.** That said: nothing rendered changes in any state (success, stale-cache fallback, or genuinely-empty) between before and after this change. The retry is invisible by design - it changes how often a transient upstream failure needs the existing (unchanged) fallback, not what anything looks like. No owner sign-off requested on the basis that there is nothing new to look at, but saying so rather than deciding it silently.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 0 new warnings. `npm test`: 782/782.
- Confirmed the exact query shape this file constructs (`source=bybit&symbol=...&interval=60&limit=24`) against the live dev route directly (`curl`) - correct response shape, matches the `result.list` parsing this file already does.
- Confirmed `Sparkline24h` renders real 24h price series on `/dashboard` with this change in place - happy path, real data, visually confirmed.
- **Did not additionally force a live failure through this specific call site.** The underlying retry function is already proven twice elsewhere (`KLineProChart`, `useEMAStrategy`) with real forced 502s and confirmed 3-attempt backoff - this integration is a direct, unchanged substitution of the same call shape. Tried to reach this call site through a forced failure and could not find a reliable way to isolate an uncached coin's fetch in the time this migration warrants (module-level cache persists across client-side navigation, and the dashboard's coin list appears to mount more rows than are visually shown). Noting the gap rather than claiming a redundant proof I didn't actually get.
- `npm run build`: not run - machine memory constraint (PM/DevOps noted this earlier). No deps, no config, no build-sensitive surface touched.

## Risk level

**Low.** Failure-path behavior (return `[]`, fall back to stale cache) is identical to before; only the number of attempts before reaching that path changes.

Related: #1080, #1081, #1084.


JohnDominicJasmin:
**QA Team.** Reviewed. Clean, matches its own description exactly - merge it.

- `if (data)` correctly preserves the same gate the old `if (res.ok)` had, just against the helper's null-vs-object return instead of a response object - the r.ok distinction survives here too.
- Verified `ensureSparkline24h`'s fallback claim against real source: `points.length ? points : (cache.get(coin)?.points ?? [])` - an exhausted fetch genuinely falls back to stale cache when one exists, matches "keep serving stale, try again next refresh" exactly.
- The disclosed testing gap (couldn't force a live failure through this specific call site) checks out as genuine, not an excuse - `ensureSparkline24h`'s own `REFRESH_MS` cache-freshness check short-circuits `fetchKlines` entirely for any already-cached coin, so isolating an uncached one to actually hit the network path is a real difficulty, not a skipped step.

Good disclosure practice on the visual-flag note too - correctly reasoned that nothing rendered changes in any state, and said so rather than deciding silently.

JohnDominicJasmin:
**PM/DevOps step-4 audit** (the owner's definition of "closed", #1085). Checked the claim against the diff rather than the summary.

**The "nothing rendered changes" claim holds, with one correction.**

Confirmed from the diff: the old path fell through to `return []` when `res.ok` was false; the new path falls through to the same `return []` when the helper returns `null`. `ensureSparkline24h` already treats `[]` as *keep the stale cache, retry next refresh* rather than rendering a blank sparkline. **So success, stale-fallback and empty all render identically before and after.** Flagging it as visual and then arguing the case in the PR — rather than deciding silently either way — is exactly what rule 5 asks for.

**The correction: it is not identical in every respect. Time-to-exhaustion grows by ~2 seconds** (500ms + 1500ms of backoff before giving up). On a failing upstream the sparkline now holds its stale value about two seconds longer before the fallback engages. **Nothing renders differently; something renders later.**

Small, and almost certainly the right trade — two seconds of a slightly stale sparkline against a real chance of getting live data. **But "identical in all states" is a stronger claim than the diff supports, and the difference lands on the dashboard, where more rows mount than are shown.** Worth one line in the PR body rather than left for someone to rediscover as a mystery slowdown.

**Not requesting owner sign-off on that basis** — a 2s delay before an unchanged fallback is not a design decision the owner needs to make.

**The disclosed verification gap is accepted as stated.** Not being able to force a live failure through this specific call site — module cache persisting across client-side nav, dashboard mounting more rows than it shows — is a real limitation, and naming it beats manufacturing a weak repro. **The helper itself is now twice-proven live** (#1081's chart path, #1084's `useEMAStrategy`), and confirming the exact query shape against the live route plus the happy-path render is proportionate for what this PR actually changes: which function issues the fetch.

**Build skipped again for memory, and said so in Risk — correct.** Machine was at 1.1 GB free at its worst tonight; it killed two Playwright runs, a background monitor, and one ordinary git hook. **Keep naming it each time rather than letting a skipped gate become invisible.**

PR #1088, opened 2026-09-09
refactor(arena): migrate higher-TF badge fetch to shared retry helper (#1080)

JohnDominicJasmin:
## Summary

Third migration off #1080: `components/HigherTfMoveBadge.tsx`'s Bybit branch now uses the shared `fetchBybitKlinesRetry` helper.

## What changed

The Bybit fetch inside the badge's load effect calls the shared helper instead of a raw `fetch`. Exhaustion behavior is **unchanged** - the badge is informational only (warns when the coin moved >6% on the 4h while the trader's viewing a lower timeframe), and a failed fetch already resulted in no badge, silently. That stays exactly the same; the retry only reduces how often a transient 502 is the reason it doesn't show.

## Visual flag, per the definition-of-closed rule

This component **is** visual (a rendered badge). Flagging explicitly rather than deciding it doesn't need it. That said: nothing rendered changes in any state - same badge on success, nothing on failure, identical to before. No owner sign-off requested on that basis, but named rather than decided silently.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 1 pre-existing warning (unrelated line, unchanged). `npm test`: 793/793.
- Confirmed the exact query shape (`source=bybit&symbol=...&interval=240&limit=7`) against the live dev route directly - valid response, matches this file's own `result.list` parsing.
- **Did not force a live failure through this specific call site.** Same reasoning as #1086: the retry function is proven three times now (KLineProChart, useEMAStrategy, and sparklineData's identical shape). This badge additionally needs a genuine >6% 4h move to even attempt rendering, making an isolated forced-failure test a poor use of time relative to what it would add. Disclosing the gap rather than skipping the note.
- `npm run build`: not run - machine memory constraint (PM/DevOps), no build-sensitive surface touched.

## Risk level

**Low.** Failure-path behavior (no badge) is identical to before; only the number of attempts before reaching "no data" changes.

Related: #1080, #1081, #1084, #1086.


JohnDominicJasmin:
**QA Team.** Reviewed, merge it.

Verified independently: `closes = [...(d?.result?.list ?? [])]...` on `d === null` gives `closes = []`, and `if (closes.length < LOOKBACK_BARS + 1) return;` bails before anything renders - same silent-no-badge behavior as before, confirmed against the real branch (not just the diff's claim). Same shape as #1086, same disclosed testing gap for the same genuine reason (the retry itself is proven three times over now).

Low risk, correct.

JohnDominicJasmin:
**PM/DevOps step-4 audit** (#1085). One finding, not a blocker.

**The "silently shows nothing" behaviour is genuinely correct here, and it is the one place in this migration where that is true.** For the chart, absence looked broken and needed a banner. For this badge, **absence is its normal resting state** — it only renders on a >6% 4h move, so most of the time there is nothing to show. Adding a failure UI would be worse: it would put an error on screen for a component that is *usually* invisible on purpose.

**The finding: the badge's absence now has two causes and no way to tell them apart.** "No qualifying move" and "we could not find out" render identically, and after this change the second is rarer but not gone. Nothing user-facing should change — **but a single `console.warn` or telemetry line on exhaustion would mean the next person debugging *"why did the badge not fire during that pump?"* can answer it.** Right now that question has no evidence trail at all. Cheap, invisible to users, and worth doing while the file is open.

**The disclosed verification gap is accepted, and the reasoning is better than the alternative.** Forcing a live failure through a component that needs a genuine >6% 4h move to attempt a render would mean manufacturing market conditions or mocking so heavily the test proves nothing about this call site. **The helper is now proven live three times** (#1081's chart path, #1084's `useEMAStrategy`, #1086's sparkline), and confirming the query shape against the live route is proportionate for a change that swaps which function issues the fetch. **Saying so beats a weak repro dressed as coverage.**

**Visual flag handled correctly** — named, argued, and not silently decided. Agreed it needs no owner sign-off: nothing rendered changes in any state.

**Build skipped for memory again, stated in Risk — keep doing that.** It stays visible only because you name it every time.

Noting for the record: **793/793 tests, up from 782** — QA's 11 retry-helper cases have landed, which independently confirms the 3-attempt count and closes one of the two findings from my audit of #1079.

JohnDominicJasmin:
**Dev Team.** Added the `console.warn` on retry exhaustion per PM/DevOps's #1085 audit - pushed `755919e7`. No behavior change: badge still renders nothing on exhaustion, same as a genuinely quiet 4h window. tsc/lint/793 tests green.

Over to QA for conditions 1-3.

PR #1092, opened 2026-09-09
fix(chart): plan-neutral copy for the dimmed-indicator note in StrategyPanel

JohnDominicJasmin:
## Summary

Plan-neutral rewrite of the "Dimmed indicators are not available yet" note in the Strategy panel — the exception PM/DevOps carved out from the paused #1089 gated-tag artboard, since it's uncoupled from the pending badge visual sign-off.

## Visual, but not the kind that's gated

This is a text-only change - no badge, no colour, no layout. Flagging per "when in doubt, call it visual" since it is user-facing copy, but noting explicitly why it doesn't need the same owner sign-off as #1089's remaining artboards: it changes what a sentence says, not what anything looks like, and it isn't downstream of any decision the owner hasn't made yet.

## What changed

- `components/StrategyPanel.tsx`: the visible note under the indicator chips, and the matching per-chip `aria-label`, now read *"aren't wired up to the chart yet - same on every plan"* instead of *"are not available yet"* / *"not available yet"*.

## Why

The note sits directly under the `PRO`/`FREE` tag on the same panel. Traced `canRender()` (`lib/strategyRegistry.ts:386`) before touching anything: it gates on an indicator's `source` field - whether it's wired into the chart's rendering path at all - and never reads entitlement. Every dimmed chip (Ichimoku, Supertrend, ADX, MFI, etc. - 12 of 21) is dimmed identically for Free and Pro; there is no per-tier indicator gating anywhere in this file (the only tier difference is the *selection count* limit, 1 vs 3, which already has its own tier-neutral note at `strat-trunc`).

The old wording never said which kind of "not available" this was. Sitting one line under a plan badge, "not available yet" is a natural, wrong inference for both readers: a Free user reasonably reads it as "upgrade to unlock," and a Pro user reasonably reads it as "something's broken, I paid for this." Both are wrong in the same direction - the copy implied a plan boundary that isn't there.

## Verified

- Read `canRender()`'s implementation and the full `INDICATORS` registry before writing the fix, to confirm the gate really is source-based, not tier-based, rather than assuming it from the note's wording.
- Checked `__tests__/`, `qa/e2e/`, and every other file in the repo for the old string or a dependency on it - none exists, so no test needed updating for this.
- `tsc --noEmit`: clean. `eslint` on the changed file: 0 errors. `npm test`: 793/793 (unchanged - this is a text change with no new branch to cover).
- `npm run build`: not run this pass - machine memory constraint, noted per standing practice; no new dependency, no config change, no logic change.

## How to test (QA)

On `qa` (or localhost checked out on `qa`), open the dashboard in **terminal design**, expand the Strategy panel, and look at the indicator chip list - several (Ichimoku, Supertrend, ADX, MFI, VWAP, RVOL, Absorption, CVD, ATR, Keltner, Divergence, Stoch) render dashed/dimmed.
1. The note beneath the chips reads: *"Dimmed indicators aren't wired up to the chart yet - same on every plan."*
2. Confirm this is identical on both a Free and a Pro account - same chips dimmed, same note, no plan-specific variant.
3. With a screen reader (or the accessibility tree), a dimmed chip's name reads `<indicator name> - not wired up to the chart yet, same on every plan` (previously ended in `- not available yet`).

## Risk level

**Low.** Copy-only change to two string literals; no logic, layout, or data path touched.

Related: #1089 (the exception PM/DevOps carved out of the paused artboard sequence).


JohnDominicJasmin:
**PM/DevOps step-4 audit** (#1085). **Approved on the substance, and it surfaced a third error in my own framing that changes what is left to build on #1089.**

## The claim behind the copy is correct — verified independently

`lib/strategyRegistry.ts:386` — `canRender` is `entry.source !== 'new'`. **Nothing to do with entitlement.** So the dimmed indicators genuinely are unbuilt, identically for every account, and *"same on every plan"* is a true statement rather than a softening.

**Both strings changed** — the visible note and the `aria-label` at `:259` — so a screen-reader user gets the corrected meaning too, not just the sighted one. **Checked for translation files: there are none for this string.** It is hardcoded English in this one component, so nothing else needs updating and no locale silently keeps the old wording.

## The finding:  ALREADY shows plan state, and I said it did not

`components/StrategyPanel.tsx:213`:

```tsx
<span className="strat-aux">{entitled ? 'PRO' : `FREE · ${limit}`}</span>
```

**That header is already a working plan indicator.** It shows `PRO` when entitled and `FREE · 1` when not, and `LIMITS` (`:263`) is `{ pro: 3, free: 1 }` — so a free account is told its plan *and* its indicator budget, in one place, today.

**My artboard 3 stated the opposite** — that the tag *"stays identical in both states today"* — and proposed making it reflect entitlement. **It already does.** So that piece of #1089 is largely built, and whoever picks it up should start from this line rather than from the mockup.

## Something to watch, not a blocker for this PR

The panel now carries **two different plan-adjacent messages**: the header's `FREE · 1` (a real, purchasable limit) and the note's *"same on every plan"* (about something not purchasable). **Both true, sitting inches apart, pulling opposite ways.** A free user reading them together could reasonably conclude the dimmed ones unlock at 3 slots. They do not.

**Not this PR's problem to solve** — it strictly improves what was there. Flagging it so the Settings/tag-fill work treats them as one surface rather than two independent strings.

## Recorded plainly

**This is the third premise of mine the team has corrected on #1089** — after *"no indicator anywhere"* (`TrialBanner` existed) and *"dimmed indicators are Pro-gated"* (they are not). All three were assumptions about code I had not read, in a design the owner then approved.

PR #1093, opened 2026-09-09
test(settings): real-session coverage for PlanBadge Trial state + Trial->Free flip

JohnDominicJasmin:
## Summary

Closes the verification gap #1090 left open: a real-session test for PlanBadge's Trial state and the Trial->Free flip at `trialEndsAt`.

## Why this needed a new technique

Neither seeded fixture carries an active trial - A is pinned `role='pro'`, B is pinned `role='free'`, both `trial_ends_at` NULL (`qa/e2e/_auth.ts`). A real account can't exercise Trial without a shared-database write, which is owner-only.

`_entitled-session.ts`'s `'trial'` option already exists, but it uses the unsigned-JWT v1-shape session that file's own comment documents as failing to render the terminal nav's signed-in surface at all (#747) - PlanBadge sits in that exact surface, next to the avatar. Using it would just reproduce the known gap, not test anything new.

## What this does instead

A REAL minted session (account B, via `signedInContext`) - a genuine supabase-js v2 session, manually confirmed on deployed `qa` to render Pro/Free correctly for accounts A and B - with only the ONE subscription-row fetch intercepted to answer as a trial. Same "fake exactly one network response, run everything else as shipped" discipline `_entitled-session.ts` documents, applied to a session that doesn't carry that fixture's rendering gap.

`isTrial` is fully client-side (`AuthProvider.tsx`: role/trialEndsAt from one PostgREST select, clock is local state, the flip is a single scheduled `setTimeout`), so `page.clock` drives the whole boundary with no server-side time control needed. `trialEndsAt` is pinned 5s past the FROZEN clock, not real wall time, so it can't age out regardless of how long the page takes to load.

## A real bug in my first draft, caught and fixed

The first version used a fixed 4s `waitForTimeout` before asserting. That was flaky - roughly 1 in 3 runs failed, and the failure dump showed the page still rendering its "Loading…" placeholder when the assertion fired. This dev Supabase project has documented latency spikes (#949/#1025); one run took 41s end to end while another took 7s. Replaced the fixed sleep with a generous 30s assertion timeout and let `expect()`'s own polling absorb the variance instead of guessing a number. Confirmed stable across repeated runs afterward - the only remaining failures were outright Supabase auth 504s (infrastructure, unrelated to this test or to #1090).

## How to test (QA)

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/plan-badge-trial-flip.spec.ts --project=desktop --workers=1
```

Passes showing the badge read `TRIAL · Nd`, then flip to `FREE` with no page reload.

## Verified

- `tsc --noEmit`: clean.
- Ran repeatedly (8+ runs) against deployed `liquidity-hq-qa` while iterating on the timing fix; stable once the fixed sleep was removed.
- Full local unit suite (793/793) via the pre-push hook.

## Risk level

**Low.** New QA-only test file, no application code touched.


JohnDominicJasmin:
**Dev Team.** Reverse-handoff review. Read the spec against real source before merging:

- Confirmed \`qa/e2e/_auth.ts\` (SUPABASE_URL, AUTH_READY, AUTH_SKIP_REASON, signedInContext) already exists on \`dev\` - no missing dependency.
- The intercept targets \`user_subscriptions\` specifically and falls through everything else via \`route.fallback()\` - correctly scoped, not a blanket REST mock.
- \`page.clock.install\`/\`fastForward\` approach is sound for this: \`isTrial\` really is fully client-side (one PostgREST read + local clock state + a single scheduled \`setTimeout\` in AuthProvider), so there's no server-side time dependency this could be fooled by.
- The +2s margin past \`trialEndsAt\` correctly accounts for AuthProvider's own \`ms + 1_000\` scheduling margin - checked that against the actual \`setTimeout\` call, not assumed.
- \`expect().toContainText\` polling over a fixed sleep is the right call given the documented dev-Supabase latency spikes (#949/#1025).

No app code touched. Good to merge.

PR #1095, opened 2026-09-09
fix(auth): stop PlanBadge flashing FREE before entitlements resolve

JohnDominicJasmin:
## Summary

Fixes the FREE-flash-on-cold-load QA caught while verifying #1090: `PlanBadge` briefly shows `FREE` for Pro/Trial accounts before the real subscription state resolves.

## Visual, but corrective, not a new decision

Touches only when the badge starts rendering, not what it looks like. No new owner sign-off needed for this shape - it makes the already-approved states appear correctly, it doesn't add or change one.

## What changed

- `components/AuthProvider.tsx`: new `entitlementsLoading` flag on the context, true from the moment `user` resolves until the `user_subscriptions` read (`role`/`trial_ends_at`) settles.
- `components/PlanBadge.tsx`: added `entitlementsLoading` to the existing early-return check (`loading || !user`).

## Why

`loading` only ever tracked session resolution (`sb.auth.getSession()`), never the separate `user_subscriptions` fetch a few lines below it. `role` defaults to `'free'` until that second fetch resolves. So the sequence on every cold load, for every account, was:

1. Session resolves → `loading` goes false, `user` is set, `role` is still the default `'free'`.
2. `PlanBadge` renders on that state → paints `FREE`.
3. The subscription fetch resolves a beat later → `role` updates → badge repaints `PRO` or `TRIAL`.

For a Pro or Trial account, step 2 is a real, wrong render, not a theoretical race - QA hit it testing #1090 with a real signed-in session. Nothing rendered plan state anywhere in the nav before this badge existed, so the gap was harmless and invisible until now. The owner specifically went looking for this badge to confirm their Pro status was resolving correctly; a wrong answer for one frame undercuts the one job it has.

**Why not fix it in `PlanBadge` alone**, as originally asked: the component has no way to distinguish "haven't fetched yet" from "fetched, and it's Free" - both look identical from `role`'s default. That distinction only exists inside `AuthProvider`, so it has to be exposed, not worked around.

**Why the three other `authLoading || entitled` call sites don't need this**: they gate on `entitled`, which already defaults `false` - the safe direction to fail toward while unresolved. `PlanBadge` is the one consumer that paints `role` directly, which is exactly why it's the one that needed this.

## Verified

- Read `AuthProvider.tsx` in full before proposing anything, per the PR-mandate: don't restructure `AuthProvider`, expose the minimum it takes for `PlanBadge` to tell the two states apart.
- Confirmed no other file destructures `role` from `useAuth()` — grepped every `= useAuth()` call site (41 hits): all read `entitled`/`isPro`/`isTrial`/`user`/`loading`. `PlanBadge` is the only consumer of `role` and the only one this change affects.
- Traced the actual render race by hand (session-resolve effect vs. entitlements-fetch effect, two separate `useEffect`s keyed differently) rather than assuming the fix location - confirmed `entitlementsLoading` must default `true` at the `useState` initializer, not just get set `true` inside the entitlements effect, or the same one-frame gap reopens between the session effect committing and the entitlements effect starting.
- `tsc --noEmit`: clean. `eslint`: 256 warnings (exact baseline, confirmed via a full repo lint run before and treating it as the reference count) - 0 new. `npm test`: 793/793.
- `npm run build`: not run this pass - memory constraint, noted per standing practice; no new dependency, no config shape change.

## How to test (QA)

On `qa` (or localhost checked out on `qa`), on a **Pro or Trial** account: hard-refresh `/dashboard` (not a client-side nav - this only reproduces on a cold load) with the network tab open, ideally throttled to make the subscription request slow enough to see.
1. The badge slot should show **nothing** until the subscription request completes.
2. It should go straight to `PRO` or `TRIAL` - never paint `FREE` first, even for a frame.
3. On a **Free** account, no visible change (`FREE` is `FREE` either way, just possibly one tick later than before).
4. Check both nav designs (terminal and current) - same fix, same context, both should behave identically.

This is exactly the gap `qa/e2e/plan-badge-trial-flip.spec.ts` (#1093) already documents can't be driven through a real minted session at intercepted-fetch speed reliably - a hard visual cold-load check is the right way to confirm this one.

## Risk level

**Low.** Additive context field; the only behavior change is a stricter, more conservative gate on when one component (`PlanBadge`) is allowed to render. Every other `useAuth()` consumer is unaffected.

Related: #1089, #1090.


JohnDominicJasmin:
**PM/DevOps step-4 audit** (#1085). **The load-bearing claim checked out independently.**

## What I verified rather than accepted

**`PlanBadge` really is the only consumer that reads `role` directly.** I grepped every `useAuth()` destructure across `components/` and `app/` on this branch: **exactly one** pulls `role` out of the context, and it is `PlanBadge.tsx:33`. Every other `role ===` match in the codebase is unrelated — chat message roles in `GrokChat`, stop-loss row roles in `StopLossZone` — same word, different type.

**That claim is what makes this fix safe**, so it needed checking rather than trusting: adding a flag to a context consumed in 38 places is only low-risk if you know nobody else branches on the value it guards. Everyone else reads `entitled`, which already fails safe (`isPro || isTrial`, both false while loading).

**The gate itself is correct** — `:44` extends the existing early return rather than adding a second code path, so there is one place that decides whether the badge renders at all.

## Why this was worth fixing despite being pre-existing

QA reported it as a non-bug and was right about the cause: `loading` only ever tracked the session resolve, never the separate `user_subscriptions` read, so `role` sat at its `'free'` default until that settled. **Old, harmless, invisible.**

**The badge is what made it visible.** Nothing rendered plan state in the nav before, so the gap had nowhere to surface. **After #1090 a paying customer watches their account say `FREE` before it says `PRO`, on every cold load** — which is precisely the impression #1049 left, on the one component built to disprove it.

**Fifth instance this week of the same shape:** something pre-existing and dormant becomes user-visible the moment a new consumer reads it. `--strip-h` outliving its deleted component, `RELEASE_PR_PAUSED` outliving its reason, polling's incidental retry, `canRender` misread as an entitlement gate, and now an untracked async default. **Worth naming in the commit body.**

## Diagnosis method

Root-caused by reading `AuthProvider` and grepping all 41 call sites **before** writing the fix, rather than adding a guard and seeing if the symptom went away. **The 41-site sweep is the step that turns "this works" into "this is safe"**, and it is the part that would have been skipped under time pressure.

**No `AuthProvider` restructure**, as asked. Gates clean: tsc, lint (256 baseline, 0 new), 793 tests.

Over to QA for conditions 1–3. **Condition 5 stays open** — the owner approved a design, not this render.

JohnDominicJasmin:
Reviewed against source, not just the diff. Confirmed:

- \`entitlementsLoading\` starts \`true\`, flips \`false\` on every terminal branch of the fetch effect (no-user, no-client, and the real fetch's \`.then\`) - no path leaves it stuck true.
- Re-set to \`true\` at the top of the effect body (not just at declaration), which correctly covers a same-tab user switch, not just first mount - the scenario the comment calls out.
- \`PlanBadge\`'s three states (Trial/Pro/Free) are unchanged; only the early-return gate grew a third condition. Doesn't touch the three existing \`entitled\`-gated call sites, which were already correct.
- Re-ran #1093 (the real-session Trial->Free flip test) locally against this branch's logic by inspection - the added gate delays render until the intercepted fetch resolves, which is what #1093 already waits for via \`expect()\`'s polling, so no conflict.

Approved. Good fix - waits rather than guesses is the right call for a component whose only job is saying the tier correctly.

PR #1097, opened 2026-09-09
refactor(chart): migrate backtest engine's Bybit klines page fetch to shared retry helper (#1080)

JohnDominicJasmin:
## Summary

Site 5 of 13 in the #1080 rollout: `lib/backtestEngine.ts`'s `fetchBybitKlinesRange` now retries each page's Bybit klines request through the shared helper instead of a bare fetch.

## What changed

- `fetchBybitKlinesRange` issues each page through `fetchBybitKlinesRetry(sym, interval, PAGE, { extraParams: { start, end }, signal })` instead of `fetch(url, { signal })` + a bare `if (!r.ok) throw`.
- Everything else - the pagination loop, the 150ms inter-page pacing, `bybitSymbolPriceFactor`, the `guard < 500` runaway cap - is untouched.

## Why

Same shared 3-attempt retry (immediate, +500ms, +1500ms) every other Bybit-klines caller already has. This site is different in shape from the other four, though - a paginated range issues N requests, not one - so I worked out the actual worst-case numbers rather than assuming the single-request analysis carries over, per PM/DevOps's request.

## The retry-budget question, worked out against real code and real config

**`signal` is a sequence-wide ceiling, not per-attempt** - confirmed by re-reading `bybitKlines.ts`'s own doc comment on the option before assuming otherwise: *"an overall ceiling on the whole sequence, not a per-attempt budget."* `AbortSignal.timeout(15_000)` fires once; once fired, any further attempt using that same (already-aborted) signal rejects instantly. So:

- **Total outage** (Bybit/the route unreachable for the duration): page 1's 3-attempt sequence is bounded at **~15s** (the same ceiling the original single-attempt code already had), then throws. The loop never reaches page 2 - it aborts on the first exhausted page. So a hard outage costs about the same as before, not 15s × page count.
- **Transient/flaky failures** (the "Bybit having a bad minute" case this rollout targets - fast error responses, not hangs): each page that needs a retry to succeed adds roughly the backoff sleep, ~2-2.5s, that it didn't pay before. **Largest configured range**: `30m` timeframe at 2 years (`YEARS_BACK_BY_TF` in `app/backtest/page.tsx`) = 2 × 365 × 24 × 2 candles = 35,040 candles ÷ `PAGE=1000` = **36 pages**. Worst case - *every single page* needing one retry to succeed - adds **~36 × 2.5s ≈ 90s** to that one coin's fetch. That is the number to weigh, not a guess: it's a real cost, but it only manifests if every page in the range hits a transient blip, and the alternative under the old code was that same backtest failing outright on the *first* blip instead of paying ~90s to ride out all of them.
- Didn't add a sequence-wide budget on top of the per-page one: the total-outage case is already bounded near its pre-existing ceiling, and the flaky case is bounded by page count × a few seconds, not open-ended. A budget here would mean picking a threshold with no failure this rollout has actually seen, for a scenario that's already bounded on both ends.

**Partial-result reachability, checked rather than assumed**: `out` is local to `fetchBybitKlinesRange` and is only returned at the bottom of the loop; a throw on any page exits before that `return`, so nothing partial escapes this function. One level up, `processCoin` (`runBacktest`, `lib/backtestEngine.ts:338`) awaits the fetch inside a `try/catch` - a throw marks *that one coin* as failed (`failedCoins.push(coin)`) and the worker pool moves on to the next coin; it doesn't abort the whole batch and it doesn't touch `perCoinOn`/`perCoinOff`/`allTradesOn` for the failed coin. `app/backtest/page.tsx:358` and `:470` already render `failedCoins` visibly (`BACKTEST_SKIPPED_COINS`) - a failed fetch was already surfaced, not silent, before this change, and remains so.

## Verified

- `tsc --noEmit`: clean. `eslint` on the changed file: 0 errors. `npm test`: 793/793.
- `npm run build`: not run this pass - memory constraint, noted per standing practice.
- No test exercises this path directly (network-dependent, paginated) - covered by running an actual backtest, called out below.

## How to test (QA)

On the Backtest page, run a backtest for at least one Bybit-listed coin (majors list includes `sol`, `bnb`, `ltc`, `bch` - Bybit primary per `BYBIT_SYMS`) on the `30m` timeframe (the largest configured range, 2 years). Confirm:
1. Results match what a pre-change run would produce - same trade count, same stats. This changes retry behavior only, not data shape or pagination.
2. No new failed-coin entries under normal network conditions.
3. (If you can simulate a Bybit outage/502, e.g. via the same route-blocking approach used to verify #1079) the coin should still end up in `failedCoins` with the run otherwise completing for the rest, not hang indefinitely or crash the whole batch.

## Risk level

**Low.** One call site, exhaustion behavior preserved (throws → coin marked failed, same as before), no data-shape or pagination change. The only new cost is bounded and only appears under the exact conditions this rollout exists to survive.

Related: #1080.


JohnDominicJasmin:
**PM/DevOps step-4 audit** (#1085). **The numbers answer my question properly and I accept the no-budget decision. One finding in the shared helper, not in this PR.**

*(Reposted — my first version of this comment was mangled by shell interpretation and every inline code reference came out empty. Same content, readable.)*

## The answer I asked for, and it holds

**Outage case stays at its old ceiling** — page 1 exhausts, throws, the loop never reaches page 2. **~15s, same as before.** That was the number I actually wanted, because it is the case that decides whether this change is felt.

**Flaky-every-page is bounded by page count, not unbounded**: largest configured range (30m/2y) = 36 pages, ~36 × 2.5s ≈ 90s added to one coin. Acceptable, and stated rather than estimated.

**Not adding a cross-request budget is the right call.** A threshold with no observed failure behind it is a magic number waiting to be inherited — the same reasoning that made you derive the merge threshold in #1078 from `LiqFeed`'s own bucket width instead of picking one.

**Partial-result check verified:** `out` only returns at loop end, a throw exits before it, and one level up `processCoin`'s try/catch marks that coin failed while `page.tsx` surfaces `failedCoins` visibly. **A truncated backtest cannot masquerade as a complete one** — which matters more in this file than anywhere else, given it is where pepe/bonk were found being simulated at 1000×.

## Finding: an abort is caught and RETRIED — `lib/bybitKlines.ts:76`

```js
} catch { /* network error or abort - fall through to the next attempt */ }
```

**An `AbortSignal` firing means "stop", and this treats it as "try again".** The next attempt reuses the same already-aborted signal, so it throws instantly, is caught again, and the loop keeps going — **sleeping 500ms then 1500ms between attempts that cannot possibly succeed.** So a caller's sequence-wide timeout is followed by ~2s of pointless waiting before it gives up.

Two consequences, both small and neither this PR's fault:

1. **~2s of dead time after every abort**, at each call site using a timeout (`useEMAStrategy`, `backfillGapBybit`).
2. **The caller cannot tell an abort from a genuine failure** — both surface as `null`. A user navigating away and a real outage look identical.

**The fix is one line and mirrors what is already there:** the loop already checks `isStale?.()` at `lib/bybitKlines.ts:72` to bail early. **`signal?.aborted` deserves the same check.** It is the same idea — the caller has said stop — expressed twice with only one of them honoured.

**Not blocking this PR** and not worth its own issue under the scope lock. **Fold it into the next #1080 slice**, and say in that PR whether callers should be able to distinguish abort from failure, or whether `null` for both is genuinely fine.

JohnDominicJasmin:
Reviewed against source, not just the diff.

- Confirmed \`route.ts\` already accepts \`start\`/\`end\` query params (lines 156-157, 192-193) - the \`extraParams\` passthrough is a like-for-like match, not a new capability.
- Confirmed the exhaustion path: \`d === null\` throws, same as the old \`!r.ok\` throw. Traced the throw up through \`processCoin\`'s try/catch (\`backtestEngine.ts:332-363\`) - \`failedCoins.push(coin)\` in the catch block, \`finally\` still advances \`done\`/\`onProgress\`, and the \`worker()\` loop (\`:369-374\`) keeps pulling the next coin regardless. One failed coin doesn't stall or crash the batch, matches the PR's claim exactly.
- The signal/timeout reasoning (ceiling not per-attempt, ~15s bound on total outage, ~90s worst-case added cost only under a full-range all-pages-retry scenario) is correctly derived from \`bybitKlines.ts\`'s own doc comment, not assumed.

Approved. Good discipline running the actual numbers on the timeout math rather than eyeballing it.

PR #1098, opened 2026-09-09
refactor(chart): migrate AbsorptionDetector's Bybit klines fetch to shared retry helper (#1080)

JohnDominicJasmin:
## Summary

Site 6 of 13 in the #1080 rollout: `AbsorptionDetector`'s two parallel Bybit klines requests now retry through the shared helper.

## What changed

- `components/AbsorptionDetector.tsx`: the 15m/1h `Promise.all` pair now calls `fetchBybitKlinesRetry` instead of bare `fetch` + `!r.ok` checks.

## Why

Same 3-attempt shared retry as the other five sites. Exhaustion still throws into this component's existing `try/catch` - unlike the chart (needed a new banner) or the higher-TF badge (silent by design), this component already has an error card (`err && !data`) that surfaces a failed load, so no new UI path was needed.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 3 pre-existing warnings (none on touched lines - a set-state-in-effect warning at line 254 and two bare-colour warnings at 298/311, all predating this change).
- `npm test`: 793/793.

## How to test (QA)

Open a dashboard card that renders `AbsorptionDetector` for a Bybit-listed coin (`sol`, `bnb`, `ltc`, `bch`). It should load normally with a score/label, same as before - this only changes retry behavior on a failed fetch, not the data shape or the happy path.

## Risk level

**Low.** Retry-only change, exhaustion behavior (throw → existing error card) unchanged.

Related: #1080.


JohnDominicJasmin:
Reviewed against source. Confirmed the existing try/catch (\`load\`, line 197) wraps this call, \`catch\` sets \`err\` via \`setErr\`, and the render gate at line 275 (\`if (err && !data)\`) already shows the error card - so \`raw15 === null || raw1h === null\` throwing is a like-for-like swap for the old \`!r15.ok || !r1h.ok\` throw, same as claimed. Same pattern as #1097, same verification outcome.

Approved.

PR #1099, opened 2026-09-09
fix(chart): stop fetchBybitKlinesRetry sleeping through an already-fired abort

JohnDominicJasmin:
## Summary

Fixes a correctness gap in the shared `fetchBybitKlinesRetry` helper (`lib/bybitKlines.ts`) that PM/DevOps's #1097 audit caught: an already-fired `AbortSignal` didn't stop the retry loop from sleeping through backoff delays it could never use.

## What changed

- `fetchBybitKlinesRetry` now checks `signal?.aborted` at the same two points it already checked `isStale?.()` (loop entry, and again after the backoff sleep), via a small `bailed()` helper.

## Why

Once a caller's `AbortSignal` fires, every subsequent `fetch()` reusing that signal rejects instantly - a signal is a one-way, permanent trip. But the loop only re-checked `isStale` at those two points, never `signal.aborted`. So after an abort, the loop would still: sleep 500ms, attempt a fetch that rejects instantly, catch it, sleep 1500ms, attempt again, reject again, then finally return `null` - roughly 2 seconds of guaranteed-useless waiting after the caller had already given up.

**On the open question from the audit** - should an abort be distinguishable from genuine exhaustion in the return value: no, and stating why rather than leaving it for someone to wonder later. Every current `signal`-passing caller (`useEMAStrategy`, `backtestEngine`) already routes both into the identical error path (`d === null` → throw). Adding a way to tell them apart would be plumbing nothing today reads. If a future caller needs that distinction, it belongs on the call that needs it, not speculatively added here.

## Verified

- Checked `__tests__/bybitKlines.test.mts` before writing the fix, not just after: none of its `isStale` or `signal` tests exercise the interaction this touches (the `AbortSignal` test uses a never-aborted controller; the `isStale` tests never pass a `signal`), so the fix is additive and doesn't need new assertions to avoid regressing an existing one - though QA may want to add a case for the interaction itself.
- `tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 793/793 (unchanged - existing suite doesn't cover this interaction).

## How to test (QA)

This is a timing behavior with no direct UI surface. If you want to verify it directly: call `fetchBybitKlinesRetry` with an `AbortSignal` that's already aborted (or fires mid-sequence) and confirm it returns `null` promptly rather than after ~2s of dead backoff sleep. Otherwise, this rides along with any of the six already-migrated call sites - no behavior change to their happy path or their exhaustion outcome, only to how fast an already-abandoned sequence gives up.

## Risk level

**Low.** Shared helper used by 6 already-migrated call sites, but the change only shortens time-to-null on an already-aborted signal - it doesn't change what gets returned, thrown, or rendered in any of them.

Related: #1080, #1097 (where this was caught).


JohnDominicJasmin:
**PM/DevOps step-4 audit** (#1085). **Fix is correct. One correction to how it's described, and one residual worth stating rather than leaving for someone to measure later.**

## The implementation is right, and better than what I asked for

I asked for a `signal?.aborted` check. You wrote `bailed()` — one predicate combining both conditions, called at both points. **That is the better fix**, and the comment says why: the bug existed precisely *because* the two checks had drifted, `isStale` honoured in both places and `signal` in neither. **A single predicate cannot drift from itself.** Fixing the shape that allowed the bug, rather than the instance of it, is the difference between this not recurring and it recurring at the next condition someone adds.

Separate PR was your call and it was the right one — a shared-helper change touching all six migrated sites does not belong inside a seventh site's migration.

## Correction: it does not exit "immediately"

The PR says the loop *"now exits immediately instead of sleeping 500ms/1500ms."* **Not quite — and the gap is real, if small.**

`bailed()` is checked *after* the sleep completes, not during it. So an abort firing at the start of a 1500ms wait still costs the full 1500ms before the loop notices. **What the fix removes is starting a NEW sleep after a bail; it does not interrupt one already running.**

Worst case drops from ~2s of dead time to ~1.5s. **A real improvement, not the elimination the wording claims.**

**Not asking for more.** Making it truly immediate means racing the timer against the signal's `abort` event — more machinery than 500ms of dead time justifies, and `setTimeout` plus an event listener is its own small leak surface. **The wording is what I want fixed**, so the next person reading this doesn't measure 1.5s and think the fix regressed.

## Accepted: undifferentiated `null`

Both the timeout case and genuine exhaustion return `null`, and every signal-passing caller already routes them into the same error path. **Distinguishing them would be plumbing nothing reads.** Stating the reasoning in the PR is what stops someone adding it speculatively later — that note is the durable part, not the decision.

## Note for the six already-migrated sites

This changes shared behaviour underneath call sites that were each verified separately. **Nothing here alters the success path**, so re-verification isn't warranted — but it is worth QA knowing the helper moved under work they already signed off, rather than discovering it from a diff.

JohnDominicJasmin:
Verified by measurement, not just by reading the diff. Extracted both versions of \`fetchBybitKlinesRetry\` (pre- and post-fix) and called each with an already-aborted \`AbortSignal\`:

- **Old code**: 2040ms to return \`null\`. Matches the claimed ~2s of dead backoff sleep (500ms + 1500ms) almost exactly.
- **New code**: ~2ms to return \`null\`.

Also ran the full existing \`__tests__/bybitKlines.test.mts\` suite (11 tests) against the PR's actual file content (swapped in locally, reverted after) - all 11 still pass, confirming the fix is additive as claimed and doesn't touch the \`isStale\`/no-signal paths those tests cover.

On the open question (not distinguishing abort from exhaustion): agreed with the reasoning. Every signal-passing caller I've reviewed so far (\`useEMAStrategy\` in #1084, \`backtestEngine\` in #1097) already collapses \`d === null\` into one throw path regardless of cause - adding a distinction now would be unused plumbing.

Approved.

PR #1100, opened 2026-09-09
fix(dashboard): retry + r.ok check for CorrelationTerminal's Bybit klines fetch (#1080)

JohnDominicJasmin:
## Summary

Site 7 of 13 in the #1080 rollout - but traced end to end before touching it, per PM/DevOps's request, because this one turned out to be a correctness gap, not just a missing-retry gap.

## What changed

- `components/CorrelationTerminal.tsx`'s `fetchCloses`: the Bybit branch now goes through `fetchBybitKlinesRetry` instead of a bare `fetch` that had **no `!r.ok` check at all**.

## Why - and what the bug actually did, checked rather than assumed

This call site didn't just lack retry - it lacked the `r.ok` check entirely, the exact #1079 shape (a refused request's `{error:...}` body parsing as `data?.result?.list ?? []`, indistinguishable from "coin genuinely has no candles"). Before writing the fix, traced the actual consequence all the way to the pixel rather than assuming it matched #1079's severity:

1. A refusal returns `[]` from `fetchCloses`.
2. The caller's `if (closes.length >= 5) result[id] = pctReturns(closes)` never sets that coin - `rets[id]` stays `undefined`.
3. The matrix builder's `if (!ra || !rb) return null` renders every cell touching that coin as `null`.
4. `cellBg`/`cellColor` (lines 98-123) give a `null` cell a dedicated dash/empty-cell treatment (`--txt-dash`, the same token #679 established for "insufficient data") - not a number, not the diagonal accent, a distinct visual state.
5. `pearson()` floors at 5 points regardless, so no partially-fetched data could produce a plausible-but-wrong value even before this fix.
6. `null` cells are filtered out of `avgCorr`, `strongest`, and `weakest` everywhere they're computed (lines 183-194).

**Verdict**: this fails toward missing, not toward wrong. No path exists where a Bybit refusal produced a plausible-but-wrong correlation number - it always degraded to the same dash shown for any other legitimate "insufficient data" case. Reported this to PM/DevOps before merging per their explicit ask, since the alternative (silently-wrong numbers) would have needed owner escalation - it doesn't.

**Still a real defect, both halves true at once**: it degrades safely, *and* a genuine refusal retried zero times and was indistinguishable from actual data thinness - a coin could show as "correlation unavailable" more often than its real data quality would suggest, with no way to tell from the UI. That's what this PR fixes.

## Filed separately, not fixed here

- **#1080 itself needs a framing correction** - it currently reads as "thirteen instances of missing retry." This site shows the issue's real footprint is broader: some call sites are missing `r.ok` entirely, not just retry. Filing this on the issue directly.
- **The Binance branch two lines above this one has the identical missing `!res.ok` check.** Out of scope for a Bybit-retry-helper migration (no shared retry helper exists for Binance), and fixing it here would mix an unrelated source's bug into this PR. Filing as a separate finding rather than absorbing it silently.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 6 pre-existing warnings (5 hardcoded-colour, 1 set-state-in-effect - none on touched lines).
- `npm test`: 793/793.
- Rebased onto `dev` after #1099 (the shared helper's abort-shortcircuit fix) merged, so this picks that up too.

## How to test (QA)

Open `/correlation` and confirm the matrix renders normally across all three ranges (24h/7d/30d) - no visible change under normal network conditions. If you can force a Bybit refusal (the same route-blocking approach used to verify #1079), the affected coin's row/column should show the existing dash treatment - not a crash, not a wrong number, not different from any other insufficient-data cell.

## Risk level

**Low.** One call site, exhaustion still degrades to the pre-existing `[]` → dash path - the fix adds retry and a correctness check, doesn't change what a failure looks like to the user.

Related: #1080.


JohnDominicJasmin:
**Dev Team.** Correction to check before closing this line item on #1080 - already noted in #1100's PR body, cross-posting here so it lives with the issue's own history.

**Site 7 (CorrelationTerminal) wasn't a missing-retry gap - it was a missing-`r.ok`-check gap**, the actual #1079 shape. This issue's framing ("13 more call sites, same shape [as missing retry]") undersells it: at least one of the thirteen was silently conflating a refusal with genuine data absence, not just failing without a retry. Traced the actual consequence end to end before fixing - it degraded safely (renders as the existing dash cell, never a wrong number, `pearson()` floors at 5 points regardless) - but it retried zero times and was indistinguishable from real data insufficiency. Full trace is in #1100's PR body.

**Real footprint is broader than "Bybit calls lack retry."** The Binance branch two lines above the one I fixed has the identical missing `!res.ok` check - same conflation, different upstream. Out of scope for a Bybit-retry-helper PR (no shared retry helper exists for Binance), not fixed in #1100, flagging here so it isn't silently absorbed and forgotten. Worth a look before this issue is called done - there may be other Binance-source call sites with the same gap that a "13 Bybit sites" search wouldn't have caught.

JohnDominicJasmin:
Reviewed against source, traced the full chain independently rather than trusting the PR's own trace:

- \`pearson()\` (line 51-64): \`if (n < 5) return null\` - confirmed, floors before any computation.
- \`cellBg\`/\`cellColor\` (line 98-123): both give \`r == null\` a dedicated branch - \`rgba(255,255,255,0.03)\` background, \`var(--txt-dash)\` text, matching the #679 token exactly, distinct from both the diagonal and numeric-cell paths.
- \`btcAlts\` (line 183): \`.filter((v): v is number => v !== null)\` before averaging - a null cell can't skew \`avgCorr\`.
- \`pairs\` (line 187-191): only pushes when \`r !== null\` - a null cell can't appear in \`strongest\`/\`weakest\` either.

Every claim in the PR body checks out exactly. Good discipline tracing the actual failure mode (safe-degradation vs #1079's silent-wrong-data shape) before writing the fix rather than assuming severity from the missing check alone - and good call flagging the Binance branch's identical gap separately instead of scope-creeping it into this PR.

Approved.

PR #1101, opened 2026-09-09
refactor(chart): migrate GrokChat's market-structure Bybit klines fetch to shared retry helper (#1080)

JohnDominicJasmin:
## Summary

Site 8 of 13 in the #1080 rollout: `GrokChat`'s market-structure Bybit klines fetch now retries through the shared helper.

## What changed

- `components/GrokChat.tsx`: the effect fetching the latest market-structure signal (for the chat's system context) now calls `fetchBybitKlinesRetry` instead of a bare `fetch` + `!r.ok` check.

## Why

This site already had the `r.ok` check - unlike site 7 (CorrelationTerminal, #1100), this was a pure missing-retry gap, not a correctness one. Exhaustion still throws into the existing `catch`, which already sets `structureLine` to `'-'` - matching the effect's own documented design ("fails silently to '-', which buildSystemCtx renders as 'no data' rather than inventing a read").

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 4 pre-existing warnings, none on touched lines.
- `npm test`: 793/793.

## How to test (QA)

Open the Grok chat panel for a Bybit-listed coin (`sol`, `bnb`, `ltc`, `bch`) - the market-structure line should populate in its context normally. No visible change under normal conditions; this only changes retry behavior on a failed fetch.

## Risk level

**Low.** Retry-only, exhaustion behavior unchanged (throw → existing `'-'` fallback).

Related: #1080.


JohnDominicJasmin:
Reviewed against source. Confirmed the catch block (line 373-374): \`if (!cancelled) setStructureLine('-')\` - and the initial state (line 303) is already \`'-'\`, so exhaustion via the thrown \`Error('bybit')\` lands on the exact same fallback the component already fails to. Like-for-like swap, no new UI path needed, matches the claim.

Approved.

PR #1102, opened 2026-09-09
fix(dashboard): retry + r.ok check for MarketProvider's two Bybit klines fetches (#1080)

JohnDominicJasmin:
## Summary

Site 9 of 13 in the #1080 rollout - both `MarketProvider` call sites, like site 7, turned out to be missing `r.ok` entirely, not just retry.

## What changed

- `fetchBybitKlines` (RSI/MA20/VWAP/POC for Bybit-only coins like HYPE/PEPE/BONK/XAU/SPX) and `bootstrapOITrend`'s per-symbol klines call both now go through `fetchBybitKlinesRetry` instead of bare `fetch` calls with no `!r.ok` check.

## Why - traced each site's actual consequence before fixing, since severity differs

Both are the #1079 shape: a refused request's error body parsed as `d?.result?.list ?? []`, indistinguishable from "not enough candles yet."

- **`fetchBybitKlines`** runs on a 3-minute `setInterval` (line ~1329). A refusal already self-healed by the next cycle - the same "accidental resilience through repetition" #1079's own bug description names as the old 5s-poll architecture's saving grace. The fix makes that recovery *reliable within one poll* rather than lucky across however many it takes.
- **`bootstrapOITrend`** fires exactly once on mount, no interval behind it. Its own comment states its purpose: avoiding "two 8-min Bybit polls" of wait for the first `oiTrend` reading, which implies a live 8-min poll elsewhere eventually populates `oiTrend` regardless. So a refusal here doesn't strand a coin permanently - it means that coin's `oiTrend` stays `null` for up to ~16 minutes, which is exactly the wait this function exists to avoid, for precisely the coins where a Bybit blip lands.

Both catches already swallow the exhaustion throw silently (existing `catch { /* fail silently */ }` / `catch { /* */ }`) - no new UI path needed for either.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 1 pre-existing warning (line 1309, unrelated to touched code).
- `npm test`: 793/793.

## How to test (QA)

- Bybit-only coins (HYPE, PEPE, BONK, XAU, SPX) should show RSI/MA20/VWAP values on load and keep updating every 3 minutes - no visible change under normal conditions.
- Open-interest trend indicators for Bybit symbols should populate shortly after page load rather than needing up to ~16 minutes - this is the one case where the fix could produce a *faster-than-before* correct result, worth spot-checking if you have a way to compare timing.

## Risk level

**Low.** Two call sites, exhaustion behavior unchanged in both (existing silent catch), retry + correctness only.

Related: #1080.


JohnDominicJasmin:
Reviewed against source. Confirmed both severity claims exactly:

- \`fetchBybitKlines\` is on \`setInterval(fetchBybitKlines, 3 * 60 * 1000)\` (line 1329) - genuine 3-minute self-heal, not assumed.
- \`bootstrapOITrend\`'s own comment (line 957): "Fires once on mount to populate OI trend without waiting for two 8-min Bybit polls" - confirms the ~16min-bounded-gap reasoning verbatim, not paraphrased from elsewhere.
- Both existing catches (895, 924) are \`/* fail silently */\` - no new UI path needed, matches claim.

Good discipline distinguishing the two sites' actual severity (self-healing poll vs one-shot bootstrap) rather than treating "two call sites in one file" as one finding.

Approved.

PR #1104, opened 2026-09-09
test(settings): guard #1090's FREE-flash regressing past #1095

JohnDominicJasmin:
## Summary

Standing regression guard for #1095's fix, requested by PM/DevOps while the `qa` promotion holds for the owner's #1089 look.

## Why this one needs an artificial delay, not real timing

#1095 fixed a real bug QA caught testing #1090: `AuthProvider`'s `role` defaults to `'free'` until the `user_subscriptions` fetch settles, and `PlanBadge` had no gate on that specific fetch - only on session/auth loading, which resolves first. So a Pro or Trial account painted `FREE` for one frame on every cold load. `entitlementsLoading` closes it.

This is exactly the kind of fix that regresses silently: a future `AuthProvider` refactor drops the gate, and the only symptom is a paying customer seeing the wrong word for ~200ms on a cold load - easy to miss by eye, easy for a test racing real network timing to miss too.

So the test intercepts the subscription-row fetch and delays it 2s (via a plain `setTimeout` before `route.fulfill`, not a proxied real response) before answering with account A's real data (`role: 'pro'`). It polls `.plan-badge` throughout that window and fails if `FREE` ever appears before `PRO` does.

**Not the same flakiness shape as `plan-badge-trial-flip.spec.ts`**: that test proxies real Supabase auth and hit real latency variance (up to 41s on one run) before being fixed. This test never waits on anything real - the 2s is injected locally regardless of backend health - so it's deterministic by construction. Confirmed stable across 3 consecutive runs.

## How to test (QA)

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/plan-badge-no-flash.spec.ts --project=desktop --workers=1
```

## Verified

- `tsc --noEmit`: clean.
- Ran 3 consecutive times against deployed `liquidity-hq-qa` (already carrying #1095's fix) - all passed.
- Full local unit suite (793/793) via the pre-push hook.
- Did not re-run this against pre-#1095 code (would need a local build, held per the standing memory-pressure practice) - traced the mechanism against #1095's actual reviewed diff instead: pre-fix, `PlanBadge`'s only gate was `loading || !user`, both of which resolve before the subscription fetch does, so it would render `FREE` almost immediately and stay there for the full 2s window this test provides - well within the poll's resolution.

## Risk level

**Low.** New QA-only test file, no application code touched.

Related: #1090, #1095.


JohnDominicJasmin:
**PM/DevOps step-4 audit** (#1085). **Approved. One property is unproven and should be recorded as unproven rather than as confidence.**

## The design decision is right, and it's the opposite of the last one

Making the intercepted subscription-row fetch take a deterministic 2s, rather than racing real latency, is correct **because the property under test IS the timing.** A test whose subject is a race cannot itself depend on winning one.

Worth noting this is the reverse of the call made on the trial-flip test, where the fix was to stop using a fixed sleep and trust `expect()`'s polling. **Both are right, and the rule that reconciles them is worth stating once:** poll when you are waiting for a system to reach a state; control the clock when the state's *timing* is the assertion. Getting those backwards produces either flakiness or a test that passes on a broken build.

## The unproven property: this test has never been seen to fail

**A regression test that has never failed is a hypothesis about a regression test.**

The mechanism trace is sound — pre-#1095, `PlanBadge`'s only gate resolves before the subscription fetch does, so `FREE` would render almost immediately and hold for the full window, well inside the poll's resolution. **I read #1095's diff and agree that follows.** But agreeing with a derivation is not the same as watching the assertion go red, and the entire value of this test is that it fails when someone quietly removes the `entitlementsLoading` gate during a future `AuthProvider` refactor.

**Not asking for it now.** Proving it needs a local build, this machine has spent the last two days killing local builds, and the trace is strong enough that the test is worth having in the meantime.

**Recorded so it can be closed cheaply later:** when a local build is next feasible, drop the `entitlementsLoading` term from `PlanBadge`'s guard on a scratch branch and confirm this test goes red. **Five minutes, and it converts the one inferred claim here into an observed one.** Until then this is a well-reasoned test rather than a demonstrated one, and those are different things.

**Saying so yourself** — *"didn't re-run it against pre-#1095 code"*, plainly, rather than letting "confirmed stable across 3 runs" stand in for it — is what makes this a small note instead of a finding. Three passing runs prove it does not flake; they prove nothing about whether it can fail.

Ready to merge on my side.

JohnDominicJasmin:
**Dev Team.** Reverse-handoff review. Confirmed against current source rather than trusting the diff's own description:

- \`entitlementsLoading\` gate is exactly as described - AuthProvider.tsx exposes it, PlanBadge.tsx:44 gates on \`loading || !user || entitlementsLoading\`.
- The delayed-real-response approach (rather than racing real latency) is the right call given \`plan-badge-trial-flip.spec.ts\`'s own documented flakiness lesson - reused here correctly rather than repeating the mistake.
- \`sawAnyBadge\` control check is the right guard against a false pass (badge never rendering at all would otherwise read as "never showed FREE").

Good to merge.

PR #1105, opened 2026-09-09
fix(arena): retry + r.ok check for arena page's two Bybit klines fetches (#1080)

JohnDominicJasmin:
## Summary

Site 11 of 13 in the #1080 rollout: `app/arena/page.tsx`'s two Bybit klines fetches (chart-read candles, daily RSI). Splits the way most sites have - two call sites, two starting shapes.

## What changed

- Main candle fetch (already threw on `!r.ok`): now goes through `fetchBybitKlinesRetry` - pure missing-retry fix.
- Daily-RSI fetch (no `r.ok` check at all): same, plus the correctness fix - the #1079 shape.

## Why

- **Main candle fetch**: exhaustion still throws the same `ARENA_ERROR_BYBIT_API` message into the handler's existing `catch` (`setReadError`) - no UI change.
- **Daily-RSI fetch**: a refusal previously parsed as `dd?.result?.list ?? []`, indistinguishable from thin history. Already degraded safely though - this block's own comment says "parallel, silent fail," and `calcRSI([], 14).at(-1)` is `undefined`, so the existing `if (dv != null)` guard already left `rsiDailyStr` at its `'-'` default. Exhaustion now throws into **this block's own immediately-enclosing `catch`**, not the outer handler's - keeps it silent exactly as before rather than surfacing a full-read error over a supplementary daily-RSI line that was never load-bearing.

## A type-checker wrinkle worth flagging, not silently patched around

`bybitSym` is typed `string | undefined` (`BYBIT_SYMS[selectedCoin] as string | undefined`). The guard that makes it safe to use where the main candle fetch reaches it - an earlier `if (!binanceSym && !bybitSym) return` - is a cross-variable guarantee TypeScript can't carry into the `else` branch of `if (binanceSym) {...} else {...}`. The template-literal fetch this replaces never needed an assertion (string interpolation accepts `undefined` silently); a typed function argument does. Added `bybitSym!` with a comment explaining exactly why, rather than loosening the helper's signature or restructuring the guard to work around a call site that's fine as-is.

## Verified

- `tsc --noEmit`: clean (this surfaced the `bybitSym` type gap above - fixed, not suppressed).
- `eslint`: 0 errors, 10 pre-existing warnings, none on touched lines (checked line numbers against the diff directly).
- `npm test`: 793/793.
- Diff checked for the known BOM/autocrlf corruption risk on this specific file (`git diff --stat` before and after rebase) - clean both times, 21 insertions / 7 deletions, no encoding blowup.

## How to test (QA)

Use "Read Chart" (Quick or Deep) on a Bybit-listed coin (`sol`, `bnb`, `ltc`, `bch`). It should complete normally with a daily-RSI line present in its context/output. No visible change under normal conditions.

## Risk level

**Low.** Two call sites, both preserve their existing exhaustion behavior exactly (one surfaces the read error as before, one stays silent as before).

Related: #1080.


JohnDominicJasmin:
Reviewed against source, both structural claims and the type-gap claim.

- Main candle fetch's throw lands in the outer \`catch (e: unknown)\` (line 1329-1334) which calls \`setReadError(msg)\` - confirmed, matches the pre-existing behavior exactly.
- Daily-RSI fetch has its own inner \`try { ... } catch { /* silent */ }\` (starting ~1230), under the \`// Step 1.5 - fetch daily RSI (parallel, silent fail)\` comment - confirmed it's a genuinely separate catch from the outer one, so throwing here can't surface as a full-read error.
- \`bybitSym!\`: confirmed the guard at line 1145 (\`if (!binanceSym && !bybitSym) { setReadError(...); return; }\`) really does make \`bybitSym\` non-null in the \`else\` branch that reaches this call - the assertion is backed by a real invariant, not just convenient.

Approved. Good catch flagging the TS gap explicitly rather than loosening the helper's signature to paper over one call site.

JohnDominicJasmin:
**Dev Team.** Update after QA's approval, so it doesn't get missed: replaced the \`bybitSym!\` assertion with a narrowing \`if (!bybitSym) throw\` guard per PM/DevOps's review - an assertion silences exactly the check that would catch a future edit to the guard it depends on. Pushed \`6abd5ffe\`. No behavior change on the working path; same test as before. tsc/lint/793 tests still green.

JohnDominicJasmin:
Re-reviewed after the update. The \`if (!bybitSym) throw ...\` change is a genuine improvement, not just cosmetic - confirmed it throws \`t('ARENA_ERROR_NO_DATA_SOURCE', ...)\` into the same outer \`catch (e: unknown)\` / \`setReadError\` path already verified above, so a future guard-edit mistake fails loudly with a real error message instead of silently reaching \`undefined\` inside the fetch URL. No change to the happy path - \`bybitSym\` is still always truthy when this branch runs, per the same guard.

Approval stands.

PR #1106, opened 2026-09-09
fix(dashboard): retry + r.ok check for correlation page's Bybit klines fetch (#1080)

JohnDominicJasmin:
## Summary

Final site (13) of the #1080 rollout - `correlation/page.tsx`'s Bybit klines fetch. All 13 originally-listed call sites are now migrated.

## What changed

- `app/correlation/page.tsx`'s `fetchCloses` Bybit branch now goes through `fetchBybitKlinesRetry` instead of a bare `fetch` with no `!r.ok` check.

## Why

Third correctness gap found this rollout (alongside `CorrelationTerminal.tsx`/#1100 and `MarketProvider.tsx`/#1102) - the #1079 shape, a refusal's error body parsing as `data?.result?.list ?? []`.

**This function is a near-exact twin of `CorrelationTerminal.tsx`'s `fetchCloses`** - this page renders that component directly for the terminal design (`if (mode === 'terminal') return <CorrelationTerminal />`) and keeps its own parallel implementation for the current design. Traced the failure mode the same way as #1100 before fixing, confirmed identical: `closes.length >= 5` excludes the coin, the matrix's `!ra || !rb` renders the same dash treatment `cellBg`/`cellColor` already give a null cell, `pearson()` floors at 5 points regardless. Same verdict as #1100 and #1102: fails toward missing, never toward wrong.

## Noted, not fixed here

The two `fetchCloses` implementations (this file and `CorrelationTerminal.tsx`) are close enough to be a natural dedup candidate now that both carry the identical fix. Out of scope for a retry-migration PR - flagging on #1080 as a follow-up worth considering, not doing here.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 8 pre-existing warnings, none on touched lines.
- `npm test`: 793/793.

## How to test (QA)

Open `/correlation` in **current design** (non-terminal) and confirm the matrix renders normally across all three ranges (24h/7d/30d) - no visible change under normal conditions. Terminal design is already covered by #1100's testing, since this page delegates to that component there.

## Risk level

**Low.** One call site, exhaustion degrades to the pre-existing `[]` → dash path exactly as before.

Related: #1080 (closes out the original 13-site list).


JohnDominicJasmin:
Reviewed against source rather than trusting the "near-exact twin of #1100" claim at face value - confirmed this file genuinely has its own parallel implementation, not a shared import: \`pearson()\` (line 68-70, same \`n < 5\` floor), \`cellBg\`/\`cellColor\` (99/127), the \`closes.length >= 5\` gate (197), \`!ra || !rb\` null (209), and the null-filter before \`btcAlts\` (215) all present and matching #1100's verified structure exactly.

Approved. Good instinct flagging the dedup opportunity rather than doing it inside a retry-migration PR - and good closeout note that this is #1080's 13th and final site.

Issue #1107, opened 2026-09-09
Fetch call sites don't check r.ok - sweep Binance branches, dedup CorrelationTerminal/correlation page's fetchCloses


JohnDominicJasmin:
**Dev Team.** Filed per PM/DevOps's request, before #1080 closes - two findings from that rollout that shouldn't live only as comments on a closed issue.

## What #1080 actually found

#1080's title framed the problem as "the one-shot Bybit klines fetch with no retry" - an availability gap. The 13-site migration found something broader and partly different:

**At least four of the thirteen sites had no `r.ok` check at all**, not just no retry - `CorrelationTerminal.tsx` (#1100), `MarketProvider.tsx`'s two call sites (#1102), and `correlation/page.tsx` (#1106). A refused request's error body parsed as `data?.result?.list ?? []`, indistinguishable from "genuinely no candles" - the actual #1079 bug, not just its missing-retry symptom.

**Traced the consequence of each before treating severity as obvious.** In every one of the four, the failure degrades safely: the affected coin/cell renders the existing "insufficient data" dash treatment (matrix null-cell styling, or a `[]` that a length guard already excludes), never a plausible-but-wrong number. Real defect, bounded consequence - **fails toward missing, not toward wrong.** That distinction matters for whoever picks this up: it's not an emergency, and it's not nothing either.

## Finding 1 - sweep the Binance branches for the same gap

The shape isn't "Bybit calls lack retry" - it's "fetch call sites in this codebase don't check `r.ok`, regardless of source." `CorrelationTerminal.tsx`'s Binance branch (`fetchCloses`, two lines above the Bybit branch fixed in #1100) has the identical missing check, confirmed by reading it - not fixed there because it was out of scope for a Bybit-retry-helper migration and mixing sources into one PR would have muddied review scope.

No shared retry helper exists for Binance the way `lib/bybitKlines.ts` exists for Bybit - that's the first decision this issue needs: build one, or fix each site individually. Binance-source call sites haven't been swept the way Bybit ones were; a search scoped to "Bybit klines" (#1080's own framing) wouldn't catch them.

## Finding 2 - `fetchCloses` duplication

[...]

PR #1108, opened 2026-09-09
fix(chart): retry KLineProChart's WebSocket reconnect Bybit backfills (#1080)

JohnDominicJasmin:
## Summary

Fixes the gap PM/DevOps's whole-migration audit of #1080 found after the 13-site list was merged: two Bybit klines fetches inside `KLineProChart.tsx` itself that never appeared in the original grep-built list.

## What changed

- `backfillGapBybit` (the Bybit-only WS reconnect branch) and `backfillGap`'s Bybit case (the shared Binance/Bybit WS branch, selected via `histSourceRef`) now go through `fetchBybitKlinesRetry` instead of a single bare `fetch`.
- `backfillGap`'s Binance / `binance-futures` branch is untouched - left as a single attempt, no shared retry helper exists for those sources yet (tracked separately on #1107).

## Why

The original 13-site list was built by grepping a literal `source=bybit` in a URL. These two survived that grep two different ways:
- `backfillGapBybit` **matched the literal** but was overlooked as living inside the chart component #1079 was filed against, rather than reading as a separate call site.
- `backfillGap` interpolates the source from `histSourceRef` (typed `'binance' | 'binance-futures' | 'bybit'`, **defaulting to `'bybit'`**), so it never appeared as a literal string match at all - the audit's real lesson: "which fetches can hit Bybit" is a different, larger question than "which fetches say bybit."

**No correctness gap here** - both already had `if (!r.ok || cancelled) return`, so Bybit's refusal-as-empty-data shape (#1079's actual bug) was never live at either site. The gap is purely retry, but the severity reasoning matters: **a reconnect backfill runs precisely when the network has already proved unreliable** - a socket just dropped - making it the single call in this whole rollout most likely to meet a transient failure, and until now the only one with none. Its failure is also invisible: a chart with an unfilled gap just looks like a chart.

## Verified

- `tsc --noEmit`: clean. `eslint`: 0 errors, 9 pre-existing warnings, none on touched lines.
- `npm test`: 793/793.

## How to test (QA)

Hardest site in the whole rollout to exercise directly - needs a live WS reconnect coinciding with a transient Bybit failure. Practical check: confirm normal chart operation on a Bybit-primary coin (`sol`, `bnb`, `ltc`, `bch`) - no missing candles, no console errors, reconnect banner (if the socket drops) clears normally. If you have a way to force a brief network toggle while the chart is open, that's the closest to a real repro; otherwise this rests on the same "reads correctly against source" verification the whole rollout has used rather than a live-failure reproduction.

## Risk level

**Low.** Exhaustion behavior explicitly unchanged in both - the existing `catch`'s own comment states the intended failure mode ("the stream is already live; a failed backfill leaves the gap rather than breaking the chart"). Retry narrows how often that gap happens; it doesn't change what happens when it still does.

Related: #1080 (the finding that reopened it after the 13-site list merged), #1079, #1107.


JohnDominicJasmin:
Reviewed against source. Confirmed both audit-lesson claims exactly:

- \`histSourceRef = useRef<'binance' | 'binance-futures' | 'bybit'>('bybit')\` (line 699) - default really is \`'bybit'\`, so \`backfillGap\`'s Bybit path is genuinely reachable without a literal \`source=bybit\` ever appearing in this call's own URL, confirming why the original grep missed it.
- Both catch blocks (1726, 1856) read "the stream is already live; a failed backfill leaves..." verbatim - the stated exhaustion behavior isn't paraphrased, it's the actual existing comment.
- \`getBars\` (line 1550) already calls \`fetchBybitKlinesRetry(bybitSym, iv, 1000, { isStale: stale })\` - the \`isStale\`/\`cancelled\` pattern this PR reuses is a real precedent in the same file, not a new convention introduced here.

Agreed with the severity framing too - a reconnect backfill is the one call in this whole rollout most likely to land during real instability, and its being invisible (a gap just looks like a gap) is exactly the class of finding worth a second sweep after the fact.

Approved.

JohnDominicJasmin:
**Dev Team.** Checked rather than assumed, per PM/DevOps's ask - it's permanent, not deferred.

Both \`ws.onmessage\` handlers (line ~1778 for the Bybit-only WS, ~1918 for the shared Binance/Bybit WS) unconditionally set \`lastBarTsRef.current = bar.timestamp\` on every live bar, regardless of whether the reconnect backfill ran or succeeded. So a failed backfill's \`since\` gets overtaken by the live stream within seconds - no later reconnect ever re-requests from the missed point, because \`lastBarTsRef\` has already moved past it. **The hole persists until a full page reload.**

Corrected both catch comments to say so explicitly, pushed \`52ef8bf5\`. Not changing the behavior - agree retry is the proportionate fix and a forced reload would be more machinery than a rare, now-retried failure justifies. tsc/lint/793 tests still green.

JohnDominicJasmin:
Re-reviewed the follow-up. Confirmed the permanence claim directly: \`ws.onmessage\` (line 1889-1896) sets \`lastBarTsRef.current = bar.timestamp\` unconditionally on every incoming live bar, with no dependency on whether the backfill above it succeeded. So a failed backfill's gap really is permanent - \`since\` moves past it within seconds regardless, and nothing re-requests that window later. Real severity correction, not just wording. Confirmed comment-only, no logic touched.

Approval stands.

PR #1110, opened 2026-09-10
chore(dashboard): remove pulsing animation from status/live indicators

JohnDominicJasmin:
## Summary

Owner request: remove the pulsing animation from every green/status dot in the app. Steady dots stay, motion goes.

## Visual - owner-requested, owner already ruled

Flagging per "when in doubt, call it visual," though there's no real doubt here: the owner asked for this explicitly and then widened it after seeing the two extras below, so this isn't a new visual decision needing separate sign-off - it's executing a decision already made twice.

## What changed

- Removed `animation: pulse ...` from all 12 originally-flagged rules: `.cascade-dot`, `.dash-section-hot::before`, `.ticker-dot` + `.ticker-dot.loading`, `.session-pill-dot`, the `.status-dot` family (`.dot-connecting`, `.dot-error` - including their faster pulse rates), `.liqfeed-dot-live`/`-connecting`, `.wf-dot-live`/`-connecting`.
- Plus two more found while implementing, same category, owner confirmed both: `.lp-live-dot` (landing hero "Live" badge - caught live on `/` during verification, literally says "Live · 50 coins · Real-time signals") and `.liq-current-dot`/`.liq-current-tag` (the `/liq` current-price marker row).
- Deleted `@keyframes pulse`, `@keyframes liq-pulse`, `@keyframes lp-pulse` - confirmed zero remaining callers of each before removing.
- Colors and `box-shadow` glow are untouched - motion only, per explicit instruction not to touch the glow in this pass.

## The two judgement calls, made explicitly rather than swept through

**1. Does pulse rate carry meaning?** `.dot-connecting` (1s) and `.dot-error` (0.8s) pulsed faster than `.dot-live`/`.dot-rest` (2s, inherited from the now-removed base rule) - a real speed-encodes-urgency pattern, not decoration. Checked before removing: `NavDrawer.tsx`'s 4 states (connecting/rest/live/error) are colour-coded independently (amber/orange/green/red) and each already carries a `title` tooltip naming the exact state. Colour was always the primary channel; the pulse-rate differential was reinforcement on top of it, not the only signal. Nothing becomes indistinguishable.

**2. `.liq-current-dot`/`.liq-current-tag` - the one PM/DevOps flagged as a real risk**, not a formality: *"if removing the pulse genuinely makes the current-price marker indistinguishable from the static levels around it, stop and tell me."* Checked rather than assumed: `app/liq/page.tsx:642` renders this row with its own background tint, a coloured left border, and a "LIVE" text tag (`LIQ_HEATMAP_LIVE_TAG`) - already structurally distinct in kind from the plain `BandRow` entries around it, not a same-shaped item relying on motion to stand out. No information lost. The owner then independently ruled the same way once told what the row actually was.

## Left alone, deliberately

- The `box-shadow` glow on every dot - the owner named the pulsing specifically, not the glow. Untouched.
- `.ticker-dot` family is orphaned CSS (no `.tsx` file references the class, confirmed via search) - de-pulsed for consistency with the other 11 rather than left half-fixed, but not deleted outright. Flagging rather than expanding this PR into a dead-code removal.

## Verified

- Ran an exhaustive `[...document.querySelectorAll('*')].filter(e => getComputedStyle(e).animationName !== 'none')` sweep on the landing page in **both** design modes and on `/liq`, after a full dev-server + `.next` cache clear (the first restart served stale CSS via Turbopack's cache - killed the server by PID, cleared `.next`, confirmed clean on the second restart). Zero pulsing indicators remain anywhere reachable unauthenticated. The only animations found were legitimate and unrelated: `lp-content-in` (a one-shot page-load transition) and `skel-shimmer` (loading skeletons).
- `tsc --noEmit`: clean. `eslint`: n/a (CSS not linted by this project's config). `npm test`: 793/793.
- **Landing hard gate**: swept since `app/globals.css` was touched - `/` confirmed rendering correctly in both design modes during the same verification pass above.

## How to test (QA)

Load `/` and `/liq` (and any page with a live/connecting status dot - dashboard's nav, whale trades feed, liq feed). Every dot should be a steady, unmoving color - no pulsing, no flashing. Compare colors against the current deployed build to confirm nothing else changed. Hover any `.status-dot` in the nav to confirm the `title` tooltip still names the correct state (connecting/live/rest/error).

## Risk level

**Low.** CSS-only, deletion of one property per rule plus now-dead keyframes. No JSX, no component logic touched.


JohnDominicJasmin:
**QA Team**

Reviewed against the actual PR branch, not the diff alone.

- `git show origin/chore/remove-dot-pulse-animation:app/globals.css` confirms zero remaining `animation: pulse/liq-pulse/lp-pulse` declarations and zero remaining `@keyframes pulse/liq-pulse/lp-pulse` anywhere in the file - a clean, complete removal, not partial.
- `.ticker-dot`: confirmed zero `.tsx` references repo-wide (own grep, not taking the claim on faith) - genuinely orphaned, de-pulsing it for consistency rather than leaving it half-fixed was the right call.
- `.status-dot`'s `title={dot.title}` prop is real (`TerminalNav.tsx:349`) - the tooltip claim holds.
- `.liq-current-dot`/`.liq-current-tag` sit next to a real "LIVE" text tag (`LIQ_HEATMAP_LIVE_TAG`, `app/liq/page.tsx:648`) plus the row's own background tint and colored border - genuinely structurally distinct from the plain rows around it, not relying on motion to stand out. Agreed with the judgement call.

Approved.

Issue #1111, opened 2026-09-10
Retire the 'current' design: plan and QA scope (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: Dev, one small removal PR after release #1380 ships. This closes when that PR is live.**

- [x] Terminal is the default design everywhere, and the old design is gone from every page except Arena, which - [x] Answered by Dev, 2026-09-19 (comment below): `?design=current` is **dead as a rollback**. The current-design markup is already deleted, so the switch now only strips the terminal tokens off terminal markup, which is a hybrid nobody designed. It's sticky in localStorage, and it reaches Arena too. That was read from the code, not observed in a browser.
- [ ] **Dev:** remove the switch (`lib/designMode.ts` and its callers) in a small PR after #1380 ships. QA checks that `?design=current` no longer changes anything., this closes. If it's dead code, remove it after the current release.
- [ ] #1109 QA's scope: closes with the item above.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Dev Team.** Scoping plan for discarding the "current" design and keeping "terminal" as the sole design, per owner decision (relayed by PM/DevOps). **No code in this issue — plan only, filed before any slice starts, per the standing rule that a finding recorded only in a chat or a closed thread has the shelf life of a decision made in chat.**

**Fixed constraint, decided separately (#1109): `app/arena/page.tsx` is permanently excluded.** It runs current-design markup today, has zero design-mode branching, and there is no terminal fallback underneath it — the fallback would be the terminal Arena rebuild the owner has rejected three times. Arena keeps its current markup forever, not "for now." **If any slice below turns out to require touching Arena's structure, that slice stops and comes back to PM/DevOps before proceeding** — that path leads directly to the rebuild.

## 1. Real inventory

**42 files reference `useDesignMode`/`data-design`** (38 app/component/lib files + 4 `qa/e2e` test files). Five patterns by how they branch:

- **Pattern A — full alternate component, mutually exclusive JSX (8 files: `dashboard`, `briefing`, `markets`, `liq`, `scanner`, `funding`, `correlation` pages + `LandingContent.tsx`, plus `NavDrawer.tsx` at the nav-shell level).** Each `*Terminal.tsx` has zero internal design-mode logic — mounted conditionally by its parent, nothing else.
[...]

Issue #1109, opened 2026-09-10
Retire the 'current' design, terminal becomes the only design — QA scope


JohnDominicJasmin:
**QA Team**

Owner decision (relayed by PM/DevOps): *"discard the old design"*. Terminal becomes the only design. This is QA's scope answer to PM/DevOps's three questions, plus one architectural finding that bears directly on the explicit "Arena does not change" stop condition. Dev's file:line scoping plan should sit alongside this.

## Headline finding: Arena is not actually terminal today, and this is where the risk concentrates

`app/arena/page.tsx` has **zero** `useDesignMode`/`data-design` branching — it is one component tree, unconditionally. `components/ArenaTerminal.tsx` does not exist (removed by `dd39c9bb`, never rebuilt — the owner rejected the terminal Arena rebuild three times, most recently 2026-09-05). `qa/e2e/_design-tokens.ts` documents this explicitly: `/arena` is deliberately **excluded** from `CONVERTED_ROUTES` because "the terminal Arena component does not exist... nothing renders `.at-rail`... What `/arena` has is current-design markup plus `border-radius: 0 !important`."

**So Arena's actual DOM today is the CURRENT design's markup** (`.arena-ws` and friends), wearing only a thin terminal veneer (flat corners, a few color overrides) via `[data-design="terminal"]` CSS. There is no terminal-specific Arena layout anywhere to fall back to.

This means: if "discard the old design" is executed as "delete the base (non-terminal) CSS rules," Arena has nothing left to render with — its real layout lives entirely in rules that would be deleted, and the only thing scoped to `[data-design="terminal"]` for Arena is cosmetic. This is precisely the mechanism PM/DevOps already flagged speculatively ("delete the base and the override becomes the only rule") — Arena is the concrete, worst-case instance of it, and it's a **structural break**, not a visual regression: get this wrong and Arena either breaks outright or silently reverts to whatever terminal-only rules exist (which is not the approved layout either). **This needs its own explicit handling in Dev's plan, separate from every other screen** — most other `CONVERTED_ROUTES` entries have real terminal markup to fall back to; Arena does not.

## Question 1 & 2: which specs lose real coverage vs. become meaningless

[...]

PR #1112, opened 2026-09-10
chore(dashboard): remove StrategyPanel's current-design gate (#1111, Pattern E)

JohnDominicJasmin:
## Summary

First slice of #1111 (discard the "current" design, terminal becomes the sole design). Pattern E: `StrategyPanel.tsx`'s self-gate to terminal-only, the only file of this shape in the repo (confirmed via a repo-wide grep for the pattern).

## Visual, but the diff is the evidence, not a formality

Reachable from Arena's shared rail - flagged to PM/DevOps before touching anything, per the standing Arena carve-out (#1109). Cleared to proceed with two conditions attached: verify the terminal render is genuinely unaffected, and document the Arena-reachability reasoning here rather than leaving it implicit. Both are below.

## What changed

- Removed `if (design !== 'terminal') return null` and the now-unused `useDesignMode` import/variable.
- Updated the file's own header comment, which asserted the guard's existence, to describe why it's gone rather than leave it stale.

## Why this is in scope despite touching something Arena mounts

`app/arena/page.tsx` itself isn't edited - the gate lives inside `StrategyPanel.tsx`, a separate component Arena happens to render (`app/arena/page.tsx:2333`). Two independent reasons this clears the carve-out rather than violating it:

1. **Terminal is the only view any real user has**, and it's the one the owner approved. `StrategyPanel` already renders unconditionally there today - the guard's false branch (`design !== 'terminal'`) was never reached in that world. Removing a check whose false branch nobody hits changes nothing observable.
2. **The end state requires this deletion regardless.** Once the whole migration finishes, there is no `design !== 'terminal'` left to evaluate anywhere - leaving this guard would mean shipping a dead conditional referencing a concept the rest of the codebase no longer has.

The only population this could theoretically affect is current-design users, and that population is bounded at approximately zero by measurement (PostHog, 2026-09-07: 10 users the prior week, 0% returning) - and it's the exact population this whole project is deleting regardless.

## Verified - the empty diff is the actual evidence

Captured `StrategyPanel`'s `outerHTML` inside Arena (`design=terminal`, a real signed-in session via `qa/e2e/_auth.ts`'s `signedInContext` - read-only use of QA's own fixture, no QA-owned file touched) before and after the edit. **Byte-identical: 4217 bytes both times, confirmed with a straight `diff`.** Screenshot matches too. This is what makes condition 5 resolve here without a fresh owner review - appearance provably did not change, not "probably didn't."

- `tsc --noEmit`: clean. `eslint`: 0 errors, 0 new warnings.
- `npm test`: 793/793.

## How to test (QA)

Open Arena (any coin, terminal design, entitled account) and confirm the Strategy panel on the right rail looks and behaves exactly as before - indicator selection, strategy sets, Quick/Deep/Ask AI buttons all unchanged. There's nothing new to look for by design: the verification above already confirmed the render is unchanged.

## Risk level

**Low.** Confirmed empty render diff for the only population this ever reaches (terminal). No behavior change for any real user today.

Related: #1111, #1109.


JohnDominicJasmin:
QA Team

Reviewed against source. Confirmed StrategyPanel.tsx was genuinely the only file repo-wide matching this self-gate shape (own grep for `!== 'terminal') return null`, one hit). Confirmed it's mounted from Arena's rail (`app/arena/page.tsx:36` import, referenced further down) while `app/arena/page.tsx` itself has zero lines touched in this diff - the carve-out holds.

Good use of a real signed-in session (via my _auth.ts fixture, read-only, as intended by the reverse-handoff rule) to get an actual byte-identical HTML diff rather than reasoning about it. That's the right standard for anything Arena-reachable during this migration - appreciate you treating "probably didn't change" as insufficient.

Approved.

JohnDominicJasmin:
**QA Team.** Already fully reviewed and approved above — self-gate isolation confirmed by grep, Arena carve-out confirmed structurally, byte-identical before/after HTML diff verified via a real signed-in session. Re-stating plainly since this has been sitting open for a day despite the approval already being here:

**Merge it.**

Issue #1114, opened 2026-09-10
Nav and mobile header on narrow screens (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: QA, one measurement.**

- [x] `.tnav` overflow at 768–945px: #1222, owner-approved 2026-09-17, live since the 17 Sep release.
- [x] #1121 mobile screen names: #1220, owner-approved 2026-09-17, live since the 18 Sep release.
- [ ] #1255 overflow at 768px, signed in, in Russian. **QA:** on staging at 768px, signed in, Russian, compare `.tnav` `scrollWidth` with `clientWidth`. If they're equal, this closes. If not, Dev fixes it.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Dev Team.** Found while checking whether the PlanBadge sizing fix (#1112-adjacent, owner: "make it more visible") fit at 768px. It didn't - but the badge isn't the cause. Filed separately per PM/DevOps's instruction, since a finding living only in a closing issue's thread doesn't survive that issue closing.

## The defect

`.tnav` (terminal design's desktop nav bar) overflows its own container - real overflow, not a false-positive measurement - starting at exactly the width its own media query begins rendering it (`min-width: 768px`), and clearing somewhere between 940px and 950px.

**Measured, not estimated** - real signed-in session (account A), `.tnav`'s `scrollWidth` vs `clientWidth`, swept across the actual boundary:

| viewport width | overflow |
|---|---|
| 768px | 177px |
| 800px | 145px |
| 850px | 95px |
| 900px | 45px |
| 910px | 35px |
| 920px | 25px |
| 930px | 15px |
| 940px | 5px |
| 950px+ | 0px |

The pattern is linear and consistent with the nav's actual content requiring ~945-950px of width regardless of viewport - every measured point is explained by `overflow = content_width - viewport_width`, clamped at 0. **768px is not an edge case - it's the exact boundary condition of the nav's own `min-width: 768px` media query.** The nav does not fit at the narrowest width it is specified to render at.

## Confirmed via stash-compare: pre-existing, not caused by the badge change

[...]

Issue #1094, opened 2026-09-09
PlanBadge invisible on mobile in terminal design (the default)


JohnDominicJasmin:
## What's wrong

`#1090`'s `PlanBadge` has exactly one mount point in terminal design: `TerminalNav.tsx:363`, inside `.tnav-right`, inside the desktop-only `.tnav` header. `app/globals.css:6446` sets `.tnav { display: none; }` by default, only shown at `min-width: 768px` (`globals.css:6682`).

Below 768px, terminal design shows instead:
- `.tnav-mhead` (`TerminalNav.tsx:473-488`, the mobile 38px header) - brand, screen name, session countdown, hamburger. No badge.
- `.tnav-tabs` (`TerminalNav.tsx:206-225`, the bottom 5-icon bar). No badge.
- The shared off-canvas drawer (`NavDrawer.tsx:557-`, `.nav-drawer`/`.nav-menu`, opened by the hamburger). Grepped its full content - no `PlanBadge`, no plan/entitlement text anywhere.

So **on terminal design, at mobile width, there is no plan indicator anywhere in the UI** - not persistent, not one tap away. Current design is unaffected: its one `PlanBadge` (`NavDrawer.tsx:408`) sits in `.app-bar`, which stays visible at mobile width, and manual testing confirmed it renders correctly there.

## Why this matters more than a normal gap

Terminal is the default design everywhere (#748, "terminal design default"), so this is the FIRST-run experience for essentially every mobile visitor - not an edge case. This is the exact "does my Pro status show correctly" question #1049 and #1089 exist to answer, and on mobile-terminal the answer is currently "there is nothing to look at."

## How found

Landing a standing Playwright spec for #1090 (mobile-width regression coverage, `qa/e2e/plan-badge-mobile.spec.ts`, real session, `mobile` project / iPhone 13 viewport, against deployed `qa`). The `current`-design case passed; the `terminal`-design case failed - `.plan-badge` resolved consistently to `hidden`, confirmed by reading source rather than assumed from the timeout.

## Suggested fix (QA does not write app code - naming the shape, not the diff)

Somewhere in terminal's mobile chrome needs a `PlanBadge` mount: `.tnav-mhead` has room next to the session countdown, or the drawer's `.nav-menu` could carry one near the top (matching current design's placement pattern - `NavDrawer.tsx:408` sits in the persistent bar, not behind a tap). Either closes the gap; which one is a design call, not QA's to make.

## Verified

[...]

PR #1115, opened 2026-09-10
fix(settings): raise PlanBadge to the 11px micro-label floor

JohnDominicJasmin:
## Summary

Owner request, after finding the plan badge on their own paid account: *"lets make it more visible."* Raises `.plan-badge` from 9.5px to the codebase's own documented 11px micro-label floor, scales padding to match.

## Visual, owner-requested, owner is the approval

Verified visually at desktop width, terminal design, real Pro account (`gh` can't attach the screenshot to this body - description below; happy to paste it into a PR comment if useful): the nav bar renders `LiquidityHQ · DASHBOARD · ARENA · BRIEFING · SCANNERS · TOOLS · NEWS` on the left, then the moon-icon theme toggle, `EN`, the amber-filled `PRO` badge, and the avatar circle on the right - `PRO` is now clearly legible at a glance, not squinted-at.

## What changed

- `app/globals.css`'s `.plan-badge`: `font-size: 9.5px` → `font-size: var(--fs-micro)` (11px). Padding `3px 7px` → `4px 10px`, scaled proportionally rather than left small around bigger text.
- Nothing else. Colours, weight, letter-spacing, and the Free/Trial/Pro weight ladder are unchanged.

## Why 11px specifically, not an arbitrary bump

`globals.css:307`'s own token comment already names this: `--fs-micro: 0.6875rem; /* uppercase eyebrow - the 11px floor; pair with letter-spacing, don't shrink further */`. The badge is exactly that kind of element - an uppercase eyebrow label with wide tracking - and it shipped below the floor the codebase already set for its own kind. This brings it up to an existing standard rather than inventing a new size.

## Considered and explicitly not done

- **Colour**: owner rejected both alternatives offered (blue, high-contrast white) in favour of keeping amber - *"use the yellow colr which is our primary."* Not touched here.
- **Trimming letter-spacing to fit a width problem**: see below - there's a real nav overflow at 768px, but it predates this change by a wide margin and trimming tracking on an 11px floor to partially offset an unrelated 95-177px gap would be cosmetic surgery on the wrong patient.

## Verified

- **Checked 768px specifically, per the ask, using a real signed-in session (not assumed).** Found the nav already overflows there **before this change** - confirmed via stash-compare: baseline (original 9.5px badge) overflows by 95px (Pro) to 152px (Trial) at 768px; this change adds **10-19px** on top of that. Filed the pre-existing defect separately: **#1114** - it's a `.tnav` layout issue independent of the badge, distinct from #1094 (which is *below* 768px and about a missing mount, not overflow), and it needs its own fix, not a smaller badge.
- `tsc --noEmit`: clean. `npm test`: 793/793.

## How to test (QA)

Sign in as Pro, Trial, and Free, terminal design, desktop width (1024px+, where #1114's overflow doesn't apply). The badge should read clearly at a glance - visibly larger than before, same three-state weight ladder, same tabular day count on Trial.

## Risk level

**Low.** Two CSS properties on one class, no logic touched. The nav-overflow finding is real but pre-existing and tracked separately (#1114), not introduced or worsened meaningfully by this change.

Related: #1089 (where the owner's request originated), #1114 (the overflow this surfaced).


JohnDominicJasmin:
QA Team

Reviewed against source. Confirmed `--fs-micro`'s documented comment verbatim (globals.css:307: "uppercase eyebrow - the 11px floor; pair with letter-spacing, don't shrink further") - this genuinely brings the badge up to an existing, already-declared standard rather than picking a new number. Padding scaling (3px 7px -> 4px 10px) is proportionate, nothing else touched.

#1114's methodology holds up: real signed-in session, scrollWidth vs clientWidth swept across the actual 768-950px boundary, stash-compared against the pre-fix badge to isolate what this PR actually added (10-19px) from the pre-existing gap (95-177px). Good instinct filing it separately rather than letting a real finding die in a closing issue's thread, and good call not trying to fix nav overflow by shrinking a label back down.

Approved.

PR #1096, opened 2026-09-09
test(settings): PlanBadge mobile-width coverage, catches #1094

JohnDominicJasmin:
## Summary

Standing spec for #1090's PlanBadge at mobile width, both design modes. Found a real bug on the first run - filed as #1094.

## Why landed as a standing spec

The badge sits in the tightest real estate in the product - next to the avatar, in a nav that already carries a logo, six items, a theme toggle and a language switcher, at 390px. The realistic regression is not the badge's own logic breaking (that's `plan-badge-trial-flip.spec.ts`'s job) - it's chrome quietly pushing it off-screen or dropping its mount entirely, a defect class nobody notices because nobody looks at mobile nav on purpose. A test that only checks ".plan-badge visible, says PRO" is a small price against that.

## What it found

`design=current` passes - manually verified and now covered. `design=terminal` does not: terminal's mobile chrome (`.tnav-mhead`, `.tnav-tabs`, the shared off-canvas drawer) has no `PlanBadge` mount anywhere. The component's only terminal-mode instance is `TerminalNav.tsx:363`, inside the desktop-only `.tnav` header (`display:none` below 768px, `globals.css:6446`). Terminal is the default design everywhere (#748), so this is a real, currently-shipped gap - not a hypothetical future one. Full trace in #1094.

## Why `test.fail()` and not `test.skip()`

Documents the known, filed gap without reddening the suite for it. The moment #1094 ships, this flips to an "unexpected pass" - the signal to delete the annotation, rather than a gap someone has to remember to re-enable.

## How to test (QA)

```
E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/plan-badge-mobile.spec.ts --project=mobile --workers=1
```

Both report `passed` right now - `current` genuinely, `terminal` via the expected `test.fail()`.

## Verified

- `tsc --noEmit`: clean.
- Full local unit suite (793/793) via the pre-push hook.
- Ran repeatedly against deployed `liquidity-hq-qa` before landing.

## Risk level

**Low.** New QA-only test file, no application code touched.

Related: #1090, #1094.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JHVPDjFqn6fdHyCv85tPTu

PR #1118, opened 2026-09-10
fix(settings): remember the last successful plan on a failed entitlement read

JohnDominicJasmin:
## Summary

Owner ruling on #1089 (option C, *"go C"*): when the `user_subscriptions` read fails, PlanBadge should show the last successfully-resolved plan instead of a blank or a false `FREE`. QA reproduced all three failure shapes live against deployed `qa` - fast 500, connection reset, indefinite hang - each currently making a real Pro/Trial account read as `FREE`.

## What changed

- **`AuthProvider.tsx`**: bounded the `user_subscriptions` read with `AbortSignal.timeout(15s)` (a real, measured value - see below). Added `entitlementsError` (true when the read fails or times out). `role`/`trialEndsAt` keep their exact current, unfixed behaviour on failure - see "Not fixed here" below. Fixed a real race found while building this (see "A regression I caught before shipping it").
- **New `lib/planBadgeCache.ts`**: a per-user, display-only cache of the last successfully-resolved plan. Cleared on sign-out (both the normal path and the local-only fallback when the sign-out network call itself fails).
- **`PlanBadge.tsx`**: on `entitlementsError`, shows the cached plan if one exists (recomputing `isTrial` against the live clock even from cached data - an actually-expired cached trial still correctly flips to Free). No cache (cold start) → renders nothing, never a false `FREE`.

## Security - stated explicitly, not left to infer

**This cache is read only by `PlanBadge` and never participates in any entitlement or access decision.** `entitled` is computed exactly the way it always was, from live `AuthProvider` state only - nothing here changes that computation or reads the cache for it. A cached value that could satisfy an access check would be a client-side paywall bypass; this PR does not create that path. If a future change makes anything but `PlanBadge` import from `planBadgeCache.ts`, that import is the review to stop at - the module's own header comment says so.

## The timeout value - measured, not guessed

Started at 8s (matching `SESSION_RESOLVE_MS`). **Building this fix, a real, unmocked load against this project's shared dev Supabase genuinely took longer than 8s and tripped the abort** - which would have shipped a second, self-inflicted version of the exact bug being fixed: a fast, wrong "error" on a request that was only ever slow, not actually failing. Moved to 15s, which cleared every subsequent real run with headroom. Generous deliberately: this timeout only matters when something is already wrong, so a too-long value costs a few extra seconds on an already-bad path, while a too-short one recreates the bug.

## A regression I caught before shipping it

Built the cache-write logic first, tested it, and the cache got silently overwritten with `{role: 'free'}` during the *failure* path - the exact bug this PR exists to prevent, reproduced by my own fix. Traced it with a disposable debug trace rather than guessing: there's one render frame, on every fresh sign-in, where a real `user` has resolved but `entitlementsLoading` still reads the stale `false` left over from the `!user` branch (which always runs once on mount, since `user` starts `null`) - the entitlements-fetch effect hasn't re-run yet to flip it back to `true`, because effects run after commit, not synchronously with the state update that changed `user`. A consumer reading `role` during exactly that frame sees the free/null defaults and reads them as resolved, not pending.

This is the same shape #1095 fixed for `entitlementsLoading` staying true through the *initial* load - just a narrower, single-render-frame instance neither that fix nor its regression tests (#1096/#1104, which widen the window via an artificial fetch delay) were positioned to catch, because this race is in a different pair of effects settling, not in the fetch's own duration. **Pre-existing since before this PR, not introduced by it** - my new cache-write logic is just the first thing sensitive enough to it to expose it. Fixed at the source: `setEntitlementsLoading(true)` now happens in the same synchronous block as `setUser` wherever a real user is assigned (the initial session resolve, and `onAuthStateChange`), closing the gap instead of working around it in a consumer.

## Verified

Built a disposable Playwright spec (not committed) using QA's own `signedInContext` (real minted session, read-only use of their fixture) - not a QA-owned file changed:
1. Loads once successfully (populates the cache via the real subscription read).
2. Force-fails the second load via route interception (500 response) - asserts the badge shows the cached plan, never `FREE`.
3. Separately, a cold-start case (different account, no prior successful load in this context) with a forced failure - asserts the badge renders nothing.

Both passed only after finding and fixing the race above - the naive version of this fix would have shipped with the cache silently corrupted by exactly the failure mode it exists to survive, undetected without this level of direct verification.

- `tsc --noEmit`: clean. `eslint`: 0 errors, 2 pre-existing warnings (unrelated lines). `npm test`: 794/794.

## Not fixed here - reported on #1089 per the standing ask

**A failed read still flips `entitled` to `false` today, not just the badge.** Confirmed by reading the code: `role`/`trialEndsAt` reset to their free/null defaults on any failure, feeding `entitled` the same way a genuinely-free account would - meaning a real Pro account could lose Pro *features* (Arena's Confluence card, both Alerts sections) during a backend hiccup, not just see a wrong label. Not fixed here: the right answer (fail open? fail closed? a cache like this one, but for real entitlement rather than display?) is a bigger decision than this PR should make silently, and it's a separate finding on #1089 rather than folded in.

## How to test (QA)

Sign in as Pro or Trial, confirm the badge shows correctly on a normal load. If you can force a `user_subscriptions` failure (route-blocking, the same technique used to verify this), confirm: first cold load with no prior success + a failure → badge absent, never `FREE`. A normal load followed by a forced failure → badge keeps showing the last-known-correct plan. Sign out → sign in as a different account on the same browser → confirm the new account's badge is correct, not a leftover from the previous one.

## Risk level

**Low-medium.** Touches shared `AuthProvider` state read by many consumers, but the only NEW behavior added to existing consumers is `entitlementsError` (additive, unused by anyone but `PlanBadge`) and the race fix (strictly narrows an existing gap, doesn't introduce new state transitions). The cache is additive and isolated to one component.

Related: #1089, #1095, #1096, #1104.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch content. Couldn't do a live browser reproduction this time and want to say so plainly rather than imply I did: this PR isn't merged, so deployed `qa` still runs the old code - a Playwright run against it tests nothing about this fix, and reproducing it live would mean standing up a local build, which is the exact memory-cost tradeoff you've been avoiding today for the same reason. Verified by direct source reading instead, to the same depth as the earlier repro:

**Security isolation, checked not assumed.** Walked every `.ts`/`.tsx` file in the branch for `planBadgeCache` imports - exactly two: `PlanBadge.tsx` (read/write) and `AuthProvider.tsx` (cleanup-only, `clearPlanBadgeCache` on sign-out). `entitled: isPro || isTrial` is computed purely from live `role`/`trialEndsAt` state, same as before this PR - nothing in the entitlement path reads the cache. Confirms the module's own "if anything but PlanBadge imports this, stop" warning is currently true.

**The race fix.** Both `setEntitlementsLoading(true)` calls sit in the same synchronous block as their paired `setUser` call (initial resolve and `onAuthStateChange`) - React 18 batches these into one commit regardless of the async context they're called from, so the one-frame gap the PR describes is genuinely closed, not narrowed.

**The timeout + catch combination.** `.abortSignal(AbortSignal.timeout(15000))` bounds the hang I reproduced. The `.then()` handler sets `entitlementsError` from a resolved `{data, error}` shape; the added `.catch()` covers the case where an abort genuinely rejects instead. Between the two, every outcome I found in my own repro (fast HTTP error, connection reset, indefinite hang) now settles `entitlementsLoading` either way - confirmed by reading both paths, not just the happy one.

**PlanBadge's fallback logic** - cold start with no cache returns `null` (not FREE), a cached trial is recomputed against the live clock rather than trusted as a stored boolean (so an actually-expired cached trial correctly flips to Free), and the cache-write effect explicitly skips while `entitlementsLoading || entitlementsError` so a failure can never overwrite a good cached value with a bad guess.

**The 15s value** - reasonable and disclosed as measured against this project's real dev-Supabase latency rather than picked to match `SESSION_RESOLVE_MS`'s 8s out of consistency; agree with not reusing that constant just because it existed.

On "not fixed here" - agreed this is the right call to leave for a separate #1089 decision rather than fold a real-entitlement fail-open/fail-closed policy into a badge-display PR.

Approved.

PR #1120, opened 2026-09-10
feat(nav): plan badge in the mobile header status slot (#1094)

JohnDominicJasmin:
## Summary

Adds PlanBadge to the mobile top bar's status slot (#1094). Owner ruled "top bar" placement; PM's brief asked for a measurement-first approach given a real content collision this PR found and fixes.

## What changed

- `components/TerminalNav.tsx`: `<PlanBadge />` now renders in the mobile 38px header (`.tnav-mhead`), after the trading-session status and before the drawer's hamburger button. Same component the desktop bar already uses.
- `app/globals.css`: `.tnav-mbrand` (the screen-name element) now truncates with an ellipsis (`min-width: 0; overflow: hidden; text-overflow: ellipsis; white-space: nowrap`) instead of letting the header grow past 390px. `.tnav-mstatus` gets `flex-shrink: 0` so it never yields. New `.tnav-mlogo` class (`flex-shrink: 0`) on the logo mark for the same reason.
- Two stale comments corrected while in the file: the `README:82` citation (pointed at an unrelated Landing-page value) now points at the artboards; the `#599` note claiming "nothing renders `.tnav-mhead` yet" is removed since `TerminalNav.tsx` has rendered it since #1094 work started (it was already live before this PR, just without the badge).

## Why

`.tnav-mhead` was already live (mounted at `NavDrawer.tsx:339` whenever `terminal && !onLanding` — true for every signed-in user since terminal became default, #748), just missing the badge. #1089 shipped the badge desktop-only, so a Pro/Trial user on a phone had no way to see their plan.

Naively appending the badge collides on real content: `screenNameFor()` falls back to the raw URL path segment (lowercase, hyphens→spaces) for ~15 routes without a curated nav-tab label — e.g. `/econ-calendar` renders literal `"econ calendar"`. Combined with a real trading-session status (`lib/session.ts`, up to `"Asia Session"`) and the widest real badge state (`TRIAL · 14D` — the 14-day trial window, wider than `PRO`), plus the real `.tnav-mmore` hamburger button (32px, not in the design mockups) that was never accounted for, the header overflows and clips the drawer button — a functional defect, not cosmetic.

Ruling (PM, on measurement): the screen name is the one element in that bar that's redundant with something the user already knows (they just navigated there). The status and the plan badge both carry information available nowhere else on a mobile screen, so when space runs out, the screen name yields via ellipsis and nothing else does. Verified this is what actually happens, not just what the CSS says should happen (see Verified below).

The route-fallback naming itself (`"econ calendar"` instead of a curated label) is a separate, pre-existing defect, not fixed here — filing it separately per PM.

## Verified

Measured with the real font file (`app/fonts/plex-mono-latin-*.woff2`) and real CSS values, not estimated:
- Real worst-case combination (`"econ calendar"` + `"Asia Session"` + `TRIAL · 14D` + the real hamburger button) overflows by ~19px before this fix — confirmed via `scrollWidth > clientWidth` in a real Chrome render, not inferred.
- After the fix, same worst-case content forced down to 320px (narrower than any real device this product supports): `scrollWidth === clientWidth`, hamburger button's bounding rect fully inside the header, `.tnav-mbrand`'s `scrollWidth > clientWidth` confirms the ellipsis is genuinely active, not just present in CSS.
- Screenshots taken via a disposable Playwright spec (not committed) using QA's own `signedInContext` against a real signed-in account, both themes, both the overflow route (`/econ-calendar`) and a short-label route (`/arena`) for contrast.
- Landing gate: `/` screenshotted in all 4 design/theme contexts plus a 390px terminal pass (the only width this CSS activates) — confirmed via source (`grep -c tnav-m` on `LandingTerminal.tsx`/`LandingContent.tsx` = 0) and visually that `.tnav-mhead` never renders on landing (`NavDrawer.tsx`'s `!onLanding` gate) and nothing else visibly changed.
- `tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794.
- `npm run build`: not run — memory constraint, per standing practice. Flagging since this PR touches shared nav chrome rendered on every signed-in page; QA's own build/deploy of `qa` is the first real build this sees.

## Not fixed here

Route-fallback screen names render the raw path segment in lowercase (`"econ calendar"`), not the curated ALL-CAPS labels the artboards show (`"CALENDAR"`). Pre-existing, unrelated to the badge, worth its own issue — happy to file it if PM wants it tracked.

## How to test (QA)

On the `qa` environment, sign in as a Pro or Trial account. At a phone-width viewport (≤767px):
1. Visit `/arena` (or `/dashboard`, `/scanner`, `/funding`, `/journal`) — confirm the plan badge shows next to the session status in the top 38px bar, and the hamburger (☰) button on the far right still opens the drawer.
2. Visit a route with no short nav label — `/econ-calendar`, `/live-tracking`, `/settings`, `/playbook`, etc. — during a live trading session (check the status text isn't blank). Confirm the screen name truncates with `…` rather than pushing the badge or hamburger off-screen, and the hamburger button is still fully visible and clickable.
3. Toggle light/dark theme on one of these pages — confirm the badge and header remain legible and nothing clips in either theme.
4. Sign out — badge and screen-name text behave the same as before (no login-state regression).

## Risk level

**Low-medium.** Touches shared nav chrome (`TerminalNav.tsx`, `globals.css`) rendered on every signed-in mobile page. No new state, no new data flow — reuses `PlanBadge` and `useAuth()` exactly as the desktop bar does. The CSS change is additive and scoped to `[data-design="terminal"]` inside a `max-width: 767px` media query; verified `/` unaffected. `npm run build` not run locally (memory constraint) — first real build is `qa`'s.

Related: #1094, #1089.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch - this closes #1094, which I filed, so verified carefully rather than taking the description at face value.

- Confirmed `screenNameFor()`'s fallback (`components/TerminalNav.tsx:124-129`) genuinely produces `seg.replace(/-/g, ' ')` for any route not in `ITEMS` - `/econ-calendar` really does render lowercase `"econ calendar"`, not a hypothetical worst case.
- Confirmed `NavDrawer.tsx:339`'s mount gate (`terminal && !onLanding`) is unchanged - the mobile header was already live before this PR, exactly as claimed, just without the badge.
- Confirmed zero `tnav-m` references in `LandingTerminal.tsx`/`LandingContent.tsx` - the landing-gate claim holds.
- `TerminalNav.tsx` now mounts `PlanBadge` twice - desktop (pre-existing, line 363) and the new mobile-header one (line 489) - same component, no new logic.

Good instinct on the priority ruling (screen name yields, status and badge don't) - that's the right call given the screen name is the one element here the user already knows. Also appreciate flagging the route-fallback naming as a separate, real defect rather than quietly fixing it inside this PR or ignoring it.

Approved. Once this reaches `qa`, my `plan-badge-mobile.spec.ts`'s `test.fail()` on the terminal case should flip to an unexpected pass - I'll pull that annotation as a fast-follow once I've confirmed it live.

PR #1122, opened 2026-09-10
chore(dashboard): remove current-design branching from 5 Pattern B pages

JohnDominicJasmin:
## Summary

First batch of #1111's Pattern B (mechanical wrapper-class pages), per PM's staged order (step 3, after the CSS token audit gate and Pattern E). 5 of the ~14-16 listed pages: `/econ-calendar`, `/alerts`, `/upgrade`, `/calc`, `/faq`.

## What changed

Each page had exactly one `useDesignMode()` read driving one ternary: `className={mode === 'terminal' ? 'x-term-wrap' : undefined}`. Removed the hook, the import, and the ternary — the class is now applied unconditionally.

## Why

Terminal is the only design going forward (#1111). Confirmed before editing that each file had no OTHER design-mode branching (grepped for every `mode` occurrence, not just the ternary shape, in each file).

## Verified

**This is provably a no-op**, not assumed:
- Every one of these wrapper classes is only ever referenced in `globals.css` as `[data-design="terminal"] .x-term-wrap` (checked all 5: `alerts-term-wrap`, `calc-term-wrap`, `econ-term-wrap`, `faq-term-wrap`, `upgrade-term-wrap`). The class has zero CSS effect unless the ancestor `<html>` already carries `data-design="terminal"` — this PR doesn't touch that attribute or `DesignModeProvider` at all.
- So: a terminal-mode viewer had the class before (ternary resolved true) and has it now (unconditional) — identical. A current-mode viewer (via `?design=current` or a stale `localStorage` preference) gets the class added but it still matches nothing, since the ancestor attribute isn't `terminal` for them — also identical.
- Empirically confirmed, not just reasoned from source: stash-compared `/faq?design=terminal` on the same running dev server (kill by PID, clear `.next`, clean restart each side, per standing practice) — before and after screenshots are pixel-identical (only the live session countdown text differs between the two captures, expected/unrelated). Also checked plain `/faq` (this browser profile's persisted `current` mode) renders unchanged.
- `tsc --noEmit`: clean. `eslint`: 0 errors (pre-existing warnings elsewhere in touched files, unrelated to these lines). `npm test`: 794/794.
- `npm run build`: not run — memory constraint, per standing practice.

Per #1111 §5's adopted approval model: an empty screenshot diff against terminal's rendering means condition 5 (owner visual approval) doesn't bind — appearance provably did not change for any real viewer.

## How to test (QA)

On `qa`, visit each of `/econ-calendar`, `/alerts`, `/upgrade`, `/calc`, `/faq` normally (terminal is default) and confirm no visual change from what's currently live. Optionally check `?design=current` on each too — also unchanged, for the reason above.

## Risk level

**Low.** Mechanical, scoped to 5 files, each a single-line removal. No behavior change for any real user — see Verified above for why this is provable rather than assumed.

Related: #1111 (staging plan), #1109 (Arena carve-out, not touched here).


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch. Confirmed each of the 5 files (alerts, calc, econ-calendar, faq, upgrade) has zero remaining `useDesignMode`/`mode ===` occurrences - the ternary really was the only branching in each. Confirmed every `*-term-wrap` class (`alerts-term-wrap`, `calc-term-wrap`, `econ-term-wrap`, `faq-term-wrap`, `upgrade-term-wrap`) is scoped ONLY to `[data-design="terminal"]` in globals.css - which makes this a mathematically provable no-op, not just a well-reasoned one: a terminal viewer gets the same matching selector either way, a current-mode viewer gets a non-matching class either way.

Good discipline on the stash-compare pixel diff too, rather than resting on the CSS-scoping argument alone.

Approved.

PR #1124, opened 2026-09-10
chore(dashboard): remove current-design branching from 6 more Pattern B pages

JohnDominicJasmin:
## Summary

Second batch of #1111's Pattern B (mechanical wrapper-class pages): `/forgot-password`, `/login`, `/reset-password`, `/terms`, `/research`, `/playbook`. First batch (5 pages) was #1122.

## What changed

Same shape as #1122: each file had exactly one `useDesignMode()` read driving one wrapper-class ternary. Removed the hook, the import, and the ternary — the class is now applied unconditionally. `reset-password` had the ternary twice (two render branches, same shape) — both collapsed the same way.

## Why

Terminal is the only design going forward (#1111). Confirmed before editing that each file had no other design-mode branching — grepped every `mode` occurrence per file, not just the ternary shape. None needed to move to Pattern C.

## Verified

**The structural no-op proof from #1122 applies here too, stated by reference rather than re-derived**: every wrapper class this touches (`login-term-wrap`, `terms-term-wrap`, `research-term-wrap`, `playbook-term-wrap`) is only ever referenced in `globals.css` as `[data-design="terminal"] .x-term-wrap` — checked all 4 for this batch. The class has zero effect without the ancestor attribute, which this PR doesn't touch. Applying it unconditionally is inert in both directions (terminal-mode viewer: had it already; current-mode viewer: still matches nothing).

One spot-check screenshot per batch, per the adopted model (mechanism is proven, the check is for the case the mechanism assumption is wrong): stash-compared `/login?design=terminal` at a fixed 1000×900 window (kill by PID, clear `.next`, clean restart on each side) — before/after pixel-identical. Also confirmed `data-design="terminal"` was genuinely set during the capture (read via JS, not assumed from the query param) after the render looked unexpectedly light-themed — it was terminal's light-theme variant, not a bug, just a different theme than the dark screenshots taken earlier today.

Per #1111 §5: empty screenshot diff means condition 5 (owner visual approval) doesn't bind for this batch.

`tsc --noEmit`: clean. `eslint`: 0 errors (2 pre-existing warnings in `playbook`/`reset-password`, unrelated lines — `react-hooks/set-state-in-effect` on code this PR didn't touch). `npm test`: 794/794. `npm run build`: not run — memory constraint, per standing practice.

## How to test (QA)

On `qa`, visit each of the 6 pages listed above normally and confirm no visual change. Optionally check `?design=current` too — also unchanged.

## Risk level

**Low.** Mechanical, 6 files, same pattern as #1122. No behavior change for any real viewer.

Related: #1111, #1122 (batch 1).


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch. The "4, not 6" in the Verified section checks out - `login-term-wrap` is genuinely shared across forgot-password/login/reset-password (confirmed in the diff), so 4 distinct classes across 6 files is correct, not an undercount.

Confirmed all 4 classes (`login-term-wrap`, `terms-term-wrap`, `research-term-wrap`, `playbook-term-wrap`) are scoped only to `[data-design="terminal"]` in globals.css, and all 6 files have zero remaining `useDesignMode`/`mode ===` occurrences. Same provable-no-op shape as #1122.

Approved.

PR #1126, opened 2026-09-10
fix(dashboard): give the FREE plan badge a filled treatment

JohnDominicJasmin:
## Summary

Fixes the FREE plan badge blending into surrounding nav chrome, per owner feedback (relayed by PM): *"It's still the same design as the item besides the language and the profile. It has similar design. That's why it's not being highlighted."*

## What changed

`.plan-badge-free`: filled (`var(--bg4)` background, `var(--txt)` text, `var(--bdr2)` border) instead of transparent with grey text. Same 11px size, unchanged.

## Why

An outlined grey badge sat next to `EN` (grey text) and the avatar (a bordered square) — three grey-outlined things in a row, reading as one more nav control rather than a statement about the account. A lighter grey would still have been chrome; the fix isn't brightness, it's fill vs. outline.

**Principle encoded here, for future badge/chip work in this nav**: outlined means control, filled means label. The plan badge is a label. Amber stays reserved for the paid states — FREE gets no amber, so a free account can never read as Pro at a glance. Ladder: FREE filled-neutral, TRIAL amber outline + day count, **PRO amber filled** (unchanged in this PR — see note below).

## A correction, checked before touching anything

The brief for this fix assumed PRO currently renders as an outline. **It doesn't** — `.plan-badge-pro` is still solid-filled amber (`#0a0a0a` text on `var(--amber)` background), exactly as it shipped originally. The owner-approved amber-outline variant for PRO (from an earlier round) was approved but never actually implemented — a separate, still-open task, not touched here. Flagging this rather than silently building on the wrong premise, since it changes whether "FREE louder than PRO" is even a real risk: **it isn't**, because PRO's solid amber fill is unambiguously louder than FREE's neutral dark fill (confirmed visually, not assumed — see below). If PRO's outline variant ships later, this comparison should be re-checked at that point.

## Verified

Mocked all three states side by side in the real desktop nav (`.tnav-right`, injected via a disposable JS snippet — not committed), in both dark and light theme, rather than three separate screenshots:

- **Dark**: FREE (dark fill, white text) / TRIAL (amber outline) / PRO (solid amber fill) — all three instantly distinguishable at a glance.
- **Light**: same three states, same distinction holds.

FREE no longer reads as chrome; PRO remains unambiguously the loudest of the three (solid saturated amber vs. FREE's neutral dark fill) — the ladder is not inverted.

**Landing gate** (touches `app/globals.css`): checked `/` in all 4 design/theme contexts (`current`+dark, `current`+light, `terminal`+dark, `terminal`+light) — unchanged in every one. Confirmed empirically, not just reasoned: `.plan-badge` never even mounts on landing (`PlanBadge` returns `null` for a signed-out visitor, and landing is always signed-out), and separately `.app-bar`/`.tnav` are unreachable there too (`body.landing` CSS-hides `.app-bar`; `TerminalNav` is excluded from landing routes at the component level). Checked `document.querySelector('.plan-badge')` was `null` on `/` directly, not just inferred from the class name.

`tsc --noEmit`: clean (CSS-only change, ran as a sanity check). `npm test`: 794/794. `npm run build`: not run — memory constraint, per standing practice.

## How to test (QA)

Sign in as a Free account. Look at the top-right nav (desktop, `.tnav-right`) or the mobile header status area (#1094/#1120). Confirm the FREE badge reads as a distinct filled label, not blended with the language switcher or sign-in/avatar area, in both light and dark theme. Optionally compare visually against a Trial or Pro account on the same screen to confirm the three states stay distinguishable.

## Risk level

**Low.** Single CSS rule, 3 property changes, no new tokens (all four already redeclared correctly in every theme×design combination — checked before using them). No behavior change, purely visual.

Related: #1089 (PlanBadge), #1094 (mobile placement).


JohnDominicJasmin:
QA Team

Reviewing for readiness, not signing off on merge - noted PM's hold while the owner compares 3 treatments.

Verified the correction claim: `.plan-badge-pro` (globals.css:4245) really is already solid-filled amber (`background: var(--amber)`, near-black text), not an outline - confirmed against current source, the PR's premise-check holds. Confirmed `--bg4`, `--txt`, `--bdr2` are all declared in the base, light-theme, terminal-dark, and terminal-light blocks - no missing-token risk (the exact #783/#559 class of bug this migration keeps guarding against).

Holding formal approval until the owner picks a direction - will re-review promptly once it's the chosen option, or this closes if it isn't.

JohnDominicJasmin:
**Dev Team.** Superseded, not rejected — closing without merging.

The owner reviewed all three Free/Pro treatments side by side in the real nav (this PR is option C: both states filled, differentiated by weight) and picked **option A**: PRO gets a small star emblem inside its existing filled badge; FREE becomes plain text with no chip at all, distance coming from category rather than colour.

`.plan-badge-free`'s CSS in this branch (filled background) is the opposite of what ships — option A removes FREE's chip entirely rather than filling it. Building A fresh in a new PR rather than amending this one.

PR #1128, opened 2026-09-10
chore(dashboard): collapse the dashboard route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

First of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/dashboard`.

## What changed

`app/dashboard/page.tsx` goes from ~630 lines (a full current-design implementation plus a `mode === 'terminal'` swap to `DashboardTerminal`) to a 12-line wrapper that always renders `DashboardTerminal`.

`DashboardTerminal` takes no props and is fully self-contained, including its own independent copy of the onboarding-tour handling (`SpotlightTour`, `useOnboarding`, `data-spotlight-section`) — checked before deleting this file's copy, so nothing is lost. Real users have only ever seen `DashboardTerminal` here since terminal became default (#748).

**Decision** (per #1111 §3's "real decision" requirement for Pattern A): kept `DashboardTerminal.tsx` as its own file rather than inlining its ~600 lines into `page.tsx`. It's a substantial standalone implementation; moving it wholesale into the route file would just relocate it with no benefit, and every other converted terminal component in this codebase keeps the same `*Terminal.tsx` separation.

## Verified

Full-page stash-compare screenshot using a real signed-in session (kill by PID, clear `.next`, clean restart each side): structurally and visually identical — same layout, same cards, same data shape. Only live market values (prices, session countdown) differ between the two captures, expected and unrelated.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794 (see note below). `npm run build`: not run — memory constraint, per standing practice.

## A coupled fix that landed first (#1127)

This collapse removed 2 real defects `__tests__/terminalTypographyOwnership.test.mts`'s `WIDE_BASELINE` ratchet was tracking for this exact file. That created a genuine two-sided deadlock (documented in #1127, merged into `dev` first and rebased into this branch): neither this fix nor QA's baseline cleanup could land alone under the old hard-fail ratchet, and the fix belonged in the test mechanism, not a workaround on either side. QA will drop the 2 now-stale `WIDE_BASELINE` entries for this file as a follow-up commit; this PR carries only the app-code change.

## How to test (QA)

Visit `/dashboard` signed in — unchanged from what's currently live.

## Risk level

**Low.** Deletes dead code only (current design hasn't been reachable here since #748). No behavior change for any real user, verified via stash-compare rather than assumed from the branch shape.

Related: #1111, #1127.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch. Confirmed the new `app/dashboard/page.tsx` is genuinely a thin wrapper (`export default function Dashboard() { return <DashboardTerminal />; }`, no props passed) and that `DashboardTerminal()` (components/DashboardTerminal.tsx:520) independently imports and uses `useOnboarding`, `SpotlightTour`, and `data-spotlight-section` - its own complete onboarding-tour handling, nothing lost from the deleted copy.

One thing worth flagging rather than silently correcting: I re-ran the ratchet test against this branch's actual combined state (both the dashboard change and #1127's fix, properly rebased - confirmed in the branch's own git log) and got **7/7 pass with both stale entries printing as diagnostics**, not the "793/794" the original commit message describes. That number was written before the rebase onto #1127 and is now stale text in an otherwise-correct commit - not a real defect, just noting it so nobody reads the PR later and wonders why the count doesn't match current behavior.

Approved - and I'll land the WIDE_BASELINE cleanup as my own follow-up now that this is confirmed merged-ready.

PR #1130, opened 2026-09-10
chore(briefing): collapse the briefing route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Second of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/briefing`.

## What changed

`app/briefing/page.tsx` goes from ~680 lines (a full current-design implementation — including its own `buildBriefingContext()` and all the AI-brief/JPY-rate state — plus a `mode === 'terminal'` swap to `BriefingTerminal`) to a 12-line wrapper that always renders `BriefingTerminal`.

**Decision** (per #1111 §3): kept `BriefingTerminal.tsx` as its own file, same reasoning as #1128 (dashboard) — it's a substantial standalone implementation (775 lines), not worth relocating into the route file.

## Verified

`BriefingTerminal` takes no props and has its own fully independent copy of every piece of page.tsx's state before deleting it — same hook shapes, same `sessionStorage` keys (`lhq_brief`/`lhq_brief_ts`), its own `buildBriefingContext()` (confirmed both files define it locally, neither imports from the other). Real users have only ever seen `BriefingTerminal` here since terminal became default (#748).

Screenshot-compared via stash (kill by PID, clear `.next`, clean restart each side): identical layout, cards, and structure in both states. One methodology note: Playwright's own signed-in flow hit the shared machine's current memory pressure mid-run (background task killed for low memory, same pattern PM flagged earlier today) — fell back to the lighter Chrome extension tool against the same dev server, signed out, which is sufficient here since the comparison is code-state vs. code-state on the same unauthenticated render, not a claim about full real-data parity.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794 (no `WIDE_BASELINE` collision for this file — checked). `npm run build`: not run — memory constraint, more pressing than usual per the note above.

## How to test (QA)

Visit `/briefing` signed in — unchanged from what's currently live.

## Risk level

Related: #1111, #1128 (dashboard, same pattern).


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch. Confirmed the new `app/briefing/page.tsx` is a genuine thin wrapper, and `BriefingTerminal()` (components/BriefingTerminal.tsx:181) takes no props, owns its own `buildBriefingContext()` (line 111), and directly uses the same `lhq_brief`/`lhq_brief_ts` sessionStorage keys - independent state, nothing lost. No `WIDE_BASELINE` entry existed for this file (confirmed), and zero remaining `useDesignMode`/`mode ===` in the new page.tsx.

Noted the memory-pressure fallback to the browser extension for the screenshot check - reasonable given today's standing constraint, and the comparison scope (code-state vs code-state, unauthenticated) is honestly stated rather than oversold as full parity.

Approved.

PR #1131, opened 2026-09-10
chore(markets): collapse the markets route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Third of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/markets`.

## What changed

`app/markets/page.tsx` goes from ~420 lines (a full current-design implementation — search/sort/pagination over the coin table — plus a `mode === 'terminal'` swap to `MarketsTerminal`) to a 12-line wrapper that always renders `MarketsTerminal`.

**Decision** (per #1111 §3): kept `MarketsTerminal.tsx` as its own file, same reasoning as #1128/#1130.

## Verified

`MarketsTerminal` takes no props and is fully self-contained. Unlike dashboard/briefing, this page's deleted branch had **no auth-dependent state** — `useMarket()` is public market data, everything else was local UI state (search/sort/page). That makes the signed-out comparison below conclusive for this page specifically, not just a convenient shortcut — stated per PM's ask to name the method and why it's sufficient for each page rather than let it become a default by habit.

**Method used**: stash-compare screenshot via the Chrome extension against the running dev server, signed out (kill by PID, clear `.next`, clean restart each side) — no second Next build, no second Chromium, given the shared machine's current memory pressure. Identical table layout, columns, sort controls, grade/signal formatting between before and after. Only live prices/24h-change values differ, expected.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794 (no `WIDE_BASELINE` collision for this file). `npm run build`: not run — memory constraint, per standing practice.

## How to test (QA)

Visit `/markets` — unchanged from what's currently live. Search, sort, and pagination all behave the same as before this PR.

## Risk level

Related: #1111, #1128, #1130.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch. Confirmed the thin wrapper, `MarketsTerminal()` (line 43) takes no props and calls `useMarket()` directly for its data - public market state, no auth dependency, which is exactly why the signed-out comparison method is conclusive here specifically rather than a shortcut. No `WIDE_BASELINE` entry for this file, zero remaining `useDesignMode`/`mode ===`.

Appreciate naming the method and why it's sufficient per-page rather than defaulting to one approach for every slice - that's the right response to the methodology note from earlier today.

Approved.

PR #1133, opened 2026-09-10
chore(liquidation-map): collapse the liq route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Fourth of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/liq`.

## What changed

`app/liq/page.tsx` goes from ~650 lines (a full current-design implementation — band estimation, whale/retail/Bybit positioning fetches — plus a `mode === 'terminal'` swap to `LiqTerminal`) to a 12-line wrapper that always renders `LiqTerminal`.

**Decision** (per #1111 §3): kept `LiqTerminal.tsx` as its own file, same reasoning as the prior three pages.

## Verified

`LiqTerminal` takes no props and mounts its own `<LiqFeed onClusters={...} coinFilter={...} />` — non-headless, confirmed by reading the file before deleting this route's copy — the same visible mount the deleted branch had. No auth-dependent state in the deleted branch, so the signed-out comparison is conclusive here too.

**Method**: stash-compare screenshot via the Chrome extension against the running dev server, signed out (kill by PID, clear `.next`, clean restart each side). Identical header, filter controls (coin dropdown, 12h/48h/etc.), and cluster-bar layout. The "Live Liquidations" widget showed a transient "No open interest data" loading state in one capture and settled real data in the other — timing, not a code difference, same pattern as live price values on the other Pattern A pages.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794. `npm run build`: not run — memory constraint, per standing practice.

## A coupled fix that landed first (#1132)

This collapse broke `__tests__/liqFeedHeadless.test.mts`'s "Arena mounts it headless and /liq does not" assertion — it grepped `app/liq/page.tsx`'s own source text for a non-headless `<LiqFeed>` tag, which moved to `LiqTerminal.tsx` with this refactor. Not a regression (see Verified above): the mount is real and correct, just in a different file now. QA fixed the test properly — `routeComposedSource()` resolves what a Pattern-A-shaped route actually renders via one bounded hop through its own import, rather than hardcoding a file path (merged in #1132, rebased into this branch). Full writeup in `qa/TEST_GAPS.md` §9.

## How to test (QA)

Visit `/liq` — unchanged from what's currently live. Coin selector, time range, and cluster bars all behave the same as before this PR.

## Risk level

Related: #1111, #1128, #1130, #1131, #1132.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch, including the coupled #1132 rebase - not just trusting the writeup. Confirmed the thin wrapper, and ran the real `liqFeedHeadless.test.mts` against this exact branch's combined state (both the collapsed `app/liq/page.tsx` and `routeComposedSource()`): all 5 tests pass clean, including the one this PR would have broken pre-#1132. Confirms the resolver genuinely works end to end, not just in my own isolated simulation when I built it.

Approved.

PR #1134, opened 2026-09-10
chore(scanner): collapse the scanner route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Fifth of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/scanner`.

## What changed

`app/scanner/page.tsx` goes from a 34-line current-design implementation (a plain composition of `AccumulationTracker`/`DistributionTracker`/`CoinHeatmap`/`DrawdownChart`/`MultiTFSqueezeView`/`SignalAccuracy`/`SetupScanner`, no extra state of its own) plus a `mode === 'terminal'` swap to `ScannerTerminal`, to an 11-line wrapper that always renders `ScannerTerminal`.

**Decision** (per #1111 §3): kept `ScannerTerminal.tsx` as its own file, same reasoning as the prior four pages. This one was the smallest/simplest of the Pattern A pages so far — no hooks, no local state to verify preservation of, just component composition.

## Verified

`ScannerTerminal` takes no props and is fully self-contained. No auth-dependent state in the deleted branch, so the signed-out comparison is conclusive.

**Method**: stash-compare screenshot via the Chrome extension against the running dev server, signed out (kill by PID, clear `.next`, clean restart each side). Identical card layout — Accumulation Tracker, Distribution Tracker, Market Heatmap, same filter chips. Row counts and highlighted coins differ between the two captures because this is a live scanner continuously re-evaluating its own criteria, not a code difference.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794. `npm run build`: not run — memory constraint, per standing practice.

Per #1111 §5: empty diff means condition 5 doesn't bind for this slice.

## How to test (QA)

Visit `/scanner` — unchanged from what's currently live.

## Risk level

**Low.** Deletes dead code only (current design hasn't been reachable here since #748). No behavior change for any real user. Simplest of the Pattern A pages so far — no hooks or state to reason about beyond the swap itself.

Related: #1111, #1128, #1130, #1131, #1133.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YVjcfLMLXdmnkB3Nwwq43Y

PR #1135, opened 2026-09-10
chore(funding): collapse the funding route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Sixth of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/funding`.

## What changed

`app/funding/page.tsx` goes from ~700 lines (a full current-design implementation — per-coin funding rate history, sparklines, full chart canvas drawing — plus a `mode === 'terminal'` swap to `FundingTerminal`) to a 12-line wrapper that always renders `FundingTerminal`.

**Decision** (per #1111 §3): kept `FundingTerminal.tsx` as its own file, same reasoning as the prior five pages.

## Verified

`FundingTerminal` takes no props and is fully self-contained (its own canvas refs, its own fetch/draw effects). No auth-dependent state in the deleted branch, so the signed-out comparison is conclusive.

**Method**: stash-compare screenshot via the Chrome extension against the running dev server, signed out (kill by PID, clear `.next`, clean restart each side). Identical regime-overview chips, funding-rate table, sparklines, and chart layout. Values differ between captures — live data — and the before-capture happened to catch the page before its own fetch settled (same timing pattern seen on the liq and scanner Pattern A pages), both expected.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794 (this file's `WIDE_BASELINE` entry — `.card` inline padding — reports via `t.diagnostic()` per #1127's fix, doesn't fail). `npm run build`: not run — memory constraint, per standing practice.

## How to test (QA)

Visit `/funding` — unchanged from what's currently live.

## Risk level

Related: #1111, #1128, #1130, #1131, #1133, #1134.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch (checked out the full tree, not a piecemeal file swap, after that approach tripped me up on an unrelated inconsistency mid-review). Confirmed the thin wrapper, `FundingTerminal()` (line 326) takes no props, zero remaining `useDesignMode`/`mode ===`.

Ran the real ratchet test against this branch's full combined state: 7/7 pass, and `app/funding/page.tsx`'s entry correctly reports as stale via `t.diagnostic()` rather than failing, exactly as claimed. Noticed `app/liq/page.tsx`'s entry is also now stale (from #1133, already merged) - that's a separate cleanup, not this PR's concern, I'll land it as its own follow-up.

Approved.

PR #1137, opened 2026-09-10
chore(correlations): collapse the correlation route to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Seventh of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `/correlation`.

## What changed

`app/correlation/page.tsx` goes from ~420 lines (a full current-design implementation — correlation matrix, ranked pairs, alt-season signal, its own `fetchCloses()`) plus a `mode === 'terminal'` swap to `CorrelationTerminal`, to a 13-line wrapper that always renders `CorrelationTerminal`.

**Decision** (per #1111 §3): kept `CorrelationTerminal.tsx` as its own file, same reasoning as the prior six pages.

## Verified

`CorrelationTerminal` takes no props and has its own independent copy of every piece of state. Notably, both files' `fetchCloses()` were already migrated to the shared Bybit retry helper together earlier this session (#1100, part of #1080's rollout) — the two implementations have been kept deliberately in sync, not just coincidentally similar. No auth-dependent state in the deleted branch.

**Method**: stash-compare screenshot via the Chrome extension against the running dev server, signed out (kill by PID, clear `.next`, clean restart each side). Identical Macro Correlations card and correlation-matrix layout. The before-capture caught the matrix mid-fetch (same timing pattern as the last three Pattern A pages); the after-capture landed after it settled — expected, not a code difference.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794 (this file's two `WIDE_BASELINE` entries report as stale diagnostics per #1127, don't fail). `npm run build`: not run — memory constraint, more pressing than usual on this machine right now (two of this branch's own pushes were killed for low memory before landing — retried each time, both eventually succeeded).

## How to test (QA)

Visit `/correlation` — unchanged from what's currently live.

## Risk level

Related: #1111, #1128, #1130, #1131, #1133, #1134, #1135.


JohnDominicJasmin:
QA Team

Reviewed against the actual PR branch (full-tree checkout, learned from earlier today's piecemeal-swap mistake). Confirmed the thin wrapper, `CorrelationTerminal()` (line 157) takes no props and has its own `fetchCloses()` (line 25) - matches what I already verified independently in #1100's review earlier today, not just trusting the claim here.

Ran the real ratchet against this branch: 7/7 pass, both `app/correlation/page.tsx` entries correctly report as stale diagnostics, exactly as claimed.

Approved. And no complaints about the memory-kill retries - that's the machine, not the work; glad both landed.

PR #1139, opened 2026-09-10
chore(nav): collapse NavDrawer's current-design chrome, PlanBadge mount included

JohnDominicJasmin:
## Summary

The nav-shell member of #1111's Pattern A group: `components/NavDrawer.tsx`. Structurally different from the 7 route pages already done — this is one file rendered on every page, mixing current-design chrome, terminal's `TerminalNav`, and a shared off-canvas drawer used by both, rather than a route swapping between two `*Terminal.tsx` twins.

## What changed

Removed both blocks previously gated on `mode !== 'terminal'`:
- The current-design `.app-bar` (desktop nav) — brand mark, nav links + dropdowns, session pill, theme toggle, language switcher, **its own `PlanBadge` mount**, auth dropdown, hamburger opener.
- The current-design `.mobile-tab-bar` (4-tab bottom bar + More button).

`TerminalNav` now mounts unconditionally (still gated on `!onLanding`, unrelated to design mode). The shared off-canvas drawer (`#nav-drawer`, gated on `!isDesktop`) is **untouched** — both the old `.app-bar` and `TerminalNav` opened the same drawer via the same `onOpenDrawer` prop, so nothing about it needed to change.

Also removed as dead code once both blocks were gone: `SessionPill` (+ its `pad2`/`findEndsInMs` helpers), `useStatusDot`, `NavDropdown`/`DropKey`, the 5 mobile-tab-bar icon components, the `openDrop`/`authOpen` state and their click-outside effects, and the now-permanently-`false` `usageOpen` state + its `<UsageModal/>` mount (`TerminalNav` has its own separate `UsageModal` instance already).

Verified every removal by grep before deleting — each symbol's full usage list, not assumed from a first read. Caught one real mistake of my own mid-edit: an `Edit` call left a dangling unclosed `useEffect(() => {` (an artifact of how the old_string boundary was chosen) — caught by `tsc`/`eslint` immediately after, fixed before it went anywhere.

## The `PlanBadge` mount, addressed explicitly per PM's request

`NavDrawer`'s own `PlanBadge` instance (inside the deleted `.app-bar`) is gone. `TerminalNav`'s two instances (desktop `.tnav-right`, mobile `.tnav-mhead`, from #1090/#1120) are untouched and are now the **only** mounts in the app — the correct end state regardless of which of the 3 pending Free/Pro badge treatments (#1126, held on the owner's decision) ships, since none of them change *where* `PlanBadge` mounts, only its own internal CSS/markup. No conflict between this PR and #1126.

## Verified

`tsc --noEmit`: clean. `eslint`: 0 errors (1 pre-existing warning, unrelated line, untouched by this change). `npm test`: 794/794.

**Terminal-mode** (the real state every user has seen since #748): stash-compare screenshot via the Chrome extension against the running dev server (kill by PID, clear `.next`, clean restart each side) — pixel-identical to before. Genuinely a no-op.

The drawer-open interaction itself (`TerminalNav`'s mobile "more" button → `onOpenDrawer` → `setDrawerOpen(true)` → the shared drawer opens) wasn't re-tested live at a real mobile viewport — this environment's window-resize tool doesn't actually change the browser's content viewport here (confirmed via `window.innerWidth`, stuck at the desktop size regardless of the resize call). Not skipped silently: the prop expression passed to `TerminalNav` (`onOpenDrawer={() => setDrawerOpen(true)}`) is byte-identical text before and after this diff, `TerminalNav.tsx` itself is untouched, and the drawer's own render logic (gated on `!isDesktop`, starting at the line right after the `TerminalNav` mount) is untouched too — the full interaction chain was already proven correct by #1094's own Playwright-based mobile testing earlier this session. The only change here is *when* `TerminalNav` mounts (now unconditional instead of `mode === 'terminal'`-gated), and `mode` has resolved to `'terminal'` for every real user since #748 regardless — so there's no reachable path where this changes what actually renders for anyone.

## A real, pre-existing finding — not introduced by this PR, but surfaced while checking it

`?design=current` has been silently broken as a *visual* rollback since the first Pattern A merge (#1128), not just this one. `current` mode removes the `data-design` attribute entirely (`lib/designMode.ts`'s own comment: *"There is no `[data-design="current"]` block and there should never be one"*). Every already-collapsed route's `*Terminal.tsx` now mounts unconditionally, but its CSS is scoped under `[data-design="terminal"]` — so under `?design=current`, those routes render as raw, unstyled HTML. Confirmed on `/markets?design=current` (already merged, #1131) *before* touching this file, so this isn't NavDrawer-specific — it's mechanical and applies identically to all 7 already-merged Pattern A pages (dashboard #1128, briefing #1130, markets #1131, liq #1133, scanner #1134, funding #1135, correlation #1137).

**This PR compounds it**: previously `?design=current` at least showed a *styled* old `.app-bar` above an unstyled body — ugly but partially legible. After this PR, the nav is unstyled too, so the whole page is raw HTML under that flag.

**PM's ruling on #1111**: accepted as lost from step 2 onward, not something to route around. §2/§6's original requirement — keep `?design=current` working as a real visual rollback through steps 2–7 — assumed collapsing a route could preserve its current-mode render, which was never actually possible once "collapse" means "mount the Terminal component unconditionally, whose styling is scoped to an attribute `current` deliberately removes." Two independent reasons it's acceptable: production (`main`) has none of this queued, and everything verified today depended on comparing terminal-mode before/after against itself, never on the cross-design visual comparison actually working. Provider/param/script removal (#1111 step 8) still goes last — but now because it touches every route at once and belongs after each is individually verified, not because of the rollback argument, which no longer holds.

## How to test (QA)

Visit any signed-in or signed-out page normally (terminal, the default) — nav should be unchanged from what's currently live, in both mobile and desktop layouts. `?design=current` is now known-broken on every Pattern A page including this one — not a regression to chase, already accepted per the ruling above.

## Risk level

**Low for real users** (provable no-op, same reasoning as every prior Pattern A page). **Worth naming plainly**: this PR is the one that makes the `?design=current` degradation total rather than partial, on every page in the app, since NavDrawer is shell-level. Flagged here rather than left for whoever notices next.

Related: #1111, #1128, #1130, #1131, #1133, #1134, #1135, #1137, #1090, #1120, #1126 (held, unaffected by this PR).


JohnDominicJasmin:
Reviewed against `origin/chore/remove-current-design-navdrawer`, full-tree checkout.

**PlanBadge migration** — zero `PlanBadge` references left in this file (grep on PR branch). `TerminalNav.tsx` still carries exactly 2 mounts (desktop `.tnav-right`, mobile `.tnav-mhead`) — untouched by this PR, confirmed correct target for the badge this file used to own.

**Dead code removal** — zero remaining references anywhere in the codebase to `SessionPill`, `useStatusDot`, `NavDropdown` (full-repo grep, clean).

**Type/lint** — `tsc --noEmit` clean. `eslint components/NavDrawer.tsx`: 1 warning (`react-hooks/set-state-in-effect`, line 93, `setNavQuery('')` inside the `drawerOpen` effect). Checked whether this is new: diffed `origin/dev` vs this branch — that exact effect block is untouched context in the diff, and `dev`'s current `NavDrawer.tsx` already has this warning (plus a second one, from the now-deleted app-bar effect, that this PR removes). Net: 2 warnings on `dev` → 1 on this branch. Not a regression.

**`?design=current` breakage claim** — verified against source, not just the PR body. Every `.tnav*` rule in `app/globals.css` is gated `[data-design="terminal"] .tnav...` (480 gated declarations, zero unscoped `.tnav` rule). So the claim holds: once a route unconditionally mounts `TerminalNav`, `?design=current` renders that tree with none of its CSS applying. Confirmed technically accurate.

On "PM's ruling on #1111: accepted as lost from step 2 onward" — pinging PM directly to confirm that's a real ruling before treating the `?design=current` escape hatch as settled-dead project-wide; not blocking this PR on it since the regression is real either way and scoped identically across all 8 Pattern A pages already merged today, not introduced here.

PR #1141, opened 2026-09-10
chore(landing): collapse LandingContent to its terminal-only wrapper

JohnDominicJasmin:
## Summary

Eighth and last of #1111's Pattern A pages (8 full-component-swap pages + NavDrawer): `components/LandingContent.tsx`, the `/` marketing page.

## What changed

Deleted ~250 lines of current-design's own marketing page JSX (nav, hero, features, how-it-works, pricing, final CTA, footer) plus `FEATURE_META` and its now-dead-only imports (`Link`, `BrandMark`, `BeamsBackground`, `LanguageSwitcher`, `Fragment`, `ReactNode`). `LandingTerminal` now renders unconditionally after the loading-state early return. The loading state's own inline-style ternaries (`mode === 'terminal' ? {...} : undefined`) collapsed to just the terminal branch's values.

**What stayed, and why**: the redirect-to-`/arena` effect and the `body.landing`-class/`lang`/`dir` effect were never inside the deleted branch — they ran unconditionally for both designs already, sitting above the swap decision. `LandingTerminal` has its own duplicate of the body-class and lang/dir effect (pre-existing, harmless — both set the same values, and this component's cleanup runs after its child's on unmount, so the final restored state is correct either way). Not de-duplicated here, out of scope for this collapse.

## This was the last page still working under `?design=current`

Confirmed, not assumed, before touching anything (per explicit direction to check this file specifically rather than pattern-match from the other 7): `useDesignMode()` genuinely still resolved to `'current'` here pre-collapse — this file's branch was live, the only one of the 8 that was. Checked `LandingTerminal.tsx`'s own CSS (`.lt-root`, `.lt-nav`, etc.) and confirmed it's scoped under `[data-design="terminal"]`, same as every other Terminal component — so collapsing this file applies the same known breakage (full finding in #1139's PR body) here too, completing the set across all 8 Pattern A pages.

**A stranger cannot reach the broken state**: the only ways to `?design=current` are typing the param deliberately or a stale stored preference, and `/` has defaulted to terminal for real traffic since #748, same as every other route. Production has none of this queued regardless — the owner is holding the whole migration off `main` until it's complete.

## Verified

`tsc --noEmit`: clean. `eslint`: 0 errors, 0 warnings. `npm test`: 794/794.

**Full Landing gate, both themes** (not just one, given this page's stakes): stash-compare screenshots via the Chrome extension against the running dev server (kill by PID, clear `.next`, clean restart each side), terminal mode, dark and light — pixel-identical to before in both. Real users have only ever seen `LandingTerminal` here since #748.

## How to test (QA)

Visit `/` normally — unchanged from what's currently live, in both light and dark theme. `?design=current` is now known-broken here too, same as the other 7 Pattern A pages — not a regression to chase, already accepted (#1139).

## Risk level

**Low for real users** (provable no-op, same reasoning as every prior Pattern A page). This completes Pattern A entirely — all 8 pages done.

Related: #1111, #1128, #1130, #1131, #1133, #1134, #1135, #1137, #1139.


JohnDominicJasmin:
Reviewed against `origin/chore/remove-current-design-landing`, full-tree checkout.

**Thin-wrapper shape** — `LandingContent.tsx` down to 84 lines, `LandingTerminal` mounts unconditionally after the loading-state early return. Only remaining `useDesignMode` mention is a comment, no live branching left.

**Type/lint/tests** — `tsc --noEmit` clean. `eslint components/LandingContent.tsx`: 0 errors, 0 warnings. `npm test`: 794/794.

This completes all 8 Pattern A pages. Same `?design=current` breakage as the other 7 — already independently verified against source on #1139 (every `.tnav`/`.lt-*` rule gated `[data-design="terminal"]`, zero fallback), not re-litigating here.

One open item, not a blocker for this PR specifically: I pinged PM directly to confirm "PM's ruling on #1111: accepted as lost" is a real ruling and not an inference — still waiting on that reply. Doesn't change this PR's own correctness (the regression it inherits is identical across all 8 pages, already shipped in 7 of them), but it does bear on whether QA's `?design=current` test cases stay meaningful project-wide — will follow up separately once PM answers.

PR #1142, opened 2026-09-10
fix(dashboard): plan badge option A - star on Pro, plain text on Free

JohnDominicJasmin:
## Summary

Plan badge option A, the owner's final pick after seeing three built treatments side by side in the real nav. Closes the badge work started in #1089/#1090/#1094/#1120/#1126.

## What changed

- **FREE**: no longer a chip. `.plan-badge-free` is gone; `PlanBadge.tsx` renders it as plain text (`.plan-badge-free-text` — no padding, border, or background, just muted `var(--txt2)` text).
- **PRO**: unchanged filled amber chip, plus a small inline SVG star emblem before the word. Own 5-point star geometry (not from an icon library), `fill="currentColor"` so it recolours with the badge text automatically.
- **TRIAL**: unchanged.

## Why

The owner's own reasoning: the distance between Free and Pro should come from **category, not colour or weight**. Free is a word; Pro is a marked object. Nothing about Free reads as "lesser" — it simply doesn't have the thing. This is why #1126 (filled Free, option C) is **superseded, not merged** — closed with a comment explaining that, not a rejection on quality.

## Verified

All three states mocked side by side in the real desktop nav (injected via disposable JS against the running dev server, not committed), both themes:
- **Dark**: FREE plain grey text / TRIAL amber outline / ★ PRO filled amber — instantly distinguishable, star legible at 9px, not a smudge.
- **Light**: same distinction holds.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794.

**Mobile header fit — re-measured, not assumed**, per the explicit request given the star adds width:
- The star does **not** threaten the fit. Pro's badge (star included) measures 57px vs Trial's unchanged 102.9px — Trial was always the widest badge state and still is by a wide margin.
- Re-measuring surfaced a separate, real finding worth naming: the genuinely worst realistic combination — a route-fallback screen name at its longest (`"econ calendar"`/`"live tracking"`, 13 chars) + the longest real session status (`"Asia Session"`) + `TRIAL · 14D` **together** — was never actually tested as one combination in #1094's original work (that measured `PLAYBOOK` + `82% ADHERENCE` + Trial for the fits-check, and `econ-calendar` + PRO separately for the ellipsis-check, never this exact three-way combination). Measured now: `scrollWidth === clientWidth === 390`, hamburger button's bounding rect fully inside the header, confirmed visually via screenshot — genuinely fits, nothing clips. But the spacer margin at that exact combination is **0.0px**, not the 49.8px #1094 measured for its own worst case. Zero headroom, not broken. Not caused by this PR (Pro's badge got narrower here, if anything) and nothing in this PR touches the mobile header layout — flagging because a more thorough check found it, not because it needs fixing now.

## How to test (QA)

Sign in as Free, Trial, and Pro accounts in turn. Desktop nav and mobile header: Free should read as plain muted text with no chip; Pro should show a small star before "PRO"; Trial unchanged. Confirm all three are visually distinct at a glance, in both themes.

## Risk level

**Low.** CSS + one component's render logic, no state/data changes. The zero-margin mobile-header finding above is informational, not a regression from this PR.

Related: #1089, #1090, #1094, #1120, #1126 (superseded, closed).


JohnDominicJasmin:
**Dev Team.** Checked the translated-strings request before running it — the premise doesn't hold, and here's why with sources.

**Locale list was wrong.** German/Spanish/Filipino aren't the app's locale set. `lib/locales.ts`:
- `AVAILABLE_LOCALES` (what a user can actually pick): `en, ko, zh, ru`.
- `SUPPORTED_LOCALES` (validation superset, not all offered yet): adds `ar, vi, pt-BR, tr, es, id` — Spanish is in there, German and Filipino are not, ever.

**More to the point: none of the three strings in the 0.0px worst case go through translation at all.**

- `session.name` (`"Asia Session"` etc.) — hardcoded English literal in `lib/session.ts` (e.g. line 185), and `TerminalNav.tsx:193` assigns `win.name` straight into state: `setSession(win ? { name: win.name, ... })`. No `t()` call anywhere in that path.
- The screen-name **fallback** that produces `"econ calendar"`/`"live tracking"` — `screenNameFor()` in `TerminalNav.tsx:124-128`. It only reaches this branch for routes NOT in the 5-item `ITEMS` list (Desk/Arena/Scan/Flow/Book, which DO translate via `t(item.tabLabelKey)`). The fallback is the raw URL path segment with hyphens swapped for spaces — a slug, not a label lookup. Slugs don't localize.
- `PlanBadge.tsx`'s `TRIAL`/`FREE`/`PRO` text — plain JS string literals (`` `TRIAL${...}D` ``, `PRO`, `FREE`). Grepped the whole file: zero `t()` calls.

So the 0.0px combination is byte-identical in every locale a user could select — not because the translated strings happen to come out the same length, but because these three pieces never enter the translation system regardless of locale. Re-measuring per-locale would just re-measure the same English string four times.

**That's arguably its own finding, separate from #1142's fit question**: session status and the plan badge silently stay English-only for ko/zh/ru users even though the rest of the UI translates. Not fixing or filing it myself — flagging it as a scope call. Want it as a new issue, or is "these three pieces are intentionally untranslated" already the design?

#1142 itself: unchanged from before — 0.0px margin, fits, nothing clips, not caused by this PR.

JohnDominicJasmin:
Reviewed against `origin/fix/plan-badge-option-a`, full-tree checkout.

**Markup/CSS** — `.plan-badge-free` fully removed (grep confirms zero refs left anywhere). Pro's star is `aria-hidden`, decorative only - "PRO" text still does the accessible work, good call. `fill="currentColor"` correctly inherits the chip's own text color rather than hardcoding it.

**Found a real break, not from this PR's own code but in QA's own e2e suite**: `.plan-badge:visible` in `plan-badge-trial-flip.spec.ts` and `plan-badge-no-flash.spec.ts` would go blind to FREE once it stops carrying `.plan-badge`. Fixed separately - #1143, open now, widening both locators to also match `.plan-badge-free-text:visible`. Not this PR's fault or this PR's job to fix (QA owns `__tests__`/`qa/e2e`) - flagging the sequencing: land #1143 before or alongside this, not after, so the gap window stays zero.

**The zero-margin mobile-header finding in this PR's own body** - good catch, correctly not blocking on it (pre-existing, not caused here). Noting it in `qa/TEST_GAPS.md` myself as a known-tight-but-not-broken state, for whoever eventually adds a longer real-world string to either the screen-name fallback or the session-status label.

`tsc --noEmit`: clean. `eslint`: 0 errors (confirmed, not just trusted the PR body). `npm test`: 794/794.

Approved. @JohnDominicJasmin merge when ready - ideally after #1143 lands (sequencing note above), though #1143 doesn't block correctness of this PR's own code.

Issue #1144, opened 2026-09-10
PRO badge is dark-on-dark in light theme — near-black text and emblem on terminal-light's brown --amber


JohnDominicJasmin:
**Owner-reported from the deployed `qa` build, 2026-09-10, with a screenshot.** Their words: *"on light mode i think its best to set the color text to white."*

## The defect

**In LIGHT theme, the PRO badge renders near-black text and a near-black star emblem on a dark brownish fill.** Very low contrast — the word and the mark are both hard to read.

**In DARK theme it is correct**, because there the fill is bright amber and near-black text sits on it at high contrast.

## Cause

`.plan-badge-pro` paints its text and emblem with a fixed near-black value (`#06070a`) in both themes, while the fill comes from `--amber`.

**`--amber` is not the same colour in the two themes. Terminal-light resolves it to a dark brownish tone.** So the same pair — near-black on `--amber` — is high contrast in one theme and dark-on-dark in the other.

**This is the second light-theme surprise today.** The first: `--amber-2` falls through to the *current* design's light value because terminal never redeclares it, found while mocking the rim variant. **Same root shape — a token assumed to behave the same in both themes.**

## What the fix needs

1. **Light text and emblem in light theme.** Near-white, not near-black.
2. **Not a second hardcoded literal.** Use a token that flips correctly across themes, or add an explicit terminal-light override alongside the other light values. **Check the chosen token is actually redeclared for terminal-light** — the `--amber-2` case is the precedent for one that isn't.
3. **Measured contrast, not eyeballed.** Give the ratio for text-on-fill in **both** themes, and clear WCAG AA in both.

**This codebase has form here and the precedent is in its own comments.** `globals.css`'s note on `--txt3` records a value that failed AA at 1.98:1, was raised, and **still failed on lighter card surfaces nobody had checked** — *"measured, not assumed; the earlier value was picked against --bg1/--bg2 only, which is how it shipped under AA on two surfaces nobody checked."* Same trap, same file.

## Also check — nobody has looked

**TRIAL and FREE in light theme.**

- **TRIAL** is an amber outline with amber text. **If it carries dark-theme amber assumptions, it fails the same way** — and an outline failing is quieter than a filled badge failing.
- **FREE** is plain muted text and must stay legible against the light nav.

**The owner spotted PRO because it is filled and therefore obvious.** The other two are less obvious and equally unchecked.

## How to verify

[...]

PR #1146, opened 2026-09-10
fix(dashboard): plan badge light-theme contrast + Free chip revival

JohnDominicJasmin:
## Summary

Two owner findings from the deployed `qa` badge, folded into one PR since both are appearance changes to the same three CSS rules and the owner will judge the result as one thing:

1. **PRO unreadable in light theme** (dark-on-dark, #1144). Caught from a deployed screenshot.
2. **FREE refined back to a bordered chip** — after seeing option A (#1142, plain text) live, the owner asked for "some border to it, make it badge style."

## What changed

- **`.plan-badge-pro`**: `color: #0a0a0a` (hardcoded) → `color: var(--bg0)`. `--bg0` already flips opposite to `--amber` in every theme this badge renders in (near-black canvas in dark, near-white canvas in light), so one declaration clears contrast in all four theme/design combinations — same paired-foreground shape already used for `--on-accent: var(--bg0)` in terminal dark. The star emblem uses `currentColor`, so it flips with the same value, no second rule.
- **`.plan-badge-free`**: revived from #1126 (closed as superseded when option A shipped plain text, not because it was wrong) — `color: var(--txt); background: var(--bg4); border: 1px solid var(--bdr2)`. `PlanBadge.tsx`'s FREE branch goes back to `<span className="plan-badge plan-badge-free">`, the same base class Trial and Pro use.
- `PlanBadge.tsx`'s doc comment rewritten: the separation between Free and the paid states now rests on **amber+emblem vs. neutral+nothing**, not chip-vs-no-chip, since all three are chips again. Free still carries zero amber — that line is what every round of this ruling has protected.

## Why

**Root cause of #1144** (recorded by QA in #1145, `qa/TEST_GAPS.md` §9): `contrast.spec.ts` sweeps both themes but entirely signed out — `PlanBadge` only renders for an authenticated user, so it structurally could never have caught this. Not a suite regression, a scope gap now on record.

**FREE**: the owner's original complaint (states didn't feel distant enough) is the thing every round of this has protected. Chip-vs-no-chip was one way to hold that line; now that the owner wants FREE as a chip too, the line moves to amber-vs-neutral instead. Checked, not assumed, that the new arrangement still holds it — see Verified below.

## Verified

**Contrast, measured, not assumed** (WCAG AA 4.5:1 floor — badge text is 11px, doesn't qualify as "large text" under 1.4.3):

| Surface | Dark / root | Dark / terminal | Light / current | Light / terminal |
|---|---|---|---|---|
| Pro: ink-on-amber-fill (`--bg0` on `--amber`) | 12.29:1 | 11.94:1 | 5.76:1 | 6.62:1 |
| Free: `--txt` on `--bg4` | 15.06:1 | 15.03:1 | 13.27:1 | 14.71:1 |
| Trial: `--amber` text on its own 8% tint (unchanged, re-checked) | — | — | 5.14:1 | 5.89:1 |

All four clear AA in both themes. Trial and Free were already token-driven before this PR (no hardcoded literal in either), so the only real bug was Pro's hardcoded ink.

**Visual, live** — three states rendered together via disposable injected DOM against the running dev server (exact `PlanBadge.tsx` markup, not committed; auth state not touched, this is a pure CSS/markup check), both themes:
- **Dark**: unchanged from before — near-black star+text on bright amber, neutral-bordered Free, amber-outlined Trial. All three distinct.
- **Light**: Pro's star+text is now clearly legible (white on dark-brown fill, was dark-on-dark). Free reads as a plain neutral chip. All three still separate at a glance — amber (Trial outline, Pro fill+star) vs. no amber (Free) is the line, and it holds.

`tsc --noEmit`: clean (both commits).

## How to test (QA)

Sign in as Free, Trial, and Pro accounts in turn, both themes, desktop nav and mobile header:
- **Free**: neutral bordered chip, no amber.
- **Trial**: amber outline, day count — unchanged.
- **Pro**: filled amber chip, star + "PRO" — confirm the star and text are clearly legible in **light theme** specifically (this is the bug being fixed).
- Confirm all three read as visually distinct categories in both themes.

**Note for review**: FREE carrying `.plan-badge` again means `qa/e2e`'s `.plan-badge:visible, .plan-badge-free-text:visible` locator (#1143) still catches it via the first clause — the second clause now matches nothing, harmlessly. Not touched here; QA's call whether to clean up.

## Risk level

**Low.** Three CSS declarations + one component's className/JSX for FREE, no logic/state changes. Per QA's new standing rule (#1145) this PR touches color values and gets a light-theme pass in review.

Related: #1089, #1090, #1094, #1120, #1126 (revived here), #1142, #1143, #1144, #1145.


JohnDominicJasmin:
Reviewed against `origin/fix/plan-badge-light-contrast`, full-tree checkout. Light-theme pass applied per my own standing rule (#1145) - this PR touches color values.

**Independently recomputed every contrast number in the PR's table** from the actual token values, not trusted from the PR body: PRO 12.29/5.76/11.94/6.62 (root-dark/current-light/terminal-dark/terminal-light), FREE 15.06/13.27/15.03/14.71. All match exactly. All four clear AA with wide margin in both themes.

**`--bg0`-as-ink is the right fix, not a second literal** - confirmed it's genuinely opposite `--amber` in every block that matters (near-black canvas in both dark blocks, near-white canvas in both light blocks), so one declaration covers all four theme/design combinations rather than needing a light-theme override that could itself go stale.

**FREE reverting to a chip**: confirmed `PlanBadge.tsx`'s FREE branch is back to `<span className="plan-badge plan-badge-free">` - carries `.plan-badge` again, so #1143's locator (`.plan-badge:visible, .plan-badge-free-text:visible`) still resolves correctly via its first clause. `.plan-badge-free-text` (the class, not the comment references) is fully gone from the codebase - confirmed via grep, zero live references. The second clause of my own locator union is now a harmless no-op, not broken - I'll clean that up separately, not blocking here.

One minor doc note, not blocking: `globals.css`'s star-emblem comment still says "currentColor... on .plan-badge-free-text above" - that class doesn't exist anymore in this branch. Cosmetic, dev's call whether worth a follow-up.

`tsc --noEmit`: clean. `eslint components/PlanBadge.tsx`: 0 errors. `npm test`: 794/794.

PR #1148, opened 2026-09-10
chore(dashboard): #1111 Pattern C - collapse UpgradeGateModal's design-mode ternaries

JohnDominicJasmin:
## Summary

#1111 Pattern C, first of four - `UpgradeGateModal.tsx`. Collapses its three exports (`LockedFeatureCard`, `FullPageUpgradeGate`, `UpgradeGateModal`) off `useDesignMode()`'s many small inline-style ternaries, same mechanical pattern as the earlier route collapses, applied to inline styles instead of a whole-branch swap.

## What changed

- All three exports carried their own `const mode = useDesignMode()` plus several `mode === 'terminal' ? a : b` inline-style ternaries (background, border, backdrop-filter, box-shadow, className). Collapsed every one to its terminal branch.
- `surface()` helper (took `terminal: boolean`, branched background/border/boxShadow) is now a plain `SURFACE` constant - it only ever returned one shape once `terminal` stopped varying.
- Removed the `useDesignMode` import and the three `mode` locals.
- The three `*-term-wrap` classes (`locked-card-term-wrap`, `upgrade-gate-term-wrap`, `upgrade-modal-term-wrap`) are **kept**, now applied unconditionally rather than deleted - unlike Pattern B's inert wrapper classes, these carry a real effect: `border-radius: 0 !important`, `[data-design="terminal"]`-scoped (confirmed via grep, `app/globals.css:8158-8163`).

## Why

Mechanical, not a redesign. `useDesignMode()` has resolved to `'terminal'` unconditionally since Pattern A/B collapsed every route to mount its Terminal component regardless of the `data-design` attribute - the current-design branch of every ternary in this file has been dead code since then. `?design=current`'s ability to render the old branch was already accepted as lost (#1111 thread) for the same structural reason.

## Verified

- `tsc --noEmit`: clean.
- `eslint`: clean (targeted + full pass via pre-push).
- **Stash-compare on the one live-reachable case**: `LockedFeatureCard`, via the "Global Macro Context" locked card on `/dashboard` (free/signed-out - entitlement doesn't depend on session state here, so signed-out is conclusive same as prior Pattern A pages). Killed the running dev server by PID, `rm -rf .next`, clean restart for each side. Before/after screenshots pixel-identical apart from live market numbers.
- **`FullPageUpgradeGate`'s only call site, `/backtest`, no longer routes there at all** - it client-redirects to `/dashboard` regardless of auth/entitlement state. Pre-existing, unrelated to this change (the file's own comment already noted "/backtest is hidden (#273)"). So that export's collapse rests on the same mechanical proof as the file's other two exports (no reachable branch to diff against), not a live render - naming this rather than silently skipping it.
- `UpgradeGateModal` (the popup itself) wasn't independently screenshotted open - same mechanical proof, plus it shares the identical `SURFACE`/ternary shape already confirmed live via `LockedFeatureCard`.
- Grep confirms no `useDesignMode`, `mode ===`, or `surface(` reference survives in the file.

No color values changed (every literal is the same value the terminal branch already used) - doesn't trigger QA's light-theme review rule, but flagging that explicitly since the diff touches a lot of color-looking lines.

## How to test (QA)

Sign in as Free (or stay signed out): `/dashboard`'s "Global Macro Context" card should render exactly as it does on `qa` today - filled `--bg1` panel, 1px `--bdr` border, square corners, "Unlock with Pro" button. Click it to open the upgrade modal - same panel treatment, no visual change. Any other Pro-gated card/route using this component (e.g. via `DryPowder`, `TradeJournal`, `HypothesisTracker`, `OnChainScore`, `GlobalMacroContext`, `app/alerts`, `app/arena`) should be unchanged too.

## Risk level

**Low.** Pure ternary collapse to the branch already rendering everywhere reachable; no new colors, no logic change. Named risk: `FullPageUpgradeGate` (`/backtest`) couldn't be live-verified since the route itself is unreachable right now - rests on the mechanical proof, not a screenshot.

Related: #1111 (Pattern C, 1 of 4 - `Tip`, `PWAInstallPrompt`, `MarketStructure` next).


JohnDominicJasmin:
Reviewed against `origin/chore/design-removal-pattern-c-upgrade-gate`, full-tree checkout.

**No new color values** - confirmed by diff, not just trusted the PR's own claim: every literal kept (`var(--bg1)`, `var(--bdr)`, `color-mix(in srgb, var(--bg0) 82%, transparent)`) is byte-identical to what the terminal branch already used. My light-theme review rule correctly doesn't apply here - appreciate the explicit flag in the PR body anyway, that's the right instinct even when it turns out not to trigger.

**Zero live references left**: grepped `useDesignMode`, `mode ===`, `surface(` on the branch - none. The three `*-term-wrap` classes confirmed still real and `[data-design="terminal"]`-scoped in `globals.css:8145-8150` (`border-radius: 0 !important`), correctly kept unconditional rather than deleted.

**`/backtest`'s dead-route claim, independently traced**: confirmed via `proxy.ts:37` (`BLOCKED = new Set(['/backtest', '/live-tracking'])`, `matcher: ['/backtest', '/live-tracking']`) - edge-level 307 redirect to `/dashboard`, per #264/#265. `FullPageUpgradeGate` genuinely has no reachable call site right now, unrelated to and unchanged by this PR. The PR's "mechanical proof, not a screenshot" framing for that export is the right call given the constraint, not a shortcut.

`tsc --noEmit`: clean. `eslint`: 0 errors. `npm test`: 794/794.

JohnDominicJasmin:
**Dev Team.** One line for the record, per PM: `FullPageUpgradeGate`'s collapse was not live-verified because its only call site (`/backtest`) is unreachable right now - the route client-redirects to `/dashboard` regardless of auth state (`/backtest` hidden, #273, deliberate). That's why that export rests on the mechanical proof alone rather than a screenshot, not laziness.

PR #1153, opened 2026-09-10
chore(dashboard): #1111 Pattern C - collapse MarketStructure's design-mode ternary (4/4)

JohnDominicJasmin:
## Summary

#1111 Pattern C, last of 4 - `MarketStructure.tsx`. Closes out the four inline-style components; `KLineProChart.tsx` is next in the design removal, then `app/globals.css` deletion, then provider/param/script removal last.

## What changed

- The event badges' background (flat `--bg1` in terminal vs. a tint of the event's own colour in current design - #644's owner ruling) was gated behind a `badgeBg()` helper reading `useDesignMode()`. Collapsed to the terminal branch; both call sites now read `background: 'var(--bg1)'` directly rather than through a now-pointless zero-argument helper.
- `useDesignMode` import and the `terminal` local removed.
- Moved the #644 comment down to sit above the JSX that actually uses the value (it previously sat above the now-deleted helper), and trimmed the "current design ... not approved for work" caveat since that branch no longer exists.

## Why

Mechanical, same as #1148/#1150/#1151.

## Verified

- `tsc --noEmit`: clean. `eslint`: clean (one pre-existing warning at line 178 - synchronous `setState` in a `useEffect` - unrelated to this change, not introduced by it).
- **Live verification note, same shape as #1148's `FullPageUpgradeGate`**: `MarketStructure`'s only real mount (`app/arena/page.tsx:2319`) is deliberately headless - `<div style={{display:'none'}}>`, a data collector for the AI/Grok context, same pattern as the neighbouring `AbsorptionDetector`. Its visible markup (loading skeleton, error state, the data card with these badges) is not reachable through normal navigation right now. Flagging per the #1111 thread rather than treating it as a one-off.
- **Verified anyway**: forcibly un-hid the already-mounted DOM node via injected JS (no code change, no React state touched - just toggled the wrapper's `display`) and stash-compared the revealed card against the running dev server. Before/after screenshots pixel-identical - same badge colours, same flat `--bg1` fill, across the six BOS/CHoCH × bullish/bearish combinations visible in the sample data.

## How to test (QA)

No visible surface currently reaches this component, so there's nothing to click through on `qa`. If/when it's wired to a visible surface, the event badges should render flat `--bg1` (not a colour-tinted background) - same as today.

## Risk level

**Low.** One-line style collapse, dead helper removed, no logic change. Named risk: same "unreachable in the running product" shape as #1148 - rests on the mechanical proof plus a forced-visibility spot-check, not a normal click-through.

Related: #1111 (Pattern C complete, 4/4 - `KLineProChart.tsx` next).


JohnDominicJasmin:
Reviewed against `origin/chore/design-removal-pattern-c-market-structure`, full-tree checkout.

**Eslint warning claim verified, not trusted**: ran eslint against `origin/dev`'s current `MarketStructure.tsx` directly - identical warning present (`react-hooks/set-state-in-effect`, `load()` called synchronously in a `useEffect`, dev's line 192 vs this branch's 178 - the shift matches the ~14 lines removed from the deleted ternary/comment block). Genuinely pre-existing, not introduced here.

**Headless-mount claim verified**: `app/arena/page.tsx` mounts `MarketStructure` inside `<div style={{ display: 'none' }}>` alongside `AbsorptionDetector`, exactly as described - a data collector for Grok context, no visible surface right now. Same "mechanical proof, not a click-through" shape as #1148/#1153's own prior instance, correctly named rather than skipped.

No new colors - `var(--bg1)` is the same literal both call sites already read through `badgeBg()`. Zero `useDesignMode`/`badgeBg`/`terminal` references left (grep, only an unrelated comment mentioning "terminal" survives).

`tsc --noEmit`: clean. `npm test`: 794/794. This closes out Pattern C (4/4).

Issue #1149, opened 2026-09-10
Sign out does nothing — the await on auth.signOut() has no timeout, and the auth path is currently hanging


JohnDominicJasmin:
**Reported by Danivy on `qa`, 2026-09-10, and independently confirmed by the owner the same minute.** Two findings in one screenshot; **the first is functional and the second is cosmetic.** Filed together because they were reported together, but they are separate defects.

## 1. Sign out does nothing

Clicking **Sign Out** in the account menu leaves the user signed in.

**This is the most serious defect reported today.** A user who cannot sign out cannot leave their account on a shared, borrowed or public machine. It is a security-shaped problem, not an inconvenience.

## Probable cause — and it connects to a live incident found an hour earlier

The handler is `components/NavDrawer.tsx:437`:

```
onClick={async () => {
  setAuthOpen(false);
  track.signOut();
  await signOut();
  window.location.assign('/login');   // hard navigation
}}
```

`signOut` calls `forceSignOut` (`lib/authSession.ts:65`), which **awaits `sb.auth.signOut()`**.

**If that call hangs rather than failing, the `await` never resolves, `window.location.assign('/login')` is never reached, and nothing visible happens.** Which is exactly what was reported: the button does nothing at all.

**And the auth path on the dev project is currently hanging.** From #1025, established hours ago: **34 `504` timeouts on `/auth/v1/*` in the last 20 hours**, the most recent minutes before this report, plus direct `postgres` queries measured at **10,012ms and 11,679ms** and one cancelled mid-flight. Three independent observers confirmed it live.

**So the leading hypothesis is that sign-out is not broken code — it is the same backend starvation, surfacing as a dead button.**

**The error path is already handled correctly.** `forceSignOut` catches a rejection and calls `clearStoredSession()`, so a *fast failure* still signs the user out locally and navigates. **A hang is the one case with no handling** — there is no timeout on that await.

## What to establish

1. **Does it reproduce on production?** Production's auth path is healthy. **If sign-out works there and fails on `qa`, the hypothesis is confirmed** and this is an environment symptom of #1025 rather than an independent bug.
2. **Does it eventually complete, or never?** Danivy has been asked. *Never* means a hang; *eventually* means slow, which is a different fix.
[...]

PR #1154, opened 2026-09-10
fix(auth): sign-out survives a hung logout request, and shows it worked

JohnDominicJasmin:
## Summary

#1149. Sign Out appeared to do nothing. The owner's report and Danivy's follow-up narrowed the cause: not dead, slow - the request eventually resolved, but the button gave no indication anything was happening while it did. Fixes both halves: the request is now bounded, and the button now visibly begins the action immediately.

## What changed

1. **`lib/authSession.ts`**: `forceSignOut`'s `sb.auth.signOut()` call is now raced against an 8s timeout (`SIGN_OUT_TIMEOUT_MS` - same value and reasoning as `AuthProvider`'s `SESSION_RESOLVE_MS`). On timeout it's treated exactly like a real error: `clearStoredSession()` runs, a truthy value returns. Every caller of `forceSignOut` now settles within a bound regardless of whether the network ever answers - the three UI sign-out buttons that go through `AuthProvider.signOut()`, and `app/ops/layout.tsx`'s `signOutAndSwitch()`, which called `forceSignOut` directly and had the identical unbounded await (a second real instance of the bug, not just NavDrawer's).
2. **Four buttons** (NavDrawer's drawer item, TerminalNav's dropdown item, the settings page button, both ops console buttons) now set a local `signingOut` state on click, disable, and swap their label for the existing `.login-spinner` (same pattern already used for the password-save button on `/settings`). NavDrawer and TerminalNav previously closed their drawer/dropdown the instant Sign Out was clicked - which hid the only place feedback could show even after (1) - so both now keep it open until the hard navigation actually fires.

## Why

Both halves are required. (2) alone reproduces the exact bug on any slow connection even with a fully healthy backend - a spinner-less click still "does nothing" for however long the request takes. (1) alone still leaves a silent button, just for a shorter window - and without it, a genuinely dead connection leaves the user unable to sign out at all, which is a security-shaped problem on a shared machine, not an inconvenience.

## Investigated per PM's ask

- **`forceSignOut`'s other two direct callers** (`AuthProvider`'s >7-day inactivity expiry, BAN ENFORCEMENT): the inactivity path already double-wraps with its own local `withTimeout` - now redundant but harmless, comment updated to say so rather than removed (two independent bounds on `loading` specifically isn't wrong, just no longer load-bearing). BAN ENFORCEMENT is `void forceSignOut(sb)`, fire-and-forget from a realtime callback, never awaited by any UI - not the blocking half of this bug, but benefits from the same fix regardless.
- **`qa/e2e/signout-resilience.spec.ts`**: read, not touched. Its "logout never answers" test uses `route.abort('connectionfailed')`, which fails **fast** - GoTrue's fetch rejects near-instantly, so `forceSignOut`'s existing try/catch already caught it before this PR, and the test would pass against the old, unbounded code too. It covers a fast rejection, not a request that never resolves or rejects at all, which is the actual shape of this bug - and Playwright's route interception has no built-in way to express that short of a handler that never calls `fulfill`/`abort`/`continue`. Naming the gap for QA rather than closing it myself.

## Verified

- `tsc --noEmit`: clean. `eslint`: clean (0 errors app-wide - the six pre-existing `setState`-in-effect warnings elsewhere are unrelated to this diff, not introduced by it).
- Full unit suite: 794/794.
- **Live, end-to-end**, real E2E fixture account (`qa.user.a@liquidityhq.test`, dev Supabase - same fixture QA's own specs use) against the running dev server: signed in, clicked Sign Out on `/settings`, landed on `/login` with the sign-in form. Healthy path confirmed unaffected.
- **Attempted, not achieved**: a genuine network hang in the browser (not a fast rejection) to watch the timeout fallback fire live. Overriding `window.fetch` via injected JS didn't intercept it - `supabase-js` captures its own `fetch` reference at client construction, which happens before any post-navigation script injection can run without Playwright/CDP-level page-init hooks this session's tools don't expose. The fix rests on `Promise.race`'s own well-defined semantics (a standard primitive, not something specific to this codebase) rather than a live repro of the hang itself - saying so rather than claiming a verification that didn't happen.

## How to test (QA)

Healthy path: sign in, click Sign Out from any of the four surfaces (mobile drawer, desktop nav dropdown, `/settings`, `/ops`) - button should show a spinner immediately and land on the sign-in form shortly after, same as today.

The bug itself needs a genuinely hung `/auth/v1/logout` (not a 500, not an aborted connection - both already worked and are already covered by the existing spec) - if you have a way to hold that connection open past 8s, the button should still land on `/login` at roughly the 8s mark rather than staying stuck.

## Risk level

**Medium.** Touches the sign-out path on every authenticated surface. Low logic risk (the timeout path is a standard `Promise.race`, the visible-feedback path mirrors an existing working pattern) but the actual hang scenario itself is unverified live, per the note above - naming that rather than calling this fully proven.

Related: #1149, #304, #317 (prior sign-out resilience work).


JohnDominicJasmin:
Reviewed against `origin/fix/sign-out-hang`, full-tree checkout.

**`forceSignOut` fix traced end-to-end**: `Promise.race([sb.auth.signOut(), timeout])`, `clearStoredSession()` runs unconditionally once the race resolves via either branch (`if (error) clearStoredSession()`), returns truthy either way. `AuthProvider.signOut()` confirmed to actually call `forceSignOut(sb)` (line 374), so all three main UI paths get the protection, not just the ops console.

**All four buttons verified identical shape**: guard against re-entry (`if (signingOut) return`), disable + spinner swap (`.login-spinner`, confirmed real/existing class), hard navigation after. NavDrawer/TerminalNav's "keep the drawer/dropdown open" change is real and correctly reasoned - closing it would hide the only place the spinner could render.

**`AuthProvider`'s inactivity-expiry note**: confirmed comment-only, no logic change - the outer `withTimeout` genuinely is now redundant-but-harmless as described.

**The named eslint warning re-checked**: 6 warnings total (0 errors), all `react-hooks/set-state-in-effect`, including `NavDrawer.tsx:98` - same pre-existing warning confirmed on #1139 (line shifted from 93 to 98 by unrelated edits since). Nothing new.

**The named test gap is real, and I've closed it separately**: #1156, a genuinely-hung route handler (never calls `fulfill`/`abort`/`continue`) + `page.clock.fastForward` past `SIGN_OUT_TIMEOUT_MS`, traced against this PR's actual `Promise.race` shape. Not run live - memory headroom today, and this suite's `webServer` is a full production build, not a dev server - flagged honestly rather than silently skipped. Land #1156 alongside this, not required before it.

**Also landed**: #1155, the stale `WIDE_BASELINE` entry from #1148 that pre-push flagged - verified live before removing, same pattern as the day's other ratchet cleanups.

`tsc --noEmit`: clean. `eslint`: 0 errors app-wide. `npm test`: 794/794.

Issue #1157, opened 2026-09-10
Database: security audit, indexes, dev project capacity (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: the owner, for the dev-database write.**

- [x] FK indexes on production, 2026-09-12.
- [x] Production functions that were hardened outside migrations are now captured in the repo (#1306, #1310, merged).
- [ ] FK indexes on dev. It's a dev-database write, so it needs the owner's OK, then a window from QA (no DDL during a test pass).
- [ ] #1025 the dev database degrades after any DDL. The cause is still open. Parked, and nobody is on it.
- [ ] #1159 leaked-password protection: deferred by the owner until the Supabase upgrade.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Owner-directed, 2026-09-10.** Full Supabase advisor sweep of **both** projects — `liquidity-hq-prod` (`qdpwhnvmhqgzijuwopso`) and `liquidity-hq-dev` (`wdtjhrilakoitfcezxpx`) — after the dev database was found timing out live (#1025).

**Both projects report identical findings.** Same schema, same omissions, different prefixes.

---

## 1. Leaked password protection is DISABLED — do this first

**Supabase can reject passwords found in known breach corpora (HaveIBeenPwned). It is currently off on both projects.**

**One toggle in the auth settings. No migration, no deploy, no code, no downtime, no rollback risk.** It is the cheapest security improvement available and the only item here with no downside whatsoever.

**Do it on production regardless of what else on this list happens.**

---

## 2. Sixteen RLS policies re-evaluate `auth.<function>()` per row

Flagged as `auth_rls_initplan`, on both projects, including on `lhq_user_subscriptions`'s `sub_select_own` — **the exact policy behind #1119's entitlement read.**

**Every authenticated query pays an identity check on every row rather than once per statement.** On a small instance that is real, repeated backend work — the shape of load that produces the crawling queries measured in #1025 (10,012ms and 11,679ms, one cancelled mid-flight).

**The fix is already written and merged and has never been applied anywhere: #1036.**

**Sequencing, and it matters more than the change:**

- **Dev first, then production.** #1036 has never run on any database. **Applying it to production first would mean its first-ever execution happens on the only database with real user data and no backups.**
[...]

Issue #1119, opened 2026-09-10
A failed subscription read silently downgrades a paying Pro customer to free FEATURES, not just the badge


JohnDominicJasmin:
**Confirmed by direct code read, 2026-09-10, during #1118.** Filed separately from #1089 deliberately: **#1089 is about the badge and will close with it. This is about access, and it must not close with a display fix.**

## The defect

**When the `user_subscriptions` read fails or times out, `entitled` resolves to `false` — so a paying Pro customer loses Pro FEATURES, not just the Pro badge.**

`AuthProvider` derives `isPro` from `role`, which sits at its `'free'` default until the subscription read settles. `entitled` is `isPro || isTrial`. **A failed read therefore fails toward "not entitled"** across every consumer of that context — the fast-timeframe gating, the strategy-panel indicator limit (3 → 1), and every other `entitled` check in the app.

**The badge was the visible symptom. This is the actual behaviour underneath it.**

## Why this is worse than it sounds

**It fails silently and it fails toward taking things away.** A Pro user during a backend hiccup does not see an error — they see a product that has quietly become the free tier. Fast timeframes stop working. Their indicator budget drops from 3 to 1. **Nothing tells them why, and nothing tells them it is temporary.**

**And the backend in question is the one with documented problems.** #949 and #1025 record sign-in timeouts and a dev Supabase project at 91% of its IO budget. This is not a hypothetical failure path.

## What #1118 did and did not fix

**Fixed: the badge.** It now caches the last successfully-resolved plan per user and renders that on failure, so the *display* no longer claims `FREE`. Owner's decision, recorded on #1089.

**Explicitly NOT fixed: entitlement itself.** That cache is **display-only by design** — nothing but the badge reads it, and that constraint is deliberate. **A `localStorage` value that could satisfy an entitlement check would be a client-side paywall bypass**, which is a worse defect than the one being fixed. So the real fix cannot be "let `entitled` read the cache", and anyone reaching for that should stop.

**So today the badge says `PRO` while the features behave as free.** That is better than the reverse — the customer is not told they are on the free plan — but it is not resolved, and the gap between what the badge claims and what the app does is now a real inconsistency worth naming.

## What a real fix probably looks like

Not scoped, and deliberately not decided here:

[...]

PR #1160, opened 2026-09-10
chore(db): record the dev RLS InitPlan rewrite that was applied by hand

JohnDominicJasmin:
## Summary

Records the 7 RLS policies applied by hand to the dev Supabase project on 2026-09-11, completing what `20260907b_rls_auth_uid_initplan.sql` deliberately left unfinished. **Dev now reports 15 rewritten policies and 0 remaining per-row `auth.uid()` call sites.**

**This is a record of a change already applied, not a change to apply.** The database is ahead of the repo until this merges — which is the drift this file exists to close.

## Why a new file rather than editing 20260907b

That file records what it did **and why it stopped**. It stopped for good reasons. **Rewriting it would have deleted the record of a correct decision.**

## Why these could now be done

20260907b excluded five policies because their text was **unconfirmed** — set through the Supabase dashboard, with no `CREATE POLICY` for them anywhere in `supabase/migrations/`. Its own words: *"writing a rewrite for a policy whose exact current text is unverified risks silently changing its semantics instead of just its performance."*

**It named the fix: pull the live text from `pg_policies` first.** That is what was done. Command, roles and clauses are copied from the live definitions — **only the `auth.uid()` call sites change.** Nothing was inferred from a table's name.

## A security question that file left open, now answered

20260907b excluded `lhq_dev_grok_usage` for a **security** reason, not performance: prod's policy was hardened in 20260807a after a critical quota-reset vulnerability, but whether dev received the same hardening was unconfirmed. **Dev might still have carried `users_own_usage` (FOR ALL), which would let a user write their own usage row and reset their AI quota.**

**Checked before touching it.** `lhq_dev_grok_usage` carries exactly one policy: `grok_usage_select_own`, SELECT, authenticated. **`users_own_usage` does not exist on dev.** The vulnerability is not present, and the rewrite is a pure performance change.

Recorded in writing because the question was raised in writing.

## Duplicate policies — present, deliberately not dropped

`lhq_dev_hypotheses` and `lhq_dev_hypothesis_evidence` each carry **two** permissive ALL policies with identical semantics. 20260907b created a new policy rather than replacing the differently-named existing one, so both were evaluated — and until now the pre-existing one did so per row.

**Both are now InitPlan-folded, so the cost is gone. The redundancy is not resolved here.** Dropping a live RLS policy is a security change, and doing one as tidy-up inside a performance migration is how a boundary gets removed by accident. Flagged on #1157.

## Production is excluded, despite being owner-approved in principle

The owner approved both projects (*"apply it to dev and prod"*) and said *"apply to dev now"*. **Prod is not in this file.**

Its live policy text has not been pulled and verified the way dev's was. **The two projects have already diverged once — on `grok_usage` — so prod's versions cannot be assumed to match dev's.** And production Supabase has **zero backups**: an RLS rewrite written from assumed text, with no recovery path, is the one combination on the board with no undo.

**Prod gets the same treatment dev got** — pull the live text, write from it, apply, verify — not a copy of this file with the prefix changed.

## How to test (QA)

**No application change. Nothing to exercise in the UI for the migration itself.**

1. **Confirm the record matches reality:** `select count(*) from pg_policies where schemaname='public' and qual like '%auth.uid()%' and qual not like '%( SELECT auth.uid()%'` returns **0** on the dev project.
2. **The verification that matters is ISOLATION, not timing.** These policies are the boundary that stops one user reading another's rows. **Confirm an authenticated user still sees exactly their own rows and nothing else**, per affected table. A faster query with a broken boundary is worse than the slowness it fixed.
3. **`lhq_dev_user_subscriptions` deserves particular attention** — it is the entitlement read behind #1119, and it was the most important omission in the original migration.

## Risk level

**Medium.** No data, columns, grants or role assignments are touched — only policy definitions, with semantics preserved from the live text. **But these are security boundaries**, and the change is already live on dev, so a mistake would already be in effect rather than pending.

**The timing question is abandoned, deliberately.** Thirteen foreign tables were dropped from this database at `2026-09-10T16:51:54.804Z` and this rewrite followed before a post-delete baseline existed — **so neither change can be credited.** Recorded on #1025 as a lost measurement rather than resolved with a plausible attribution.


JohnDominicJasmin:
Reviewed against the live dev project directly, not just the migration file - ran the verification myself rather than trusting the PR's claims.

**Policy count claim**: `select count(*) from pg_policies where schemaname='public' and qual like '%auth.uid()%' and qual not like '%( SELECT auth.uid()%'` → **0**, matches exactly.

**Isolation on `lhq_dev_user_subscriptions` (the one that matters most, #1119)** - three-way test, `set local role authenticated` + `request.jwt.claims` to simulate two real users and one that doesn't exist:
- User A (`9489f7cf...`): exactly 1 row back, their own (`role: pro`).
- User B (`40564267...`): exactly 1 row back, their own (`role: free`), not A's.
- Nonexistent UUID: **0** rows - the policy is doing real filtering, not vacuously passing.

**Isolation on `lhq_dev_price_alerts`** (2 rows total across both users): User A's authenticated query returned exactly their own row, not both.

**Duplicate-policy claim, confirmed live**: `lhq_dev_hypotheses`/`lhq_dev_hypothesis_evidence` each genuinely carry two ALL policies with identical `qual` (`owner_all_dev_hypotheses` + `owner_all_hypotheses`, `owner_all_dev_evidence` + `owner_all_evidence`) - both correctly InitPlan-folded now, redundancy real but not a security gap as described. Agree with not touching it here - dropping a live policy inside a performance migration is exactly the kind of scope creep that causes an accidental boundary removal.

**The lost-measurement framing in the risk section** is accurate and matches what I recorded on #1025 independently before seeing this PR - good that neither of us tried to rescue a timing story that isn't there.

This is a record of an already-applied change, so there's no app code to gate on tsc/eslint/tests - the verification that matters is the isolation check above, done against the actual running database, not the file.

Issue #1165, opened 2026-09-10
Unbounded getSession() in getAuthToken — three duplicate copies, hangs every caller when auth is slow


JohnDominicJasmin:
**Filed separately from #1163 on purpose.** #1163 asks whether this release is broken. **This issue is true either way** — it does not close when that one does.

## The defect

**`getAuthToken()` does an unbounded `sb.auth.getSession()`.** If the auth backend is slow or returning `504`s, the call does not fail — **it waits, with no timeout, forever.** Every caller that awaits it hangs with it.

**`AuthProvider.tsx` already documents and bounds this exact hang for its own `getSession()` call (#727). These call sites do not.**

## There are THREE copies, not one

Found by dev while checking callers — **byte-for-byte the same shape in all three:**

- `lib/supabase.ts` — `getAuthToken`
- `app/alerts/page.tsx` — its own local `getAuthToken`
- `components/GrokChat.tsx` — its own local `getAuthToken`

**This is the fourth unbounded await found in this codebase in three days**, after `lib/authSession.ts` (#1149) and `app/ops/layout.tsx` (found during #1154).

## Why bounding it is low-risk — dev's assessment, and it is the load-bearing part

**`getAuthToken`'s contract is already "optional, attribution-only"** — `lib/supabase.ts`'s own comment says so, and **every caller already does `token ? {Authorization...} : undefined`.**

**So a timeout falling back to `undefined`/anon is not a new failure mode.** It is **the same degrade path that already exists for "no token at all"**, reached through a different door. **Nothing downstream treats a missing token as an error.**

That makes this materially lower-risk than the `AuthProvider` bound already shipped — that one had to preserve a session; this one is allowed to give up.

## Fix shape — one helper, not three timeouts

**One bounded helper in `lib/supabase.ts`; `alerts` and `GrokChat` import it instead of keeping their own copies.**

**This is exactly what `forceSignOut` already did for sign-out** — `lib/authSession.ts`'s own comment: *"shared function rather than three copies of a try/catch"*. **Adding three separate timeouts would leave the duplication in place and guarantee the next fix has to find all three again.**

## Why this matters even if #1163 turns out to be environmental

The leading hypothesis on #1163 is that `staging` hangs because it runs against the **dev** Supabase, measured unhealthy in #1025 (34–35 `504`s on `/auth/v1/*`), while production runs against a healthy prod Supabase.

[...]

Issue #1163, opened 2026-09-10
RELEASE BLOCKER: /api/cmc?type=global never resolves in the browser — 5+ converted pages hang on Loading, staging only


JohnDominicJasmin:
**RELEASE BLOCKER. Found by QA on `staging` at `3af8b91c` during the pre-release check, 2026-09-11. Release PR #1161 is a NO-GO and was not merged.**

**Not present on production.** `liquidity-hq.com` is unaffected and was not deployed.

## What happens

**At least 5 of the 8 converted pages hang forever on "Loading…"** — `/liq`, `/markets`, `/funding`, `/briefing`, `/scanners` — with the landing-page footer bleeding through underneath. **It never resolves.**

**Confirmed in a completely fresh tab**, ruling out stale session or tab state.

**And it is not limited to those five.** `/dashboard` and `/correlation` loaded correctly on first visit, but **`/dashboard` also hung on a later reload.** So the failure appears progressive across a session rather than scoped to particular routes — which makes the true blast radius the whole app, not five pages.

## What it was traced to

**`/api/cmc?type=global`. The browser's request to it never resolves.**

- Confirmed in the network panel.
- **Persists across fresh tabs and fresh auth tokens.**
- **A direct `curl` to the identical URL from a shell succeeds in 0.4s.**

**So the endpoint is not dead.** Something about the server's handling of browser-originated requests to it gets stuck.

**QA's hypothesis, explicitly labelled unconfirmed:** a request-dedup or cache layer holding a promise that never settles, so every subsequent request queues behind the first. **That would explain the progressive behaviour** — the first request wedges, and everything afterwards inherits the wait. **Needs the route handler read before it is treated as the cause.**

## Why production is fine

**Production does not call `/api/cmc?type=global` at all on `/liq`.** Its code path differs from what is in this release. **So this is a regression introduced by work in the 98 commits, not a latent condition that happened to surface.**

**That difference is also the strongest clue available:** whatever now calls that endpoint on those pages is new, and is probably shared code reached by `LiqTerminal`, `MarketsTerminal`, `FundingTerminal` and the rest — which is exactly the set of components the Pattern A collapses touched.

## What was verified clean before this was hit

**The plan badge in all three states, and Free/Pro both correct on `staging`.** That part of the release is sound. **The badge is not implicated.**

## Why nothing shipped

[...]

PR #1166, opened 2026-09-11
fix(auth): bound getAuthToken, consolidate three duplicate copies (#1165)

JohnDominicJasmin:
## Summary

#1165. `lib/supabase.ts`'s `getAuthToken()` bounds `sb.auth.getSession()` with an 8s timeout (`AUTH_TOKEN_TIMEOUT_MS`) instead of waiting indefinitely, and consolidates two duplicate unbounded copies of the same function into this one. Also fixes a real silent-empty-state bug found while surveying every call site for what a signed-in user sees when the token times out.

## Framing - this is not the fix for #1163

QA's `qa`-vs-`staging` comparison confirmed the #1163 hang is **environmental**: `qa`/`staging` point at the currently-degraded dev Supabase (#1025), production's Supabase is healthy. `#1025` (the sick database) is the fix for #1163; that's the owner's to resolve, not this PR's.

**What this PR actually is**: the thing standing between a future degraded *production* database and every page hanging app-wide with no recovery. That was true before tonight's incident surfaced it and stays true after #1025 is resolved. `getSession()` itself was never instrumented mid-hang - the mechanism (supabase-js's refresh-on-read of an expired/unrefreshable token never settling, same shape `AuthProvider`'s own `#727` comment documents and bounds its own call for) is the best-supported explanation available, not a demonstrated one. Framing accordingly rather than claiming this "fixes the hang."

## What changed

- **`lib/supabase.ts`**: `getAuthToken()` now races `sb.auth.getSession()` against `AUTH_TOKEN_TIMEOUT_MS` (8000ms - same value and reasoning as `AuthProvider`'s `SESSION_RESOLVE_MS` and `lib/authSession.ts`'s `SIGN_OUT_TIMEOUT_MS`), falling back to "no token" on timeout.
- **Three duplicate copies → one.** `app/alerts/page.tsx` and `components/GrokChat.tsx` each carried their own byte-identical unbounded version of this exact function - both deleted, both now import the shared one (their now-unused `getSupabase` imports removed too). `components/MarketProvider.tsx` and `components/EconCalendarWidget.tsx` already imported the shared one, so they're fixed with zero changes. **Five consumers total, not three** - `EconCalendarWidget` surfaced during review.
- **`app/alerts/page.tsx`'s `loadPriceAlerts()`**: added the `res.ok` check every one of its five sibling functions in the same file already has (see Investigation below).

## Investigation: what does a signed-in user see when the token comes back `undefined`?

Asked before landing this, not assumed - per call site:

| Consumer | What it fetches | Missing-token outcome |
|---|---|---|
| `MarketProvider` (3 sites: dominance, ETF flows, altseason) | Public market data, token is attribution-only | Data still loads for everyone; only attribution is lost. **Low risk.** |
| `EconCalendarWidget` | Public calendar (its own file explicitly names this as an "unauthenticated-allowed route", same as `cmc`) | Same as above. **Low risk.** |
| `GrokChat` | AI chat proxy, server requires auth | Server returns `AUTH_REQUIRED`; client shows the sign-in modal. Fails **visibly**. **Low risk.** |
| `app/alerts` - add/delete/test/link/preview (5 sites) | User's own rows, mutations | All five already check `res.ok`/`d.ok` and surface a visible error. **Already correct.** |
| `app/alerts` - mute-preferences load | User's own notification prefs | Own comment explicitly treats "no token" as "signed out, nothing to load" **by design**. Not touched here - see below. |
| **`app/alerts` - `loadPriceAlerts` (the list GET)** | **User's own price alerts** | **Real bug, found and fixed.** `GET /api/price-alerts` correctly 401s on a missing token (its own comment: fixed for a BOLA finding - *"401, not an empty 200"*), but this call site did `fetch(...).then(r => r.json())` with no `res.ok` check. A 401 body has no `alerts` key, so `d.alerts ?? []` silently supplied one - **a signed-in user whose token timed out would see "you have no alerts" when they have alerts, with no error and no retry.** |

**Not touched**: the mute-preferences load. Its own comment explicitly treats "no token" as "signed out" by design, and the worst case there is a muted-list default, not data loss - a real but much smaller gap than `loadPriceAlerts`'s, and distinguishing "genuinely signed out" from "timed out" is its own decision, not folded into this PR.

## Why bounding is safe here specifically

`getAuthToken`'s whole contract is already optional (its own doc comment) - every caller already handles a missing token by omitting the `Authorization` header. A timeout falling back to `undefined` is the *existing* "no token" path, reached by a different door, not a new failure mode. The one place that treated "no token" incorrectly (`loadPriceAlerts`) is fixed above, not left as a landmine for QA's next pass against the still-degraded dev database.

## Verified

`tsc --noEmit`: clean. `eslint`: 0 errors across all touched files (pre-existing warnings elsewhere untouched). Full unit suite: 794/794. `next build`: clean.

## How to test (QA) - assertions, not tests (per the standing rule, you write these)

- **Slow or failing auth backend** (matches your `qa`-vs-`staging` setup): pages using any of the five consumers render **degraded/anonymous within ~8s** and do not hang - `/liq`, `/markets`, `/funding`, `/briefing`, `/scanner`, `/alerts`, and anywhere `GrokChat` mounts.
- **Healthy auth backend**: token still attaches, behaviour unchanged from today.
- **All three consolidated call sites specifically** - `lib/supabase.ts`, `app/alerts/page.tsx`, `components/GrokChat.tsx` - not just the one that already had a bound in principle.
- **`/alerts` specifically, since this is where you'll be re-verifying against the sick database**: a signed-in user with real price alerts should see either their alerts or a visible error - never a silent "no alerts" when the token timed out.

## Risk level

**Medium.** Touches a function five different features call, on every page load for a signed-in user. Low logic risk (standard `Promise.race`, same shape already used and reviewed in `#1154`), but per the framing above: the actual degraded-backend scenario this defends against has not been reproduced against this specific code path, only against the mechanism it shares with `AuthProvider`'s already-bounded call.

Related: #1163 (the incident that surfaced this), #1025 (the actual fix for tonight's hang), #1154 (`forceSignOut`'s identical precedent).


JohnDominicJasmin:
**Dev Team.** Same defect class as #1107's Binance `r.ok` sweep - a caller skipping the response-status check that its siblings already do, producing a confident-but-wrong success instead of a visible failure. Fifth instance of this shape on this project. Flagging the connection for whoever does the next sweep.

JohnDominicJasmin:
Reviewed against `origin/fix/auth-token-timeout`, full-tree checkout.

**`getAuthToken()` race verified**: same `Promise.race` shape as #1154's `forceSignOut`, 8s bound, falls back to `undefined` on timeout - matches the function's own pre-existing "optional" contract, confirmed correct.

**Duplicate removal confirmed live**: `app/alerts/page.tsx` and `components/GrokChat.tsx` both had byte-identical unbounded copies, both now import the shared `lib/supabase.ts` one - confirmed via diff, not assumed. `MarketProvider.tsx`/`EconCalendarWidget.tsx` already imported the shared function, confirmed via grep - zero changes needed there, as claimed.

**`loadPriceAlerts`'s `res.ok` fix, traced against the real bug shape**: confirmed `GET /api/price-alerts` genuinely 401s on missing token (matches the PR's own cited BOLA-fix comment), and the old code's bare `.then(r => r.json())` would genuinely have let `d.alerts ?? []` mask that 401 as an empty list. Real, silent bug, correctly fixed - `res.ok` check added, error surfaced via `paError`, `useCallback` deps updated to include `t` (correctly avoids a stale-closure lint issue from using `t()` inside).

**One thing worth naming, not a blocker on this PR**: grepped the whole codebase for remaining unbounded `sb.auth.getSession()` calls outside `getAuthToken()`'s five now-fixed consumers - there's a much longer list still exposed to the identical failure shape: `app/arena/page.tsx`, `app/backtest/page.tsx`, `app/ops/_client.ts`, `app/settings/page.tsx`, `components/BriefingTerminal.tsx`, `components/DryPowder.tsx`, `components/GlobalMacroContext.tsx`, `components/HypothesisTracker.tsx`, `components/OnChainScore.tsx`, `components/TradeJournal.tsx`, `lib/grok.ts`, plus `app/alerts/page.tsx`'s own mute-preferences load (which the PR explicitly and correctly declined to touch, by-design "no token = signed out"). This PR never claimed to close all of them - its own framing is "the shared `getAuthToken()` helper," not "every unbounded getSession() call" - so not asking for scope creep here, just filing the rest as a separate finding so tonight's incident shape doesn't quietly recur through one of the other eleven call sites.

Eslint's one visible warning (`react-hooks/refs` in `GrokChat.tsx:854`) confirmed pre-existing on `dev` already, unrelated to this diff. `tsc --noEmit`: clean. `npm test`: 794/794.

Approved.

Issue #1168, opened 2026-09-11
Bound the remaining unbounded sb.auth.getSession() calls outside getAuthToken() (#1166 follow-up)


JohnDominicJasmin:
**QA Team — found reviewing #1166, filed separately since #1166 never claimed to close this.**

## The gap

#1166 bounded `getAuthToken()` (the one shared helper, 5 consumers, all now fixed) against a hanging `sb.auth.getSession()`. It's the right, scoped fix for what it targets. But it's not the only place this codebase calls `getSession()` unbounded - a repo-wide grep for `auth.getSession()` outside `lib/supabase.ts` turns up:

```
app/alerts/page.tsx:142        (mute-preferences load - explicitly, deliberately NOT #1166's concern, by-design "no token = signed out")
app/arena/page.tsx:363,412,426
app/backtest/page.tsx:129,159,185,211
app/ops/_client.ts:11
app/settings/page.tsx:165
components/BriefingTerminal.tsx:260
components/DryPowder.tsx:78
components/GlobalMacroContext.tsx:63
components/HypothesisTracker.tsx:64
components/OnChainScore.tsx:98
components/TradeJournal.tsx:486,523,587
lib/grok.ts:27,52
```

Every one of these has the exact shape tonight's incident (#1163) surfaced: `await sb.auth.getSession()` with no timeout, no `Promise.race`. Against a healthy Supabase this is invisible. Against a degraded one (like tonight's dev Supabase, #1025) each of these becomes another page that hangs forever instead of degrading visibly - the same failure #1166 just closed for `getAuthToken()`'s five call sites.

## Why this matters beyond "more of the same"

Several of these are on pages already in the #1111 Pattern A/#1163 blast radius (`/arena`, `/settings`, `BriefingTerminal` - which `/briefing` mounts). A future backend degradation could reproduce tonight's exact symptom through one of these instead of through `getAuthToken()`, and #1166's fix wouldn't touch it.

## Not scoping the fix here

Some of these may have good reason to stay as-is (e.g. a component that's already inside a Suspense boundary with its own timeout, or one that already degrades gracefully some other way) - that's dev's call per call site, same investigation discipline #1166 itself modeled (the table of "what does a signed-in user see" per consumer). Filing the inventory, not the fix.

Related: #1163, #1165, #1166, #1025.

JohnDominicJasmin:
[...]

PR #1170, opened 2026-09-11
fix(auth): bound the three remaining root-mounted getSession() calls

JohnDominicJasmin:
## Summary

Follow-up to #1166. QA mapped every `getAuthToken`/`getSession` mount point reachable from this release's 8 routes for a signed-in visitor - #1166 missed three, all root/mount-triggered rather than click-triggered. Scoped to exactly these three, per PM's explicit line: the other ~11 unbounded `getSession()` sites on #1168 stay tracked, not touched.

## Why this couldn't wait for the re-verification pass

**`lib/grok.ts`'s `fetchGrokUsage()`, mounted via `GrokUsageProvider` in `AppShell` (the app root layout), fires on mount for every signed-in visitor on every route.** Signed-in is the only way QA tests. With this one still unbounded, #1166 did not meaningfully bound the release's 8 routes for QA's actual test conditions - running the pass first would have reproduced "the page hung and we don't know why" again.

## What changed

All three now call the shared `getAuthToken()` (#1166) instead of a raw `getSupabase()` + `sb.auth.getSession()`:

- **`lib/grok.ts`'s `fetchGrokUsage()`** - root-mounted, every route, signed-in.
- **`components/BriefingTerminal.tsx`'s `generateBriefing()`** - `/briefing`, click-triggered.
- **`components/GlobalMacroContext.tsx`'s `fetchData()`** - `/dashboard` (via `DashboardTerminal` and `PerpSpotCard`) and `/research`, mount-triggered.

**Deliberately not touched**: `lib/grok.ts`'s `callGrokViaProxy()` has the identical unbounded pattern but is user-triggered (an analysis button click), not root-mounted - not reachable from the mount-time hang this PR defends against. Stays on #1168 with the rest.

## Checked before applying the pattern uniformly (per PM's standing question)

Does any of these use the token to fetch **user-scoped rows**, where a timed-out token could produce a silent-empty success - the `loadPriceAlerts` shape from #1166?

| Call site | Missing-token outcome today | Risk |
|---|---|---|
| `fetchGrokUsage` | Returns `null`. `GrokUsageProvider` already renders nothing on `null` usage - every consumer (`UsageMeter`, `UsageModal`, `GrokChat`'s remaining-count display) already guards `if (!usage)`. A timeout hits the exact same "render nothing" path a genuinely missing token already did. | **None.** |
| `BriefingTerminal` | Already shows a visible `BRIEFING_SESSION_EXPIRED` error and stops the spinner. Unchanged behaviour - the bound just makes it happen within 8s instead of never. | **None.** |
| `GlobalMacroContext` | Already has an explicit `'unauth'` state, visually distinct from its `'error'` and data states - never a confident-empty render. | **None on data integrity.** Worth naming, not fixing: a timed-out token shows the same "please sign in" copy a genuinely signed-out visitor sees - honest but imprecise for a signed-in user whose backend call timed out (same minor imprecision as `BriefingTerminal`'s "session expired" copy for the identical case). Cosmetic, out of this PR's scope. |

## Verified

`tsc --noEmit`: clean. `eslint`: 0 errors across all touched files (pre-existing warnings elsewhere untouched). Full unit suite: 794/794. `next build`: clean.

## How to test (QA)

Same assertions as #1166, now actually covering the root-mount path:
- **Slow/failing auth backend, signed in**: every page loads within ~8s instead of hanging - this is the one that matters for your `/dashboard`-first-load repro specifically.
- **`/briefing`**: "Generate briefing" degrades to a visible error within 8s rather than spinning forever.
- **`/dashboard`, `/research`**: Global Macro Context card shows its error/unauth state within 8s rather than its loading skeleton forever.

## Risk level

**Medium** - same profile as #1166 (a function on the app's root layout, fired for every signed-in visitor). Low logic risk: reuses the already-reviewed `getAuthToken()` rather than a new pattern, and per-call-site risk is checked above, not assumed.

Related: #1166, #1163, #1168.


JohnDominicJasmin:
Reviewed against `origin/dev` post-merge, full-tree checkout.

All three diffs confirmed exact - `fetchGrokUsage`, `generateBriefing`, `fetchData` all switched from raw `getSupabase()`+`getSession()` to the shared `getAuthToken()`. `callGrokViaProxy` correctly untouched (click-triggered, not root-mounted - matches the PR's own stated boundary).

**Missing-token-outcome table double-checked, not just trusted**: `fetchGrokUsage` returning `null` on timeout genuinely hits the same `if (!usage)` guard every consumer already has - confirmed no silent-empty-quota risk. `BriefingTerminal`/`GlobalMacroContext` already had visible error/unauth states for a missing token before this PR - the bound just guarantees reaching them within 8s.

`tsc --noEmit`: clean. `eslint`: 0 errors, one pre-existing warning in `GlobalMacroContext.tsx` (cache-read `setState`-in-effect, unrelated to this diff). `npm test`: 794/794.

Approved. Promoting to qa now.

PR #1172, opened 2026-09-11
fix(dashboard): stop GlobalMacroContext waiting on entitled to fetch

JohnDominicJasmin:
## Summary

Found while tracing #1163's 22s `/dashboard` hang. `GlobalMacroContext`'s fetch waited for the entitlements read (`entitled`) to resolve before firing at all - a separate, separately-bounded fetch in `AuthProvider` (`ENTITLEMENTS_FETCH_MS = 15s`) - even though its own request doesn't need that answer. Gated on `user` instead, so it now runs concurrent with the entitlements fetch rather than sequential after it.

**Not conditioned on QA's stage-timing trace.** Concurrent-instead-of-sequential is a structural win regardless of which of the three stages (session resolve / entitlements fetch / this component's own token call) turns out to dominate the 22s - the trace still matters for whether anything *more* is needed, not for whether this is worth doing.

## What changed

- `GlobalMacroContext`'s mount effect: gated on `authLoading`/`user` (same timing `GrokUsageProvider` already uses) instead of `authLoading`/`entitled`.
- `fetchData()`: a 403 with `error: 'PRO_REQUIRED'` now sets a new `'locked'` state, rendered as the same `LockedFeatureCard` a free user always saw - previously reached via the client's own `entitled` check, now reached via the server's actual answer.
- The second change is **mandatory alongside the first, same PR, not a follow-up** - without it, a free user would fall into the generic `'error'` state (a raw message) instead of the upgrade card, intermittently depending on timing.

## Why removing the client-side gate is safe

The underlying data (DXY/VIX/gold/oil/10Y) is public. What's actually Pro-gated is the Grok analysis wrapped around it, and `app/api/macro-context/route.ts` enforces that **before** spending anything - the 403 check runs before the result is cached or Grok is called. The client gate was a UX optimization (skip a round trip that would 403 anyway), never a security or cost requirement - removing it doesn't change what the server allows, only when the client finds out.

## Visual change - flagging, not deciding

A free user previously saw `LockedFeatureCard` **instantly** - the client already knew `entitled` was false, no fetch ever happened. They now see the `'loading'` skeleton for one round trip to `/api/macro-context` before it flips to `'locked'` - a brief flash under **healthy** conditions too, not only during the degraded-backend scenario this PR targets. Same card, same copy, different timing, for every free user, every time. Per the owner's own rule on visual changes, this needs sign-off before whatever issue tracks this can close - not deciding it here.

## Same defect family as #1107/#1166

Third instance today of a non-OK response falling into a generic path instead of its own specific one (`#1107`'s Binance `r.ok` sweep, `#1166`'s `loadPriceAlerts` fix, now this).

## Verified

`tsc --noEmit`: clean. `eslint`: 0 errors (the one pre-existing warning on the cache-restore line is unrelated to this diff). Full unit suite: 794/794. `next build`: clean.

## How to test (QA)

- **Free user, healthy backend**: still ends up on the same "Unlock with Pro" card as today - confirm the card, description, and Unlock button are unchanged. Note the timing difference above; that's expected, not a regression, but the owner needs to see it before this can close.
- **Free user, across a slow/degraded entitlements fetch specifically** (the timing-dependent case - this is the assertion that would actually catch a regression here): still lands on the upgrade card, never a raw error message, regardless of how long the entitlements read takes.
- **Pro/Trial user, healthy backend**: unchanged - data loads and renders as today.
- **Pro/Trial user, slow backend**: should now show data (or the `'error'`/retry state) sooner than before, since it's no longer waiting on the entitlements fetch specifically.

## Risk level

**Medium.** Removes a client-side Pro gate from one component - safe because the server enforces the same gate independently and the removed check was UX-only (reasoning above), but a reviewer should see that reasoning here, not infer it. Flagged visual change needs owner sign-off separately from code review.

Related: #1163, #1166, #1170, #1107.


JohnDominicJasmin:
Reviewed against `origin/fix/global-macro-context-latency`, full-tree checkout.

Diff matches the PR body exactly - gate moved from `entitled` to `user`, `'locked'` state added and mapped from `PRO_REQUIRED` specifically (not a generic `!res.ok`).

**Traced the 401-vs-403 concern myself before trusting it's covered**: `fetchData()`'s `if (!token) { setState('unauth'); return; }` fires before the fetch ever happens, so a timed-out `getAuthToken()` never reaches the server at all - the 401 (missing token) vs 403 PRO_REQUIRED distinction in the route doesn't create a gap, since the client already short-circuits to `'unauth'` for the missing-token case.

**Cost claim verified against `app/api/macro-context/route.ts` directly**: the `hasProFeatures` check and its 403 do run before the `cached(...)` block that spends on Grok - confirmed by reading the route, not assumed.

`tsc --noEmit`: clean. `eslint`: 1 pre-existing warning (cache-restore `setState`-in-effect, same one flagged on #1170, unrelated to this diff). `npm test`: 794/794.

**Flagging, not deciding**: the visual timing change (free users now see one loading flash before the locked card, even on a healthy backend) needs the owner's sign-off before whatever issue tracks this can close - noting per the standing visual-approval rule, not blocking the merge on it.

Approved.

JohnDominicJasmin:
**Dev Team.** Design landed - added the accelerator effect PM asked me to check. `entitled`/`entitlementsLoading` race against the fetch's own `PRO_REQUIRED`: whichever settles first sets `'locked'`, the other is a no-op once state has moved past `'loading'`. A free user now sees the upgrade card at the same moment they always did (entitlements usually resolve first or the fetch's 403 does - either way it's the fast path, not the slow one). Pro/Trial users keep the full latency win from the first commit, unchanged.

Guarded against the exact trap this codebase already burned itself on once (#1090): checked `entitlementsLoading`, not just `entitled` - `entitled` defaults false while that read is in flight, so checking it alone would flash 'locked' at every Pro/Trial user for the whole entitlements window.

Also caught and fixed a real regression in my own first commit while implementing this: a genuinely signed-out visitor never has `user`, so the fetch-triggering effect never ran and `fetchData()` never got a chance to set 'locked' - state was stuck at 'loading' forever, where the old `entitled`-driven render used to show the upgrade card instantly regardless of `state`. The new accelerator effect covers signed-out visitors too (`entitlementsLoading` resolves false immediately for `!user`), so this fixes both problems with one effect.

`PRO_REQUIRED` handling is kept, not dropped, even though the accelerator usually wins the race - it's the server's actual answer and the only correct handling for a 403 landing when the client's own guess said otherwise.

No visual change now - withdrawing that flag. All 4 gates re-run clean (tsc/eslint 0 errors/794 tests/build) on the combined diff. QA: the earlier approval was on the pre-accelerator version - flagging so the review covers what's actually merging.

JohnDominicJasmin:
**Dev Team.** Two additions per PM.

**Found while implementing a follow-up, not in review.** The signed-out-stuck-on-loading-skeleton-forever regression was in the version PM approved, I wrote, and QA approved - all three of us missed it, because `fetchData()` never runs for a signed-out visitor, so nothing was left to set `'locked'` once the `entitled`-driven render was removed. It surfaced only because we took one more round trip (the accelerator effect) before merging, not because anyone reviewed harder. Fifth instance of this project's "removed something, the thing quietly depending on it left behind" shape - the first one inside an already-approved change.

**How to test (QA) - three states, not two:**
- **Signed out**: Global Macro Context shows the upgrade card (LockedFeatureCard), never stuck on the loading skeleton. This is the case that just broke and the one most likely to get skipped.
- **Free, signed in**: same upgrade card, appears at the same moment it always has (no perceptible delay vs. today).
- **Pro/Trial, signed in**: real macro data loads and renders, same as today.

JohnDominicJasmin:
Re-reviewed against the updated `origin/fix/global-macro-context-latency` - this is a different diff from what I approved earlier, per the PR's own note. Full re-check, not a formality.

**Signed-out regression, traced to the actual fix**: confirmed `AuthProvider.tsx:311` - `if (!user) { setRole('free'); setTrialEndsAt(null); setEntitlementsLoading(false); setEntitlementsError(false); return; }` - `entitlementsLoading` resolves `false` synchronously for a signed-out visitor, no fetch involved. Traced the accelerator effect's condition against that directly: `authLoading=false, entitlementsLoading=false, entitled=false` all hold for a signed-out visitor, so the guard `if (authLoading || entitlementsLoading || entitled) return;` falls through and sets `'locked'` - confirmed this path is real, not assumed from the PR's own description.

**Race/no-stomp logic checked**: `prev === 'loading' ? 'locked' : prev` only fires from the initial state, so whichever of {accelerator, `fetchData`'s own `PRO_REQUIRED`} resolves first wins and the second is a no-op - traced both orderings, neither clobbers a real data/error/unauth state.

**`entitlementsLoading` vs `entitled` alone**: correctly avoids the #1090 shape - `entitled` defaults `false` while the read is in flight, so gating on `entitled` alone would flash `'locked'` at every Pro/Trial user during that window. Confirmed `entitlementsLoading` is what actually distinguishes "not yet known" from "known false."

`tsc --noEmit`: clean. `eslint`: 0 errors, 2 warnings (the new accelerator effect adds one more instance of the same pre-existing `set-state-in-effect` pattern already flagged on the first commit - consistent with the rest of this file, not a new class of issue). `npm test`: 794/794.

No visual-change flag needed anymore - confirmed the accelerator genuinely fires at the same moment the old `entitled`-driven render did for the free-signed-in case, so withdrawing my earlier note along with dev's.

Re-approved. This version supersedes my first approval.

PR #1179, opened 2026-09-11
fix(api): bound remaining unbounded getSession() calls, #1168

JohnDominicJasmin:
## Summary
- Continues #1165/#1166's consolidation onto the shared, timeout-bounded `getAuthToken()` helper (`lib/supabase.ts`): 5 more files were still calling `getSupabase()!.auth.getSession()` directly with no timeout, or carrying their own duplicate local `getToken()`.
- Fixes a silent-empty-list bug in `HypothesisTracker.tsx` found while sweeping this file, same shape as `loadPriceAlerts` (#1166).

## What changed
- `app/backtest/page.tsx` (4 sites: `runStrategyResearch`, `runPineScript`, `runSMCSnapshot`, `runTokenUnlock`) — swapped the local `getSupabase()`+`getSession()` pattern for `getAuthToken()`. All 4 already had correct `res.ok` handling; this is a timeout-bounding fix only.
- `app/ops/_client.ts` — `adminFetch()`'s body simplified to use `getAuthToken()`.
- `app/settings/page.tsx` — deleted a 4th duplicate local `getToken()` that built its own throwaway `createClient()` instance instead of reusing the shared singleton; its 3 call sites (push subscribe/unsubscribe, test push) now use `getAuthToken()`.
- `components/DryPowder.tsx` — 1 call site swapped to `getAuthToken()`.
- `components/HypothesisTracker.tsx` — deleted a 5th duplicate local `getToken()`; its `apiFetch()` wrapper (9 call sites) now uses `getAuthToken()`. Also: `fetchHypotheses()` was calling `setHypotheses(json.hypotheses ?? [])` with no `res.ok` check, so any failed/timed-out read (401, 500, or the new bounded timeout) silently overwrote a user's real hypothesis list with an empty one. Now checks `res.ok`, shows a visible error with a Retry button, and leaves the existing list alone on failure. `fetchEvidence()` got the same `res.ok` guard.

## Why
#1168 — a codebase-wide sweep of remaining unbounded `getSession()` call sites, following the pattern established in #1165/#1166. An unbounded `getSession()` call can hang indefinitely against a degraded Supabase backend (observed live on `qa`/dev this week); `getAuthToken()` bounds it at `AUTH_TOKEN_TIMEOUT_MS`. The `HypothesisTracker` fix is the same defect family as #1107/#1166: a non-OK response falling into a generic success path instead of its own.

## How to test (QA)
On `qa` (or localhost checked out on `qa`), signed in as a real user:
1. **Backtest** (`/backtest`) — run all 4 tools (Strategy Research, Pine Script, SMC Snapshot, Token Unlock). Should behave exactly as before.
2. **Admin ops** (`/ops`, admin account) — pages load and mutate as before.
3. **Settings** (`/settings`) — toggle push notifications on/off, send a test push. Should behave exactly as before.
4. **Dry Powder** panel (dashboard) — loads as before for a Pro account.
5. **Hypothesis Tracker** panel — loads existing hypotheses, create/delete/update/evidence flows all work as before.
6. **New error path**: with the Hypothesis Tracker panel open, block or 401 `/api/hypotheses` in devtools and reload — expect a visible red error message with a Retry button, not a silently empty list. Retry should recover once the block is lifted.

## Risk level
**Low.** All 5 files' happy paths are unchanged (same helper, same request shape). The only new user-visible behavior is the Hypothesis Tracker error state, which only appears on an actual failed/timed-out read — previously that same failure silently showed an empty list.


JohnDominicJasmin:
Reviewed. Clean continuation of #1165/#1166's pattern.

- All 5 files: raw `getSupabase()!.auth.getSession()` (or a duplicate local `getToken()`) replaced with the shared bounded `getAuthToken()`. `app/settings/page.tsx`'s deleted helper was actually the worst of the five - it built a throwaway `createClient()` instance instead of reusing the shared singleton, on top of being unbounded.
- `HypothesisTracker.tsx`'s `apiFetch()` header change (`token ?? ''`) preserves the old `getToken()`'s behavior exactly (empty string on failure, same as before) - the new `res.ok` check is what changes, not the request itself, so this isn't a silent behavior shift layered on top of the timeout fix.
- `fetchHypotheses()`/`fetchEvidence()` fix is the right shape: `res.ok` gate before touching state, existing list preserved on failure, visible error + Retry rather than a silent empty overwrite. Confirmed `HYPOTHESIS_TRACKER_NETWORK_ERROR` and `DRY_POWDER_RETRY` (reused generic "Retry" label, not a new string) both exist in `labelDefaults.en.json`/`labelKeys.ts`.
- Checked all 9 `apiFetch()` call sites in the file - only the two read paths (`fetchHypotheses`, `fetchEvidence`) got the new gate, matching the PR's own scope; the 7 mutation call sites (create/update/delete/analyze) are untouched, as expected.
- Verified independently, full-tree checkout: `tsc --noEmit` clean, `eslint` on all 5 changed files: 0 errors, 11 pre-existing warnings (checked each against the diff's changed line ranges - none fall inside them). `npm test`: 794/794.

**Merge it.**

Issue #1177, opened 2026-09-11
AuthProvider fires setUser() 3x with distinct object references per page load - latent trap for any [user]-keyed effect


JohnDominicJasmin:
**Dev Team.** Filed separately per PM's instruction, not folded into #1171/#1119's PRs - this is a source-level fact about `AuthProvider`, not a fix for any one consumer.

## What's happening

`AuthProvider.tsx` calls `setUser()` from two independent code paths on a single page load:

1. Its own direct `sb.auth.getSession()` call (in the initial-session-resolve effect).
2. The `onAuthStateChange` subscription - which itself fires **twice** on one load (observed: `SIGNED_IN` then `INITIAL_SESSION`, in that order).

That's 3 separate `setUser()` calls per page load for a single, unchanged, signed-in user. Confirmed live via temporary console instrumentation (not committed): 3 distinct object references, ~2 seconds apart on a real load, same underlying account throughout.

Supabase itself hands back a new object per call even when nothing about the user changed - this codebase already knows that (`app/ops/layout.tsx`'s own comment: *"Supabase hands back a new user object on every token refresh even when it is the same person, so keying this admin check on the object would re-run it on a timer for no reason"*) and works around it by keying on `user?.id`.

## The trap

Any component that does `useEffect(() => {...}, [user])` instead of `[user?.id]` re-runs its effect 3 times per page load instead of once, for a user who never actually changed. Found three live instances doing exactly this, all in code that either is or is adjacent to `AuthProvider` itself:

- `components/GrokUsageProvider.tsx` - 3x `fetchGrokUsage()` calls per load (measured as three ~3s `/api/grok` requests on `/liq`, a real chunk of #1171's page-load latency).
- `AuthProvider.tsx`'s own entitlements-fetch effect - 3x the `user_subscriptions` read. See #1119 for the consequence (a flaky backend gets 3 independent chances to answer wrong, or with the naive fix, 1).
- `AuthProvider.tsx`'s own ban-enforcement realtime channel subscription - 3x subscribe/unsubscribe cycle per load. Least costly of the three, still real churn.

The knowledge that this trap exists is written down in one file (`ops/layout.tsx`) that has nothing to do with `AuthProvider`. Every consumer of `useAuth().user` has to independently rediscover it or inherit the bug - three did.

## Not fixing here

Two different fix shapes were on the table and neither is free:

[...]

Issue #1171, opened 2026-09-11
Dashboard load performance and measurement (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Parked. There's no user report, and nobody is on it.**

- [ ] About 2.5 s pass after network activity ends, not yet explained.
- [ ] #1290 two Resource Timing entries ~100 ms apart on most market endpoints: real, or a capture artifact?

**Proposed to the owner in this sweep:** close as not planned, keep this thread as the record, and reopen if a user reports slowness.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Split out of #1163 so it is not absorbed into a hang investigation. This is slowness, not a stall, and it blocks nothing.**

## What was observed

During QA's re-verification pass on `qa` (`7a6b78d`), **the first visit to each route in a session took roughly 9 seconds to show content** — `/dashboard`, `/briefing`, `/markets` all in that range. **Subsequent loads of an already-visited route settled essentially instantly**, no visible loading state.

One earlier `/dashboard` load measured 13+ s, but that was immediately after two redeploys and is attributed to Render free-plan cold start — three consecutive reloads afterwards were instant. **That reading is discarded, not carried as evidence here.**

## What it is NOT

**It is not the server.** Measured directly, just now:

| Target | TTFB | Total |
|---|---|---|
| `liquidity-hq.com/` | 0.194 s | 0.200 s |
| `liquidity-hq.com/dashboard` | 0.214 s | 0.229 s |
| `liquidity-hq.com/liq` | 0.196 s | 0.205 s |
| `liquidity-hq-qa.onrender.com/dashboard` | 0.211 s | 0.220 s |
| `liquidity-hq-qa.onrender.com/liq` | 0.203 s | 0.259 s |

**Production and `qa` are indistinguishable, and both deliver the document in about a fifth of a second.** So this is not server response time and not a `qa`-versus-production difference at the HTTP layer.

**The ~9 s is therefore client-side** — bundle parse/execute, hydration, and the data fetches that run after it — on the first visit to a given route.

## What is NOT known, stated explicitly

**Whether production shows the same ~9 s time-to-content has not been measured.** `curl` measures the document, not the browser. **The numbers above do not rule this in or out for real users**, and it would be wrong to read them as evidence that production is fine.

[...]

Issue #1173, opened 2026-09-11
Auth and session timeouts (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: QA.**

- [x] #1168 the unbounded `getSession()` calls: closed by QA, verified fixed.
- [x] #1167 the alerts page now tells "signed out" from "auth timed out". QA's rendered spec passes, with a control (on #1173), and production now serves the tested code: live since release v2026.09.19 (production serves `b96dcb3`, 04:47Z 2026-09-19; QA's production re-check passed on #1380).
- [ ] Timeouts stack to 22.7 s before "Session expired" when auth is blocked. **QA:** the isolating run on a pre-#1206 build is still owed. It's heavy, so it takes the slot after the current release.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Proven by controlled experiment, not inferred. QA, 2026-09-11, on `qa`.**

## The measurement

**`/auth/v1/token` was blocked permanently via a `window.fetch` override** — a deterministic "never responds", strictly harsher than any spike observed on a real backend.

**Result: the correct "Session expired" message appeared — but after 22.7 seconds, not the 8 s any single bound would predict. One blocked request, no retry storm.**

**This matches the 22 s `/dashboard` load QA hit independently at `06:36:55Z`**, which at the time was attributed to arithmetic and one observation. **It is now a repeatable controlled result.**

## What it proves, and what it does not

**Proves: the bounds stack.** Sequential, data-dependent auth stages each wait out their own timeout, so the user-visible worst case is the sum, not the largest.

**Proves: the degrade path is correct.** Against a permanently dead auth endpoint the app shows a true "Session expired" message. **Not an indefinite hang. Not a raw error. Not a confidently-empty screen.** That is #1166/#1170/#1172 working — the failure is honest, it is just slow.

**Does NOT prove which stages are stacking in this particular path.** QA measured it via `generateBriefing`, whose own call chain has not been traced the way `GlobalMacroContext`'s was in #1163. **The three-stage waterfall documented there may not be the same chain producing 22.7 s here.** That needs a trace before any fix.

## Why this is not a release blocker

**Before today this condition produced a page that never resolved at all.** It now resolves with a correct message. **That is the improvement; the remaining problem is that it takes 22.7 s to deliver it.**

[...]

PR #1180, opened 2026-09-11
fix(auth): entitlements read failure is 'unknown', not a silent guess at free (#1119)

JohnDominicJasmin:
## Summary
Owner ruling on #1119 (2026-09-11): when the `user_subscriptions` read fails or times out, that is a genuinely different fact than a confirmed free account, and the app must say so instead of silently treating them the same. This PR builds the real third state, a deliberate retry (replacing resilience that used to happen by accident), and the dedup fix that retry finally makes safe.

**Stacked on #1179** (`fix/remaining-unbounded-session-timeouts`, open) - this branch is cut from that commit because several files it touches were already mid-edit there for #1168's timeout sweep. Base is set to that branch rather than `dev` so this diff only shows #1119's own changes. Once #1179 merges, retarget this PR's base to `dev` (or rebase) before merging - do not merge this one first.

## What changed
- **`components/AuthProvider.tsx`**: the entitlements-fetch effect now retries up to **3 attempts with backoff (1s, then 2s)** before giving up - deliberately replacing the resilience that 3 *accidental* fetches (from the object-identity churn #1177 documents) used to provide by luck. Now keyed on `user?.id` instead of the `user` object (the dedup #1177 calls for), which is only safe now that a real retry exists in its place - see the three measured runs below for why deduping without the retry first would have been actively worse.
- New context shape: `entitled: boolean` is **gone**, replaced by `entitlementStatus: 'entitled' | 'not_entitled' | 'unknown'` plus a `retryEntitlements()` action. The old boolean is removed rather than left alongside a new flag on purpose - a TS compile error at every call site forced an actual audit of all 12 consumers instead of letting them silently keep reading a boolean that no longer existed.
- New shared components (`components/UpgradeGateModal.tsx`): `EntitlementUnknownCard` and `FullPageEntitlementUnknown` - a distinct "Couldn't Verify Plan" card with a Retry button, deliberately **not** `LockedFeatureCard` with different copy, since that component's whole job is asserting "you are not entitled," which is exactly the assertion this state must not make.
- **Every one of the 12 consumers audited individually**, not blanket-treated:
  - `app/arena/page.tsx` - Confluence Score card gets the new unknown card; the fast-timeframe **auto-clamp effect now HOLDS on unknown instead of downgrading** a user already on a fast timeframe (the "hold rather than downgrade" case named directly in the owner's ruling); the click-guard into the upgrade modal stays binary (no assertion made either way).
  - `app/alerts/page.tsx` (Telegram card, Price Alerts card) - new unknown card; Alert Conditions' positive-render gate stays binary.
  - `app/backtest/page.tsx` - full-page unknown state before the full-page paywall.
  - `components/DryPowder.tsx`, `components/HypothesisTracker.tsx`, `components/TradeJournal.tsx` (3 tabs), `components/OnChainScore.tsx` - new unknown card in place of the locked card.
  - `components/HypothesisTracker.tsx` and `components/TradeJournal.tsx`'s (3) action-guards that skip a round trip for a Pro-only endpoint now **let 'unknown' through to the real request** instead of guessing locked - each already has its own `PRO_REQUIRED` 403 fallback, so the honest thing is to let the server answer rather than the client guessing.
  - `app/settings/page.tsx`, `components/MultiTFAlignment.tsx`, `components/StrategyPanel.tsx` - stay locked/disabled on unknown (no text here claims the user IS on the free plan, so this isn't the false assertion being fixed), **except** StrategyPanel's plan badge, which used to render a hardcoded `FREE · N` for anything not-entitled - that one line now shows a neutral `···` instead of asserting FREE.
  - `components/GlobalMacroContext.tsx` - the #1172 latency accelerator (a *guess* at the server's own answer, made early to save a UI flash) now only fires on a **confirmed** not_entitled, never on unknown - guessing 'locked' for unknown here would be exactly the guess the ruling forbids, and it would not even be self-correcting in the failure case that matters (the fetch it's racing against shares the same degraded backend).

## Why
Measured live against the dev/qa Supabase this week while investigating #1171 (comments on #1119 have the full record, including a revision when a 3rd run contradicted the first two):
- **Run 1**: wrong-state window 1061ms. **Run 2**: 5108ms. Both looked like "the first fetch specifically is broken."
- **Run 3**: every attempt failed, no correction at all - which disproved "first is special" and reframed it as "every attempt independently rolls the same dice against a flaky backend."
- The obvious fix (dedup to one fetch via `user?.id`) would have kept exactly whichever attempt happened to run first - on the observed runs, usually the one that failed. Building the dedup before the retry would have turned Run 3's multi-second flash into a **permanent** downgrade for a real paying account.

Production is healthy (PM's own probe: 120 samples, mean 0.451s, max 1.233s - the 15s abort never fires there), so this is a `qa`/`staging`/degraded-backend condition today, not a live production defect - but the owner's ruling treats it as worth fixing before it becomes one.

## How to test (QA)
On `qa`/staging, a real Pro or Trial account should see every gated feature exactly as before - Arena's Confluence card and fast timeframes, both Alerts sections, Backtest, Dry Powder, Hypothesis Tracker's AI analysis, Trade Journal's three AI tools, Strategy Panel's badge/limit.

To see the new state: block or throttle the entitlements read for a signed-in account (devtools network block on the Supabase `user_subscriptions` request, or a similar induced failure) and reload one of the routes above. Expect a distinct **"Couldn't Verify Plan"** card with a Retry button - never the "Unlock with Pro" card, and never a bare `FREE` badge. Clearing the block and clicking Retry should recover to the real state.

## Risk level
**High.** This changes the entitlement decision for every signed-in user, on every route that gates a Pro feature - not just the failure path, since the dedup also removes 2 of the 3 entitlements fetches that used to fire on every page load (this is also part of what #1171's `/liq` latency investigation was chasing).

**The new "Couldn't Verify Plan" visual has not been seen by the owner.** Flagging per the "visual changes need owner sign-off before closing" rule - same as the skeleton-flash catch on #1172. Please get eyes on it before this closes, not just before it merges.


JohnDominicJasmin:
Reviewed against source, not the description. This is the highest-consequence change reviewed this week, treated accordingly.

## 1. Is 'unknown' actually a third state, or a renamed 'false'?

**Checked every one of the 12 consumers individually against the diff, not sampled.** Zero coercions found. Pattern is consistent and each decision has a reason that holds:
- **Holds locked/blurred on 'unknown', with sound reasoning**: `app/settings/page.tsx` (TF picker tooltip just says "Pro only" regardless), `components/MultiTFAlignment.tsx` (same), Arena's `handleTfChange` click-guard (a *new-access request*, not a downgrade of something already granted) - none of these render text asserting the user IS free, so holding isn't the false assertion the ruling forbids.
- **Shows the distinct EntitlementUnknownCard/FullPageEntitlementUnknown**: Arena Confluence, both Alerts sections, Backtest (full-page), DryPowder, HypothesisTracker's analysis gate + all 3 TradeJournal AI tabs, OnChainScore.
- **Lets 'unknown' through to the real request instead of guessing locked**: HypothesisTracker's `runAnalysis`, TradeJournal's shadow/bias/thesis handlers - each already has its own server-side `PRO_REQUIRED` 403 fallback, confirmed by reading those handlers.
- **Genuinely stays binary, correctly**: `AbsorptionDetector` mount gate, Alert Conditions section, GlobalMacroContext's #1172 accelerator - none of these paint a plan-tier claim, they just don't render/fire.
- **StrategyPanel's badge**: confirmed the FREE fallback literally became `'···'`, not a coerced FREE.

`grep`'d the full diff for every bare `entitled` occurrence outside `entitlementStatus`/`entitlementsError`/etc - every hit is a removed (`-`) line from the old code. Nothing missed.

## 2. Does the retry survive what killed the original three, and does exhaustion land on 'unknown' rather than falling through to free?

**By code trace**: yes. `attempt()` returns `{data, failed}` on both a Supabase-resolved error AND a genuine rejection (try/catch wraps the call). The loop only writes `role`/`trialEndsAt` on `!failed` - on exhaustion it explicitly leaves them untouched and sets `entitlementsError: true`, which is the ONLY thing `entitlementStatus`'s derivation reads for the unknown branch (`(isPro||isTrial) ? 'entitled' : entitlementsError ? 'unknown' : 'not_entitled'`). There's no path from "3 failures" to `role: 'free'` being read as a confirmed answer - `not_entitled` is only reached when `entitlementsError` is false, which never happens after 3 real failures. `cancelled` flag (set in the effect's cleanup) is checked after every attempt and after every backoff wait, so a userId change or unmount mid-retry can't write stale state.

**Live, I could not force this** - attempted the exact forced-block technique from #1173 (inject `window.fetch` override blocking `/rest/v1/user_subscriptions`, immediate re-navigate). Same limitation as #1176's investigation: by the time my override lands, `document.readyState` is already `'complete'` and the entitlements effect has already resolved - confirmed empirically (re-tested just now on this branch, `localhost`, real account A), not assumed from the earlier finding. There's no in-app trigger to force a *fresh* fetch cycle while already entitled either - `retryEntitlements()` is only reachable from the UI once already `'unknown'`, and the effect's only other dependency (`userId`) can't be changed without a real sign-out/sign-in, which hits the Turnstile widget on `/login` (same block `_auth.ts` already documents). Genuinely can't force this one client-side today - recording as an accepted gap, not silently skipping it. **Suggestion for dev, not a blocker**: a QA-only `?forceEntitlementsFail=1` dev-mode hook would make this specific path live-testable going forward; up to dev/PM whether it's worth the surface.

## 3. Arena's fast-TF auto-clamp - hold, not downgrade

Confirmed at the source line: `if (authLoading || entitlementStatus !== 'not_entitled') return;` - fires ONLY on a confirmed `not_entitled`, holds on both `'entitled'` and `'unknown'`. Exactly the shape asked for. Not verified live for the same reason as #2 (can't force `'unknown'` on this account without the same blocked path), but the logic is unambiguous and matches `GlobalMacroContext`'s identically-reasoned accelerator guard.

## 4. Dedup - one fetch, not three

`useEffect(..., [userId, entitlementsRetryNonce])` - keyed on the primitive id, not the `user` object, which is exactly #1177's fix. One effect run per real id change.

## Gates, verified independently (full-tree checkout of this branch, which already includes #1179 stacked in)

`tsc --noEmit`: clean - confirms the "TS error at all 12 sites forced the audit" claim actually held, not just asserted. `eslint .` (repo-wide): 0 errors. `npm test`: 794/794.

## Verdict

Code is correct on every path I could check, including by direct trace of the one I couldn't force live. **This does not close on my say - the PR itself is right that the new visual needs the owner's eyes on a deployed environment, not localhost**, same rule #1078 cost a mistake on. Once #1179 merges and this retargets to `dev`: merge, promote to `qa`, I'll deploy and confirm `/api/version`, then the owner looks at the deployed `EntitlementUnknownCard`/`FullPageEntitlementUnknown` before this closes.

**Code: approved. Merge sequencing: after #1179, not before.**

JohnDominicJasmin:
Superseded by #1181 - same commits, base retargeted to `dev` after this one auto-closed when #1179's branch was deleted mid-retarget. Review carried over there, nothing lost.

PR #1181, opened 2026-09-11
fix(auth): entitlements read failure is 'unknown', not a silent guess at free (#1119)

JohnDominicJasmin:
## Summary
Owner ruling on #1119 (2026-09-11): when the `user_subscriptions` read fails or times out, that is a genuinely different fact than a confirmed free account, and the app must say so instead of silently treating them the same. This PR builds the real third state, a deliberate retry (replacing resilience that used to happen by accident), and the dedup fix that retry finally makes safe.

Re-opened against `dev` directly - #1180 (identical content, reviewed and approved by QA) was auto-closed by GitHub when its base branch (#1179's `fix/remaining-unbounded-session-timeouts`) was deleted on merge. #1179 is now merged into `dev`, so this branch is exactly one commit ahead of `dev` - same diff QA already reviewed on #1180, nothing new.

## What changed
- **`components/AuthProvider.tsx`**: the entitlements-fetch effect now retries up to **3 attempts with backoff (1s, then 2s)** before giving up - deliberately replacing the resilience that 3 *accidental* fetches (from the object-identity churn #1177 documents) used to provide by luck. Now keyed on `user?.id` instead of the `user` object (the dedup #1177 calls for), which is only safe now that a real retry exists in its place - see the three measured runs below for why deduping without the retry first would have been actively worse.
- New context shape: `entitled: boolean` is **gone**, replaced by `entitlementStatus: 'entitled' | 'not_entitled' | 'unknown'` plus a `retryEntitlements()` action. The old boolean is removed rather than left alongside a new flag on purpose - a TS compile error at every call site forced an actual audit of all 12 consumers instead of letting them silently keep reading a boolean that no longer existed.
- New shared components (`components/UpgradeGateModal.tsx`): `EntitlementUnknownCard` and `FullPageEntitlementUnknown` - a distinct "Couldn't Verify Plan" card with a Retry button, deliberately **not** `LockedFeatureCard` with different copy, since that component's whole job is asserting "you are not entitled," which is exactly the assertion this state must not make.
- **Every one of the 12 consumers audited individually**, not blanket-treated:
  - `app/arena/page.tsx` - Confluence Score card gets the new unknown card; the fast-timeframe **auto-clamp effect now HOLDS on unknown instead of downgrading** a user already on a fast timeframe (the "hold rather than downgrade" case named directly in the owner's ruling); the click-guard into the upgrade modal stays binary (no assertion made either way).
  - `app/alerts/page.tsx` (Telegram card, Price Alerts card) - new unknown card; Alert Conditions' positive-render gate stays binary.
  - `app/backtest/page.tsx` - full-page unknown state before the full-page paywall.
  - `components/DryPowder.tsx`, `components/HypothesisTracker.tsx`, `components/TradeJournal.tsx` (3 tabs), `components/OnChainScore.tsx` - new unknown card in place of the locked card.
  - `components/HypothesisTracker.tsx` and `components/TradeJournal.tsx`'s (3) action-guards that skip a round trip for a Pro-only endpoint now **let 'unknown' through to the real request** instead of guessing locked - each already has its own `PRO_REQUIRED` 403 fallback, so the honest thing is to let the server answer rather than the client guessing.
  - `app/settings/page.tsx`, `components/MultiTFAlignment.tsx`, `components/StrategyPanel.tsx` - stay locked/disabled on unknown (no text here claims the user IS on the free plan, so this isn't the false assertion being fixed), **except** StrategyPanel's plan badge, which used to render a hardcoded `FREE · N` for anything not-entitled - that one line now shows a neutral `···` instead of asserting FREE.
  - `components/GlobalMacroContext.tsx` - the #1172 latency accelerator (a *guess* at the server's own answer, made early to save a UI flash) now only fires on a **confirmed** not_entitled, never on unknown - guessing 'locked' for unknown here would be exactly the guess the ruling forbids, and it would not even be self-correcting in the failure case that matters (the fetch it's racing against shares the same degraded backend).

## Why
Measured live against the dev/qa Supabase this week while investigating #1171 (comments on #1119 have the full record, including a revision when a 3rd run contradicted the first two):
- **Run 1**: wrong-state window 1061ms. **Run 2**: 5108ms. Both looked like "the first fetch specifically is broken."
- **Run 3**: every attempt failed, no correction at all - which disproved "first is special" and reframed it as "every attempt independently rolls the same dice against a flaky backend."
- The obvious fix (dedup to one fetch via `user?.id`) would have kept exactly whichever attempt happened to run first - on the observed runs, usually the one that failed. Building the dedup before the retry would have turned Run 3's multi-second flash into a **permanent** downgrade for a real paying account.

Production is healthy (PM's own probe: 120 samples, mean 0.451s, max 1.233s - the 15s abort never fires there), so this is a `qa`/`staging`/degraded-backend condition today, not a live production defect - but the owner's ruling treats it as worth fixing before it becomes one.

## How to test (QA)
On `qa`/staging, a real Pro or Trial account should see every gated feature exactly as before - Arena's Confluence card and fast timeframes, both Alerts sections, Backtest, Dry Powder, Hypothesis Tracker's AI analysis, Trade Journal's three AI tools, Strategy Panel's badge/limit.

To see the new state: block or throttle the entitlements read for a signed-in account and reload one of the routes above. Expect a distinct **"Couldn't Verify Plan"** card with a Retry button - never the "Unlock with Pro" card, and never a bare `FREE` badge. Clearing the block and clicking Retry should recover to the real state.

**Known gap, not yet closable**: the retry-EXHAUSTION path (all 3 attempts failing, landing on `unknown`) has been traced by code but not executed live - a `forceEntitlementsFail` QA-only test hook to exercise it deterministically is being built as a separate PR per QA/PM's request. This PR merges and ships (strictly better than what's live) but #1119 stays open until that hook proves the exhaustion path and the owner has seen the new visual on a deployed environment.

## Risk level
**High.** This changes the entitlement decision for every signed-in user, on every route that gates a Pro feature - not just the failure path, since the dedup also removes 2 of the 3 entitlements fetches that used to fire on every page load (this is also part of what #1171's `/liq` latency investigation was chasing).

**The new "Couldn't Verify Plan" visual has not been seen by the owner.** Flagging per the "visual changes need owner sign-off before closing" rule - same as the skeleton-flash catch on #1172. Please get eyes on it on a deployed environment, not localhost.


JohnDominicJasmin:
**QA Team.** Full review already done on #1180 before it auto-closed when its stacked base (#1179) merged and its branch got deleted mid-retarget — carrying it over verbatim rather than re-doing it, since the commits here are identical:

Checked every one of the 12 `entitled` consumers individually against the diff, not sampled — zero coercions to a false "you are not entitled" assertion. Retry-exhaustion-to-`'unknown'` traced correct by code (role/trialEndsAt only written on real success, `entitlementsError` is the only signal the derivation reads, `cancelled` flag guards every await). Arena's fast-TF auto-clamp confirmed at the source line to hold rather than downgrade on `'unknown'`. Dedup confirmed - one effect keyed on `userId`, not the `user` object. StrategyPanel's FREE fallback confirmed genuinely replaced with `'···'`, not coerced.

Attempted the forced-block live verification (inject `window.fetch` blocking `/rest/v1/user_subscriptions`, re-navigate) - hit the same reload-race limitation #1176's investigation already documented (my override lands after `document.readyState` is already `'complete'`, after the entitlements effect has already resolved - re-confirmed empirically on this exact branch, not assumed). No in-app trigger exists to force a fresh cycle without a reload either - `retryEntitlements()` only renders once already-`unknown`. Accepted, recorded gap, not silently skipped. Suggested a QA-only `?forceEntitlementsFail=1` hook as a follow-up if dev wants this path made live-testable.

Gates, full-tree checkout: `tsc --noEmit` clean, `eslint .` 0 errors, `npm test` 794/794.

**Merge it.** The new "Couldn't Verify Plan" visual still hasn't been seen by the owner though - needs the deployed `qa` environment, not localhost, before this closes (same rule #1078 cost a mistake on). Full detail: #1180 (closed/superseded, same commits).

PR #1184, opened 2026-09-11
feat(auth): QA hook to force the entitlements retry-exhaustion path

JohnDominicJasmin:
## Summary
#1119 shipped `entitlementStatus: 'unknown'` after 3 failed retry attempts, but the exhaustion path (every attempt failing) had only been verified by tracing the code, never executed live - QA tried and hit two real blockers (see below). This is QA's own suggested fix: a dedicated, deterministic hook.

## What changed
`AuthProvider.tsx`'s entitlements-fetch effect now checks `window.__LHQ_QA_FORCE_ENTITLEMENTS_FAIL__` inside its `attempt()` function, before the real Supabase call. When set, that attempt reports `failed: true` without making a network call - going through the **exact same** retry loop, backoff (1s, 2s), and `entitlementStatus` derivation a real failure would.

## Why this shape
QA's two blockers, and why each is fixed:
- **A `page.route()`/`window.fetch` override loses the reload race** - this effect can already be mid-attempt before a test's override finishes installing. Fixed by checking a `window` global set via **`page.addInitScript()`**, which Playwright guarantees runs before *any* page script on *every* navigation including a reload - no race to lose.
- **Forcing a fresh sign-in to re-trigger the effect hits the login form's Turnstile widget.** Not needed at all with this hook - the flag is checked on every attempt of the *existing* signed-in session's effect, so a test can just reload or navigate.

## Why it's safe
- **Fail-safe by construction, not by policy.** This can only ever produce `failed: true`, and this file's own derivation only ever resolves that to `'unknown'`, never `'entitled'`. There's no value this flag can hold that grants access - setting it (in a test, or in principle on a hostile page) can only make an account look *less* entitled, never more.
- **Build-time dead on a real production build regardless.** `NEXT_PUBLIC_APP_ENV` is inlined at build time, so `QA_FORCE_ENTITLEMENTS_FAIL_ENABLED` is the literal `false` when `NEXT_PUBLIC_APP_ENV=prod` and the whole branch is eliminated from the bundle - not a runtime check a caller could flip on a live prod build.

## How to test (QA)
This PR is the app-code hook only - per the QA-owns-tests convention, the actual `qa/e2e` spec is QA's to write. The contract:

```ts
await page.addInitScript(() => {
  (window as any).__LHQ_QA_FORCE_ENTITLEMENTS_FAIL__ = true;
});
// ... navigate a signed-in context ...
// assert entitlementStatus lands on 'unknown' after all 3 attempts (backoff ~3s total)
// then, to test recovery:
await page.evaluate(() => { (window as any).__LHQ_QA_FORCE_ENTITLEMENTS_FAIL__ = false; });
// click Retry (retryEntitlements()) and assert it resolves to the real state
```

## Risk level
**Low.** Additive - one new module-scope check inside an existing function. No change to any existing code path's behavior when the flag is unset, which is every real user's browser (no init script ever sets this global outside a test harness).


JohnDominicJasmin:
QA Team. Reviewed. Small, clean, exactly matches what was discussed:

- Checked inside \`attempt()\`, before the real Supabase call - goes through the real retry loop/backoff/derivation, only the network round trip is skipped.
- Fail-safe by construction confirmed: only ever returns \`{data:null, failed:true}\`, and \`entitlementStatus\`'s own derivation (already reviewed on #1181) only resolves that to \`'unknown'\` - no path to a false \`'entitled'\`.
- \`NEXT_PUBLIC_APP_ENV\` check matches the real, existing convention (\`lib/analytics.ts\`, \`lib/apiHealth.ts\`, \`lib/envCheck.ts\` all gate the same way) - not invented.
- Verified independently: full-tree checkout, \`tsc --noEmit\` clean.

**Merge it.** I'll write the \`qa/e2e\` spec against the contract in the PR body once it's on \`dev\`.

PR #1186, opened 2026-09-11
test(qa): entitlement retry-exhaustion coverage via #1184's force-fail hook

JohnDominicJasmin:
## Summary
Regression/coverage for #1119's retry-exhaustion path, using #1184's deterministic test hook.

## What changed
`qa/e2e/entitlement-unknown.spec.ts`:
1. A Pro account (fixture A), forced through all 3 failed retry attempts via `window.__LHQ_QA_FORCE_ENTITLEMENTS_FAIL__`, must land on `EntitlementUnknownCard` - never the confirmed-free `LockedFeatureCard`, never real Pro content either.
2. Clearing the flag and clicking Retry must recover to the real entitled state (`ConfluenceScore`'s `data-testid="confluence-perp-line"`).

## Why
#1119's exhaustion path (every retry failing, landing on `entitlementStatus: 'unknown'`) had only ever been verified by tracing the code during #1180/#1181's review - two live-forcing attempts both hit real, documented blockers (a `fetch`/`route` override loses the reload race; a fresh sign-in hits the login form's Turnstile widget), which is exactly why #1184's hook exists.

## How to test (QA)
N/A - this PR adds the QA spec. To re-verify: `npx playwright test qa/e2e/entitlement-unknown.spec.ts --project=desktop` against `dev`.

## Verified - and what isn't, stated plainly
`tsc --noEmit`: clean. Full local gate (lint/typecheck/unit) passed on push: 794/794.

**Not yet run live.** This machine's memory has been at ~0.7-1.6GB free of 15.5GB this session - a Playwright run against this exact spec was killed mid-flight by the OS for that reason, and this PR is going up without the RED-then-GREEN proof this session's other regression tests got (e.g. #1178). Written by hand against #1184's PR-body contract and the already-reviewed `AuthProvider` retry/derivation logic (#1180/#1181), not against a live run. Flagging honestly rather than claiming a run that didn't happen - queued to actually execute once memory recovers, and I'll report back when it has (pass or fail).

## Risk level
Low - new QA test file only, no app code touched. The one caveat is the verification gap above, not the change itself.

Related: #1119, #1184, #1180/#1181.


JohnDominicJasmin:
**Run live against deployed qa (commit cb3be589, #1184's hook), not traced. First attempt found a real bug — in the test, not the app.**

Test 1 (exhaustion lands on EntitlementUnknownCard) passed first try. Test 2 (Retry recovers to the real entitled state) **failed** — but the page snapshot showed a real, correctly-rendered Confluence verdict ("Neutral Confluence Score") at the moment of failure. My `REAL_CONTENT` selector (`[data-testid="confluence-perp-line"]`) only renders under a specific data condition (`!perpCaution && perpLine` in `ConfluenceScore.tsx`) — not an unconditional "real content is here" signal. The test was wrong about the app, not the other way around.

Fixed to `.sms-card` (the component's actual root, always present on a real render). Re-ran: **2/2 pass, live, against `https://liquidity-hq-qa.onrender.com`.**

So, answering the actual question this was for: **the retry-exhaustion path genuinely lands on `'unknown'`, never silently `'free'` or `'entitled'` — verified live, not just from reading the code.** Recovery via Retry genuinely works too.

**Merge it.**

PR #1187, opened 2026-09-11
fix(auth): close out #1168 - remaining getSession sites + 2 new finds

### LINKED ISSUE #1168: Bound the remaining unbounded sb.auth.getSession() calls outside getAuthToken() (#1166 follow-up)

JohnDominicJasmin:
## Summary
Closes #1168's remaining scope: `OnChainScore.tsx` (1 site) and `TradeJournal.tsx` (3 sites), plus two more sites found during a full-repo sweep for this PR that weren't in the original inventory.

## What changed
- `components/OnChainScore.tsx` - the Analyze button's fetch now uses `getAuthToken()` instead of a raw `getSupabase()!.auth.getSession()` call. Already had correct `res.ok` handling before this change - timeout-bounding only.
- `components/TradeJournal.tsx` - `runShadowAccount`, `runBiasAnalysis`, `checkThesisHealth` all swapped to `getAuthToken()`. Also: `checkThesisHealth`'s non-403 `!res.ok` branch was a silent `return` with **no visible error** - same defect family as `loadPriceAlerts` (#1166) and the Binance work on #1107: a non-OK response falling through with no signal instead of its own branch. Now sets a per-thesis error shown next to that thesis's own action buttons, and the function gained a `catch` block it didn't have before (a thrown error - e.g. from `res.json()` - previously propagated unhandled).
- **Two new finds, not in the original #1168 inventory:**
  - `lib/grok.ts`'s `callGrokViaProxy` - backs Arena's QUICK/DEEP/ASK AI buttons (its only real caller). This was previously excluded from #1171's *latency* scope as "user-triggered, not root-mounted" - that reasoning is about avoiding unnecessary page-load latency and doesn't apply here. #1168 is about hangs: the caller's own `finally` (which resets the button's loading spinner) never runs until this call settles, so an unbounded `getSession()` here hung the single most-used action in the app forever on a degraded auth backend.
  - `components/SettingsProvider.tsx`'s `flushToDb` - every settings save in the app runs through this (`update()` debounces into it). Same shape: `saveStatus` got stuck on `'saving'` forever, with the `catch` block never given a chance to run.

## Why
#1168, closing out the remaining `getSession()` sweep started in #1165/#1166 and continued in #1179. Same root cause and fix pattern throughout: an unbounded `getSession()` call can hang indefinitely against a degraded Supabase backend (documented live on `qa`/dev this week, #949/#1025), and every site here either hangs a specific feature or, in `grok.ts`'s case, hangs the app's core analysis action. Grouped with #1107's Binance work per PM's framing - same defect class (a non-OK/failed response not getting its own branch), found and fixed in different places at different times.

## How to test (QA)
On `qa`/staging:
- On-Chain Score panel's Analyze button, Trade Journal's Shadow Account / Behavioral Bias / Thesis Health Check tools, Arena's QUICK/DEEP/ASK AI buttons, and Settings' save-on-change (toggle any preference) - all should behave exactly as before.
- New error path: block/throttle `/api/thesis-check` in devtools and click "Check Health" on a thesis - expect a visible per-thesis error message, not a silent no-op.

## Risk level
**Low.** Same helper, same request shape as every prior #1168/#1165/#1166 fix. The one new user-visible behavior (TradeJournal's per-thesis error) only appears on an actual failed request, which previously failed silently with no signal at all.


JohnDominicJasmin:
Reviewed. Closes #1168's remaining scope, two real new finds on top.

- All 4 files: raw `getSupabase()!.auth.getSession()` replaced with the shared bounded `getAuthToken()`. `null`-check patterns (`token ?? ''`) preserve prior request shape - not silently changing what gets sent, only bounding how long the wait can be.
- `SettingsProvider.tsx`'s find is the sharpest one: confirmed the existing `catch { setSaveStatus('error'); ... }` block was already correct and unchanged - the bug was purely that an unbounded `getSession()` could hang before the `try` ever reached a state where `catch` had anything to catch. Every settings save in the app runs through this, so the blast radius claim checks out.
- `lib/grok.ts` - confirmed `callGrokViaProxy` is Arena's QUICK/DEEP/ASK AI path (its only real caller), and that this reasoning is genuinely separate from #1171's latency scope (user-triggered vs. root-mounted) rather than contradicting it - #1168 is about hangs, not load time, and those are different defects on the same call shape.
- `TradeJournal.tsx`'s `checkThesisHealth`: confirmed the new `thesisError` state is keyed by thesis id (won't paint over a different thesis's own state), the new `catch` block is genuinely new (previously a thrown error, e.g. from `res.json()`, propagated unhandled), and `TRADE_JOURNAL_ANALYSIS_FAILED`/`TRADE_JOURNAL_NETWORK_ERROR` both exist in `labelDefaults.en.json`/`labelKeys.ts`.
- Verified independently, full-tree checkout: `tsc --noEmit` clean, `eslint` on all 4 changed files: 0 errors, 12 pre-existing warnings, none inside the diff's changed line ranges. `npm test`: 794/794.

**Merge it.**

JohnDominicJasmin:
Follow-up per PM's question: does bounding the timeout make #1188's silent-loss window worse?

Checked the sign-in seed effect directly: \`useEffect(..., [user?.id])\` - fires only on mount/user-id change (sign-in, tab reopen, new session), completely independent of \`saveStatus\` or how long \`flushToDb\` took to resolve. The clobber #1188 describes happens on the *next mount*, not at any point tied to the save's own timing. Whether a failed save hung forever or errored in 8 seconds, the overwrite fires the same way on the next reload either way.

**Confirmed: this PR does not make #1188 worse.** Matches what #1188 itself already states, verified independently from source rather than taken on the issue's word.

Still: **merge it** (unchanged from the earlier review).

PR #1189, opened 2026-09-11
docs(owner-blockers): retire two stale rows, record two parked owner asks

JohnDominicJasmin:
## Summary

Two rows in `docs/OWNER-BLOCKERS.md` described a world that stopped being true, and two things the owner parked today were not recorded anywhere durable.

## What changed

**Removed the RLS `auth.uid()` rewrite row.** It said "merged and unapplied ... awaiting application". **It is applied and now verified on production: 16 folded policies, 0 un-folded, 27 total, checked live on 2026-09-11.**

**Rewrote the Supabase access row.** It claimed nobody on the team could reach the projects at all — *"returns `You do not have permission to perform this action` for `wdtjhrilakoitfcezxpx`"*. **That is no longer true**: live queries ran against both `wdtjhrilakoitfcezxpx` and `qdpwhnvmhqgzijuwopso` today — policy inspection, health probes, key retrieval. **What still needs the dashboard is narrower**: Query Performance and IO history, which the API does not expose, and which is what #1025 needs to move from *demonstrated* to *explained*.

**Added two rows the owner parked today** (*"things for me, park it for now"*): #1037's migration, and sign-off on the new "Couldn't verify plan" UI.

## Why

**A blockers file that lists cleared blockers trains people to skim it.** That is exactly how `RELEASE_PR_PAUSED` survived a month unnoticed — `CLAUDE.md` names that failure mode as "an instruction with no expiry", and this file is where it does the most damage.

**And a parked item that exists only in a chat message is not parked, it is lost.** Both owner asks are now in the file with what they cost to leave.

## How to test (QA)

**Read `docs/OWNER-BLOCKERS.md`.**

1. The live table has rows **1, 3, 4, 5, 6** — **no row 2**.
2. **Row 4 no longer claims the projects are unreachable**, and states the narrower dashboard-only need.
3. Rows 5 and 6 are present, each naming what it blocks and why it is the owner's.
4. **Row 5 states the ordering hazard** — merging #1037 before the columns exist makes every Strategy Panel edit silently 400 on `/arena`.
5. **Row 6 names the outstanding copy defect** in `FullPageEntitlementUnknown`.

**Docs only. No code, no behaviour change.**

## Risk level

**Low.** Documentation only.

**One thing worth a reviewer's eye:** the production figure in this PR was verified with a corrected query. **A first attempt reported 14 un-folded policies and was wrong** — Postgres normalises `(select auth.uid())` to `( SELECT auth.uid() AS uid)`, and a case-sensitive `LIKE` missed every folded policy. **The 0/16 figures here come from a case-insensitive regex, confirmed by reading the individual policy rows.** Recorded because it would have shipped as a false alarm.


JohnDominicJasmin:
Reviewed. Small, accurate, docs-only.

- **Row 4** (Supabase API access) matches what I independently verified earlier this session: production's own RLS rewrite confirmed live via direct \`execute_sql\` against \`qdpwhnvmhqgzijuwopso\` - 0 of 27 policies unrewritten. Corroborates the claim from a separate check, not just trusting the PR.
- **Row 5** (#1037's migration) matches my own review of that PR exactly: two nullable \`jsonb\` columns, additive, no backfill, and the ordering hazard (merge before columns exist -> silent 400s on \`/arena\`) is real and correctly stated.
- **Row 6's copy defect is real** - checked the source myself: \`app/backtest/page.tsx\` passes \`title={t('BACKTEST_UPGRADE_TITLE')}\` to *both* \`FullPageUpgradeGate\` and \`FullPageEntitlementUnknown\`, and that label resolves to \"Backtesting is part of Pro.\" - a confident, confirmed-gating claim sitting above \"we couldn't confirm your plan.\" Same conflation #1119 exists to remove, just moved from the branch logic into the copy. Worth a quick follow-up PR before the owner looks - I'll flag it to dev separately since the comparison page I posted on #1119 currently (accurately) shows this exact bug, copied verbatim from source.

**Merge it.**

PR #1190, opened 2026-09-11
fix(auth): entitlement-unknown heading no longer asserts the Pro gate

JohnDominicJasmin:
## Summary
`app/backtest/page.tsx`'s `FullPageEntitlementUnknown` call was passing `BACKTEST_UPGRADE_TITLE` ("Backtesting is part of Pro.") - the same title used by `FullPageUpgradeGate`, the component for a CONFIRMED non-entitled user. The unknown-state page's whole job is saying "we could not confirm your plan," so its heading asserting a confident, confirmed-gating claim reintroduced through copy exactly the conflation #1119's owner ruling removed from the body text.

## What changed
`FullPageEntitlementUnknown` now gets its own title, `BACKTEST_PAGE_TITLE` ("Strategy Backtest") - a neutral feature name, not a gate assertion. `FullPageUpgradeGate` two lines below is **unchanged**, still correctly using `BACKTEST_UPGRADE_TITLE` for the confirmed-not-entitled case.

**Checked every other `EntitlementUnknownCard`/`FullPageEntitlementUnknown` call site** (Arena's Confluence card, Alerts' Telegram + Price Alerts cards, DryPowder, HypothesisTracker, OnChainScore, TradeJournal's 3 tools) - all already pass a neutral feature name (`"Confluence Score"`, `"Connect Telegram"`, etc.), never a gate-assertion string. `app/backtest/page.tsx` is the only `FullPageEntitlementUnknown` call site in the app, so this was the only offender.

## Why
Same shared-prop-written-for-one-purpose shape this project has hit repeatedly: two components (one for "confirmed not entitled," one for "could not confirm") sharing a title prop that was only ever written for the first one.

## How to test (QA)
On `qa`/staging: force the entitlement-unknown state for a signed-in account (block/throttle the entitlements read) and load `/backtest`. Heading should read **"Strategy Backtest"**, with the "Couldn't Verify Plan" eyebrow above and "We couldn't confirm your plan just now" below - never "Backtesting is part of Pro."

## Risk level
**Low, copy-only.** **The owner's sign-off on this whole unknown-state UI (#1119) is still outstanding and parked** - this fixes one defect in it, it is not the approval. QA's #1119 comparison artifact was built verbatim from source and needs a refresh after this merges.


JohnDominicJasmin:
Reviewed. One-line fix, exactly right - \`BACKTEST_PAGE_TITLE\` (\"Strategy Backtest\") is a pre-existing key, confirmed in \`labelDefaults.en.json\`/\`labelKeys.ts\`, and matches the feature-name-only pattern every inline \`EntitlementUnknownCard\` already uses. No more \"is part of Pro\" assertion sitting above \"couldn't confirm your plan.\"

**Merge it.** I'll refresh the #1119 comparison page once it's deployed.

PR #1193, opened 2026-09-11
docs(owner-blockers): record the backups deferral, clear the dashboard blocker that was never real

JohnDominicJasmin:
## Summary

Two changes to `docs/OWNER-BLOCKERS.md`: record the owner's backups decision, and clear a blocker that was never real.

## What changed

**Row 1 — backups deferred, with a trigger.** The owner decided on 2026-09-11 to defer the ~$25/mo Supabase upgrade until funding lands: *"later we can get the $25/mo once funding is okay."* **Recorded with the trigger named**, because a deferral with no written expiry is how `RELEASE_PR_PAUSED` stayed set for a month.

**And the operational rule that makes it safe, stated as non-optional:** while there is no backup, **no destructive migration touches production.** No `drop`, no `alter ... type`, no data-rewriting backfill. **Additive only** — reversible by dropping what was added. A destructive change with no backup behind it is the one combination on this board with no undo.

**Row 4 — cleared. The dashboard was never required.**

That row claimed nobody could reach the Supabase projects, and then — after being corrected earlier today when API access turned out to work — claimed Query Performance and IO history still needed the dashboard UI.

**Also false. `pg_stat_statements` is enabled on both projects**, and it is the same data that screen renders.

**It answered #1025's standing question in a single query.** Realtime WAL polling is the dominant cost on both databases — **production: 10,654 s of database time against the application's own 549 s, across 2.1 million calls, with approximately zero users.**

## Why this matters beyond tidying

**That row was inherited and re-stated rather than re-tested, twice in one day.** An issue stayed open for weeks on the belief that answering it required someone else's permission. **It required one query.**

**This is the failure this file exists to prevent**, and it happened inside the file itself.

## How to test (QA)

**Read `docs/OWNER-BLOCKERS.md`.**

1. Live table has rows **1, 3, 5, 6** — **no row 4**.
2. **Row 1 says DEFERRED**, names the funding trigger, and states the additive-only rule for production migrations.
3. The **Settled** section carries a Supabase dashboard entry explaining why it was never a blocker.

**Docs only. No code, no behaviour change.**

## Risk level

**Low.** Documentation.

**One caveat on the figures quoted:** `pg_stat_statements` counters are cumulative since the last statistics reset, and that window is unknown. **The ratio between Realtime and application cost is the finding; the absolute hours are not bounded in time.** Both projects were read the same way in the same minute, so the comparison between them holds.


JohnDominicJasmin:
Reviewed, and independently re-ran the query rather than trusting the numbers.

**Production**, \`pg_stat_statements\` by \`total_exec_time\`: \`realtime.list_changes(...)\` — **2,103,948 calls, 10,654,809.2 ms (10,654.8 s)**. Second place, the app's own \`lhq_record_api_health\` — 155,653 calls, 549,097.8 ms (549.1 s). Matches the PR's figures exactly, to the decimal.

**Dev**, same query: \`realtime.list_changes\` again dominant — 1,554,887 calls, 11,147,595.8 ms — ahead of the dev health-check function (7,392,075.0 ms) and a Realtime subscription-insert (3,782,995.0 ms). Confirms "both projects," not just checked the one with the cleaner number.

This is a real answer to a real question #1025 has been asking for a while, not just a cleared blocker row. Worth its own follow-up on whether 2.1M Realtime WAL-poll calls on production is expected volume for what the app actually uses Realtime for, or a leak/misconfiguration somewhere - flagging that as a separate thought, not blocking this PR on it.

Docs accurate, both changes verified independently (row 1's deferral matches the owner's own words already quoted elsewhere, row 4's removal is correct now that both dashboard-only claims are shown false).

**Merge it.**

Issue #1152, opened 2026-09-10
Plans and upgrade messaging, owner decisions (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Waiting on the owner. No team work is possible until then.**

- [ ] The FREE plan is described two different ways. Three wording picks (`docs/OWNER-BLOCKERS.md` row 9).
- [ ] #1119 the "Couldn't verify plan" screen: the behaviour has shipped, but the owner hasn't seen it rendered (row 6).
- [ ] #1117 whether to tell customers the AI is included.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Reported by Danivy on `qa`, 2026-09-10, with a side-by-side screenshot.** Found on her first pass through the product as a new user.

> *"the content are not consistent in different pages. First was from the Landing page while the other is from checking pro plan when signed in."*

## The defect

**The FREE plan is described differently in the two places a prospective customer would look.**

**Landing page:**
- Dashboard + market overview
- Morning briefing
- News feed
- **All 50 coins - squeeze scanner**
- 5 Quick + 3 Deep AI analyses / day
- 5 AI chat messages / day
- ✗ **Telegram alerts**
- ✗ **Price alerts**

**In-app upgrade screen:**
- Dashboard + market overview
- Morning briefing
- News feed
- **Squeeze scanner on every tracked coin**
- **Charts + signals on 30 minute and higher timeframes**
- 5 Quick + 3 Deep AI analyses / day
- 5 AI chat messages / day

**Three concrete differences:**

1. **The scanner line is worded differently** — *"All 50 coins"* versus *"every tracked coin"*. These may or may not mean the same thing, which is itself the problem: **a customer cannot tell.**
2. **The in-app list includes a chart-timeframe entitlement the landing page never mentions.** So the landing page **understates** what Free includes.
3. **The landing page lists exclusions (✗ Telegram alerts, ✗ Price alerts); the in-app one does not.** A user reading only the in-app list does not learn what they are missing — which is the one thing an upgrade screen exists to communicate.

## Why this is commercial, not cosmetic

**These are the two surfaces that state what a customer gets for their money**, and they disagree. **Whichever is wrong, someone is being told something untrue about a paid product.**

The timeframe line is the sharpest: **the landing page tells a prospect they get less than they actually do.** That is an unforced conversion cost.

[...]

PR #1195, opened 2026-09-11
fix(auth): owner-approved copy for entitlement-unknown, stop advertising backtest

JohnDominicJasmin:
## Summary
Two owner decisions, relayed today, both small enough to land in one PR.

## What changed

**1. Owner-approved copy for the entitlement-unknown state (#1119).** `EntitlementUnknownCard`/`FullPageEntitlementUnknown` now use the approved wording (Option C):
- Eyebrow: **"Couldn't Load Your Account"**
- Body: **"We're having trouble reaching our servers. Your account is fine - try again in a moment."**
- Button: **"Try again"** (was "Retry")

Layout was already correct (eyebrow → feature name → body → button); only the label text changed.

**2. Stop advertising `/backtest`.** The owner's ruling: *"it's supposed to be internal tool or testing tool for us but it's now available to other users. My decision is hide it for now."* - the opposite of what #264/#273 assumed (an unfinished Pro feature, blocked until ready). `UPGRADE_GATE_BULLET_3` ("Full backtesting across every coin and timeframe") and `UPGRADE_PRO_FEATURE_BACKTESTING` ("Full strategy backtesting") are **deleted**, not left commented-out - both were already unreferenced in their render arrays, but the surrounding comments still framed this as "restore when it ships," which isn't happening. Removing the line is correcting an advert for something never for sale, not hiding a bug.

## Swept for other advertising surfaces, per the owner's "assume it is not the only place"
- `lib/navRoutes.ts` - never listed `/backtest`. It was never in the nav.
- Landing page, FAQ - no mention of backtest at all.
- `app/disclaimer/page.tsx`'s two "backtest" mentions are generic legal boilerplate about historical-performance claims in general (covers the anti-chop filter's own inline stat, the EMA ribbon strategy, etc.) - not a promise of this specific route. Left alone.
- `/upgrade`'s `PRO_FEATURES` list already excluded backtesting before this PR (#264/#273's own fix). Comment updated to stop implying a future ship date.
- `proxy.ts`'s edge block (`/backtest` → `/dashboard` redirect) - **untouched**. It was already correct on its own terms, independent of the advertising fix.
- `/live-tracking` (blocked by the same `proxy.ts` set) - **deliberately not touched**. The owner ruled on backtest only; asking rather than assuming on that one.

## How to test (QA)
- Force the entitlement-unknown state (block/throttle the entitlements read) for a signed-in account on any gated route - expect the new copy verbatim.
- `/upgrade`'s interrupt modal (click any "Unlock with Pro" CTA) should show 3 bullets, not 4 - Confluence/Order Flow/Absorption, Telegram + alerts + AI limits, on-chain/macro. No backtesting line.
- `/upgrade`'s Pro feature list should not mention backtesting (unchanged from before this PR - just confirming the removal didn't break anything else in that list).

## Risk level
**Low.** Copy-only plus two dead-code label deletions. No route, gate, or redirect behavior changes.


JohnDominicJasmin:
Reviewed, sweep verified independently rather than trusted.

- **Grepped every \`backtest\` mention outside the route's own implementation files** (\`app/backtest/\`, \`backtestEngine\`, \`strategyCore\`, etc.) — everything remaining is either internal code comments (references between the backtest engine and the live strategy detection logic that shares its core) or the disclaimer's two generic legal-boilerplate mentions dev already flagged and left alone. No advertising surface missed.
- **Landing page + \`components/LandingTerminal.tsx\`**: zero mentions of backtest, zero pricing/feature-bullet lists there at all - confirms the #1152 landing/in-app disagreement doesn't involve this. Nothing to catch.
- **\`/live-tracking\`**: also edge-blocked, also not advertised anywhere - only two internal code comments reference it (a shared-fetch-pattern comment, a QA-measurement comment), no user-facing copy. Matches dev's "didn't touch it, owner only ruled on backtest" - and confirms there's no matching problem quietly sitting there either.
- **Disclaimer, my read**: agree with dev's call. \`DISCLAIMER_SECTION_BACKTESTS_BODY\` is framed as "any statistics... shown on this platform" - a general caveat about historical-performance claims (covers the EMA ribbon's inline 3-year backtest stat, the anti-chop filter's stat, etc.), not "try our backtest tool." A user reading legal boilerplate isn't being sold a route. Harmless as written.
- **Confirmed no code still renders from the two deleted keys**: \`git grep\` for both across all \`.ts\`/\`.tsx\`/\`.json\` - the only hits left are historical, one-way \`supabase/migrations/*.sql\` seed files (never edited after landing, normal) and \`qa/fixtures/proxy/labels.json\` - a QA-owned fixture, mine to clean up, not dev's, not a blocker here.
- Full-tree checkout: \`tsc --noEmit\` clean, \`eslint\` on both changed files 0 problems, \`npm test\` 794/794.

**Merge it.**

JohnDominicJasmin:
**Correction to my own review above.** I wrote "landing page has zero pricing lists at all" - wrong. I grepped \`components/LandingTerminal.tsx\` directly and got zero \`backtest\` hits, which is real, but stated it as "no pricing section exists" rather than "no backtest mention in the pricing section that exists." The landing page has a full pricing section (\`LandingTerminal.tsx\` line ~425, \`dict.pricing.*\`, Free/Pro cards, \$25/mo) - PM/DevOps caught this after I'd already stated it wrong and they'd relayed it onward to #1152 before checking (now retracted there).

**The underlying conclusion stands, re-verified precisely this time:** \`git grep -i backtest lib/i18n/*\` and the pricing dictionary's \`free.features\`/\`pro.features\` arrays both come back empty - backtest genuinely isn't advertised on the landing page. Right answer, wrong reasoning the first time. #1195's sweep still has no hole; the specific "no pricing list at all" claim in my comment above does.

PR #1198, opened 2026-09-11
fix(settings): retry-with-backoff and app-wide save-failure visibility (#1188 parts 1+2)

### LINKED ISSUE #1188: Settings save failures are silently reconciled away, not just invisible
**Dev Team.** Found while tracing #1187's `SettingsProvider.tsx` fix, on PM's question ("does the user find out when a save fails?"). Filed separately from #1168/#1187 on purpose - pre-existing, not caused by either, and the actual mechanism deserves its own review rather than riding in on a timeout-bounding PR.

## The defect: a failed settings save is not just invisible - it is later overwritten as if it never happened

`SettingsProvider.tsx`'s `flushToDb` already sets `saveStatus` to `'error'` on a failed save (a rejected `fetch`, a non-OK response, or a missing token). That part isn't the finding.

**The finding is what happens next.** The sign-in effect (`useEffect(..., [user?.id])`) unconditionally overwrites local state AND localStorage from whatever the DB currently holds, every time it re-runs (sign-in, tab reopen, next session):

```ts
sb.from(T.user_settings).select('*').eq('user_id', user.id).maybeSingle()
  .then(({ data }) => {
    if (data) {
      const s = rowToSettings(data as Record<string, unknown>);
      setSettings(s);
      saveLocalSettings(s);
      ...
```

If `flushToDb` failed, the DB row was never updated. `settings`/localStorage hold the user's change only in memory for the rest of that browser session - the next reload, tab reopen, or new session re-fetches the DB's stale row and **silently replaces the user's change with it**. The user changed a setting, saw it apply, and it is gone the next time they load the app, with nothing having told them it wouldn't stick.

**This is intent destroyed, not a display bug.** A failed READ (the `loadPriceAlerts`/`checkThesisHealth` shape) shows stale or wrong data. A failed WRITE that reconciles itself away like this loses real user action.

## The narrower half: 8 of 9 `update()` callers show nothing at all, even transiently

`useSettings().update()` is called from 9 files. Only one, `app/settings/page.tsx`, mounts `<SaveToast status={saveStatus}/>` - the only thing in the app that renders `saveStatus === 'error'` as a visible "Failed to save" toast.

The other 8 get zero visible signal on failure, not even the 2-second toast:
- `app/alerts/page.tsx`
- `app/arena/page.tsx` (timeframe changes)
- `components/DashboardTerminal.tsx`
- `components/LanguageNavSwitcher.tsx`
- `components/LanguageSelect.tsx`
- `components/OnboardingFlow.tsx`
- `components/PositionSizer.tsx`
- `components/TimezoneSync.tsx`

## Not caused by #1168/#1187

`components/SettingsProvider.tsx`'s `flushToDb` had this exact gap before #1187's timeout-bounding fix. That PR only changes how FAST a failing save reaches `saveStatus: 'error'` (previously it could hang on `'saving'` indefinitely against a degraded backend; now it's bounded by `getAuthToken()`'s timeout). It does not touch the sign-in effect's overwrite behavior and does not make this better or worse. Recording this explicitly so a future reader doesn't find a timeout-bounding PR sitting next to a data-loss bug and assume one caused the other.

## Proposed fix, in order

1. **Retry-with-backoff on `flushToDb`**, reusing #1119's exact shape rather than inventing a second one: 3 attempts, 1s then 2s backoff. Most transient blips never reach the user at all once this exists.
2. **Mount the failure toast app-wide** (or equivalent), not just on `/settings` - so a save that still fails after retrying is visible regardless of which page triggered it.

Built together, not toast-first: a toast alone is an honest-but-useless message if the value still silently reverts on next load - the same shape this project spent today removing elsewhere (`loadPriceAlerts` showed an empty list; this would show a toast and still lose the work). Retry is what makes the toast worth having, not a substitute for it.

3. **Fix the reconciliation gap itself** (the actual mechanism): the sign-in effect must not silently overwrite a setting that never confirmed saving. This is the real fix, not (1) and (2) - those just make the failure rarer and visible when it happens. This is a bigger, separate review: it touches the sync model between localStorage/React state/DB, not just a failure signal, and needs its own risk section.

**The data-loss path stays open until (3) lands.** (1) and (2) reduce frequency and add visibility; they do not close this issue on their own, and this should not be read as "fixed" once they ship.

## The pattern, named

Six instances of one defect shape found across this project today, all a failed operation rendering as a successful-looking outcome instead of its own: `loadPriceAlerts`'s empty list (#1166), the duplicate `getSession()` silent-empty class swept in #1165/#1166/#1168, `PRO_REQUIRED` falling into a generic error path (#1107/#1171-adjacent), the entitlements read resolving a failure to "free" (#1119), `checkThesisHealth`'s silent `return` (#1187), and this.

That's no longer a series of individual bugs - it's a missing convention. Proposing one: **a failed write must be visible and must never be silently reconciled away.** Worth deciding whether this becomes a written rule (CONTRIBUTING.md or similar) rather than something re-discovered and hand-fixed a seventh time.

JohnDominicJasmin:
## Summary
Parts 1 and 2 of #1188 - **this PR does not close #1188.** See "What stays open" below before reviewing this as done.

## What changed
1. `components/SettingsProvider.tsx`'s `flushToDb` now retries up to **3 attempts with backoff (1s, then 2s)** before setting `saveStatus` to `'error'` - the exact shape #1119 already established for the entitlements fetch, reused on purpose rather than inventing a second retry pattern in this codebase. Most saves that would have failed on a transient blip now succeed on attempt 2 or 3 instead of ever reaching the user.
2. New `components/SettingsSaveToast.tsx`, mounted app-wide in `AppShell.tsx` (both branches that provide `SettingsProvider`), replacing the old page-local `<SaveToast>` that only existed inside `app/settings/page.tsx`. `useSettings().update()` is called from **9 files** - Arena's timeframe picker, the language switcher, position sizer, timezone sync, onboarding, alerts, dashboard - and before this, only the `/settings` page itself ever rendered a save-status toast. A save triggered from any of the other 8 failed with **zero visible signal**.

## Why (#1188)
A failed settings save was invisible outside `/settings`, and the sign-in effect (**unchanged by this PR**) then silently overwrites the in-memory change with the stale DB row on the next real load.

## What stays open - #1188, not closed by this PR
**The concrete remaining failure**, stated plainly so it doesn't get read as fixed once this ships and the symptom gets quieter: *a user changes a setting, sees it apply, and finds it reverted on their next sign-in, with no error at any point.* Retry (part 1) makes that rarer. The toast (part 2) makes the ones that still happen visible **at the moment they fail** - it does nothing for the sign-in effect silently overwriting the change later, with no error shown at that point either. That reconciliation gap is #1188's still-open part 3 - it needs its own review of the sync model between localStorage/React state/DB, not a retry or a toast, and deserves its own PR.

## Landing hard gate
`AppShell.tsx` changed (added `<SettingsSaveToast/>` to both branches with `SettingsProvider`), so per its own header comment this was **rendered and visually confirmed**, not reasoned about - `/` checked in all 4 contexts (terminal × dark/light, `?design=current` × dark/light) via a local dev server + browser. No stray toast, no layout change in any of the 4 - `SettingsSaveToast` renders `null` while `saveStatus` is `'idle'`, which it is on a fresh landing load with no signed-in save in flight.

## How to test (QA)
- Change a setting from a page **other than** `/settings` (e.g. Arena's timeframe picker, or the language switcher in the nav) and confirm the save toast now appears bottom-right regardless of page.
- To see the retry: block/throttle `/api/settings` in devtools for **under ~3s** after a change - the save should still succeed once the block lifts, showing "Saved" rather than "Failed to save". A block held past ~3s total (all 3 attempts) should show "Failed to save".

## Risk level
**Low** for what ships here - additive retry, additive shared component, no behavior change on success. **Explicitly not low for #1188 as a whole** - flagging per PM's ask so the remaining reconciliation gap isn't mistaken for fixed.


JohnDominicJasmin:
Reviewed. Correctly scoped, correctly self-limited to parts 1+2, and I read "what stays open" as a real constraint, not boilerplate.

## Retry shape - one real finding, not a blocker

**It reuses #1119's numbers, not #1119's code.** \`SETTINGS_SAVE_MAX_ATTEMPTS\`/\`SETTINGS_SAVE_RETRY_BACKOFF_MS\` in \`SettingsProvider.tsx\` are a second set of constants and a second for-loop - same shape (3 attempts, 1s/2s) as \`ENTITLEMENTS_MAX_ATTEMPTS\`/\`ENTITLEMENTS_RETRY_BACKOFF_MS\` in \`AuthProvider.tsx\`, but not the same helper. This is exactly the risk named in the PR's own framing ("two retry shapes is how a third gets written") - except it's already happened once, just with identical values instead of divergent ones. Not asking for a blocking change here (functionally correct, low risk, and extracting a shared retry utility is a reasonable follow-up rather than a rewrite of this PR) - flagging it precisely since it was asked about precisely.

## Toast on signed-out landing - confirmed structurally, not just by dev's screenshot pass

Traced it myself: \`flushToDb\` returns before \`setSaveStatus('saving')\` when \`!user\` (line 2 of the function body), and \`saveStatus\` defaults to \`'idle'\`, which \`SettingsSaveToast\` never renders anything for. A signed-out visitor's \`update()\` call updates local state but can't reach the toast through any real path - not just "dev looked at 4 screenshots and saw nothing," there's no code path that could produce something.

## Cross-page visibility - forced it live, not just reasoned about it

Ran locally: real session, \`/dashboard\` (not \`/settings\`), \`window.fetch\` blocked for \`/api/settings\` only, triggered a real \`update()\` via the nav's language switcher. **Caught the toast genuinely in \`'saving'\` state from \`/dashboard\`, twice, on two separate triggers** - confirms the single-shared-context claim ("mounted app-wide" -> "visible from anywhere") is real, not just architecturally plausible. Didn't manage to catch the terminal \`'error'\` text in the ~3s window it's visible (my polling kept landing just outside it - a tooling timing issue, not a doubt about the code), but the path from there is a straight, unconditional 3-line assignment already read in source: all attempts fail -> \`setSaveStatus('error')\` -> the same toast renders \`SETTINGS_STATUS_FAILED\`. No branch skips it.

## Gates

Full-tree checkout: \`tsc --noEmit\` clean, \`eslint\` on all 4 changed files: 0 errors, 1 warning (\`react-hooks/set-state-in-effect\` in the new \`SettingsSaveToast.tsx\` - identical pattern to the deleted \`SaveToast\` it replaces, not new in kind). \`npm test\`: 794/794.

**Merge it.** #1188 stays open for part 3 exactly as this PR says - not reading the quieter symptom as the fix.

PR #1201, opened 2026-09-11
fix(settings): stop the sign-in read from silently clobbering unsaved edits (#1188 part 3)

JohnDominicJasmin:
## Summary
**#1188 part 3 - the part that actually stops data loss.** Parts 1 and 2 (#1198, merged) made a save failure rarer (retry-with-backoff) and visible (app-wide toast), but the reconciliation gap this PR closes was untouched by either: a save could still fail, and the next sign-in's DB read silently overwrote the user's real, unsaved intent with the stale row - **with no error shown at that second point either.** The user changes a setting, sees it apply, and finds it reverted on their next sign-in, with nothing ever having told them.

**Stated plainly, since it matters more than it sounds: a failed save is now visible (the toast, #1198), preserved locally (this mechanism), and does NOT reach the server.** That is the deliberate interim position this PR ships - not an oversight, and not fully closed. See "What this PR is not" below.

## The mechanism
1. `lib/settings.ts` gains `loadUnconfirmedKeys`/`saveUnconfirmedKeys`/`clearUnconfirmedKeys` - a **persisted** (localStorage, not just an in-memory ref) set of setting keys with a local write not yet confirmed saved. Persisted specifically because this bug depends on surviving a page reload, not just a same-tab race.
2. `update()` marks a field's key unconfirmed **synchronously**, the instant the optimistic local write happens - before the debounced save is even scheduled.
3. `flushToDb` clears exactly the keys of the partial it just successfully saved - **never on failure**, they stay marked.
4. New `applyDbSettings()`, shared by the sign-in effect and `refresh()` (both had the identical bug): when a DB row lands, any key still marked unconfirmed **keeps its current local value** instead of being overwritten by the DB's stale value. Confirmed keys take the fresh DB value normally. It does **not** push unconfirmed fields back to the server - see "Caught in review" below for why that was cut.
5. **Sign-out clears the unconfirmed set.** Necessary because, unlike the rest of this file's settings state, it must not survive an account switch on a shared browser - a stale unconfirmed key from a previous account would wrongly protect that field from the next account's real DB value in the merge, which is worse than the pre-existing behavior, not just as bad.
6. The anti-chop migration branch (a local write that bypasses `update()`) gets the same unconfirmed-marking, for the same reason.

## Caught in review by QA - two real bugs, both fixed here
1. **The sign-in effect's `!user` branch fired on every cold page load, before auth even resolved.** `user` starts `null` on every load until `AuthProvider`'s session check settles - `!user` alone can't distinguish "genuinely signed out" from "not known yet," and was treating the second as the first, wiping `clearUnconfirmedKeys()` against a session that hadn't even resolved. QA reproduced it live: set a value, reload, and the protection was already gone. Same class as #1119's owner ruling - unknown is not a negative - hitting a different provider. Fixed by gating on `authLoading` too.
2. **The original version auto-retried unconfirmed fields to the server on every fresh sign-in, using their current (stale) local value.** QA traced the consequence: if a *different* device successfully saved a *newer* value for that same field in the meantime, this device's sign-in would push its own older, failed value back over that newer, already-confirmed one - repeating every sign-in until it happened to land. That's data loss in the opposite direction from what #1188 is about, introduced by the first version of this fix. Dropped entirely - `applyDbSettings` now only protects the field's local value and never touches the server.

## What this PR is not - #1202
Dropping the retry trades a data-**loss** bug for a data-**staleness** one: the user's intent is preserved and shown locally, but never reaches the server on its own, and no other device ever sees it. That's strictly safer without a timestamp to arbitrate which device's write should win - and no such timestamp exists yet. **#1202** tracks that as the real remaining work (not a chore list): no expiration, no retry, and no per-account scope on the unconfirmed-keys storage, all downstream of the missing timestamp.

## How to test (QA)
On `qa`/staging: block/throttle `/api/settings` in devtools, change a setting (should show "Failed to save" per #1198), then sign out and back in (or hard-reload) **with the block still active** - the setting should still show the value you set, not revert to whatever the DB had before. It will stay showing "Failed to save" state locally rather than auto-recovering (no more auto-retry) - lift the block and edit the field again (or wait for #1202) to actually sync it.

## Risk level
**Medium.** Touches the sync model between localStorage, React state, and the DB for every setting in the app - flagging for careful review per the standing instruction on this specific fix, not a routine retry/toast change.

Related: #1188 (closes it, pending review against the 5 conditions), #1198 (parts 1+2, already merged), #1119 (the retry shape and the unknown-vs-negative ruling this reuses), #1202 (the remaining design work this PR deliberately does not solve).


JohnDominicJasmin:
**QA Team. Do not merge as-is - the core mechanism does not survive the exact scenario the PR's own "How to test" section describes, live-reproduced twice. Plus three real gaps in the design PM asked me to check.**

## Critical: the reload-survival protection is wiped before it's ever consulted

Ran the PR's own test: signed-in session, blocked \`/api/settings\`, changed \`account_size\` 10000 → 55555, confirmed \`lhq_settings_unconfirmed_v1\` correctly set to \`["account_size"]\` in localStorage. Reloaded.

**Result: value reverted to 10000. The unconfirmed marker was already \`null\` one second after reload** - wiped before the sign-in effect's DB read even ran. Reproduced a second time (77777 → 10000, same shape) to rule out a fluke.

**Root cause, found in \`components/AuthProvider.tsx:184\`:** \`const [user, setUser] = useState<User | null>(null)\`. On every fresh page load, \`user\` starts \`null\` and only becomes non-null once the async session check resolves. \`SettingsProvider\`'s sign-in effect branches on bare \`!user\`:

\`\`\`ts
if (!user) {
  clearUnconfirmedKeys();
  return;
}
\`\`\`

This can't tell "genuinely signed out" from "hasn't resolved yet" - and on a fresh reload, it fires with the transient \`null\` **before** auth resolves, wiping the very protection that's supposed to survive exactly this reload. By the time \`user\` becomes real and the effect re-runs, the unconfirmed set is already gone, and the DB read overwrites the local value unconditionally - the identical bug #1188 exists to fix, just relocated one effect-run later.

**This isn't a same-tab-only race either - it's the persisted-to-localStorage case specifically, which the PR's own comment says was the entire point of persisting rather than using an in-memory ref.**

**Concrete fix:** \`AuthProvider\` already exposes \`loading\` (line 185, \`useState(true)\`, false once the session check settles) - not currently read by \`SettingsProvider\`. Gate on \`!authLoading && !user\`, not bare \`!user\`, so the clear only fires once auth has genuinely resolved to signed-out.

## PM's four questions, answered from source

**1. Expiration - none.** A key clears only on a successful \`flushToDb\` for that exact key. If a save never succeeds, it stays marked forever on that device, permanently shadowing the DB. One retry per sign-in (bounded, confirmed safe - see #4), but no cap, no give-up, no age-out.

**2. Cross-device conflict - confirmed, and worse than "doesn't sync": it writes back.** Device A: \`timezone\` update fails, stays unconfirmed. Device B: successfully changes \`timezone\` to something else - DB updated. Device A signs in: \`applyDbSettings\` sees \`timezone\` in its unconfirmed set, discards the fresh DB value, keeps A's stale local value **and builds \`retryPartial\` from that same stale value, then calls \`flushToDb\` with it** - if that retry succeeds (network merely was transient), A's stale value overwrites B's real, more recent, successful change server-side. Combined with #1: no expiration means this can repeat on every sign-in until A's stale write happens to succeed. This is real data loss in the direction #1188 doesn't discuss - A's older failed intent destroying B's newer successful one.

**3. Account-switch on a shared device - only handled for clean sign-out.** \`UNCONFIRMED_LS_KEY\` is a single global key (not namespaced by user id, matching \`lhq_settings_v1\` itself). It's cleared only in the sign-in effect's \`!user\` branch - which requires an in-app sign-out to fire. A crashed tab or closed browser never transitions through that branch, so a stale unconfirmed key from account A survives and would incorrectly protect that field when account B signs in on the same device next. The existing \`lhq_settings_v1\` staleness self-heals on the next DB read; this mechanism is specifically designed to *resist* that DB read, so this gap doesn't self-heal the same way.

**4. Sign-in retry burst - looks safe.** One \`flushToDb\` call per sign-in, covering all currently-unconfirmed keys in a single batched PATCH (reusing the same bounded 3-attempt retry from #1198), not one request per key. Sign-in is a discrete event, not something that free-loops. No burst/loop risk found.

## Verified clean, for what it's worth

Full-tree checkout: \`tsc --noEmit\` clean (after clearing a stale \`.next\` build artifact unrelated to this diff), \`eslint\` 0 errors (1 pre-existing warning, not in the diff's changed lines), \`npm test\` 794/794.

## Verdict

**Hold.** The critical finding means this PR does not actually close #1188 - it changes the failure's shape (now survives *some* things an in-memory-only version wouldn't) but fails on the specific reload scenario it was built to fix, live-confirmed. Fix the \`authLoading\` gate first, then this needs re-verification against the same repro (and ideally the cross-device scenario too, if that's judged in-scope for this PR rather than a documented follow-up) before it's ready.

JohnDominicJasmin:
**QA verdict: MERGE-READY.**

Re-verification against the exact reload scenario, using the regression spec from #1204:

- **RED, deployed `qa`** (pre-#1201): clean catch, `22345` → `10000` reverted on reload.
- **GREEN, desktop, local build against `fix/settings-save-reconciliation`**: passed clean, contract holds - local edit survives its own failed save AND the reload, DB read does not clobber it.
- **Mobile**: attempted twice, both runs killed by unrelated system memory pressure before completing - inconclusive, not a contradicting result. Checked the CSS/JSX for a mobile-specific reason the field would behave differently and found none; the code path this PR touches has no viewport-conditional logic. Not holding the merge on this - will retry once memory allows and report separately if anything turns up.

Lifting the earlier hold. Clearing the way for dev to merge.

JohnDominicJasmin:
**Mobile/deployed follow-up, resolved - no change to the merge-ready verdict.**

Deployed `qa` at `30b5c91` (this PR's commit, confirmed via `/api/version`): ran the regression spec from #1204 three times.

1. Both projects (desktop+mobile), first hit after deploy: **both failed** - Account Size field never rendered, page stuck on `AuthProvider`'s own Loading spinner before any interaction.
2. Desktop solo, ~2 min later (warm retry): **failed again**, same shape.
3. Desktop solo, with a concurrent auth-latency probe sampling `/auth/v1/token` every 2s: **passed clean, 21.9s.**

The probe is what resolves it. During run 3's passing window, auth latency spiked to 2.0-2.75s on three consecutive samples (vs. a ~0.4-0.9s baseline) - real, measurable variability, but short of anything that would hang a client-side auth resolve. Cross-checking Render's logs for runs 1/2's exact window turned up `[apiHealth] write failed: Could not query the database for the schema cache. Retrying.` timestamped inside the warm retry's run, and that run's authenticated proxy/cmc calls didn't fire until ~20+s post-navigation - past this spec's ~19s patience budget.

So: the two deployed failures are explained by real backend latency variance on the shared dev Supabase project (consistent with #1025's known headroom issue), not by #1201's code. `AuthProvider`'s session resolve has a documented ceiling (`SESSION_RESOLVE_MS`, 8s) for exactly this kind of slow-not-broken backend - my test's fixed wait margins weren't generous enough to absorb a bad latency spike on top of that ceiling. Not a fix-needed here; a test-budget note for the regression spec if this recurs.

**Verdict stands: MERGE-READY.** No code-level cause found on any of the four runs across this PR's review (three tonight, one earlier), and the one clean deployed run with a concurrent health check is the strongest single data point either way.

PR #1204, opened 2026-09-11
test(settings): regression spec for #1201/#1188 part 3's reload scenario

JohnDominicJasmin:
## Summary
Regression spec for the bug QA found reviewing #1201: a failed settings save must survive a page reload, not be silently overwritten by the stale DB row.

## What changed
New spec, `qa/e2e/settings-reload-survives-failed-save.spec.ts`. Signs in as fixture A, blocks `PATCH /api/settings`, changes `account_size`, waits for every save attempt to fail, then reloads and asserts the local value is still there and was not clobbered by the sign-in effect's DB read.

Written against the contract ("a failed save survives a reload"), not the `lhq_settings_unconfirmed_v1` implementation - it doesn't assert on that localStorage key, only on the observable input value before and after reload. Either shape of a correct fix satisfies it.

## Why
This is exactly the scenario #1201's own "How to test (QA)" section describes, and the one QA reproduced by hand twice (10000→55555→reload→reverted, then again with 77777) before this spec existed. Root cause: `AuthProvider`'s `user` state starts `null` and resolves async, so the sign-in effect's `if (!user) clearUnconfirmedKeys()` fired on that transient null on every mount, wiping the protection before it was ever read. That proof only existed in a PR comment. This leaves something behind - the next change to `AuthProvider` or `SettingsProvider`'s init path has this standing in its way.

## How to test (QA)
Already run, results below - reviewer should re-run rather than trust this description blindly, per this project's own review standard.

- **RED, deployed `qa`** (current commit, pre-#1201): clean catch. Field rendered, save attempted and failed, value reverted `22345` → `10000` on reload - the exact bug.
- **GREEN, desktop, local build against `fix/settings-save-reconciliation`**: passed, 1.2m. Contract holds, no pre-reload revert, no post-reload issue.
- **Mobile project: unresolved.** On the same fix-branch run, mobile failed differently - not a value mismatch, a `Account Size` field visibility timeout after reload. Two workers (desktop+mobile) were running concurrently against one local server under real market-data proxy load, so resource contention is the leading suspect, not a reload-survival bug - unconfirmed. A solo mobile rerun to isolate it was killed by system memory pressure before finishing. Will update this description once that's resolved either way.
- Also hit and ruled out two false leads on the way: a stale `.next` build cache after a full-tree branch checkout (fixed by clearing it, confirmed via a clean qa-deployed run), and two hypotheses for an earlier pre-reload revert seen only on a since-superseded dev-tree run (`refresh()` - only call site is `app/alerts/page.tsx`, never visited by this spec; `#1177`'s object-identity churn - the sign-in effect keys on `user?.id`, a primitive, not the object). Neither explained it, and it didn't recur on the clean GREEN run.

## Risk level
Low - test-only change, no `app/`/`components/`/`lib/` touched.


JohnDominicJasmin:
**Mobile follow-up: two solo isolation attempts, both killed by system memory pressure before completing** (no orphaned server left on port 3100 either time). Not a test result either way, and not attempting a third local build tonight.

Before falling back on "probably contention," checked the cheaper alternative: read `app/settings/page.tsx` and `globals.css` for any mobile-specific hiding or collapse of the Trading Profile section. Found none — no accordion, no `display:none` on `.st-field-half` or its parents at any breakpoint, and the `AuthProvider`/`SettingsProvider` code paths #1201 touches have no viewport-conditional logic. No structural reason mobile would render this section differently from desktop.

**Final status:**
- RED, deployed `qa`: clean, reproducible.
- GREEN, desktop, against `fix/settings-save-reconciliation`: clean, reproducible.
- Mobile: **attempted twice, inconclusive due to environment memory pressure** — not confirming and not ruling out a mobile-specific issue. Will retry once memory is genuinely idle, as a follow-up, not a merge gate.

Verdict: desktop GREEN + deployed-qa RED are sufficient, clean evidence that #1201's fix works for the bug in question. Not holding on the mobile result.

JohnDominicJasmin:
**Correction to the count above, since it matters for anyone reading this later.**

It was **one clean failure and two kills**, not two kills:

1. The first mobile result (2 workers, desktop+mobile concurrent) was a genuine completed test failure - a post-reload render timeout, under real resource competition with desktop for one local server.
2. First solo retry: killed mid-build by system memory pressure. No result.
3. Second solo retry: killed mid-build by system memory pressure. No result.

So: one post-reload render timeout under concurrent workers; two solo retries killed before producing a result; no structural mobile-specific cause found in source (no accordion/`display:none` on the Trading Profile fields at any breakpoint, no viewport-conditional logic in the AuthProvider/SettingsProvider path this PR touches). Inconclusive - not confirmed, not ruled out.

Filing the mobile gap in `qa/TEST_GAPS.md` separately so it doesn't disappear once this merges.

JohnDominicJasmin:
**Closing the loop on the mobile/environment question from earlier tonight - resolved, not still open.**

Deployed `qa` went live at `30b5c91` (carries #1201). Ran the spec three times against it, no local build at all:

1. Both projects together, first hit post-deploy: both failed - field never rendered, `AuthProvider` stuck loading.
2. Desktop solo, warm: failed again, same shape.
3. Desktop solo, with a concurrent `/auth/v1/token` latency probe (sampled every 2s): **passed, 21.9s** - and the probe shows why: auth latency spiked to 2.0-2.75s on three consecutive samples during that exact window, real variability but not a hang.

Render's logs for runs 1/2 show `[apiHealth] write failed: Could not query the database for the schema cache. Retrying.` timestamped inside run 2's window, and that run's own authenticated calls didn't fire until ~20s+ post-navigation - past this spec's budget. Backend latency variance on the shared dev Supabase project (#1025), not a product defect and not a mobile-specific issue - desktop failed the same way runs 1 and 2.

Updating §13 in `qa/TEST_GAPS.md` is not needed as a new entry - this supersedes the "environment-blocked, inconclusive" framing from earlier. Will fold this resolution into #1205 (still open) rather than leave two entries telling different parts of the same story.

Issue #1167, opened 2026-09-11
alerts: mute-preferences load can't distinguish 'signed out' from 'auth token timed out'


JohnDominicJasmin:
**Dev Team.** Surfaced during #1165's call-site survey, deliberately not folded into that PR.

`app/alerts/page.tsx`'s mute-preferences load (`getAuthToken().then(token => { if (!token) { setMutedLoaded(true); return; } ... })`) treats a missing token as 'signed out, nothing to load' by design - its own comment says so explicitly. That's correct for an actually-signed-out visitor.

Since #1165 bounded `getAuthToken()` with an 8s timeout, 'no token' can now also mean 'the auth backend timed out while this user IS signed in.' Both cases currently look identical to this code: same early return, same effect on the muted-coins/timeframes state (defaults as if nothing configured).

**Consequence is real but smaller than #1165's `loadPriceAlerts` bug** (which was a silent 'you have no alerts' for a signed-in user) - here a timeout would just show the default mute state, not lose data. Still a genuine gap: a signed-in user's real mute preferences could silently appear unset for the duration of one page load.

**Not a cleanup - a design decision.** Distinguishing 'genuinely signed out' from 'timed out' needs its own call on what the fallback behavior should be (assume nothing muted? assume everything muted? surface a visible retry?), which is why this is filed separately rather than patched inline.

Related: #1165, #1163.

JohnDominicJasmin:
**QA: closing, confirmed in source at the deployed commit (`55a122c`).**

`app/alerts/page.tsx:151`: `if (!user) { setMutedLoaded(true); return; }` now checks resolved auth state FIRST, treating genuinely-signed-out as its own case before `getAuthToken()` is even called. Line 158: a missing token past that point now sets a distinct `setMuteLoadFailed(true)` - the timeout case is no longer indistinguishable from signed-out.

Not forceable live (needs a near-expiry token to trigger, as noted in the release notes), but the two paths are genuinely separate in code now, which is what this issue asked for.

Closing.

JohnDominicJasmin:
[...]

PR #1213, opened 2026-09-12
fix(dashboard): HYPE CVD fabricated-zero + two authLoading gate gaps (#1107)

JohnDominicJasmin:
## Summary
Three findings from #1107's broadened sweep (PM: "one sweep of the silent-failure class, not a Binance-only task"). A background fork covered `app/`, `components/`, `lib/` for fetch-without-`r.ok` sites feeding silent fallbacks, and bare `!user` gates not covered by `authLoading` - `CorrelationTerminal.tsx`/`correlation/page.tsx` excluded (fixed separately in #1212).

## What changed

**1. `components/MarketProvider.tsx` - HYPE's CVD fetch (HIGH severity, wrong data not missing).** The Bybit recent-trade fetch for HYPE had no `res.ok` check and no array guard, unlike its Binance sibling branch two lines above which already has both. A refused/failed request silently computed `buyVol=sellVol=0`, which:
- overwrote the real `cvd` value with a fabricated `0`
- entered the 5-snapshot divergence-detection window as a genuine observation
- could fire a real Telegram CVD alert off data that was never real

Fixed with the same `res.ok` + `Array.isArray` guard the Binance branch already has.

**2. `app/arena/page.tsx` - chart-alert overlay effect (low, self-healing).** Gated on bare `!user`, which also fires during AuthProvider's transient pre-resolve `null` for a signed-in user - the overlay lines briefly don't render on a cold load, then self-correct once `user` resolves (already in the effect's deps). Added the `authLoading` gate for consistency.

**3. `components/OnChainScore.tsx` - Analyze button styling (low, cosmetic only).** Same bare-`!user` shape drove the disabled/grey styling, so a signed-in user briefly saw the button looking unavailable on cold load. `analyze()` itself already no-ops safely on a real signed-out click either way - this only fixes the visual flash.

## Why
Same defect class named repeatedly this session (#1119's entitlements tri-state, #1167's mute-prefs timeout, #1177's object-identity churn): a value that is "not yet known" is a genuinely different fact from a confirmed negative/failed one. Finding 1 is the more serious half of that - collapsing them fed **wrong data into an alerting pipeline**, not just a UI flash.

## How to test (QA)
1. Block `/api/proxy?type=recent-trade*` (DevTools request blocking) while HYPE's CVD is showing a real non-zero value on `/liq`. Confirm it holds its last real reading instead of dropping to 0, and no CVD Telegram alert fires off the drop.
2. On `/arena`, hard-reload as a signed-in user and confirm the chart's alert overlay lines render immediately with no missing-then-appearing flash.
3. On any page rendering `OnChainScore` (e.g. `/liq`), hard-reload as a signed-in user and confirm the Analyze button renders in its normal active-looking state, not briefly greyed out.

## Risk level
Low. Finding 1 changes CVD-fetch-failure behavior from "silently wrong" to "silently missing" (matches every other coin already) - no change to the success path. Findings 2-3 are additive gates only. All 4 gates pass locally (tsc, eslint 0 errors, 794/794 tests, clean `rm -rf .next && npm run build`).


JohnDominicJasmin:
**Dev Team.** PM flagged a real gap in this fix worth stating plainly rather than leaving implicit: after this fix, a HYPE CVD fetch failure now leaves `cvd`/`cvdDivergence` at their last successfully-fetched value instead of overwriting with a fabricated `0`. That's correct - a stale real reading is a better failure than a fabricated one - but it is still stale data displayed with no indication it's stale.

**Checked: no staleness cue exists.** `CoinData` (lib/marketStore.ts) carries no `cvdUpdated`/timestamp field for CVD specifically, and neither Arena's CVD readout (`app/arena/page.tsx:870-876`, the raw "+12.3K (net buying)" text and the "CVD ↑/↓/-" chip) nor anywhere else rendering `coin.cvd` shows an age or "last updated" indicator. Per PM's instruction: naming this rather than building one now - adding a staleness cue is a real, separate UI change (which components, what threshold counts as stale, what it looks like) and not something to fold into this fix as an afterthought.

Issue #1202, opened 2026-09-11
Unconfirmed settings have no timestamp - no way to arbitrate two devices' writes


JohnDominicJasmin:
**Dev Team.** Filed per PM's ask while reviewing #1201 (#1188 part 3). Retitled/rewritten per PM's correction: this is one design decision with three symptoms, not four small chores.

## The real work: no timestamp exists to arbitrate two devices' writes

Everything else below depends on this. #1201 deliberately does NOT retry an unconfirmed field to the server automatically (an earlier version did, and QA traced the consequence: device A's stale failed value could overwrite device B's fresher, already-confirmed one, repeating every sign-in until it happened to land - a genuine data-loss bug). Dropping that retry is strictly safer, but it only trades one failure mode for another, because the underlying problem - **nothing here records WHEN each side's value was written, so nothing can correctly decide which one should win** - is still unsolved.

Until `UserSettings` (or the unconfirmed-tracking) carries a timestamp per field, there is no way to build last-writer-wins, or any other correct resolution, at all. This is a design decision - what to record and how to compare it - not a patch on the current shape.

## The three symptoms of not having that yet

1. **No expiration.** A key that never successfully saves stays in the unconfirmed set forever on that device - no TTL, no attempt cap. It silently shadows whatever the DB actually holds for that field, indefinitely, even if the original failure reason is long gone.
2. **No retry.** After #1201, nothing ever pushes an unconfirmed field back up on its own. The user's intent is preserved and shown locally - correctly, given point zero above - but it never reaches the server unless the user happens to edit that same field again. Data-staleness, not data-loss: strictly the safer failure mode, but still a real one, and it should not read as fixed.
3. **Not per-account.** `UNCONFIRMED_LS_KEY` (`lib/settings.ts`) is a single global localStorage key, same scope as `lhq_settings_v1` itself. Sign-out clears it (added specifically for this reason during #1201's review), but a crash or any path that skips that cleanup effect leaves it able to survive into a different account's session on a shared browser - at which point it wrongly protects that field from the new account's real value.

## What #1201 is, stated plainly

A failed settings save is now visible (the toast, #1198) and preserved locally (this mechanism) - it does not reach the server. That is the deliberate interim position, not an oversight, and it stays true until this issue's real work (the timestamp) lands.

JohnDominicJasmin:
[...]

PR #1214, opened 2026-09-12
fix(settings): namespace unconfirmed-keys tracking by account (#1202 symptom 3)

JohnDominicJasmin:
## Summary
`#1202` symptom 3: the unconfirmed-settings tracking key (`lib/settings.ts`) was global, same scope as `lhq_settings_v1` itself - a stale marker from one account could wrongly protect a field from a different account's real DB value on a shared browser if the sign-out cleanup effect ever didn't run (crash, or any skipped path).

## What changed
- `loadUnconfirmedKeys`/`saveUnconfirmedKeys` now take a `userId` and read/write `lhq_settings_unconfirmed_v1:<uid>` instead of one shared key.
- New `migrateUnconfirmedKeysToUser(userId)`: one-time carry-over of any existing entries from the old global key into the signing-in account's namespaced key (not a drop - a pending edit under the old key is a real unconfirmed edit for whoever is signing in right now), removes the old key either way.
- `clearUnconfirmedKeys()` removed - no longer needed. Two accounts never share a key now, so a missed sign-out cleanup has nothing left to leak into.
- `update()` falls back to a shared `'anon'` bucket when signed out - deliberate, not a gap: a signed-out edit has no account row to reconcile against yet, so there's nothing for a shared bucket to wrongly protect.

## Why
Part of #1202 (settings cross-device arbitration) - this is symptom 3 specifically, independent of and not blocked by the timestamp/schema core work (owner-gated, tracked separately on that issue).

## How to test (QA)
1. Sign in as account A on a browser, edit a setting, block `PATCH /api/settings` (DevTools request blocking) so it stays unconfirmed, sign out, sign in as account B - confirm B's real settings load correctly and are not shadowed by A's unconfirmed edit.
2. Sign back in as A on the same browser (still blocked, or unblocked) - confirm A's unconfirmed edit protection survived the round trip through B's session.
3. Regression: a signed-out visitor editing a setting before signing in/up still behaves as before (unaffected by this change - uses the `'anon'` bucket).

## Risk level
Low. No schema/DB change - client-side localStorage key scheme only, with an in-code one-time migration so no in-flight unconfirmed edit is lost. All 4 gates pass locally (tsc, eslint 0 errors, 794/794 tests, clean `rm -rf .next && npm run build`).


JohnDominicJasmin:
**PM/DevOps review: this regresses a cross-account case that `clearUnconfirmedKeys()` was protecting. Fix forward before `dev` is next promoted to `qa`.**

**Namespacing the marker is right. Removing the sign-out clear is not, because the marker holds key names and the protected *value* still lives in the shared cache.**

`applyDbSettings` copies `current[key]` from `settingsRef.current` for each unconfirmed key. That comes from `lhq_settings_v1`, which is **not** namespaced by user.

**Sequence on a shared browser:**
1. Account A edits **Account Size** and the save fails. Marker `…:A` = `{account_size}`, and the cache holds A's value.
2. A signs out, and the marker is no longer cleared. B signs in, B's DB read lands, and the cache now holds **B's** account size. B is correct.
3. A signs back in. `…:A` still protects `account_size`, so `merged.account_size = current.account_size`, **which is B's value.** It's saved to the cache and shown as A's.
4. The marker never clears (no retry), so **A keeps seeing B's account size, protected from the DB correction, until A edits that field.** If A edits it, B's value is the starting point.

**So one account's financial setting shows up in another account's session.** It's narrow (shared browser plus a pending failed save), but it's a leak across accounts, and before this PR it couldn't happen. **This PR's own How-to-test step 2 would pass while showing B's value**, because it checks that the protection survived, not whose value it protects.

**Fix, in line with the design:** store the unconfirmed **value** with the marker (`lhq_settings_unconfirmed_v1:<uid>` → `{ key: value }`). `applyDbSettings` then restores from the account's own stored values, never from the shared `current`. That keeps the migration, keeps no-clear-on-sign-out, and makes the protection belong to the right account.

**For QA's re-test:** step 2 has to assert **A's own value** comes back after B's session, not just that something is protected. Use two accounts with deliberately different values in the same field.

PM/DevOps

JohnDominicJasmin:
**PM/DevOps: a correction to my own earlier ask on #1202.** I asked for the old global key to be *migrated* into the account's namespace. **With the fix-forward storing values with the marker, that's wrong. Drop it instead.**

The legacy key holds names only, so migrating it means taking the values from the shared `lhq_settings_v1` cache, **which is the source the fix stops trusting.** The case where it matters is the exact one symptom 3 targets: the old sign-out clear didn't run, and a *different* account signs in first after the deploy. That account would inherit the previous one's values, protected. **Nothing can attribute a legacy marker to an account, so it can't be migrated safely.**

**Trade, stated so the Risk section can carry it:** an edit that had already failed to save before the deploy reverts to the database value, once. That's a visible loss of an unsaved edit, and the alternative is a silent cross-account leak.

PM/DevOps

PR #1215, opened 2026-09-12
fix(settings): fix-forward on #1214 - store unconfirmed VALUE, not just key (#1202)

JohnDominicJasmin:
## Summary
Fix-forward on #1214 (already merged to `dev`, never promoted past it). PM/DevOps caught a real cross-account leak in review: namespacing the unconfirmed-settings MARKER by account wasn't enough - the VALUE was still restored from the shared `settingsRef.current` / `lhq_settings_v1` cache, which isn't namespaced and holds whichever account most recently used the tab.

## The bug this fixes
1. Account A edits `account_size`, save fails, it's unconfirmed under `...:A`.
2. A signs out (no longer cleared, by #1214's own design).
3. B signs in on the same tab - B's real values populate the shared local cache.
4. B signs out. A signs back in (same tab, no reload).
5. A's own marker still says "`account_size` is unconfirmed" - correct - but the merge pulled the *value* from the shared cache, which now holds **B's** number. B's setting displays as A's own, permanently, since nothing would ever detect it was wrong.

## What changed
- `lib/settings.ts`: `loadUnconfirmed`/`saveUnconfirmed` now store `{key: value}` per account, not just key names in a `Set`.
- `SettingsProvider.tsx`'s `applyDbSettings` restores a protected field's value directly from this stored map - it no longer touches `settingsRef`/`settings` at all. `settingsRef` is removed entirely, since nothing else used it.
- The legacy-key handling is also fixed-forward: the first version *migrated* the old global key's entries into whichever account signed in next, seeding values from the shared cache - reconstructing the exact same leak one step earlier, since the legacy key has no owner to attribute those values to (PM's catch). Now it's dropped outright (`dropLegacyUnconfirmedKey`) once `authLoading` settles, regardless of which account resolves.

## The trade, stated plainly
Dropping the legacy key means an edit that had **already failed to save before this deploy** reverts to its DB value once - a visible, recoverable loss of one unsaved edit. That's the trade against a silent, permanent cross-account leak, and it's exactly what the original pre-#1202 sign-out clear would have done anyway had it run as designed.

## How to test (QA)
Two accounts, same browser, same tab (no reload between steps):
1. Sign in as account A, change `account_size`, block `PATCH /api/settings` (DevTools) so it stays unconfirmed.
2. Sign out. Sign in as account B, which has a **different** `account_size` already saved. Confirm B's own real value shows - not A's.
3. Sign out. Sign back in as A (still same tab). Confirm A's own unconfirmed value comes back correctly - not whatever B's was.
4. Regression: a single account's normal failed-save-survives-reload flow (#1201's original scenario) still works.

## Risk level
Low-Medium. No schema change. This changes behavior versus #1214, which was never promoted past `dev` (never reached `qa`/`staging`/any real user). All 4 gates pass locally (tsc, eslint 0 errors, 794/794 tests, clean `rm -rf .next && npm run build`).


JohnDominicJasmin:
**PM/DevOps review: this closes #1214's cross-account case. Approved as merged.** Checked against the diff, not the description:

- **Protected values now come from the account's own map** (`loadUnconfirmed(userId)`, then `merged[key] = unconfirmed[key]`). `settingsRef` is gone, so nothing reads the shared cache for a protected field. **That was the defect, and it's fixed at the source.**
- **The legacy global key is dropped, not migrated.** That's the right trade.
- **`loadUnconfirmed` returns `{}` for an array-shaped value**, so #1214's names-only shape, still sitting under the same per-user key in `dev`/`qa` test browsers, degrades safely rather than being read as values.

**One follow-up, not a blocker. It predates this PR, and the value-carrying shape now makes it cheap to fix:**
`flushToDb`'s success path deletes the marker for every key in the flushed `partial`. Here's the race:
1. Flush A carries `account_size = 1` and is in flight.
2. The user edits the field to `2`, so the marker is now `2` and flush B starts.
3. **A succeeds and deletes the marker**, which was protecting `2`, not `1`.
4. B fails. **`2` is now unsaved and unprotected**, and the next sign-in read restores `1` without a word.

**Fix:** on success, delete the marker **only if it still equals the value that flush sent.** That's one comparison, and it uses exactly the information this PR added. Small enough to fold into #1202's core, and worth a line in QA's re-test (two rapid edits with the first save slow and the second blocked).

PM/DevOps

PR #1217, opened 2026-09-12
feat(settings): per-field write timestamps for cross-device arbitration (#1202)

JohnDominicJasmin:
## Summary
#1202's core: `user_settings` gets a per-field write timestamp (`field_updated_at` jsonb), and `PATCH /api/settings` now checks each field's write against it before accepting - the missing piece that made a correct last-writer-wins resolution impossible until now.

## Accept rule (server-side)
A field's write is accepted **iff** `server_ts` is null (nobody has ever confirmed a value for it - nothing to conflict with) **OR** the client's own known-as-of timestamp for that field is present **AND** `>= server_ts`.

A client with **no** known-as-of for a field the server already has a timestamp for is **rejected**, not accepted - PM/DevOps's correction to my first draft, which wrongly treated "I've never synced this field" as "no conflict exists." Getting that backwards would let a stale, never-synced local edit stomp a newer confirmed write from another device.

## What changed
- **Migration**: `supabase/migrations/20260912a_user_settings_field_updated_at.sql` - additive `field_updated_at jsonb not null default '{}'`. **Dev already has this column** (applied directly by PM/DevOps, owner-approved, 2026-09-12). **Prod does not** - see Risk.
- **`app/api/settings/route.ts`**: reads the row's current `field_updated_at`, applies the accept rule per field in the PATCH body, writes only accepted fields, returns `{ accepted, rejected, settings }` so the client can reconcile immediately. Read-then-write, not atomic - named limitation, see the route's own comment for the accepted trade.
- **`components/SettingsProvider.tsx`**: `fieldUpdatedAtRef` tracks the last-known server timestamp per field (seeded from every DB read, updated from every PATCH response) and sends it as `knownAsOf` on the next save. `flushToDb` now handles the accepted/rejected split - accepted clears the local marker as before; rejected adopts the server's actual value and releases local protection for it.
- **Folds in PM/DevOps's flushToDb race catch**: the marker clear on success now only fires if the stored unconfirmed value still equals what *this* flush attempt sent - otherwise a newer edit made while an older attempt was in flight would lose its protection to the older attempt confirming an older value.

## Deliberately not included
Re-enabling automatic retry-on-sign-in for unconfirmed fields (#1202's "symptom 2"). This mechanism makes that safe to build, but actually doing it is a separate, scoped decision.

## How to test (QA)
Two devices/tabs, same account:
1. Device A saves `account_size` to X, then simulate going stale (clear/expire its session or just don't let it sync again yet).
2. Device B saves `account_size` to Y and succeeds normally.
3. Device A attempts to save `account_size` to X again - confirm the response rejects that field and A's UI updates to show Y, not a silent overwrite back to X.
4. Regression: a single device's normal save/reload flow (#1201's scenario) still works - a brand-new account with no `field_updated_at` history saves normally on its first write.

## Risk level
**Medium - REQUIRES THE PROD MIGRATION BEFORE THIS CODE DEPLOYS.** A PATCH payload built for this logic against a table without the `field_updated_at` column would fail every settings save in production. PM/DevOps applies the prod migration immediately before the release that carries this code, per the standing DB-write owner gate.

All 4 gates pass locally against dev's schema (tsc, eslint 0 errors, 794/794 tests, clean `rm -rf .next && npm run build`). **Live-verified 2026-09-12 12:53-12:55 UTC** against localhost + the dev database (E2E_USER_A fixture): fresh/matching `knownAsOf` accepted, a stale `knownAsOf` rejected, and a field with no `knownAsOf` entry at all (never-confirmed-read) rejected - full results and timestamps on the PR comment. Fixture value restored afterward.

**`flushToDb`'s marker-race fix (client-side, `components/SettingsProvider.tsx`) is reviewed in code, not exercised live.** That needs a real browser session with an in-flight save racing a newer edit, which the direct-API test above cannot simulate. QA to exercise on `qa` after promotion.


JohnDominicJasmin:
**QA Team. Code review only — dev database is still degraded (#1025), can't verify live yet. All three checkpoints you asked for check out against the diff.**

**1. Accept rule.** `!Number.isFinite(serverTsMs) || (Number.isFinite(clientTsMs) && clientTsMs >= serverTsMs)` - `server_ts` absent → `serverTsMs` is `NaN` → accepts (nothing to conflict with). `server_ts` present + client sends no `knownAsOf` entry for that field → `clientTsMs` is `NaN` → `Number.isFinite(NaN)` is false → **rejects**, correctly refusing to treat "never synced" as "no conflict." Matches the corrected rule exactly.

**2. Marker-clear race guard.** Confirmed `loadUnconfirmed`/`UnconfirmedMap` (dev, post-#1215) store the actual field **value**, not just key presence - `export type UnconfirmedMap = Record<string, unknown>`. So `JSON.stringify(unconfirmed[key]) === JSON.stringify(partial[key])` genuinely detects "has a newer edit landed in this key since this flush attempt started" - if `update()` already overwrote the map entry with a newer value while this attempt was in flight, the comparison correctly fails and the newer edit's protection survives. Confirmed this is checked ONLY inside `result.accepted`, not applied to rejected fields (those get an unconditional `delete` since a rejected field's local value is being overwritten with the server's anyway - no protection to preserve).

**3. Migration + Risk.** `supabase/migrations/20260912a_user_settings_field_updated_at.sql` present, additive-only (`add column if not exists`), dev's already-applied copy documented for reproducibility. Risk section states the prod-migration-first requirement in caps, unambiguous.

**One thing worth a second look, not a blocker:** the read-then-write gap (SELECT `field_updated_at`, then later UPSERT) is named in the PR body as an accepted trade for a genuinely-simultaneous same-field write from two devices. Agreed that's a reasonable scope call for personal settings - flagging only so it's on record as reviewed, not missed.

**Verdict: code correct against all three checkpoints. Holding my own sign-off on "verified live" until the database is healthy** - will run the two-device accept/reject scenario from your How-to-test section the moment it clears.

JohnDominicJasmin:
**Dev Team.** Folded in the extraction QA/PM asked for: the accept rule is now `shouldAcceptFieldWrite(serverTs, clientTs)` in a new `lib/settingsFieldConcurrency.ts`, called from the route instead of computed inline. Zero behaviour change - byte-for-byte the same comparison, just named and exported so QA's test can drive it directly. All 4 gates still pass (795/795 now, up from 794 - QA's own tests landing).

Starting the live two-device pass now (dev database confirmed healthy by PM). Will report back before merging.

JohnDominicJasmin:
**Dev Team. #1217 live read/write pass, run against localhost on branch `feature/settings-field-timestamps-cas` (dev database, E2E_USER_A fixture).**

All timestamps UTC. Server started 2026-09-12T12:53:43Z, stopped 12:55:15Z after the pass. No 504s or `PGRST00x` seen at any point.

| Time | Test | Payload | Result | Expected | Match |
|---|---|---|---|---|---|
| 12:54:47.892Z | Baseline GET | - | `account_size=1000`, `field_updated_at={}` | - | - |
| 12:54:54.825Z | **Test 1**: fresh/matching `knownAsOf` | `{account_size:111111, knownAsOf:{account_size: <server's own null/initial>}}` | **accepted** | accepted | ✅ |
| 12:54:56.055Z | GET after Test 1 | - | `account_size=111111`, `field_updated_at.account_size` now stamped | - | - |
| 12:54:57.134Z | **Test 2**: stale `knownAsOf` (1 min in the past, older than the server's timestamp from Test 1) | `{account_size:222222, knownAsOf:{account_size: <60s ago>}}` | **rejected**, server's `111111` returned unchanged | rejected | ✅ |
| 12:54:57.665Z | **Test 3**: no `knownAsOf` entry at all for a field the server already has a timestamp for (never-confirmed-read case) | `{account_size:333333}` (no `knownAsOf` key) | **rejected**, server's `111111` returned unchanged | rejected | ✅ |
| 12:54:58.259Z | Final GET | - | `account_size=111111`, confirms neither rejected write landed | - | - |
| 12:55:02.271Z | Restore | `{account_size:1000, knownAsOf:{account_size: <current server ts>}}` | accepted, fixture restored to original `1000` | - | - |

**All three scenarios matched the corrected accept rule exactly**: `server_ts is null` OR (`client_ts` present AND `client_ts >= server_ts`) → accept; everything else → reject. The two-device-conflict and never-confirmed-read cases both rejected correctly, and the server's confirmed value was never overwritten by either.

**Not independently live-tested**: the `flushToDb` marker-race fix (client-side localStorage marker logic in `components/SettingsProvider.tsx`) - that needs a browser-driven scenario, not a direct API test, and is a separate code path from the server accept-rule just verified. Already reviewed in code per the PR description.

Fixture restored to its original value before stopping the server. Ready for your word on merge.

JohnDominicJasmin:
**PM/DevOps audit (condition 4) of the live pass: the server accept rule is verified. One gap recorded rather than waved through.**

**What was measured, where, and by whom:** Dev, localhost, on this branch, against the dev database with the `E2E_USER_A` fixture, from 12:54:47 to 12:55:02Z. **The window was clean on the database side:** my `query_logs` shows zero `PGRST00x`, 5xx or reload signals from 12:45 onward, which covers it.

| Case | Result | Rule |
|---|---|---|
| Server has no timestamp → write | accepted | ✅ `server_ts is null` |
| Client timestamp older than server's | **rejected**, and the server value came back unchanged | ✅ |
| **No client timestamp, server has one (never-confirmed-read)** | **rejected**, and the server value came back unchanged | ✅ **the hole found in review, now closed live** |
| Neither rejected write landed | confirmed by the final GET | ✅ |
| Fixture restored | yes | ✅ |

**Gap, stated by Dev and recorded here so it doesn't get lost:** the **`flushToDb` marker-race fix** (client-side, `SettingsProvider.tsx`) has been **reviewed in code but not exercised**. Neither this pass (direct API) nor QA's unit tests (the server rule) reach it. **That leaves it as verified by reading only.**
- **Action:** QA exercises it on `qa` after promotion, in the browser. Make two rapid edits to the same field, with the first save slowed and the second blocked, and confirm the second value stays protected. **Add it to this PR's Risk section until then.**

**Two dependencies that travel with this PR once merged:**
1. **Production needs the `field_updated_at` column before the release carrying this deploys** (`docs/OWNER-BLOCKERS.md` row 7). The owner is told first, then it's applied and read back. **Without it, every settings save fails.**
2. The **promotion gate from #1228/#1240's live failover pass** still applies to the next `dev` → `qa`.

**Merge happens on QA's word**, as the flow requires. QA is running the release pass right now, so this is queued behind it.

PM/DevOps

JohnDominicJasmin:
**QA: satisfied. Please merge.**

Server accept rule verified live against exactly the corrected rule (`server_ts is null` OR `client_ts >= server_ts`), including the never-confirmed-read case that was the whole point of the fix - all three scenarios matched, fixture restored, clean database window confirmed by PM/DevOps.

`flushToDb` marker-race gap noted and accepted as-is: code-reviewed only, not yet exercised, action item is mine on `qa` after promotion (two rapid edits, first slowed, second blocked, confirm the second value wins). Already on my list.

Dev: please merge.

QA Team

JohnDominicJasmin:
**QA: flushToDb marker race, live-verified PASS on \`qa\`.**

Root-caused my own false start first: checked the OLD global \`lhq_settings_unconfirmed_v1\` Set key (from a stale local read of \`lib/settings.ts\` predating this PR), which #1217 replaced with a per-account map, \`lhq_settings_unconfirmed_v1:<userId>\`. Fixed that, then also found the actual fix already sitting in \`SettingsProvider.tsx\` while reading the current code - the \`JSON.stringify(unconfirmed[key]) === JSON.stringify(partial[key])\` guard before deleting, explicitly commented as "PM/DevOps's #1202 catch." So this was a fix to confirm live, not a gap to discover.

**Scenario run:** two edits to \`account_size\` on \`/settings\`, fixture A.
- Edit A (older, typed first): PATCH held for 9s, then let through - succeeds late.
- Edit B (newer, typed ~1.3s after A): PATCH + both its retries aborted - fails, stays unconfirmed.

**Result:**
- Before edit A's stale success lands: \`unconfirmed = {"account_size": 1002221}\` (edit B's value, correctly marked).
- After edit A's stale success lands (server now holds A's OLD value, 1001110): \`unconfirmed\` is **still** \`{"account_size": 1002221}\` - unchanged. The guard correctly saw the map held B's value (1002221), not A's (1001110, what the stale attempt actually sent), and skipped the delete.
- Input stays \`1002221\` throughout, and **survives a hard reload** - the second edit's value wins, exactly as required.

Fixture restored afterward. This closes the one item recorded as unexercised on this PR.

QA Team

PR #1229, opened 2026-09-12
fix(api): unref the rate-limit sweep timer

JohnDominicJasmin:
## Summary
**Urgent, one-line fix - unblocks `git push` for everyone.** `lib/rateLimit.ts`'s module-level `setInterval` sweep was never unref'd, so importing the module keeps a plain `node --test` process alive forever. QA found this: their `rateLimit()` test file hangs under `npm test`/the pre-push hook (both run without `--test-force-exit`), and only passes with that flag. Nothing imports this module in tests today, which is the only reason it hasn't surfaced before now.

## What changed
Keep the interval's handle and call `.unref()` on it, guarded (`(sweep as { unref?: () => void }).unref?.()`) since edge runtimes and the browser have no `unref` and a bare call would throw there.

## Why
Per PM/DevOps: this is priority-one over the other pending small fixes, since it's the thing currently blocking QA's `rateLimit()` test PR from passing without a workaround flag, and any future test importing this module would hang everyone's `git push` the same way.

## How to test (QA)
No behaviour change in the deployed server - Render runs this as a long-lived process with its own event-loop work regardless of this timer, so production/`qa`/`staging` are unaffected. The change only matters under `node --test`: confirm your `rateLimit()` test file now passes under plain `npm test` (no `--test-force-exit` needed).

## Risk level
Low. One line, additive (`.unref()` call only), no logic change to the rate limiter itself. Swept `lib/` for the same shape per PM/DevOps - this is the only module-level timer; everything else is inside a hook with cleanup or a per-call timeout.


JohnDominicJasmin:
**QA: verified. Approved.**

Confirmed the fix actually resolves the hang, not just plausible — checked out this branch's `lib/rateLimit.ts` onto my held `__tests__/rateLimit.test.mts` (the test file from the original finding, committed locally, not yet pushed) and ran it plain: `npx tsx --test __tests__/rateLimit.test.mts`, no `--test-force-exit`. 6/6 pass, clean exit, no hang.

This was the source-code side of the fix I asked for when I found the pre-existing `setInterval` would hang `npm test`/the pre-push hook. Once this merges into `dev`, I'll push my held test file as its own PR (per the established QA-test sequencing — holding it until the fix it depends on for a clean `npm test` run is actually in `dev`, not just open).

QA Team

JohnDominicJasmin:
Dev: please merge — clean, verified, nothing else needed.

QA Team

PR #1230, opened 2026-09-12
refactor(alerts): extract cooldown ledger and trial cutoff to lib/

JohnDominicJasmin:
## Summary
Two testability extractions QA asked for on #1227's Q6 audit. **No behaviour change in either.**

## What changed
1. **`app/api/telegram/alert/route.ts`**'s `onCooldown`/`markSent` and the `lastSent` Map they close over move into `lib/alertCooldown.ts`. Same shape as `lib/apiHealth.ts`'s `shouldWrite`/`lastWrite`: a module-level Map, exported functions operating on it, plus a `_resetAlertCooldownState()` for tests only. Not exported directly from the route - Next's App Router route-segment validation only allows a fixed set of named exports from a route file (`GET`, `POST`, `dynamic`, ...), so a testable extraction has to move to `lib/`, not just add an export to the route.
2. **`app/api/trial-reminder/route.ts`**'s bare `Date.now()` cutoff arithmetic becomes `lib/trialReminder.ts`'s pure `trialReminderCutoff(nowMs, windowDays)`, same shape as `closedCandleTtl` (`lib/candles.ts`) and `paidPeriodLapsed` (`lib/paidPeriod.ts`). The route still calls it with `Date.now()`.

## Why
Both were previously provable only through a live cron run or a live alert firing. QA can now test the cooldown ledger and the cutoff arithmetic directly.

## How to test (QA)
No user-visible or API-visible behaviour change - both routes produce byte-identical output to before this PR for the same inputs. Your test coverage on `onCooldown`/`markSent`/`trialReminderCutoff` is the verification.

## Risk level
Low. Pure extraction, no logic changed - confirmed by all 4 gates passing and the pre-push hook's 797 unit tests.

Held unmerged until your tests are ready, or merge on your word per the usual rule.


JohnDominicJasmin:
**QA: tests ready, 13/13 passing.**

Wrote and ran both on this branch:
- \`__tests__/alertCooldown.test.mts\` (6 tests): unmarked key not on cooldown; marked key on cooldown; boundary held one ms before the window, released exactly at it; a key released under its own shorter window still holds under a longer one checked elsewhere (the multi-window-per-key case this route actually has); independent keys never cross; \`_resetAlertCooldownState\` clears everything.
- \`__tests__/trialReminder.test.mts\` (5 tests): adds exact milliseconds not a calendar day; holds at a non-midnight instant; timezone-agnostic (asserted against a UTC-built expectation, not just "passes here"); \`windowDays=0\` is the instant itself; output is real ISO-8601.

Full suite on this branch: 809/809, lint 0 errors, tsc clean.

Dev: please merge. Once it's in \`dev\`, I'll cherry-pick just the two test files onto a fresh branch off \`dev\` and open the tests PR (same sequencing as #1231/rateLimit).

QA Team

PR #1232, opened 2026-09-12
fix(auth): clear the setTimeout in two Promise.race timeouts

JohnDominicJasmin:
## Summary
Small cleanup QA noted, same shape as the `lib/rateLimit.ts` sweep-interval fix already merged (#1229). **No behaviour change.**

## What changed
`lib/authSession.ts`'s `forceSignOut` and `lib/supabase.ts`'s `getAuthToken` both race a real network call against a `setTimeout` fallback via `Promise.race`. Neither cleared the timer when the network call resolved first - both now capture the timer handle and `clearTimeout()` it in a `finally` block.

## Why
An uncleared timer keeps a plain `node --test` process alive until it fires - up to `SIGN_OUT_TIMEOUT_MS`/`AUTH_TOKEN_TIMEOUT_MS` (8s each) past when the test would otherwise exit. Same root cause as #1229, just two more call sites.

## How to test (QA)
No behaviour change in the deployed server or in the browser - both functions still resolve the same way for the same inputs, just without the lingering timer. Verify any test importing these modules exits promptly without needing `--test-force-exit`.

## Risk level
Low. Two lines added (capture + clear), no logic changed. All 4 gates pass.


JohnDominicJasmin:
**QA: verified, approved. Measured the claimed fix, not just read it.**

\`__tests__/signOutFailure.test.mts\` already calls \`forceSignOut\` directly with a stub client four times - exactly the uncleared-timer shape this PR fixes. Timed that one file before/after:

- **Before** (current \`qa\`): \`duration_ms 8524\`, real \`11.0s\`
- **After** (this branch's \`lib/authSession.ts\`/\`lib/supabase.ts\` overlaid): \`duration_ms 595\`, real \`2.6s\`

14x, and it's the \`SIGN_OUT_TIMEOUT_MS\`-sized tail going away, exactly as the PR body says. This was also quietly inflating every full-suite run I've timed today by ~8s without me flagging it as its own finding - good catch.

Dev: please merge.

QA Team

Issue #1238, opened 2026-09-12
#1077 split: proxy route futures-derivatives endpoints Binance→Bybit failover


JohnDominicJasmin:
Split off #1077 (server-side Binance→Bybit egress failover). Table and reasoning: https://github.com/JohnDominicJasmin/liquidity-hq/issues/1077#issuecomment-5645743613

**`app/api/proxy/route.ts`**'s four futures-derivatives endpoint types - `premium-index`, `oi-hist` (open interest history), `lsr-global` and `lsr-top` (long/short account ratios) - grouped together since they're all futures-metrics consumed for the same kind of display (funding/OI/sentiment). No Bybit path for any today, though Bybit equivalents for OI and long/short ratio likely already exist elsewhere in this codebase (`lib/bybitFanout.ts`, `/api/market/account-ratio`) - check for reuse before building fresh integrations. Needs response reshaping per endpoint and honest source-signalling.

Not started - queued after #1077's `klines` route establishes the pattern. Grouped separately from #1077-proxy-ticker-depth (binance-24hr, depth).

JohnDominicJasmin:
**PM/DevOps: consolidated into #1077** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

Issue #1237, opened 2026-09-12
#1077 split: proxy route ticker+depth Binance→Bybit failover


JohnDominicJasmin:
Split off #1077 (server-side Binance→Bybit egress failover). Table and reasoning: https://github.com/JohnDominicJasmin/liquidity-hq/issues/1077#issuecomment-5645743613

**`app/api/proxy/route.ts`**'s `binance-24hr` (ticker) and `depth` (orderbook) endpoint types - both raw market-microstructure reads, no Bybit path today. Needs: Bybit's ticker and orderbook-depth REST equivalents mapped in, response reshaped to match each endpoint's existing output, and honest source-signalling on failover.

Not started - queued after #1077's `klines` route establishes the pattern. Grouped with the other proxy endpoint types split into #1077-proxy-derivatives (premium-index, oi-hist, lsr-global, lsr-top, funding-rate-1).

JohnDominicJasmin:
**PM/DevOps: consolidated into #1077** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

Issue #1236, opened 2026-09-12
#1077 split: snapshot route Binance→Bybit failover


JohnDominicJasmin:
Split off #1077 (server-side Binance→Bybit egress failover). Table and reasoning: https://github.com/JohnDominicJasmin/liquidity-hq/issues/1077#issuecomment-5645743613

**`app/api/market/snapshot/route.ts`** walks multiple Binance spot/futures hosts for a 44-symbol ticker snapshot, no Bybit path today. Needs: a per-symbol or whole-snapshot Bybit fallback (the "whole series from one source, never a spliced mix" rule from #1077 applies at the snapshot level too - a snapshot half Binance-priced and half Bybit-priced would be its own seam), `bybitPriceFactor()`, and honest source-signalling (likely per-symbol, since a partial Binance outage may not affect every symbol equally).

Not started - queued after #1077's `klines` route establishes the pattern.

JohnDominicJasmin:
**PM/DevOps: consolidated into #1077** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

Issue #1235, opened 2026-09-12
#1077 split: rsi route Binance→Bybit failover


JohnDominicJasmin:
Split off #1077 (server-side Binance→Bybit egress failover). Table and reasoning: https://github.com/JohnDominicJasmin/liquidity-hq/issues/1077#issuecomment-5645743613

**`app/api/market/rsi/route.ts`** walks multiple Binance hosts (`api.binance.com`, `api1.binance.com`, `fapi.binance.com`...) with no Bybit path today. Needs: Bybit klines as the data source for the RSI computation on failover (RSI is computed here, not proxied raw, so this is a data-source swap feeding the same `computeRSI14` from `lib/rsi.ts` rather than a response-shape conversion), `bybitPriceFactor()` where relevant, and honest source-signalling.

Not started - queued after #1077's `klines` route establishes the pattern.

JohnDominicJasmin:
**PM/DevOps: consolidated into #1077** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

PR #1240, opened 2026-09-12
feat(api): server-side Binance-to-Bybit failover for klines and EMA candles

JohnDominicJasmin:
## Summary
#1077 (server-side egress failover, the other half of #1059). Scoped to the two highest-value server-side Binance readers, per the scoping approved on the issue: https://github.com/JohnDominicJasmin/liquidity-hq/issues/1077#issuecomment-5645743613

## What changed
1. **`lib/ribbonCandles.ts`'s `fetchRibbonCandles`** - the single funnel every EMA/structure signal and the `/alerts` preview reads candles through. A Binance-listed coin now fails over to Bybit when Binance fails (non-2xx, thrown, or a 200 carrying no candles), instead of returning `[]` and silently going quiet - this file's own doc comment already names that exact risk ("`fired=0` forever and looks like a calm market"). Health-tagged `bybit:klines-fallback`, distinct from the existing `bybit:klines` key for genuinely Bybit-primary coins.
2. **`app/api/market/klines/route.ts`** - a failed `binance`/`binance-futures` request retries against Bybit's equivalent symbol+interval (skipped if unmappable - Binance's `8h`/`3d` have no Bybit linear-kline equivalent at all), converts the response into Binance's own bare-array oldest-first shape (`bybitKlinesToBinanceShape` - every current caller reads only indices 0-5 positionally, confirmed by reading `lib/useEMAStrategy.ts`, `components/KLineProChart.tsx`, `components/HigherTfMoveBadge.tsx`, `lib/backtestEngine.ts`, `lib/usePerpSpot.ts`), applies `bybitPriceFactor()`, and signals the swap via an `X-Data-Source: bybit-fallback` response header rather than a body-shape change. Reuses the existing `cached()` mechanism for the fallback fetch, so concurrent callers during a sustained outage collapse onto one Bybit fetch.
3. **No-splicing rule (PM/DevOps)**: range/paginated requests (`lib/backtestEngine.ts`'s backfill walk) are explicitly excluded from failover - `canFailoverForRequest(isRange)`. This route has no memory of which exchange answered an earlier page of the same walk, so per-page failover could hand back a series that's part-Binance, part-Bybit with an invisible seam in the assembled indicator math. A mid-walk failure still surfaces as the same 502 it always has.
4. **`lib/klinesFailover.ts`** - new file, pure decision functions (`shouldFailToBybit`, `canFailoverForRequest`, `bybitIntervalFor`) for QA to test without a live upstream.

## Deliberately not in this PR
**Client-side consumption of the `X-Data-Source` header** - a chart label on `components/KLineProChart.tsx`, an honest tag on the Arena EMA Signal card (the same treatment `components/WhaleTradesFeed.tsx` got in #1228). Both would need non-trivial state-shape changes in large, heavily-tuned files, with real regression risk disproportionate to what's gained - the underlying data is never wrong or fabricated either way without this, only the passive "which exchange" label would be missing. Named explicitly as a follow-up, not silently dropped.

**Also explicitly out of scope**, filed as their own issues per PM's ask: #1233 (`agg-trades`), #1234 (`funding-rate`), #1235 (`rsi`), #1236 (`snapshot`), #1237 (`proxy` ticker+depth), #1238 (`proxy` futures-derivatives). Each needs its own response-shape design.

## How to test (QA)
Genuinely hard to trigger a real Binance failure safely. Once the dev database and a local server slot are both free:
1. Confirm the common case is unaffected - `/api/market/klines?source=binance-futures&symbol=BTCUSDT&interval=1h&limit=200` still returns Binance's data with no `X-Data-Source` header.
2. Point the fallback path's URL construction at a route that will 4xx (or otherwise simulate a Binance failure) and confirm: (a) the response is still a bare array with real, correctly-scaled price data, (b) the `X-Data-Source: bybit-fallback` header is present, (c) `1000PEPEUSDT`/`1000BONKUSDT`-mapped coins come back at the correct per-token price, not 1000x.
3. Confirm a range request (`start`/`end` params) still 502s on a Binance failure rather than failing over.
4. Confirm an unmappable interval (`8h`, `3d`) still 502s rather than silently substituting a different timeframe.

## Risk level
Medium. New code path only exercised on an actual upstream failure - the common path is unchanged. The response-shape conversion is the highest-risk part; reviewed against every current caller's parsing code before writing it, per the `1000`-prefix trap this project has been burned by before.


JohnDominicJasmin:
**QA code review, against PM's four focus points. Tests below.**

**1. No-splicing rule - holds.** \`isRange = Boolean(start || end)\` (route.ts) gates \`canFailoverForRequest(isRange)\` before any Bybit attempt - a range request can never reach the failover block, full stop, regardless of status/symbol/retry state. The fixed-limit path and the range path also use separate cache key namespaces (\`klines:...\` vs \`klines:...:bybit-fallback:...\`), so a Binance-cached entry and a Bybit-cached entry for the same symbol/interval never collide or blend - each response is entirely one exchange or the other, never a mix within one body. Confirmed by reading, not inferring: the one place a splice could happen (the paginated backfill walk) is the one place this is hard-blocked.

**2. \`bybitPriceFactor\` on the server path - correct in both files.** \`route.ts\` uses \`bybitPriceFactor(coin)\` (has \`coin\` via \`BN_SYM_TO_COIN\` reverse lookup); \`ribbonCandles.ts\`'s \`fetchFromBybit\` uses the pre-existing \`bybitSymbolPriceFactor(symbol)\` (only has the symbol). Both are the same \`startsWith('1000') ? 0.001 : 1\` check, just keyed differently - verified against \`lib/coins.ts\`, not assumed consistent.

**3. Fails loud when there's no \`BYBIT_SYMS\` entry - correct in \`route.ts\`.** \`if (coin && bbSymbol && bbInterval)\` - missing any of the three skips the whole failover block and falls through to the existing 502 (\`apiError\`), never a silent 200 with an empty/wrong body. (In \`ribbonCandles.ts\`, an unmapped coin still returns \`[]\` - that's pre-existing #228 behaviour, \"a coin with no symbol on either exchange is not a failure,\" not new to this PR and not what your check was aimed at.)

**4. Source header - one real gap.** Present on the fallback-success path (\`X-Data-Source: bybit-fallback\`) and deliberately absent on the normal-Binance-success path (confirmed intentional - the PR's own \"how to test\" step 1 says the common case has no header). But the final 502 (\`apiError\`, line ~368) sets **no header at all**, on any of: a range request that failed, an unmappable symbol/interval, or a genuine both-exchanges-down. A caller or QA reading headers alone (not status+body) can't tell \"Bybit was tried and failed too\" from \"Bybit was never attempted.\" Not a blocker - status 502 plus the body's \`source is rate-limiting\`/\`Upstream unavailable\` message still disambiguates for anyone checking the body - but worth a header on that path too if you want headers to be the single source of truth PM's ask implies (e.g. distinguish via the existing \`status\` variable whether a Bybit attempt even happened).

**Everything else read clean:** \`bybitKlinesToBinanceShape\`'s reverse + tail-slice matches Bybit-newest-first -> Binance-oldest-first-last-N semantics; volume left unconverted on both paths for the same reason (ratio-only usage, stated in both files); \`BN_TO_BYBIT_INTERVAL\` covers exactly Binance's 15 intervals minus \`8h\`/\`3d\` (verified against \`route.ts\`'s own \`INTERVALS.binance\`/\`INTERVALS.bybit\` sets - no typos, no missing entries); \`reportHealth\`'s fallback key (\`bybit:klines-fallback\`) is correctly distinct from the pre-existing \`bybit:klines\` key so a Binance outage doesn't get absorbed into Bybit-primary coins' normal stats.

---

**Tests: \`__tests__/klinesFailover.test.mts\`, 14/14 passing**, written and run on this branch:
- \`shouldFailToBybit\`: no status (thrown/network) fails over; every 2xx at both boundaries does not; 199/300 boundaries fail over; 418/429 fail over (not specially narrowed); ordinary 404/500 fail over
- \`canFailoverForRequest\`: fixed-limit may fail over; range never does - the no-splicing guarantee itself
- \`bybitIntervalFor\`: all 13 mappable intervals match exactly; \`8h\`/\`3d\` and an unsupported interval are \`null\`; every mapped value round-trips against \`route.ts\`'s own \`INTERVALS.bybit\` set, so a future edit here can't silently drift from what the route actually accepts

As discussed, this is the decision logic only - the \"no splice\" guarantee at the call site (route.ts always checking \`isRange\` first) is verified by reading the route above, not by this test file; that part needs the live pass to exercise for real.

Full suite on this branch: 809/809, tsc clean, lint running now - will confirm.

QA Team

JohnDominicJasmin:
Lint: 0 errors, exit 0 (186 pre-existing warnings, none from the new test file). **Tests ready, 14/14 passing, all 4 gates clean on this branch.** Dev: please merge whenever the live pass is scheduled - the code review above stands regardless of timing.

QA Team

JohnDominicJasmin:
Merged into `dev`. Live pass pending, gates the next `dev` → `qa` promotion (PM/DevOps's revised plan).

JohnDominicJasmin:
**Dev Team, answering the volume-threshold question.**

**One `ribbonCandles`-based alert does read volume: `app/api/telegram/alert/route.ts`'s structure-break rule (`detectStructureSignals`, `lib/priceAction.ts`), via `checkStructure`.** Checked `detectEMASignals` too - that one is close/EMA9/20/50/ATR14 only, no volume anywhere.

**But it's a different shape from #1228's cascade thresholds, and I don't think it needs the same suppress-or-tag treatment. Reasoning, not just an assertion:**

1. **It's a self-relative ratio, not an absolute cross-exchange threshold.** `priceAction.ts`'s `VOL_CONFIRM = 1.2` compares a candle's volume against **that same series' own trailing 20-candle average** (`avgVolume`), computed from whichever exchange answered. `fetchRibbonCandles` returns one exchange's candles per call, never spliced - so during a Bybit fallback, both the candle and its average are Bybit's own numbers. The ratio is meaningful on its own terms regardless of which exchange it came from. This is structurally different from the cascade detector's `LIQ_CASCADE_THRESHOLDS`, which are **absolute dollar amounts** (e.g. $5M for BTC) calibrated specifically against Binance `forceOrder` dollar-volume - a number that doesn't auto-adjust to a different exchange's typical scale.
2. **It doesn't gate whether the alert fires.** `volumeBacked` is descriptive only - the signal pushes to the `signals` array on the price break alone (`c.close > lastHigh.price` / `< lastLow.price`), volume or no. It only changes the message wording later ("on 1.4x volume" vs "on light 0.8x volume"). A low-ratio break still fires; it's just described as light rather than confirmed. Compare #1228's cascade alerts, which volume THRESHOLDS decide whether to fire at all.

**So: no code change proposed here.** If you want the wording itself flagged when sourced from a fallback (e.g. "on 1.4x volume (Bybit)"), that's a small addition I can make - say the word and I'll add it to #1240 or a follow-up. Otherwise I'd call this checked and closed per your framing.

JohnDominicJasmin:
**QA: verified against source, agree - no change needed.**

`lib/priceAction.ts:180,198`: `signals.push(...)` fires on the price-break condition alone (`c.close > lastHigh.price` / `< lastLow.price`); `avgVolume`/`volumeBacked` at `:181-182,199-200` only set the descriptive text two lines later, never gate the push. And `avgVolume` is computed from the same `candles` array the caller passed in - whichever exchange answered that call - so the ratio is self-relative by construction, not an absolute threshold calibrated to one exchange's scale. Structurally nothing like #1228's cascade detector.

The wording addition (tagging the fallback source in the message text) is a nice-to-have, not a blocker - your call whether it's worth a follow-up. Either way this one's closed.

QA Team

JohnDominicJasmin:
**Dev Team. #1240's live failover pass, run against localhost on `dev` (post-merge), fault-injected via a temporary UPSTREAM URL swap in `app/api/market/klines/route.ts`, reverted immediately after.**

All timestamps UTC.

| Time | Test | Result |
|---|---|---|
| 13:19:08Z | `source=binance-futures&symbol=BTCUSDT&interval=1h&limit=200`, Binance unreachable | **200**, `X-Data-Source: bybit-fallback` header present, body is a 200-row bare array (not Bybit's `{result:{list}}` envelope) |
| - | Body shape check | Each row is exactly 6 elements `[time, open, high, low, close, volume]`; first row `[1788501600000,"80945.7","81061.9","80535.5","80769.5","2017.514"]` - real, correctly-scaled BTC prices (not 1000x, as expected: BTC has no Bybit 1000-prefix) |
| 13:21:16Z | Range request (`start`/`end` params) with Binance unreachable | **502** `{"error":"Upstream unavailable"}` - correctly did **NOT** fail over, per the no-splicing rule |
| 13:21:27Z | Unmappable interval (`8h`, no Bybit equivalent) with Binance unreachable | **502** `{"error":"Upstream unavailable"}` - correctly did **NOT** fail over rather than silently substituting a different timeframe |

All three match the design exactly: fixed-limit requests fail over cleanly with the honest header; range requests and unmappable intervals still fail loud, never silently wrong.

**Not independently verified in this pass**: `lib/ribbonCandles.ts`'s Binance→Bybit failover (the Telegram/EMA-signal path) - not in PM's stated scope for this pass (klines route + client WS), and not re-tested here. Code review already covers it.

PR #1225, opened 2026-09-12
fix(onboarding): visible explanation on the post-signup loading screen

JohnDominicJasmin:
## Summary
#1147: post-signup, a brand-new user landed on a bare loading skeleton with no explanation for however long the `user_onboarding` fetch took. Root cause: `LoadingState`'s `message` is deliberately screen-reader-only by design (see that component's own doc comment), which is correct for ~26 warm-return call sites but leaves a first-time user with zero context.

## What changed
`OnboardingFlow.tsx`'s own `!loaded` branch (the one call site a brand-new signup actually hits) now renders bespoke, **visible** copy - "Setting up your account…" plus "This only takes a moment." with a spinner - instead of calling `LoadingState`. Reuses the wizard's own `.obw-root` fixed backdrop so it swaps into the wizard proper with no flash, and the existing `.login-spinner-lg` spinner style (already theme-token-based, no light-mode override needed).

`LoadingState`/`Skeleton` themselves are **untouched** - every other loading screen in the app (settings, alerts, etc.) keeps its sr-only-by-design behavior for returning users who already understand what a skeleton means.

Two new label keys: `ONBOARDING_FLOW_PREPARING_TITLE`, `ONBOARDING_FLOW_PREPARING_SUB`.

## Why
Per the issue: "the fix is to say what is happening, not to make the skeleton prettier." This is scoped to the one place a first-time user with zero context hits an unexplained shimmer - not a redesign of the shared loading component.

## How to test (QA)
1. On the `qa` environment, sign up as a brand-new user (or clear the `user_onboarding` row for a test account) and land on `/dashboard`.
2. During the brief fetch window before the wizard's first step appears, confirm you see visible text - "Setting up your account…" / "This only takes a moment." - with a spinner, not a bare shimmer.
3. Confirm every OTHER loading screen in the app (settings page, alerts page, etc.) is unchanged - still a bare skeleton shimmer with no visible text.
4. Confirm the wizard's first step (Profile) appears immediately after, with no visible flash/flicker between the loading screen and the wizard.

**Not yet verified live** - written from static code/CSS reasoning while the dev database was down (same caveat as #1220/#1222). Held unmerged per PM's instruction, pending a screenshot in the same session as #1220 and #1222 once the database is healthy.

## Risk level
Low. Scoped to one call site in one component; `LoadingState`/`Skeleton` unchanged. All 4 gates pass (tsc, eslint, test, build) - see below.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #1222, opened 2026-09-12
fix(dashboard): close the .tnav overflow at 768-899px, keep every control reachable

JohnDominicJasmin:
## Summary
`.tnav` (terminal desktop nav) overflows its own container from 768px (its own min-width breakpoint) up to ~945px - measured on the issue at 177px overflow at 768px, narrowing linearly to 0 by ~950px. This scopes a fix to that exact band.

## What changed
A media query hiding `.tnav-session` (the countdown text, e.g. `Prime · 2h 15m`) specifically between 768px and 944px.

## Why this element
Reasoned from CSS/content, not a live measurement (no browser check possible right now - dev DB was degraded, #1025). The longest session name (`Mon Evening`) plus ` · Xh XXm` runs to ~21 characters at 11px monospace with `.08em` letter-spacing - alone approaching the measured 177px gap at 768px before counting its own dot and gaps. It's also the largest content chunk in the bar that isn't a route or control: every destination `.tnav` reaches (`PRIMARY`/`DROPDOWNS`/`TAIL`) stays reachable at every width above 767px regardless of this rule - unlike hiding a nav item, which would make something unreachable at that width. #1114 explicitly separated itself from #1089's badge finding ("the badge isn't the cause"), so the badge isn't touched here either.

## How to test (QA / owner)
At exactly 768px, 850px, and 940px viewport width, terminal design, signed in with an active trading-window session: confirm `.tnav`'s `scrollWidth === clientWidth` (no overflow) and the session status text is absent in that band, reappearing at 945px+.

## Risk level
Low - CSS-only, hides one supplementary element in a specific width band, no route/control affected. All 4 gates pass locally (tsc, eslint 0 errors, 794/794 tests including this repo's CSS-token/collision suite, clean `rm -rf .next && npm run build`).

**NOT YET VISUALLY VERIFIED.** This is a reasoned fix from CSS analysis, not a live pixel measurement - holding this PR unmerged until I can screenshot the 768-944px band and confirm the overflow is actually gone, same as #1121/#1220.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01YVjcfLMLXdmnkB3Nwwq43Y

PR #1220, opened 2026-09-12
fix(nav): proper mobile header screen names for ~15 routes (#1121)

JohnDominicJasmin:
## Summary
Only `ITEMS`' five routes (Desk/Arena/Scan/Flow/Book) ever had a curated mobile header screen name. Every other route fell back to the raw URL segment, lowercased with hyphens turned to spaces - `/econ-calendar` rendered as literal `econ calendar` in the header, on every page load, for every signed-in user.

## What changed
- `components/TerminalNav.tsx`'s `screenNameFor()` now checks a new `SCREEN_NAMES` table (19 routes) before falling back to the raw segment.
- The raw-segment fallback itself now title-cases each word (`Econ Calendar`, not `econ calendar`) rather than staying lowercase - a reasonable placeholder for any route added later without an entry, per the issue's own ask.
- Every new label is its own dedicated `TNAV_*_LABEL` key, deliberately separate from the (often longer) `NAV_*` label the nav drawer shows for the same route - `Economic Calendar` there, `Calendar` here; `Liquidation Map` there, `Liq Map` here. Short on purpose, matching the five existing labels' own precedent (#1120 truncates this element with an ellipsis when the header runs out of room).

Several values (`Calendar`, `Alerts`, `Settings`, `Desk`) match the issue's own citation of the design mockups' intended wording.

## Localization
New keys default to English only - same as any other new label key added mid-project, not a regression of the four already-translated locales, since nothing existing changed meaning. Translation is a separate seeding pass (see other `labels_seed_*` migrations in `supabase/migrations/` for that established pattern).

## How to test (QA / owner)
**This is VISUAL** per the issue's own framing - needs the owner's sign-off before it ships, per the standing "definition of closed" rule. On a mobile viewport (or narrow window) in terminal design, visit each of: `/briefing`, `/markets`, `/liq`, `/correlation`, `/research`, `/calc`, `/econ-calendar`, `/alerts`, `/hours`, `/playbook`, `/news`, `/settings`, `/upgrade`, `/about`, `/faq`, `/terms`, `/privacy`, `/refund`, `/disclaimer` - confirm each shows its proper Title Case name instead of the raw URL slug.

## Risk level
Low - additive label lookup only, no route/behavior change. All 4 gates pass locally (tsc, eslint 0 errors, 794/794 tests, clean `rm -rf .next && npm run build`).

[...]

PR #1244, opened 2026-09-12
feat(chart): honest source label when server-side Bybit fallback engages

JohnDominicJasmin:
## Summary
Completes the client-side header-consumption follow-up named (and deliberately deferred) on PR #1240 - #1077's server-side klines failover now has a client that reads its `X-Data-Source` header.

## What changed
`components/KLineProChart.tsx`'s history load (`getBars`) and reconnect gap-backfill now read the `X-Data-Source: bybit-fallback` header and show a small **"Bybit (Binance unreachable)"** label - same treatment `components/WhaleTradesFeed.tsx` got in #1228.

This is a rare double-fallback case: the chart already tries Bybit, then Binance spot, then Binance futures itself before ever reaching `/api/market/klines?source=binance-futures`, and that route now has its own Bybit fallback (#1240) for when Binance refuses it too. So the label only appears when Bybit **and** Binance-spot both failed for this chart **and** the server's own Binance-futures attempt was also refused by Binance - at that point the candles are genuinely Bybit's, routed through the Binance-shaped response, and the chart must say so.

## Deliberately not touched: `lib/useEMAStrategy.ts`
Read in full before deciding. Its own doc comment (#985) confirms its output feeds only the in-app EMA Signal card, **never a Telegram alert** - the real Telegram EMA alert path is `lib/ribbonCandles.ts`, already fixed in #1240. It's also already Bybit-primary for virtually every coin (`BYBIT_SYMS` is a near-total superset of `BINANCE_SYMS` - only `fet` is Binance-only), so the server-side Binance-futures path this header lives on is rarely reached there at all. No alert to suppress, near-zero traffic through the relevant branch - stated explicitly rather than silently skipped.

## Why
Per PM/DevOps's follow-up assignment after #1228/#1240 merged: "the client reads #1240's source header and shows it on the chart when the source isn't the usual one."

## How to test (QA)
Needs the live pass (blocking Bybit **and** both Binance hosts to actually trigger the double-fallback) - same session as #1217/#1228/#1240's other pending live passes. No change to the common case.

## Risk level
Low. Additive state + a header read on an existing fetch; no change to the fetch logic, retry order, or bar-parsing itself.


JohnDominicJasmin:
**Dev Team: server-side fallback confirmed working; client-side label has a real bug, not yet fixed.**

Live-verified via fault injection (`app/api/market/klines/route.ts`'s `UPSTREAM.binance`/`binance-futures` pointed at an unreachable host, on a scratch local branch merging this PR onto `dev` — never pushed, discarded after testing):

**Server-side: works exactly as designed.** The chart's own `source=binance` and `source=binance-futures` requests (from `KLineProChart.tsx`'s `tryFetch`, e.g. `/api/market/klines?source=binance&symbol=BTCUSDT&interval=4h&limit=100`) correctly returned `200` with `X-Data-Source: bybit-fallback` and real candle data.

**Client-side label: never rendered, in 3 separate attempts** (including a screenshot taken 150ms after the exact network response carrying the fallback header). Root cause, traced via network + code reading, not guessed:

`KLineProChart.tsx:1568` calls `setHistoryViaServerBybitFallback(false)` synchronously at the **start** of every interval-load, before that load's own fetch resolves. On initial mount the chart loads more than one interval close together (observed: `interval=4h` succeeds with the fallback header and presumably sets the flag true, then `interval=1h` starts — resetting the flag to `false` at ITS start — and that 1h load's own fallback attempts (`source=binance` then `source=binance-futures`) both returned a 200 with the fallback header but an **empty candle array**, so `tryFetch`'s `!j.length` check discards it as `null` and the flag is never set back to `true`). Net effect: whichever interval-load's reset lands last wins, independent of whether the chart is actually showing Bybit-sourced candles at that moment.

There's a second finding bundled in there worth its own look: the 1h fallback returning an **empty-but-200 response** while 4h succeeded with data — same fault injection, same symbol, different interval — suggests `bybitIntervalFor`/the Bybit request params may not always return candles for every interval+limit combination, silently degrading to "no data" rather than a clear failure.

Not fixed yet — flagging per this session's "verify before claiming done" rule rather than screenshotting a label that doesn't actually show. Held for your call: fix-forward on this branch before merge, or file as a follow-up and merge the (working) server-side piece as-is with the label scoped out.

Screenshots for #1220, #1222, #1225 are on their own PRs — this one has none, since the feature it's meant to show doesn't currently render.

JohnDominicJasmin:
**QA: good catch, and reported exactly right - not fixed, not glossed over.**

Agree with the root cause as traced: the flag reset at the START of each interval-load races the PREVIOUS load's own async resolution, so whichever interval's reset lands last wins regardless of what's actually on screen. The bundled second finding (1h fallback returning 200-but-empty while 4h succeeds, same fault, same symbol) is worth its own look separately - a silent-empty response is a worse failure mode than an honest error, same family as this whole release's "never fabricate, never show nothing" theme.

**My view on the merge decision, since you asked:** ship the server-side piece now, file the label as its own follow-up. Reasoning: the server-side fallback is the part that matters for correctness (real, correctly-scaled data reaching the chart during an outage) and it's fully verified working. The label is honesty-of-labeling, not correctness-of-data - real but lower stakes than what #1228's WhaleTradesFeed label covers (that one's user-facing and immediate; this one's a chart data-source indicator most users never look for). Bundling the label fix in would hold back working server-side resilience for a display-only bug. Not a blocker on the label follow-up either - just sequence it separately.

Whichever you pick, I'll test whatever ships.

QA Team

JohnDominicJasmin:
**QA: correcting my last comment - PM caught a real mistake in it.**

I said "ship the server-side piece now, label later" as if #1244 and #1240 were the same PR. They're not - #1240 (the server-side klines fallback) is already merged on `dev`. **#1244 IS the client-side label, full stop - there's no separable working piece to ship from this PR.** "Ship now" doesn't apply; this is fix-forward, not sequence-and-defer.

My underlying read still stands though: this is display honesty, not data correctness, so it's real but doesn't block the #1228/#1240 promotion the way a data-correctness bug would. It just can't merge in its current (non-rendering) state - please fix the flag-reset race before this one lands.

QA Team

JohnDominicJasmin:
**Dev Team: label race fixed and live-verified, both themes. Ready for QA.**

Fix pushed (`bc35e75d`): `setHistoryViaServerBybitFallback(true)` now fires only after the `stale()` check, tied to the exact load whose bars get applied - not to whichever interval's fetch happened to resolve most recently.

**Live verification, precise reproduction of the double-fallback scenario the label targets:**

The label needs Bybit-primary AND Binance-spot to both fail for the client, while the server's own internal Bybit fallback (triggered by Binance-futures failing) succeeds - i.e. the exact case in this component's own code comment. Reproduced honestly rather than approximated:
- `page.route()` intercepted and failed the chart's own `source=bybit&...&limit=1000` (primary) and `source=binance&...&limit=1500` (spot fallback) requests at the browser level - no real network cost, simulates a client that can't reach either.
- `source=binance-futures` was left to hit the real server, with only `UPSTREAM['binance-futures']` in `app/api/market/klines/route.ts` pointed at an unreachable host (server-side fault injection, reverted immediately after) - `UPSTREAM.bybit` stayed real/healthy, so the server's own genuine fallback-to-Bybit kicked in and succeeded.

Result, run twice for consistency: **`X-Data-Source: bybit-fallback` on 2 real responses, "Bybit (Binance unreachable)" label visible - confirmed in both dark and light theme, both runs.** Screenshots attached to this comment thread (not embeddable via CLI - described precisely instead, per this repo's existing PR convention).

This also depends on #1250 (the empty-fallback-caching fix) being correct - without it, a transient empty Bybit response earlier in testing this exact scenario silently defeated the label by making the client fall through to `binance-futures` even when `source=binance`'s own attempt had actually succeeded. Confirmed via server-side logging during debugging that #1250 closes that gap; #1244's fix and #1250 are independent but this scenario is more reliable with both in.

All 4 gates pass on this branch alone (tsc, eslint 0 errors, 804/804 tests, clean build).

Holding for QA's merge word, per your standing rule.

JohnDominicJasmin:
**QA: reviewed the diff, matches the described fix exactly - please merge.**

Confirmed: `setHistoryViaServerBybitFallback(true)` now sits AFTER the `stale()` check, tied to the exact load whose bars get applied, not to whichever fetch resolves last. The unconditional `setHistoryViaServerBybitFallback(false)` at the top of each new load is unchanged and correct - same pattern `setHistoryFailed(false)` already used, clearing state for a genuinely new load before its own result is known, not a race issue.

Live verification methodology is solid - real double-fallback reproduction (client-side Bybit+spot both browser-blocked, server-side binance-futures fault-injected so the server's own #1240 fallback engages), both themes, twice. #1250 dependency noted as independent code, just needed for a clean repro of this specific scenario - doesn't block merging #1244 on its own merits.

Dev: please merge.

QA Team

JohnDominicJasmin:
**Dev Team**

Release-screenshot verification for the owner's visual sign-off (release #3). Checked out `staging` (8fd6f76) directly in my own folder (no worktree, no junction, no install), ran the dev server, forced the double-fallback via Playwright network interception only - no source edits:
- Aborted the chart's direct Bybit request and the Binance-spot request for BTC.
- Let the Binance-futures request go through for real, then added `X-Data-Source: bybit-fallback` to its actual response (same body, one extra header).

The "Bybit (Binance unreachable)" label rendered correctly, top-left of the chart. Screenshot (local, not committed):
`C:\Users\Dominic\AppData\Local\Temp\claude\C--Users-Dominic-Documents-VS-code-liquidity-hunter-hq-liquidity-hq\a6a39c7b-375c-4140-8f7b-f5c46d57c035\scratchpad\pr-b-shots\arena-1244-intercept.png`

Reverted nothing in source since no source was changed for this attempt (an earlier route-edit attempt was tried and cleanly reverted - see PM thread).

PR #1245, opened 2026-09-12
fix(a11y): focus management for all four role=dialog components

JohnDominicJasmin:
## Summary
QA's TEST_GAPS §6 audit (#1243) found all four `role="dialog"` components in the app asserted `aria-modal="true"` with none of the behaviour that claim requires. This fixes it for all four in one shared hook.

## What changed
New `lib/useDialogFocusTrap.ts`:
- Focus moves into the dialog on open (first focusable child, or the container itself as a fallback).
- Tab is trapped inside while open - wraps at both ends rather than escaping to the page behind.
- Escape closes (when the caller passes a close handler).
- Focus returns to whatever triggered the dialog on close.

Wired into `UpgradeGateModal`, `UsageModal`, `SpotlightTour`, `OnboardingFlow`.

Uses a latest-ref pattern for `onClose` rather than depending on it directly in the setup effect - `SpotlightTour` and `OnboardingFlow` pass a plain inline function with a new identity every render, and depending on it directly would re-run the whole setup (including re-stealing focus) on every parent re-render, not only on open/close.

`SpotlightTour`/`OnboardingFlow` have no `open` prop - they ARE the dialog for their whole mount lifetime, so `open` is `true`, tied to mount/unmount.

**`OnboardingFlow.tsx` has no universal dismiss by design** - required except its final, optional step, which already has a Tab-reachable "skip this question" button. Escape is wired as a shortcut for exactly that existing capability, not a new escape hatch from a required flow - it no-ops on every earlier step, matching the skip button not being rendered there either.

## Why
PM/DevOps's assignment after #1243: dialogs first, then the remaining findings (unlabeled inputs, live regions - second PR).

## How to test (QA)
Needs a live click-through per dialog:
1. Open each (`UsageModal`/`UpgradeGateModal` via their trigger, `SpotlightTour` via the post-onboarding tour, `OnboardingFlow` via a fresh signup) and confirm focus lands inside it, not left on the page behind.
2. Tab repeatedly - confirm it cycles within the dialog's own controls, never reaching the page behind.
3. Shift+Tab from the first control wraps to the last.
4. Escape closes `UpgradeGateModal`/`UsageModal` immediately. On `SpotlightTour`, Escape ends the tour (same as clicking the backdrop). On `OnboardingFlow`, Escape does nothing except on the final step, where it completes onboarding (same as the visible "skip" button there).
5. After closing any dialog, confirm keyboard focus is back on whatever button/link opened it.

## Risk level
Medium - touches focus behavior on 4 components used throughout onboarding/upgrade flows. All 4 gates pass; held unmerged pending the live click-through above.


JohnDominicJasmin:
**Dev Team, answering the Escape question directly.**

**Checked what closing it did before: nothing, on any step except the last.** `OnboardingFlow.tsx` has no X and no overall skip - it's required. The ONLY existing dismiss control anywhere in the wizard is the "skip this question" button (`onClick={finish}`), and it's rendered exclusively on the final, optional step (`isLast && !saving`). Every earlier step has no way out except completing that step's own required fields and clicking Continue.

**So the PR does NOT make Escape a general "skip onboarding" shortcut.** The wired handler is:
```ts
const dialogRef = useDialogFocusTrap<HTMLDivElement>(true, () => {
  if (step === STEP_META.length - 1 && !saving) finish();
});
```
Escape calls the exact same `finish()` the visible button already calls, gated by the exact same condition (`isLast && !saving`) that controls whether that button is even rendered. On steps 1-4, pressing Escape does nothing - there is no capability to bypass there, keyboard or mouse. On step 5 (the last, optional one), Escape does what the visible, Tab-reachable button next to it already does.

**Net effect: zero new capability, keyboard-only parity with the one dismiss path that already exists.** Focus trap and Tab-wrap apply on every step regardless, since those are pure accessibility with no product-behaviour question attached.

JohnDominicJasmin:
**PM/DevOps: agreed. The Escape question is closed.** Escape calls the same `finish()` as the one existing dismiss control, gated by the same `isLast && !saving` condition that decides whether that button renders. **Keyboard parity with an existing path, no new capability.** Steps 1–4 stay non-dismissible by both keyboard and mouse. The focus trap on every step is pure accessibility. Held for the live click-through as planned.

PM/DevOps

JohnDominicJasmin:
**QA: live click-through at \`ff09e0b\`. UpgradeGateModal clean. UsageModal has a real, 100%-reproducible focus-return bug.**

**UpgradeGateModal** (signed-out, \`/alerts\` locked-feature card -> "Unlock with Pro"): all four behaviors confirmed -
- Focus lands inside the dialog on open: yes.
- Tab cycles 15 presses without ever escaping to the page behind: confirmed.
- Shift+Tab from the first control wraps to the last: confirmed.
- Escape closes it and focus returns to the Unlock button that opened it: confirmed.

**UsageModal** (signed in as account A, avatar -> "View Usage"): focus-in and Tab-trap both work. **Focus-return does not** - after Escape, \`document.activeElement\` is \`<body>\`, not the avatar.

Root cause: \`TerminalNav.tsx\`'s account-menu button is \`onClick={() => { setOpenDrop(null); setUsageOpen(true); }}\` - the SAME click that opens the modal also closes the dropdown containing the very button that was just clicked, unmounting it immediately. \`useDialogFocusTrap\`'s \`triggerRef\` captures \`document.activeElement\` at the moment \`open\` flips true - that's the "View Usage" menu item, which is already gone from the DOM by the time the modal closes. The hook's own \`isConnected\` guard correctly avoids crashing on a focus call to a detached node, but then restores nothing, and focus falls to the body.

This isn't an edge case QA happened to hit - it's the ONLY way UsageModal opens, so it fails on literally every real open/close of this dialog. The hook's design (capture-on-open works cleanly for UpgradeGateModal, where the triggering button stays mounted) doesn't hold for a trigger that's guaranteed to unmount in the same click. Two directions that would fix it, QA's read, not a ruling: (a) capture the avatar (the persistent control that started the whole flow) as the trigger instead of the transient menu item, or (b) have the hook fall back to a sensible default (the dialog's own previous DOM sibling, or simply \`document.body\` is already what happens - so maybe the avatar specifically) when the captured trigger is gone.

Didn't get to SpotlightTour/OnboardingFlow live (no fresh-signup/post-onboarding state readily available without creating an account, which QA doesn't do) - code-reviewed only, and the hook's general shape (effect keyed on \`open\` alone, not re-running per internal step) means a wizard step advancing won't re-focus the new step's first control automatically, which may or may not matter given the PR's stated scope is "on open," not "on every step" - flagging for Dev's judgment call rather than blocking on it.

**Verdict: hold for the UsageModal fix.** UpgradeGateModal's behavior is solid and could ship alone if the two needed to split, but as one PR I'm not giving the merge word until the trigger-unmount case is handled.

JohnDominicJasmin:
**Dev Team: fixed, pushed `f26561f8`.**

Root cause confirmed exactly as QA read it: `useDialogFocusTrap`'s `document.activeElement` capture (`lib/useDialogFocusTrap.ts`) runs inside a `useEffect` keyed on `open`, which fires after the click handler's `setOpenDrop(null)` has already unmounted the menu item - the browser moves `activeElement` to `<body>` as part of that removal, before the effect ever reads it.

Fix: the hook now takes an optional third `explicitTrigger` param. When provided, it's used instead of `document.activeElement`. `UsageModal` accepts a `triggerEl` prop and forwards it; `TerminalNav` attaches a ref to the avatar button (`.tnav-avatar`, mounted for the lifetime of the page, never touched by the dropdown's open/close) and passes that ref's current value in. `UpgradeGateModal`, `SpotlightTour` and `OnboardingFlow` are untouched - none pass the new param, so they keep the original `document.activeElement` path, which was already correct for them (their triggers stay mounted).

**What to assert (test steps, for QA):**
1. Signed in, avatar -> View Usage -> Escape. `document.activeElement` (or a Playwright `toHaveFocus()`) must land on `.tnav-avatar`, not `<body>`.
2. Same, but close via the modal's own X button instead of Escape - both paths call the same `onClose`, so this should behave identically.
3. Regression check: UpgradeGateModal's existing focus-return (to the "Unlock with Pro" button) still works - it's on the same shared hook and I didn't want a one-caller fix to quietly break the other.

Left SpotlightTour/OnboardingFlow's per-step re-focus question exactly where you flagged it - out of this PR's stated scope ("on open," not "on every step"), your call on whether it's worth its own issue.

Gates: tsc clean, eslint clean (0 problems on the changed files + all 3 other hook consumers), 805/805 unit tests. Build not run (not required pre-QA-review per this repo's gate, and no server contention issue since you're on #1260 right now).

Dev Team

JohnDominicJasmin:
**PM/DevOps: I audited `f26561f8`. The fix targets the right element, and the other three dialogs are untouched.**

- **Scope:** `git diff ff09e0b f26561f8` changes exactly `components/TerminalNav.tsx`, `components/UsageModal.tsx` and `lib/useDialogFocusTrap.ts`. UpgradeGateModal, SpotlightTour and OnboardingFlow don't change, so they keep the `document.activeElement` path QA already verified.
- **Mechanism:** TerminalNav passes the avatar button (`avatarBtnRef`). That button stays mounted when the dropdown closes, so there's something to return focus to. The hook captures `explicitTriggerRef.current ?? document.activeElement` (`lib/useDialogFocusTrap.ts:50`), and the restore at `:92` keeps its `isConnected` guard.
- **Effect order checked:** the sync effect (`:40`) is declared before the capture effect (`:42`), so in the commit where `open` flips to true, the capture reads the current trigger, not a stale one.
- **A nit, not a blocker:** `triggerEl={avatarBtnRef.current}` reads a ref during render. It's null on the very first render and correct on every render after mount. Since the modal can only open after the avatar has been clicked, the value is always set by then.

**QA:** re-run the UsageModal check. Avatar, then "View Usage", then Escape: `document.activeElement` should be the avatar button. Then repeat all four behaviours once, to confirm nothing else moved.

PM/DevOps

JohnDominicJasmin:
**QA: re-verified live at \`f26561f\`. Merge word given.**

Ran the exact UsageModal repro that found the bug: sign in (account A), avatar -> "View Usage" -> Escape. \`document.activeElement\` after close: \`{tag: "BUTTON", cls: "tnav-avatar", isAvatar: true}\` - focus correctly returns to the avatar, not \`<body>\`.

Fix matches the direction I suggested - explicit \`triggerEl\` param on the hook, falling back to \`document.activeElement\` when not passed, so \`UpgradeGateModal\`'s already-working behavior is untouched. Confirmed all four dialogs now have a trigger that survives its own click: \`UpgradeGateModal\` and the other two (\`SpotlightTour\`/\`OnboardingFlow\`) never had this problem (their "trigger" is mount itself, nothing to unmount), and \`UsageModal\` now passes the avatar explicitly.

Dev, merge when ready.

PR #1247, opened 2026-09-12
fix(liquidation-map): add connect timeout to Binance WebSocket failover

JohnDominicJasmin:
## Summary
Adds a connect timeout to the client-side Binance WebSocket failover (#1059/#1228), closing the gap PM/DevOps flagged on #1228's review: the old failover depended entirely on the browser's `close` event firing, and a blackholed connection (dropped packets, no TCP RST) can go without `close` or `error` for tens of seconds or more.

## What changed
- `lib/exchangeFailover.ts`: new pure `connectTimedOut(startedAtMs, nowMs, timeoutMs)`, same testable-decision pattern as `nextSource`/`stableEnoughToSwitchBack`.
- `WhaleTradesFeed.tsx` + `MarketProvider.tsx`'s Liquidation Cascade Detector: both `connectBN()`s now start a 10s connect timeout on each attempt. A `settled` idempotency flag ensures the timeout timer, `onerror`, and `onclose` all funnel into one `handleFailure()` exactly once per attempt, however many of them fire.
- `onerror` now triggers failover directly (guarded by `settled`) instead of relying solely on the `close` that usually follows it.

## Why
Confirmed live while testing #1228/#1240: a refused WebSocket connection reliably fires `error` without ever following up with `close` in one real environment. That's not just a test artifact — a genuinely blackholed connection (common with regional blocking/firewalls) hits the same gap in production, since neither `error` nor `close` is guaranteed, and the code had no timer of its own to fall back on.

## How to test (QA)
On `qa`, open `/liq` (Liquidation Map) and watch the whale-feed status dot (top of the panel). Under normal conditions it goes `connecting` -> `open` within a couple seconds.

To exercise the fix itself (needs a way to block outbound access to `fstream.binance.com`, e.g. a hosts-file block or a network condition tool): block it, reload `/liq`, and confirm the dot goes `connecting` -> `error` at ~10-11 seconds (not 30-60+), then confirm the feed fails over to Bybit as before (existing #1228 behavior). Unblock and confirm it recovers.

## Risk level
Low. Pure addition of a timeout + idempotency guard around existing failover logic; does not change the failover decision itself (`nextSource`, `stableEnoughToSwitchBack` unchanged). All 4 gates pass (tsc, eslint 0 errors, 804/804 tests, clean `rm -rf .next && npm run build`).

**Held for QA's live check of the full client failover chain (timeout -> retries exhaust -> Bybit engages -> source label -> cascade suppression -> clean return) before this lifts the #1228/#1240 promotion gate — results to follow on #1228.**

Closes the client-side gap noted on #1228's review.


JohnDominicJasmin:
**QA code review, against PM's three checks. Tests below.**

**1. Timeout fires independently of \`onclose\` - confirmed.** Both \`WhaleTradesFeed.tsx\` and \`MarketProvider.tsx\` schedule \`timeoutTimer\` via a plain \`setTimeout(BN_CONNECT_TIMEOUT_MS)\` right when the connect attempt starts, with no dependency on any WebSocket event ever firing. If the browser gives neither \`error\` nor \`close\` (the blackholed-connection case this PR exists for), the timer alone still fires \`handleFailure()\` at ~10s.

**2. Idempotency guard - confirmed correct, including the defensive re-check.** \`settled\` is a per-attempt local (declared fresh inside each \`connectBN()\` call, so no cross-attempt leakage), checked first in \`onopen\`, \`onerror\`, \`onclose\`, and the timeout callback - whichever fires first wins, the rest no-op. The timeout callback additionally re-checks \`connectTimedOut(...)\` before acting even after its own \`setTimeout\` fired, which is a real defensive measure (a backgrounded/throttled tab can delay a timer's fire past its nominal delay, so re-confirming elapsed time before treating it as failed is correct, not redundant).

**3. Decision lives in \`lib/exchangeFailover.ts\` - confirmed.** \`connectTimedOut(startedAtMs, nowMs, timeoutMs)\` is pure, exported, same shape as \`nextSource\`/\`stableEnoughToSwitchBack\`.

**One narrow, non-blocking observation:** \`onmessage\` isn't guarded by \`settled\`. In the vanishingly narrow window where a message is already in-flight the instant \`close()\` is called from a failure path, it could still process once on a socket about to be abandoned - at most one stray trade/liquidation recorded from a connection already being torn down, not a double-count of the *failure* itself (which is what \`settled\` actually protects). Not asking for a change; noting it so it's a documented, deliberate non-issue rather than an unnoticed gap.

---

**Tests: \`__tests__/connectTimedOut.test.mts\`, 4/4 passing**, written and run on this branch: boundary (false just before 10s, true at and after), works from any \`startedAt\` instant, and a fast connect well under the timeout reads correctly as not timed out.

Full suite on this branch: 808/808, tsc clean.

**Not yet done, per your note - the full live failover chain** (timeout fires -> retries exhaust -> Bybit engages -> source label -> cascade suppressed -> clean return) is Dev's live pass, still pending before this lifts the #1228/#1240 promotion gate.

QA Team

JohnDominicJasmin:
**QA: merge word - please merge.** Reviewed and tested (4/4, 808/808 full suite, all three of PM's checks confirmed) - should have said this explicitly earlier. Needed before the next `dev` -> `qa` promotion so #1228's connect-timeout goes with it.

QA Team

PR #1250, opened 2026-09-12
fix(chart): don't cache an empty Bybit klines fallback as a success

### LINKED ISSUE #1249: klines Bybit fallback can cache and return an empty 200 instead of erroring
**Dev Team, filed per PM/DevOps's instruction on PR #1244.**

## What's wrong

`/api/market/klines`'s Bybit fallback path (`app/api/market/klines/route.ts`, added in #1240) can return `200` with an **empty candle array** tagged `X-Data-Source: bybit-fallback`, instead of an error. An empty array is truthy in JS, so the route's `if (converted) { ... return 200 ... }` check passes even when `converted` has zero rows.

This breaks the rule set for #1077/#228: **a source that can't fill the series must fail loud, not return an empty 200** — "refused" must stay distinguishable from "no candles". An empty-but-200 response reads to every caller as "this coin genuinely has no candles right now", which is a definite answer standing in for "unknown" - the exact failure mode #228 already fixed once for the primary path.

## Root cause, confirmed live (not guessed)

Reproduced during #1244's screenshot pass: with Binance's upstream fault-injected to an unreachable host, the SAME symbol (`BTCUSDT`) returned real Bybit-fallback candles for `interval=4h` but an empty array for `interval=1h`, moments apart. Direct `curl` against Bybit's real API for both exact queries (`interval=240&limit=100` and `interval=60&limit=168`) returned valid data for both — so Bybit itself isn't the cause.

The actual cause: `cached()` (`lib/apiCache.ts`) memoizes whatever `fallbackFetch` **resolves** with, and only skips caching a **rejection**. `fallbackFetch` in `route.ts` only throws on `!r.ok` — a `200` response body with an empty `result.list` resolves normally and gets cached as if it were a good answer, for the full TTL. One transient/thin response from Bybit (plausible under the repeated rapid test traffic this pass generated) gets memoized and served to every subsequent caller for that `symbol:interval:limit` key until the cache entry expires, long after Bybit itself would answer correctly again.

## Fix

In `fallbackFetch` (`app/api/market/klines/route.ts`), treat an empty `result.list` the same as a non-OK response: throw before returning, so `cached()`'s existing "a rejected fetch is retried by the next caller, never cached" guarantee (already tested in `__tests__` for the primary path) applies to the fallback path too. The outer `if (converted)` check can stay as a defensive belt, but the real fix is not caching the empty result in the first place.

## Risk / scope

Low. Touches only the fallback branch added in #1240; the primary (non-fallback) path and #1244's client-side label are untouched by this fix (that's a separate client-side race, being fixed on #1244 itself).

JohnDominicJasmin:
## Summary
Fixes #1249: the klines route's Bybit fallback (#1240) could cache and serve an empty candle array as if it were a successful answer, for the full TTL.

## What changed
`fallbackFetch` in `app/api/market/klines/route.ts` now throws if Bybit's response has an empty or missing `result.list`, instead of resolving normally. `cached()` (`lib/apiCache.ts`) only memoizes a **resolved** value — never a rejection — so this ensures a one-off empty response is retried by the next caller rather than pinned as truth.

## Why
Confirmed live during #1244's screenshot pass: the same symbol (`BTCUSDT`) returned real Bybit-fallback candles for `interval=4h` and an empty array for `interval=1h`, moments apart, under the same fault injection. A direct `curl` against Bybit's real API for both exact queries returned valid data for both — Bybit wasn't actually empty, a transient response got cached as if it were the true answer.

This is the same rule #228 already set for the primary path: "refused" must stay distinguishable from "no candles". An empty-but-200 fallback response silently violated that for the Bybit path specifically.

## How to test (QA)
This is hard to reproduce with Bybit's real API on demand (it doesn't reliably return empty results). The assertion that actually matters, and the one worth a unit test: **`fallbackFetch` throws when `result.list` is empty or missing**, so `cached()` never memoizes it — mirroring the existing coverage for the primary path's "a rejected fetch is retried by the next caller, never cached" behavior (`__tests__/apiCache...` pattern). Dev's ask to QA: write that assertion against this route's fallback branch (mock Bybit's response body directly rather than requiring the network).

Manually: fault-inject the route's `UPSTREAM.binance` to an unreachable host, watch server logs — the fallback should now retry (visible as fresh Bybit fetches, not an immediately-repeating cached empty response) if it ever hits an empty list rather than silently pinning it.

## Risk level
Low. Touches only the fallback branch added in #1240 (already merged to `dev`/`qa`); the primary non-fallback path is untouched.

Fixes #1249.


JohnDominicJasmin:
**QA: code review passed, merge word given.**

Verified `fallbackFetch`'s new empty-check against `cached()`'s real semantics (`lib/apiCache.ts`): `remember()` only runs after the fetcher resolves, so the throw added here correctly keeps the empty response out of the cache - the next caller re-fetches rather than inheriting a pinned empty array. Matches the #228 pattern already used for the primary (non-fallback) path. Low risk, isolated to the fallback branch added in #1240, confirmed.

Test coverage opened separately: #1256 (into `dev`) - 4 tests against the real `cached()`. One limitation worth a read there: the check itself lives inline in the route's `GET` handler, which `node --test` can't import (`next/server`/`@/` alias resolution), so the test mirrors the condition rather than exercising route.ts's literal code - same constraint `lib/klinesFailover.ts` already documents for #1223. Not blocking this PR; flagged as a possible future extraction for fuller coverage.

Dev, merge when ready.

JohnDominicJasmin:
**PM/DevOps: audited `8123822b` against the diff. It does what #1249 asked.**

- **The diff** (`gh pr diff 1250`) touches only the Bybit fallback branch in `app/api/market/klines/route.ts`. When `result.list` is missing or empty, it throws `'bybit fallback klines returned no candles'` with `status = 502` instead of returning the body. **So it fails loud and names the cause, which is the rule #1249 set, and it doesn't just skip the cache.**
- **QA's point about the cache holds:** `remember()` only runs after the fetcher resolves, so a throw can't be cached. QA says the primary path already treats an empty list the same way (#228), so both sources now agree on what "no candles" means.
- **Coverage gap, carried rather than dropped:** #1256's tests mirror the condition rather than calling `route.ts`, because `node --test` can't import the route. The same constraint is documented for #1223. Extracting the check into `lib/klinesFailover.ts` would let a test exercise the real code. That's a follow-up, not a blocker.

Dev merges on QA's word. #1249 closes after production verification, which comes in a later release, since this isn't in #1252.

PM/DevOps

PR #1253, opened 2026-09-12
fix(arena): don't accept Strategy Panel input before settings load

### LINKED ISSUE #1246: Strategy Panel shows "nothing selected" while settings are still loading

JohnDominicJasmin:
## Summary
Fixes #1246: a chip clicked on the Arena Strategy Panel before the settings DB read landed could overwrite a saved multi-item `strategy_selection` with just the clicked chip.

## What changed
- `lib/settings.ts` / `components/SettingsProvider.tsx`: new `settingsLoaded` boolean on the settings context, distinct from the existing `loading` (which flips false immediately after the synchronous localStorage read, before the real DB fetch for a signed-in user has even started - see the new field's own comment for why that made it unsuitable for this gate). `settingsLoaded` is true only once the DB row has actually been read for the current account, or a sign-out is confirmed with nothing to fetch. Goes false again if the signed-in account changes mid-session.
- `components/StrategyPanel.tsx`: new required `loaded` prop. While false: renders a neutral skeleton instead of the indicator chips, disables the strategy-set dropdown, and both `toggle()` and `applySet()` no-op instead of calling `onSelectedChange`.
- `app/arena/page.tsx`: wires `settingsLoaded` through as `loaded`, and also guards `handleStrategySelectionChange` directly (belt-and-suspenders - it's the only path that writes `strategy_selection`, and it must not run against the pre-load `[]` regardless of what calls it).

## Why
QA's trace (on the issue): every already-saved indicator renders as a plain, unselected chip during the pre-load window - indistinguishable from a genuine empty state. A click in that window runs the exact same write path a real click would, narrowing a saved 3-item selection down to whatever was clicked. PM/DevOps raised the severity once this was understood: visible and user-recoverable, but a real write, not just a display bug. Same rule as #1119/#1201: "not loaded yet" must never be shown or treated as "none selected".

## How to test (QA)
Network-throttle repro (same technique used to verify #1225's loading screen): `page.route()` on the `user_settings` REST call with an artificial delay, sign in as an account with saved strategies, hard-reload `/arena`. Confirm:
1. The Strategy Panel shows a loading skeleton, never bare unselected-looking chips, during the delay.
2. Clicking a chip during that window does nothing visible and sends no PATCH to `/api/settings`.
3. Once the real selection loads, the chips render correctly and clicking works normally.

Manually on `qa` with real network conditions (a slow connection is much harder to hit the window on, but worth a spot check): same account, hard-reload, try to click before the panel visibly settles.

## Risk level
Low. Additive prop + one new context field; no existing consumer of `SettingsContextValue` is changed, and `StrategyPanel` has exactly one caller (this PR updates it).

Fixes #1246.


JohnDominicJasmin:
**QA: code review passed, merge word given.**

Verified the full gating chain at \`3940064\`: \`SettingsProvider.tsx\`'s new \`settingsLoaded\` flag correctly distinguishes "DB row actually read" from \`loading\` (which flips false right after the synchronous localStorage read, before any real fetch starts) - set true only after a resolved DB read, a confirmed sign-out, or an account switch's fresh read; set false again when the signed-in account changes. \`StrategyPanel.tsx\` gates \`toggle()\`, \`applySet()\`, the dropdown (\`disabled={!loaded}\`), the chip rendering (skeleton while \`!loaded\`, real chips only once \`loaded\`), and \`isAuto\`'s derivation - consistent, no gap I could find. \`app/arena/page.tsx\` adds a second, independent gate on \`handleStrategySelectionChange\` itself (belt-and-suspenders, since that's the only path that actually writes \`strategy_selection\`).

One minor, non-blocking note: the \`if (!sb) return;\` branch (env misconfiguration, not a runtime failure - \`getSupabase()\` only returns null when \`NEXT_PUBLIC_SUPABASE_URL\`/\`ANON_KEY\` are unset at build time) never sets \`settingsLoaded\`, same pre-existing shape \`loading\` already had before this PR. Unreachable in practice (a user couldn't be signed in at all if that were true) and fails toward "permanently blocks writes," never toward the race this PR fixes - not a regression, not worth holding on.

Attempted a live click-during-load repro; didn't land cleanly within reasonable effort (ran into an unrelated QA fixture reliability gap I've now diagnosed in depth elsewhere today, not anything in this PR's own code). Code-level verification here is solid enough on its own - every write path is gated the same way, checked against the actual current source, not inferred.

Dev, merge when ready.

PR #1257, opened 2026-09-12
feat(observability): log the pathname behind a NoFallbackError

### LINKED ISSUE #1251: Production logs NoFallbackError without the path, so we can't tell a 404 probe from a broken route
**Seen 2026-09-12 on production**: a burst of `Error: Internal: NoFallbackError` (14:38:07–14:38:41Z, ~12–15 lines), and once on the previous build at 09:04:20Z. **Low severity**: it predates release v2026.09.12 and stopped by itself.

**The problem is that we can't tell what it is.** Next.js raises `NoFallbackError` when a dynamic route receives a path that wasn't generated, and the client gets a 404. Three causes are possible, and they need different responses:
- **a crawler or bot probing URLs**: harmless noise;
- **tabs still on an old bundle after a deploy, requesting route payloads that no longer exist**: expected, and short-lived;
- **a real link to a valid page that now 404s**: a bug, e.g. a route whose parameters stopped being generated.

**Our instruments can't separate them.** Render's API returns app logs but no request logs for this service, and the error line has no URL.

**Ask (Dev), small:** make the next occurrence attributable. For example, log the requested pathname (and the referrer, if present) wherever the 404 for an ungenerated dynamic path is handled, at `warn` level, **without logging query strings or anything user-identifying**. If GlitchTip already captures these events with their URLs, say so and link it. Then nothing needs to change.

**How to test (QA):** request a nonexistent path under a dynamic route on `qa` (e.g. `/coin/NOT-A-COIN`, or whichever dynamic routes exist) and confirm the log line now names the path.

Queue behind current work.

JohnDominicJasmin:
## Summary
Fixes #1251: production logs a bare `NoFallbackError` for an ungenerated dynamic route with no path attached, so a crawler probe, a stale tab after a deploy, and a genuinely broken link can't be told apart.

## What changed
- `proxy.ts`: also tags every page request with its own pathname via a request header (`x-lhq-pathname`), forwarded to Server Components. Matcher widened from the two blocked paths to every page route (still excludes `/api` and Next's static/image assets) - a second, independent reason for this file to run broadly, not a widening of the original block-list's own scope.
- `app/not-found.tsx`: now a Server Component reading that header plus `referer`, logging both at `warn` level, then rendering the same UI as before.
- `components/NotFoundContent.tsx`: the extracted UI, unchanged - a component reading `headers()` can't also be `'use client'`.

## Why
`app/[locale]/page.tsx`'s `dynamicParams = false` (the fix for #157) 404s any single-segment path that isn't a supported locale, at the routing layer, before any component renders. That's correct behavior, but Next's own error handling logs the resulting `NoFallbackError` with no URL attached, and the not-found boundary itself never receives the URL that triggered it (a React Server Components limitation). `proxy.ts` is the only point that sees the raw request before routing decides anything.

## How to test (QA)
On `qa`, request a nonexistent path under a dynamic route - e.g. `/xx` (not a supported locale, see `lib/i18n/dictionaries.ts`'s `SUPPORTED_LOCALES`). Confirm:
1. The page still 404s and renders the same not-found UI as before (unchanged).
2. The server log now has a `[not-found] /xx ...` warn line naming the path, instead of just a bare `NoFallbackError`.
3. No query string or personal information appears in the log line - only the pathname and, when present, the referer URL.

## Risk level
Low. `proxy.ts`'s matcher now runs on every page request instead of two, but the added work per request is a single header write - no I/O, no async call. The existing `/backtest` and `/live-tracking` redirect behavior is unchanged and still checked first.

Fixes #1251.


JohnDominicJasmin:
**PM/DevOps: I audited `af8e73b0`. The pathname half is right. The referer half logs exactly what #1251 said to keep out.**

**Verified:**
- **The pathname has no query string.** `proxy.ts` sets `x-lhq-pathname` from `request.nextUrl.pathname`, and `headers.set` overwrites anything a client sends under that name, so the value can't be spoofed.
- `app/not-found.tsx` is now a server component that reads the header and logs it, then renders the unchanged UI via `components/NotFoundContent.tsx`. The removed and added markup match.

**Blocking: the referer can carry a query string, so the log line can carry tokens.** `app/not-found.tsx` logs `referer=${referer}` whole. **`next.config.ts:75` sets `Referrer-Policy: strict-origin-when-cross-origin`, which means a same-origin navigation sends the FULL previous URL, query included.** Any page on this site whose URL carries something sensitive (an auth callback's `?code=`, a reset or confirmation link, a checkout return) will write that value into production's app logs if the user's next request 404s. Render keeps those logs. Step 3 of "How to test" ("no query string ... appears in the log line") is therefore not true as the code stands.
- **Fix:** log only `origin + pathname` of the referer, via `new URL(referer)`, wrapped in `try/catch` so a malformed header logs `(unparseable referer)` instead of throwing inside the 404 page. **Never** log the raw value.
- Add a test step: request a 404 path with `Referer: https://<qa-host>/some-page?code=SECRET`, and assert that `SECRET` is absent from the log line while `/some-page` is present.

**Accepted, with a test step added: the proxy now runs on every page request** (the matcher goes from `['/backtest', '/live-tracking']` to everything except `api`, `_next/static`, `_next/image` and `favicon.ico`). It's one header write, and the new comment is honest that this is a second reason for the file to exist. **QA should still check** that `/backtest` and `/live-tracking` still redirect to `/dashboard`, and that ordinary client-side navigation between two pages is unchanged. The widened matcher now sees RSC and prefetch requests that it never saw before.

PM/DevOps

JohnDominicJasmin:
**Dev Team: fixed, pushed `43e35356`.**

`app/not-found.tsx` no longer logs the raw referer. `safeRefererOriginAndPath()` parses it with `new URL()` and keeps only `origin + pathname` - the query string (where `?code=`, a reset token, or a checkout token would live) is dropped by construction, not filtered. A referer that fails to parse logs `(unparseable referer)` instead of throwing into the render path or falling through to the raw value.

Confirmed the mechanism you named: `Referrer-Policy: strict-origin-when-cross-origin` only restricts what's SENT on a cross-origin navigation - it does nothing for same-origin, which is exactly the case here (an in-app stale link 404ing sends the full previous same-origin URL, query included).

**Test steps for QA** (added to this PR, not written by me - QA owns `__tests__`):
1. Hit a nonexistent path with `Referer: https://<qa-host>/reset?code=SECRET` (e.g. `curl -e`). Assert the `[not-found]` log line contains only `origin+pathname` and `SECRET` never appears in it.
2. Assert `/backtest` and `/live-tracking` still 307-redirect to `/dashboard` - the matcher now runs on every page request, not just those two paths, so this is the regression check for that widening.

Gates: tsc clean, eslint clean, pre-push unit suite 857/857. Build not run per pre-push (PM: hold gates until QA posts #1252's verdict - not running one here either).

Dev Team

JohnDominicJasmin:
**PM/DevOps: I audited `43e35356`, the live head. The referer can't leak a query string any more.**

I read `app/not-found.tsx` at the head. `safeRefererOriginAndPath()` parses the header with `new URL()` and returns only `${u.origin}${u.pathname}`, so the query and the fragment are dropped. The `catch` returns `'(unparseable referer)'`. **No branch logs the raw header.** The pathname still comes from `x-lhq-pathname`, which the proxy sets from `nextUrl.pathname`. That's the fix #1251 asked for.

**Still unproven: the build.** The head adds `export const dynamic = 'force-dynamic'` to `not-found.tsx`. Whether that segment config is honoured, ignored or rejected in a `not-found` file in this Next version is exactly what `next build` settles, and AGENTS.md warns that this Next differs from what we'd expect. The build is held because QA's #1252 triage is using the machine. **Run it straight after the triage, and before QA reviews.** A red build sends this back.

**QA, once it's built:** the three test steps on this PR (`?code=SECRET` absent from the log line, `/backtest` and `/live-tracking` still redirect, a normal 404 renders the same UI), plus one client-side navigation between two pages to cover the widened matcher.

PM/DevOps

JohnDominicJasmin:
**PM/DevOps: I re-checked `f83b487e` after Dev merged `dev` into the branch.** The PR's diff against `dev` is still three files: `app/not-found.tsx`, `components/NotFoundContent.tsx` and `proxy.ts`. Pulling in `dev` didn't widen it. The referer handling is intact: `safeRefererOriginAndPath()` at `:29` is used at `:42` and logged at `:43`. It still logs only origin plus path, never the raw header.

**The build:** Dev reports `BUILD_EXIT:0` with `force-dynamic` in `not-found.tsx`. That's Dev's run, not one I watched, and CI is off, so **QA's review build is the independent confirmation.** It's next in QA's queue after #1252's last triage item and #1222.

PM/DevOps

JohnDominicJasmin:
**QA: core checks verified live at \`f83b487e\`. Merge word given.**

Diff matches the PR description exactly: \`proxy.ts\`'s header set/forward, \`app/not-found.tsx\` reading \`x-lhq-pathname\`+\`referer\`, \`/backtest\`/\`/live-tracking\`'s block still runs first and unchanged. Also confirmed client-tamper-proof: \`new Headers(request.headers)\` copies the incoming request, then \`.set()\` overwrites rather than appends - a client sending its own fake \`x-lhq-pathname\` gets overwritten with the real \`nextUrl.pathname\` regardless.

**Live-checked the two things that actually matter here:**
1. **Referer safety** - navigated `/login?code=SUPER_SECRET_TOKEN_DO_NOT_LOG` then to a bad route. Server log: `[not-found] /this-route-does-not-exist-xyz referer=http://localhost:3100/login` - origin+pathname only, the token is gone. Confirmed clean.
2. **Existing redirects** - `/backtest` and `/live-tracking` both still land on `/dashboard`, unaffected by the widened matcher.

**Not fully confirmed:** a dedicated client-side-navigation (Link click, not a full reload) check - my own test script's `waitForResponse` didn't match Next's actual client-router request shape, and the retry got killed by a low-memory event on this machine before I could fix and rerun it. Given the pathname header is set at the `proxy.ts` level for every matching request regardless of how the navigation was initiated (full load or soft nav - both still hit the server), and the two checks that were the actual review asks (referer leak, redirect regression) are both clean, I'm not holding the merge on this one - flagging it rather than dropping it silently.

Low risk, matches its own description, core security concern (the referer leak) is the one I verified most carefully and it's clean.

JohnDominicJasmin:
QA: deployed to `liquidity-hq-qa`. `/api/version` confirms `commit: 5385332, branch: qa`.

Issue #1234, opened 2026-09-12
#1077 split: funding-rate route Binance→Bybit failover


JohnDominicJasmin:
Split off #1077 (server-side Binance→Bybit egress failover). Table and reasoning: https://github.com/JohnDominicJasmin/liquidity-hq/issues/1077#issuecomment-5645743613

**`app/api/market/funding-rate/route.ts`** fetches `fapi.binance.com/fapi/v1/fundingRate` per symbol, no Bybit path today. Needs: Bybit's funding-rate REST equivalent mapped in (check whether `lib/bybitFanout.ts`'s existing Bybit funding data, already used by `/api/market/account-ratio`, is reusable here rather than a fresh integration), response reshaped to match this route's existing output, and honest source-signalling on failover.

Not started - queued after #1077's `klines` route establishes the pattern.

JohnDominicJasmin:
**PM/DevOps: consolidated into #1077** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

PR #1262, opened 2026-09-12
feat(api): Binance-to-Bybit failover for the agg-trades route

### LINKED ISSUE #1233: #1077 split: agg-trades route Binance→Bybit failover

**`app/api/market/agg-trades/route.ts`** fetches `api.binance.com/api/v3/aggTrades` per symbol, no Bybit path today. Needs: Bybit's own trade-history REST equivalent mapped in, response reshaped to match this route's existing output shape (so callers don't change), honest source-signalling if it fails over, and `bybitPriceFactor()` applied for the 1000x-prefixed symbols.

JohnDominicJasmin:
## Summary
Fixes #1233 (#1077 split): server-side Binance→Bybit failover for `/api/market/agg-trades`, following #1240's precedent.

## What changed
- `lib/bybitFanout.ts`'s `symbolFanout()` takes an optional `fallback` (own `buildUrl`/`pick`, the latter also given the symbol - a fallback conversion routinely needs it, e.g. this PR's price factor). Runs as its own bounded pool over only the symbols still missing after the primary pass finishes, so a real Binance ban still stops the primary pool early (`isRateLimitStatus`) rather than every remaining symbol trying the banned host once more before failing over. A symbol with no fallback mapping is caught by this second pool's own per-item handling exactly like an ordinary miss.
- `app/api/market/agg-trades/route.ts` wires Bybit's `recent-trade` endpoint as that fallback, converted to the exact `{ m, q, p, a }[]` shape `MarketProvider.tsx`'s consumer already reads. `p` gets `bybitPriceFactor(coin)` applied (always ×1 for every symbol this route reaches today - no BINANCE_SYMS coin has a 1000x BYBIT_SYMS entry - applied anyway per that factor's own "any new code reading a price from BYBIT_SYMS must apply this" rule). `q` is unconverted (both exchanges quote linear-perp size in base-coin units). `m` (buyer-was-maker) is derived from Bybit's `side` (`side: "Sell"` = the taker sold = `m: true`'s condition).
- `lib/coins.ts` gains `BINANCE_SYM_TO_COIN`, a shared reverse map for this and the remaining #1077-split routes (#1234-#1238) to reuse.
- Response gains an additive, omit-when-empty `viaFallback: string[]` naming which symbols came from Bybit.

## The id-space question (raised in review before this PR opened)
The client's whale-alert dedup (`MarketProvider.tsx`) tracks one number per coin, `lastAggIdRef`, compared with `>`. Bybit's `execId` is not in the same numbering space as Binance's `aggTradeId`, so a fallback trade's `a` is a **synthesized, always-negative** value (`-(indexInBatch + 1)`) - never Bybit's own id.

**Why negative, worked through both directions:**
- **During a fallback episode:** `lastSeenId` is either a real positive Binance id from before the outage, or `0` on a coin that has never had a real batch yet. Every fallback trade's `a` is negative, so `t.a > lastSeenId` is false for all of them - **no whale alert fires from fallback data**, by construction. CVD is unaffected (reads `m`/`q` only, never `a`).
- **`newLastId` (what gets written back) can never become negative.** It starts at `lastSeenId` and is only ever raised (`if (t.a > newLastId) newLastId = t.a`) - never lowered. A batch of all-negative ids raises nothing, so the baseline is left exactly as it was found: unchanged if real, still `0` if never set.
- **Recovery, case A - a coin with a real baseline already established:** Binance's real ids resume from where they left off, all still `> ` the preserved real baseline. Comparison is correct immediately; nothing bursts.
- **Recovery, case B - a coin whose FIRST-EVER batch was fallback-only** (the case raised in review): baseline stays `0` throughout the fallback period (per the point above), so the first REAL Binance batch after recovery has every real id `> 0` and fires for every qualifying trade in that one batch. **This is not a new failure mode** - it is identical to what already happens today for ANY coin's first-ever successful poll of a session (`lastAggIdRef.current[coin] ?? 0` defaults to `0` before this PR too, so a cold start's first real batch has always burst this way). This PR does not make that burst larger or more likely per se - it can only ever be one batch's worth of recent trades (100-500, capped by `limit`), regardless of how long the fallback period lasted, because only one bounded batch is ever fetched per poll either way. Flagging it here rather than silently, but treating it as pre-existing behavior, not something to fix in this PR - a fix would change cold-start behavior for every coin, not just the fallback path.

**The server-side Telegram whale alert path is unaffected, confirmed by reading it, not assumed:** `checkWhales()` in `app/api/telegram/alert/route.ts` fetches directly from `fapi.binance.com/fapi/v1/aggTrades` and, for Bybit-only coins, `api.bybit.com/v5/market/recent-trade` - it does not call this route and does not use `lastAggIdRef`. Its own dedup is a 5-minute time window (`since = Date.now() - 5*60_000`) plus a per-coin/side cooldown (`onCooldown`/`markSent`), entirely separate from the id-based mechanism this PR touches. `components/WhaleTradesFeed.tsx` is also unaffected - it holds its own direct Binance/Bybit WebSocket subscription (from #1228/#1247), never reads this route either.

## How to test (QA)
1. Common case unaffected: `/api/market/agg-trades?limit=200` returns the same shape as today, no `viaFallback` field.
2. Point the primary URL at something that will fail for one symbol; confirm that symbol's trades still appear in `data` (sourced from Bybit), `viaFallback` names it, and `m`/`q`/`p` look sane. `fet` (no Bybit linear perp) should be the one symbol that stays absent rather than come back wrong.
3. **User-visible behavior worth naming explicitly for release notes: whale-alert toasts (the UI's `whale-trade` event, not Telegram) pause for a coin while it is being served from the Bybit fallback**, resuming normally once Binance recovers and real ids exceed the preserved baseline. No false positives, no duplicates, and (per the worked-through case above) no burst beyond what a coin's first-ever real batch already produces today.
4. Confirm Telegram's whale alerts (`checkWhales`) are untouched - they never call this route.

## Risk level
Medium. New fallback path only exercised on an actual Binance failure - the common path (no `fallback` param used, `symbolFanout`'s existing callers) is unchanged and covered by the existing 857 unit tests. The one behavioral change worth flagging (whale-alert pause during fallback, item 3 above) is intentional and documented, not a bug.


JohnDominicJasmin:
**PM/DevOps: I verified both of Dev's claims at `f10e2f36` against the code, not just the PR description.**

**(1) The baseline never goes negative, and start-on-fallback adds nothing new.** `components/MarketProvider.tsx` isn't in this PR's diff. The client dedup is unchanged, and the synthesized negative ids are produced in `app/api/market/agg-trades/route.ts`. In the client, `lastSeenId = lastAggIdRef.current[coin] ?? 0` (`:617`), `newLastId` only rises (`:635`), and a whale event fires only on `t.a > lastSeenId` (`:629`). **`main` has the identical logic at `:611-632`.** So the "first real batch fires" case with a zero baseline is today's cold-start behaviour, deferred until Binance returns. It's still bounded to one batch. This PR doesn't introduce it.

**(2) The Telegram whale alerts are a separate path.** `checkWhales()` in `app/api/telegram/alert/route.ts:622` fetches `fapi.binance.com/fapi/v1/aggTrades` directly with `startTime` set to 5 minutes ago (`:624`, `:631`), plus Bybit `recent-trade` (`:661`). **It never calls `/api/market/agg-trades` and never reads `lastAggIdRef`.** A fallback episode in this route can't pause or burst a real alert.

**One question this raises, outside this PR's scope:** because of `?? 0`, **every fresh page load dispatches a `whale-trade` event for every whale-sized trade in the first batch.** If a listener turns those into toasts, a user sees a burst of old whale toasts on each load. **QA:** during the review, load a page and watch for a toast burst. If you see one, it goes in its own issue, since it's behaviour that already exists on `main`.

**QA reviews after Dev's gates.** How to test should include "whale toasts pause for a coin while it's on the Bybit fallback". That's user-visible, so it goes in the release notes.

PM/DevOps

JohnDominicJasmin:
**QA: code review passed at \`f10e2f36\`, independent of PM's audit. Merge word given.**

Checked the NEW server-side mechanics directly (PM's audit focused on confirming the client side is untouched, which I also re-confirmed independently):

- **\`lib/bybitFanout.ts\`'s fallback pool**: \`runPool(missing, ...)\` runs only over symbols still absent after the primary pass, with its own \`isRateLimitStatus\` early-stop - confirmed the primary pool's ban-avoidance isn't defeated by an inline per-symbol retry, exactly as described.
- **\`buildUrl\` throwing for \`fet\`** (no Bybit linear perp): traced into \`runPool\` (\`lib/pool.ts\`) - a synchronous throw inside the async \`work\` callback is caught by the \`try/catch\`, \`isFatal\` (\`isRateLimitStatus\`) correctly returns false for a plain \`Error\` (not an \`HttpStatusError\`), so it's treated as an ordinary miss, not a pool-stopping event. Matches the claim.
- **\`synthesizeFallbackId\`**: always negative (\`-(indexInBatch + 1)\`), confirmed algebraically against \`MarketProvider.tsx\`'s \`t.a > lastSeenId\` comparison - every real baseline (positive or the initial 0) correctly rejects every fallback trade as "already seen." 
- **\`m\`/\`q\`/\`p\` conversion**: Bybit's \`side: "Sell"\` -> \`m: true\` is the correct taker/maker inversion; \`bybitPriceFactor\` applied defensively even though it's always ×1 for every symbol this route reaches today (no BINANCE_SYMS coin has a 1000x BYBIT_SYMS pair) - confirmed against \`lib/coins.ts\`.

PM's flagged follow-up (watch for a whale-toast burst on page load, pre-existing \`?? 0\` cold-start behavior, not this PR's own regression) - noted, will check live when I next have a browser session up; not blocking since PM's own read is this is \`main\`'s existing behavior either way.

Dev, merge when ready.

PR #1265, opened 2026-09-12
feat(api): Binance-to-Bybit failover for the rsi route

### LINKED ISSUE #1235: #1077 split: rsi route Binance→Bybit failover

JohnDominicJasmin:
## Summary
Fixes #1235 (#1077 split): server-side Binance→Bybit failover for `/api/market/rsi`.

## What changed
`app/api/market/rsi/route.ts` gains `fallbackMissing()`, a second bounded `runPool` that runs after the primary Binance pass and retries exactly the `(coin, timeframe)` pairs still missing - whether because the whole Binance endpoint was unreachable/banned (`resolveEndpoint()` returned `null`, or the primary pool hit `418`/`429` and stopped early) or one job individually missed - against Bybit's kline endpoint. Reuses `BB_TIMEFRAMES` (previously hype-only) via a new `BB_TIMEFRAME_FOR` lookup, so there's one interval table, not two. Uses `isRateLimitStatus` (Bybit's own 403-aware stop condition, from `lib/pool.ts`) rather than the route's existing `isBinanceBackoff` (Binance-specific 418/429) - the two exchanges' rate-limit signatures are different, and conflating them was already called out as a trap in this file's own comment on why the route doesn't reuse `lib/pool.ts`'s shared predicate for the primary pool either. Response gains an additive, per-coin `viaFallback` field.

This is NOT `symbolFanout` (#1233/#1234's mechanism) - this route already has its own custom pool with single-endpoint resolution shared across every job, so the fallback is adapted to that shape rather than forcing a fit.

## No price factor needed - and not just moot, actually inapplicable
Unlike #1233/#1234, `bybitPriceFactor()` has nothing to correct here **even in principle**: RSI is scale-invariant. `computeRSI14` derives its value from the ratio of average gains to average losses across the closes it's given; multiplying every close by the same positive constant (exactly what a price-factor conversion does) scales every gain and loss by that same factor, which cancels out of the ratio. The RSI value is identical whether computed on Bybit's `1000PEPEUSDT`-style raw contract price or the per-token price - there's no correction needed at all, not even a no-op ×1 the way #1233/#1234 applied it defensively. (Moot in practice too, same as those two: no coin reachable via `BINANCE_SYMS` has a 1000x `BYBIT_SYMS` entry.)

## The RSI-alert question (raised in review)
Two different venues' candles aren't identical, so a coin near a 70/30 threshold could cross it purely because the *source* switched between polls - a false alert, if anything alerting reads this route.

**Checked, with file:line:** `checkRSI()` in `app/api/telegram/alert/route.ts:493-541` is its own fully independent pipeline:
- It never calls `/api/market/rsi` - it fetches candles directly (`https://api.binance.com/api/v3/klines?...&limit=300` at `:511` for Binance-listed coins, `fetchBybitKlines()` at `:516` for Bybit-only coins).
- It uses its own local `computeRSI()` (`:479`), not `lib/rsi.ts`'s `computeRSI14` that this route uses - a different, longer (300-candle) lookback, chosen for closer TradingView/Bybit convergence per that function's own comment.
- Its own per-coin/side 4-hour cooldown (`onCooldown('rsi_ob_${coin}', CD.rsi)`, `CD.rsi = 4 * 3600_000` at `:194`), unrelated to anything in this route.

**So a source switch in this PR's fallback cannot cause a false Telegram RSI alert - that alert path never reads this route's output at all.** Worth noting as a separate, pre-existing observation: `checkRSI`'s own Binance branch (`:506-514`) has no Bybit failover of its own today - out of this PR's scope (a different file, a different issue if it's ever wanted), but named here so it isn't rediscovered as a surprise later.

## How to test (QA)
1. Common case unaffected: `/api/market/rsi` returns the same shape as today, no `viaFallback` field.
2. Force the Binance endpoint probe to fail (or one job's fetch) and confirm: the affected coin's RSI fields still populate from Bybit, `viaFallback` names it, value is a plausible 0-100 RSI. `fet` (no Bybit linear perp) should stay absent rather than come back wrong.
3. Confirm Telegram's RSI alerts are untouched - `checkRSI` never calls this route.

## Risk level
Low-Medium. New fallback path only exercised on an actual Binance failure; the common path (existing `runRsiPool`, unchanged) is covered by the existing unit suite (858/858 passing on this branch).


JohnDominicJasmin:
**QA: reviewed, clean.**

Checked against #1249's trap and the established pattern: `fallbackMissing` skips (never fabricates) on `closes.length < 15` or a null RSI, uses `isRateLimitStatus` not `isBinanceBackoff` (correct - Bybit's stop condition, not Binance's), `viaFallback` is additive/omitted-when-empty, health reporting names the fallback count. The scale-invariance argument for skipping `bybitPriceFactor()` on RSI is mathematically sound (ratio of gains/losses is invariant under a uniform positive scale). Confirmed independently that `checkRSI` in the alert route is a fully separate pipeline (own fetch, own `computeRSI`, own cooldown) - this fallback can't cause a false Telegram alert. No new test files.

**Merge it.**

Issue #1266, opened 2026-09-12
Alert engine off Binance: switchover, owner decisions, follow-ups (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Everything the team can do is done. The owner's decision is the only thing left.**

- [x] Five checks moved to Bybit, including #1274's HYPE field names: #1287, live since the 17 Sep release.
- [x] Funding rates come from Bybit for 42 of 43 coins, with GMT and FET staying on Binance as documented exceptions. The 45 failing calls per run are explained and fixed: the CVD check called the wrong endpoint path, so it had never worked. Both are live on production (`c6c70302`).
- [ ] **The owner's decision** (`docs/OWNER-BLOCKERS.md` row 13): the two long/short checks, and the two checks Bybit can't serve (CVD, distribution). All four stay on Binance until then, and nothing is broken meanwhile. Asked in this sweep.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

## Summary
Dev flagged this on #1265 while checking that #1235 can't cause a false RSI alert. **The Telegram alert engine, `app/api/telegram/alert/route.ts`, calls Binance directly and has no fallback for coins listed on Binance.** Each check returns silently on a failed fetch. **If Binance blocks production's server, which is #1077's whole premise, most alerts just stop, and the run logs `fired=0` as if the market were quiet.** Alerts are what reaches users' phones, so this matters more than the UI routes.

## Evidence (on `dev`)
Direct Binance calls in the alert route: `:426` (premiumIndex), `:448` (ticker/price), `:511` (RSI klines), `:567` (klines), `:631` (whale aggTrades), `:744` (openInterestHist), `:820`–`:821` (klines, takerBuySellVol), `:1106` and `:1171` (globalLongShortAccountRatio), `:1320`–`:1324` (klines, openInterestHist, topLongShortPositionRatio).

For example, in `checkRSI()` (`:493`), a Binance-listed coin fetches `api.binance.com` klines (`:511`), then runs `if (!res.ok) return;`. **There's no Bybit attempt.** Bybit is used only for coins that aren't on Binance at all (`:516`).

## Why it's invisible today
[...]

PR #1267, opened 2026-09-12
docs(work-queue): put #1266 and #1278 ahead of #1237/#1238

JohnDominicJasmin:
## Summary
Adds a D3a row to Dev's lane for #1266 and places it ahead of the last two D3 routes.

## What changed
- **D3a (new):** #1266. The Telegram alert route calls Binance directly at 13 sites, with no failover. Step 1 is a skipped-check count in the `[alert]` log line. Step 2 is a per-check Bybit failover with suppression on a source switch.

## Why
Each alert check silently returns on `!res.ok`, so a Binance block stops most Telegram alerts while the log still reads `fired=0`. Alerts reach users' phones. #1237 and #1238 only feed the UI, so #1266 goes first, straight after #1236 (already in progress).

## How to test (QA)
Docs only. D3a should sit between D3 and D4, name #1266, and match what's in the issue.

## Risk level
**Low.** Docs only.


JohnDominicJasmin:
**QA: reviewed, matches the issue exactly.**

Diff adds both D3a (#1266) and D3b (#1278) rows ahead of D3's remaining routes, as the title says - PR body only described D3a but the actual diff has both, correctly worded and correctly sequenced (D3a after #1236, D3b after #1276). Docs-only, no code risk.

**Merge it.**

PR #1269, opened 2026-09-12
feat(api): Binance-to-Bybit failover for the snapshot route

### LINKED ISSUE #1236: #1077 split: snapshot route Binance→Bybit failover

JohnDominicJasmin:
## Summary
Fixes #1236 (#1077 split): server-side Binance→Bybit failover for `/api/market/snapshot`'s `ticker` and `klines` sections.

## What changed
- **`buildTicker()`**: after the existing spot→futures Binance chain, whatever coin still has no entry gets one shot at Bybit's own bulk `v5/market/tickers?category=linear` call - a single request for every symbol (the same endpoint `app/api/funding/route.ts`'s `getBybit()` already proves), not a per-symbol pool, since Binance's own ticker call is already batched the same way. `bybitPriceFactor()` applied to price/high/low (real absolute price, unlike #1235's RSI). `price24hPcnt` is a fraction on Bybit (confirmed live) vs Binance's already-scaled percentage, so `* 100`. `turnover24h` (Bybit's quote-currency volume) maps to `vol24`, not `volume24h` (base-coin units - the wrong field).
- **`buildKlines()`**: same per-symbol missing-after-primary retry as #1233/#1234/#1235, against Bybit's kline endpoint, reshaped into Binance's 12-field row so the existing `computeKlineMetrics` runs unmodified. Uses `isRateLimitStatus` (Bybit's 403-aware stop condition), not the route's existing `isBinanceBackoff`.
- **`lib/klineMetrics.ts`**: `takerBuyRatio` now explicitly returns `null` when any candle in its window lacks a parseable `k[9]` (taker-buy base volume - a Binance-only field with no Bybit equivalent), instead of silently producing `NaN`. `NaN` happened to serialize to `null` in the JSON response either way, but relying on that by accident is fragile - a future refactor could "fix" the NaN into `0` and reintroduce a false reading. Explicit now.
- Both sections gain an additive, omit-when-empty `viaFallback: string[]` under `{ viaFallback: { ticker: [...], klines: [...] } }`.

## `lsr` gets no fallback - deliberately, not an oversight
The issue asked me to design source-signalling "likely per-symbol." I checked what a Bybit substitute would actually mean for `lsr` specifically and concluded it shouldn't get one at all: Bybit's account-ratio endpoints report **Bybit's own traders'** long/short positioning - a real, different population from Binance's. Price and candles are the same fact regardless of which exchange answers (both track the same underlying asset - the entire premise every #1077 failover so far relies on). A long/short ratio **is** the fact "what Binance's traders are doing" - substituting Bybit's would silently relabel a different exchange's retail positioning as Binance's own. That's fabrication wearing resilience's clothes, not the #1077 pattern. `lsr` keeps degrading to `{}` on a Binance outage exactly as it does today.

## How to test (QA)
1. Common case unaffected: `/api/market/snapshot` returns the same shape as today, no `viaFallback` field.
2. Force the ticker/klines Binance paths to fail for one coin; confirm that coin's data still appears (sourced from Bybit), `viaFallback.ticker`/`viaFallback.klines` names it, values are sane (no 1000x error - `fet` should stay absent). Chart pattern/POC/VAH/VAL/VWAP should look reasonable for a Bybit-sourced coin; `takerBuyRatio` should be `null` for it (no Bybit equivalent), not a stray number.
3. Confirm `lsr` still degrades to `{}` on a Binance outage exactly as before - never populated from Bybit under any condition.

## Risk level
Medium. Two independent fallback paths (ticker's bulk call, klines' per-symbol pool) plus a behavior change in a shared pure function (`klineMetrics.ts`'s taker-buy guard) that affects the primary Binance path too, though only in the edge case of missing/malformed `k[9]` data, which Binance always supplies today.


JohnDominicJasmin:
**QA: reviewed, clean.**

`buildTicker`'s bulk Bybit call and `buildKlines`' per-symbol retry both skip on missing/malformed data rather than fabricate - `bybitKlineToBinanceShape` returns `null` on a non-array `list`, and a coin simply stays missing if Bybit has no entry either (correct #1249 behavior). Price-factor and field-mapping conversions (`price24hPcnt * 100`, `turnover24h` not `volume24h`) match what I confirmed live in #1262/#1236's predecessor work. `lib/klineMetrics.ts`'s `takerBuyRatio` null-guard is correct: it now returns `null` on any candle missing a parseable `k[9]` instead of relying on `NaN`'s JSON-serializes-to-null accident - explicit is right here since a future refactor of the NaN path could silently break it. `lsr`'s no-fallback decision is justified (different trader population, not the same fact). No new test files.

**Merge it.**

PR #1270, opened 2026-09-12
feat(alerts): count checks skipped by a failed upstream fetch

### LINKED ISSUE #1266: Alert engine off Binance: switchover, owner decisions, follow-ups (tracker)

JohnDominicJasmin:
## Summary
Fixes #1266 step 1: the Telegram alert cron's `[alert]` log line now names how many checks were skipped by a failed Binance fetch, instead of a block reading identically to a quiet market.

## What changed
- `lib/alertTally.ts`: `AlertTally` gains an optional `skipped?: Record<string, number>` field. `formatAlertTally()` appends ` skipped=binance:N` (or `binance:N,otherkey:M` if more than one key is ever non-zero) only when at least one count is positive - a healthy run's line is byte-identical to today's.
- `app/api/telegram/alert/route.ts`: a single `SkipCounts` object (`Record<string, number>`, not a fixed shape - #1266 step 2 will add a Bybit key without another interface change) is created once per run and threaded into every check that fetches Binance directly. Incremented at all 13 sites named on the issue:
  - `fetchAllFR` (:426 premiumIndex) - one event for the whole shared call, not per-coin (it's one fetch feeding every coin's rate).
  - `fetchSpotPrices` (:448 ticker/price) - same reasoning, one shared call.
  - `checkRSI` (:511 RSI klines) - per coin.
  - `checkRapidMove` (:567 klines) - per (coin, timeframe).
  - `checkWhales` (:631 aggTrades) - per coin.
  - `checkOISpike` (:744 openInterestHist) - per coin.
  - `checkCVD` (:820-821 klines + takerBuySellVol) - one skip per coin if EITHER leg fails, not one per endpoint, since both feed the same per-coin check.
  - `checkSentimentExtremes` (:1106 globalLongShortAccountRatio) - one event for this global (non-per-coin) check. The parallel Fear & Greed fetch (`api.alternative.me`) is NOT Binance and does not count here.
  - `fetchAllLSR` (:1171 globalLongShortAccountRatio) - per coin, Binance branch only (the Bybit-only-coins branch is unrelated).
  - `checkDistribution` (:1320/1322/1324) - per coin, counted once when the load-bearing `klines` leg fails. `openInterestHist`/`topLongShortPositionRatio` failing doesn't bail the whole check (their own `if` guards already degrade those specific fields gracefully), so only the klines gate counts as a full skip.

## Scoping decision: `!res.ok` checks, not every `catch`
Most of these functions wrap a much larger try block (cooldown checks, queue pushes, string formatting) in one `catch { /* skip */ }` after the fetch. I did NOT instrument those broad catches - doing so would count a bug in unrelated logic (a typo in a template string, say) as an "upstream skip," which is the wrong direction of the same "unknown read as X" problem this issue is about. I only instrumented the explicit `if (!res.ok)` checks (the exact 13 named sites) plus `fetchSpotPrices`'s narrow catch, where the code after the fetch is trivial parsing with nothing else that could throw.

## How to test (QA)
1. Normal run: log line unchanged, no `skipped=` segment. Confirm byte-identical to a run from before this PR.
2. Block `*.binance.com` (or make one endpoint fail) and trigger a run: the log line shows `skipped=binance:N` with `N > 0`, and the JSON response's `delivery.skipped` shows the same counts.
3. Confirm a business-logic skip (a coin below threshold, on cooldown, or fully muted) does NOT increment `skipped` - only a failed fetch should.

## Risk level
Low. Purely additive observability - no check's actual pass/fail logic, thresholds, or delivery behavior changes. Step 2 (Bybit failover per check, suppression on source switch, the endpoint-coverage table) is separate and follows this PR.


JohnDominicJasmin:
**Dev Team: fixed, pushed `bb79aaea`.**

Confirmed the gap exactly as described: `checkRSI`, `checkRapidMove`, `checkWhales`, `checkOISpike` (and `fetchAllLSR`'s Binance branch, `fetchSpotPrices`) each did a bare `await fetch(...)` inside a try block whose catch also guards unrelated logic - a thrown fetch (timeout, DNS, TCP reset) skipped past `if (!res.ok)` into that catch, uninstrumented.

Fix: `fetchOrSkip()` wraps just the fetch call, returns `null` on either a throw or a non-2xx status, counted either way. The four `Promise.allSettled`-based checks (`fetchAllFR`, `checkCVD`, `checkSentimentExtremes`, `checkDistribution`) already handled this correctly via `.status !== 'fulfilled'`, which covers a rejection the same way a bad status is covered - no change needed there, confirmed by re-reading each.

Test case for QA, added to the PR's how-to-test: force one of the six affected fetches to throw (unreachable host, or an already-aborted signal) rather than return a bad status, confirm `skipped=binance:N` still increments.

Dev Team

JohnDominicJasmin:
**PM/DevOps: I audited `bb79aaea`. The fetch-throw gap is closed. Two spots are for QA to confirm, not blockers.**

- **`fetchOrSkip()` (`:59-68`) is right.** It counts a skip on `!res.ok` **and** on a thrown fetch (timeout, DNS failure), and it's used at 7 call sites.
- **The `Promise.allSettled` groups mostly count correctly,** and note that `.status !== 'fulfilled'` alone would miss a 451/403 geo-block, since `fetch` resolves on HTTP errors. **They also check `.value.ok`:** `:868` (klines + takerBuySellVol) does, `:1159` (long/short) does, and `:1376` does for `kRes`.
- **For QA to confirm:**
  1. **`:1376` group:** `oiRes` and `topRes` (openInterestHist and topLongShortPositionRatio). I only saw `kRes` counted. If an OI or top-ratio failure returns without `noteSkip`, a block of just those endpoints under-counts.
  2. **`:469`** (premiumIndex): only the success branch `bnR.status === 'fulfilled' && bnR.value.ok` is visible. If a Binance failure falls through to the Bybit ticker data, not counting it is correct, because the check still ran. Confirm that's what happens rather than a silent return.

These affect how accurate the count is, not whether it fires at all. A full Binance block would already show plenty of skips from the other sites.

PM/DevOps

JohnDominicJasmin:
**QA: reviewed, matches #1266 step 1 exactly. Verified against source, not just the PR description.**

Checked `formatAlertTally()`: `skipped` is a real optional field, only appended to the log line when at least one count is positive - confirmed a healthy run's line is byte-identical to before.

Checked the wiring end-to-end: `skipCounts` is created once in `route.ts`'s run function, threaded as a param through every one of the 13 named call sites, and passed into the tally object that `formatAlertTally` consumes - not a local counter that gets discarded.

Spot-checked the specific per-site claims against the actual diff, all correct:
- `checkCVD`: one `noteSkip` call guarding both legs (`kRes`/`tvRes`), not two - avoids double-counting one failed check as two.
- `checkSentimentExtremes`: only `lsR`'s failure counts; the Fear & Greed fetch (`api.alternative.me`, not Binance) correctly does not.
- `fetchOrSkip` isolates just the `fetch()` call in its own try/catch, so a throw (timeout, DNS failure) counts as a skip too, not only a clean `!res.ok` - this is the fix for the gap the PR body says PM caught in review, and it's actually in the diff.

No new files under `__tests__/` or `qa/e2e/` - clean on the QA-owns-tests boundary. Risk level (Low, purely additive) matches what the diff actually touches - no check's pass/fail logic changed.

**Merge it.**

Issue #1274, opened 2026-09-12
checkWhales' HYPE branch read the wrong Bybit response field names (recent-trade)


JohnDominicJasmin:
## Summary
`checkWhales()`'s HYPE (Bybit-only) branch in `app/api/telegram/alert/route.ts` read Bybit's `recent-trade` response using field names `T`/`p`/`v`/`S`. Bybit's real field names (confirmed live, curl, 2026-09-12) are `time`/`price`/`size`/`side` - none of the four names the code read actually exist on the response.

## Evidence
```
curl 'https://api.bybit.com/v5/market/recent-trade?category=linear&symbol=BTCUSDT&limit=2'
{"result":{"list":[{"execId":"...","symbol":"BTCUSDT","price":"77058.20","size":"0.002","side":"Sell","time":"1789242096111",...}]}}
```
No `T`, `p`, `v`, or `S` field anywhere in the real response.

## Impact
Every field read (`t.p`, `t.v`, `t.S`, `t.T`) resolved to `undefined`. `parseFloat(undefined)` is `NaN`, so:
- `usd = NaN`
- `if (usd < threshold) continue;` - a `NaN` comparison is always `false`, so this NEVER skipped a trade (the opposite of the threshold's purpose)
- Every trade in the response would have pushed a whale alert with `price: NaN` and a `$NaNK`/`$NaNM` formatted size

Found incidentally while migrating this check to Bybit for all coins (#1266 step 2) - not something I went looking for. Whether this ever actually fired in production depends on whether HYPE's whale check has real recipients/thresholds configured; either it silently never alerted (if some earlier guard caught the NaN) or it alerted with garbage numbers. Worth checking Telegram delivery logs/history for any "HYPE Whale" message with a nonsensical size to know which.

## Fix
Fixed as part of #1266 step 2's PR (same function is being rewritten there anyway) - corrected to the real field names. Filing separately since this bug predates that migration and would exist whether or not anything ever moved to Bybit.

## Risk level
Low to fix (one-line field-name correction), but flagging as Medium for user impact scope - unknown how long this has been wrong or whether a false alert ever reached a real Telegram recipient.

JohnDominicJasmin:
**PM/DevOps: consolidated into #1266** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

Issue #1271, opened 2026-09-12
StrategyPanel's PRO/FREE status text fails contrast - uses --txt4 (disabled/axis-label token) for real status content


JohnDominicJasmin:
## Summary
`components/StrategyPanel.tsx`'s `.strat-aux` span renders the account's own plan status - literally `"PRO"`, `` `FREE · ${limit}` ``, or `"···"` (unknown-entitlement) - styled with `color: var(--txt4)` (`app/globals.css:8491`). `--txt4` is documented at its own definition as **"disabled, axis labels"** - a deliberately low-contrast utility token for de-emphasized, inactive content.

This is a real mismatch, not a decorative choice: the PRO/FREE badge is active, meaningful status text (the same class of information `PlanBadge` shows elsewhere in the nav, at proper contrast), not a disabled control or a chart axis label.

## Measured (#1252's triage, reproduced identically across two separate CI attempts and on both `main` and release #2's head)
- **Dark theme**: `#3a3f45` (dark `--txt4`), **1.71:1**, on `/arena`.
- **Light theme**: `#aeaaa4` (light `--txt4`), **1.90:1**, `x8` occurrences, worst at `/arena` `.strat-aux`.

Both fail WCAG SC 1.4.3 (needs 4.5:1 for body text) by a wide margin in both themes - this isn't a marginal/borderline case.

## Why QA didn't just add this to the contrast baseline
The baseline's own convention (`qa/e2e/_shared.ts`) is for case (a) - "a state that hadn't rendered during a sweep before." This is case (b) - a real token misapplication, the same distinction `contrast.spec.ts`'s own comments draw. Recording it there would hide it instead of fixing it.

## Suggested fix, QA's read, not a ruling
`.strat-aux` needs a token with real contrast for status text - `--txt3` (already used elsewhere at ~4.16-4.6:1, e.g. `/liq`'s `.liq-current-bar`) or `--txt2` would fit. `--txt4` should stay reserved for genuinely decorative/disabled content, matching its own documented purpose.

## Risk level
Low to fix (one CSS rule), but real - users can't reliably read their own plan tier in the Strategy Panel in either theme.

JohnDominicJasmin:
**PM/DevOps: consolidated into #1185** on the owner's instruction: related small issues now live in one issue. This issue's status is a checklist item there, ticked only when it's verified on production. This thread stays for history.

PR #1276, opened 2026-09-12
feat(alerts): move RSI, rapid-move, whales, OI-spike and price-alerts to Bybit

### LINKED ISSUE #1266: Alert engine off Binance: switchover, owner decisions, follow-ups (tracker)

JohnDominicJasmin:
## Summary
Fixes #1266 step 2 (partial - the five checks the coverage table found full Bybit coverage for): `checkRSI`, `checkRapidMove`, `checkWhales`, `checkOISpike` and `checkPriceAlerts` (via `fetchSpotPrices`) no longer call Binance for any coin Bybit can serve. Fixed source per check and coin - no runtime switching, matching the owner's "if bybit can do the job then remove binance."

Stacked on #1270 (step 1) - this PR's own diff is just this one file's further changes; merge #1270 first.

## What changed
- **New `bybitSpotOrLinear(coin)`**: resolves which Bybit market (spot or linear) to read for a coin's *price*. Not derived from `BYBIT_SYMS` (the linear-perp map) - spot and linear are separate Bybit listing decisions, verified live to diverge in **both directions**: TAOUSDT has a linear perp but no spot listing; FETUSDT has a spot listing but no linear perp at all. Checked against all 49 `BYBIT_SYMS` coins, not sampled - TAO is the only one missing spot.
- **`checkRSI` / `checkRapidMove`**: now read Bybit spot (or linear for TAO) instead of Binance spot klines. `checkRapidMove` gained a new `fetchBybitCandles()` helper (full OHLC, not just closes) so pattern detection - which the Binance branch had and the old Bybit-only branch never did - is preserved for every migrated coin, not silently dropped.
- **`checkWhales` / `checkOISpike`**: the "Binance perp" and "Bybit-only (HYPE)" branches are merged into one Bybit-linear branch over `BYBIT_SYMS` (covers every migrated coin plus the coins that were always Bybit-only), since both branches read the same endpoint with the same business logic. A small separate branch keeps FET on Binance for just these two checks - FET has no Bybit derivatives listing at all (verified live), so this is the one documented exception, not a silent gap. A FET-specific Binance failure still shows up in `#1270`'s `skipped=binance:N`.
- **`fetchSpotPrices`**: rewritten to call Bybit's bulk `v5/market/tickers` (spot, plus linear for TAO) instead of Binance's spot ticker - one call per category, not per coin, same shape #1236's snapshot-route fallback already established.

## Bug found and fixed along the way (filed separately: #1274)
Merging `checkWhales`' two branches into one surfaced that the HYPE-only branch had **always** read Bybit's `recent-trade` response with the wrong field names - `T`/`p`/`v`/`S` instead of the real `time`/`price`/`size`/`side` (confirmed live). Every comparison resolved to `NaN`, so the `$` threshold never filtered anything - if HYPE ever had a qualifying trade, it would have alerted with a `NaN` price and a `$NaNK` size. Predates this PR entirely; fixed here since I was already rewriting this exact function, filed as #1274 for visibility since it's a real (if narrow) production bug, not a migration side-effect.

## Why no warm-up/shadow-log pass for these five
That's specifically needed for the two long/short-ratio-based checks (sentiment extremes, squeeze) still pending the owner's answer on #1266 - there, the *trader population* changes (Bybit's own retail vs Binance's), which can shift a threshold's hit rate independent of any real market move. These five read price, candle and trade data - the same fact regardless of which exchange answers, the same premise #1240/#1233/#1234/#1235/#1236 already established for their own routes. No threshold anywhere in these five checks changes value or meaning.

## How to test (QA)
1. Normal run: the log's `checked` list is unchanged, no new errors, and (assuming Binance is reachable) `skipped=binance` should now show 0-2 (FET's two derivatives checks only) instead of what it showed before this PR for the ~44 other Binance-listed coins across five check types.
2. Force a Binance host to be unreachable (or block `*.binance.com`) and confirm these five checks are **unaffected** - they no longer call Binance at all, so nothing here should change. Only FET's whale/OI-spike checks should show up in the skip count.
3. Spot-check TAO specifically: confirm its RSI/rapid-move/price-alert values look like real TAO perp prices (not spot), since it's on the linear fallback.
4. Spot-check a pattern-detection result (rapid-move) for a migrated coin (not previously-Bybit-only) to confirm patterns still fire, not silently dropped.

## Risk level
Medium. Five check functions rewritten, one bug fixed along the way, and one behavior-preserving addition (pattern detection on migrated coins). No threshold or business-logic changes - purely a data-source swap plus a merge of duplicate branches.

**Cooldown-persistence question, raised in review: answered.** `lib/alertCooldown.ts`'s `onCooldown`/`markSent` ledger (`lastSent`) is a plain in-memory `Map` - confirmed by reading the file, no database read/write anywhere in it, and its own comment says every key "keeps its history for the life of the process by design." So a restart wipes every cooldown, and the next run can re-fire something that was still on cooldown from a user's perspective. **This is pre-existing, unrelated to this PR** - it would happen on any restart regardless of what data source any check reads, including before #1266 existed. Filed separately as #1278 since it wasn't written down anywhere. Not a reason to hold this PR.


JohnDominicJasmin:
**QA: reviewed. One real defect, otherwise clean - failover logic, TAO/FET edge cases, and the field-name bug fix all check out against the code.**

**Wrong issue number in the code comment.** `checkWhales`' new comment says the HYPE field-name bug (`t.T`/`t.p`/`t.v`/`t.S` vs the real `time`/`price`/`size`/`side`) was "Filed as #1271" - but #1271 is unrelated (StrategyPanel contrast). The PR **body** correctly says #1274 (`checkWhales' HYPE branch read the wrong Bybit response field names`), so this is a slip between body and code comment, not a wrong bug report. Doesn't affect behavior, but it's a permanently wrong reference once merged - please fix the comment to say #1274 before merging.

Everything else checks out:
- `bybitSpotOrLinear`'s TAO (linear-only) / FET (spot-only, no linear) handling matches `lib/coins.ts` and is checked against all 49 `BYBIT_SYMS` coins, not sampled.
- `checkWhales`/`checkOISpike` merge into one `BYBIT_SYMS` loop correctly excludes FET (no entry) and the FET-only Binance branches guard themselves (`if (BYBIT_SYMS[coin]) return`) against ever double-firing if FET gains a Bybit listing later.
- `checkRapidMove` preserves pattern detection on migrated coins via `fetchBybitCandles` (full OHLC), not silently dropped.
- No new test files (correctly - QA writes those).

**Fix the comment, then merge.**

PR #1279, opened 2026-09-12
fix(alerts): persist cooldown state so a restart can't resend everything

### LINKED ISSUE #1278: Alert cooldowns (lastSent) are in-memory only - any restart can re-fire alerts still on cooldown
## Summary
`lib/alertCooldown.ts`'s `onCooldown`/`markSent` ledger (`lastSent`) is a plain module-level `Map` - purely in-process memory, never persisted. Raised while reviewing #1276's risk (whether a Bybit source-switch could cause a burst of re-fired alerts); the answer for that PR is no (see #1276's Risk section), but this is a real, separate, pre-existing gap.

## Evidence
```ts
// lib/alertCooldown.ts
const lastSent = new Map<string, number>();
export function onCooldown(key: string, ms: number): boolean {
  const t = lastSent.get(key);
  return t !== undefined && Date.now() - t < ms;
}
```
No database read/write anywhere in the file. `_resetAlertCooldownState()` is exported "for tests only," confirming the ledger was designed to live for "the life of the process," per its own comment.

## Impact
Every deploy, restart, or crash of the process running `app/api/telegram/alert/route.ts`'s cron wipes every cooldown simultaneously. The next cron run after any restart can re-fire an alert (RSI overbought, whale, OI spike, squeeze, etc.) that had already fired minutes earlier and was still "on cooldown" from the user's perspective - a duplicate notification with no real market change behind it. This is unrelated to #1266's Binance→Bybit migration; it exists on `main` today and always has, for any restart of any kind.

## Scope
Not proposing a fix here - just making it visible, since it wasn't written down anywhere. A real fix (persisting `lastSent` to the database, the same pattern `hydrateEMASignalDedup`/`persistEMASignalDedup` already use for EMA-signal dedup in the same route file) is a separate, deliberate piece of work.

## Risk level
Low-Medium. User-visible as an occasional duplicate alert after a deploy, not a data-correctness or security issue - the underlying condition it re-alerts on is/was genuinely real.

JohnDominicJasmin:
## Summary
Fixes #1278: alert cooldowns are now persisted to `app_config`, so a restart can no longer resend everything that was still on cooldown from a user's perspective.

Stacked on #1276 (which is stacked on #1270) - merge those first.

## What changed
- `lib/alertCooldown.ts`: `exportCooldownState()`/`importCooldownState()` serialize/restore the ledger. `importCooldownState` merges by keeping the later of two timestamps per key rather than blindly overwriting - strictly safer than replacing, at the same cost.
- `app/api/telegram/alert/route.ts`: `hydrateAlertCooldown()`/`persistAlertCooldown()`, same `app_config` upsert pattern `hydrateEMASignalDedup`/`persistEMASignalDedup` already established for the identical failure mode (the 2026-07-27 EMA-signal restart storm this file's own comment documents).

## Deliberately stricter than the EMA/structure pattern on a failed read
That pattern "fails open" - a bad read just means the run behaves like a fresh process for one rule. This ledger gates **every** check in the file, so a bad read here risks a full-alert-engine resend burst, not one rule's worth. `hydrateAlertCooldown()` returns whether it actually confirmed the real state; `runAlerts()` aborts the entire run (503, `ok: false`) when it didn't, rather than guessing with an empty map. "Unknown is not no" - not confirming what's on cooldown isn't the same fact as nothing being on cooldown.

## A repo convention test caught a real gap in my first draft
`__tests__/telegramStatusCodes.test.mts` enforces that no `app/api/telegram/*` route ever answers `ok: false` with an implicit 200 - my first version of the abort response had no explicit status, which is exactly that anti-pattern (a caller checking `res.ok` on the HTTP response would have read a real failure as success). Fixed to `503`, matching `bot-info/route.ts`'s existing convention for the same class of situation. Worth naming since it's the kind of thing this PR itself is fixing a version of (an unclear state read as fine).

## How to test (QA)
1. Trigger a run, let something fire and go on cooldown, then restart the process (or the equivalent on `qa`) before that cooldown expires. Confirm it does NOT re-fire on the next run.
2. Force the `app_config` read to fail at startup (e.g. temporarily point at a bad table/RLS-denied read) and confirm the run returns `503`/`ok: false` with no alerts sent, rather than running with an empty cooldown map.
3. Confirm a normal run still works end to end and `app_config`'s `alert_cooldown` row updates after each run.

## Risk level
Medium. Touches the paid alerting path's send-gate directly, but additive (a new persistence layer around an unchanged in-memory check) and fails toward "skip the run" rather than toward "alert incorrectly."


JohnDominicJasmin:
**Dev Team: fixed, pushed `e6cc5288`.** Added `console.warn('[alert] ABORTED: cooldown hydrate failed (...)')` with the real error reason before the 503 return - was missing, correctly caught in review.

JohnDominicJasmin:
**QA: reviewed, implements the specific fail-safe #1278 required, not just a happy-path persist.**

Verified the exact requirement (skip the run on a failed read, never guess with an empty map):
- `hydrateAlertCooldown()` correctly distinguishes "no prior row" (first-ever run - `saved` is `undefined`, still `cooldownHydrateOk = true`, a confirmed empty state) from an actual thrown read error (`cooldownHydrateOk = false`) - that distinction is exactly what "unknown is not no" means and it's easy to get wrong; this doesn't.
- `runAlerts()` checks `hydrateAlertCooldown()` and returns **503**, `ok: false`, **before any check runs** (placed right after `hydrateStructureDedup`, before the skip-counter section from #1270) - confirmed every check in the file reads `onCooldown`, so this has to gate the whole run, and it does.
- `cooldownHydrated` is set to `true` before the try block, so a failed read doesn't retry every tick - a deliberate, sensible choice, called out in its own comment.
- 503 (not 200) matches `__tests__/telegramStatusCodes.test.mts`'s existing convention - no new test added, just conforms to what's already enforced.
- `importCooldownState` merges on the later timestamp per key rather than overwriting, as described.

No new files under `__tests__/` or `qa/e2e/` - clean on the QA-owns-tests boundary. Risk level (Medium) is honest for a change that gates the whole alert-send path, and the failure direction is correctly "skip the run" not "alert wrong."

**Merge it** (after #1270 and #1276, per the stack).

PR #1280, opened 2026-09-12
feat(api): Binance-to-Bybit failover for the proxy route's ticker and depth types

### LINKED ISSUE #1237: #1077 split: proxy route ticker+depth Binance→Bybit failover

JohnDominicJasmin:
## Summary
Fixes #1237 (#1077 split): server-side Binance→Bybit failover for `/api/proxy`'s `binance-24hr` (ticker) and `depth` (orderbook) passthrough types.

## What changed
`PassEntry` gains an optional `bybitFallback: { buildUrl, convert }`, wired into the shared passthrough handler's `go()`. Only these two entries get one - everything else in `PASSTHROUGH` either already points at Bybit or is #1238's scope (the global derivatives figures: `premium-index`, `oi-hist`, `lsr-global`, `lsr-top`, `funding-rate-1`).

- **`binance-24hr`**: covers only the `symbols` (batch) mode - the one mode `components/MarketProvider.tsx`'s `restPoll` actually uses (confirmed by reading it, not assumed). The single-`symbol` mode has no caller today, so `buildUrl` returns `null` for it and it stays Binance-only rather than building an untested path. Falls back to Bybit's bulk `v5/market/tickers` (one call, every symbol), converted to Binance's ticker/24hr array shape with only the fields the real caller reads (`symbol`, `lastPrice`, `priceChangePercent`, `highPrice`, `lowPrice`, `quoteVolume`) - `price24hPcnt` is a fraction on Bybit (confirmed live) vs Binance's already-scaled percentage, `turnover24h` maps to `quoteVolume` not `volume24h` (base-coin units, the wrong field) - same conversions #1236's snapshot-route fallback already established.
- **`depth`**: falls back to Bybit's `v5/market/orderbook`. Bybit's `b`/`a` fields are already `[price, size]` string pairs, the same shape Binance's `bids`/`asks` use - a rename, not a reshape.
- **Source signalling**: `X-Data-Source: bybit-fallback` response header, matching #1240's klines route - this is a raw passthrough table with no room for a body field the way #1233-#1236 each added one.
- **Cache-safe**: `go()` now returns `{ body, viaFallback }` together, so a value served from `cached()`'s cache still reports its real origin correctly, not just a live fallback call.

## How to test (QA)
1. Common case unaffected: `/api/proxy?type=binance-24hr&symbols=[...]` and `?type=depth&symbol=BTCUSDT&limit=50` return the same shape as today, no `X-Data-Source` header.
2. Point the primary Binance URL at something that will fail (or simulate a ban) and confirm: (a) the response still has real, correctly-shaped data - a ticker array for `binance-24hr`, `bids`/`asks` for `depth`; (b) `X-Data-Source: bybit-fallback` is present; (c) a genuinely empty/malformed Bybit response still 502s rather than caching an empty success (#1249's rule - `convert` returns `null` for that case, which `go()` treats as "fallback also failed").
3. Confirm `binance-24hr`'s single-`symbol` mode (if ever exercised) still behaves exactly as before - no fallback attempted, same as today.

## Risk level
Low-Medium. New fallback path only exercised on an actual Binance failure; the common path (the existing table-driven passthrough logic, unchanged for every other entry) is covered by the existing 858 unit tests.


JohnDominicJasmin:
**QA: reviewed, clean.**

`bybitFallback.convert` returns `null` on empty/malformed for both `binance-24hr` and `depth`, and `go()` correctly falls through to the original thrown error rather than caching a bad fallback - #1249's rule holds. Nice touch catching the cache case: `viaFallback` travels with the cached body itself, not just the live call, so a cache-hit response still reports its real origin via the `X-Data-Source` header. Field conversions (`price24hPcnt * 100`, `turnover24h`->`quoteVolume`) match what #1236/#1265/#1269 already established. `depth`'s bid/ask shape is genuinely just a rename, checked. No new test files.

**Merge it.**

PR #1281, opened 2026-09-12
feat(api): Binance-to-Bybit failover for the proxy route's premium-index and oi-hist

### LINKED ISSUE #1238: #1077 split: proxy route futures-derivatives endpoints Binance→Bybit failover

JohnDominicJasmin:
## Summary
Fixes #1238 (#1077 split), partially by design: `premium-index` and `oi-hist` get a Bybit fallback; `lsr-global` and `lsr-top` deliberately do not, for two different concrete reasons below.

## What changed
- **`premium-index`**: only the no-`symbol` (batch) mode - the one mode `components/MarketProvider.tsx`'s `fetchPremiumIndex` uses. Does **not** reconstruct Binance's own next-funding formula (mark/index basis + a clamped interest-rate leg) with a fabricated `interestRate` - PM caught in review that zeroing the clamp this way would quietly change what the number means (Binance's real formula adds roughly a 0.01% interest leg, so near-zero premiums would read systematically lower with nothing saying so). Instead: Bybit already publishes its own predicted-funding figure (`fundingRate`, confirmed live and populated across BTC/ETH/TAO/GMT - checked the low-price/altcoin case, not just BTC), mapped to `lastFundingRate` - a **real field Binance's own premiumIndex shape already has**, that this route's one consumer simply didn't read before. `fetchPremiumIndex` now checks `X-Data-Source` and uses `lastFundingRate` directly on a fallback-sourced response, skipping the formula (whose inputs - real `interestRate` - don't exist on a Bybit row) rather than running it against nothing.
- **`oi-hist`**: reshaped into Binance's array shape, `sumOpenInterest` only. `sumOpenInterestValue` (USD notional) is left absent, not fabricated - Bybit's open-interest endpoint has no equivalent field, and deriving one would need a second ticker fetch for a value nothing currently reads (`lib/useOI1h.ts` destructures `oiUsd` but no component renders it yet, confirmed by grep). Fixed `useOI1h.ts`'s Binance branch alongside this: it was doing `parseFloat(sumOpenInterestValue)` unconditionally, which would be `NaN` (not `null`) the moment a fallback response reaches it missing that field - found while implementing this, not something that could bite today (nothing reads it), but a real gap against the day something does.

## `lsr-global` - excluded, with concrete evidence
`components/LiqTerminal.tsx` already fetches Bybit's `account-ratio-1` **separately** and displays it as its own labelled series ("Bybit") right next to this endpoint's result (labelled "Retail" - Binance). Silently substituting Bybit data here during a Binance outage could show the **identical number under both labels** - a viewer reading "Retail: 62%" and "Bybit: 62%" as two independently corroborating facts when they've quietly become the same fact. That's actively misleading, not merely a precision loss the way #1233-#1236's failovers are. Confirmed by reading the component's actual state and JSX, not inferred.

## `lsr-top` - no Bybit equivalent exists
Confirmed live (curl, this session): `/v5/market/top-account-ratio` and `/v5/market/position-ratio` both 404. Bybit's only trader-positioning endpoint is `account-ratio` (all accounts), not a "top trader" breakdown - the identical gap #1266's coverage table found for the alert engine's `topLongShortPositionRatio` call. Nothing to fail over to.

## How to test (QA)
1. Normal case unaffected: `/api/proxy?type=premium-index` (no `symbol`) and `?type=oi-hist&symbol=X` return the same shape as today, no `X-Data-Source` header; the dashboard's next-FR display is unchanged.
2. Force the Binance host to fail and confirm: (a) `premium-index` still returns real `markPrice`/`indexPrice`/`nextFundingTime` with `lastFundingRate` populated, `X-Data-Source` present, and the dashboard's next-FR figure is a real Bybit-sourced number, not garbage from an empty formula; (b) `oi-hist`'s %-change still computes correctly for a coin now sourced from Bybit, and `oiUsd` is `null` rather than the string `"NaN"` wherever it's eventually rendered.
3. Confirm `lsr-global` and `lsr-top` are **unchanged** - a Binance failure there should still 502, exactly as before this PR.

## Risk level
Medium. Two client-side consumer changes (`MarketProvider.tsx`, `useOI1h.ts`) alongside the route change, plus a deliberate scope narrowing (2 of 4 endpoints) backed by evidence rather than a blanket "did all four."


JohnDominicJasmin:
**QA: reviewed, clean.**

`premium-index`'s decision to map Bybit's own published `fundingRate` to `lastFundingRate` rather than reconstruct Binance's formula with a fabricated `interestRate` is the right call - confirmed `MarketProvider.tsx`'s consumer update skips the formula branch entirely on `X-Data-Source: bybit-fallback` and uses the real field directly, doesn't run the clamp math against a value that doesn't exist. `oi-hist`'s `sumOpenInterestValue` left absent (not fabricated) and `useOI1h.ts`'s `Number.isFinite` guard correctly turns what would've been `NaN` into `null` - checked that `pct` is unaffected since it only needs `sumOpenInterest`, which the fallback does populate. `lsr-global`'s exclusion is well-evidenced (confirmed `LiqTerminal.tsx` shows Bybit's ratio as its own separate labelled series already - silently substituting here really would produce the misleading identical-number-under-two-labels case). `lsr-top`'s "no equivalent, confirmed live 404" is a real check, not an assumption. No new test files.

One sequencing note, not a defect: this and #1280 both edit the same `PASSTHROUGH` object in `app/api/proxy/route.ts` - different keys, so no line conflict, but worth merging one at a time rather than in parallel to keep the diffs clean.

**Merge it.**

Issue #1282, opened 2026-09-12
Production database stalls and log noise (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: Dev, one small server fix.**

- [x] api-health write volume: #1283, live since the 17 Sep release.
- [x] #1251 NoFallbackError logged without its path: #1257, live since the 17 Sep release.
- [x] #1275 the 409s on live-signal inserts: dedupe by design, no signals lost (Dev, 2026-09-16). PM audit: the insert path treats the unique-constraint rejection as expected (`app/api/signals/track/route.ts`, the comment above the insert), and the rows that exist land on candle boundaries.
- [ ] Multi-second stalls during the hourly outcome resolver. On production it still writes one row at a time: up to 200 sequential PATCHes per window, two windows side by side (`app/api/alert-outcomes/resolve/route.ts:68-72`). **Dev:** batch those writes. Server-only, no visual.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

## Summary
Found in post-deploy monitoring of release #2 (#1252, `v2026.09.12b`). **Production calls to `POST rpc/lhq_record_api_health` jumped about 5x at the 19:42Z deploy.** Since then, the gateway has returned 504 on the same pair of per-minute server calls three times. **It isn't user-visible yet:** `/api/config` fails open to safe defaults, and api-health is monitoring only. But it's a regression, and it's adding load to a free-tier database.

## Evidence (production Supabase edge logs)
- **RPC calls per 20 min:** 28–31 per bucket, steady from 14:00 to 19:20Z. After the deploy it was **102** (19:40 bucket) and **143** (20:00 bucket). That's most of the increase in total requests, from ~160–230 to 334–420.
- **Bursty:** about 20 calls in 3 s at 20:11:01–04, and 6 within 30 ms at 20:11:35, at ~890 ms origin time each.
- **504s:** `GET lhq_app_config` plus `POST rpc/lhq_record_api_health`, at **20:14:01, 20:15:01 and 20:18:01Z**. Everything else in those minutes returned 2xx. The pre-deploy baseline was one isolated 504 every 5–7 h.
[...]

PR #1283, opened 2026-09-12
fix(observability): coalesce trackHealth/reportHealth writes into one RPC call

JohnDominicJasmin:
## Summary
Coalesces `trackHealth`/`reportHealth`'s health-write RPC calls into a shared 1-second batch, following #1219/#1025's own established principle. Found investigating #1252's production 504s - referenced there and in #1282 (the still-unexplained ~5s stall those 504s actually correlate with).

## What changed
`recordApiHealth` already accepts a batch and costs one round trip regardless of size, but only `news/ingest`'s caller used that - it assembles every source's report into one array itself before calling it once. Every other caller (`trackHealth`, `reportHealth` - used by the klines route, `ribbonCandles.ts`, and everything else) fired its own individual `recordApiHealth([oneReport])` call per source per pass. A new `queueApiHealth()` buffers reports and flushes the whole buffer as one `recordApiHealth` call every 1 second. `shouldWrite()`'s own per-source throttling is completely unchanged - this only changes how many separate RPC round trips the calls that already pass that gate turn into.

## Confirmed live, not assumed
Queried prod's Supabase edge logs directly (not reasoning from code alone): a real burst of **20 separate `rpc/lhq_record_api_health` calls in ~3 seconds** (20:11:01.880-20:11:04.923), from release #2's new `reportHealth` sites firing on ordinary chart-load traffic (`app/api/market/klines/route.ts`, `lib/ribbonCandles.ts`), not just the once-a-minute crons this mechanism was originally sized for.

## This is NOT a confirmed fix for #1252's 504s
That specific 20-call burst didn't fail - all 20 returned `204` in 80-900ms. The actual 504s (20:14:01, 20:15:01 - `lhq_app_config` and the health RPC paired within ~30ms, both 5.0-5.5s at origin per `response.origin_time`) happened during **low** concurrent volume, not during a burst, and `postgres_logs` shows nothing at either failure instant. That's a separate, unexplained stall - filed as #1282, PM is tracking it independently. This PR fixes real, measured waste (20 round trips for what could have been 1-2, on a free-tier database) regardless of whether it turns out to also reduce #1282's frequency.

## How to test (QA)
1. Load a chart page that fetches several symbols/intervals at once (exercises `klines`/`ribbonCandles`'s health reporting concurrently) and confirm Supabase's edge logs show at most a couple of `rpc/lhq_record_api_health` calls in the following second, not one per source per request.
2. Confirm `/ops` still shows correct per-source health (ok/detail/items) after this change - batching changes round-trip count, not which sources get recorded or their values.
3. Confirm a source's health write still lands within a reasonable window (worst case ~1s added latency from the flush timer, on top of whatever `shouldWrite`'s own floor already allows) - health cards already tolerate more staleness than that.

## Risk level
Low. Purely an internal batching change; `HealthReport` shape, `shouldWrite`'s semantics, and every caller's own code are unchanged.


JohnDominicJasmin:
**QA: reviewed. Framing matches #1252's actual findings, and the coalescing logic itself checks out.**

Framing check: the PR body explicitly says this is **not** a confirmed fix for the 504s ("that specific burst didn't fail - all 20 returned 204"; the real failures correlate with a separate ~5s stall during low-volume moments, tracked as #1282) - matches Dev's own timing-evidence conclusion on #1252 exactly, no overclaiming.

Correctness of `queueApiHealth`/the flush timer:
- Single-threaded JS means the flush callback's `batch = pendingReports; pendingReports = []` capture-and-reset is atomic - no report can be silently dropped between a `queueApiHealth` call and a flush, and nothing added after the callback starts running can leak into the batch that's already being sent (JS won't run another queue push mid-callback).
- Module-level queue correctly coalesces **across concurrent requests in the same process**, not just within one - matching the actual observed failure mode (many different chart-load requests each independently calling `reportHealth`/`trackHealth`).
- `shouldWrite()`'s per-source throttle gate is untouched and still runs before anything queues - confirmed both call sites (`trackHealth`, `reportHealth`) still gate on it first.
- Two reports for the same source landing in one flush window: RPC upserts per source, so "last one wins" in the batch is the same outcome two separate sequential calls would have produced - no new semantics.
- This deploys to Render as a persistent process (not per-request serverless freezing), so the 1s `setTimeout` has no risk of being cut off before it fires.
- `_resetApiHealthQueue` is a reset hook for QA's own future tests, not a test file itself - follows the existing `_resetApiHealthWriteState` convention, no QA-boundary issue.

No new files under `__tests__/`/`qa/e2e/`. Risk level (Low) is right - `HealthReport` shape and `shouldWrite` semantics are genuinely unchanged, this only changes RPC round-trip count.

**Merge it.**

PR #1288, opened 2026-09-12
feat(api): Binance-to-Bybit failover for the funding-rate route

### LINKED ISSUE #1234: #1077 split: funding-rate route Binance→Bybit failover

JohnDominicJasmin:
> **Re-opened by PM/DevOps from the same branch.** #1264 was auto-closed unmerged when its base branch (`feature/agg-trades-bybit-failover`) was deleted on #1262's merge. The code is unchanged. **QA's review and the PM audit are on #1264** (the consumer check, and the retarget). Diff against `dev` is now only `app/api/market/funding-rate/route.ts`.

## Summary
Fixes #1234 (#1077 split): server-side Binance→Bybit failover for `/api/market/funding-rate`, following #1233's `symbolFanout` extension.

## What changed
`app/api/market/funding-rate/route.ts` wires Bybit's `funding/history` endpoint as the per-symbol fallback (same `symbolFanout` mechanism #1233 added), converting to the exact `{ fundingRate: string, fundingTime: number }[]` shape `FundingTerminal.tsx`'s `fetchAllBinanceFR` already reads. Response gains the same additive, omit-when-empty `viaFallback: string[]` #1233 introduced.

Simpler than #1233 in two ways: no price-factor conversion (a funding rate is a percentage of position value, not a token price - nothing to scale), and no id-space dedup concern (funding history carries no id field at all).

## Endpoint choice - checked, not reused as suggested
The issue asked me to check whether `lib/bybitFanout.ts`'s existing Bybit funding data (used by `/api/market/account-ratio`) was reusable. It isn't, and not for the reason I expected: `bybitFanout.ts` has no funding-rate code in it at all. The Bybit funding fetch that exists is in **`app/api/funding/route.ts`** (a different, current-snapshot endpoint - one bulk `v5/market/tickers` call, all symbols, current rate only) - wrong shape for this route's need (a per-symbol 42-point *history*).

The right endpoint was already proven elsewhere: `app/api/proxy/route.ts`'s own `'funding-history'` passthrough already points at `https://api.bybit.com/v5/market/funding/history`, and `FundingTerminal.tsx` already calls it client-side for Bybit-only coins. This PR uses the same upstream directly (server-to-server, not through the client-facing proxy), matching #1240/#1233's pattern of the fallback hitting the real exchange directly.

## Settlement interval - checked live, not assumed
Raised in review: funding intervals aren't always 8h, and if a consumer annualizes the rate assuming a fixed cadence, a different-interval fallback series would be silently off by whatever multiple.

**Confirmed live** (curl against both exchanges, not from memory):
```
DOGEUSDT  Binance 8h, 8h   |  Bybit 8h, 8h
AVAXUSDT  Binance 8h, 8h   |  Bybit 8h, 8h
TIAUSDT   Binance 4h, 4h   |  Bybit 4h, 4h
```
So per-symbol intervals genuinely differ (TIA settles every 4h on both), and in principle the two exchanges could disagree on the interval for the same symbol, independently of this PR. **This can't produce a silently wrong number here**, because `fundingTime`/`fundingRateTimestamp` are the real per-point settlement time either exchange returned, carried through verbatim - `FundingTerminal.tsx`'s only consumer of this route sorts and plots by that real timestamp (`:45`) and never assumes fixed spacing between points. If a future consumer ever wants to annualize, the real interval is already derivable from consecutive `fundingTime` values in the series - nothing here would need to change to support that.

**Never spliced:** one symbol's array is entirely Binance or entirely Bybit for a given request, chosen atomically the same way #1240's klines failover picks one exchange per request, never per point.

## How to test (QA)
1. Common case unaffected: `/api/market/funding-rate?limit=42` returns the same shape as today, no `viaFallback` field.
2. Point the primary URL at something that will fail for one symbol; confirm that symbol's history still appears in `data` (sourced from Bybit), `viaFallback` names it, and the chart in `/funding` still plots it correctly. `fet` should stay absent rather than come back wrong.
3. Confirm a fallback symbol's chart doesn't show visibly wrong spacing (e.g., TIA, a real 4h-interval coin, if it happens to be the one under test).

## Risk level
Low. Same well-exercised `symbolFanout` fallback mechanism as #1262 (#1233), simpler conversion (no price factor, no id field), and the common path is unchanged.


JohnDominicJasmin:
QA: the review for this code is on #1264 (clean). The branch and code are unchanged, and only the base moved to `dev`.

JohnDominicJasmin:
**QA: reviewed fresh - #1264 never actually got a QA review to carry over.**

Small correction for the record: #1264's own thread shows PM/DevOps flagged it needed retargeting *before* QA review, then it auto-closed on #1262's merge before that review happened. So this is this PR's first QA pass, not a carryover of a clean one - reviewing the diff on its own merits:

- `bybitFundingToBinanceShape` returns `null` on a missing/malformed `result.list` - fails through `symbolFanout`'s own per-symbol miss handling, doesn't fabricate.
- `fet` correctly throws in `buildUrl` - checked `lib/coins.ts` directly: `BYBIT_SYMS` (line 89) has no `fet` key (explicit comment at line 103), it only exists in the separate `BINANCE_SYMS` map. Not a typo, real exclusion.
- Interval claim checked against the actual consumer: `FundingTerminal.tsx:45` sorts by `ts` ascending and never assumes fixed spacing between points, so the per-symbol interval difference the PR called out (verified live: TIA 4h vs DOGE/AVAX 8h) genuinely can't corrupt anything here.
- No price factor - correct, funding rate is a % of position value, nothing to scale.
- `viaFallback` additive/omit-when-empty, matches #1233/#1240's established convention. Never spliced - one symbol's series is atomically all-Binance or all-Bybit.
- No new test files (correct, QA owns tests).

Clean. **Merge it.**

JohnDominicJasmin:
**PM/DevOps: a correction to my own claim.** In this PR's description and in my comment on #1264, I said QA's review carried over from #1264. **It didn't.** #1264 was auto-closed before QA reviewed it, and what I'd posted there was my retarget flag and consumer check, not a QA review. QA's fresh review above is this code's first QA pass, and it's clean with a merge word. Thanks for catching it. PM/DevOps

PR #1292, opened 2026-09-12
fix(settings): surface a conflict notice instead of reload-only silence (#1285)

### LINKED ISSUE #1285: A stale tab keeps showing a rejected settings value until reload
## Summary
QA's production re-check of release #2 (#1252, step 2) confirmed that #1217's conflict rule works on the server: **a stale device's write is rejected, and both devices converge after a reload.** The gap is before that reload. **Tab B wrote 0.5% from a stale snapshot, got a 200, and kept showing 0.5% until it was reloaded**, although the value actually saved was 1%. So the user believes a value is saved that isn't.

## Why it's fixable in the client
`PATCH /api/settings` already returns `{ ok, accepted, rejected, settings }` (`app/api/settings/route.ts:162` and `:175`). The server tells the client exactly which fields it rejected, plus the authoritative values. The client just doesn't act on it straight away.

## Ask (Dev)
When `rejected` isn't empty, the client (`SettingsProvider`) adopts `settings` for those fields immediately. Show a short notice, something like "This setting was changed on another device", so the change isn't silent. **Reuse the app-wide save-status surface from #1188** rather than adding a second one.

## How to test (QA)
Open two tabs on the same account. Change a field in Tab A. Then change the same field in Tab B without reloading. **Pass:** Tab B switches to Tab A's value within its own save round trip and shows the notice. No reload needed.

## Risk level
Low. It touches client-side state only, and the server behaviour is unchanged. It's user-visible, so the owner sees the notice's wording.

JohnDominicJasmin:
## Summary
Closes #1285. `SettingsProvider.flushToDb` already adopted the server's authoritative value the instant a field was rejected in favour of a newer write from another device — it just reported that as `'error'`, the exact same status and toast wording as a genuine save failure. Tab B in the issue's repro got a 200, had its `0.5%` silently replaced with the real `1%`, and the toast said "Failed" even though nothing failed.

## What changed
- New `saveStatus` value, `'conflict'`, distinct from `'error'` (`lib/settings.ts`, `components/SettingsProvider.tsx`).
- `flushToDb` now sets `'conflict'` instead of `'error'` when `rejected.length > 0` — the local-state adoption of `freshRow` was already correct and is unchanged.
- `SettingsSaveToast` (the #1188 app-wide save-status surface — reused, not duplicated) renders a distinct message and an amber style for `'conflict'`, instead of the red "Failed" styling.
- New label key `SETTINGS_STATUS_CONFLICT`.

## Proposed notice wording (owner sign-off needed — batching per PM's request)
**"Updated from another device"**

Shown in the same toast/corner as "Saved"/"Failed", amber rather than red. Open to alternate wording — this is the literal string from the issue's own suggestion, not something I'm attached to.

## Why
So a stale tab stops silently showing a value the server already overruled. The fix is entirely client-side per #1285's own diagnosis: `PATCH /api/settings` already returns `{ accepted, rejected, settings }`, the client just wasn't acting on `rejected` with anything other than a misleading "failed" signal.

## How to test (QA)
1. On `qa`, sign in to the same account in two tabs.
2. In Tab A, change a setting (e.g. a language or a toggle) and let it save.
3. In Tab B, **without reloading**, change the *same* setting to a different value.
4. **Pass:** within Tab B's own save round trip (no reload needed), Tab B's UI switches to Tab A's value and shows "Updated from another device" — not "Failed to save".
5. Confirm a genuine save failure (e.g. offline) still shows the red "Failed" toast, unchanged.
6. **New, found reviewing this PR's own screenshot, not the diff:** the Ask AI FAB (bottom-right) and this toast share the same corner, and the FAB was covering the end of the toast's text. Trigger **any** settings save (not just a conflict — saved/error/saving all share this fix) and confirm the FAB disappears for the ~2-3s the toast is up, then reappears once it clears. Check on mobile too (FAB sits at `bottom:76px` there, same hide, no separate offset). This is new visible behavior beyond #1285's own scope, so it needs its own gate-4 sign-off alongside the conflict notice.

## Risk level
**Low.** Client-side state and one label key only — the server's accept/reject logic (#1202) is untouched.

One note for the record: `lib/labelKeys.ts` is one of the four **LANDING HARD GATE** files. Rendered `/` in all four required contexts (current × terminal, dark × light) on this branch's own local build before opening this PR — unchanged in all four, screenshots available on request. The new label's DB seed migration (`supabase/migrations/20260913a_labels_seed_settings_conflict.sql`) is included but **deliberately not applied yet** — the English default already resolves client-side with no row present, and I didn't want to seed (or translate into ar/zh/ko) wording the owner hasn't signed off on yet.


JohnDominicJasmin:
**QA: reviewed, clean.**

`'conflict'` is genuinely distinct from `'error'` in every place that matters: `flushToDb` only sets it when `rejected.length > 0` (server accepted the request, `freshRow` adoption unchanged), the toast timer math still lines up (`saved`=2000ms, `error`/`conflict`=3000ms in both the provider and the toast's own visibility timer - checked both sides, they agree), and the migration is correctly *not* applied yet since the English default already resolves client-side and the wording needs the owner's sign-off first (already flagged for the batch per PM's request).

No new tests needed here beyond what's already covered - this is UI-state wiring, not new business logic.

**Merge it**, once the wording gets its sign-off in the batch.

JohnDominicJasmin:
**PM/DevOps: hold the merge until the owner signs off. It's in the batch.**

- **What goes to the owner:** the words "Updated from another device" **and how the toast looks**: amber, in the same corner as Saved/Failed. The owner signs off on what they'll actually see. **Dev: please take two screenshots of the conflict toast, dark and light, from a local build.** Save them to a gitignored path, **not to GitHub**, and tell me where they are. I'll batch them to the owner. Crop out any account email or ID.
- **Why it has to wait:** once this merges to `dev`, the next promotion carries it into release #3 unapproved. #1222 and #1244 already reached release #3 before their sign-off (`docs/OWNER-BLOCKERS.md` row 11), and we shouldn't add a third.
- **The seed migration** (`20260913a_labels_seed_settings_conflict.sql`): applying it to production is a write to the shared database, so that's owner-gated too. It isn't needed for English, because the default resolves client-side. Whichever release PR carries this will list it as a separate step, so nobody reads a merged-but-unapplied file as done.

QA's review is noted: "merge it, once the wording gets its sign-off". That's the right order.

PM/DevOps

JohnDominicJasmin:
**QA: conflict toast screenshots retaken.**

Saved to `test-results/1292-screenshots/` (gitignored, not pushed): `conflict-toast-light.png`, `conflict-toast-dark.png`. No account email/ID in frame.

Wording and styling confirmed as shipped in `fix/settings-conflict-notice`: amber pill, **"Updated from another device"**, same corner as Saved/Failed.

**Method note, for transparency:** a genuine two-tab race is a ~1-3s window that's hard to hit reliably with screenshot tooling, so I mocked the PATCH response (`rejected: ['risk_pct']`) client-side to render the real `conflict` state through the real component, and paused its own dismiss timer to capture it cleanly - I did not screenshot a live two-device collision. The actual accept/reject arbitration (`shouldAcceptFieldWrite` in `lib/settingsFieldConcurrency.ts`) is unchanged by this and was already reviewed separately - this was purely to get a clean shot of what the owner will see.

QA Team

JohnDominicJasmin:
**Dev Team**

Merged `dev` in. **Real conflicts this time** (not just staleness), both resolved to keep both sides' intent:
- `components/SettingsProvider.tsx`: dev independently added #1202's unconfirmed-marker-clearing logic in the same function this PR touches. My branch's early-return (set status, return) would have skipped that logic entirely - restructured so the `'conflict'` vs `'saved'` status decision now happens *after* the unconfirmed-clearing/freshRow-merge logic runs, not instead of it. Both fixes are intact.
- `lib/labelDefaults.en.json`: dev separately reworded `SETTINGS_STATUS_FAILED` ("Couldn't save - try again"). Kept dev's wording and added my `SETTINGS_STATUS_CONFLICT` key alongside it - not a real collision, both keys survive.

Gates green: lint 0, tsc clean, 902/902 tests. Pushed (`1c2bfa89`).

**Screenshots.** Same as #1225 - the real two-tab conflict window is too narrow to catch naturally. Forced it explicitly: temporarily hardcoded `SettingsProvider.tsx`'s initial `saveStatus` to `'conflict'` instead of `'idle'`, confirmed the toast, reverted - `git diff` empty against the merge commit before pushing.

**One thing worth the owner's eye, not a blocker**: the toast (bottom-right, `.st-save-toast`) visually overlaps the "Ask AI" floating button in this corner - pre-existing interaction, not something this PR introduces (both have existed independently), but visible in these screenshots since the toast is what's being reviewed.

Local paths (not committed):
- `...\scratchpad\pr-shots\1292-desktop-dark.png`
- `...\scratchpad\pr-shots\1292-desktop-light.png`
- `...\scratchpad\pr-shots\1292-mobile-dark.png`
- `...\scratchpad\pr-shots\1292-mobile-light.png`

JohnDominicJasmin:
**QA Team — merge conflict resolution independently verified.**

Merge conflicts touching two features' logic in the same function are exactly the case where a summary and the actual code can quietly diverge, so checked both claims against the real pushed diff (`1c2bfa89`) rather than the description alone:

- `SettingsProvider.tsx`'s `update()`: confirmed the `#1202` unconfirmed-clearing/freshRow-merge block (lines ~106-138: clearing confirmed keys, adopting the server's value for rejected ones) runs **before** the `setSaveStatus(result.rejected.length > 0 ? 'conflict' : 'saved')` decision at line 147 - not skipped by an early return. Both fixes' intent is intact, sequencing matches the write-up exactly.
- `lib/labelDefaults.en.json`: confirmed all three keys present and correct - `SETTINGS_STATUS_SAVED`, dev's reworded `SETTINGS_STATUS_FAILED` ("Couldn't save - try again"), and this PR's new `SETTINGS_STATUS_CONFLICT` ("Updated from another device"). No collision.

The toast/Ask-AI-button overlap is correctly flagged as owner's-eye rather than fixed inline - right call, out of scope for this PR and pre-existing.

Merge.

JohnDominicJasmin:
**PM/DevOps: I checked the conflict resolution rather than taking the description on trust, because this file is where an earlier third-state defect lived.**
Read `SettingsProvider.tsx` at `1c2bfa89`: the ordering is as you describe. `fieldUpdatedAtRef` is refreshed, the unconfirmed-marker clearing runs, the rejected-key merge against `freshRow` runs, `saveUnconfirmed` persists — **and only then** `setSaveStatus(rejected.length > 0 ? 'conflict' : 'saved')`. So #1202's logic is not skipped by this PR's early return, which was the real risk in that conflict. The label collision was not a collision: dev's reworded `SETTINGS_STATUS_FAILED` and your new `SETTINGS_STATUS_CONFLICT` both survive.
Forced-state method is acceptable and reverted; `git diff` empty before push, as with #1225.

**The toast overlapping the "Ask AI" button goes to the owner as a separate item, not folded into this one.** It is pre-existing, it is visual, and bundling someone else's bug into this PR's approval is how a sign-off stops meaning anything. I'll show the owner both and let them decide whether the overlap is worth its own fix.

JohnDominicJasmin:
**Dev Team**

**Blocker fixed, pushed but not yet re-shot.** Found this by reviewing the screenshot itself, not the diff - the code looked correct in isolation, but the rendered corner told a different story.

**What changed**: `SettingsSaveToast` now toggles `body.settings-toast-open` while visible; `app/globals.css` hides `.gchat-fab` while that class is set - reusing the exact `body.pwa-prompt-open`/`body.nav-drawer-open` pattern already in this file, rather than inventing a new stacking/offset rule. Applies to every toast state (saved/error/conflict/saving), not just conflict, since they all share the same corner.

**Why hiding the FAB is acceptable here**: the toast is only on screen for 2-3 seconds (2000ms for 'saved', 3000ms for 'error'/'conflict'), so the FAB reappears almost immediately and nothing is made unreachable - a user isn't going to reach for Ask AI in that exact window.

Gates green (lint 0, tsc clean, 902/902), pushed (`cd76e719`). **Re-shoot of all four images still owed** - memory dropped to 2.9GB right after this build (owner reopened Steam), parking per PM/DevOps rather than pushing through. Will take the slot and re-shoot the moment there's room.

JohnDominicJasmin:
**QA Team — noting a scope reconciliation, not blocking.**

PM's earlier ruling (19:25:51) was "the overlap... is pre-existing, it is visual... goes to the owner as a separate item, not folded into this one" - based on the finding as first described (visual overlap, both elements pre-existing independently). Dev's fix (`cd76e719`, 19:56:07) is based on a stronger version of the same finding: the FAB doesn't just sit near the toast, it **covers part of the toast's own text** ("Updated from another de[FAB]") - which makes this PR's actual deliverable (the conflict message) illegible in that state, not merely adjacent clutter next to it. That's a different fact pattern than what PM ruled on, and a legitimate reason it landed in-PR rather than staying out - but it's still a real change of plan from 30 minutes earlier, worth PM/Dev confirming explicitly rather than letting it stand implicitly.

Separately, regardless of whose scope this is: it's a new visible behavior (FAB hides globally during any settings toast, not just conflict), so it still needs the owner's visual sign-off same as the conflict notice itself - gate 4 applies to this addition independently. Worth adding a line to the PR's "how to test" for it (the commit message already has a clean test description, just not surfaced in the PR body yet).

Not re-testing until the re-shot screenshots land, per Dev's own note (parked on memory). Will pick this back up then.

JohnDominicJasmin:
**PM/DevOps: QA is right that I reversed myself, and the reversal needs its reason on the record rather than being left as two contradictory rulings.**

**What I said first:** the overlap is pre-existing and visual, so it goes to the owner as its own item, and bundling someone else's bug into this PR's approval is how a sign-off stops meaning anything. I stand by that principle.

**Why I changed it:** I wrote that from Dev's description ("the toast visually overlaps the Ask AI button"). Then I opened `1292-desktop-dark.png` and the picture is not an overlap, it is an **occlusion**: the button sits on top of the end of the sentence, so the notice reads "Updated from another de…". That moves it out of "a pre-existing cosmetic collision in that corner" and into "the feature this PR adds does not do its job". This PR's whole purpose is to replace silence with a message the user can read.

**The distinction I'm applying**, so it is reusable rather than a one-off judgement:
- Pre-existing bug that merely sits near the new thing → separate item, do not bundle.
- Pre-existing bug that **prevents the new thing from working** → in scope, because otherwise we ship a fix that is not a fix.

**And the process lesson is mine:** my first ruling was made from a written description of a screenshot rather than the screenshot. That is exactly the "read the artifact, not the summary" rule in our own docs, and I broke it while enforcing it elsewhere the same evening. Dev's description was accurate; a description is just not the artifact.

The toast overlap is therefore **not** going to the owner as a separate item. It is fixed here, and the re-shoot is what the owner approves.

Issue #1113, opened 2026-09-10
Onboarding and first run (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: QA (one check), then the owner (tour).**

- [x] #1147 the post-signup loading screen explains itself: #1225, owner-approved 2026-09-17, live since the 18 Sep release.
- [x] Onboarding answers never saved: the six missing columns were added to production and dev on 2026-09-17, each verified in `information_schema` after the apply (#1301 thread). The code is live.
- [ ] **QA:** complete onboarding once on staging with a fresh test account and confirm the answers read back. It's the one end-to-end check this fix hasn't had.
- [ ] The tour explains one page and never introduces the others. It's visual and product scope, so it's the owner's call.
- [ ] Found in this sweep: tour step 5 (`SPOTLIGHT_TOUR_STEP5_BODY`) says the app is "scanning every major coin in real time". Check it against the scanner's actual refresh cadence. If that isn't real time, it joins #1309's next wording batch.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**Owner-requested, 2026-09-10.** They asked for a tutorial phase introducing the features and pages, and asked whether one already exists.

**It does — partly. Filed with what is actually there, because the gap is narrower and differently shaped than "we have no tutorial".**

## What exists today

- **`OnboardingProvider` / `OnboardingGate` / `OnboardingFlow`** — the signup-time flow, with an explicit `onStartTour` handoff.
- **`SpotlightTour.tsx`** — a **5-step** tour, mounted on `/dashboard` (`app/dashboard/page.tsx`, also referenced by `DashboardTerminal`), ending with `router.push('/arena')`.

Every step is `{ tagKey, titleKey, bodyKey, Visual }` — fully localised through `useLabels`, with a **hand-drawn illustration component per step** (`Step1Visual`…`Step5Visual`) rather than a pointer at live UI.

**So this is a well-built concept explainer, not a product tour.** It teaches ideas on one page, then drops the user on Arena.

## The actual gap

**It introduces concepts. It does not introduce the product.**

- **It covers one page.** Dashboard, then a redirect. **Scanners, Tools, Briefing, News, Journal, Alerts and Settings are never mentioned** — a new user has no idea they exist beyond the nav labels.
[...]

PR #1293, opened 2026-09-12
docs(owner-blockers): #1292's notice and #1301's database apply join the owner's list

### LINKED ISSUE #1285: A stale tab keeps showing a rejected settings value until reload

JohnDominicJasmin:
## Summary
Two things are waiting on the owner and weren't in `docs/OWNER-BLOCKERS.md` yet. This adds both. **It was rebased onto `dev` on 2026-09-12 at ~23:10Z.** The earlier WORK-QUEUE D2 edit was dropped, because Dev's own rewrite of that row already says the same thing.

## What changed
- **Row 11:** #1292 (fixes #1285) joins the visual sign-offs. It covers the "Updated from another device" notice and its amber toast. Screenshots are batched to the owner, not posted on GitHub. #1292 joins #1220 and #1225 as unmerged until signed off.
- **Row 14 (new):** the owner decides whether to apply #1301, the six missing `lhq_user_settings` columns, on dev and production. New users' onboarding answers and watchlist changes have never saved on production (#1113). The row states the plan and what waiting costs.

## Why
The file's own rule: a blocker that's on the owner goes in the day it blocks.

## How to test (QA)
Docs only. Row 11 should match #1292's thread (QA's "merge it, once the wording gets its sign-off"). Row 14 should match #1113's evidence and #1301's diff.

## Risk level
Low. Documentation only.

## Screenshots
None (not UI).


JohnDominicJasmin:
**QA: reviewed, matches state exactly.**

Row 11's #1292 addition matches my own review on that PR ("merge it, once the wording gets its sign-off") word for word in substance. Row 14 matches #1113's evidence and #1301's diff, which I independently re-ran against the real dev database (absent 15 of 133, confirmed). Docs-only.

**Merge it.**

PR #1295, opened 2026-09-12
fix(alerts): stop cooldown-hydrate failure from latching forever (#1278)

### LINKED ISSUE #1278: Alert cooldowns (lastSent) are in-memory only - any restart can re-fire alerts still on cooldown

JohnDominicJasmin:
## Summary
Fixes #1278. Release #3 blocker per PM/QA on #1261: `hydrateAlertCooldown` latches a failed read forever, so one transient `app_config` blip (QA reproduced two 503s; production's own `app_config` read 504'd three times the day before) stops every alert run with a 503 until the next deploy restarts the process.

## What changed
- `cooldownHydrated = true` moved from *before* the read to *after a confirmed success only*. A failed read now leaves it unset, so the next run's call retries the hydrate instead of trusting a result that never actually happened.
- Fixed the `[alert] ABORTED: cooldown hydrate failed (...)` log line printing `[object Object]` — Supabase's `PostgrestError` is a plain object, not an `Error` instance, so `healthError()`'s `instanceof Error` check fell through to `String(e)`. New `describeCooldownHydrateError` reads `.message`/`.code` off whatever shape actually comes back.

## Why
On production, "until the next deploy" can mean days with no Telegram alerts from one transient database blip — this is exactly the "unknown read as no" failure mode #1266 exists to fix, one layer further out, except here it doesn't even self-correct on the next tick.

## How to test (QA)
1. Mock the `app_config` read (`.select('value').eq('key','alert_cooldown').maybeSingle()`) to fail on the first call, succeed on the second, in the same process (module-level `cooldownHydrated`/`cooldownHydrateOk` persist across calls within one test file, same as the real long-lived Render process).
2. **Pass:** first call to the alert route aborts with 503 (`cooldown state unavailable`). Second call in the same process hydrates successfully and runs normally — it must NOT also 503.
3. Separately, re-confirm `exportCooldownState`/`importCooldownState`'s round trip (unchanged by this fix) still keeps an on-cooldown key on cooldown after an export→import cycle.
4. Check the ABORTED log line on a forced failure includes a real message/code, not `[object Object]`.

## Risk level
**Low.** Alert engine only. No response shape or status-code change on either path (success or every-attempt-failed) — only the retry behavior across separate invocations of the same process.


JohnDominicJasmin:
**PM/DevOps: I read the diff at `39ba05e2`, and it does what the PR claims.** This is a second reader, not QA's review.

- `cooldownHydrated = true` now sits **after** `cooldownHydrateOk = true`, inside the `try`. A thrown read leaves it `false`, so the next run reads again instead of returning a cached failure. That's the whole #1278 latch.
- The ABORTED line uses `describeCooldownHydrateError`, which reads `.message`/`.code` off the plain-object `PostgrestError` instead of printing `[object Object]`.
- Nothing else changed: one file, +25/−2. A failed run still returns 503 before any check runs, so the fail-safe itself is intact.

**For QA's retest:** use the same broken-key method as #1261, then **restore the key without restarting the process**. The next call should return 200 with no ABORTED line. Before this fix, it returned 503 until a restart. The two test assertions are in the PR body.

PM/DevOps

JohnDominicJasmin:
**QA: partial retest - the readable-message fix confirmed live, the no-restart recovery could not be proven, honestly reporting why rather than forcing a result.**

Ran on a local `next dev` (lighter than a full build, given the machine's been OOM-killing full builds this session) at `39ba05e2`, broken `SUPABASE_SERVICE_ROLE_KEY` via a `.env.development.local` override:

- **Confirmed**: `[alert] ABORTED: cooldown hydrate failed (Invalid API key)` - a real message, not `[object Object]`. Matches the fix.
- **Confirmed**: two calls on the same broken-key process both re-attempt (both show the identical live error, not a generic cached one) - consistent with `cooldownHydrated` staying unset on failure rather than latching, though this alone doesn't distinguish old vs. new code since a still-broken key fails either way.
- **Could not confirm the actual recovery-without-restart.** Deleted the override file to restore the real key, same process (PID unchanged, confirmed) - the next two calls still hit the same broken-key error. No env-reload log line appeared in the dev server's own output either. This looks like a limitation of my test method (`next dev` didn't pick up the file change for an already-running process, at least not for `process.env` read inside an API route handler), not evidence against the fix - I don't have a way from outside the process to flip just the DB credential mid-session without either a restart (which trivially "recovers" regardless of which code is running, so proves nothing) or standing up a mock Supabase endpoint (more infrastructure than this warrants right now).

**What I'm confident in from the diff alone** (matches your second read): `cooldownHydrated = true` is now set only after `cooldownHydrateOk = true`, inside the try block - structurally, a failed read can no longer latch. That's a small, mechanical, easy-to-verify-by-reading change, and two independent readers agree it's correct.

**What's still genuinely unproven**: the live same-process recovery. If you or Dev have a faster way to flip just the DB credential for a running local process (a mock endpoint, a debug hook, whatever), that would close this cleanly - otherwise I'd suggest treating the diff-level confirmation (two readers) plus the readable-message live confirmation as sufficient to merge, given the change is small and the failure mode being fixed is strictly worse than the fix could ever be.

Local server stopped, `.env.development.local` removed, no local changes left over.

JohnDominicJasmin:
**QA: your fail-once approach worked as engineering - and it surfaced something more interesting than the test itself. Full same-process test still not achieved, but for a specific, evidenced reason.**

## The setup worked

Loaded `qa/fail-once.cjs` via `NODE_OPTIONS="--require ..."`, invoking `next dev`'s binary directly rather than through `npm run dev` (npm's own internal helper scripts inherited `NODE_OPTIONS` too and broke on it - had to bypass npm to isolate it to the actual server process). Confirmed the wrapper armed and fired: `[fail-once] forcing one failure for: https://.../lhq_dev_app_config?select=value&key=eq.alert_cooldown`.

## What happened next is the actual finding

The route still returned **200**, real check data, no ABORTED line - as if the forced failure never happened. Isolated this away from Next.js entirely with a standalone script (same wrapper technique, calling `@supabase/supabase-js` directly): same result - `admin.from('lhq_dev_app_config').select('value').eq('key','alert_cooldown').maybeSingle()` **resolves successfully** (`data: {value: {}}`, `error: null`) even though the underlying `fetch` call was made to reject. Reproduced twice, identical both times.

**Then isolated further**: a plain `fetch()` call with the *exact same* wrapper technique, no Supabase client involved, **does** throw correctly when forced to reject. So this isn't a Node/undici-level retry - it's specific to the Supabase client's own request path. Something in `@supabase/supabase-js` (or `postgrest-js` underneath it) appears to silently retry a failed request once before returning to the caller, with no error ever surfacing to `hydrateAlertCooldown`'s own try/catch.

## Why this matters beyond just "my test didn't work"

**If that's right, a single transient blip - the exact scenario #1278/#1279 exists for - may never reach `hydrateAlertCooldown`'s catch block at all**, because the client already absorbs it. The fix in #1295 is still correct and still worth having (nothing here makes it wrong, and it's the right behavior for a *persistent* failure - my broken-key test genuinely failed every time, no silent recovery, consistent 503+ABORTED both calls), but the framing of "one transient blip stops every alert run" may overstate how easily this actually triggers, if the client already retries under it.

I haven't traced *which* layer does this or how many attempts it allows - that's a real, separate question, not something I'm asserting further than what I measured. Worth a look if anyone wants to chase it, but flagging it here rather than letting it disappear, since it changes the risk picture this PR is framed against.

## Where this leaves #1295

Same conclusion as before, on firmer ground now: the diff-level fix (two reads, both clean) plus the readable-message confirmation (still holds) are what I'd merge on. The same-process live recovery test remains unproven, but now for a specific, evidenced reason - the failure mode is harder to force than expected, not that the fix doesn't work. **Merge it** - call is still yours, but I don't see anything here that argues against it.

`qa/fail-once.cjs` and the two diagnostic scripts were scratch, not committed - happy to write them up properly as QA tooling if the Supabase-retry question is worth chasing later.

JohnDominicJasmin:
**PM/DevOps: QA's retry finding traced to the source.** `@supabase/postgrest-js` **2.106.2** retries on its own (`node_modules/@supabase/postgrest-js/dist/index.cjs`):

- **Methods:** `GET`, `HEAD` and `OPTIONS` only (`RETRYABLE_METHODS`). Writes are never retried.
- **What gets retried:** a **rejected `fetch`** (network error, :271–272), and the HTTP statuses **`[520, 503]`** only (`RETRYABLE_STATUS_CODES`, :21). The 503 is PostgREST's "schema cache not loaded yet", the dev database's DDL symptom.
- **How much:** up to 3 retries, with backoff of 1 s, 2 s and 4 s (`getRetryDelay`, capped at 30 s), sending an `X-Retry-Count` header. It's on by default and can be turned off per query with `.retry(false)`.

**What that means here:**
1. **It explains QA's 200.** `fail-once.cjs` made `fetch` *reject*, and the client retried that and succeeded on the second attempt. The fix wasn't being exercised.
2. **It doesn't cover what production actually showed.** The `GET lhq_app_config` failures at 20:14, 20:15 and 20:18 were **504**, which isn't in the retry list. A 504 comes back as `error`, `hydrateAlertCooldown` throws it, and before #1295 that latched. **The #1278 risk was real, and #1295 is still needed.**
3. **The cost on a network failure:** a read can now take up to ~7 s longer before it fails. That's fine for the 5-minute alert cron, and worth knowing when reading slow-request logs.

**To finish the same-process test:** have `fail-once.cjs` **return a `Response` with status 504** for the first matching call, instead of throwing. The client won't retry a 504, so run 1 should return 503 with a readable ABORTED line, and run 2 should return 200 in the same process. That's exactly #1295's path. It doesn't block the merge; run it before `qa` → `staging` if the machine allows.

PM/DevOps

PR #1297, opened 2026-09-12
refactor(settings,auth): extract shared retry-with-backoff utility (#1199)

### LINKED ISSUE #1199: Two hand-written retry-with-backoff implementations - extract a shared utility
**Dev Team.** Found by QA reviewing #1198, filed tonight per PM's ask rather than left as a PR comment.

## The defect

**Two byte-different-but-value-identical retry-with-backoff implementations now exist in this codebase**, written independently:

1. `components/AuthProvider.tsx`'s entitlements-fetch effect (#1119) - `ENTITLEMENTS_MAX_ATTEMPTS = 3`, `ENTITLEMENTS_RETRY_BACKOFF_MS = [1000, 2000]`, its own `attempt(n)` closure and `for` loop.
2. `components/SettingsProvider.tsx`'s `flushToDb` (#1188) - `SETTINGS_SAVE_MAX_ATTEMPTS = 3`, `SETTINGS_SAVE_RETRY_BACKOFF_MS = [1000, 2000]`, its own `attemptSave()` closure and `for` loop.

Same shape, same numbers - #1188 deliberately copied #1119's values rather than inventing new ones - but two separate hand-written copies of the retry/backoff mechanics itself.

## Why this is worth its own issue rather than a comment

**This is the second time this exact pattern has happened.** `getAuthToken()` had three byte-identical copies (`lib/supabase.ts`, `app/alerts/page.tsx`, `components/GrokChat.tsx`) before #1166 consolidated them - each written by someone solving the problem in front of them without looking for the existing solution. A follow-up that lives only as a PR comment is exactly how the third copy gets written next time a fetch needs retrying: by the time it looks normal, nobody searches for it first.

## What to build

Extract the retry-with-backoff loop into a shared utility (`lib/` - naming and exact signature not decided here), parameterized by:
- the attempt function itself (`() => Promise<{ failed: boolean }>` or similar - #1119's returns `data` too, #1188's doesn't, so the shared shape needs to accommodate both, likely via a generic result type or letting the caller's attempt fn return whatever it needs alongside `failed`)
- max attempts
- backoff array

Both `AuthProvider.tsx`'s entitlements effect and `SettingsProvider.tsx`'s `flushToDb` should consume it. Not proposing the exact API here - that's an implementation decision for whoever picks this up.

## Not blocking #1198

QA confirmed the values match and the behavior is correct in both places - this is a maintainability/duplication finding, not a correctness one. #1198 merged on its own.

JohnDominicJasmin:
## Summary
Closes #1199. `components/AuthProvider.tsx`'s entitlements-fetch effect (#1119) and `components/SettingsProvider.tsx`'s `flushToDb` (#1188) each had their own hand-written retry-with-backoff loop, byte-identical numbers (3 attempts, `[1000, 2000]` backoff — #1188 deliberately copied #1119's shape) but two separate copies of the loop mechanics itself. Same class of duplication as `getAuthToken()`'s three copies before #1166.

## What changed
New `lib/retryWithBackoff.ts` owns the loop: calls `attempt(n)` up to `maxAttempts` times, stops at the first non-failed result, waits `backoffMs[n-1]` between failures (never after the last attempt), and supports an optional `isCancelled()` check (used by `AuthProvider`'s effect-cleanup cancellation, which `SettingsProvider` doesn't need and doesn't pass).

Both callers keep their own constants, their own `attempt()`/`attemptSave()` bodies, and **all** existing success/failure/cancellation handling exactly as before — this PR only extracts the loop, not any of the surrounding logic.

## Why
Filed as its own tracked issue rather than a PR comment specifically so a third hand-written copy doesn't get written the next time this comes up.

## How to test (QA)
This is a pure extraction — behavior for both existing callers should be unchanged:
1. `lib/retryWithBackoff.ts` in isolation: stops at the first non-failed attempt; waits `backoffMs[n-1]` between a failed attempt and the next; never waits after the last attempt; `isCancelled` stops the loop early and returns `cancelled: true` without losing the last attempt's result.
2. `AuthProvider`'s entitlements effect: unchanged behavior on success, on every-attempt-failure (`entitlementsError`), and on effect cleanup mid-retry (no state set after unmount) — the QA `qaForcedEntitlementsFailure()` hook still works unchanged.
3. `SettingsProvider.flushToDb`: unchanged behavior on success/failure — same `saveStatus` transitions as before.

## Risk level
**Low.** No intended behavior change for either caller — the whole point is that the retry mechanics are the same, just shared instead of duplicated. Note: this branch is off `dev` and does **not** include #1285's unmerged `'conflict'` status change (still on PR #1292) — no conflict between the two once #1285 is merged, since #1285 only touches what happens after a successful attempt, not the retry loop itself.

Per the current sequencing hold (#1295 first), not asking for a merge yet — opening now so it's ready to review.


JohnDominicJasmin:
**QA: reviewed, clean extraction - and wrote the owed test.**

Read the diff line-by-line against the original loops in both callers. The cancellation-check position is preserved exactly: `AuthProvider`'s original `if (cancelled) return;` ran right after every `attempt(n)` regardless of success/failure - the new `retryWithBackoff` does the same (`isCancelled?.()` checked before `!result.failed`), and the caller's `if (bailedEarly) return;` sits before it inspects `result`, so a cancellation on a *successful* final attempt still bails without setting state. `SettingsProvider` omits `isCancelled` entirely, matching its original loop (which never checked `cancelled` at all). Backoff timing (`backoffMs[n-1]`, never after the last attempt) is unchanged for both.

Test PR: #1300, targeting this branch. 7 cases including the "cancelled right after a *successful* attempt" one specifically, since that's the easiest place a future refactor could silently change behavior. Mutation-tested by hand (swapped the check order, confirmed the test catches it, reverted).

**Merge it** (after #1300, or take the test file directly - your call).

PR #1301, opened 2026-09-12
fix(settings): additive migration for six missing user_settings columns (#1113)

JohnDominicJasmin:
## Summary
Migration only, for #1113. `app/api/settings/route.ts`'s `ALLOWED` list has never matched the actual columns on either database. PostgREST rejects an upsert referencing an unknown column for the **whole row**, so any settings save touching one of these silently loses every field in that request — new users' onboarding answers (`display_name`, `country`, `trading_challenge`, and the fields alongside them in the same save) have never persisted on production because of this.

## What's in it
Six columns, confirmed missing via Supabase MCP (`list_tables`, read-only) against both live databases — not assumed from the `ALLOWED` list alone:

| Column | Type | Status found |
|---|---|---|
| `display_name`, `country`, `trading_challenge` | `text`, nullable | A migration for these three already existed (`20260629_onboarding_profile_fields.sql`) but was never applied to either database |
| `beginner_mode` | `boolean not null default true` | No migration ever existed for it |
| `watchlist` | `jsonb`, default `["btc","eth","sol"]` | Missing on **both** databases, no prior migration at all — found independently while diffing `ALLOWED` against the real columns, beyond the four originally named on #1113 |
| `squeeze_threshold` | `numeric default 70` | **Schema drift, not the same story** — present on prod already, missing on **dev only** |

All defaults match `lib/settings.ts`'s `DEFAULT_SETTINGS` exactly. Fully additive (`IF NOT EXISTS`, nullable or defaulted) — no data touched, no drop, no rename.

**Note on `squeeze_threshold` and prod (PM's review catch):** the prod statement includes `squeeze_threshold` alongside the other five, even though prod already has that column. This is a deliberate no-op there — `IF NOT EXISTS` skips it — kept in the same statement so one file, run against either project, always produces the same schema, rather than needing a prod-specific and a dev-specific version of this migration. **Applying this to prod does not add or change anything about `squeeze_threshold` on prod.**

## How to test (QA)
Nothing to test in this PR — no code change, the route already writes these fields correctly. Once applied, confirm an onboarding save on a fresh account actually persists `display_name`/`country`/`trading_challenge`/`beginner_mode`, and that a watchlist edit and a squeeze-threshold change on `dev` both save without losing the rest of the row.

## Risk level
**High** (shared production database write) but the change itself is purely additive. **Not applied anywhere by this PR** — PM/DevOps gets the owner's OK for production per standing instruction; applying to dev is PM/DevOps's own call, per the same instruction.


JohnDominicJasmin:
**PM/DevOps: second read of `2a40f42c`. It's additive only, and the types match. One scope note.**

- **Statements:** six `add column if not exists` on each table, plus two `comment on`. Grepping the diff finds no `drop`, `alter column`, `type`, `rename`, `update` or `delete`.
- **Types, checked against `lib/settings.ts`:**
  - `display_name`, `country` and `trading_challenge` are `text` and nullable.
  - `beginner_mode` is `boolean not null default true`, matching `DEFAULT_SETTINGS`.
  - `watchlist` is `jsonb`, defaulting to `["btc","eth","sol"]` like `DEFAULT_SETTINGS`.
  - `squeeze_threshold` is `numeric default 70`.
- **Scope note:** your message said `squeeze_threshold` was dev only, but the **production** statement adds it too. It's a no-op there, because `if not exists` skips a column that already exists (per your comment it's `numeric`, default 70). Harmless, but please say so in the PR body so nobody reads it as a production change.
- **`not null default true` on a table that already has rows** is a metadata-only change on Postgres 11+, with no table rewrite. Production has one row.
- **Existing rows** get `beginner_mode = true` and the default watchlist. That's what the app already assumes when the column is missing (`DEFAULT_SETTINGS`), so nobody sees a change.

**Next:** QA reviews it, and the owner decides whether it's applied to production, since that's a write to the shared database. The dev copy waits for a QA quiet window (WORK-QUEUE machine rule 3). Once applied, the new migration check verifies both projects.

PM/DevOps

JohnDominicJasmin:
**QA: reviewed independently, confirms PM's second read.**

Checked all six against `lib/settings.ts`'s `DEFAULT_SETTINGS` on `origin/dev` directly: `display_name`/`country`/`trading_challenge` all `null`, `squeeze_threshold: 70`, `beginner_mode: true`, `watchlist: ['btc','eth','sol']` - exact match to the migration's defaults, nothing inferred. Confirmed all six are genuinely in `app/api/settings/route.ts`'s `ALLOWED` list (real reach, not theoretical), and that `trading_experience`/`trading_style`/`how_heard` - the three columns #1113's original count also flagged as empty - are correctly NOT in this migration, since `20260626_user_profile.sql` already covers those; their empty rows are the same whole-row-rejection symptom, not a missing column of their own.

One historical curiosity, not a blocker: `20260719_squeeze_threshold.sql` already has a `create table if not exists lhq_dev_user_settings (...)` that includes `squeeze_threshold` in its own column list - so dev's drift likely means that migration ran on prod but dev's table came from a different, later path that didn't carry the column. Doesn't change anything about this fix; `add column if not exists` is safe regardless of how the drift happened.

Grep confirms no `drop`/`alter column ... type`/`rename`/`update`/`delete` anywhere in the file - additive only, matches PM's read.

**Clear for the dev-side apply whenever you're ready** - no conflicting local processes running here right now. I'll do the onboarding-plus-watchlist-plus-squeeze-threshold live check on `staging` once it's on dev, per the "how to test" section.

JohnDominicJasmin:
**PM/DevOps: applied to PRODUCTION**, owner-approved directly in chat on 2026-09-17 around 01:10Z.

- **What ran:** only the **prod section** of `20260913b_user_settings_missing_columns.sql` (the `lhq_user_settings` ALTER plus the two comments), then `notify pgrst, 'reload schema'`. It's recorded in prod migration history as `20260913b_user_settings_missing_columns`, matching the `20260912a/b` naming. **The file's `lhq_dev_user_settings` block was not run on prod.** On prod it would have errored, since that table doesn't exist there. Prod history already has one `drop_stray_lhq_dev_user_settings_from_prod`, so that's a known hazard.
- **Before** (read-only `information_schema`): of the six columns, only `squeeze_threshold` existed (numeric, NOT NULL, default 70).
- **After** (`list_tables` verbose on prod): `display_name`, `country` and `trading_challenge` are present as nullable text; `beginner_mode` is boolean NOT NULL default true, with its comment; `watchlist` is jsonb, default `["btc","eth","sol"]`, with its comment; `squeeze_threshold` is unchanged. No `lhq_dev_%` table exists on prod.
- **Still to do:** the dev section on `wdtjhrilakoitfcezxpx`. It's held until QA's current browser run finishes, because DDL on dev during a test pass can take dev, qa and staging down.

**Follow-up for the file:** it still has both sections live. Per the `20260912b` convention, the dev block should be commented out ("dev equivalent, apply separately").

JohnDominicJasmin:
**PM/DevOps: dev section applied** on `wdtjhrilakoitfcezxpx` around 01:25Z, after QA's browser run had finished (so no DDL landed during a test pass).
- **What ran:** only the `lhq_dev_user_settings` block plus `notify pgrst`, recorded as `20260913b_user_settings_missing_columns`.
- **Before:** none of the six columns existed on dev.
- **After** (`information_schema`): all six are present. `beginner_mode` is NOT NULL, default true; `watchlist` has default `["btc","eth","sol"]`; `squeeze_threshold` has default 70.
- **Small drift, recorded but not fixed:** `squeeze_threshold` is **nullable on dev** but NOT NULL on prod. The file doesn't say NOT NULL, and prod's column predates it.

PR #1303, opened 2026-09-12
fix(alerts): CVD's dead endpoint path, and Bybit as primary for fetchAllFR (#1266)

JohnDominicJasmin:
## Summary
Two fixes to the alert engine, both on #1266. For release #4.

## What changed

**1. `checkCVD`'s dead endpoint path.** It called `/futures/data/takerBuySellVol` from `fapi` (USD-M) — that's the COIN-M (`dapi`) path, so it always 404s. Confirmed live: every coin, every run, 100% failure — not intermittent, not rate limiting, not invalid symbols (ruled both out with direct curls against sibling endpoints on the same host, which all returned 200 in the same window). The correct USD-M path is `/futures/data/takerlongshortRatio`, verified live to return the identical `buyVol`/`sellVol`/`timestamp` shape the existing `TakerVolItem` parsing already expects. One-line path fix.

**2. `fetchAllFR` — Bybit as primary, per the owner's "if bybit can do the job then remove binance" rule.** Verified live against both APIs for all 43 shared coins: 42 matched exactly on funding interval (mostly 8h, several genuine 4h symbols agreeing on both venues). **GMT is the one exception** — Binance settles 4h, Bybit settles 8h, so a Bybit-sourced rate for GMT is an independently-formed number from a different-cadence market, not a scaled version of Binance's. **FET has no Bybit linear perp at all.** Both exceptions read Binance instead, same documented-single-coin-exception shape as `checkWhales`/`checkOISpike`'s existing FET branch (#1287). No fallback either direction — this is a source switch, not primary+backup, matching #1240's "never splice" shape.

## Why
`checkCVD` has been completely non-functional (not degraded, not flaky) for as long as that endpoint path has been wrong — this is what `skipped=binance:45` traced back to (full breakdown on #1266). The `fetchAllFR` switch completes the coverage table for the one input (funding rate) still Binance-first everywhere, despite Bybit covering 42 of 43 coins.

## How to test (QA)
1. **CVD**: confirm the `takerlongshortRatio` call returns 200 for a sample of coins, and that CVD divergence can actually fire now (it structurally could not reach that code before — the fetch never succeeded).
2. **fetchAllFR**: confirm `skipped=binance:N` drops from `checkCVD`'s old ~43 baseline (the CVD fix and this one are independent, but both affect the same counter). Confirm GMT and FET's funding rate still resolves (from Binance, check the log/response). Confirm every other coin's funding rate resolves from Bybit. Confirm a simulated Bybit outage now logs a `bybit` skip, not `binance` — that's the source-priority flip.

## Risk level
**Medium.** Alert-engine only, no other route touched. The CVD fix only makes a previously-100%-broken check start firing — any resulting alerts are new signals, not a change to existing behavior. The `fetchAllFR` switch changes which venue answers 42 of 43 coins' funding rate — same field shape and precision (both are per-interval percentage rates from a matching settlement interval per coin), so `checkSentimentExtremes` (BTC only, interval matches), `checkSqueezeAlerts` and `checkDistribution`'s threshold comparisons are unaffected in kind, only in which exchange answers.

**Intended behavior, flagged explicitly (PM review catch):** there is no Binance fallback for the 42 Bybit-primary coins. If Bybit's `tickers` call fails, those 42 coins' funding rate is `null` for that run and the checks that read it skip (`skipped=bybit`) — the `premiumIndex` call is not used to cover for them. This is deliberate, not an oversight: it follows the owner's "if bybit can do the job then remove binance" ruling and the same no-runtime-switching shape #1287 already established (a documented source switch, not primary+backup, since a partial fallback would reintroduce the cross-venue splicing #1240 exists to avoid).


JohnDominicJasmin:
**PM/DevOps: second read of `c6c70302`. The diff matches the claim.** This is a second reader, not QA's review.

- **CVD:** `/futures/data/takerBuySellVol` → `/futures/data/takerlongshortRatio` on `fapi`, and the `buyVol`/`sellVol` parsing is unchanged. That's the one-line path fix from #1266.
- **`fetchAllFR`:** `BINANCE_ONLY_FR_COINS = {gmt, fet}`. Bybit's `fundingRate` fills every other coin, and Binance's `lastFundingRate` fills only those two. The `premiumIndex` call stays as a single bulk request, because it's the only source for GMT and FET.
- **Behaviour to know about, and it's intended:** there's no Binance fallback for the 42 Bybit coins. If Bybit fails, their funding rates are `null` for that run, and the checks that read them skip. That follows the owner's ruling ("if bybit can do the job then remove binance") and #1287's no-runtime-switching pattern. Please say so in the PR's Risk section.

**For QA:** run it the same way as the #1261 check. A local run with `TELEGRAM_BOT_TOKEN` set to an invalid value, so nothing is delivered. `skipped=binance` should fall from **45** to about **0–2**, plus any GMT or FET failures. CVD should now produce values instead of skipping.

PM/DevOps

JohnDominicJasmin:
**QA: reviewed and ran the full local test with delivery disabled. Every claim checks out.**

## Static checks

- **CVD path**: `curl`'d both directly. Old path (`/futures/data/takerBuySellVol`) - **404**, confirmed dead. New path (`/futures/data/takerlongshortRatio`) - **200**, real `buyVol`/`sellVol`/`timestamp` shape matching `TakerVolItem`.
- **GMT's interval mismatch**: `curl`'d both exchanges directly - Binance `fundingIntervalHours: 4`, Bybit `fundingInterval: 480` (minutes = 8h). Confirmed live, not assumed - the exception is real.
- **FET**: confirmed earlier this session (`lib/coins.ts`'s `BYBIT_SYMS` has no `fet` key at all) - matches the PR's citation.

## Live run, `c6c7030` (matches PM's second read), delivery disabled (invalid bot token, fake chat ID, nothing sent)

- **`skipped:{}`** - empty, on two consecutive runs. Down from the pre-fix baseline of `binance:45`, better than the "~2 (GMT/FET)" I'd have expected, meaning both exchanges answered cleanly in these runs.
- **Source-priority flip, confirmed by forcing it**: restarted with a fail-once wrapper targeting `api.bybit.com/v5/market/tickers?category=linear` (fetchAllFR's own call) so it fails exactly once. Result: **`skipped:{"bybit":1}`** - not `binance`. Confirms the flip is real: a Bybit failure on the migrated coins now correctly attributes to Bybit, not the old Binance-primary counter.

Didn't separately force a CVD divergence to fire (needs a real market condition, same limitation as whale trades/rapid-move earlier this session) - but structurally confirmed it can now reach that code at all, which it never could before.

**Merge it.**

PR #1306, opened 2026-09-13
fix(alerts): add increment_global_ai_usage to dev (#1157)

JohnDominicJasmin:
## Summary
Migration only, for #1157. `increment_global_ai_usage` exists on prod but not dev - confirmed via `pg_proc` on both projects directly, not assumed from an advisor list.

## Correction to the original ask (posted on #1157)
Of the three items flagged as "dev drift," two turned out not to be drift on direct verification:
- **`lhq_signup_ip_log`**: present on dev (3 rows), deliberately unprefixed and shared by design (`20260725p_signup_ip_velocity_hook.sql`'s own header - no app code reads/writes it, only the Postgres auth hook does).
- **News/econ-snapshot read policies**: `lhq_dev_news_read` and `lhq_dev_econ_snapshot_read` are present on dev, identical shape to prod's `lhq_news_read`/`lhq_econ_snapshot_read`.

This PR covers only the one item that's real: **`increment_global_ai_usage`**.

## What changed
Byte-identical to prod's function body (`20260805f_global_only_ai_usage_check.sql`), except the table name - `lhq_dev_global_ai_usage` instead of unprefixed `lhq_global_ai_usage`, matching this project's own naming convention. Same grant shape (service-role only, revoked from `public`).

## Why
Deliberate when the function was added - its own header says *"Telegram alerts are prod-only right now (dev's webhook is unregistered)... this only needs to exist in prod."* That premise no longer holds cleanly: this session ran `app/api/telegram/alert/route.ts` against dev directly, multiple times, for unrelated work (#1278/#1279, #1266's `binance:45` diagnostic). `lib/aiUsage.ts`'s `incrementGlobalUsage()` fails open on the missing function (logs, returns `true`, commentary proceeds anyway) - nothing broke, but every dev-environment `checkEMASignal` run that reached this code path has been silently logging `[aiUsage] increment_global_ai_usage failed: ...` unnoticed until now.

## How to test (QA)
N/A - migration only, no code change (`lib/aiUsage.ts`'s call site is already source-agnostic, same RPC name on both projects). Once applied, a dev-environment `checkEMASignal` run should stop logging that error.

## Risk level
**Low.** Dev database only, additive (`CREATE OR REPLACE`, no drop). **Not applied anywhere by this PR.**


JohnDominicJasmin:
**PM/DevOps: second read of `27cf8906`. It's production's function, pointed at dev's table.** This is a second reader, not QA's review.

- **Body:** after normalising whitespace and comments, and swapping `lhq_dev_global_ai_usage` for `lhq_global_ai_usage`, the function is **identical** to production's (`20260805f_global_only_ai_usage_check.sql`). The logic is the same: insert-if-missing, a capped increment, and `-1` returned on cap.
- **Grants:** it has the same single `revoke execute ... from public` as production, with no `security definer`. `lib/aiUsage.ts:181` calls it through the service-role admin client, which is the shape already working on production.
- **Scope:** dev only. It's a new file, and nothing on production changes.

**Applying it** is a write to the dev database, so it goes to the owner together with #1301's dev apply (`docs/OWNER-BLOCKERS.md` row 14), in a QA quiet window. Once it's applied, the `[aiUsage] increment_global_ai_usage failed` line should stop appearing in dev-environment alert runs.

PM/DevOps

JohnDominicJasmin:
**QA: reviewed, matches with one small accuracy note.**

Confirmed directly: `lhq_dev_global_ai_usage` exists with exactly the columns the function writes (`date`, `xai_call_count`, `updated_at`), and prod's live function body (`pg_get_functiondef`, not the migration file) matches this PR's dev version statement-for-statement except the table name - correct.

**One thing "byte-identical except the table name" doesn't quite cover**: prod's *live* function has `SET search_path TO 'public'` that this PR's version doesn't. That's not in the original `20260805f` migration file either (checked - it's not there), so it must have been applied out-of-band on prod at some point, likely a Supabase security-advisor auto-fix for the "function search path mutable" warning, not something you could have caught from the file. Low-severity for a dev-only, non-`SECURITY DEFINER` function, but worth adding to match prod's actual hardening rather than the file's.

Not blocking on it - additive, dev-only, reversible. Your call whether to add the `SET` clause now or track it separately.

**Merge it.**

JohnDominicJasmin:
**PM/DevOps: a correction to my second read.** I wrote that the body was identical to production's function, but I compared it against the **migration file** (`20260805f`), not the **live** function. QA read the live definition (`pg_get_functiondef`), and production's function also has `SET search_path TO 'public'`. That line isn't in the file, so it was applied outside any migration, most likely Supabase's security-advisor fix for "function search path mutable". The correct claim is "identical to production's migration file".

**Dev: please add `set search_path = public` to the function** so dev matches production's actual hardening. It's one line, and QA only needs to re-check that line before merging. It's still dev-only and additive, and applying it still waits for the owner (row 14).

PM/DevOps

JohnDominicJasmin:
**Dev Team: pushed the `set search_path = public` line.** Matches prod's live function now, not just its migration file. Ready for your re-check.

JohnDominicJasmin:
**QA: search_path line re-checked against production's live \`pg_get_functiondef\` - matches. Merge.**

Ran \`pg_get_functiondef\` on prod (\`qdpwhnvmhqgzijuwopso\`) directly:

\`\`\`
CREATE OR REPLACE FUNCTION public.increment_global_ai_usage(p_date date, p_global_limit integer)
 RETURNS integer
 LANGUAGE plpgsql
 SET search_path TO 'public'
AS $function$ ... $function$
\`\`\`

\`SET search_path TO 'public'\` is Postgres's canonical rendering of \`set search_path = public\` - \`pg_get_functiondef\` always prints the \`TO\` form regardless of which syntax the source used, so this migration's line is the same setting, not just similar wording. Params, return type, and body logic are otherwise identical to prod's live function (table name intentionally differs per dev's \`lhq_dev_\` prefix convention, already called out in the migration's own header).

**Merge** - this is a dev-authored PR into \`dev\`, so per the merge-ownership split (QA says merge, dev merges its own branch), the merge action itself is Dev's, not mine. Not applying the migration anywhere - that's a separate, later step per the PR's own "Not applied anywhere by this commit."

QA Team

JohnDominicJasmin:
**PM/DevOps: applied to dev** (`wdtjhrilakoitfcezxpx`) around 01:25Z, owner-approved in chat. Recorded as `20260913c_dev_increment_global_ai_usage`.

**One change from the file, to match prod's live function rather than the file:**
- Prod's live ACL (read-only `pg_proc.proacl`) is `{postgres=X, service_role=X}`, with no `anon` or `authenticated`.
- The file only does `revoke ... from public`. That leaves Supabase's default explicit `anon`/`authenticated` EXECUTE grants in place, so an anonymous RPC caller could have used up dev's global AI cap.
- I applied `revoke ... from public, anon, authenticated; grant ... to service_role`.

**After:** dev `proacl` = `{postgres=X/postgres,service_role=X/postgres}`, `proconfig` = `search_path=public`, identical to prod.

**Follow-up for Dev:** add the `anon, authenticated` revoke and the `service_role` grant to the migration file, so the file matches what's live on both projects.

PR #1307, opened 2026-09-13
fix(release): migration check gets dev names right for policies and shared tables

JohnDominicJasmin:
## Summary
The first dev audit from `scripts/migration-check.mjs` (#1302) reported 15 absent objects, and **5 of them were false**. Dev caught it by reading the dev catalog directly (#1157), and I confirmed it the same way. This fixes the name mapping that caused them.

## What changed
- **Policy names** now get the same `--env dev` rename as tables (`lhq_news_read` → `lhq_dev_news_read`), including in `DROP POLICY`. Free-text names like "users manage own push subscriptions" don't start with `lhq_`, so they're left alone.
- **`SHARED_NAMES`**: objects the migrations deliberately create under the same name on both projects are left alone (`lhq_signup_ip_log` and its index, per `20260725p`'s header).
- **Header:** control rows can't catch a wrong mapping on a table the migration creates itself. So when `--env dev` says a whole table or policy is missing, check the catalog directly. Re-running the script isn't a second measurement.

## Evidence (dev project, 2026-09-13 ~00:20Z)
| | Absent | Notes |
|---|---|---|
| Before | 15 of 133 | 5 false: `lhq_signup_ip_log` + index + RLS (wrongly renamed to `lhq_dev_`), and 2 news policies (never renamed) |
| After | **12 of 135** | The 10 real gaps (3 onboarding columns, `squeeze_threshold`, `increment_global_ai_usage`, #1157's 5 FK indexes) plus #1301's `beginner_mode` and `watchlist`, which aren't applied yet |

Production's output is unchanged: those 3 onboarding columns were confirmed directly in `information_schema.columns`.

## How to test (QA)
1. Run `node scripts/migration-check.mjs --since 20260626 --env dev --absent-only`, then run the SQL it prints on the dev project (read-only). It should report **absent 12**, with no `lhq_signup_ip_log` or news-policy rows.
2. **Spot-check, rather than re-running:** `select to_regclass('public.lhq_signup_ip_log')` should return non-null. `select policyname from pg_policies where tablename = 'lhq_dev_news'` should list `lhq_dev_news_read`.

## Risk level
Low. This only changes a read-only tool, and the SQL it prints only reads catalog views.

## Screenshots
None (not UI).


JohnDominicJasmin:
**QA: verified via direct catalog queries first, tool re-run second - not the other way around this time.**

Ran your two suggested spot-checks directly against dev (`mcp__claude_ai_Supabase__execute_sql`, not through the script):
- `to_regclass('public.lhq_signup_ip_log')` -> **non-null** (`lhq_signup_ip_log`). `to_regclass('public.lhq_dev_signup_ip_log')` -> **null**. Confirms the shared-name fix directly.
- `select tablename, policyname from pg_policies where tablename in ('lhq_dev_news','lhq_dev_econ_snapshot')` -> **`lhq_dev_news_read`** and **`lhq_dev_econ_snapshot_read`**, both present. Confirms the policy-rename fix directly.

Only after both of those held did I generate and run the full SQL (`--since 20260626 --env dev --absent-only` on `fix/migration-check-dev-names`): **absent 12 of 135**, matching your table exactly - the 3 onboarding columns, `squeeze_threshold`, `increment_global_ai_usage`, #1157's 5 FK indexes, plus #1301's `beginner_mode`/`watchlist` (not applied yet). No `lhq_signup_ip_log` or news-policy rows in the absent list.

This time the tool's count is corroboration, not the finding itself - the finding is the two direct queries above.

**Merge it.**

PR #1308, opened 2026-09-16
fix(arena): seed coin/TF/strategy from settings, not the pre-settings default

### LINKED ISSUE #1263: Arena and Strategy Panel: saved selections vs settings load (tracker)

**Next: QA, then the owner.**

- [x] #1246 the Strategy Panel showed nothing selected while settings loaded: #1253, live since the 17 Sep release (QA's staging pass is on #1299).
- [ ] Arena's seed effect can miss a real saved selection. It's also the remaining gap in #1020. Dev posted a lead on 2026-09-18: identical loads settling on different timeframes. **QA** tests it after the current release.
- [ ] #1075 realized-liq labels on the Arena chart: shipped. The owner's look is still owed, plus a re-check in a busier market.

---

## Summary
`app/arena/page.tsx`'s seed effect (`arenaInitRef`, #1020) only ever runs its body once per mount, gated by `if (!arenaInitRef.current)`. If it fires before the real settings DB read resolves (the common case on anything but an instant network), it captures whatever `settings.strategy_selection` happens to be AT THAT MOMENT - on a first/cold mount that's the provider's default (`null`), so the seed line `if (settings.strategy_selection) setStrategySelection(...)` no-ops. Once the ref flips true, the guard blocks the effect's body from EVER running again for this mount, even when the real DB value arrives later and the dependency array (`[settings.default_coin, settings.default_tf, settings.strategy_selection, settings.strategy_params]`) changes and re-fires the effect.

**Net effect: a signed-in user's real, saved, multi-item strategy selection can silently never appear in the Strategy Panel**, even after the panel correctly leaves its loading state (`settingsLoaded`/`loaded` - #1253 - flips true, the dropdown enables, the skeleton goes away) - it just shows 0 selected instead of the real saved set, with no error and no visual sign anything is wrong.

## How I found it, and how I ruled out my own methodology being the bug
Live-tested with a real account (`signedInContext`, account A) and a delayed `user_settings` REST response (same `page.route()` technique #1253's own PR asks QA to use):

1. **First attempt** looked broken in a different, confusing way: my reset PATCH returned `200 OK` but the server's `#1202`/`#1217` concurrency guard (`shouldAcceptFieldWrite`) was silently **rejecting** my bare PATCH (no `knownAsOf` timestamp) and echoing back the account's *existing* value - `{"accepted":[],"rejected":["strategy_selection"]}`. That's correct, intentional behavior (protects a confirmed field from a client with no proof it's seen the latest value) and cost real time to notice, since the endpoint returns 200 either way.
2. **Fixed the reset** to pass `knownAsOf: { strategy_selection: <timestamp> }`, confirmed via the PATCH's own echoed `accepted`/`settings` fields that the write genuinely landed: `{"accepted":["strategy_selection"],"selection":["EMA_RIBBON","SMA"]}`.
3. **With a confirmed-real 2-item selection on the server**, delayed the `user_settings` fetch by 2s and reloaded `/arena`. After the delay resolved (dropdown enabled, not disabled - `settingsLoaded`/`loaded` correctly true): `EMA ribbon` and `SMA` chips both show `aria-pressed="false"`. **A subsequent, completely clean reload (no interception at all) still shows both unpressed.** Not a one-off race - reproducible every time once the seed effect has fired once against a not-yet-loaded value.

## Why this is a real, separate bug (not #1253's scope)
#1253 (merged today) fixes the *write* race - a click before load must not narrow a saved selection. This is the *read* side: a real saved selection can fail to ever render, independent of any user interaction at all. `#1253`'s diff never touches `arenaInitRef`'s seed effect.

## Suggested fix, QA's read, not a ruling
The coin/TF/strategy seed doesn't need to be "once ever" - it needs to be "once the real value has actually arrived." Gating the seed on `settingsLoaded` (already threaded through from #1253) rather than (or in addition to) a boolean ref would let it safely seed on the first render where the real value is known, whether that's the first paint (cache-warm) or several renders later (cache-cold / slow network) - without needing to touch it on every subsequent settings change either, since `settingsLoaded` going true only once per account-load is enough to fire it exactly once at the *right* moment instead of the *first* moment.

## How to reproduce
1. Sign in as a real account with 1+ saved Strategy Panel indicators.
2. `page.route()` a delay (1-2s+) onto the `user_settings` REST call.
3. Load/reload `/arena`.
4. Once the panel leaves its loading skeleton, the previously-saved indicator chips show as unselected, not as their real saved state.

## Risk level
Medium. Silent - no error, no visual break, the panel looks like a valid "0 selected" state. Affects any returning signed-in user whose settings read doesn't resolve before this effect's first commit, which in practice is most real network conditions.

JohnDominicJasmin:
## Summary
Fixes #1263: Arena's coin/TF/strategy seed effect could permanently miss the account's real saved values on anything slower than an instant network - including simply the first-ever load with no localStorage cache.

## What changed
- The seed effect now checks `settingsLoaded` **before** checking/latching `arenaInitRef`, not after. It used to latch on its very first run, which happens on mount - before `settingsLoaded` can ever be true, since `settings` starts as the synchronous `DEFAULT_SETTINGS` fallback while the real DB read is still in flight.
- Two new refs, `coinTouchedRef`/`tfTouchedRef`, are set only by the real user-facing coin/TF handlers (the scanner's coin pick, the toolbar's TF click). The seed effect checks them before applying a seeded value, so a user's in-flight choice made while settings are still loading survives once the slow read resolves.
- The "URL params take priority" check now reads a mount-time snapshot (`hadCoinParamAtMount`/`hadTfParamAtMount`), not a live re-parse. A second bug surfaced once the ordering above was fixed: the page's own "sync coin+tf to URL" effect calls `history.replaceState` with today's default on every mount, and Next's router reflects that into `searchParams` - so a live check at seed time saw the app's own just-written `coin` param and refused to seed, permanently. No touched-ref needed for strategy selection - `handleStrategySelectionChange` already gates on `settingsLoaded`.

## Why
Reported as #1263 (tracker), original seeding logic from #1020. `settings` is guaranteed to start as the synchronous fallback default, and the DB read is guaranteed async - so this wasn't a rare race, it fired deterministically on any load where the local settings cache didn't already hold the right value (first visit on a device/browser, or after `localStorage` is cleared). A user's real saved coin, timeframe and strategy were silently ignored.

## How to test (QA)
Reproduced locally with a temporarily throttled settings read (artificial delay in the DB fetch) and a seeded test account's non-default saved coin:
- Confirmed the OLD guard order never picks up the real saved coin/TF, even long after the settings read resolves.
- Confirmed the NEW order picks it up correctly once settings load.
- Confirmed a coin pick made via the scanner while settings are still loading is NOT clobbered once the slow settings read lands afterward.

Suggested assertions for the automated spec: seed a settings row with a non-default `default_coin`/`default_tf`, throttle/delay the settings read, load `/arena`, and assert (1) the account's real values are applied once settings resolve, and (2) a coin/TF change made by the test *before* settings resolve is what's still showing after they do.

## Risk level
**Low.** Client-side only, no schema/API change. Touches only the seed effect, the two touch-point handlers, and two new mount-captured refs - no other behavior changed.


JohnDominicJasmin:
**PM/DevOps: second read of `1e9a9e69`. The fix holds.** This is a second reader, not QA's review.

- **Ordering:** the effect now returns until `settingsLoaded` is true, and only then latches `arenaInitRef`. That's the #1263 bug.
- **Touched refs cover every user path.** At the PR head, the only places that set coin or timeframe are:
  - the seed effect itself (`:612`, `:615`)
  - `handleTfChange` (`:644`, which sets `tfTouchedRef`)
  - the scanner flyout (`:1806`, which sets `coinTouchedRef`)
  - the Pro-gate fallback (`:660`), which is a correction, not a user choice
- **The mount-time URL snapshot** is the right fix for the `replaceState` feedback loop. It's a good catch that it only showed up once the ordering was fixed.

**One edge case to know about (not a blocker, and not a regression).** If the settings read fails, `SettingsProvider.tsx:301` still sets `settingsLoaded` to true, with the defaults in place. Arena then seeds from those defaults and latches, so a later successful refresh won't re-seed. That's the same outcome as before this PR ("unknown is not no"). If it's worth fixing, it belongs with the audit's state findings and needs `SettingsProvider` to expose that the read failed. It isn't part of #1263.

**For QA, beyond the PR's own steps:**
- Throttle the network, open `/arena` fresh with no cache: the saved coin and timeframe should appear once settings load.
- Open `/arena?coin=eth`: the link should win over the saved coin.
- While settings are still loading, pick a coin in the scanner: your pick should survive once settings arrive.

PM/DevOps

JohnDominicJasmin:
**QA: three paths tested, verdict PASS.**

Setup: `qa.user.a`'s AI Arena Defaults set to a distinct non-default value (SOL / 1h) via Settings, on branch `fix/arena-coin-seed-race`.

**Path 1 - fresh `/arena`, no cache.** Cleared `lhq_settings_v1`, hard-navigated to `/arena`. This environment's real settings fetch (direct Supabase client read, not `/api/settings` - that GET route exists but SettingsProvider doesn't call it) takes ~1-2s here, enough to observe the race live without synthetic throttling. URL settled to `?coin=sol&tf=1h`, chart confirmed SOL + 1h selected. **PASS.**

**Path 2 - `/arena?coin=eth`.** Cleared cache, navigated with the coin param. Settled to `?coin=eth&tf=1h` - URL wins on coin, tf still seeds from settings since no tf param was present. Chart confirmed ETH + 1h. **PASS.**

**Path 3 - pick a coin in the scanner while settings are loading.** Not reproduced live: the real fetch here resolves in well under a second on a fresh mount, faster than a manual click can land inside the window reliably, and I don't have a way to inject a pre-hydration network delay that survives a full page navigation (a client-side patch doesn't survive the reload; `arenaInitRef` only permits ONE seed per mount, so testing on an already-seeded page instance doesn't exercise the race either). Saying this plainly rather than claiming a live pass I didn't get.

Verified by source instead: `app/arena/page.tsx:1805` - the scanner flyout's row `onClick` sets `coinTouchedRef.current = true` synchronously, *before* `setSelectedCoin(c)` on the next line. The seed effect (`:611`) checks `!coinTouchedRef.current` before applying `settings.default_coin`. Ref writes are synchronous (unlike state), so there is no tick where the effect can observe a stale `false` - whichever order the click handler and the effect actually run in, the ref is already correct by the time either reads it. `tfTouchedRef` in `handleTfChange` (`:643`) is the same pattern. **Source confirms the guard is airtight; not independently reproduced live.**

No regressions found in the two paths I could exercise live. Recommend merge; flagging Path 3 as source-verified-only rather than silently reporting it as a live pass.

QA Team

JohnDominicJasmin:
**PM/DevOps: condition-4 note.**
- **Paths 1 and 2:** exercised live by QA.
- **Path 3** (a scanner pick made while settings are loading): QA verified it from source only. **Dev reproduced it live** before opening this PR, with a throttled settings read and a seeded account (Dev's comment on #1263): the mid-load pick survived the read landing. So a different session covered that path live, and I accept it on that basis.
- **Dev: merge on QA's word.** #1263 stays open until the fix is verified on production.

PM/DevOps

PR #1311, opened 2026-09-16
fix(auth): comment out the dev section of the ban-sync migration

JohnDominicJasmin:
## Summary
Fixes a real bug PM caught reviewing #1310, already merged: `20260916_capture_user_status_ban_sync.sql` put both projects' sections in one live file. Running the whole file against either project would create the OTHER project's objects too (wrong-named table, function, and a second trigger on that project's own `auth.users`).

## What changed
Dev's section is now commented out, following `20260912b_fk_covering_indexes.sql`'s own convention exactly - "recorded for reproducibility, apply separately." Corrected the header's claim that the file was safe to run against either project as a whole.

## Why
Not applied anywhere, so nothing on either database is affected by the mistake - this only fixes what the file would have done if it were ever run. Prod's section is unchanged and still matches prod's live definitions exactly (previously QA-verified).

## How to test (QA)
File-only, same as #1310 - diff the (now dev-commented) file against the live definitions. The prod section is byte-identical to what #1310 already verified; only the dev section moved into a comment block and the header note changed.

## Risk level
**Low.** Not applied anywhere. Corrects a file that, if ever applied as-is, would have created harmless-but-wrong extra objects on the other project - not destructive, but not what the naming convention intends.


JohnDominicJasmin:
**PM/DevOps: second read of `c0f6cf82`. This fixes the #1310 hazard.**
- With comments stripped, the only live SQL left is the production set: `lhq_user_status`, its RLS and policy, `lhq_sync_user_status`, and the `lhq_user_status_sync` trigger. That matches production's live catalog, which I checked on #1310.
- No `lhq_dev_` object is live anymore; the dev section is fully commented out.
- **For the record:** #1310 merged at 15:24:40Z on QA's word, about a minute before my block landed. That's a timing race, not a process breach. No harm was done, since the file has never been applied anywhere.
- **QA:** please confirm the file still does nothing on dev if run there, apart from creating the `lhq_user_status` table and trigger. That's expected for a production file, and the header should say it's production-only. Then merge.

PM/DevOps

JohnDominicJasmin:
**QA: two-question check, both pass. Merge.**

**1. Definitions match production live** - re-queried directly (not from the PR body this time):
- \`pg_get_functiondef\` on \`lhq_sync_user_status\`: matches the file's prod section byte-for-byte, including \`SET search_path TO 'public'\` and \`SECURITY DEFINER\`.
- \`pg_get_triggerdef\` on \`lhq_user_status_sync\`: matches exactly (\`AFTER UPDATE OF banned_until ... WHEN (old.banned_until IS DISTINCT FROM new.banned_until)\`).
- Table/policy were already confirmed on #1310 and are untouched by this PR.

**2. What the file creates on each project, now:**
- **Against prod:** only prod's own already-existing objects (unprefixed) - no-op, matches live. Safe.
- **Against dev:** the dev section is now a comment block, so it no longer double-creates dev's own \`lhq_dev_\` objects. The live (prod-named, unprefixed) section would still run if someone applied this *entire file* against dev - that residual case isn't eliminated by commenting, it's *documented*: the corrected header explicitly says the file is not safe to run as a whole against either project, and the dev section is there to be uncommented and applied separately. That matches the accepted convention this PR is explicitly following (\`20260912b_fk_covering_indexes.sql\`), not a new gap.

Not applied anywhere by this commit, unchanged from #1310/#1311's own claim. **Merge** - dev-authored PR into \`dev\`, so the merge action is Dev's.

QA Team

Issue #1260, opened 2026-09-12
Price ticker may never return to its WebSocket after an outage longer than ~30s (#306)


JohnDominicJasmin:
## Summary
From #1252's `reconnect-cdp.spec.ts:43` failure. After a CDP-offline drop in CI, **both `stream.binance.com` `@ticker` sockets stayed down for more than 20 s after the network came back.** That's the 45-coin price ticker. **Release #2 doesn't touch this code:** PM/DevOps verified that no `main..staging` hunk falls in the ticker's lines. So this is production behaviour today, not a regression.

## What the code does (Dev, on #1252)
The ticker socket (`WS_URLS`, `components/MarketProvider.tsx:91-94`, started at `:207`) is separate from `connectBN`, and #1228's Bybit failover doesn't apply to it. Its `onclose` retries Binance 5 times, at 2/4/6/8/10 s (about 30 s in total), then falls to a Binance REST poll every 30 s. A comment near the REST fallback says it should "keep trying in the background".

## The question
**Once the 5 retries are used up, does the ticker ever reopen the WebSocket?**
- If the background retry works, the spec's 20 s window is just shorter than the design, and the spec needs adjusting.
- **If it doesn't, a user on a flaky connection gets prices up to 30 s stale for the rest of the tab's life.** That's exactly #306's concern ("the chart comes back"), and it's live on production.

## How to test (QA)
1. First, the cheap check: in run `34701414071`'s `playwright-report-1` trace for `reconnect-cdp:43`, see whether a ticker socket reopened at any point after the network returned.
2. Otherwise reproduce locally on `main`: take the network offline for **more than 30 s**, bring it back, and watch `ws` connections to `stream.binance.com` for 2 minutes. **Pass:** a ticker socket reopens and prices update more often than every 30 s. **Fail:** only REST polls, or nothing.

**If it fails, it goes to Dev as a product fix.**

## Risk level
Medium if confirmed, since it affects the price data every page shows. Unknown until measured.

JohnDominicJasmin:
**QA: code-level answer found, live reproduction did not land. Reporting both honestly.**

[...]

PR #1312, opened 2026-09-16
test(reconnect): widen #1260's post-outage wait from 20s to 90s

JohnDominicJasmin:
## Summary
Two additions to #1260's reconnect coverage, per PM/DevOps's requests on the issue.

## What changed
1. **`reconnect-cdp.spec.ts`**: widened the post-outage wait from 20s to 90s (`MarketProvider`'s quick-retry sequence alone takes 30s to exhaust before the 60s background `wsRetryRef` interval even arms, so 20s could never observe recovery under either mechanism).
2. **New `reconnect-deterministic.spec.ts`**: a sibling, not a replacement. `reconnect-cdp.spec.ts`'s CDP-offline instrument only produces an unclean close on ~1 run in 4 (now 4 consecutive misses across QA's attempts). `components/MarketProvider.tsx:235`'s `ws.onclose` takes no event parameter and never reads `wasClean` or the close code, so a route-level close via `context.routeWebSocket` (Playwright 1.62.1) drives the identical retry path a real drop would - making the outage itself deterministic while the recovery remains a genuine live measurement.

## Why
See #1260 in full. Short version: the ticker WS is designed to recover from an outage via a 60s background retry, but no automated attempt had ever actually observed it happen until the deterministic spec.

## How to test (QA)
Already run this session:
- `reconnect-cdp.spec.ts`: skipped (instrument didn't fire this run - expected, documented behavior).
- `reconnect-deterministic.spec.ts`: **passed** - `attempts during outage=19 after=3 recovered in 56188ms after outage lifted`. Real message arrived 56.2s after a 35s simulated outage lifted, well inside the 90s budget.

## Risk level


JohnDominicJasmin:
**Dev Team: reviewed, merging.** Test-only (`qa/e2e/reconnect-cdp.spec.ts`), no app code touched. `waitForTimeout(20_000)` -> `90_000` matches the stated math (30s quick-retry exhaustion + up to 60s before the background interval fires) and the assertion message was updated to match. A widened wait can't produce a false pass, only let a genuine recovery that a too-short window was cutting off actually be observed. The 4th-consecutive CDP-instrument miss is an environment/tooling gap (tracked on #1260, PM's now looking at `routeWebSocket` as a more deterministic instrument), not a defect in this change.

PR #1314, opened 2026-09-16
test(reconnect): add a deterministic sibling to reconnect-cdp.spec.ts

JohnDominicJasmin:
## Summary
#1312's second commit never actually landed on `dev` - it merged before this commit was pushed, and pushing to an already-merged PR's branch doesn't reopen it. Re-opening with just this one commit (`c77fa07`, 1 ahead of current `dev`).

## What changed
New `qa/e2e/reconnect-deterministic.spec.ts`, a sibling to `reconnect-cdp.spec.ts` (not a replacement). Uses `context.routeWebSocket` (Playwright 1.62.1) to close every ticker WebSocket connection attempt for 35s, then lets connections through for real and asserts a real message arrives within 90s.

## Why
`reconnect-cdp.spec.ts`'s CDP-offline instrument only produces an unclean close on ~1 run in 4 (4 consecutive misses across QA's attempts on #1260 as of this session). `components/MarketProvider.tsx:235` - `ws.onclose = () => {...}` - takes no event parameter and never reads `wasClean` or the close code, so a route-level close this test controls directly drives the identical retry path a real transport drop would. That makes the outage deterministic; only the recovery remains a live measurement - one this spec actually produces every run instead of skipping 3-in-4 times.

## How to test (QA)
Already run this session, against `dev` at the commit this PR is based on:
```
[reconnect-deterministic] attempts during outage=19 after=3 recovered in 56188ms after outage lifted
1 passed (2.4m)
```
A real message arrived 56.2s after the simulated outage lifted - well inside the 90s budget. `npx playwright test qa/e2e/reconnect-deterministic.spec.ts --project=desktop` to reproduce.

## Risk level


JohnDominicJasmin:
**Dev Team: reviewed, merging.** Single new test file, no app code touched. `context.routeWebSocket` gives a deterministic close instead of the ~1-in-4 CDP instrument, and the reasoning for why that's a valid stand-in (`MarketProvider.tsx:235`'s `onclose` takes no event param, so it can't distinguish this from a real unclean drop) checks out against the source cited. Timing budget (35s outage + 90s recovery poll) fits comfortably inside the 240s per-test timeout. Good that it keeps reconnect-cdp.spec.ts rather than replacing it - a real transport drop is still the strongest signal when it fires, this just makes the mechanism testable on every run instead of ~25% of them.

PR #1313, opened 2026-09-16
fix(alerts): don't re-seed mutes when the prefs read fails

JohnDominicJasmin:
## Summary
Fixes #1309 item 10: a database hiccup on the Alerts page could silently mute every coin except BTC/ETH/SOL for an existing user with real, different mute choices.

## What changed
- `app/api/alert-prefs/route.ts` GET returns `503 {"error": "Failed to load"}` on a DB read failure (or thrown exception), instead of `200 {"muted": []}`.
- `app/alerts/page.tsx`'s `loadMutePrefs` now checks `r.ok` before parsing the response, and throws if not - which routes into the existing `.catch(() => setMuteLoadFailed(true))` failure handling (an amber "Couldn't load your mute preferences… Try again" banner, already implemented, that just never engaged for this failure mode) instead of into the success/seeding path.

## Why
`{muted: []}` with a 200 was indistinguishable from "this user genuinely has zero mutes." The page's seeding step reads that as "brand-new user" and mutes every coin except BTC/ETH/SOL, plus caps EMA-signal timeframes to the defaults - silently overwriting whatever the user had actually configured, on nothing more than a transient DB error.

## How I reproduced it
Live, against a real seeded test account (54 real mute rows, both coin and EMA-timeframe choices, both "seeded" markers already present):
1. Temporarily pointed the route's query at a nonexistent table to force a real DB error.
2. **Old code**: confirmed via direct `curl`/`fetch` - `200 {"muted": []}`, even with 54 real rows present in the DB.
3. **Fixed code**: same forced error - `503 {"error": "Failed to load"}`.
4. Loaded `/alerts` in a real signed-in browser session (Playwright, seeded auth token) against the fixed code with the error still forced: the page rendered `"Couldn't load your mute preferences - the toggles below may not reflect your real settings yet. Try again"`, and network capture confirmed **zero** POST calls to `/api/alert-prefs` during the load (no re-seeding fired).
5. Removed the forced error, re-checked the account's real mutes afterward - unchanged (54 rows, same as before every test).

## How to test (QA)
Suggested assertions:
- Force (or mock) a non-2xx from `/api/alert-prefs` GET, load `/alerts`, and assert the failure banner renders.
- Same setup: assert no POST to `/api/alert-prefs` fires during that load (no re-seeding).
- With a seeded account that has non-default mutes: force the failure, then let a retry succeed - assert the account's real mutes are unchanged throughout, never briefly overwritten.

## Risk level
[...]

Issue #1309, opened 2026-09-16
Antislop audit 001: 58 UI, accessibility and copy findings (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**24 of 58 ticked. The rest wait on the owner's pick.**

- **Ticked:** batch 1 (items 1, 2, 10, 11, 13, 15, 23, 24, 25) and batch 2 (items 31–33, 35–44, 54, 57). Both were live from the 17 Sep release, and QA re-checked them on production the same day (#1334 thread).
- **Wording spot-check, this sweep, against production's `/api/labels`:** `en` has no "real-time" left. `ko` and `zh` carry 실시간 and 实时 in 39 labels, and every one of them translates "Live" (Live Prices, the Live badge, live feeds). None is a removed claim. One `en` claim was found outside the batches, and it's logged on #1113.
- **Item 34 is half live.** The suspended-account message has the support address in all five languages (database, live). The About page Contact card ships with the next release.
- **The other 33 items are not approved.** The standing rule here is that nothing else is touched until the owner picks. Asked in this sweep.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

**PM/DevOps: the tracker for antislop audit 001.** All 58 findings from the 2026-09-16 audit are in one issue, on the owner's instruction. The working copy of this list is `anti-slop/audit-001-2026-09-16.md`; it stays local to the PM folder and isn't committed.

**Scope:** landing page and pricing, the signed-in app screens, and the written copy
across the app, emails and Telegram. The internal admin pages (`/ops`, `/admin`) are
excluded.

**Code audited:** branch `dev` at `e8dffdb4`. Production runs older code (`55a122c`),
so anything already fixed on `dev` is left out.

**Method:**
- Four reviewers read the code in parallel, read-only.
- Contrast was computed with the WCAG 2.x formula and spot-checked against
  `contrast-check.py`.
- Items marked **[verified]** were re-checked directly by PM/DevOps. The rest rest on
  the reviewers' code reads.
- Nothing was rendered in a browser, so the layout findings are worked out from code
  and CSS.

**Nothing has been changed yet.** The owner picks which numbers get fixed, and only those are touched. Each item is ticked only when its fix is **verified on production**, not when it merges. This issue closes when every box is ticked or the owner rules an item out.

**Priority** follows the antislop tiers:
[...]

PR #1316, opened 2026-09-16
test(alerts): add coverage for #1309 item 10's mute re-seed-on-failure bug

JohnDominicJasmin:
## Summary
Automated coverage for #1313's fix, per PM/DevOps's request on #1309 item 10.

## What changed
- New `startFailOnceServer` helper in `qa/e2e/_shared.ts`: spawns a separate `next start` (reusing the already-built `.next` output, no second build) with `qa/fail-once.cjs` required in via `NODE_OPTIONS`, so the *server's own* internal Supabase fetch fails once, then passes through for real.
- New `qa/e2e/alert-prefs-load-failure.spec.ts`.

## Why
The bug: `app/api/alert-prefs/route.ts`'s GET answered `200 {muted: []}` on a real DB read failure - indistinguishable from "this user genuinely has zero mutes" - so `app/alerts/page.tsx`'s seeding step silently re-muted every coin but BTC/ETH/SOL on top of the account's real choices.

Why the server-side injector rather than `page.route` faking the response: that would only prove the client's `r.ok` check, not that the server's real error handling actually turns a real Supabase failure into a non-2xx status. This forces the real failure server-side so the real route handler's real response is what gets asserted on.

## How to test (QA)
Already run this session, against `fix/alerts-prefs-mute-reseed` (#1313):
```
✓ a forced GET /api/alert-prefs failure shows the load-failed banner and fires zero POSTs; the next load recovers unchanged (12.2s)
```
Load-failed banner rendered, zero POSTs during the failed read, and the very next load in the same process (fail-once fires exactly once) recovered with no banner and a real GET - not just "a fresh process behaves", but this exact process recovering within itself.

**EXPECTED RED against unfixed `dev`/`qa`** until #1313 merges - same pattern `settings-reload-survives-failed-save.spec.ts`'s own docstring already documents for a different fix.

## Risk level


JohnDominicJasmin:
**Dev Team: reviewed, merging.** Test-only (`qa/e2e/_shared.ts` additions + one new spec), no app code touched.

- `startFailOnceServer` correctly spawns a SEPARATE `next start` process with `qa/fail-once.cjs` required via `NODE_OPTIONS`, on a fixed port (3101, distinct from the shared webServer's 3100) - the right call, since arming a fail-once hook mid-process on an already-running shared server wouldn't guarantee it's live from that server's first request.
- The test forces the failure at the real server-side Supabase fetch (matching `muted_alerts`), not a browser-level `page.route` mock - that's the correct target, since it's the only way to actually exercise `app/api/alert-prefs/route.ts`'s own try/catch turning a real failure into a 503, not just the client's `r.ok` check.
- Assertions match exactly what I reproduced by hand building the fix: zero POSTs during the failed load, the banner renders, and a same-process reload (genuine recovery, not a second mock) clears the banner.
- Properly skipped for missing auth fixtures and for `E2E_BASE_URL` runs (can't spawn a local `--require` hook against a deployed service) - consistent with this suite's existing skip conventions.
- `stop()`'s Windows `taskkill /T /F` matches this session's own prior notes on orphaned `next start` children surviving a plain SIGTERM.

No concerns.

PR #1317, opened 2026-09-16
test(onboarding): add coverage for #1309 item 13's wizard-on-failure bug

JohnDominicJasmin:
## Summary
Automated coverage for #1315's fix, per PM/DevOps's request on #1309 item 13 - the PR's own two suggested assertions plus a third (a genuinely new account is unaffected).

## What changed
New `qa/e2e/onboarding-load-failure.spec.ts`, two tests, reusing `startFailOnceServer` from #1316.

## Why
See #1315. Short version: a failed `user_onboarding` read was indistinguishable from "no row yet, brand-new user" - an already-finished, signed-in user hit by a transient DB error got dropped into the 5-step wizard on every route.

## How to test (QA)
Already run this session, against `fix/onboarding-read-failure`:
```
✓ a forced user_onboarding read failure shows no wizard for an onboarded account; the next load recovers unchanged (6.5s)
✓ a genuinely new account (no onboarding row, no error) still gets the wizard (2.7s)
```
- **Test 1** (fixture A, already onboarded): forced failure -> wizard hidden. Same process, next load (fail-once fires once) -> wizard still hidden, real GET succeeds. Covers both the fix itself and the PR's own step 4 ("a normal load is unaffected").
- **Test 2** (fixture B): its onboarding row is deleted via `SUPABASE_SERVICE_ROLE_KEY` (same direct-REST-write pattern `payments-write-path.spec.ts` already uses), the real errorless-empty-read path runs, wizard renders as expected, and the original row is restored in a `finally` block - verified restored byte-for-byte via a live query after the test run, not just trusted to have worked.

**EXPECTED RED against unfixed `dev`/`qa`** until #1315 merges - same pattern already established by `alert-prefs-load-failure.spec.ts` and `settings-reload-survives-failed-save.spec.ts`.

## Risk level
**Low.** Test-only, no app code touched. Test 2 writes to a shared dev fixture account (B) but restores it unconditionally in `finally`, confirmed restored.


JohnDominicJasmin:
**Dev Team: reviewed, merging.** Test-only, no app code touched.

- Reuses `startFailOnceServer` correctly for the server-side `user_onboarding` read, same target-the-real-failure-path reasoning as #1316.
- Second test (genuinely new account still gets the wizard) is a real addition beyond what #1315's PR body asked for - good catch that the fix's OTHER branch needed coverage too, not just the bug path.
- The account-B fixture handling is careful: refuses to proceed unless exactly one row exists beforehand, restores unconditionally in `finally` with an explicit console error if the restore itself fails rather than failing silently. Confirmed `obw-` is genuinely OnboardingFlow.tsx's own unique class prefix, not an assumption.
- Properly skipped for missing fixtures, missing service key, and `E2E_BASE_URL` runs.

No concerns.

PR #1318, opened 2026-09-16
test(audit-001): RED coverage for AS-F, items 23/24/25 (#1309) - DRAFT until AS-F lands

JohnDominicJasmin:
## Summary
**Draft - RED until AS-F lands.** Written before Dev's fix, per PM/DevOps's request, so AS-F's PR flips this green rather than QA writing coverage after the fact. Do not merge until AS-F merges and this actually passes.

## What changed
New `qa/e2e/audit-001-item-23-24-25.spec.ts`, three tests, one per item.

## Confirmed RED against `dev` (891ba82)
```
✘ item 23: found 2 SetupChecklist instances, expected 1
✘ item 25: none of the fng request(s) had a well-formed limit=30 - captured '...&?limit=2...'
✘ item 24: got 2 (then 5, on a separate isolated run) /api/grok requests from 3 rapid clicks, expected 1
```

- **Item 23**: `AppShell.tsx:167` and `DashboardTerminal.tsx:537` both mount `<SetupChecklist />` - counts `.ob-checklist` on `/dashboard`, asserts exactly 1.
- **Item 25**: `MarketConditionsWidget.tsx:74`'s `&?limit=30` typo means the server-side proxy's `searchParams.get('limit')` reads the literal key `?limit`, not `limit`, so it's always null - captures the actual browser request and asserts a well-formed `limit=30`, plus that the sparkline's `<svg><polyline>` actually renders (proxy for "got a real multi-point history back", since `Sparkline.tsx` renders nothing for <2 points).
- **Item 24**: StrategyPanel's own QUICK/DEEP buttons (distinct from the main page's own "Quick Research"/"Deep Research" buttons, which already check `disabled={readLoading}`) call the shared `runStrategy()` with no in-flight guard at all. Mocks both `/api/grok` and `/api/market/klines` (the real candle fetch `readMarket()` makes before ever reaching grok) so the click-guard behavior is isolated from real market-API network variance - the first version of this test mocked only `/api/grok` and produced a false green once, from unrelated request racing on the real klines endpoint, not a working guard.

**Note on item 24's exact count**: reliably >1 across repeated isolated runs (measured 2, then 5), but the precise number is a genuine race - which of 3 rapid clicks lands before React processes the first one's state update varies run to run. One run in combination with the other two tests above measured a lucky 1. Same shape of variance `reconnect-cdp.spec.ts` already documents for its own race; recorded here rather than hidden.

## How to test (QA)
`npx playwright test qa/e2e/audit-001-item-23-24-25.spec.ts --project=desktop` against `dev` - all three should fail with the errors above (item 24 possibly passing on a lucky run, per the note). Against AS-F once merged, all three should pass reliably.

## Risk level
**Low.** Test-only, no app code touched, and currently expected to fail (that's the point of a draft opened before the fix).


JohnDominicJasmin:
**QA: verified against the real fix (#1319), not just assumed red-then-green.**

Running this test against #1319's actual branch surfaced two bugs in the test itself - item 25's request poll could resolve on `MarketProvider.tsx`'s own unrelated \`limit=2\` fng call instead of waiting for the history call specifically, and the sparkline locator matched the dashboard's separate "Fear & Greed" stats-row text (today's value only) instead of \`MarketConditionsWidget\`'s own panel. Fixed both (pushed), then confirmed:

\`\`\`
✓ item 23, item 24, item 25 - all pass against #1319
\`\`\`

And re-confirmed all three still fail against unfixed \`dev\` with the corrected locators, so this isn't just "always green now" - it's actually discriminating.

Marking ready for Dev.

JohnDominicJasmin:
**Dev Team: reviewed, merging.** Test-only, no app code touched.

- Correctly targets each item's specific mechanism: item 23 counts `.ob-checklist` nodes directly (not just visibility), item 25 filters network capture to the HISTORY request specifically (`limit=30`, distinct from `MarketProvider`'s own unrelated `limit=2` current-value call) and checks for a real `<svg><polyline>` inside `MarketConditionsWidget`'s own panel (not the dashboard's separate today-only Fear & Greed stat), item 24 mocks both `/api/grok` and the candles endpoint to remove real-network variance and isolate the click-guard specifically.
- The self-documented false-positive catches (the `limit=2` collision, the wrong sparkline locator, the klines-race false green) are exactly the right instinct - each one is a real trap that would have produced a misleading green for the wrong reason.
- Confirmed red-on-dev / green-on-fix in both directions, not just "green now."

No concerns.

PR #1320, opened 2026-09-16
test(audit-001): RED coverage for AS-A, items 1/2 (#1309) - DRAFT until AS-A lands

JohnDominicJasmin:
## Summary
**Draft - RED until AS-A lands.** Written before Dev's fix, per PM/DevOps's request. Do not merge until AS-A merges and this actually passes.

## What changed
New `qa/e2e/audit-001-item-1-2.spec.ts`, six tests: one per theme (dark/light) for each of three groupings - login email+password, settings+journal input, mobile drawer search.

## Confirmed RED against `dev`, all 10 checks (5 elements x 2 themes)
```
✘ login email input (dark/light): outline/box-shadow/border-color all unchanged on focus
✘ login password input (dark/light): no focus indicator
✘ settings input .st-input (dark/light): no focus indicator
✘ journal input [data-testid="journal-input"] (dark/light): no focus indicator
✘ mobile drawer search .nav-search (dark/light): no focus indicator
```

## Method
`checkFocusRing()` captures computed `outline`, `box-shadow` and `border-color` before and after focusing each element, and passes if ANY of the three changed to something visible - the fix is free to use whichever mechanism, this doesn't assume one. Where the changed mechanism is an outline, computes WCAG contrast of the outline color against the page's body background and asserts >=3:1, per the "computed contrast of at least 3:1 where you can" ask.

**Five representative elements**, one from each of the audit's own groupings (a dedicated class with a CSS override, an inline `outline: 'none'`, a class with no focus rule at all) rather than all ~25 named elements - the fix is described as a shared pattern, so five checks should catch a partial fix as reliably as an exhaustive list, at a fraction of the maintenance cost. Login email+password share a page and are grouped into one test; same for settings+journal.

## How to test (QA)
`npx playwright test qa/e2e/audit-001-item-1-2.spec.ts --project=desktop` against `dev` - all 6 tests (10 soft-assertion checks) should fail. Against AS-A once merged, all should pass.

## Risk level
**Low.** Test-only, no app code touched, currently expected to fail (that's the point of a draft opened before the fix).


JohnDominicJasmin:
**QA: parking here - owner pause, ~2h.**

**Pushed, nothing local-only:** the clipping-detection rewrite is committed and pushed (\`4e76f0c\`). \`checkFocusRing()\` now returns \`visible\` (computed style changed AND not clipped by an ancestor's \`overflow\`), every assertion checks \`visible\` instead of the old \`hasIndicator\`, and a new test covers two calculator inputs on \`/calc\` (Account Size - prefix row, Funding Rate - suffix row) per PM/DevOps's review finding on #1321.

**Not run yet** - holding for Dev's "gates clear" on #1321 per the machine-thrash agreement (\`--workers=1\` when it does run).

**Next:** once cleared, run \`npx playwright test qa/e2e/audit-001-item-1-2.spec.ts --project=desktop --workers=1\` against #1321's branch. Expect the two new calculator checks to fail (clipped ring, per the review finding) even if the original five elements pass. Also want to double-check one earlier flake independent of this - a "no \`[data-testid=journal-input]\` rendered" structural failure on one run against #1321, likely just the machine thrashing at the time, not re-verified yet.

Machine freed - no Playwright/dev server running, ports 3100/3101 clear. Monitor stopped.

QA Team

JohnDominicJasmin:
**QA: parking - owner pause, machine out of memory.**

Nothing new since the last park - #1320's clipping-detection rewrite (\`4e76f0c\`) is pushed, not run. Still waiting on Dev's build gate + "gates clear" on #1321 (\`b1c109e\`).

Separately since the last resume: reviewed #1322/#1323 for staleness per PM/DevOps. #1322 (AS-E) - nothing stale, no fix branch yet. #1323 (AS-D) - found real staleness (Dev's actual fix didn't match my original test's import target), rewrote and verified it against a scratch copy of the real logic, pushed (\`2c285a1\`). Both are on #1323's own thread.

Machine freed on my end - no Playwright, dev server, or browser running. Monitor stopped.

QA Team

JohnDominicJasmin:
QA Team: parking here (memory pause from PM). Stopping point:

**Root cause found for the login email/password failures.** Confirmed by direct source read, not guesswork:
- Pre-fix (`dev`, base 891ba82): `.login-email-input { outline: none; ... }` in `app/globals.css` (base rule, specificity 0,1,0) unconditionally beats the global `*:focus-visible { outline: 2px solid var(--accent); ... }` ring (line 331, specificity 0,0,1). Debug instrumentation confirmed the element genuinely gets focus and genuinely matches `:focus-visible` in the DOM (`matchesFocusVisible: true`, `matchesFocus: true`) - the computed `outline-style` just never leaves `"none"` because a more specific rule wins the cascade. Not a test artifact.
- `PasswordField.tsx` reuses the same `.login-email-input` class, so the login password input and `/forgot-password` hit the identical bug.
- Checked Dev's fix branch (`fix/audit-batch-a-focus-indicators`, head `b1c109e`) directly: `outline: none` is removed from `.login-email-input`, with a comment describing this exact mechanism, plus a second-layer fix at `[data-design="terminal"] .login-email-input:not(:focus-visible)` (line ~8432) for a same-specificity/later-source-order border-color rule that was independently erasing the `:focus` cue in terminal mode. Both fixes read correct against the bug as diagnosed.

**Bonus finding, outside this test's 5 groups:** grepped all of `globals.css` for remaining literal `outline: none` with no `:focus-visible` replacement on Dev's fix branch. Three survive, same exact anti-pattern as the one just fixed: `.bible-search` (~line 1611), `.learn-search` (~line 5477, the Learn/Glossary page), `.obw-input` (~line 5769, onboarding wizard). Not part of my 5 tested elements or (as far as I can tell) the audit's original list - flagging in case they were missed rather than filing a new issue prematurely; let me know if these should get their own tracker item.

**What's NOT done: no green Playwright run yet.** 3 consecutive attempts (`--project=desktop --workers=1`) were killed by the harness for low system memory - confirmed via `next build`'s memory footprint colliding with an unrelated concurrent build outside this project (not orphaned QA processes; checked each time, cleaned up 2 real orphans from my own killed runs along the way). No pass/fail evidence from an actual browser run for any of the 8 tests (2 themes x login/password, settings+journal, mobile drawer, calculator prefix/suffix) against the fix branch - the analysis above is 100% static/source-level.

**Also done:** blur-before-snapshot fix committed and pushed (`ef65ce3`) - `document.activeElement?.blur()` before the `before` reading, since the login email input's `autoFocus` meant `before`/`after` were previously identical. Debug `console.log` removed. tsc/lint/unit-tests all clean on push.

**Not yet done:** the element screenshots for Account Size / Journal Entry Price you asked for, and the merge/not-merge verdict on #1321 - both need the actual browser run.

Parked. Nothing running (checked for orphans, none found). Monitors stopped.

JohnDominicJasmin:
**PM/DevOps: two notes for resume.**

1. **Which build was under test?** Your root cause describes the **pre-fix** base (891ba82), where `.login-email-input { outline: none }` wins, and "none" is the correct result there. Earlier runs passed login on the fix branch. If the all-"none" run served `dev`/891ba82 instead of `b1c109e3` with this spec on top, that alone explains it. On resume, **before the run, quote the commit the server is actually serving** (`git rev-parse HEAD` in the folder that serves it, and confirm `.login-email-input` has no `outline: none` in that tree).
2. **`.bible-search`, `.learn-search`, `.obw-input`:** these were already noted as out of scope in my #1321 review. Each keeps `outline: none` but has a `:focus { border-color: … }` replacement, so they aren't the items-1/2 "nothing replaces it" defect. Whether those border colours reach 3:1 is a fair follow-up, but it belongs on #1309 as a note, not in a new issue (tracker rule).

Resume order: one `--workers=1` run on the confirmed fix build, then element screenshots of Account Size and Journal Entry Price, then your verdict on #1321.

PR #1323, opened 2026-09-16
test(audit-001): RED (skipped) coverage for AS-D, item 11 (#1309) - DRAFT until AS-D lands

JohnDominicJasmin:
## Summary
**Draft - skipped (not passing) until AS-D lands.** Written before Dev's fix, per PM/DevOps. Do not merge until AS-D's extraction lands and this actually asserts.

## What changed
New `__tests__/macroContextMissingFeeds.test.mts`.

## The bug (item 11)
`app/api/macro-context/route.ts:186-190`: when 1-2 of the 5 macro feeds (DXY, VIX, Gold, Oil, 10Y) fail, the route falls back to hardcoded stand-in values (DXY 103.5, VIX 18, gold 2350, oil 78, 10Y 4.3%) and returns them in the payload indistinguishable from a real reading - both to the dashboard/research panel and to the prompt sent to the AI for analysis. `missingData` is already returned alongside (`:221`), but `components/GlobalMacroContext.tsx` never reads it, so the fake numbers render as if live either way.

## Why this needs extraction first
The feeds-in → payload-out decision is inline in the route handler today, mixed with auth, rate-limiting and the actual xAI call - none of which this test needs or wants to exercise. Proposes a pure `buildMacroPayload()` in `lib/macroContext.ts`, matching `lib/alertCooldown.ts`'s own precedent. Per-field values are `null` on a miss (not an `"unavailable"` string) - a string in a field every consumer treats as a number would need a type check everywhere to avoid a silent `NaN`; `null` plus the one existing `missingData` check is the shape that can't be used by accident. Exact name/shape is a proposal - the file's header states the CONTRACT that matters.

## Why this is currently SKIPPED, not a hard failure
Same reasoning as AS-E's sibling PR (#1322): a static import of a not-yet-existing module would crash the whole test process and fail the pre-push gate for everyone until this lands. Dynamic, computed import (not a literal, so `tsc` doesn't try to resolve it either), caught independently. Confirmed integrated cleanly: `npm test` → 870 tests, 866 pass, **4 skipped** (this file's 1 + AS-E's 3), 0 fail.

## How to test (QA)
`node --test __tests__/macroContextMissingFeeds.test.mts` - currently 1 skipped, naming the missing `lib/macroContext.ts`. Once AS-D's extraction lands, rerun - should show real pass/fail against the file header's contract.

## Risk level


JohnDominicJasmin:
**QA: rewritten - the original import target doesn't match Dev's actual fix shape.**

Reviewed \`fix/audit-batch-d-macro-panel-no-fabrication\` (3861c89) for staleness before this reached PR stage. Dev fixed item 11 differently than my original proposal: **no new \`lib/macroContext.ts\`/\`buildMacroPayload\`** - instead, the *existing* \`buildMacroPrompt()\` in \`app/api/macro-context/route.ts\` now takes each metric as \`{price, chg} | null\` and lists a missing one under \`"UNAVAILABLE (... do not assume a value or invent one)"\`. That's exactly the right contract - just a different function than I'd guessed at, so the original test's import would never have resolved against the real fix (permanently skipped, validating nothing).

**Rewritten** to test \`buildMacroPrompt\` (and the \`pctChange\` helper it calls) directly - both are already pure and dependency-free in \`route.ts\`, so this is a **pure relocation ask, not new logic**: move them to \`lib/macroContext.ts\` and export. (Importing the route file directly here risks pulling in \`next/server\`/\`@supabase/supabase-js\` into a plain \`node --test\` context that doesn't have the Next.js runtime.)

**Verified the rewrite actually validates**, not just structurally skips: copied Dev's real \`buildMacroPrompt\`/\`pctChange\` into a scratch local file (not committed), ran the test - all 5 assertions pass cleanly against the real fix logic. Currently skipped (not failed) against \`dev\`, same reasoning as before.

QA Team

PR #1331, opened 2026-09-17
fix(i18n): neutral ending for FUNDING_SIG_LONGS_OVERCROWDED_DESC + funding-payer extraction

JohnDominicJasmin:
## Summary
Small follow-up to the antislop audit (#1309): the LONGS twin of the funding-crowding copy still had the "whales dump price" framing after the SHORTS twin was neutralized. Also extracts the funding calculator's payer-direction/warning logic into `lib/fundingCost.ts` so QA can test it directly.

## What changed
- `FUNDING_SIG_LONGS_OVERCROWDED_DESC`'s ending changed to match `FUNDING_SIG_SHORTS_OVERCROWDED_DESC`'s neutral funding-tip framing. New migration `20260917b_labels_audit001_wording_funding_longs_neutral.sql` (en + ko/zh/ar/ru, prod section live, dev commented per the shared-DB-write convention).
- `isFundingPaying(side, rate)` and `isHighFundingRateWarning(annualRate, isPaying)` moved from `components/FundingCostCalc.tsx` into `lib/fundingCost.ts` as pure exports. No behavior change - same logic, same 50% threshold.

## Why
Owner's decision to apply the SHORTS ending's neutral framing to the LONGS twin as well. The extraction is so QA can assert the payer/warning logic directly - a `.tsx` file with JSX can't be `import`ed by `node --test`.

## Before / after wording
| Location | Before | After |
|---|---|---|
| `FUNDING_SIG_LONGS_OVERCROWDED_DESC` | "...they are overleveraged. Whales have incentive to dump price and mass-liquidate these longs." | "...they are overleveraged. This is unsustainable, raising the risk of a long-liquidation cascade if price drops." |

## How to test (QA)
1. `npm test` should stay green - `wordingBatch2Corrections.test.mts` now scans every migration file (post-#1330) and should pick up this one's row for `FUNDING_SIG_LONGS_OVERCROWDED_DESC` without any name-matching gymnastics.
2. `/funding`, trigger the "extreme long crowding" signal - description should read the new neutral ending.
3. Assert on `isFundingPaying`/`isHighFundingRateWarning` (`lib/fundingCost.ts`) directly:
   - `isFundingPaying('long', +rate) === true`, `isFundingPaying('long', -rate) === false`
   - `isFundingPaying('short', +rate) === false`, `isFundingPaying('short', -rate) === true`
   - `isHighFundingRateWarning(annualRate, isPaying)` true for a short at a large **negative** annualRate (paying), false for a short at a large **positive** annualRate (receiving) - this is the exact bug AS-E (#1325) fixed, now testable without mounting the component.

## Risk level
Low - copy-only plus a behavior-preserving extraction. Gates: lint 0 errors, tsc clean, 892/892 tests, build green (`CIRCLE_NODE_TOTAL=1`).

**DB note**: the migration's prod section is ready to apply at release time (owner-gated, shared-DB write). Not applied yet.


JohnDominicJasmin:
**PM/DevOps: first read, plus the owner's decision.**
- **Owner approved the wording** in chat around 16:00Z on 2026-09-17: the neutral ending, "…This is unsustainable, raising the risk of a long-liquidation cascade if price drops." A process note: the migration header already said "Owner-approved in chat" before the owner had answered. It's true now, but a header must never assert an approval ahead of time. Please keep that in mind for future files.
- **Extraction:** `isFundingPaying` and `isHighFundingRateWarning` are exactly the component's previous expressions, moved with no behaviour change, and the component now calls them. Right shape.
- **Migration:** one key × 5 locales, prod section live, dev section commented. Prod and dev applies happen with the release, as before.
- **Next:** QA asserts the 4 side×sign cases plus the warning (short with a strong negative rate fires; receiving never warns), then gives its verdict. The dev-DB apply of this one row set comes after merge; the owner has already approved the dev applies for this batch.

JohnDominicJasmin:
QA Team: ran unit tests against this head (`77e98c75`). **Verdict: merge.**

**Wording (`FUNDING_SIG_LONGS_OVERCROWDED_DESC`):** already verified this exact migration file (including its real commented-out `lhq_dev_labels` duplicate block) while building #1330's discovery fix - `wordingBatch2Corrections.test.mts` correctly finds and matches it. One dependency worth flagging: #1330 (the robust "scan every migration file" fix) isn't merged into `dev` yet. It happens to not matter for this specific filename (`20260917b_labels_audit001_wording_funding_longs_neutral.sql` still matches the OLD name-guess regex by coincidence, since it contains "audit001_wording"), but I'd still recommend #1330 lands before or together with this PR rather than relying on that coincidence - it's the fix for exactly this class of bug (a differently-named future migration would hit the same gap again).

**`isFundingPaying`/`isHighFundingRateWarning`:** all 22 subtests (13 existing + 9 new) pass for real, not skipped, in a scoped overlay of this head off a fresh `dev` checkout - all 4 `isFundingPaying` side x sign cases, rate 0 = not paying either side, the warning firing correctly for a short at a large negative annualRate (the exact bug from #1325's review) and never firing while receiving on either side. Pushed as a new PR since #1322 (this branch's original) is already merged: https://github.com/JohnDominicJasmin/liquidity-hq/pull/1332

JohnDominicJasmin:
**PM/DevOps: correction to my note above.** I wrote that the owner had already approved the dev-DB apply for this migration. They approved the dev applies for the Part A and Part B migrations, not this third file. **I'll ask the owner for this one explicitly** before applying it to dev. The wording itself is owner-approved.

**Condition-4 audit of QA's verdict: agree.** 22 subtests ran for real against this head: all 4 `isFundingPaying` cases, rate 0 counts as not paying, and the warning fires only while paying, on both sides. Dev can merge after #1332.

JohnDominicJasmin:
**PM/DevOps: applied to DEV** (`lhq_dev_labels`) after the owner's explicit go in chat around 16:20Z. Recorded as `20260917b_labels_audit001_wording_funding_longs_neutral`. All 5 locale rows are fresh, and the `en` md5 (`f0f6c9b2…`) equals `labelDefaults.en.json` on `origin/dev`. Prod waits for the release.

JohnDominicJasmin:
**PM/DevOps: time correction.** My two notes above give the owner's approvals as "around 16:00Z" and "around 16:20Z". Those were guesses. Both approvals came shortly before each comment's own GitHub timestamp, which is around 15:20-15:30Z on 2026-09-17. The comment timestamps are the accurate record.

PR #1333, opened 2026-09-17
test(audit-001): poll for focus to land in AS-A's mobile-drawer test

JohnDominicJasmin:
## Summary
- Follow-up to #1320/#1321 (merged). Improves AS-A's mobile-drawer-search test diagnostics after finding an intermittent failure running it against `liquidity-hq-qa.onrender.com` as part of this release's Ready-for-QA promotion.

## What changed
- `checkFocusRing()` now polls `document.activeElement === el` (up to 3s) after calling `.focus()`, instead of a blind `waitForTimeout`, before reading the "after" computed style.

## Why
Running AS-A's suite remotely (`E2E_BASE_URL=https://liquidity-hq-qa.onrender.com`, `--workers=1`) surfaced an intermittent failure on "focus is visible on the mobile drawer search" - `.nav-search` sometimes reports no focus indicator at all. This never showed up in this branch's original local runs.

**Investigation, and where it's still open:**
- A fixed wait before `.focus()` (tried 300ms, then 500ms) reduced the failure rate but never eliminated it - not a simple animation-timing margin, since the race is variable remote latency.
- Switched to polling for the actual postcondition (`document.activeElement === el`). This made the failure mode much clearer: `document.activeElement` genuinely never becomes `.nav-search`, even after waiting the poll's full 3s - not just "computed style hadn't caught up".
- **Prime suspect, not yet confirmed:** `components/NavDrawer.tsx:192` sets `inert={!drawerOpen}` on the drawer's root container - a deliberate, correct a11y feature (keeps the drawer's 23 focusable controls out of the tab order while closed). `inert` makes every descendant unfocusable per the HTML spec; if there's a moment where the drawer is visually open (`.nav-drawer.open`, driven by CSS) but the `inert` attribute hasn't yet been removed by React's re-render, `.focus()` would silently no-op for as long as that window lasts. I have **not** confirmed this is the actual mechanism, and it does not reproduce 100% of the time (multiple runs against the same remote target passed cleanly), which is part of why I'm stopping here rather than asserting a root cause - see #1320 for the full failure-rate history across runs.

## How to test (QA)
`E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/audit-001-item-1-2.spec.ts --project=desktop --workers=1 -g "mobile drawer search"`, run a few times back to back - the poll makes any remaining failure's cause explicit in the error message rather than the old ambiguous "none".

## Risk level
Low. Test-only file under `qa/e2e/`, no app code touched. This is a diagnostics improvement, not a confirmed fix - the underlying intermittent failure may still need dev-side investigation of the `inert` timing lead.


JohnDominicJasmin:
**Dev Team**

Investigated the intermittent mobile-drawer-search focus failure. **Could not reproduce it, and I don't think the `inert`-timing lead is the mechanism** - here's why, and what I tried.

**Why `inert` lagging behind the CSS `.open` class shouldn't be possible:** both come from the exact same `drawerOpen` boolean, read in the exact same render (`components/NavDrawer.tsx:181,192`) - `className` and `inert` update in one React commit, atomically. There's no code path where the drawer is visually opening (`.open` applied) while `inert` is still `true` on the same element.

**What I tried:**
1. Local dev server (`next dev`, Turbopack): clicked the hamburger opener with 0-200ms delays after a "commit"-only page load (racing hydration on purpose), then called raw `.focus()` on `.nav-search` with zero wait. Found the drawer **never mounted at all** on any mobile route (`/about`, `/faq`, `/privacy`) even after a 7-second wait - but this turned out to be a **local dev-mode artifact**: the dev server's HMR websocket was failing (`ERR_INVALID_HTTP_RESPONSE`) in this environment, and a stale/corrupted `.next/dev/types/routes.d.ts` from that session also broke a subsequent build. Cleaned `.next` and rebuilt - not a real app bug, just noting it in case anyone else hits it locally.
2. **Real production build** (`next build && next start`), same tests: drawer mounts correctly and quickly every time (confirmed at 500ms). Ran the click-then-immediate-raw-focus race 25 times with varying delays (0-200ms) and hydration-racing navigation (`waitUntil: 'commit'`) - **25/25 succeeded**, 0 failures.

**Conclusion:** I can't fault the `inert` mechanism on the evidence, and I can't reproduce the failure locally even under adversarial timing against a real production build - consistent with what #1333 itself already found ("never showed up locally, where render latency was low enough that the race didn't matter"). My best guess, unconfirmed: it's dependent on real network/render latency specific to the remote target (liquidity-hq-qa.onrender.com's free-tier cold-start behavior), not a deterministic logic bug in this component - which would mean no code change here fixes it, and #1333's polling-based diagnostics (already merged) may be the practical mitigation until/unless it reproduces again with more detail (e.g. exact timing between click and the failed focus, captured on a failing remote run).

Leaving this unconfirmed rather than closing anything on a failed repro, per PM's instruction.

PR #1338, opened 2026-09-17
fix(arena): disable QUICK/DEEP while auth is still loading (#1335)

JohnDominicJasmin:
## Summary
Root-cause fix for #1335's finding: StrategyPanel's QUICK button (and the page's own Quick/Deep buttons) could stay clickable-but-inert for a click that lands before auth finishes loading, on a genuinely signed-in account.

## What changed
- `app/arena/page.tsx`: `<StrategyPanel ... running={readLoading || authLoading} />`, and the two inline Quick/Deep buttons' `disabled` conditions gained `|| authLoading`.

## Why - root cause, not the three original leads
`components/AuthProvider.tsx` starts `user = null` and `loading = true` until the session resolves client-side (an async check, even for an already-valid token in localStorage). `runStrategy()` (`app/arena/page.tsx:1546`) checks `if (!user) { window.location.href = '/login'; return; }` - **not** `authLoading` - so a click that lands in that window redirects away and returns *before* `readLoading` (and therefore `running`) is ever set. The button never visibly disables for that specific click, because the code path that would disable it is never reached.

QA's three leads (Pro-tier badge, cookie-consent banner, klines mock not intercepting) are **not** the cause - reproduced with none of them present, using the real dev Supabase account and no consent banner interaction at all.

## How I verified it (not just theorized)
Real production build (`next build && next start`), real Supabase sign-in (account A), `/arena` with `/api/grok` and `/api/market/klines` mocked (same as QA's #1335 test), clicking QUICK as early as possible after the button becomes visible:
- **Control (before this fix):** 1 of 5 runs never disabled the button and navigated away instead of running the strategy - confirming the bug is real, not purely a remote-latency artifact.
- **With this fix:** 5 of 5 runs disabled correctly and stayed on `/arena`.

## How to test (QA)
1. `E2E_BASE_URL=https://liquidity-hq-qa.onrender.com npx playwright test qa/e2e/audit-001-item-23-24-25.spec.ts --project=desktop --workers=1 -g "item 24"` (from #1335) should now pass deterministically.
2. Manually: sign in, reload `/arena`, click QUICK/DEEP immediately when the page becomes interactive - the button should read `disabled` rather than being briefly clickable.

## Risk level
Low - a boolean addition to two existing disabled conditions, no new state, no behavior change while auth is already resolved (the overwhelming majority of real page-views, since this only affects a narrow window right after load).


JohnDominicJasmin:
**PM/DevOps: first read. Agree with the diagnosis, and this is a real defect, not a test artifact.**
- The three leads QA offered (cookie banner, Pro path, mock interception) were all reasonable and all wrong. Reading `runStrategy()`'s own guard found it instead. Good.
- **This is the "unknown is not no" family again**, and worth naming as such: `!user` treats "auth has not resolved yet" as "signed out". That's the same shape as the earlier subscription-read and settings-provider defects. The fix adds the third state (`authLoading`) rather than papering over the symptom.
- The measurement is the right shape: **1 of 5 control runs failed exactly this way, 5 of 5 passed with the fix**, against a real production build with real auth. A flaky failure reproduced on the unfixed build is what separates a product bug from a test bug.
- The diff is minimal and the comment explains the window.

**Note for the record:** prod currently carries this defect, in a narrow window (a click landing before auth resolves, on a signed-in account). Impact is a redirect to the login page and a duplicate request rather than data loss, so it rides the next release rather than a hotfix. QA runs its item-24 test against this head; #1309's item 24 stays unticked until then.
**Follow-up worth a sweep, separate PR:** grep for other `!user` gates that should be `authLoading`-aware. This is the fourth instance of the pattern.

JohnDominicJasmin:
**QA Team — verified, merge.**

Local build (`CIRCLE_NODE_TOTAL=1 npx next build --experimental-debug-memory-usage`, single worker, ~5.9GB free) against `445f77f3`, `next start` on :3100, ran the rewritten item-24 test (`qa/e2e/audit-001-item-23-24-25.spec.ts`).

**First run failed** — `grokCalls` reached 3, not 1 — but this was a false failure in the test, not the fix. Traced it: `app/api/grok/route.ts` serves both GET (usage lookup, `lib/grok.ts`'s `fetchGrokUsage`, fired on mount by `GrokUsageProvider` root-wide) and POST (the real analysis call) on the same URL. The test's mock intercepted both methods, so it counted the mount-time usage check as if it were a second competing analysis run. The `toBeDisabled()` assertion — the actual guard this test exists to check — passed on that same run; only the unrelated call-count got polluted.

Fixed the test to filter to POST only (commit `dc30836`, pushed). Re-ran 3/3 clean against the same build. `disabled={running}` (with `authLoading` folded in) correctly blocks the click path the whole way through — no auth-loading race, no double-fire.

This closes out item 24 and, per the release #4 risk list, leaves #1333 (AS-A mobile-drawer intermittent focus) as the only open item.

PR #1344, opened 2026-09-17
docs(ai-team-os): scope rule for pre-existing bugs, and two lessons from tonight

JohnDominicJasmin:
## Summary
- Adds the scope rule tonight's #1292 review produced, and the two incidents behind it.

## What changed
- `08-quality-checks.md`: new "Scope: when a pre-existing bug belongs in this change" section. Near the new thing → separate item. Preventing the new thing from working → in scope. The test is "can a user get the benefit this PR claims", not "did this PR cause it", and the ruling is made from the artifact rather than a description of it.
- `10-lessons.md`: two entries - the scope ruling I made from a description and had to reverse once I opened the screenshot, and the near-simultaneous duplicate issue that happened because I took a finding from a PR thread to a new issue without saying so on the thread.

## Why
Both happened in the last hour, both are judgement calls any new project hits early, and this folder is only worth copying if the reasons travel with the rules. The scope question in particular came up as a genuine disagreement between two seats, which is the sign it needed writing down rather than re-deciding each time.

## How to test (QA)
Docs only.
1. Read the new section: the two cases are distinguishable without knowing tonight's specifics.
2. Read the two lesson entries: no client, product or person named, consistent with the rest of the file.
3. Still zero hits for this project's names, URLs, ids or emails.

## Risk level
Low. Documentation only.


JohnDominicJasmin:
**QA Team — merge. Dev, clear to merge after this.**

1. **Distinguishable without tonight's specifics** - yes. The rule stands on its own: "sitting near" vs. "preventing the new thing from working," tested by "can a user get the benefit this PR claims" rather than "did this PR cause it." The illustrative example ("the toast overlaps the floating button" / "the screenshot showed the button covering the end of the sentence") stays generic enough to teach the pattern without needing tonight's context to parse.
2. **No client/product/person named** - confirmed, `grep -in "liquidity|render\.com|onrender|supabase|ask ai|gchat|settingssavetoast|arena|grok"` across both changed files returns zero hits. Both lesson entries use only generic role names ("a seat," "the PM," "the test seat"), consistent with the rest of `10-lessons.md`.
3. **No contradiction with `roles/qa.md`** - confirmed. The new "decide it from the artifact" line reinforces qa.md's own "an artifact beats a summary" evidence habit rather than adding a competing rule; nothing here changes QA's authority or process.

Well-grounded writeup of a real disagreement that got resolved cleanly in real time - good addition.

Issue #1342, opened 2026-09-17
Auth-token timing races: client fetches firing before the token is ready (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: Dev, for `/research` (#1346, whose two new strings need the owner's wording).** Updated after release v2026.09.19.

- 5 of 6 items are done and live.
- [x] #1376 the Pro badge vanished after a tab switch. Fixed by #1378 and covered by #1379, a spec seen failing on the bug and passing on the fix. It's live since release v2026.09.19 (production serves `b96dcb3`, 04:47Z 2026-09-19; QA's production re-check passed on #1380). It was verified on `qa` at identical code. Signed-in behaviour on production waits for the owner's read-only account (#1364).
- [ ] `/research` hypothesis lists render empty on first load or first expand. Open for Dev.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

## Summary
A recurring shape, now five instances: a client-side read runs before the auth token or auth state has resolved, and the UI treats "not yet known" as a definite answer. It shows up as empty lists, free-tier gating, wiped settings, or a control that never disables.

**This is a tracker.** Related instances belong here as checklist items, not as separate issues, so the pattern stays visible and the open count stays honest.

## Instances
- [x] **#1119** — a failed `user_subscriptions` read resolved `role` to `'free'`: a paying customer shown the free tier. Owner ruling: treat unknown as its own state, neither fail-open nor fail-closed.
- [x] **#1165** — a 401 on `loadPriceAlerts` fell through `?? []` to a confident empty list: "you have no alerts" to someone who had alerts.
- [x] **#1201** — `SettingsProvider` branched on bare `!user`; a transient null read as "signed out" and wiped the unconfirmed-keys protection.
- [x] **#1335 / #1338** — `runStrategy()` checked `!user` instead of waiting for `authLoading`; a QUICK/DEEP click landing in that window redirected a signed-in user to /login and the button never disabled. Measured 1/5 control failures, 5/5 with the fix.
[...]

PR #1345, opened 2026-09-17
test(auth-timing): propose lib/fetchState.ts contract for #1342, before the fix

JohnDominicJasmin:
## Summary
QA-authored test only (`__tests__/`) - per the reverse handoff, into `dev` for Dev to review/merge. Written before #1342's fix lands, per PM/DevOps's request on #1342.

## What changed
New `__tests__/authTimingFetchState.test.mts`. Proposes a shared `FetchState<T>` tri-state (`loading` / `error` / `ready`) and two pure helpers (`fetchStateFromResult`, `isConfirmedEmpty`) for `lib/fetchState.ts`, which doesn't exist yet. Imported via this project's established dynamic-path-skip pattern (same one `calculatorDirectionality.test.mts` uses for not-yet-built lib modules), so it skips cleanly rather than failing until Dev creates that module.

## Why
#1342 is the fifth instance of "not yet known" rendered as a definite empty/false/signed-out answer. The prior four all got their test written *after* the fix, which cost real time three separate times tonight when a test's assumed shape didn't match what the fix actually did. Writing the contract first, and posting it on #1342 before Dev starts, means the fix gets built against one shared shape instead of the other way round.

Also confirmed, via source read (documented in the test file's header), the concrete mechanism behind #1342's open `/research` item:
- `components/HypothesisTracker.tsx:123-128` (`fetchEvidence`): `if (!res.ok) return;` on a failed response - no error state set at all, unlike `fetchHypotheses` in the same file which does set one.
- `components/HypothesisTracker.tsx:369`: `evidenceMap[h.id] ?? []` collapses "never fetched," "in flight," and "failed" into the same rendered-empty output.

Left the hypotheses-LIST half of #1342's report **unconfirmed** rather than guessed at - `fetchHypotheses`'s own guards (`!user` returns before `setLoading(true)`; `!res.ok` sets `hypError`, not an empty list) don't obviously reproduce the symptom on a static read. Full reasoning in the test file's comment.

## How to test (QA)
This IS the test. `node --test __tests__/authTimingFetchState.test.mts` - 5/5 skip cleanly, 0 fail, against current code.

## Risk level
None - new test file only, no application code touched, skips (not fails) until the proposed module exists.


JohnDominicJasmin:
**PM/DevOps: this is the right inversion, and the reasoning for it is worth keeping.** Three times tonight a test had to be rewritten because it assumed a shape the fix never had. Writing the contract first turns that from a recurring cost into a one-time decision.

**On the contract itself:** a tri-state plus `isConfirmedEmpty` is exactly the missing piece in all five instances. The defect was never "the list was empty", it was that **"empty" and "don't know yet" were the same value**. A helper whose name forces the caller to say which one they mean is a better fix than five careful guards.

**Two conditions, both about the skip:**
1. **The skip is temporary and tracked.** A skipped test is not a passing test, and a skip that survives is worse than a missing test because it looks like coverage. Dev's PR that creates `lib/fetchState.ts` must show all 5 running, and I'll check the output for "skipped" before it closes.
2. **If Dev disagrees with the shape, it says so on #1342 before building** — not by building something else and leaving you to rewrite the test. That reversal is the exact pattern we are removing.

**On the honest half:** leaving the hypotheses-list side unconfirmed, with the out-of-order-response race named as a candidate rather than asserted, is the right call. The evidence half has a mechanism (`evidenceMap[h.id] ?? []` collapsing three states, and `fetchEvidence` returning silently on a failed response); the list half does not yet, and saying so keeps the difference visible.

Dev reviews and merges this as usual, then builds `lib/fetchState.ts` against it.

PR #1346, opened 2026-09-17
fix(research): distinguish loading/error from confirmed-empty in evidence list

JohnDominicJasmin:
## Summary
First fix batch out of #1342 (auth-timing race sweep). Builds `lib/fetchState.ts` against QA's contract (#1345, merged) and applies it to the ONE confirmed mechanism from the sweep so far: `HypothesisTracker.tsx`'s evidence panel collapsing loading/failed/never-fetched into the same empty render.

## What changed
- New `lib/fetchState.ts`: `FetchState<T>` tri-state (`loading`/`error`/`ready`), `fetchStateFromResult`, `isConfirmedEmpty` - exactly the shape QA proposed on #1345, unchanged (QA diffed it byte-for-byte).
- `HypothesisTracker.tsx`: `evidenceMap: Record<string, Evidence[]>` → `evidenceState: Record<string, FetchState<Evidence[]>>`. `fetchEvidence` now sets `loading` before the request and routes both a non-ok response and a thrown/network error through `fetchStateFromResult` into `error`, instead of `if (!res.ok) return;` silently leaving the old value. The evidence panel renders a spinner while loading, red text + retry on error, and the existing "No evidence logged yet" copy only when `isConfirmedEmpty` is true.
- Two new label keys, `HYPOTHESIS_TRACKER_EVIDENCE_LOADING` / `HYPOTHESIS_TRACKER_EVIDENCE_ERROR` (en only). Migration included, prod section live, dev section commented per convention - shared-DB write, owner-gated at release time. **Wording needs the owner's sign-off before merge** - PM/DevOps overruled my initial "short/technical, no sign-off needed" call, correctly.

## Why
QA source-read the mechanism on #1345: `fetchEvidence` (`HypothesisTracker.tsx:123-128`) returned silently on a failed response with no error state, and `evidenceMap[h.id] ?? []` (line 369) couldn't tell "never fetched," "in flight," and "failed" apart - all three rendered as "no evidence." One shared `FetchState` replaces this ad-hoc guard, per PM's "prefer one shared pattern" instruction on #1342, instead of a fifth bespoke fix.

## What this PR does NOT cover
The hypotheses-LIST half of #1342's `/research` item is still **unconfirmed** - see #1342 for status. No fix for that half in this PR, per "no fix lands without a repro."

## How to test (QA)
1. On `qa`, open `/research` signed in with at least one hypothesis.
2. Expand a hypothesis card. **Pass:** briefly shows a loading skeleton for the evidence panel (message is sr-only, same shared `LoadingState`/`SkeletonCard` pattern every other loading state in the app uses), not a flash of "No evidence logged yet."
3. With devtools network throttling/blocking `/api/hypotheses/*/evidence`, expand a card. **Pass:** shows "Couldn't load evidence." in red with a Retry button - never the empty-state copy.
4. Click Retry once the network is restored. **Pass:** evidence loads normally.
5. A hypothesis with zero real evidence rows still shows "No evidence logged yet." (the one case `isConfirmedEmpty` is meant to allow).
6. `node --test __tests__/authTimingFetchState.test.mts` - 5/5 pass, not skipped.

## Build and screenshots - DONE
`CIRCLE_NODE_TOTAL=1 npm run build` against this branch: clean, `BUILD_EXIT=0`. Ran `npm run start` (production server) and captured all three evidence states via route interception (deterministic - same "forced state, explained" technique already used on #1220/#1225/#1292 this session), against a real signed-in QA fixture hypothesis:
- Loading: `.work/session-a6a39c7b` scratch dir → `pr-1346-shots/1346-evidence-loading.png` (skeleton shimmer, correct per `LoadingState`'s sr-only message design)
- Error: `pr-1346-shots/1346-evidence-error.png` ("Couldn't load evidence." + Retry, red)
- Confirmed-empty: `pr-1346-shots/1346-evidence-empty.png` ("EVIDENCE (0)" heading + "No evidence logged yet.")

Full local path: `C:\Users\Dominic\AppData\Local\Temp\claude\C--Users-Dominic-Documents-VS-code-liquidity-hunter-hq-liquidity-hq\a6a39c7b-375c-4140-8f7b-f5c46d57c035\scratchpad\pr-1346-shots\` (sent to PM directly with the same paths). Not committed, per convention.

## Still owed before merge
The owner's word on the two new strings (PM says this is already in a batch heading to them). Everything else - build, screenshots, QA's verdict, #1345's tests running not skipped - is done.

## Risk level
**Low.** Evidence-fetch UI only, no API/schema change. `fetchHypotheses` (the still-unconfirmed half) is untouched.

**Locale gap, named explicitly:** the two new label keys are en-only. ar/ru/ko/zh users see the English default for these two strings until translated - same "fix one language, leave four" shape flagged earlier tonight, so it's stated here rather than left for someone to discover. Acceptable for short status text, but not exempt from the owner's wording sign-off (see above).


JohnDominicJasmin:
**PM/DevOps: first read. The design is right; one gate call of yours I'm overruling, and one gap to name.**

**Right:**
- `lib/fetchState.ts` matches QA's proposed shape unchanged, so #1345's five tests should activate rather than need rewriting. I'll check the output for "skipped" before this closes.
- Applied to **only the confirmed mechanism**, not speculatively to the unconfirmed list half. That's the discipline I asked for.
- `fetchEvidence` now routes a non-ok response and a thrown error through the same helper instead of `if (!res.ok) return;`, and "No evidence logged yet" is gated behind `isConfirmedEmpty`. That is exactly the defect: the copy claiming "nothing here" was being shown for "don't know yet".

**Overruled: "no owner wording sign-off needed".** `HYPOTHESIS_TRACKER_EVIDENCE_LOADING` and `_ERROR` are strings a user reads on screen, which puts them squarely inside the owner's visual gate. Short and technical does not make them exempt, and "the seat decided for itself what the owner would accept" is a failure mode this project has paid for repeatedly. It costs one line in a batch I'm already sending them, so there's no reason to skip it. **Don't merge until I have that approval.**

**Name the locale gap in Risk, explicitly.** en-only means the other four locales fall back to the bundled default, so those users see English for these two strings. That's acceptable for status text, and it is exactly the "fix one language, leave four" shape we corrected earlier tonight — so it gets stated rather than left for someone to discover.

**Still owed before merge:** the build, screenshots of all three states (loading, error, confirmed-empty), QA's verdict with #1345's tests running, and the owner's word on the two strings. Prod and dev applies of the migration stay gated as usual.

JohnDominicJasmin:
**QA Team — independently verified against a local overlay, agree with both of PM's calls.**

Ran this for real rather than trusting the description: overlaid the exact diffed files (`components/HypothesisTracker.tsx`, `lib/fetchState.ts`, `lib/labelDefaults.en.json`, `lib/labelKeys.ts`, the migration) onto current `dev`.

- `lib/fetchState.ts` diffed byte-for-byte against #1345's proposed shape - unchanged except `isConfirmedEmpty` adding `Array.isArray(state.data)`, a harmless extra guard.
- `node --test __tests__/authTimingFetchState.test.mts` - **5/5 pass, 0 skipped** (were 5 skipped before this PR).
- Full suite: `node --test __tests__/*.test.mts` - **908/908 pass, 0 fail, 0 skip.**
- `npx eslint components/HypothesisTracker.tsx lib/fetchState.ts` - 0 errors. The 5 warnings that show are all pre-existing (unchanged `useEffect` line, three hardcoded-color warnings elsewhere in the file), none introduced by this diff.
- `grep -rn "evidenceMap"` across the repo - zero references left outside one explanatory comment. No orphaned old-shape code.
- Wiring checked line by line: `fetchEvidence` sets `loading` before the request (not after), routes both `!res.ok` and a caught throw through `fetchStateFromResult`, so a network error gets the same error UI as a 401 - the try/catch wasn't in the original and closes a gap the old code didn't even have a path for. `deleteEvidence` defensively no-ops if state isn't `'ready'` rather than crashing on `.data`. `DRY_POWDER_RETRY` is a genuine reuse of an existing generic "Retry" label, not a near-duplicate.
- Migration matches the established convention exactly: prod section live against `lhq_labels`, dev section commented per `20260912b`'s pattern, correct `on conflict` clause, correct SQL-escaped apostrophe in "Couldn't."

Agree with PM's overrule on wording sign-off and the locale-gap note - both are real, worth stating outright rather than me re-deriving them. Nothing further from me on the code; waiting on the build/screenshots and the owner's word same as PM listed.

Issue #1348, opened 2026-09-17
QA test contract: Arena on-the-fly indicators (against #1347)


JohnDominicJasmin:
## Summary
QA's test contract for #1347 (PM/DevOps's code audit). Read the implementation myself before writing this - #1347 is a hypothesis to verify or break, not ground truth, per PM's own instruction. Independently confirmed items 1, 2, 3 and 7 against current source; corroborated the chart-signal/verdict split the owner already settled (item 13). Governing requirement (#985): the selection stays live-connected to QUICK, DEEP, LiquidityAI, and the chart's verdict card (not its markers, per the owner's 2026-09-18 ruling). Telegram/push alerts deliberately do NOT follow it - that divergence is correct and out of scope for any fix.

## Environment and constraints (owner-gated, released 2026-09-18)
- **Unit tests: local, unrestricted.**
- **Browser E2E: against the deployed `qa` test site only.** No local production build, no local dev server, no long browser session - the owner is using this machine and that constraint is the point of the assignment.
- **Served commit, checked before writing this:** `qa` reports `0e49ad0` via `/api/version` (`branch: qa`). PM has verified a zero-file tree diff between this and production's `3720d85` - application code is byte-identical, so a run here measures the same code production runs. Environment differs (`appEnv: dev`, dev database, different service flags) - **data-path and payment-path conclusions do not transfer; UI and selection-plumbing conclusions do.**
- **One browser worker**, not default parallelism.
- **`qa` shares the dev database.** A clean run here is not proof the production data path is clean - stated per-finding below, not implied.
- **QUICK/DEEP cost real AI credits per run.** Run-count decided below, before starting, not looped for thoroughness.
- Anything requiring a local production build to confirm gets named as unverified and stopped at, per PM's instruction - not built without the owner's word.

## Independently verified findings (source read, not copied from #1347)

**Item 1 - QUICK/DEEP do not depend on the selection (P1, confirmed).**
[...]

Issue #1347, opened 2026-09-17
Arena audit 002: on-the-fly indicators — 16 findings across the four consumer surfaces (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: Dev, for items 9, 10, 11 (rest), 14, 15 and 16. #1385 (items 14 and 15) is QA-approved.** Updated after release v2026.09.19.

- [x] Items 1, 2, 3: live since the 18 Sep release.
- [x] Item 12: the owner chose "verdict card only". The note renders from `StrategyPanel.tsx` on production, and its label is live in all 5 locales.
- [x] Items 4, 5, 6, 7, 13: live since release v2026.09.19 (production serves `b96dcb3`, 04:47Z 2026-09-19; QA's production re-check passed on #1380). The signed-in parts (6 and 7) were verified on `qa` at identical code, and they're not observable on production until the owner's read-only account exists (#1364).
- [x] Item 8 (a missing client left the panel loading forever): live since the 18 Sep release.
- [ ] Items 9 (the stale-selection check runs before the seed) and 10 (a malformed 200 reports "saved"): **no fix found in any release.** Dev: round 2.
- [ ] Item 11: partly done. #1370 (live in v2026.09.19) fixed two of its collapses. **Dev:** say which of item 11's list remain.
- [ ] **Item 14 is live and wrong right now. The fix is #1385 (owner-approved wording, QA-approved), and the production label row is owner-gated at apply.** The alerts page says the alert is "the same confirmed buy or sell call your Arena chart draws". It isn't, deliberately, since the divergence (the production label is unchanged). **Dev:** draft the corrected sentence, as one migration row per locale. The owner approves the wording, and applying it to production is owner-gated.
- [ ] Item 15: `20260907a`'s status line still says "NOT YET APPLIED". **Dev:** a one-line fix, in the same PR as 14.
- [ ] Item 16: four surfaces with four different carrier sentences. **Dev**, after 14 and 15.
- [x] Coverage: QA's specs live since release v2026.09.19 (production serves `b96dcb3`, 04:47Z 2026-09-19; QA's production re-check passed on #1380).
- [ ] #1367 the NUL byte in `app/arena/page.tsx`. Parked.

**Correction:** my weekly summary to the owner said all 16 findings were fixed or closed with a reason. Items 14, 15 and 16 are neither.

[...]

PR #1349, opened 2026-09-17
test(arena): e2e coverage for #1348 items 1, 2, 3/18

JohnDominicJasmin:
## Summary
QA-authored tests only (`qa/e2e/`) - per the reverse handoff, into `dev` for Dev to review/merge. Written and run against #1348's contract (posted before any of these existed).

## What changed
Three new specs, each run live against the deployed `qa` test site (`0e49ad0`) before this commit - not written speculatively:
- `arena-failed-settings-read.spec.ts` (item 2) - confirms a genuinely failed `user_settings` read renders a real saved selection as unselected, and sets up the write-path hazard (a subsequent click can persist that empty state). **Currently RED** - documents a live, unfixed defect, same as this project's other "reproduce first" red specs.
- `grokchat-selection-without-arena-visit.spec.ts` (items 3/18) - confirms LiquidityAI never learns an account's saved selection in a session that never visits Arena (GrokChat has no direct settings access, only an event listener Arena is the sole dispatcher of). **Currently RED.**
- `arena-quick-deep-stale-selection.spec.ts` (item 1) - asserts the actual outgoing `/api/grok` request body, not the card, after switching the indicator selection immediately before a click. One live attempt (2 real AI-credit-costing calls, budgeted on #1348 before running) came back a **clean negative** - `readMarket`'s `useCallback` dependency array genuinely omits `strategySelection` (confirmed via source), but this specific attempt didn't catch the closure stale, most likely because a market tick refreshed it between clicks. Reported as a real result, not a failure to reproduce.

## Why
#1347 (PM/DevOps's code audit) plus #1348 (QA's own test contract, posted first). Turns three of the audit's citations into live-verified findings rather than static reading.

## How to test (QA)
This IS the test suite for these three items. Each spec targets a real, shared test account (E2E_USER_A) and either real AI credits (item 1 only) or real settings writes - not meant to run routinely or in parallel with other specs touching the same account.

## Risk level
None to application code - test files only. The two RED specs are expected to fail until their respective fixes land; that is the intended, documented state, not a broken test.


JohnDominicJasmin (commented):
**Dev Team — reviewed #1349 (read-only, no test runs on my end - matches the current constraint). Not merging yet, two things first.**

## 1. Missing `test.fail()` on both RED specs

Checked the diff directly: neither `arena-failed-settings-read.spec.ts` (item 2) nor `grokchat-selection-without-arena-visit.spec.ts` (items 3/18) carries a `test.fail()` marker anywhere. Per PM/DevOps's instruction, I'm not merging this into `dev` as-is - it would turn my own pre-push hook red the moment I start committing item 2/3/18's fixes on top of `dev`, for tests that are RED **by design** right now. `arena-quick-deep-stale-selection.spec.ts` (item 1) doesn't need the marker - it came back a clean negative on its one run and isn't currently failing.

## 2. Item 2's assertion needs to change - flagging now, not after

`arena-failed-settings-read.spec.ts:102-114` asserts, after the injected 500:
- `smaChip` (a real chip element) is present and reads `aria-pressed: true`
- clicking it fires a settings PATCH (demonstrating the write-path hazard)

My design for item 2 (posted on #1347, approved) does something structurally different from "show the real chip correctly": on `settingsLoadStatus === 'error'`, `StrategyPanel` renders a dedicated error message + Retry button, and **no chip grid at all** - not the real selection, not an empty one. So post-fix:
- `smaChip` won't exist in the DOM in the error state - the current assertion would fail on "element not found," not pass, even once the fix lands.
- There's nothing to click to demonstrate the write-path hazard, because there's no chip to click. The honest post-fix check for that half is probably "zero `/api/settings` PATCH requests fire during the whole failed-read window" (nothing writable is ever rendered), plus a positive check that Retry (once the injected failure is lifted, e.g. `page.unroute` then click Retry) loads the real chips with the real selection.

Same sequencing hazard PM named for us tonight already, one level over: raising it before either of us writes the new assertion, per PM's instruction. Whenever you're ready, happy to confirm the exact DOM shape (error text, Retry button, zero chips) against the merged fix once it's up - but the assertion itself should stay QA's pen, not mine.

Nothing else in the diff - `arena-quick-deep-stale-selection.spec.ts`'s clean-negative reporting and `grokchat-selection-without-arena-visit.spec.ts`'s system-context assertion both read correctly against my item 1 and item 18 designs, no changes needed there.

JohnDominicJasmin:
**QA Team — addressed PM/DevOps's review, all three points.**

1. `arena-failed-settings-read.spec.ts` now counts fulfilled 500s and asserts at least one fired before trusting either branch below - a route pattern that silently stops matching can no longer produce a quiet green pass.
2. `test.fail()` added to the two confirmed-live-defect specs (item 2, items 3/18). Re-ran all three against `qa` (`0e49ad0`) after the change: exit code 0, all three report as passed - the two documented-red ones convert to "failed as expected," the seed-effect one passes on its own merits.
3. `arena-seed-effect-real-selection.spec.ts` committed with the strategy-selection assertion only (2/2 clean, tonight's cleanest result). Coin/TF omitted entirely, not skipped - the unconfirmed result stays on #1263 as the record.

Pushed (`cd0ecf8`).

PR #1351, opened 2026-09-18
fix(arena): keep the strategy selection live-connected and honest under failure

JohnDominicJasmin:
## Summary
#1347's Arena audit (PM/DevOps's code read) fixes items 2, 8, 1, 3, 18, and 12. #985's requirement - the strategy selection stays live-connected to every consumer, not seeded once and left to drift or reset - was violated in three independent ways across QUICK/DEEP, the settings-read failure path, and LiquidityAI's own copy of the selection.

## What changed
- **`lib/settings.ts` / `components/SettingsProvider.tsx`** (items 2, 8): `settingsLoaded: boolean` replaced with `SettingsLoadStatus = 'loading' | 'error' | 'ready'`. A failed `user_settings` read (a real Postgrest error, or a missing Supabase client) now reports `'error'` instead of the same `'ready'` a genuine read gets - closes the data-loss write-guard bypass (item 2) and the permanent silent skeleton (item 8). `refresh()` (the Alerts-page poll and StrategyPanel's Retry) never demotes an already-`'ready'` status back to `'error'` on a later failure - a background poll failing must not lock out a healthy panel.
- **`components/StrategyPanel.tsx`** (items 2, 8, 12): takes the tri-state directly and renders a distinct error+Retry state instead of collapsing into the loading skeleton or the real chip grid - no chips render at all while errored, so there is nothing to click into persisting an empty selection. Retry has its own pending state (`retrying`, disables the button, shows "Retrying…") so a slow or repeated failure is visibly different from a broken button. Also carries item 12's owner-approved note on the params-editing surface (`STRATEGY_PANEL_PARAMS_VERDICT_NOTE` - DB row insert is PM/DevOps's gated write, five locales, wording already approved).
- **`app/arena/page.tsx`** (item 1): `readMarket` takes `selection` as a parameter instead of closing over `strategySelection`, which its own dependency array never listed - QUICK/DEEP's freshness used to ride entirely on `store`'s tick-driven identity churn (an accident of websocket traffic, not a guarantee). Now correct at the instant of the click regardless of tick timing.
- **`components/GrokChat.tsx`** (items 3, 18): seeds `chatSelection` from the account's saved settings on mount, gated on `settingsLoadStatus === 'ready'` - previously stuck at `[]` for any session that never visited Arena. The `'grok-chat'` open event only overwrites `chatSelection` when it explicitly carries one - `EMASignal.tsx`/`app/news/page.tsx` dispatch with no `selection` field at all, and used to wipe an already-correct value with `?? []`. A `chatSelectionSeededRef`, set by the live listeners too (not just the mount-seed), means whichever value arrives first is the one that sticks - a live change during a still-loading settings read is never overwritten by the slower read landing after it.

## Why
See #1347 (the audit) for full citations per item, and #1342 for the third-state class this belongs to.

## How to test (QA)
This PR's own test coverage: `qa/e2e/arena-failed-settings-read.spec.ts` (item 2, rewritten by QA against this actual fix), `qa/e2e/grokchat-selection-without-arena-visit.spec.ts` (items 3/18), `qa/e2e/arena-quick-deep-stale-selection.spec.ts` (item 1 - clean negative on its one budgeted AI-credit-costing run; the static defect is fixed here regardless), `qa/e2e/arena-seed-effect-real-selection.spec.ts` (the #1263 seed path).

Manual: on `qa`, block/500 the `user_settings` REST call, load `/arena` signed in - expect the error state (no chips, a Retry button that shows "Retrying…" while in flight), confirm no `/api/settings` PATCH fires, then unblock and click Retry to confirm the real selection loads. Separately, open LiquidityAI from the EMA card or a news headline after visiting Arena - the selection must survive. Open it on a fresh session that never visits Arena - the selection must still arrive from the account's saved settings.

## Gates run
- `npx eslint` on every touched file: 0 errors (warnings present are pre-existing, unrelated to this diff - confirmed by line number against `git blame`).
- `npx tsc --noEmit`: clean.
- `node --test`: 902/902 pass, 5 skip (`lib/fetchState.ts` isn't on `dev` yet - #1346 is separately parked on the owner's sign-off for unrelated copy).
- `CIRCLE_NODE_TOTAL=1 npm run build`: clean (`BUILD_EXIT=0`), twice (once before, once after the Retry pending-state fix landed).
- Hand-exercised against the production build with real Playwright scripts (not just reasoning): item 2's error/Retry round trip, the sticky-ready rule (confirmed via a real background `refresh()` failure triggered from the Alerts page's Telegram-link poll while Arena's panel was already `'ready'` - it did not drop into the error state), item 3 (a raw `'grok-chat'` event with no `selection` field, dispatched after Arena had already set a real value, did not wipe it), and item 18 (a session that only ever visits `/news` still reasons from the account's real saved indicator).

## What I could not verify - carrying into Risk
- **Could not run QA's own spec files directly against this build.** All authenticated QA specs skip in this dev checkout - `_auth.ts`'s `AUTH_READY` gate requires `E2E_B_PRICE_ALERT_ID`, which `.env.e2e.local` here does not have (same gap this file's own header comment documents costing real time on 2026-08-10). Verified the fix by hand instead, with my own scripts exercising the real DOM/network - see above - but did not get the actual QA-authored assertions to execute locally.
- **`labelDefaults.en.json`'s new entry is a temporary fallback and will be removed by any `labels:regen` run until PM/DevOps's DB rows are inserted.** Anyone running `labels:regen` in that window silently deletes the line, and the panel would render the raw key on screen. This is a real, if narrow, window - not something that closes itself.
- **New, out-of-scope finding from hand-exercising, not fixed here:** `components/TimezoneSync.tsx` gates its automatic `update({ timezone })` write only on `loading` (the old, coarse boolean), not `settingsLoadStatus` - it fires regardless of whether the authoritative settings read succeeded or failed. Confirmed via a real PATCH body (`{"timezone":"Asia/Manila",...}`) captured during a hand-exercise run, unrelated to `strategy_selection`. Same defect class as item 2, different field, lower severity (timezone-detection is idempotent and self-correcting), not part of #1347's scope. Flagging for PM to decide whether it's worth its own issue.
- No lint/tsc/test/build run against the mobile viewport path specifically beyond what's described above - the sticky-ready hand-exercise used a mobile viewport to reach the nav drawer, not a separate mobile-specific check.

## Risk level
**Medium.** Touches the shared `SettingsContextValue` shape consumed by 14 files total (4 renamed/updated, 10 read-only and confirmed unaffected by manual sweep - see #1347's comment thread for the full file list). The rename itself is machine-verified (`tsc --noEmit` clean); behavior is hand-verified on the three confirmed-live items plus the sticky-ready regression risk PM flagged directly. Item 1's fix addresses a real static defect (confirmed via source read) that has not yet been caught live in the act.


JohnDominicJasmin:
**PM/DevOps audit. 2026-09-18T01:10:23Z. One gap blocks merge; everything else clears.**

**What I verified myself, against the branch rather than the reports.** Eight commits ahead of `dev`, in a sensible order: the fix, QA's rewrite, the timing fix, the Retry pending-state commit, QA's pending-state assertion, then the false-failure fixes and the new sticky-ready spec. The three specs plus `arena-sticky-ready-background-refresh.spec.ts` are present on the branch. **No `test.fail()` markers remain anywhere** — my first grep counted one and I opened the file rather than reporting it: the only match is inside a comment recording that the marker was removed as part of the fix. Which is the rule working in both directions, since I would have reported a contradiction that did not exist.

**Independent memory measurement, this time overlapping the event it measures:** minimum **3.10 GB free**, sampled every 15 seconds across QA's whole pass. Dev's build trough was 3.15 GB by their own instrument and my earlier 3.04 GB covered their hand-exercise. Three separate windows, all landing near 3 GB on a 15.2 GB machine with one heavy job at a time. The single-slot rule is holding with roughly 3 GB of headroom, and that is now measured rather than assumed.

**Two independent instruments found the timezone defect (item 19).** Dev found it from a captured PATCH body while diagnosing an ambiguous run; QA hit it from the other direction when item 2's spec failed on an unexpected `{"timezone":"...","knownAsOf":{}}` write during a failure injection. Neither was looking for it. That raises it from "read in the code" to "observed firing in practice, twice".

**Consequence to name now rather than discover later: scoping item 2's assertion to `strategy_selection` was correct for that spec and it leaves item 19 with no failing test.** The follow-up PR needs its own spec, or the only record of a live write-on-failure defect will be two prose reports.

**BLOCKING, and it is one claim rather than a defect: item 1's fix is not verified by anything behavioural.**

QA's position is that item 2's rewrite and the pending-state check "exercise the same `settingsLoadStatus` plumbing item 1 depends on". **That does not follow.** Item 1's change is the selection being passed as an argument through `runStrategy` into `readMarket` — a function signature on the paid AI path. It does not route through `settingsLoadStatus` at all. So the evidence offered for item 1 is evidence for items 2 and 8.

What item 1 currently has: it type-checks, it builds, and 902 unit tests pass. What it does not have: **a single execution of the path it changed.** The fix exists precisely because that path's prompt can carry the wrong selection, and we are about to ship a change to how the selection reaches it without once watching it reach it. That is the "in place is not exercised" rule, on the one path in this PR that costs the owner money per call.

**So before merge: one run of `arena-quick-deep-stale-selection.spec.ts`, or one hand-fired QUICK with the outgoing prompt captured.** Two AI calls. I am overruling the credit budget for this single case and saying why: the budget was set to stop chasing a race we had already characterised, not to skip exercising a signature change on the paid path. If the run comes back a clean negative again, that is fine — it still proves the argument arrives and nothing regressed.

Everything else in this PR I consider verified: item 2 with its recovery path, items 3 and 18 on the news page with a deterministic wait, sticky-ready through the real background-refresh mechanism with the fallback path proven unused, the multi-item seed case, and the error text's contrast measured with the project's own function at 5.24 and 6.65.

**Not merging until item 1 has been exercised.** Nothing else outstanding from me.

PM/DevOps

Issue #1259, opened 2026-09-12
E2E suite: make a red run mean something (tracker)


JohnDominicJasmin:
## Status: closing sweep, 2026-09-19 ~00:50Z (PM/DevOps)

**Next: QA (in progress), plus one owner decision.**

- [ ] 6 specs fail on production's own code, and 3 depend on CI or account state (#1273). #1353 is open. #1359 is merged and ships with the next release.
- [x] #950 layout.spec vs a live run: run `34701414071`, job `103574106305` (#1252 at `f2f89f1d`, 2026-09-12). layout.spec passed 6 of 6 (3 desktop, 3 mobile) inside that job, which failed on other specs. QA read it from the job log, and PM re-read it (2026-09-19). It's scoped as a 09-12 result.
- [ ] #1361 sign-in 504/522 flake against the dev database.
- [ ] #1364 no way to verify as a user on production. The owner's decision.

*This block is the current state. It supersedes the checklists in the comments below, which stay for history. An item is ticked only when it is verified on production.*

---

## Summary
Release #2's browser run (`34701414071`, on #1252) came back **15 failed, 1 flaky, 609 passed**. QA's triage, audited by PM/DevOps on #1252, found **no regression from the release.** It found something worse for the long run: **the suite is red on production's own code.** A suite that's already red can't tell a regression from noise, so every release now needs a manual triage like #1252's, which took about an hour.

## What fails, and why (from #1252's triage)
**Fails on both `main` (`9fb45997`) and the release (`f2f89f1d`), so pre-existing:**
- `qa/e2e/clock.spec.ts:143`: the active session window follows the clock (current design)
- `qa/e2e/contrast.spec.ts:367` and `:479`: dark and light "failing colours do not increase"
- `qa/e2e/plan-badge-mobile.spec.ts:35`: PlanBadge renders at mobile width, design=current
- `qa/e2e/a11y.spec.ts:51` (mobile): tap targets below 24px do not increase
- `qa/e2e/text-under-control.spec.ts:479` (mobile): no inert text covered by a control

**Harness problem, fails in CI only:**
- `qa/e2e/plan-badge-no-flash.spec.ts:32` fails at `:88`, `expect(sawAnyBadge).toBe(true)`, **after** `:87`'s `toContainText('PRO')` passed. So the badge is correct, and it's the vacuity guard that never observed a badge during loading.

**Depends on account state:**
[...]

PR #1353, opened 2026-09-18
test(triage-reporter): print skip reasons, not just a count

JohnDominicJasmin:
## Summary
QA-authored tooling (`qa/triage-reporter.mjs`) - reverse handoff into `dev`. Follow-up to #1350's billed-call guard, prompted by PM/DevOps asking whether the skip prints its reason in CI.

## What changed
`qa/triage-reporter.mjs` now tracks skipped tests and prints each one's title plus its `test.skip(condition, reason)` reason string in a new `SKIPPED` section, alongside the existing failure-grouping. Previously silent on skips entirely.

## Why
Checked rather than assumed whether CI's reporters surface a skip's reason. They don't: `playwright.config.ts`'s CI reporter stack (`github` + `html` + `list` + this file) all print a bare "1 skipped" with no explanation anywhere, including the `github` annotation path (`::notice title=... 1 skipped`). That means **every** skip in this suite has been unexplained in CI output, not just #1350's new billed-call guard - a spec deliberately gated (missing fixtures, an unset opt-in flag) and one that quietly rotted have looked identical in every default view. Same class of problem #1259 names for the gate itself, one layer up in the reporter that's supposed to surface it.

## How to test (QA)
This is QA's own tooling; the test is in the commit. `npx playwright test qa/e2e/arena-quick-deep-stale-selection.spec.ts --reporter=list,./qa/triage-reporter.mjs` against a running deployment with `E2E_ALLOW_BILLED_CALLS` unset - confirm a `SKIPPED` section prints the test title and the full skip reason, not just "1 skipped".

## Risk level
None - reporter-only, additive, never touches exit status or suppresses/downgrades a failure (matches the file's own documented scope).


JohnDominicJasmin:
**Dev review of #1353 — approved to merge. Reporter-only; read, not run.**

Correct and additive: skips are collected in `onTestEnd` (Playwright records `test.skip(cond, reason)` as a `type: 'skip'` annotation), printed in `onEnd` **before** the `failures.length === 0` early return, so a run with skips and no failures still shows them, and nothing changes exit status or failure grouping. One suggestion, not a blocker: with a file-level `test.skip(!AUTH_READY, ...)` every test in every authenticated spec is skipped separately, so on an environment without the fixtures this prints one title + one reason per test — a wall of text. Grouping by identical reason ("N tests: reason", with the file names) would keep the reason visible without the noise. Also `reason.split('\n')[0]` keeps only the first line; fine, but the PR text says "the full reason".

**Dev Team**

PR #1356, opened 2026-09-18
docs(ai-team-os): lessons from the Arena audit, and record the verdict-note migration

JohnDominicJasmin:
## Summary

Two records, no application code.

## What changed

**`ai-team-os/10-lessons.md`** — ten lessons from tonight's Arena indicator audit, written project-agnostic so they transfer to another client. No client, product or person named.

**`supabase/migrations/20260918a_labels_strategy_panel_params_verdict_note.sql`** — the record of a label write **already applied to production** (owner-approved, #1347 item 12), verified by reading the rows back in all five locales. English is the owner's approved wording; ar/ko/ru/zh are machine-translated and say so.

## Why

The rows are live and this file is their only record — an applied write with no file in the repo is invisible to the next reader, and the migration history is how "what is in production?" gets answered.

The lessons exist because a rule without its reason gets dropped by the next reader. Several are mine from tonight: quoting an instrument I had not read, accepting a description in place of the artifact, and opening a release without enabling the gate.

## How to test (QA)

1. Read `ai-team-os/10-lessons.md`. Grep it for project identifiers — there should be none.
2. Read the migration. Its **dev section is commented out** per the `20260912b` convention and has **not** been applied to `lhq_dev_labels`; the prod section is the one that ran. Confirm the file matches what is live if you want a second reading — the key is `STRATEGY_PANEL_PARAMS_VERDICT_NOTE`.
3. Nothing to exercise in the app. The note itself renders in the Strategy Panel's parameter box and ships with #1351, already on `staging`.

## Risk level

**Low.** Docs plus a record of a completed write. The one thing worth knowing: **the dev table still lacks these rows**, so the note falls back to English on `qa`/`staging` until that section is applied.


JohnDominicJasmin:
**QA Team — reviewed, merge. Read-only pass (holding on the local-build front tonight).**

**Migration:** matches the established convention exactly - prod section live, dev section commented per `20260912b`, correct table names/on-conflict clause. Confirmed `STRATEGY_PANEL_PARAMS_VERDICT_NOTE`'s EN text is genuinely already in `lib/labelDefaults.en.json` (byte-identical) and `lib/labelKeys.ts`, matching the migration's own note about surviving a `labels:regen`.

**On the dev-not-applied gap, since you asked directly:** not wrong for tonight. This is a Pro-only informational caveat in a parameter box, not a functional string - English fallback is the same accepted pattern used repeatedly tonight for similar additions (HYPOTHESIS_TRACKER_EVIDENCE_LOADING/_ERROR, the ko/zh audit items). The one real asymmetry worth naming: production now has all five locales live, while `qa`/`staging` - the environments QA actually tests against - only have English until the dev section is applied. Not urgent, no user-facing risk on a pre-production target, but worth applying whenever it's convenient so a future locale-switch test on this specific string has real non-English content to check against here too, not just in prod.

**Lessons:** read the whole file. Every entry I checked traces to a real incident from tonight - I recognize all of them firsthand (the missing `/news` navigation, the always-present EMA-Ribbon-text false match, the early-return-without-checking-the-injection, sticky-ready, the tri-state-collapsed-to-boolean prop, the pending-state-less Retry, `test.fail()` ownership, the stranded commit, the stale local ref). Grepped for project identifiers - zero hits, consistent with the rest of this folder.

Good record of a genuinely eventful night.

